PR #201, opened 2016-04-20
Cleanups noticed during investigation of cloning code

lgolding:
While spending a few minutes figuring out how much the SARIF SDK depends on the generated cloning code, I found a few places where it did, but didn't need to. The resulting changes were a nice cleanup.

@michaelcfanning


lgolding:
@michaelcfanning This shouldn't cause you any merge conflicts with your Stacks work.

michaelcfanning:
+1 ship it

PR #203, opened 2016-04-20
Fix #176, fix #202

### LINKED ISSUE #176: BuildAndTest doesn't clear Autogenerated directory
When we change a class name, say ToolInfo => Tool, the old and the new files ToolInfo.cs and Tool.cs both end up in the Autogenerated directory. This causes a problem since the .targets file includes everything in the Autogenerated directory. The .targets file has a line that's supposed to clear that directory but it doesn't seem to be happening.

### LINKED ISSUE #202: NotYetAutogenerated feature of build is not working
This is the feature that allows you to make a copy of one of the generated files in the NotYetAutoGenerated directory, and make manual changes to it for features that JSchema's code gen doesn't (yet) support. @michaelcfanning reports that it's not working.

lgolding:
I was deleting files at the wrong time (after object model had been built). Frankly don't know why it was working at all.

176: BuildAndTest doesn't clear AutoGenerated directory
202: NotYetAutogenerated feature of build is not working


src/Sarif/Microsoft.Json.Schema.ToDotNet.targets
```diff
@@ -16,6 +16,7 @@
   <!-- This target runs only when we actually need to build the object model files, which should be
        only when the JSON schema changes. -->
   <Target Name="BuildObjectModel" Inputs="@(JsonSchemaFile)" Outputs="$(_ObjectModelOutputDirectory)\*.cs">
+    <RemoveDir Directories="$(_ObjectModelOutputDirectory)" />
```
lgolding: Just moving the "Delete" task here wasn't enough. I had to remove the entire directory to get it to work. Don't know why.

lgolding:
@michaelcfanning Please review

michaelcfanning:
nice fix and thank you
+1

PR #205, opened 2016-04-21
Stacks

michaelcfanning:


src/Sarif.Driver/Sdk/Errors.cs
```diff
@@ -238,31 +238,34 @@ public static void LogExceptionLoadingPlugIn(string plugInFilePath, IAnalysisCon
         public static void LogUnhandledRuleExceptionAssessingTargetApplicability(
```
michaelcfanning: We use the new stacks functionality to preserve stack details for issues that might require active investigation by a tool owner. We do not emit stack details for exceptions that indicate configuration errors.

src/Sarif.UnitTests/Core/StackTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Linq;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    [TestClass]
+    public class StackTests
```
michaelcfanning: Currently, the tests and implementation ensure that stack representations match formatting of .NET StackTrace instances and Exception.StackTrace string values. After this basic verification, we could consider adding additional formatting (.NET doesn't show offsets in its frames, for example).

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -223,6 +225,33 @@ private void WriteRunToErrorList(Run runLog)
             }
         }
 
+        private IEnumerable<IEnumerable<AnnotatedCodeLocation>> CreateAnnotationsFromStacks(IList<Stack> stacks)
```
michaelcfanning: Note, the viewer is currently broken for results that don't populate the Location object (which is permitted by spec). That fix coming in a second change.

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+        private const string IN = " in ";
+        private const string AT = "   at ";
+        private const string LINE = ":line";
+
+        /// <summary>
+        /// Creates a SARIF Stack instance from a .NET StackTrace instance
+        /// </summary>
+        /// <param name="stackTrace"></param>
+        /// <returns></returns>
+        public static Stack Create(StackTrace stackTrace)
```
michaelcfanning: Could have used constructors for these instead of factory pattern. Any comments?

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            StringBuilder sb = new StringBuilder(255);
+
+            for (int iFrameIndex = 0; iFrameIndex < this.Frames.Count; iFrameIndex++)
+            {
+                StackFrame sf = this.Frames[iFrameIndex];
+
+                if (iFrameIndex > 0) { sb.AppendLine(); }
+
+                sb.Append(sf.ToString());
```
michaelcfanning: I put the StrackTrace formatting of StackFrames into the StackFrame ToString() override. .NET doesn't do this, though, it formats frames specifically for stacks and then provides a more comprehensive ToString() for StackFrame (that is really only suitable for debugging due to failure to handle missing data conditions in a readable way).

src/Sarif/Core/StackFrame.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        /// Creates a SARIF StackFrame instance from a .NET StackFrame instance
+        /// </summary>
+        /// <param name="stackTrace"></param>
+        /// <returns></returns>
+        public static StackFrame Create(System.Diagnostics.StackFrame dotNetStackFrame)
+        {
+            // This value is -1 if not present
+            string fileName = dotNetStackFrame.GetFileName();
```
michaelcfanning: Nick, would be interested in talking to you about some style considerations when you have a chance. Such whether a consistent style around locals declaration, init patterns, etc. makes sense.

michaelcfanning:
@lgolding @nguerrera 
Nick, would appreciate your eyes on this one, on style considerations as well as content.

src/Sarif.Driver/Sdk/Errors.cs
lgolding: That makes sense. It should be a code comment :-)

src/Sarif/Core/Stack.cs
nguerrera: I'd use constructors here. Some good reasons for going with factories are: (1) to return subclasses, (2) to reuse shared immutable instances, (3) To give more descriptive names to the methods.

None of those apply and I'm not seeing a fourth reason that would.

---
In reply to: [60623108](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60623108) [](ancestors = 60623108)

michaelcfanning: Two other reasons (only one of which apply), if you want to produce a TryXXX pattern (and therefore match it with a potentially exception raising Create). Second reason (which prompted me here), we turn a single exception into multiple stack instances (one for the root exception and all inner exceptions we encounter).

Glad to switch to ctors if you don't find that single scenario compelling to drive a general pattern.

nguerrera: The last one is interesting, but I don't think seeing var stack = Stack.Create(ex) vs.  var stack = new Stack(ex) would give me much of a hint about it. Something like Stack.CreateChain perhaps might do that. I don't have a very strong feeling here.

---
In reply to: [60625329](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60625329) [](ancestors = 60625329)

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+            return stack;
+        }
+
+        public static IEnumerable<Stack> Create(Exception exception)
+        {
+            List<Stack> stacks;
+            Queue<Exception> exceptions;
+
+            stacks = new List<Stack>();
+            exceptions = new Queue<Exception>(new Exception[] { exception });
```
nguerrera: I probably would have gone for a recursive implementation since N=number of inner exceptions is generally quite small and it would use the stack rather than extra allocations on the heap for bookkeeping.  To what extent are we reviewing for small efficiencies in this library?

src/Sarif.Driver/Sdk/Errors.cs
```diff
@@ -238,31 +238,34 @@ public static void LogExceptionLoadingPlugIn(string plugInFilePath, IAnalysisCon
...
+            Exception exception)
         {
             string ruleId = context.Rule.Id;
             string ruleName = context.Rule.Name;
             context.Rule = Errors.RuleDisabled;
 
+            Result result = RuleUtilities.BuildResult(ResultKind.InternalError, context, null,
+                                nameof(SdkResources.ERR0998_ExceptionInCanAnalyze),
+                                ruleName,
+                                exception.ToString());
```
lgolding: Doesn't `exception.ToString()` give you the whole stack as well as the message? Would `exception.Message` be better here?

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+                Stack stack;
+                Exception current;
+                
+                current = exceptions.Dequeue();
+
+                if (current.InnerException != null)
+                {
+                    exceptions.Enqueue(current.InnerException);
+                }
+
```
nguerrera: We probably want to special case AggregateException and its multiple InnerExceptions.

src/Sarif/Core/Stack.cs
michaelcfanning: A good question and one that will come up again when you examine the StackTrace line parsing code (which creates many strings rather than parsing the line character by character as could be done).

I favored readability, assuming as you allude to that these small inefficiencies aren't worth tuning for, given the relative rarity of these code paths. Otoh, if these are easily unit-tested methods (as they are) and readability is not at great risk (it probably is not), why not author the most efficient implementation possible? Glad to hear your thoughts and Larry's.

src/Sarif/Core/Stack.cs
michaelcfanning: Yes, nice catch, thanks. I'm expecting a rewrite of this code after collecting feedback, will file an issue.

src/Sarif.Driver/Sdk/Errors.cs
michaelcfanning: Yes, I think it makes sense not to recapitulate this information if we're producing a stack. Will fix.

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+                stack = Create(current.StackTrace);
+
+                stack.Message = current.Message;
+
```
nguerrera: I strongly recommend capturing exception.GetType() at every level. I often find just the message (either because it's in a language I don't understand [;)] or just not very clearly worded) unhelpful.

src/Sarif.UnitTests/Core/StackTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        [TestMethod]
+        public void Stack_CreateFromStackTrace()
+        {
+            var dotNetStack = new StackTrace();
+            Stack stack = Stack.Create(dotNetStack);
+
+            // The .NET StackTrace.ToString() override must preserve a trailing NewLine
+            // for compatibility reasons. We do not retain this behavior in ToString()
+            // but provide a facility for adding the trailing NewLine
+            Assert.AreEqual(dotNetStack.ToString(), stack.ToString(StackFormat.TrailingNewLine));
```
lgolding: I suggest we use FluentAssertions in new tests.

lgolding: Does this facility exist for any other reason than to make this test pass? If not, I suggest
```
dotnetStack.ToString().Should().Be(stack.ToString() + "\r\n");
```

src/Sarif/Core/Stack.cs
michaelcfanning: A nice improvement. Will add this to the second round where I improve/refine the output (will add a knob to retain formatting that is consistent with .NET).

src/Sarif.UnitTests/Core/StackTests.cs
michaelcfanning: I'm fine with fluent assertions. Nick? The ToString formatting facility will be built out to add other knobs, including whether to express the Module, add offsets, etc. If that wasn't in plan, I think your suggestion makes sense.

src/Sarif.UnitTests/Core/StackTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            catch (ArgumentException exception)
+            {
+                Exception containerException = new InvalidOperationException("test exception", exception);
+
+                IEnumerable<Stack> stacks = Stack.Create(containerException);
+                Assert.AreEqual(2, stacks.Count());
+
+                Assert.AreEqual(exception.StackTrace, stacks.ElementAt(1).ToString());
+                Assert.AreEqual(containerException.StackTrace, stacks.First().ToString());
+                caughtException = true;
```
lgolding: If you wrote
```
Stack[] stacks = Stack.Create(...).ToArray();
```
you could just write
```
stacks[0].Should().Be(containerException.StackTrace);
stacks[1].Should().Be(exception.StackTrace);
```

michaelcfanning: nice, will do

src/Sarif.UnitTests/Core/StackTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            try
+            {
+                File.Create(Path.GetInvalidFileNameChars()[0].ToString(), 0);
+            }
+            catch (ArgumentException exception)
+            {
+                Exception containerException = new InvalidOperationException("test exception", exception);
+
+                IEnumerable<Stack> stacks = Stack.Create(containerException);
+                Assert.AreEqual(2, stacks.Count());
```
lgolding: I've read ahead to your discussion about `Create` _vs._ ctor. It never would have occurred to me that a method named `Stack.Create` would return an `IEnumerable<Stack>`. I agree with Nick (I think he's the one that said it) that a name like `CreateChain` would be better -- except that it doesn't create a "chain" it creates Stacks, so I suggest `CreateStacks`.

michaelcfanning: I propose to switch to constructors everywhere except this case. Will use CreateStacks as you suggest here.

lgolding: Oh, one more thing: I think it should return `Stack[]`. You're always going to have to "realize" the `IEnumerable` with `ToArray` or `ToList` anyway, so returning `IEnumerable` doesn't save you anything. And there's nothing in the design guidelines against methods returning arrays (just properties).

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -163,9 +163,11 @@ private void WriteRunToErrorList(Run runLog)
                         HelpLink = rule?.HelpUri?.ToString()                        
                     };
 
-                    CaptureAnnotatedCodeLocationCollections(result.Stacks, AnnotatedCodeLocationKind.Stack, sarifError);
+                    IEnumerable<IEnumerable<AnnotatedCodeLocation>> stackLocations = CreateAnnotationsFromStacks(result.Stacks);
+
```
lgolding: Again I suggest returning `AnnotatedCodeLocation[][]`.

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+                stacks.Add(stack);
+            }
+
+            return stacks;
+        }
+
+        public static Stack Create(string stackTrace)
+        {
```
nguerrera: So the non-.NET-specific stack creation: e.g. building up from list of plain SARIF locations is not shown here because it is generated from schema, right? I'd need to look at the full API as it shows in IntelliSense, but maybe this method would benefit from a better name (though I can't come up with anything pretty -- CreatFromDotNetString? yuck)  to leave room for helpers from other well-known formats from other environments? Or maybe have a StackFormat input arg and we can have StackFormat.DotNet?

I also wonder if we shouldn't be careful about adding signifcant dependencies directly as members of the core schema OM. e.g. Should helpers like this go in a different assembly from the pure OM? That would be my BCL instinct, but maybe it's not worth the complexity here. Also, do we have a task to make the SARIF SDK a portable class library? It will definitely need to work in .NET Core for x-plat tools and it's better to do early than late when you may have taken too many dependencies. (I thought for a minute that StackTrace had been gutted too much for this in .NET Core, which prompted this line of thinking, but looking at corefx repo now and I think this is OK.  Still, it would be better to let the compilation handle future portability rather than just thinking about it.)

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                stack.Message = current.Message;
+
+                stacks.Add(stack);
+            }
+
+            return stacks;
+        }
+
+        public static Stack Create(string stackTrace)
```
nguerrera: No doc comment?

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
lgolding: Below you use a method named `CaptureAnnotatedCodeLocationCollections`. I think this one should be named `CreateAnnotatedCodeLocationCollections`. (and I don't think you need "FromStacks" since the argument tells you that).

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -163,9 +163,11 @@ private void WriteRunToErrorList(Run runLog)
...
+                    CaptureAnnotatedCodeLocationCollections(stackLocations, AnnotatedCodeLocationKind.Stack, sarifError);
                     CaptureAnnotatedCodeLocationCollections(result.CodeFlows, AnnotatedCodeLocationKind.CodeFlow, sarifError);
-                     CaptureAnnotatedCodeLocations(result.RelatedLocations, AnnotatedCodeLocationKind.Stack, sarifError);
+                    CaptureAnnotatedCodeLocations(result.RelatedLocations, AnnotatedCodeLocationKind.Stack, sarifError);
```
lgolding: Ah yes, I see -- when we introduce the intermediate `codeFlow` object we've discussed, you'll need to have an overload of `CreateAnnotatedCodeLocationsCollections` that takes `result.CodeFlows`.

src/Sarif/Core/StackFrame.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+            sb.Append(methodBase.Name);
+
+            // deal with the generic portion of the method
+            if (methodBase is MethodInfo && ((MethodInfo)methodBase).IsGenericMethod)
+            {
+                Type[] typars = ((MethodInfo)methodBase).GetGenericArguments();
+                sb.Append("[");
+                int k = 0;
+                bool fFirstTyParam = true;
```
nguerrera: I know this is inherited code from elsewhere, but let's ditch the Hungarian notation. I'd also get rid of the abbreviations. typeParameter is not so long.

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -223,6 +225,33 @@ private void WriteRunToErrorList(Run runLog)
...
 
+        private IEnumerable<IEnumerable<AnnotatedCodeLocation>> CreateAnnotationsFromStacks(IList<Stack> stacks)
+        {
+            List<List<AnnotatedCodeLocation>> codeLocationCollections = new List<List<AnnotatedCodeLocation>>();
+
+            foreach (Stack stack in stacks)
+            {
+                if (stack.Frames == null) { continue; }
+
+                List<AnnotatedCodeLocation> codeLocations = new List<AnnotatedCodeLocation>();
```
lgolding: Good place for `var`

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -223,6 +225,33 @@ private void WriteRunToErrorList(Run runLog)
...
+
+                foreach (StackFrame stackFrame in stack.Frames)
+                {
+                    codeLocations.Add(new AnnotatedCodeLocation()
```
lgolding: Nit: the parens on a default ctor call aren't needed when you use property initializer syntax.

src/Sarif/Autogenerated/Result.cs
```diff
@@ -572,7 +549,7 @@ private ISarifNode DeepCloneCore()
             return new Result(this);
         }
 
-        private void Init(string ruleId, ResultKind kind, string fullMessage, string shortMessage, FormattedMessage formattedMessage, IEnumerable<Location> locations, string toolFingerprint, IEnumerable<IEnumerable<AnnotatedCodeLocation>> stacks, IEnumerable<IEnumerable<AnnotatedCodeLocation>> codeFlows, IEnumerable<AnnotatedCodeLocation> relatedLocations, bool isSuppressedInSource, IEnumerable<Fix> fixes, IDictionary<string, string> properties, IEnumerable<string> tags)
+        private void Init(string ruleId, ResultKind kind, string fullMessage, string shortMessage, FormattedMessage formattedMessage, IEnumerable<Location> locations, string toolFingerprint, IEnumerable<Stack> stacks, IEnumerable<IEnumerable<AnnotatedCodeLocation>> codeFlows, IEnumerable<AnnotatedCodeLocation> relatedLocations, bool isSuppressedInSource, IEnumerable<Fix> fixes, IDictionary<string, string> properties, IEnumerable<string> tags)
```
lgolding: ... and won't it be nice when we can say `IEnumerable<CodeFlow> codeFlows`!

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+        Default,
+        TrailingNewLine
+    }
+
+    /// <summary>
+    /// A call stack relevant to a SARIF result.
+    /// </summary>
+    public partial class Stack
+    {
+        private const string IN = " in ";
```
lgolding: I know the design guidelines only apply to public API surface, but... "screaming caps"? Would `In`, `At`, `Line` be ok?

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+        private const string AT = "   at ";
+        private const string LINE = ":line";
+
+        /// <summary>
+        /// Creates a SARIF Stack instance from a .NET StackTrace instance
+        /// </summary>
+        /// <param name="stackTrace"></param>
+        /// <returns></returns>
```
lgolding: Fill in or remove.

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+        public static Stack Create(StackTrace stackTrace)
+        {
+            Stack stack = new Stack();
+
+            if (stackTrace.FrameCount == 0) { return stack; }
```
lgolding: Is it necessary to special-case the empty stack trace?

```
Stack stack = new Stack
{
    Frames = new StackFrame[stackTrace.FrameCount];
};
```

src/Sarif.UnitTests/Core/StackTests.cs
nguerrera: I personally like Assert.AreEqual(x, y) better than x.Should().Be(y).  I guess it's fine for unit tests, but I've seen a lot of crazy code backing fluent API and they get my head spinning around "what does Should()" do exactly?".  This line of code is roughly if (!Object.Equals(x, y)) throw new AssertException(); I personally don't see the value in making that into an English sentence. I can read code. ;)



---
In reply to: [60628907](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60628907) [](ancestors = 60628907)

src/Sarif/Core/Stack.cs
lgolding: I agree with Nick that recursion would be a little more readable; as to perf, do we know that this implementation is actually faster?

My tendency is to author the most readable code possible and optimize it if/when perf issues arise.

My 2c.

src/Sarif/Core/Stack.cs
lgolding: Hah, good point. If it weren't for that, you wouldn't need a `Queue` at all, just a single `nextExceptionToProcess`.

src/Sarif.UnitTests/Core/StackTests.cs
nguerrera: Oh I missed that this has a different return type so can't be constructor. I imagined before reading in detail that since we were talking about Create overloads that they all returned Stacks and a Stack must be chained to another for that to capture more than one. CreateStacks works for me. There is some precedent for CreateRange in immutable collections, but it's not exactly the same since it is CollectionType.CreateRange and not CollectionElementType.CreateRange.

---
In reply to: [60630251](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60630251) [](ancestors = 60630251)

src/Sarif.UnitTests/Core/StackTests.cs
nguerrera: I won't push on this if you guys both like FluentAssertions better, though.

---
In reply to: [60634645](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60634645) [](ancestors = 60634645,60628907)

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            stack.Frames = new List<StackFrame>();
+
+            foreach (string line in stackTrace.Split('\n'))
+            {
+                // at Type.Method() in File.cs : line X\r
+                string current = line;
+
+                StackFrame stackFrame;
+                stackFrame = new StackFrame();
```
lgolding: Nit: Consider
```
var stackFrame = new StackFrame();
```

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                // at Type.Method() in File.cs: line X
+                current = line.Trim('\r');
```
lgolding: Any reason not to trim all white space? `current = line.Trim();`

src/Sarif/Core/Stack.cs
nguerrera: I was actually implying (from intuition not measurement) that recursion would be faster (or probably about equal with less GC pressure).

---
In reply to: [60634687](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60634687) [](ancestors = 60634687)

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                // Type.Method() in File.cs: line X
+                current = current.Substring(AT.Length);
+                stackFrame.FullyQualifiedLogicalName = current.Split(')')[0] + ")";
+
```
lgolding: Is the intent here to change `C.F(string)` to `C.F()`? If so, why? I mean, if we're eliding the argument list (which I'm not sure I agree with), why not `C.F`?

src/Sarif.UnitTests/Core/StackTests.cs
michaelcfanning: Larry, in re: using an [], I will stick in this case with IEnumerable. The reason is that the method itself doesn't know how many stacks it will return and so uses a List internally to build up the return value. I don't think it makes sense to take the cost upfront of allocating another return value when the list itself can be returned. It's not clear that consumers would need to convert to an array (a simple enumeration over the values makes sense). If it came to it, the helper could simply return the list, there's no real benefit to the helper to return a fixed array.

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
michaelcfanning: yes

src/Sarif.UnitTests/Core/StackTests.cs
michaelcfanning: I see places in code where this enumerable is converted to a list as you say. I have just made the helper return IList<Stack>.

lgolding: Yes, `IList` is good. Basically anything except `IEnumerable`, because when I see that, if I'm going to access it more than once, the first thing I do is call `ToList` or `ToArray` (depending on how I'm feeling that day) on it, to make sure the collection only has to be "realized" once. If the helper returns `IList` then I know I don't have to do that.

src/Sarif/Core/Stack.cs
lgolding: I filed #206 to make it a portable class library.

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+                    current = current.Substring(IN.Length);
+
+                    // :line X
+                    fileName = current.Split(' ')[0];
+                    fileName = fileName.Substring(0, fileName.Length - LINE.Length);
+
+                    // X
+                    current = current.Substring(fileName.Length + LINE.Length);
+                    lineNumber = int.Parse(current);
+
```
lgolding: I think regex would make this much shorter and (IMO) more understandable:

```C#
Match match = Regex.Match(line, @"^at ([^)]+\)) in ([^:]+)(: line (.*))?");
if (match.Success)
{
    stackFrame.FullyQualifiedLogicalName = match.Groups[1];

    if (match.Groups.Count > 2)
    {
        string fileName = match.Groups[2];
        int lineNumber = int.Parse(Groups[4]);

        stackFrame.Location = ... // as before
    }
}
```

src/Sarif/Core/Stack.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        public override string ToString()
+        {
+            return ToString(StackFormat.Default);
+        }
+
+        public string ToString(StackFormat stackFormat)
+        {           
+            if (this.Frames == null) { return null; }
```
lgolding: IMO it's dangerous from `ToString` to return null. I've never seen anyone null-check the value returned form `ToString`. I suggest returning `string.Empty`.

src/Sarif/Core/StackFrame.cs
```diff
@@ -0,0 +1,160 @@
...
+                string lineNumber = Location.Region.StartLine.ToString(CultureInfo.InvariantCulture);
+                string fileName = Location.Uri.LocalPath;
+
+                result += IN + fileName + LINE + " " + lineNumber;
+            }
+
+            return result;
+        }
+
+        public static string CreateFullyQualifiedName(MethodBase methodBase)
```
nguerrera: Should this be public? StackFrame doesn't feel like the right place for it.

src/Sarif/Core/StackFrame.cs
```diff
@@ -0,0 +1,160 @@
...
+            if (fileName != null)
+            {
+                physicalLocation = new PhysicalLocation()
+                {
+                    Uri = new Uri(fileName),
+                    Region = new Region()
+                    {
+                        StartLine = dotNetStackFrame.GetFileLineNumber(),
+                        StartColumn = dotNetStackFrame.GetFileColumnNumber(),
+                        Offset = ilOffset
```
lgolding: Here we are again :-) We decided to get rid of `byteOffset` and `charOffset`, and replace them with just `Offset`, which -- depending on whether the `Region` is a "text region" (as determined by whether `StartLine` is > 0) -- is interpreted either as characters or bytes. So in this case, `Offset` would be interpreted as a character offset.

I'm more than happy to resurrect `charOffset` and `byteOffset`. This is the first scenario I've seen where a byte offset coexists with line and column information.

From the spec:

> When a region object represents a binary region, it shall contain a property named offset which shall have an integer value equal to the byte offset from the beginning of the file of the first byte in the region.

> When a region object represents a text region, the offset property may be present. In this case, it represents the character offset from the beginning of the file of the first character in the region

src/Sarif/Core/Stack.cs
nguerrera: I don't get the need for a knob to retain consistency with .NET. Exception.ToString() has the exception type at every level.

---
In reply to: [60628469](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60628469) [](ancestors = 60628469)

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -21,8 +21,99 @@
   },
   "required": [ "version", "runs" ],
   "definitions": {
+    "stack": {
+      "description": "A call stack that is relevant to a result.",
+      "additionalProperties": false,
+      "type": "object",
+      "properties": {
+        "message": {
+          "description": "A message relevant to this call stack, if applicable.",
```
lgolding: No need for "if applicable"

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -21,8 +21,99 @@
...
+          "type": "string"
+        },
+        "frames": {
+          "description": "An array of stack frames that represent a sequence of call, rendered in reverse chronological order, that comprise the call stack.",
```
lgolding: "a sequence of calls"

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -21,8 +21,99 @@
...
+          "type": "array",
+          "minItems": 1,
+          "items": {
+            "$ref": "#/definitions/stackFrame"
+          }
+        },
+        "properties": {
+          "description": "Key/value pairs that provide additional details about the code location.",
```
lgolding: "about the stack frame"

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -21,8 +21,99 @@
...
+          "type": "object",
+          "default": { }
+        },
+        "tags": {
+          "description": "A unique set of strings that provide additional information for the code location.",
```
lgolding: "for the stack frame"

src/Sarif/Core/StackFrame.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                physicalLocation = new PhysicalLocation()
+                {
+                    Region = new Region() { Offset = ilOffset }
+                };
+            }
+
+            StackFrame stackFrame = new StackFrame()
+            {
+                FullyQualifiedLogicalName = fullyQualifiedName,
```
lgolding: The schema includes a `LogicalLocationKey` property, but you don't populate it here -- and I don't even know how you'd figure it out!

src/Sarif/Core/StackFrame.cs
lgolding: (although in this case the name would be something like `isFirstTypeParameter`. Still works for me.)

src/Sarif/Core/StackFrame.cs
lgolding: Happy to have that conversation. I think you know my starting position on this :-)

src/Sarif/Core/StackFrame.cs
lgolding: I agree that _if_ it were public, it doesn't belong here. More likely an extension method on `MethodBase`. But I don't think it needs to be public.

src/Sarif/Core/Stack.cs
nguerrera: Even string.Empty is discouraged (but still much better than null). See notes to inheritors: [https://msdn.microsoft.com/en-us/library/system.object.tostring%28v=vs.110%29.aspx.](https://msdn.microsoft.com/en-us/library/system.object.tostring%28v=vs.110%29.aspx.) 

---
In reply to: [60651545](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60651545) [](ancestors = 60651545)

lgolding: TIL! Yes, there is is in black and white: "Your ToString override should not return String.Empty or a null string."

src/Sarif/Core/StackFrame.cs
nguerrera: I don't know that the question is here. This line looks fine. ;)

---
In reply to: [60653352](https://github.com/Microsoft/sarif-sdk/pull/205#discussion_r60653352) [](ancestors = 60653352)

lgolding:
:+1: , modulo all the little nits, the issue with `ToString` returning `null`, and the problem with packing line/column info and a byte offset into the same `region`.

src/Sarif.UnitTests/Core/StackTests.cs
michaelcfanning: I never saw an issue with the Assert api myself. Larry has the strongest feelings on it, I think.

lgolding: There's nothing wrong with the Assert API _per se_. It's just that FluentAssertions has two features which I consider compelling. One is objective, the other subjective:

1. Objective: FluentAssertions has a far larger set of assertions than the Assert API. Take a look at its [wiki page](https://github.com/dennisdoomen/fluentassertions/wiki). To really appreciate it, keep scrolling down. And down. It's amazing how much good stuff is in there.

2. Subjective: Reasonable people can disagree whether `stacks.Count.Should().Be(2)` is easier to read than `Assert.AreEqual(2, stacks.Count)`. But when you get down to something like this, IMO the case for the fluent syntax becomes much more compelling:

```C#
Action act = () => subject.Foo(null)); 
act
  .ShouldThrow<ArgumentNullException>()
  .WithMessage("?did*");
```

My 2c

src/Sarif/Core/Stack.cs
michaelcfanning: These are good thoughts. The constructors and factory methods live in partial classes integrated with the code emit. You could imagine going with a decoupled factory type to help keep unwanted type coupling at bay.

src/Sarif/Core/Stack.cs
michaelcfanning: because I believe I have a precise sense of what is in hand and don't want to trim anything else off. probably, I should split on Environment.NewLine, though, not '\n' and could therefore get rid of this. will do.

src/Sarif/Core/Stack.cs
lgolding: Oh, I see, the intent is to get everything up to and including the closing paren. Below I suggested regex for this whole exercise.

michaelcfanning: No, that's the right parenthesis we're parsing on (ends the params list) and then we add it again. Preserves args, as you say

src/Sarif/Core/Stack.cs
michaelcfanning: sure, more readable

src/Sarif/Core/Stack.cs
michaelcfanning: IN screams at you? :) using all caps is a convention most teams I've been on have used to distinguish these private const values from properties.

lgolding: Was just reading Framework Design Guidelines, 2nd Ed., p. 38, quoting Brad Abrams:

> To facilitate these [naming] debates, we coined a number of terms. ... We used the term SCREAMING_CAPS to indicate an all-uppercase style.

I don't object, just hadn't noticed it in the code base before.

src/Sarif/Core/Stack.cs
michaelcfanning: fixed

src/Sarif/Core/StackFrame.cs
michaelcfanning: We only require a subset of members. Path of least resistance is to hoist them into StackFrame and eliminate physical location use.

lgolding: Agreed, xclnt

src/Sarif.UnitTests/Core/StackTests.cs
michaelcfanning: I am ripe for convincing. However, fluent assertions fail in one basic case, they convert Environment.NewLine to escaped char sequences (to render output on a single line). This tanks easy comparison of stack ToString(). So I reverted to Assert in this case. I have other examples.

IOW while the readability is interesting and powerful, the experience grokking failures when you examine output text in VS is inferior to MSTEST API. in my experience to date.

src/Sarif/Core/StackFrame.cs
michaelcfanning: fixed

src/Sarif/Core/StackFrame.cs
michaelcfanning: Right. It's too late! Now we have to parse the fully qualified name created by the Exception/StackTrace classes. :) That's the justification for providing a structured way to document the components of a logical name. I am punting on recreating this. An exercise for the future.

src/Sarif/Core/StackFrame.cs
michaelcfanning: closed down visibility.

src/Sarif/Core/StackFrame.cs
michaelcfanning: I added braces, should have cleaned up names as well. Done now.

src/Sarif/Schemata/Sarif.schema.json
michaelcfanning: fixed here and in 4 other places.

src/Sarif/Schemata/Sarif.schema.json
michaelcfanning: yes

src/Sarif/Schemata/Sarif.schema.json
michaelcfanning: 'about the stack', yes, fixed

src/Sarif/Schemata/Sarif.schema.json
michaelcfanning: fixed

src/Sarif.UnitTests/Core/StackTests.cs
lgolding: Yes, that's definitely a problem. What I've taken to doing is copying the output to an editor window (where, as you say, you see "foo\r\nbar\r\n" and so on). The I CTRL-H and, with "regex" turned on in the search and replace window, I replace "\\r\\n" with "\r\n". Yes, it's a pain.

PR #207, opened 2016-04-22
Support arrays with uniqueItems

lgolding:
@michaelcfanning @rtaket @nguerrera


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -14,6 +14,7 @@
       "description": "The set of runs contained in this log file.",
       "type": "array",
       "minItems": 1,
+      "uniqueItems": true,
```
michaelcfanning: I don't think you can say this, given the preponderance of optional values. Unless a run contains start time and end time, you could easily imagine two runs with the same result (empty, for example).

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -241,6 +243,7 @@
         "fileChanges": {
           "description": "A message relevant to this annotation.",
           "type": "array",
+          "uniqueItems": true,
```
michaelcfanning: this is wrong. file changes needs to be a list because ordering is important, changes need to be applied in a precise sequence. also, the changes themselves could be duplicated.

'If a single fileChange object (§6.20) specifies more than one replacement, then the effect of the replacements shall be as if they were performed in the order they appear in the replacements array (§6.20.3).'

michaelcfanning:
We keep adding the converter dll to the project and editing it. That's bad, there shouldn't be a checked in binary. I added a gitignore flag that i hope clears this up. Can you under your change to that file and delete it?

src/Sarif/Schemata/Sarif.schema.json
lgolding: No, it's the `replacement` objects that must be in an ordered list, not the `fileChange` objects.

src/Sarif.Driver.UnitTests/AnalyzeCommandTests.cs
```diff
@@ -11,6 +11,7 @@
 
 using Xunit;
 using System.Collections.Generic;
+using System.Linq;
```
michaelcfanning: namespaces in this file need a sorting

src/Sarif.Driver.UnitTests/AnalyzeCommandTests.cs
```diff
@@ -463,7 +464,7 @@ public Run AnalyzeFile(string fileName)
                 Assert.NotNull(log);
                 Assert.Equal<int>(1, log.Runs.Count);
 
-                run = log.Runs[0];
+                run = log.Runs.First();
```
michaelcfanning: bad change, see my other comment. runs needs to remain an array

lgolding:
I thought I already had done a git -rm on the converted dll and pushed the change.

src/Sarif/Schemata/Sarif.schema.json
lgolding: What does a log with two completely empty runs tell anybody? I think the schema _should_ insist, and the SDK enforce, that there is some way to tell one run from another. startTime/endTime would be a fine way to do it.

src/Sarif.Driver.UnitTests/AnalyzeCommandTests.cs
lgolding: Fixed.

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
```diff
@@ -215,8 +215,12 @@ private IEnumerable<IEnumerable<AnnotatedCodeLocation>> CreateAnnotationsFromSta
                         Message = stackFrame.ToString(),
                         PhysicalLocation = new PhysicalLocation
                         {
-                            Uri = stackFrame.Location.Uri,
-                            Region = stackFrame.Location.Region
```
michaelcfanning: Oops. Sorry. We need a CI system.

michaelcfanning:
If you agree with my comments on reverting back to array for a small # of items, you are good to go. +1 really great to get this one done. in the event, the breaking change wasn't as intrusive as I expected.

src/Sarif.Viewer.VisualStudio/ErrorListService.cs
lgolding: :)

src/Sarif/Schemata/Sarif.schema.json
lgolding: Per our discussion, agreed. Run's can't be completely empty. They always have at least a tool name. So it's useful to say "I ran FxCop twice" (and there were no results).

PR #208, opened 2016-04-22
Make isSuppressedInSource a bool?. When null, this indicates the log …

michaelcfanning:
…file is not attempting to provide an answer re: whether the result is suppressed or not. @lgolding 

This nullable value is the most important one. I will scan others. We don't need this for Region as we have clear guidance on how to interpret values there (because 0 is invalid for StartColumn). 

Nullable type would probably be useful for the stack frame offset


.gitignore
```diff
@@ -215,3 +215,4 @@ GeneratedArtifacts/
 _Pvt_Extensions/
 ModelManifest.xml
 
+PREfastXmlSarifConverter.dll
```
lgolding: Hah. Good change. I wish the build didn't put it there.

src/Sarif/Autogenerated/Stack.cs
```diff
@@ -0,0 +1,293 @@
+// Copyright (c) Microsoft.  All Rights Reserved.
```
lgolding: Please don't check this in. My PR already has it.

src/Sarif/Autogenerated/StackFrame.cs
```diff
@@ -0,0 +1,434 @@
+// Copyright (c) Microsoft.  All Rights Reserved.
```
lgolding: Likewise.

src/Sarif/NotYetAutogenerated/Result.cs
```diff
@@ -0,0 +1,685 @@
...
+        /// A grouped set of locations and messages, if available, that represent code areas that are related to this result.
+        /// </summary>
+        [DataMember(Name = "relatedLocations", IsRequired = false, EmitDefaultValue = false)]
+        public IList<AnnotatedCodeLocation> RelatedLocations { get; set; }
+
+        /// <summary>
+        /// A flag indicating whether or not this result was suppressed in source code.
+        /// </summary>
+        [DataMember(Name = "isSuppressedInSource", IsRequired = false, EmitDefaultValue = false)]
+        public bool? IsSuppressedInSource { get; set; }
```
lgolding: Right, this is the "hand tweak", got it.

lgolding:
:+1:

PR #213, opened 2016-04-23
Sarif driver package incorrectly refers to SARIF sdk.

michaelcfanning:
@lgolding 

Investigating BinSkim issue and noticed this problem. Hopeful this resolves the strange dependency collisions we have observed.


src/Nuget/Sarif.Driver.nuspec
```diff
@@ -15,7 +15,7 @@
     <tags>command line driver utilities static analysis sarif framework utils driver.utilities</tags>
     <dependencies>
       <group>
-        <dependency id="Sarif.Sdk" version="1.4.$patch$-beta" />
+        <dependency id="Sarif.Sdk" version="$major$.$minor$.$patch$-beta" />
```
lgolding: Oh, my!

lgolding:
:+1:

PR #214, opened 2016-04-24
Update to JSchema 0.15.0 to avoid arbitrary code execution

lgolding:
As originally designed, the code hint mechanism allowed the user to specify an arbitrary `CodeGenHint`-derived type in the hints file. This was convenient because it allowed JSchema to use Json.NET to directly deserialize the hints, but it potentially allowed for arbitrary code execution as the type was being deserialized.

So JSchema now uses a mechanism where there is a fixed set of hint types. The hints file contains an enumeration value to specify the hint type, and a property bag used to initialize the hint object. JSchema has explicit code to initialize each type of hint from the property bag.

In this change, we upgrade the SARIF SDK to use this new version of JSchema, and modify the hints file accordingly.

See lgolding/jschema#52

@michaelcfanning


src/Sarif/CodeGenHints.json
```diff
@@ -1,156 +1,166 @@
 {
   "Rule.Options": [
     {
-      "$type": "Microsoft.Json.Schema.ToDotNet.DictionaryHint, Microsoft.Json.Schema.ToDotNet"
+      "kind": "DictionaryHint"
     }
   ],
   "AnnotatedCodeLocation.Properties": [
     {
-      "$type": "Microsoft.Json.Schema.ToDotNet.DictionaryHint, Microsoft.Json.Schema.ToDotNet"
```
lgolding: Don't let Json.NET deserialize an arbitrary type.

src/Sarif/CodeGenHints.json
```diff
@@ -1,156 +1,166 @@
...
+      "kind": "DictionaryHint"
```
lgolding: Instead, specify one of a fixed number of supported hints.

src/Sarif/CodeGenHints.json
```diff
@@ -1,156 +1,166 @@
...
+      "kind": "DictionaryHint" 
     }
   ],
   "file": [
     {
-      "$type": "Microsoft.Json.Schema.ToDotNet.ClassNameHint, Microsoft.Json.Schema.ToDotNet",
-      "className": "FileData"
+      "kind": "ClassNameHint",
+      "arguments": {
+        "className": "FileData"
```
lgolding: The rest of the changes involve wrapping the other properties of the hint object -- which Json.NET would previously have directly deserialized into the specified type -- in a property bag named `arguments`. JSchema constructs each type "by hand" from the contents of the property bag.

src/Sarif/CodeGenHints.json
```diff
@@ -1,156 +1,166 @@
...
-      "zeroValue": "Unknown",
-      "enum": [
-        "OneZeroZeroBetaThree"
-      ]
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "SarifVersion",
+        "description": "Possible values for the SARIF schema version.",
+        "zeroValue": "Unknown",
+        "enumValues": [
```
lgolding: Oh, and also, I changed the name of the JSchema property `EnumHint.Enum` to `EnumHint.EnumValues`.

michaelcfanning:
+1

PR #216, opened 2016-04-25
Remove extraneous file and package references.

michaelcfanning:
@lgolding More erroneous stuff hanging around, now removed.


lgolding:
:+1:

PR #217, opened 2016-04-25
Optimistic push of dictionary rewriting visitor fix. Update to 1.5.12

michaelcfanning:
@lgolding fyi, pushed your fix into SDK in order to permit easier downstream tool testing. will follow up in mail.


lgolding:
:+1:

PR #219, opened 2016-04-25
Update to JSchema 0.16.0 to fix dictionary traversal bug

lgolding:
The JSchema-generated `SarifRewritingVisitor` class had a bug where it was modifying the elements of dictionaries as it looped over the dictionary's keys. JSchema 0.16.0 fixes this by making a copy of the keys and looping over the copy.

Also:
* Add a unit test for `SarifRewritingVisitor`.

@michaelcfanning


SetCurrentVersion.cmd
```diff
@@ -1,7 +1,7 @@
 set MAJOR_PREVIOUS=1
 set MINOR_PREVIOUS=5
 set PATCH_PREVIOUS=12
-set PRERELEASE_PREVIOUS=-beta
+set PRERELEASE_PREVIOUS=-developer
```
michaelcfanning: We should never release a -developer version. Did that happen? If not, this needs to remain 'beta'

michaelcfanning: +1 after reverting the PRERELEASE_PREVIOUS change.

lgolding: Just a mistake. I didn't understand the semantics of your file. Fixed.

michaelcfanning: The _PREVIOUS data is consumed by the helper script that delists the last published version from Nuget. So, on releasing, you update the _PREVIOUS with the last shipped verion. And introduce -developer to the current version to mark it as a working version. Before release, -developer is changed to -beta. Run the script to delist previous. Run the script to publish current. Then repeat as above.

PR #220, opened 2016-04-25
proposed schema changes for invocation object

michaelcfanning:
@rtaket @lgolding 

Please sanity check this schema definition for the invocation object (basically resurrected runInfo). 

Larry, note that this object is where we collect sensitive information about the tool run, another advantage of having a distinct object. For processing systems that want to sanitize logs, this entire node can get zapped.

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -188,7 +188,7 @@
       "required": [ "value", "algorithm" ]
     },
     "invocation": {
-      "description": "The runtime environment of the analysis too run.",
+      "description": "The runtime environment of the analysis tool run.",
       "additionalProperties": false,
       "type": "object",
       "properties": {
@@ -215,14 +215,14 @@
           "type": "string"
         },
 
-        "architecture": {
-          "description": "The hardware architecture for which the run was targeted.",
+        "machine": {
+          "description": "The machine that hosted the analysis tool run.",
           "type": "string"
         },
 
         "processId": {
           "description": "The process identifier for the analysis tool run.",
-          "type": "string"
+          "type": "integer"
         },
 
         "workingDirectory": {
@@ -619,7 +619,7 @@
           "additionalProperties": {
             "$ref": "#/definitions/rule"
           }
-        },
+        }
       },
       "required": [ "tool" ]
     },
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -610,8 +610,8 @@
         },
 
         "invocation": {
-          "description": "A string containing the runtime parameters with which the tool was invoked. For command line tools, this string may consist of the completely specified command line used to invoke the tool.",
-          "type": "string"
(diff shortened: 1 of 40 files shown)


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -692,56 +705,75 @@
...
         }
       },
-      "required": [ "tool" ]
+      "required": [ "name" ]
     },
-    "tool": {
-      "description": "The analysis tool that was run.",
+
+    "invocation": {
+      "description": "The runtime environment of the analysis too run.",
```
lgolding: "too" => "tool"

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -692,56 +705,75 @@
...
           "type": "string"
         },
 
-        "fileVersion": {
-          "description": "The binary version of the tool's primary executable file (for operating systems such as Windows that provide that information).",
+        "workingDirectory": {
+          "description": "The working directory for the analysis rool run.",
           "type": "string",
-          "pattern": "[0-9]+(\\.[0-9]+){3}"
+          "default": { }
```
lgolding: No need for a default, but if there were one, it would have to be a `string` since that's the `type` of this property (so would be "`""`".

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -692,56 +705,75 @@
...
           "items": {
             "type": "string",
             "default": [ ]
           }
-        }
-      },
-      "required": [ "name" ]
-    }
+        },
+        "required": { [ "name" ] }
```
lgolding: This change is wrong... not sure how it happened (since it doesn't look like something you'd've done by hand). The value of `required` must be an array of string: `"required": [ "name" ]`.

But... what is this doing here at all? `invocation` doesn't have a property called "`name`".

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -692,56 +705,75 @@
...
           "type": "object",
           "default": { }
         },
 
+        "properties": {
+          "description": "Key/value pairs that provide additional information about the run.",
+          "type": "object",
+          "default": { }
+        },       
+
```
lgolding: You'll need to add a code gen hint:

```json
  "Invocation.Properties": [
    {
      "kind": "DictionaryHint"
    }
  ],
```

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -692,56 +705,75 @@
...
-          "description": "The tool version.",
+        "architecture": {
+          "description": "The hardware architecture for which the run was targeted.",
           "type": "string"
         },
 
-        "semanticVersion": {
-          "description": "The tool version rendered as Semantic Versioning 2.0.",
+        "processId": {
+          "description": "The process identifier for the analysis tool run.",
```
lgolding: I'd've expected this to be `"integer"`, not `"string"`.

lgolding:
Looks basically right.

PR #222, opened 2016-04-26
Invocation

michaelcfanning:
@lgolding @rtaket @nguerrera


src/Sarif/Autogenerated/SarifNodeKind.cs
```diff
@@ -84,6 +80,14 @@ public enum SarifNodeKind
         /// </summary>
         Run,
         /// <summary>
+        /// A value indicating that the <see cref="ISarifNode" /> object is of type <see cref="Stack" />.
+        /// </summary>
+        Stack,
```
michaelcfanning: I noticed the schema previously alphabetized these items and that I'd compromised sort order, so corrected.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -287,6 +187,73 @@
...
+          "type": "string",
+          "format": "date-time"
+        },
+
+        "correlationId": {
+          "description": "An identifier that allows the run to be correlated with other artifacts produced by a larger automation process.",
+          "type": "string"
+        },
+
+        "machine": {
```
michaelcfanning: machine is expressed in an environment variable, if that dictionary is present, on windows machines. does this warrant its own top level property? what about current user account name (also expressed in env vars)?

src/Sarif.Driver/Sdk/SarifLogger.cs
```diff
@@ -64,20 +63,35 @@ public class SarifLogger : IDisposable, IAnalysisLogger
...
             }
 
-            string invocation = Environment.CommandLine;
+
+            run.Invocation = Invocation.Create();
 
             if (invocationTokensToRedact != null)
             {
-                foreach (string tokenToRedact in invocationTokensToRedact)
+                run.Invocation.Machine = Redact(run.Invocation.Machine, invocationTokensToRedact);
```
lgolding: I don't understand. If we consider machine name to be sensitive information, then why fill it in at all?

src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/SimpleBundleDoubleNesting.sarif
```diff
@@ -320,61 +320,45 @@
...
-                "region": {
-                  "startLine": 1
-                }
-              }
-            }
-          ]
-        },
-        {
-          "ruleId": "JS2073",
+          "ruleId": "JS2074",
```
lgolding: I don't understand these differences. Is this some sort of result-ordering inconsistency?

src/Sarif.UnitTests/ResultLogObjectWriter.cs
```diff
@@ -43,9 +43,8 @@ public void WriteTool(Tool tool)
         }
 
 
-        public void WriteRunProperties(string invocation, DateTime startTime, DateTime endTime, string correlationId, string architecture)
+        public void WriteInvocation(Invocation invocation)
```
lgolding: I'm surprised this method is empty.

src/Sarif/Autogenerated/SarifNodeKind.cs
lgolding: Good, I'd considered doing that the last time I was in here.

src/Sarif/Schemata/Sarif.schema.json
lgolding: I didn't even notice "machine" in the original proposal, but... if we consider something important enough to be a first-class member of `invocation`, the decision shouldn't depend on whether that thing happens to exist as an environment variable on a particular operating system.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -287,6 +187,73 @@
...
+          "default": { }
+        },
+
+        "tags": {
+          "description": "A set of distinct strings that provide additional information about the run.",
+          "type": "array",
+          "uniqueItems": true,
+          "items": {
+            "type": "string",
+            "default": [ ]
```
lgolding: this "default" belongs on "tags", not under tags.items. It is the default value for the tags property, not the default value for each of its items. (The items are strings, not arrays.)

lgolding: (I corrected this everywhere else in my last PR)

src/Sarif.UnitTests/ResultLogObjectWriter.cs
lgolding: D'oh. This is ResultLogObjectWriter, a unit test class. Never mind.

lgolding:
:+1: modulo the question about machine name, and the problem with "tags.default".

src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/SimpleBundleDoubleNesting.sarif
michaelcfanning: yes, new issue: sarif log emits issues in a non-deterministic order due to lack of sorting

src/Sarif.Driver/Sdk/SarifLogger.cs
michaelcfanning: just a 'secure by default' placeholder until we can drop in wiring to make this option. I was interested in seeing the redaction behavior. instead of this, I will simply make a bool value for emitting more sensitive information and make it false by default.

src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/SimpleBundleDoubleNesting.sarif
lgolding: I published jSchema 0.18.0 this morning that solves that problem by going away from sets and back to lists, but no need to fix in this PR.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -51,6 +51,12 @@ public class AnalyzeOptionsBase : IAnalyzeOptions
         public bool ComputeTargetsHash { get; set; }
 
         [Option(
+            'e',
+            "environment",
+            HelpText = "Log machine environment details of run to output file. WARNING: This option records potentially sensitive information (such as all environment variable values) to any emitted log.")]
+        public bool LogEnvironment { get; set; }
+
```
lgolding: We log other sensitive things (working directory, machine name...). Why does "environment" in particular get its own flag? Why not `--log-sensitive-information` that covers everything? Or alternatively, separate flags (`--log-environment`, `--log-machine-name`, etc.)?

lgolding: Oh, I see. I finally got down to the implementation of `Invocation.Create`. It doesn't mean "environment variables", it means "the whole execution environment". Ok.

src/Sarif/Schemata/Sarif.schema.json
lgolding: This has not yet been corrected.

PR #226, opened 2016-04-28
Add codeSnippet to result. Drop shortMessage and rename fullMessage to message.

michaelcfanning:
@lgolding intrusive change due to rename of one of the most commonly exercised properties (a result message).


lgolding:
This is going to present some conflicts for me since I'm working in the same area (and I'd in fact made the same change) as part of my "notifications" PR. I guess go ahead and merge this but -- do you have other SDK PRs planned for today?

michaelcfanning:
I assumed from your mail that your work was restricted to notifications, sorry. Also, you don't have the private webskim bits required to updates tests, so I thought this was work I could clearly parallelize. 

Are you covering the remainder of what we talked about yesterday? (but which was erased from your board)?

I can stand down until you close your current SDK work, will do that. I've got some other things I can pick up.

lgolding:
Yes, you were certainly entitled to assume that. I opportunistically did the other work without telling you I was doing it. Let me get what I have in this afternoon -- which is notifications only, now that you've merged the "message" change -- and then let's talk about who picks up the remaining small tasks.

PR #227, opened 2016-04-28
Change isSuppressedInSource to an enum to better support for 'unknown…

michaelcfanning:
…' condition (when in-source suppression details aren't available).

@lgolding @rtaket


src/Sarif/Autogenerated/IsSuppressedInSource.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Indicates whether a result has been suppressed in source (or if that information is unknown).
+    /// </summary>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.16.0.0")]
+    public enum IsSuppressedInSource
+    {
+        Unknown,
+        True,False
```
lgolding: I'm intrigued that the code gen put those on the same line.

Not crazy about the names. How about: type = SourceSuppressionState, with values Unknown, Suppressed, NotSuppressed?

src/Sarif/CodeGenHints.json
```diff
@@ -22,11 +22,29 @@
...
   ],
+  "Result.IsSuppressedInSource": [
+    {
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "IsSuppressedInSource",
+        "description": "Indicates whether a result has been suppressed in source (or if that information is unknown).",
+        "zeroValue": "Unknown",
+        "enumValues": [
+          "True,False"
```
lgolding: Oh wow! did it put them on them same line... oh!

That should be:

```JSON
"enumValues": [
  "True",
  "False"
]

Comment (lgolding) [inline src/Sarif/CodeGenHints.json]: (modulo naming question)

**Comment by lgolding** [comment]
:+1: once the bug I pointed out it fixed, and once we agree on the names.

**Comment by michaelcfanning** [comment]
So this is what it looks like in code:
I am fine with your names. enum values of true and false are odd.

**Comment by lgolding** [comment]
Not sure what you meant by "so this is what it looks like..." (I don't see what you're referring to). But once that bug is fixed, then great, we agree on the changes, so go ahead and merge it.

File: src/Sarif/Autogenerated/SuppressionStates.cs [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+using System;
+using System.CodeDom.Compiler;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// The state of a result relative to a baseline of a previous run.
+    /// </summary>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.16.0.0")]
+    [Flags]
```
michaelcfanning: Note: with the new FlagsEnumConverter I wrote, the only change you'd need to make to your jschema code is to add this [Flags] attribute to an enum as required. You could look for a plural name or (preferable to you, I'm sure) add a new code hint making this explicit. [Flags] is required to get the correct ToString() behavior that allows parsing multiple values OR'd together.

src/Sarif/Converters/FileInfoFactory.cs
```diff
@@ -84,7 +92,7 @@ private void AddFile(PhysicalLocation physicalLocation)
                     {
                         new FileData
                         {
-                            MimeType = _mimeTypeClassifier(key)
+                            MimeType = _mimeTypeClassifier(filePath)
```
michaelcfanning: bad bug: MIME type broken here. will file a bug to add unit tests.

michaelcfanning:
@lgolding take a look. I included the AlgorithmKind change here and there are also new tests to review. Please merge this tomorrow or make any required changes to this branch that you think are warranted and merge.

src/Sarif.FunctionalTests/ConverterTestData/FxCop/FxCopReportAllMessageStates.xml.sarif
```diff
@@ -0,0 +1,357 @@
...
+          {
+            "name": "wpfapplication5.exe",
+            "kind": "module"
+          },
+          {
+            "name": "WpfApplication5",
+            "kind": "namespace"
+          },
+          {
+            "name": "Foo",
```
lgolding: Discouraged in public-facing code.

src/Sarif.FunctionalTests/ConverterTestData/FxCop/FxCopReportAllMessageStates.xml.sarif
```diff
@@ -0,0 +1,357 @@
...
+            "kind": "type"
+          }
+        ],
+        "WpfApplication5.Properties.Resources..ctor()": [
+          {
+            "name": "wpfapplication5.exe",
+            "kind": "module"
+          },
+          {
+            "name": "WpfApplication5.Properties",
```
lgolding: Should this be two namespaces?

```json
{
  "name": "WpfApplication5",
  "kind": "namespace"
},
{
  "name": "Properties",
  "kind": "namespace"
}
```

If we specify that namespace components should be separated, then we can can ask the question "How many issues are in namespace N _and all of its sub-namespaces_", without having to parse a compound name. If we specify that namespace components should be combined (as you did here), then we can ask the question "How many issues are in namespace N, _excluding its sub-namespaces_", without having to examine the next logical location component to ensure that it isn't a namespace.

So the questions are:

1. Which is the more useful question to be able to ask easily?
2. If there's no clear answer to (1), do we give producers the choice of whether to separate or combine the components?

src/Sarif.UnitTests/Converters/FxCopConverterTests.cs
```diff
@@ -188,12 +188,13 @@ public void FxCopLogReader_Context_RefineMessage()
         {
             var context = TestHelper.CreateProjectContext();
 
-            context.RefineMessage("CA0000", "VeryUsefulCheck", "1", "MyCategory", "Breaking");
+            context.RefineMessage("CA0000", "VeryUsefulCheck", "1", "MyCategory", "Breaking", "ExcludedInSource");
```
lgolding: I thought we had names `"Unknown"`, `"SuppressedInSource"`, and `"SuppressedInBaseLine"`.

src/Sarif.UnitTests/Readers/AlgorithmKindConverterTests.cs
```diff
@@ -43,36 +43,88 @@ private static string GetJson(Action<ResultLogJsonWriter> testContent)
...
-                var run = new Run();
+                string expected = 
+                    "{\"version\":\"1.0.0-beta.4\",\"runs\":[{\"tool\":{\"name\":null},\"files\":{\"http://abc/\":[{\"hashes\":[{\"value\":null,\"algorithm\":\"" +
+                    testTuple.Item2 +  
+                    "\"}]}]},\"results\":[{}]}]}";
 
-                uut.WriteTool(s_defaultTool);
+                string actual = GetJson(uut =>
+                {
+                    var run = new Run();
```
lgolding: Doesn't appear to be used

src/Sarif/Readers/SarifContractResolver.cs
```diff
@@ -34,8 +35,14 @@ protected override JsonContract CreateContract(Type objectType)
             else if (objectType == typeof(NotificationLevel))
                 contract.Converter = EnumConverter.Instance;
 
-            else if (objectType == typeof(AlgorithmKind))
-                contract.Converter = AlgorithmKindConverter.Instance;
+            if (objectType == typeof(AlgorithmKind))
```
lgolding: Why remove the else's? It causes unnecessary tests (minor), and it suggests to the reader that the cases are not exclusive, which they are.

src/Sarif/Readers/FlagsEnumConverter.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class FlagsEnumConverter : JsonConverter
```
lgolding: Shouldn't be called this if in fact it only treats`SuppressionStates`. Did you intend to make it a generic type?

src/Sarif/NotYetAutogenerated/SuppressionStates.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// The state of a result relative to a baseline of a previous run.
+    /// </summary>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.16.0.0")]
+    [Flags]
+    public enum SuppressionStates
+    {
+        None,
```
lgolding: Shouldn't it be "Unknown"? I thought that was the result of our discussion -- that if you didn't see one of the positive values, that you just couldn't tell if it was suppressed. "None" sounds like it means "not suppressed", which we don't know.

michaelcfanning: None means we have no SuppressionState information of any kind. None is a typical name for an enum zero value

src/Sarif.FunctionalTests/ConverterTestData/FxCop/FxCopReportAllMessageStates.xml.sarif
michaelcfanning: fixed

src/Sarif.UnitTests/Converters/FxCopConverterTests.cs
michaelcfanning: These are FxCop terms, not SARIF.

src/Sarif/Autogenerated/SuppressionStates.cs
lgolding: Sounds good.

src/Sarif/NotYetAutogenerated/SuppressionStates.cs
lgolding: I agree that it's typical. I'm saying two things:

1. In this particular case, "None" didn't convey _to me_ the meaning "no information about suppression". To me, it conveyed "no suppression".

2. The guidelines allow you wiggle room on this point:

> **DO** provide a value of zero on simple enums.
>
> _Consider_ calling the value _something like_ "None" [my emphasis]

Given that, please consider renaming to Unknown, but I'll accept your decision.

michaelcfanning: If we have a value of Unknown, then we need an explicit value for NotSuppressedInSource. Now it becomes unclear how to manage this as a [Flags] enum. We would have Unknown, SuppressedInSource, NotSuppressedInSource as a non-flags enum and SuppressedInBaseline would be handled elsewhere. With the current approach, we maintain a set of exclusion conditions in the event that we have some positive signal of this kind. I think this is a better approach but am open to continued discussion.

src/Sarif.FunctionalTests/ConverterTestData/FxCop/FxCopReportAllMessageStates.xml.sarif
michaelcfanning: Yes. separate issue. The FxCop converter should break these apart. There is also the open question of whether the log file should contain the root container as an item if it matches the analysis target.

michaelcfanning: Opens #236 and #237 to help track fxcop converter fixes.

src/Sarif/Readers/SarifContractResolver.cs
michaelcfanning: just a mistake. fixed.

src/Sarif/Readers/FlagsEnumConverter.cs
michaelcfanning: changed code to consume the objectType argument, good catch.

PR #234, opened 2016-05-02
Add language/region information to tool data.

michaelcfanning:
@lgolding 

please merge any of these that stand as far as you're concerned.

### DIFF
--- a/src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/LocalizedBundleDoubleNesting.sarif
+++ b/src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/LocalizedBundleDoubleNesting.sarif
@@ -1,0 +1,78 @@
+{
+  "version": "1.0.0-beta.4",
+  "runs": [
+    {
+      "tool": {
+        "name": "ModernCop",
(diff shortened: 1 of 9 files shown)


src/Sarif.FunctionalTests/DirectProducerTestData/WebSkim/LocalizedBundleDoubleNesting.sarif
```diff
@@ -0,0 +1,78 @@
...
+    {
+      "tool": {
+        "name": "ModernCop",
+        "fullName": "mod50 7.0.7000.0",
+        "version": "7.0.7000.0",
+        "semanticVersion": "7.0.7000",
+        "fileVersion": "7.0.20811.0",
+        "language": "en-US",
+        "properties": {
+          "Language": "Language Neutral",
```
lgolding: `language` is `en-US` but `properties.Language` is `Language Neutral`?

michaelcfanning:
A language neutral assembly contains the resources for the fallback language, which in this case is en-US. This is an optimization that allows the CLR to avoid probing for satellite assemblies in some cases. We probably don't need to emit this property, I can remove it.

PR #235, opened 2016-05-02
Add default severity to rule metadata.

michaelcfanning:
@lgolding

### DIFF
--- a/src/Sarif/CodeGenHints.json
+++ b/src/Sarif/CodeGenHints.json
@@ -50,6 +50,23 @@
       "kind": "DictionaryHint"
     }
   ],
+  "Rule.DefaultLevel": [
+    {
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "ResultLevel",
+        "description": "Values specifying the level of a result.",
+        "zeroValue": "Unknown",
+        "enumValues": [
+          "NotApplicable",
+          "Pass",
+          "Note",
+          "Warning",
+          "Error"
+        ]
+      }
+    }
+  ],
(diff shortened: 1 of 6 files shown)


src/Sarif/CodeGenHints.json
```diff
@@ -50,6 +50,23 @@
       "kind": "DictionaryHint"
     }
   ],
+  "Rule.DefaultLevel": [
```
michaelcfanning: Of course, what we'd really like to do is to create a single enum, ResultLevel, but the serializer doesn't support this.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -658,6 +658,12 @@
           "type": "object"
         },
 
+        "defaultLevel": {
```
michaelcfanning: would be good here to have a reference to a separate type that contained a shared definition of resultLevel

michaelcfanning:
please merge if correct.

src/Sarif/CodeGenHints.json
lgolding: Yes, we need the same Schema feature -- the ability to specify that a defined type in the schema is an enum rather than a class -- to properly support SuppressionState (because that's what would allows us to define an array of SuppressionState). I'll do it this week.

src/Sarif/Schemata/Sarif.schema.json
lgolding: Yes, as above.

lgolding:
👍

PR #239, opened 2016-05-02
Add sarifLog.$schema and notification.analysisTarget

lgolding:
@michaelcfanning


src/Sarif/NotYetAutogenerated/SarifLog.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                return SarifNodeKind.SarifLog;
+            }
+        }
+
+        /// <summary>
+        /// The URI of the JSON schema corresponding to <see cref="Version"/>.
+        /// </summary>
+        [DataMember(Name="$schema")]
+        public Uri SchemaUri { get; set; }
```
lgolding: This file is in NotYetAutogenerated because we don't have a code gen hint that specifies the data member name (the generator always just camel-cases the C# property name), and the property can't be named "$schema".

PR #242, opened 2016-05-03
Migrate to object model with reference equality

lgolding:
Update to JSchema 0.20.0, which generates object model classes with reference equality rather than value equality semantics. The value equality logic is pushed into separate comparer classes that implement `IEqualityComparer<T>`.

@michaelcfanning @nguerrera


src/Sarif.UnitTests/Converters/AndroidStudioConverterTests.cs
```diff
@@ -251,8 +251,14 @@ public void AndroidStudioConverter_ConvertSarifResult_RecordsModuleAsTopLevelIfP
 
             LocationInfo locationInfo = GetLocationInfoForBuilder(builder);
 
-            locationInfo.Location.Should().Be(expectedLocation);
-            locationInfo.LogicalLocationComponents.SequenceEqual(expectedLogicalLocationComponents).Should().BeTrue();
+            LocationEqualityComparer.Instance.Equals(locationInfo.Location, expectedLocation)
+                .Should().BeTrue();
+
```
lgolding: In future, will be `locationInfo.Location.EqualsByValue(expectedLocation)` or something like that.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -56,131 +56,6 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "tags", IsRequired = false, EmitDefaultValue = false)]
         public IList<string> Tags { get; set; }
 
-        public override bool Equals(object other)
```
lgolding: Overrides of `Equals` and `GetHashCode` moved out of object model classes.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -12,8 +12,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// A code annotation that consists of single physical location and associated message, used to express code flows through a method, or other locations that are related to a result.
     /// </summary>
     [DataContract]
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.19.0.0")]
-    public partial class AnnotatedCodeLocation : ISarifNode, IEquatable<AnnotatedCodeLocation>
```
lgolding: Object model classes no longer implement `IEquatable<T>`.

src/Sarif/Autogenerated/AnnotatedCodeLocationEqualityComparer.cs
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) Microsoft.  All Rights Reserved.
```
lgolding: The new comparer classes.

src/Sarif/Autogenerated/AnnotatedCodeLocationEqualityComparer.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                return true;
+            }
+
+            if (ReferenceEquals(left, null) || ReferenceEquals(right, null))
+            {
+                return false;
+            }
+
+            var physicalLocationEqualityComparer = new PhysicalLocationEqualityComparer();
```
lgolding: I wrote the code to generate this line before I added the `Instance` properties (above), and forgot to change it. Will change to `left.PhysicalLocation.EqualsByValue(right.PhysicalLocation)` or something like that when I introduce that API.

src/Sarif/NotYetAutogenerated/SarifLogEqualityComparer.cs
```diff
@@ -0,0 +1,95 @@
...
+                return true;
+            }
+
+            if (ReferenceEquals(left, null) || ReferenceEquals(right, null))
+            {
+                return false;
+            }
+
+            var runEqualityComparer = new RunEqualityComparer();
+            if (left.SchemaUri != right.SchemaUri)
```
lgolding: Still in NotYetAutoGenerated because I don't yet have "PropertyNameHint" that will map the JSON property `"$schema"` to the .NET property `SchemaUri`.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -753,7 +753,6 @@
           "description": "The set of results contained in an SARIF log. The results array can be omitted when a run is solely exporting rules metadata. It must be present (but may be empty) in the event that a log file represents an actual scan.",
           "type": "array",
           "minItems": 0,
-          "uniqueItems": true,
```
lgolding: Because the AndroidStudio converter generates non-unique items, and we're no longer using HashSets, we have two choices:

1. Remove `uniqueItems` from the `results` array.
2. Check each result for equality with all existing results before adding to the list (N^2 behavior).

src/Sarif.UnitTests/Converters/AndroidStudioConverterTests.cs
```diff
@@ -345,8 +363,14 @@ public void AndroidStudioConverter_ConvertSarifResult_GeneratesLocationWithOnlyP
...
+            locationInfo.LogicalLocationComponents
+                .SequenceEqual(
+                    expectedLogicalLocationComponents,
+                    LogicalLocationComponentEqualityComparer.Instance)
```
nguerrera: For the cases where you need the comparer (like this), consider having something a little shorter (and maybe less cluttering, and easier to find). 

Maybe:

``` C#
public class LogicalLocationComponent {
     ...
     public static IEqualityComparer<LogicalLocationComponent> ValueComparer { get; }
}
```

You could then make the concrete comparer types internal.

lgolding: That's nice. I'll do that in a separate PR, at the same time I introduce `bool EqualsByValue(T other)`.

lgolding: Huh, actually `ValueEquals` is a better name, and if you type `foo.Eq` it pops up in the Intellisense right next to `Equals`. What do you think?

nguerrera: I like it. I didn't even realize that IntelliSense has gotten smarter about that. It does read better.

So, taken together, we'd have two value-comparison members per class and no extra public types.

``` C#
public class LogicalLocationComponent {
        ...
     public static IEqualityComparer<LogicalLocationComponent> ValueComparer { get; }
     public bool ValueEquals(LogicalLocationComponent other) => ValueComparer.Equals(this, other);
}

That looks pretty nice to me. :)

Comment (lgolding) [inline src/Sarif.UnitTests/Converters/AndroidStudioConverterTests.cs]: It does. I can implement the `ValueComparer` property as well:

```C#
public class LogicalLocationComponent {
     ...
     public static IEqualityComparer<LogicalLocationComponent> ValueComparer => LogicalLocationComponentValueComparer.Instance;
     public bool ValueEquals(LogicalLocationComponent other) => ValueComparer.Equals(this, other);
}
```

(`LogicalLocationComponentValueComparer.Instance` can now be internal)

nguerrera: :+1:

src/Sarif/Schemata/Sarif.schema.json
nguerrera: Note that in circumstances where a caller might want unique items, they can get them in O(N) with very little code e.g. by passing the comparer to https://msdn.microsoft.com/en-us/library/bb338049(v=vs.110).aspx

nguerrera:
This overall change looks good to me. (And we also agree on next steps.)

src/Sarif/Schemata/Sarif.schema.json
lgolding: Oh, nice!

src/Sarif/Autogenerated/InvocationEqualityComparer.cs
```diff
@@ -10,10 +10,10 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Defines methods to support the comparison of objects of type Invocation for equality.
     /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.20.0.0")]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.21.0.0")]
     public sealed class InvocationEqualityComparer : IEqualityComparer<Invocation>
     {
-        public static readonly InvocationEqualityComparer Instance = new InvocationEqualityComparer();
+        internal static readonly InvocationEqualityComparer Instance = new InvocationEqualityComparer();
```
nguerrera: I suggest making the whole concrete comparer classes internal rather than just the Instance properties.

src/Sarif/Autogenerated/AnnotatedCodeLocationEqualityComparer.cs
```diff
@@ -10,10 +10,10 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Defines methods to support the comparison of objects of type AnnotatedCodeLocation for equality.
     /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.20.0.0")]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.21.0.0")]
     public sealed class AnnotatedCodeLocationEqualityComparer : IEqualityComparer<AnnotatedCodeLocation>
     {
-        public static readonly AnnotatedCodeLocationEqualityComparer Instance = new AnnotatedCodeLocationEqualityComparer();
+        internal static readonly AnnotatedCodeLocationEqualityComparer Instance = new AnnotatedCodeLocationEqualityComparer();
```
nguerrera: Commenting again since last one is "out-of-date" but not fixed.  I would make the whole concrete comparer classes internal rather than just the Instance properties.

src/Sarif/Autogenerated/RunEqualityComparer.cs
```diff
@@ -0,0 +1,339 @@
...
+                if (obj.Rules != null)
+                {
+                    // Use xor for dictionaries to be order-independent.
+                    int xor_2 = 0;
+                    foreach (var value_11 in obj.Rules)
+                    {
+                        xor_2 ^= value_11.Key.GetHashCode();
+                        if (value_11.Value != null)
+                        {
+                            xor_2 ^= value_11.Value.GetHashCode();
```
michaelcfanning: We shouldn't be calling the default Sarif object GetHashCode here but the one from the associated equality comparer, right?

nguerrera:
I'm not sure why GitHub is marking new comments out-of-date so re-stating this at top-level: I suggest making the whole concrete comparer classes internal rather than just the Instance properties.

src/Sarif/Autogenerated/RunEqualityComparer.cs
lgolding: Yes, bad bug, will need another iteration.

lgolding:
@nguerrera Yes, I will make the comparers themselves internal.

lgolding:
Pushed change to fix bug found by @michaelcfanning, which was that the generated GetHashCode was incorrectly calling the sub-object's GetHashCode rather than the sub-object's comparer's GetHashCode.

As it stands one unit test is failing. The validation test is complaining that the list of results in AndroidStudio_issueLog5_raw.xml.sarif contains duplicates. That's true, but the schema no longer requires the results array to be unique, so I don't know what's going on. Some kind of mismatch, but I haven't found it yet.

lgolding:
@michaelcfanning figured it out. The `$schema"` property in the test files means that the validation library we use is now using the schema on schemastore.org -- which has the constraint, even though our local copy does not.

So:
1. Restore the `uniqueItems` constraint on `run.results` in the source tree (which I had removed in the course of this PR).
2. Make AndroidStudio use the `Result.ValueComparer` as it constructs its `HashSet<Result>`.

PR #244, opened 2016-05-03
Provide converter to turn IRule into Rule

michaelcfanning:
@lgolding


src/Sarif/Readers/IRuleConverter.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class IRuleDictionaryConverter : JsonConverter
```
lgolding: Class name doesn't match file name.

src/Sarif/Readers/IRuleConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    DefaultLevel = iRule.DefaultLevel,
+                    FullDescription = iRule.FullDescription,
+                    HelpUri = iRule.HelpUri,
+                    Id = iRule.Id,
+                    MessageFormats = iRule.MessageFormats,
+                    Name = iRule.Name,
+                    Properties = iRule.Properties,
+                    ShortDescription = iRule.ShortDescription,
+                    Tags = iRule.Tags
```
lgolding: A bit of a maintenance problem if we add properties to `IRule`, but I don't see a way around it.

lgolding:
:+1: (+/- the file name change)

PR #245, opened 2016-05-03
Sarif logger was not persisting notifications.

michaelcfanning:
@lgolding 

unit test hole. will close asap.


src/Sarif.Driver/Sdk/SarifLogger.cs
```diff
@@ -197,6 +197,16 @@ public void Dispose()
...
+                if (_run.ConfigurationNotifications != null)
+                {
+                    _issueLogJsonWriter.WriteConfigurationNotifications(_run.ConfigurationNotifications);
+                }
+
+                if (_run.ToolNotifications != null)
+                {
+                    _issueLogJsonWriter.WriteToolNotifications(_run.ToolNotifications);
+                }
+
```
lgolding: *jaw drops* I would have sworn I wrote those lines. Good catch.

PR #252, opened 2016-05-05
Persist mime type in SarifLogger. Remove stacks from configuration ex…

michaelcfanning:
…ception notifications. Reclassify 'could not parse target' as a configuration issue, not a tool error (since the most likely scenario is that an invalid target is being inspected, not that the tool parser is broken). @lgolding @rtaket


src/Sarif.UnitTests/JsonTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public JsonTests()
+        {
+            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
+            {
+                ContractResolver = SarifContractResolver.Instance
+            };
+        }
+    }
+}
```
lgolding: Funny you should be doing this; I did the same thing in a branch recently. I suggest you move the `GetJson` method that's duplicated in a few test classes into this base class as well.

src/Sarif.UnitTests/Writers/ResultLogObjectWriterTests.cs
```diff
@@ -1,84 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
```
lgolding: Why removed?

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -66,5 +68,42 @@ public void SarifLogger_RedactedCommandLine()
...
+        public void SarifLogger_WritesFileData()
+        {
+            var sb = new StringBuilder();
+            string file;
+
+            using (var tempFile = new TempFile(".cpp"))
+            using (var textWriter = new StringWriter(sb))            
+            {
+                File.WriteAllText(tempFile.Name, "#include \"windows.h\";");
+                file = tempFile.Name;
```
lgolding: Reverse these two lines and then use `file` in the call to `WriteAllText`.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -66,5 +68,42 @@ public void SarifLogger_RedactedCommandLine()
...
+
+                using (var sarifLogger = new SarifLogger(
+                    textWriter,
+                    analysisTargets: new string[] { tempFile.Name },
```
lgolding: ... and use it here as well.

lgolding:
:+1:

src/Sarif.UnitTests/Writers/ResultLogObjectWriterTests.cs
michaelcfanning: matching product type was deleted. test file removed from project but lingered on disk

PR #253, opened 2016-05-05
Diff using value type comparison semantics instead of reference equality

michaelcfanning:
@lgolding. Oops! Test was not authored in a way such that reference equality worked. Now fixed.

### DIFF
--- a/src/Sarif/Readers/ResultDiffingVisitor.cs
+++ b/src/Sarif/Readers/ResultDiffingVisitor.cs
@@ -9,9 +9,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Readers
     {
         public ResultDiffingVisitor(SarifLog sarifLog)
         {
-            this.AbsentResults = new HashSet<Result>(ResultEqualityComparer.Instance);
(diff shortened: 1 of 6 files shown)


src/Sarif/Readers/ResultDiffingVisitor.cs
```diff
@@ -31,7 +31,7 @@ public bool Diff(IEnumerable<Result> actual)
 
             this.AbsentResults = this.SharedResults;
 
-            this.SharedResults = new HashSet<Result>();
+            this.SharedResults = new HashSet<Result>(ResultEqualityComparer.Instance);
```
lgolding: This works, but that property is actually `internal`. The public API is `Result.ValueComparer`.

lgolding: ... so I suggest the change for two reasons:

1. Tests should use the public API as far as possible, so they test what customers could do.
2. The public API is shorter and more readable, and does not introduce additional classes (the class `ResultEqualityComparer` and all the concrete comparer classes are internal.

lgolding:
:+1:

PR #256, opened 2016-05-06
Ids

michaelcfanning:
@rtaket @lgolding


src/Sarif/IResultLogWriter.cs
```diff
@@ -11,7 +11,9 @@ public interface IResultLogWriter
         /// <summary>
         /// Initialize the current output log.
         /// </summary>
-        void Initialize();
+        /// <param name="id">A string that uniquely identifies a run.</param>
+        /// <param name="correlationId">A global identifier for a run that permits correlation with a larger automation process.</param> 
+        void Initialize(string id, string correlationId);
```
lgolding: In the email we discussed making these parameters optional; did you decide against that?

PR #257, opened 2016-05-10
Support arbitrary properties

lgolding:
As the code stands, values of all types can be read and round-tripped. But there is not yet a SetProperty<T> API that would allow you to set a property to an arbitrary type T and then write it out.

@michaelcfanning 

Adding @nguerrera @rtaket FYI


src/Sarif/CodeGenHints.json
```diff
@@ -17,11 +69,6 @@
       }
     }
   ],
-  "ExceptionData.Properties": [
-    {
-      "kind": "DictionaryHint"
-    }
-  ],
```
lgolding: Forgot to remove this when we removed `exception.properties` from the spec.

src/Sarif/CodeGenHints.json
```diff
@@ -30,19 +77,204 @@
...
+      "kind": "PropertyModifiersHint",
+      "arguments": {
+        "modifiers": [
+          "internal",
+          "override"
+        ]
+      }
+    }
+  ],
+  "Hash.Algorithm": [
```
lgolding: Sorted into alphabetical order. `Hash.Algorithm` used to be further down int the file.

src/Sarif/CodeGenHints.json
```diff
@@ -60,6 +292,44 @@
       }
     }
   ],
+  "Result.Level": [
```
lgolding: Again, sorted into order (used to be further down).

src/Sarif/CodeGenHints.json
```diff
@@ -116,29 +413,11 @@
...
-    {
-      "kind": "DictionaryHint"
-    }
-  ],
   "Run.Rules": [
     {
       "kind": "DictionaryHint"
     }
   ],
-  "Invocation.EnvironmentVariables": [
```
lgolding: Sorted into alphabetical order (moved up in the file).

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -12,8 +13,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// A code annotation that consists of single physical location and associated message, used to express code flows through a method, or other locations that are related to a result.
     /// </summary>
     [DataContract]
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.22.0.0")]
-    public partial class AnnotatedCodeLocation : ISarifNode
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.28.0.0")]
+    public partial class AnnotatedCodeLocation : PropertyBagHolder, ISarifNode
```
lgolding: Every class with a property bag derives from `PropertyBagHolder`, which implements the `{Get,Set}Property` APIs.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -47,7 +48,7 @@ public SarifNodeKind SarifNodeKind
         /// Key/value pairs that provide additional information about the code location.
         /// </summary>
         [DataMember(Name = "properties", IsRequired = false, EmitDefaultValue = false)]
-        public IDictionary<string, string> Properties { get; set; }
+        internal override IDictionary<string, SerializedPropertyInfo> Properties { get; set; }
```
lgolding: Those property access APIs in the base class need access to the actual `Properties` dictionary. So declare the dictionary `abstract` in the base class, and override it here. (When discussed this on Friday, I couldn't get it to work. Turned out I needed to add some attributes to the property in the base class to get it to serialized properly.)

src/Sarif/Autogenerated/AnnotatedCodeLocationEqualityComparer.cs
```diff
@@ -37,7 +38,7 @@ public bool Equals(AnnotatedCodeLocation left, AnnotatedCodeLocation right)
                 return false;
             }
 
-            if (!Object.ReferenceEquals(left.Properties, right.Properties))
+            if (!object.ReferenceEquals(left.Properties, right.Properties))
```
lgolding: Unrelated cleanup. No functional change.

src/Sarif/Autogenerated/AnnotatedCodeLocationEqualityComparer.cs
```diff
@@ -46,20 +47,20 @@ public bool Equals(AnnotatedCodeLocation left, AnnotatedCodeLocation right)
...
                 {
-                    string value_1;
+                    SerializedPropertyInfo value_1;
                     if (!right.Properties.TryGetValue(value_0.Key, out value_1))
                     {
                         return false;
                     }
 
-                    if (value_0.Value != value_1)
+                    if (!object.Equals(value_0.Value, value_1))
```
lgolding: Note how the generator knows to call `object.Equals` to compare two `SerializedPropertyInfo` objects, as opposed to the `!=` that it emitted when the dictionary has `string` values. The newly enhanced version of the `DictionaryHint` in the code generator knows how to do that (see the `comparisonType` argument to the `DictionaryHint`s in the code gen hints file later on).

src/Sarif/Autogenerated/IRule.cs
```diff
@@ -10,8 +11,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Interface exposed by objects that provide information about analysis rules.
     /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.22.0.0")]
-    public interface IRule
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.28.0.0")]
+    public partial interface IRule
```
lgolding: The SDK includes code in which an `IRule`-derived object needs access to the properties (below, you'll see a line of code that references `iRule.Properties`). But `Properties` can't be part of the public interface. What we want instead is for the property accessor methods `{Get,Set}Property` to be part of the `IRule` interface. But the code generator can't generate those methods, which are specific to the SDK.

So I modified the code generator to emit _partial_ interfaces. Now the client of the generator -- the SDK -- can add whatever it wants to the generated interfaces. You'll see below that `IRule` now extends `IPropertyBagHolder`, which in turn declares the property accessor methods.

src/Sarif/CodeGenHints.json
```diff
@@ -1,12 +1,64 @@
 {
+  "annotatedCodeLocation": [
+    {
+      "kind": "BaseTypeHint",
+      "arguments": {
```
lgolding: The code generator now supports a "base type" hint, which allows the classes with properties to derive from `PropertyBagHolder`.

src/Sarif/CodeGenHints.json
```diff
@@ -1,12 +1,64 @@
...
+      }
+    }
+  ],
   "AnnotatedCodeLocation.Properties": [
     {
-      "kind": "DictionaryHint"
+      "kind": "DictionaryHint",
+      "arguments": {
+        "valueTypeName": "SerializedPropertyInfo",
+        "namespaceName": "Microsoft.CodeAnalysis.Sarif.Readers",
```
lgolding: `DictionaryHint` now has a `ValueTypeName` property (that thing you were trying to use weeks ago but it didn't exist at the time). That allows the property bags to be of type `IDictionary<string, SerializedPropertyInfo>`.

src/Sarif/CodeGenHints.json
```diff
@@ -1,12 +1,64 @@
...
+        "comparisonKind": "ObjectEquals",
+        "hashKind": "ScalarReferenceType",
+        "initializationKind": "SimpleAssign"
```
lgolding: The generated code needs to know how to compare, hash, and initialize objects of type `SerializedPropertyInfo`. This isn't a schema type, so the generator can't figure it out by looking at the schema. So the code gen hint tells the generator what kind of code to generate for comparing, hashing, and initializing.

src/Sarif.Driver/Sdk/ExportRulesMetadataCommandBase.cs
```diff
@@ -130,12 +130,16 @@ private void OutputSarifRulesMetada(string outputFilePath, ImmutableArray<IRule>
...
                 newRule.HelpUri = rule.HelpUri;
-                newRule.Properties = rule.Properties;
                 newRule.FullDescription = rule.FullDescription;
                 newRule.MessageFormats = rule.MessageFormats;
 
                 newRule.ShortDescription = rule.ShortDescription;
 
+                foreach (string propertyName in rule.PropertyNames)
+                {
+                    newRule.SetProperty(propertyName, rule.GetProperty(propertyName));
```
lgolding: You'll see below that `SetProperty` creates the underlying dictionary if it doesn't already exist, so you'll see other places where I could get rid of SDK code that created an empty dictionary.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            if (typeof(T) == typeof(string))
+            {
+                throw new InvalidOperationException(SdkResources.CallNonGenericGetProperty);
+            }
+
+            return JsonConvert.DeserializeObject<T>(Properties[propertyName].SerializedValue);
+        }
+
+        public void SetProperty(string propertyName, string value)
```
lgolding: We only handle setting string-valued properties so far, but we successfully round-trip properties of all types.

src/Sarif/Core/StackFrame.cs
```diff
@@ -55,10 +55,7 @@ public static StackFrame Create(System.Diagnostics.StackFrame dotNetStackFrame)
 
             if (nativeOffset != -1)
             {
-                stackFrame.Properties = new Dictionary<string, string>
-                {
-                    { "NativeOffset", nativeOffset.ToString(CultureInfo.InvariantCulture) }
-                };
+                stackFrame.SetProperty("NativeOffset", nativeOffset.ToString(CultureInfo.InvariantCulture));
```
lgolding: This could be an integer property if we supported explicitly setting non-string-valued properties.

src/Sarif/Readers/IRuleConverter.cs
```diff
@@ -44,11 +44,15 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
                     Name = iRule.Name,
-                    Properties = iRule.Properties,
                     ShortDescription = iRule.ShortDescription,
                     Tags = iRule.Tags
                 };
 
+                foreach (string propertyName in iRule.PropertyNames)
+                {
+                    rule.SetProperty(propertyName, iRule.GetProperty(propertyName));
+                }
```
lgolding: We might consider a adding a convenience method `IPropertyBagHolder.CopyPropertiesFrom(IPropertyBagHolder other)`.

src/Sarif/Readers/SerializedPropertyInfo.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// is represented by
+        /// <code>
+        /// new SerializedPropertyInfo("[ 1, \"b\" ]", JTokenType.Array)
+        /// </code>
+        /// The object-values JSON property <code>"o": { "a": 1, "b": false }</code>
+        /// is represented by
+        /// <code>
+        /// new SerializedPropertyInfo("{ \"a\": 1, \"b\": false }", JTokenType.Object)
+        /// </code>
+        /// </example>
```
lgolding: Although I did try it out in my little test console app, I see that I don't actually have unit tests for round-tripping properties of all types. I'll add them now.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -149,11 +149,9 @@ private static string Redact(string text, IEnumerable<string> tokensToRedact)
 
         private void SetSarifLoggerVersion(Tool tool)
         {
-            tool.Properties = tool.Properties ?? new Dictionary<string, string>();
```
lgolding: Example of a place where we don't need the initialization because `SetProperty` takes care of it.

src/Sarif/Readers/SerializedPropertyInfo.cs
lgolding: Added PropertyBagConverterTests.cs.

src/Sarif/CodeGenHints.json
michaelcfanning: What's the namespace hint for?

---
In reply to: [62749197](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62749197) [](ancestors = 62749197)

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: nice

---
In reply to: [62750560](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62750560) [](ancestors = 62750560)

src/Sarif/Readers/SerializedPropertyInfo.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// The string representation of the JSON value of the property.
+        /// </param>
+        /// <param name="jTokenType">
+        /// The JSON type of the property's value.
+        /// </param>
+        /// <remarks>
+        /// This representation allows properties to be read from JSON into memory and
+        /// then round-tripped back to JSON, without storing "live" Json.NET objects.
+        /// Live JSON objects hold references to their parent container, which would consume
+        /// alot of memory for objects that are part of a large JSON file.
```
michaelcfanning: don't think 'alot' is a discrete term. :)

src/Sarif/CodeGenHints.json
lgolding: That is the namespace in which SerializedPropertyInfo resides. It causes the directive "using Microsoft.CodeAnalysis.Sarif.Readers" to appear at the top of any file that declares an IDictionary<string, SerializedPropertyInfo>. Otherwise you get a compiler error because it can't find SerializedPropertyInfo.

It would also work to specify "Microsoft.CodeAnalysis.Sarif.Readers.SerializedPropertyInfo" as the value of the "valueTypeName" argument, and to omit the "namespaceName" argument, but then the generated property declaration would have the type IDictionary<string,Microsoft.CodeAnalysis.Sarif.Readers.SerializedPropertyInfo>. I liked this better.

---
In reply to: [62769424](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62769424) [](ancestors = 62769424,62749197)

src/Sarif/Readers/SerializedPropertyInfo.cs
michaelcfanning: Including Date, Guid, et al?

---
In reply to: [62750443](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62750443) [](ancestors = 62750443)

src/Sarif/Readers/SerializedPropertyInfo.cs
lgolding: True. Fixed.

---
In reply to: [62770040](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62770040) [](ancestors = 62770040)

src/Sarif/Readers/SerializedPropertyInfo.cs
lgolding: Ah. No, I didn't add tests for those. Will add.

---
In reply to: [62770222](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62770222) [](ancestors = 62770222,62750443)

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+                    jTokenType = JTokenType.Integer;
+                }
+                else if (propertyType == typeof(double))
+                {
+                    jTokenType = JTokenType.Float;
+                }
+                else if (propertyType == typeof(string))
+                {
+                    jTokenType = JTokenType.String;
+                }
```
michaelcfanning: What about other JToken types? Date, TimeSpan, Guid?

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+
+                if (jTokenType == JTokenType.Undefined)
+                {
+                    throw new ApplicationException(
```
michaelcfanning: Don't use ApplicationException, not advised. See the remarks section in the docs that basically say, don't use this thing as designed. InvalidOperationException is a good choice here.

[https://msdn.microsoft.com/en-us/library/system.applicationexception(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/system.applicationexception(v=vs.110).aspx)

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -0,0 +1,110 @@
...
+                    key,
+                    new SerializedPropertyInfo(serializedValue, jTokenType));
+            }
+
+            return propertyDictionary;
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteStartObject();
```
michaelcfanning: Does this code write out Dates in our expected format?

src/Sarif/Core/StackFrame.cs
michaelcfanning: Coming next, yes?

---
In reply to: [62750080](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62750080) [](ancestors = 62750080)

src/Sarif/Readers/PropertyBagConverter.cs
lgolding: I need to think about that. JSON doesn't define (for example) a Guid type, so in the file you'd just have a string:

{
  "properties": {
    "g": "{12345678-90ab-cdef-1234-567890abcdef}"
  }
}

That means that when the converter gets hold of it, all it will know is that there's a string. You can ask for it as a Guid and that will work. I just wrote this unit test and it passes:

            _input = "{\"properties\":{\"g\":\"{12345678-90ab-cdef-1234-567890abcdef}\"}}";

            PerformRoundTrip();

            Guid expectedGuid = new Guid("{12345678-90ab-cdef-1234-567890abcdef}");

            _inputObject.GetProperty<Guid>("g").Should().Be(expectedGuid);

The same should work for Uris and DateTimes; I'll check.

---
In reply to: [62770570](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62770570) [](ancestors = 62770570)

src/Sarif/Core/StackFrame.cs
lgolding: Yes, that's next.

---
In reply to: [62771090](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62771090) [](ancestors = 62771090,62750080)

src/Sarif/Readers/PropertyBagConverter.cs
lgolding: It "just works" because GetProperty<T> just calls through to JsonConvert.DeserializeObject<T>, which does the right thing.

---
In reply to: [62771104](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62771104) [](ancestors = 62771104,62770570)

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -82,8 +82,6 @@ internal static string FirstSentence(string fullDescription)
 
         public IDictionary<string, string> Options { get; }
```
michaelcfanning: Interesting post v1 thought: express tool parameterization via an options JSON blob that's handled with the same machinery as the properties. Your code could replace/provide an alternate for the current approach (which uses XML serialization). This would also allow us to express this parameterization directly in a SARIF log file.

src/Sarif.Driver/Sdk/ExportRulesMetadataCommandBase.cs
michaelcfanning: A nice increase in usability.

---
In reply to: [62749682](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62749682) [](ancestors = 62749682)

src/Sarif/VersionConstants.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     public static class VersionConstants                                       
     {                                                                          
         public const string Prerelease = "-beta";                       
-        public const string AssemblyVersion = "1.5.21";       
-        public const string FileVersion = "1.5.21" + ".0";    
+        public const string AssemblyVersion = "1.5.22";       
+        public const string FileVersion = "1.5.22" + ".0";
```
michaelcfanning: How'd this get updated? Why don't I see a chance somewhere to SetCurrentVersion.cmd?

src/Sarif.Driver/Sdk/SkimmerBase.cs
lgolding: Sure, that sounds good.

---
In reply to: [62771555](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62771555) [](ancestors = 62771555)

michaelcfanning:
Generally, this looks like great forward progress. Please review my comments and act as you see fit. I'm comfortable with this going in as a partial implementation while you wrap up other loose ends.

src/Sarif/VersionConstants.cs
lgolding: No idea. I didn't touch it. Is it possible you checked in a change to SetCurrentVersion.cmd but didn't check in the corresponding changes to the generated files?

---
In reply to: [62771652](https://github.com/Microsoft/sarif-sdk/pull/257#discussion_r62771652) [](ancestors = 62771652)

lgolding:
As we discussed: I'll merge, then fix the other small PR comments, then move on to `SetProperty<T>`.

PR #259, opened 2016-05-11
SetProperty supports arbitrary types

lgolding:
Also: `EnumHint` supports the `Flags` attribute, and allows enum values to be explicitly specified. So we don't need NotYetGenerated\SuppressionStates.cs any more.

Note: The enum values are generated as decimal (`1`, `2`) rather than hex (`0x1`, `0x2`). This suffices, but we can always make the generator support an option to emit hex values if we care enough.

@michaelcfanning


src/Sarif/Autogenerated/SuppressionStates.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// The state of a result relative to a baseline of a previous run.
     /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.28.0.0")]
     [Flags]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.30.0.0")]
```
lgolding: `[Flags]` is now auto-generated by an argument to the `EnumHint`. (The attributes happen to come out in this order. I don't think it's worth controlling the order.)

src/Sarif/Autogenerated/SuppressionStates.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.CodeAnalysis.Sarif
...
     public enum SuppressionStates
     {
         None,
-        SuppressedInSource = 0x1,
-        SuppressedInBaseline = 0x2
+        SuppressedInSource = 1,
+        SuppressedInBaseline = 2
```
lgolding: `EnumHint` also now supports the ability to specify enum values. They always come out in decimal, but we can give `EnumHint` the ability to emit them in hex if we care enough.

src/Sarif/CodeGenHints.json
```diff
@@ -114,8 +114,8 @@
       "arguments": {
         "typeName": "AlgorithmKind",
         "description": "Values specifying different hashing algorithms.",
-        "zeroValue": "Unknown",
-        "enumValues": [
+        "zeroValueName": "Unknown",
+        "enumNames": [
```
lgolding: I changed these names when I introduced the `enumValues` argument (see below) to specify the numeric values.

src/Sarif/CodeGenHints.json
```diff
@@ -336,11 +336,13 @@
...
-        "zeroValue": "None",
-        "enumValues": [
+        "flags": true,
+        "zeroValueName": "None",
+        "enumNames": [
           "SuppressedInSource",
           "SuppressedInBaseline"
-        ]
+        ],
+        "enumValues": [1, 2]
```
lgolding: Here's the new thing.

src/Sarif/NotYetAutogenerated/SuppressionStates.cs
```diff
@@ -1,20 +0,0 @@
-// Copyright (c) Microsoft.  All Rights Reserved.
```
lgolding: No longer needed since it's autogenerated.

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -106,5 +81,37 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
 
             writer.WriteEndObject();
         }
+
+        public static JTokenType DotNetTypeToJTokenType(Type dotNetType)
+        {
+            JTokenType jTokenType = JTokenType.None;
```
lgolding: Having factored this out so that it could also be used by `PropertyBagHolder.SetProperty` (see above), I now see that it doesn't handle the case where `SetProperty` is called with anything other than one of the Json.NET types -- it doesn't handle an arbitrary class. So I have more work to do here and in `SetProperty`. You can defer finishing the review until I've addressed this.

src/Sarif.UnitTests/Readers/PropertyBagConverterTests.cs
```diff
@@ -5,11 +5,10 @@
 using System.Collections.Generic;
 using System.Linq;
 using FluentAssertions;
-using Microsoft.CodeAnalysis.Sarif.Readers;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
 
-namespace Microsoft.CodeAnalysis.Sarif.Core
+namespace Microsoft.CodeAnalysis.Sarif.Readers
```
lgolding: Fixed namespace because I'd previously moved the file to a different folder but forgot to adjust namespace.

src/Sarif.UnitTests/Readers/PropertyBagConverterTests.cs
```diff
@@ -25,7 +24,7 @@ public class PropertyBagConverterTests
         private ConverterTestClass _roundTrippedObject;
 
         [TestMethod]
-        public void PropertyBagConverter_RoundTripsStringValuedProperty()
+        public void PropertyBagConverter_RoundTripsStringProperty()
```
lgolding: Just renames.

src/Sarif/Readers/PropertyBagConverter.cs
lgolding: OK, this actually works as is. Turns out I don't need to save the actual token type. All I need to save is whether the token is a string or not, so I know whether or not to add an extra pair of quotes around it when I serialize it out. Everything else just works.

src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        const string PropertyName = "prop";
+        const string Input = "{\"properties\":{\"" + PropertyName + "\":12}}";
+
+        internal static void ShouldSerializeAs<T>(this T value, string serializedValue)
+        {
+            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
+            {
+                ContractResolver = SarifContractResolver.Instance
+            };
```
lgolding: This is what makes `DateTime` formatting "just work".

src/Sarif/Core/IPropertyBagHolder.cs
```diff
@@ -12,7 +12,7 @@ public interface IPropertyBagHolder
     {
         string GetProperty(string propertyName);
         T GetProperty<T>(string propertyName);
-        void SetProperty(string propertyName, string value);
+        void SetProperty<T>(string propertyName, T value);
```
lgolding: `SetProperty` can now take any type.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -43,15 +41,9 @@ public bool TryGetProperty(string propertyName, out string value)
 
         public string GetProperty(string propertyName)
         {
-            if (Properties[propertyName].JTokenType != JTokenType.String)
+            if (!Properties[propertyName].IsString)
```
lgolding: As we discussed, we don't have to remember the "token type"; we just have to remember if we started out with a string.

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -31,64 +31,24 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             var propertyDictionary = new Dictionary<string, SerializedPropertyInfo>();
             foreach (string key in objectDictionary.Keys)
             {
-                JTokenType jTokenType = JTokenType.Undefined;
                 Type propertyType = objectDictionary[key].GetType();
 
-                if (propertyType == typeof(JObject))
```
lgolding: Again, we don't need the detailed logic to map between .NET types and JSON token types.

src/Sarif/CodeGenHints.json
michaelcfanning: 'memberNames' and 'memberValues' might have been clearer. take this comment or leave it. :)

---
In reply to: [62882321](https://github.com/Microsoft/sarif-sdk/pull/259#discussion_r62882321) [](ancestors = 62882321)

src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        [TestMethod]
+        public void PropertyBagHolder_SetProperty_SetsObjectProperty()
+        {
+            new TestObjectClass(42, "abc").ShouldSerializeAs("{\"n\":42,\"s\":\"abc\"}");
+        }
+
+        [TestMethod]
+        public void PropertyBagHolder_SetProperty_SetsGuidProperty()
+        {
+            const string GuidString = "{12345678-1234-1234-1234-123456780abc}";
```
michaelcfanning: weird guid value. this last bit should be '1234567890ab' :)

src/Sarif/CodeGenHints.json
lgolding: Yes, would be better. Requires a JSchema code generator change so will handle in a separate PR.

---
In reply to: [62932360](https://github.com/Microsoft/sarif-sdk/pull/259#discussion_r62932360) [](ancestors = 62932360,62882321)

src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs
lgolding: Fixed.

---
In reply to: [62933258](https://github.com/Microsoft/sarif-sdk/pull/259#discussion_r62933258) [](ancestors = 62933258)

michaelcfanning:
Looks good! +1

lgolding:
Awesome. Will push the GUID change, make a note of the EnumHint argument name changes, and merge.

---
In reply to: [218601941](https://github.com/Microsoft/sarif-sdk/pull/259#issuecomment-218601941) [](ancestors = 218601941)

michaelcfanning:
Beautiful. Next requirement: tags helper that dips property that dips into properties member. I am cleaning up the PREfast converter. Getting close.

PR #260, opened 2016-05-11
Optional unique identifier for a result. Intended to permit correlati…

michaelcfanning:
…on of results with external data (such as side-car annotation files). @lgolding (please merge if this looks right).

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -565,6 +565,11 @@
           "type": "string"
         },
 
+        "id": {
+          "description": "A unique identifer for the result.",
+          "type": "string"
+        },
+
         "toolFingerprint": {
           "description": "A string that contributes to the unique identity of the result.",
           "type": "string"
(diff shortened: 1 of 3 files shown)


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -565,6 +565,11 @@
           "type": "string"
         },
 
+        "id": {
+          "description": "A unique identifer for the result.",
+          "type": "string"
+        },
+
```
lgolding: Tough to decide whether to put this first (typically things put their ID properties first, ID columns in databases tend to come first...) or whether to put it with things its most-likely-to-be-used-with, like `toolFingerprint`. Doesn't matter much... I wonder if there's a naming issue. We have `correlationId` for `run`. I wonder if this isn't just plain `id` but `externalDataLinkageId` or the like. Thoughts?

michaelcfanning: I tend to think 'id' could be purposed in several ways, so it's fine to remain general. We also have a log 'id' member and I tend to think result.id name should match that.

PR #261, opened 2016-05-11
Provide optional 'kind' on annotatedCodeLocation that helps describe/…

michaelcfanning:
…categorize the annotation. @lgolding : need a final call on this. We are updating the PREfast converter, which will populate annotatedCodeLocation.kind if it exists. it's a near call but given that we intend to provide first-class support in a viewer experience that will consume this data, I am comfortable standardizing on this set. what's your thought?

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -40,6 +40,11 @@
           "type": "string"
         },
 
+        "kind": {
+          "description": "A descriptive identifier that categorizes the annotation.",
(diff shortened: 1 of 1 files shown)


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -244,7 +249,7 @@
         },
 
         "algorithm": {
-          "description": "A string specifying the name of the algorithm used to compute the hash value specified in the 'value' property.",
+          "description": "The name of the algorithm used to compute the hash value specified in the 'value' property.",
```
lgolding: Yay :)

lgolding:
:+1:

PR #262, opened 2016-05-11
Embed tags in property bag.

michaelcfanning:
@lgolding


lgolding:
As discussed, we can't DRY out the property bag "type" right now because the code generator won't emit a dictionary if you do.

PR #263, opened 2016-05-11
Schema changes for properties.tags, and EnumHint argument renames

lgolding:
@michaelcfanning 

This PR upgrades sarif-sdk to the next version of JSchema (0.31.0) to pick up the necessary change to `DictionaryHint` to support the schema change (allowing an object with properties to be typed as a dictionary).

0.31.0 also includes the argument name changes on `EnumHint` (`memberNames`, `memberValues`).


michaelcfanning:
+1 looks nice. let's talk next about the Tags helper we provide. If you call SetProperty<HashSet<string>> do the right things magically happen (as an interim solution to the helper property)?

lgolding:
Yes, I have a unit test for that:

        [TestMethod]
        public void PropertyBagHolder_SetProperty_SetsStringArrayPropertyFromHashSet()
        {
            new HashSet<string> { "ab", "cde" }.ShouldSerializeAs("[\"ab\",\"cde\"]");
        }

So for example:

        SetProperty("tags", new HashSet<string>{ "js", "spelling" });

But nothing stops you from deserializing it into any other container (and that will continue to be true even when we add the tag-handling convenience APIs).

        string[] tags = GetProperty<string[]>("tags");

And of course SetProperty is not convenient for updates because it replaces the entire array. I suggest

        HashSet<string> GetTags();
        bool HasTag(string tag);
        void SetTags(IEnumerable<string> tags);
        void ClearTags();
        void AddTag(string tag);
        void AddTags(IEnumerable<string> tags);
        void RemoveTag(string tag);
        void RemoveTags(IEnumerable<string> tags);

Internally AddTag/AddTags can remove duplicates before serializing, and GetTags will remove them when deserializing (since it returns a HashSet).

Sound ok?

---
In reply to: [218623448](https://github.com/Microsoft/sarif-sdk/pull/263#issuecomment-218623448) [](ancestors = 218623448)

lgolding:
The absolute minimum is

        HashSet<string> GetTags();
        void SetTags(IEnumerable<string>);

so I'll start with those.

---
In reply to: [218776381](https://github.com/Microsoft/sarif-sdk/pull/263#issuecomment-218776381) [](ancestors = 218776381,218623448)

michaelcfanning:
My suggestion is to create a Tags class that implements ISet<string>. Expose this item as a property on the classes that provide a property bag object. This type should have a single parameterized constructor that accepts a property bag instance, which it wraps. All ISet operations should do the appropriate thing with the backing store.

PR #267, opened 2016-05-13
Finish Tags implementation

lgolding:
@michaelcfanning


michaelcfanning:
+1 let's get this in!

PR #270, opened 2016-05-13
Add unit test for RuleDictionaryConverter

lgolding:
Also:
* DRY out tests

### DIFF
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Microsoft.CodeAnalysis.Sarif.Readers;
@@ -115,6 +116,22 @@ namespace Microsoft.CodeAnalysis.Sarif
                 : JsonConvert.SerializeObject(value);
              
             Properties[propertyName] = new SerializedPropertyInfo(serializedValue, isString);
+        }
+
+        public void SetPropertiesFrom(IPropertyBagHolder other)
+        {
+            // We need the concrete class because the IPropertyBagHolder interface
+            // doesn't expose the raw Properties array.
+            PropertyBagHolder otherHolder = other as PropertyBagHolder;
+            Debug.Assert(otherHolder != null);
+                
+            Properties.Clear();
+            foreach (string propertyName in other.PropertyNames)
+            {
+                SerializedPropertyInfo otherInfo = otherHolder.Properties[propertyName];
+
+                Properties[propertyName] = new SerializedPropertyInfo(otherInfo.SerializedValue, otherInfo.IsString);
+            }
         }
 
         public void RemoveProperty(string propertyName)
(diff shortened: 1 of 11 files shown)


src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -117,6 +118,22 @@ public void SetProperty<T>(string propertyName, T value)
             Properties[propertyName] = new SerializedPropertyInfo(serializedValue, isString);
         }
 
+        public void SetPropertiesFrom(IPropertyBagHolder other)
```
lgolding: Oh right, this is the reason for this whole PR. Adding this API to ensure properties of all types (not just strings) can be copied from one holder to another, as required by the RuleDictionaryConverter.

src/Sarif/Readers/RuleDictionaryConverter.cs
```diff
@@ -6,9 +6,13 @@
...
 {
-    public class IRuleDictionaryConverter : JsonConverter
+    /// <summary>
+    /// This converter only exists because the version of Json.NET we're using (6.0.8)
+    /// can't write out a property whose static type is an interface.
+    /// </summary>
+    public class RuleDictionaryConverter : JsonConverter
     {
-        public static readonly IRuleDictionaryConverter Instance = new IRuleDictionaryConverter();
+        public static readonly RuleDictionaryConverter Instance = new RuleDictionaryConverter();
```
lgolding: Renamed because by .NET naming guidelines, something starting with `I` is an interface.

src/Sarif/Readers/RuleDictionaryConverter.cs
```diff
@@ -47,10 +51,7 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                     ShortDescription = iRule.ShortDescription,
                 };
 
-                foreach (string propertyName in iRule.PropertyNames)
-                {
-                    rule.SetProperty(propertyName, iRule.GetProperty(propertyName));
-                }
+                rule.SetPropertiesFrom(iRule);
```
lgolding: Here's where the new API is used.

src/Sarif/Readers/RuleDictionaryConverter.cs
michaelcfanning: I think the real (verbose) convention is to specify both the key and value, so StringIRuleDictionaryConverter would be the name. 'RuleDictionaryConverter' is fine with me, though (and preferable to previous.

---
In reply to: [63249962](https://github.com/Microsoft/sarif-sdk/pull/270#discussion_r63249962) [](ancestors = 63249962)

michaelcfanning:
+1 nice change.

src/Sarif/Readers/RuleDictionaryConverter.cs
lgolding: I don't see that in "Framework Design Guidelines," 2nd Ed., Sec. 3.5.2, "Names of common types". It just says

IDictionary or IDictionary<TKey, TValue>: DO add the suffix "Dictionary".

---
In reply to: [63250824](https://github.com/Microsoft/sarif-sdk/pull/270#discussion_r63250824) [](ancestors = 63250824,63249962)

lgolding:
Thanks

---
In reply to: [219159219](https://github.com/Microsoft/sarif-sdk/pull/270#issuecomment-219159219) [](ancestors = 219159219)

PR #271, opened 2016-05-13
Populate tool.sarifLoggerVersion property.

michaelcfanning:
@lgolding @rtaket

### DIFF
--- a/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
+++ b/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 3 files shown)


src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
```
lgolding: What is this for?

lgolding:
:+1: with a question.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
michaelcfanning: FileVersionInfo lives in this namespace. Odd but true.

---
In reply to: [63250501](https://github.com/Microsoft/sarif-sdk/pull/271#discussion_r63250501) [](ancestors = 63250501)

PR #272, opened 2016-05-13
location.decoratedName, annotatedCodeLocation.essential, ACL.Kind enum hint

lgolding:
@michaelcfanning

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -49,6 +49,11 @@
           "enum": [ "branch", "call", "callReturn", "declaration", "functionEnter", "functionExit", "sanitizer", "sink", "source", "usage" ]
         },
 
+        "essential": {
+          "description": "True if this location is essential to understanding the code flow in which it occurs.",
+          "type": "boolean"
+        },
+
         "properties": {
           "description": "Key/value pairs that provide additional information about the code location.",
           "type": "object",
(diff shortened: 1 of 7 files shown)


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -386,6 +391,11 @@
           "type": "string"
         },
 
+        "decoratedName": {
+          "description": "The fully qualified name for the logical location, in the compiler's native format.",
```
michaelcfanning: 'The non-readable fully qualified name of the location where the analysis tool produced the result, such as the mangled function name provided by a C++ compiler that encodes calling convention, return type and other details along with the function name.'

Also suggest adding the qualifier 'readable' to the 'fullyQualifiedLogicalName' description. What do you think?

michaelcfanning: Do what you think is best, then +1. I'm also glad to review your revision if you'd prefer before merging. 

---
In reply to: [63254879](https://github.com/Microsoft/sarif-sdk/pull/272#discussion_r63254879) [](ancestors = 63254879)

lgolding: I'm ok with all that. I'd like to describe the FQLN as "human-readable" and the decoratedName as "machine-readable". I'm a little uncomfortable with "non-readable".

Also I'm sure someone will take offense at "mangled" but I'm prepared to defend it ;-).

---
In reply to: [63254920](https://github.com/Microsoft/sarif-sdk/pull/272#discussion_r63254920) [](ancestors = 63254920,63254879)

PR #275, opened 2016-05-14
Autogenerate SarifLog $schema property

lgolding:
Upgrade to JSchema 0.32.0, which introduces `PropertyNameHint`. This allows us to map the JSON schema property `sarifLog.$schema` to the .NET property `SarifLog.SchemaUri`. The `SarifLog` files in the `NotYetGenerated` directory are no longer necessary.

@michaelcfanning


src/Sarif/Autogenerated/SarifLog.cs
```diff
@@ -31,10 +31,7 @@ public SarifNodeKind SarifNodeKind
             }
         }
 
-        /// <summary>
-        /// The URI of the JSON schema corresponding to <see cref="Version"/>.
-        /// </summary>
-        [DataMember(Name="$schema")]
+        [DataMember(Name = "$schema", IsRequired = false, EmitDefaultValue = false)]
```
lgolding: The mapping is now generated automatically. The doc comment needs to be added to the schema.

src/Sarif/Autogenerated/SarifLogEqualityComparer.cs
```diff
@@ -10,8 +10,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Defines methods to support the comparison of objects of type SarifLog for equality.
     /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.31.0.0")]
-    public sealed class SarifLogEqualityComparer : IEqualityComparer<SarifLog>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.32.0.0")]
+    internal sealed class SarifLogEqualityComparer : IEqualityComparer<SarifLog>
```
lgolding: Here's an example of how the files in the `NotYetGenerated` directory drifted out of sync from the generated files. When we made the equality comparers internal, we neglected to fix this one.

src/Sarif/Autogenerated/SarifLogEqualityComparer.cs
```diff
@@ -84,7 +84,7 @@ public int GetHashCode(SarifLog obj)
                         result = result * 31;
                         if (value_0 != null)
                         {
-                            result = (result * 31) + value_0.GetHashCode();
+                            result = (result * 31) + value_0.ValueGetHashCode();
```
lgolding: Here's another example, and this one's a bug.

This is why I'm so adamant about not letting files linger in `NotYetAutogenerated`. It's not just debt, it can cause bugs, and it's just too hard to make sure it's kept up to date.

src/Sarif/CodeGenHints.json
```diff
@@ -429,6 +429,14 @@
...
     }
   ],
+  "SarifLog.$schema": [
+    {
+      "kind": "PropertyNameHint",
+      "arguments": {
+        "dotNetPropertyName": "SchemaUri"
+      }
+    }
+  ],
```
lgolding: The new hint

michaelcfanning:
+1

PR #277, opened 2016-05-16
Provide Properties instance in the PropertyBagHolder base class. 

michaelcfanning:
Also fix up auto-generated code to conform to final schema. @rtaket @lgolding


lgolding:
:+1:

PR #280, opened 2016-05-16
Binskim update

michaelcfanning:
@lgolding . Enabled all binskim tests and refreshed then. Fixed serialization issue where we assumed a non-null Properties instance on Rule class.


lgolding:
:+1:

PR #282, opened 2016-05-17
Convert logicalLocationComponents array to logicalComponent (each of …

michaelcfanning:
…which refers to a parent component).

@lgolding Please review this schema change. I'd like to publish it to schemastore. Then I will work on the FxCop converter.

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -442,8 +442,8 @@
           "type": "string"
         },
 
-        "parent": {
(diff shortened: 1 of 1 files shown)


michaelcfanning:
@lgolding couple of points:
'name' refers to the discrete innermost name of the thing the logical location points to. note that we don't recapitulate the fully qualified name in this construct (we didn't preserve the FQN in the previous approach either).

'parentKey' is the identifier that can be used to find the immediate parent within the dictionary. a more descriptive but lengthier name would be 'parentLogicalLocationKey'

PR #284, opened 2016-05-18
File data

michaelcfanning:
Physical locations dictionary now matches logical locations dictionary. Each dictionary value is a discrete node, which potentially participates in a nested file location. Each value refers to either a root value or has a parentKey that identifies the nested file parent. @lgolding @rtaket

### DIFF
--- a/src/Sarif.UnitTests/Readers/AlgorithmKindConverterTests.cs
+++ b/src/Sarif.UnitTests/Readers/AlgorithmKindConverterTests.cs
@@ -69,16 +69,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Readers
         ""name"": null
       },
       ""files"": {
-        ""http://abc/"": [
-          {
-            ""hashes"": [
-              {
-                ""value"": null,
-                ""algorithm"": """ + testTuple.Item2 +  @"""
-              }
-            ]
-          }
-        ]
+        ""http://abc/"": {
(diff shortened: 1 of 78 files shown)


src/Sarif.UnitTests/Readers/AlgorithmKindConverterTests.cs
```diff
@@ -95,11 +93,9 @@ public void AlgorithmKind_AllMembers()
 
                     uut.WriteTool(DefaultTool);
 
-                    var files = new Dictionary<string, IList<FileData>>
+                    var file = new Dictionary<string, FileData>
```
lgolding: It's still a container even though it's only got one thing in it.

src/Sarif/IResultLogWriter.cs
```diff
@@ -42,7 +42,7 @@ public interface IResultLogWriter
         /// A dictionary whose keys are the strings representing the locations of scanned files
         /// and whose values provide information about those files.
         /// </param>
-        void WriteFiles(IDictionary<string, IList<FileData>> fileDictionary);
+        void WriteFiles(IDictionary<string, FileData> filesDictionary);
```
lgolding: Why this rename?

lgolding:
:+1: with a couple of naming quibbles.

My main concern is I don't see any tests for `parentKey`.

PR #290, opened 2016-05-19
Allow member count mismatch on EnumHint

lgolding:
In the latest SARIF schema change, we removed two values from the enum set for `rule.defaultLevel`. We still want the type of the .NET property `Rule.DefaultLevel` to be `ResultLevel`, even though the `ResultLevel` enum has two more values than `Rule.DefaultLevel` now permits. But until now, the JSchema code generator would throw an exception if an `EnumHint` specified a different number of values from the JSON schema.

So we upgrade to JSchema 0.34.0. This version allows an `EnumHint` to specify an argument `"allowMemberCountMismatch"`. When this argument is specified with a value `true`, the code generator allows the mismatch.

@michaelcfanning


michaelcfanning:
GTG (good to go) :)

PR #291, opened 2016-05-19
Notifications

michaelcfanning:
@lgolding @rtaket 

Remove 'module' as required from 'stackFrame'. Move notes to tool notifications. Correct output of 'configuration missing' rule.


src/Sarif/ConsoleLogger.cs
```diff
@@ -55,7 +55,7 @@ public void AnalyzingTarget(IAnalysisContext context)
             if (this.Verbose)
             {
                 Console.WriteLine(string.Format(
-                    SdkResources.MSG1001_AnalyzingTarget,
+                    SdkResources.MSG001_AnalyzingTarget,
```
lgolding: I don't see the difference on this line

lgolding: Oh! 1001 -> 001

lgolding:
:+1:

PR #296, opened 2016-05-20
Provide uri validation and improved uri serialization/deserialization

michaelcfanning:
@lgolding @rtaket @nguerrera

### DIFF
--- a/src/Sarif.UnitTests/Readers/UriConverterTests.cs
+++ b/src/Sarif.UnitTests/Readers/UriConverterTests.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using FluentAssertions;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class UriConverterTests
+    {
+        [Theory]
+        [InlineData("/relative/path", "/relative/path")]
+        [InlineData("standalone", "standalone")]
+        [InlineData("relative/path/", "relative/path/")]
+        [InlineData("http://abc", "http://abc/")]
+        [InlineData(@"file:///c:/space space", "file:///c:/space%20space")]
+        public void UriConverter_UriToString(string input, string expected)
+        {
+            string actual = UriConverter.UriToString(new Uri(input, UriKind.RelativeOrAbsolute));
+            actual.Should().Be(expected);
+        }
+    }
+}
--- a/src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
+++ b/src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using FluentAssertions;
+using System.Text;
+using Newtonsoft.Json;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    [TestClass]
+    public class ValidatingVisitorTests
+    {
+        [TestMethod]
+        public void ValidatingVisitor_InvalidUri()
+        {
+            // NOTE that for validating tests, we need to deserialize SARIF text 
+            // without benefit of the many converters that transform specified
+            // JSON values to enum equivalents, etc. The Sarif text below and 
+            // for all these tests is rendered, in some cases, as it the converter
+            // had done some processing before conversion to eventual type. We
+            // express the version as OneZeroZeroBeta5, for example, because the
+            // default JSON.NET enum conversion will do the right thing as far
+            // as converting this value to the C# OM.
+            string input =
+@"{
+  ""version"": ""OneZeroZeroBetaFive"",
+  ""runs"": [{
+      ""tool"": { ""name"": ""toolName"" },
+      ""rules"" : { ""id1"" : { ""id"" : ""id1"", ""helpUri"" : ""??[][][]??""} },
+      ""files"": { ""!!not a valid uri!!"" : { ""uri"" : ""!!not a valid uri!!""} },
+      ""results"": [{
+            ""locations"": [{""resultFile"": { ""uri"": ""file:///C:/space space.cpp"" }}],
+            ""stacks"": [{ ""frames"" : [{ ""uri"" : ""c:\\test.cpp"", ""fullyQualifiedLogicalName"" : ""fqn"" }]}],
+            ""fixes"": [{ ""fileChanges"": [{ ""uri"" : ""file://c:/missing/slash"", ""replacements"" :[{ ""offset"" : 0, ""deletedLength"" : 1 }] }], ""description"" : ""ok""}]
+        }]
+      }]
+}";
+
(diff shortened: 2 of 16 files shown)


src/Sarif/SdkResources.resx
```diff
@@ -294,4 +294,7 @@
   <data name="PropertyDoesNotExist" xml:space="preserve">
     <value>Property '{0}' does not exist. Consider calling TryGetProperty instead.</value>
   </data>
+  <data name="SARIF001_InvalidUri" xml:space="preserve">
+    <value>'uri' member of '{0}' instance is not valid: '{1}'</value>
```
lgolding: You'll know the name of the member you're validating, and it won't always be "uri", so I suggest parameterizing this and passing nameof(Foo) to string.Format.

src/Sarif.UnitTests/Readers/UriConverterTests.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using FluentAssertions;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class UriConverterTests
+    {
+        [Theory]
```
lgolding: I had no idea that VS would execute both XUnit tests and MSTest tests in the same assembly.

src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
```
lgolding: Sort usings

src/Sarif/Visitors/ValidatingVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+            _sarifLog = new SarifLogger(writer, verbose: true, tool: tool);
+        }
+
+        public override FileData VisitFileData(FileData node)
+        {
+            ValidateUri(node.Uri);
+            return base.VisitFileData(node);
+        }
+
+        public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
```
lgolding: Missing VisitRun to validate analysisTargetUri.

src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
```diff
@@ -0,0 +1,64 @@
...
+      ""files"": { ""!!not a valid uri!!"" : { ""uri"" : ""!!not a valid uri!!""} },
+      ""results"": [{
+            ""locations"": [{""resultFile"": { ""uri"": ""file:///C:/space space.cpp"" }}],
+            ""stacks"": [{ ""frames"" : [{ ""uri"" : ""c:\\test.cpp"", ""fullyQualifiedLogicalName"" : ""fqn"" }]}],
+            ""fixes"": [{ ""fileChanges"": [{ ""uri"" : ""file://c:/missing/slash"", ""replacements"" :[{ ""offset"" : 0, ""deletedLength"" : 1 }] }], ""description"" : ""ok""}]
+        }]
+      }]
+}";
+
+            // TODO do we want to validate that http:// Uris have a trailing slash?
```
lgolding: No. RFC 3986 doesn't require a trailing slash. "[http://example.com/data](http://example.com/data)" and "[http://example.com/data](http://example.com/data)/" are both valid URIs. They might even point to different resources (although many web servers will redirect the one to the other).

src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
```diff
@@ -0,0 +1,64 @@
...
+    [TestClass]
+    public class ValidatingVisitorTests
+    {
+        [TestMethod]
+        public void ValidatingVisitor_InvalidUri()
+        {
+            // NOTE that for validating tests, we need to deserialize SARIF text 
+            // without benefit of the many converters that transform specified
+            // JSON values to enum equivalents, etc. The Sarif text below and 
+            // for all these tests is rendered, in some cases, as it the converter
```
lgolding: "as it" => "as if"

src/Sarif/Readers/UriConverter.cs
```diff
@@ -26,12 +26,18 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+        internal static string UriToString(Uri uri)
+        {
+            if (uri.IsAbsoluteUri)
             {
-                uri = new Uri(uri.ToString());
+                return uri.AbsoluteUri;
             }
-
-            writer.WriteValue(uri.ToString());
+            return uri.ToString();
```
nguerrera: This won't necessarily be escaped. See https://github.com/dotnet/roslyn/commit/c66413016d28dde42a1cd6cc669b3c4d47835deb

src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
lgolding: My point being, we can't know that a particular URI we find in a file would behave the same if a slash were added to it, so our validator can't insist on it.

---
In reply to: [64092768](https://github.com/Microsoft/sarif-sdk/pull/296#discussion_r64092768) [](ancestors = 64092768)

src/Sarif/Visitors/ValidatingVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+        public override FileChange VisitFileChange(FileChange node)
+        {
+            ValidateUri(node.Uri);
+            return base.VisitFileChange(node);
+        }
+
+        private void ValidateUri(Uri uri)
+        {
+            if (!uri.IsWellFormedOriginalString())
+            {
```
nguerrera: I don't get this validation. It will, for example, trip on new Uri(@"C:\MyPath\MyFile.cs"). I would think that producers would do that for convenience. This will force them to do new Uri(new Uri(@"c:\MyPath\MyFile.cs").AbsoluteUri) or am I missing something?

src/Sarif.UnitTests/Visitors/ValidatingVisitorTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            // had done some processing before conversion to eventual type. We
+            // express the version as OneZeroZeroBeta5, for example, because the
+            // default JSON.NET enum conversion will do the right thing as far
+            // as converting this value to the C# OM.
```
lgolding: Ohh! Lightbulb. I had to read this several times before I understood. You want to examine the result of deserializing a SARIF file without the benefit of any of our converters that might "help out" and mask problems in the raw file. But if (for example) sarifLog.version had a "real" value like "1.0.0", Json.NET couldn't deserialize it into a property whose type is SarifVersion. So you fake up a file that *can* be deserialized without benefit of our converters.

Ok, I get that, but how will this work in real life? A hypothetical command line tool that performs this kind of detailed validation is going to be deserializing real files with values like "version": "1.0.0". So the tool will need to use a contract resolver that does at least some of what the "real" resolver does, like convert enums.

So there are two choices. Either the tool uses the "real" resolver, and maybe misses a chance to point out some problems, or it creates its own resolver that uses only a subset of the converters used by the real resolver -- like maybe only the enum converter.

Either way, the tests ought to be doing what a "real" validation tool would do.

lgolding:
I suggest closing this PR since it's cluttering the open PR list and it doesn't look like we're taking this right now.

michaelcfanning:
I left this open by design because I intend to preserve half of the changes and want to track the PR feedback notes. Any problem leaving it open for a bit? Will close on this by EOW.

lgolding:
No problem with that.

PR #299, opened 2016-05-23
Update WebSkim files to reflect changes to nested files. Enable webskim tests.

michaelcfanning:
@lgolding 

webskim updated with all SDK changes. Look at WebSkim population of nested files, the deepest tool usage we have for this concept. WebSkim in one case is analyzing an appx bundle, that includes an appx, that includes many analysis targets.


lgolding:
That all looks right. :+1:

PR #301, opened 2016-05-23
Upgrade to Json.Schema.ToDotNet 0.35.0

lgolding:
This is the version of Json.Schema.ToDotNet that corresponds to the new Json.Schema 0.35.0, which has the new validator tool. But we're not actually pulling down Json.Schema at all -- yet.

@michaelcfanning


michaelcfanning:
Not sure why the last schema changes are flowing in as deltas for this one. :) but you are gtg

PR #303, opened 2016-05-24
WebSkim adds snippets to format and exercises its right to generate a…

michaelcfanning:
… synthesized ruleKey (which packages both the opaque and readable ids).
@lgolding 
@rtaket

### DIFF
--- a/src/Sarif/Writers/SarifLogger.cs
+++ b/src/Sarif/Writers/SarifLogger.cs
@@ -188,17 +188,19 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
             {
                 _issueLogJsonWriter.CloseResults();
 
-                if (_run.ConfigurationNotifications != null)
+                if (_run != null && _run.ConfigurationNotifications != null)
                 {
                     _issueLogJsonWriter.WriteConfigurationNotifications(_run.ConfigurationNotifications);
                 }
 
-                if (_run.ToolNotifications != null)
+                if (_run != null && _run.ToolNotifications != null)
                 {
                     _issueLogJsonWriter.WriteToolNotifications(_run.ToolNotifications);
                 }
 
-                if (_run?.Invocation?.StartTime != new DateTime())
+                if (_run != null && 
+                    _run.Invocation != null &&
+                    _run.Invocation.StartTime != new DateTime())
                 {
                     _run.Invocation.EndTime = DateTime.UtcNow;
                 }
@@ -211,9 +213,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                     _issueLogJsonWriter.WriteRules(_rules);
                 }
 
-                if (_run.Files != null) { _issueLogJsonWriter.WriteFiles(_run.Files); }
+                if (_run != null && _run.Files != null) { _issueLogJsonWriter.WriteFiles(_run.Files); }
 
-                if (_run.Invocation != null)
+                if (_run != null && _run.Invocation != null)
                 {
                     _issueLogJsonWriter.WriteInvocation(invocation: _run.Invocation);
                 }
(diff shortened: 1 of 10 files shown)


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -188,17 +188,19 @@ public void Dispose()
             {
                 _issueLogJsonWriter.CloseResults();
 
-                if (_run.ConfigurationNotifications != null)
+                if (_run != null && _run.ConfigurationNotifications != null)
```
lgolding: Consider null conditional operator: `if (_run?.ConfigurationNotifications != null)`

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -241,14 +243,24 @@ public void AnalysisStopped(RuntimeConditions runtimeConditions)
 
         public void Log(IRule rule, Result result)
         {
+            if (rule.Id != result.RuleId)
+            {
+                //The rule id '{0}' specified by the result does not match the actual id of the rule '{1}'
+                string message = string.Format(SdkResources.ResultRuleIdDoesNotMatchRule,
+                    result.RuleId.ToString(),
+                    rule.Id.ToString());
```
lgolding: No need to call `ToString()` on arguments passed to `String.Format`.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -241,14 +243,24 @@ public void AnalysisStopped(RuntimeConditions runtimeConditions)
...
+
             if (!ShouldLog(result.Level))
             {
                 return;
             }
 
             if (rule != null)
             {
-                Rules[rule.Id] = rule;
+                Rules[result.RuleKey ?? result.RuleId] = rule;
```
lgolding: This means you could end up with the same rule in the dictionary multiple times (if some results for a given rule have a `ruleKey` and others don't). Unlikely; how serious if it happens?

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: I don't ever call string.format(object), instead, I explicitly convert all arguments to string before making the call, so that reviewers can understand what culture-specific operations, if any, are performed. These ToString() calls are still a bug, though, as the 'Id' properties are already strings. :P

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: I could throw here or the logger could automatically disambiguate these items (would require mutating the passed in result as well). preference? (I will do this in a future change, btw)

src/Sarif/Writers/SarifLogger.cs
lgolding: ? `string.Format` calls `ToString()` on its arguments automatically (and it uses whatever `string.Format` was called with, in this case `CurrentCulture` by default (although FxCop thinks, and I agree, that you should specify it explicitly). So this works:

```
string message = string.Format(
                    CultureInfo.CurrentCulture,
                    SdkResources.ResultRuleIdDoesNotMatchRule,
                    result.RuleId,
                    rule.Id);
```

... whether or not the arguments happen to be strings. Am I misunderstanding you?

src/Sarif/Writers/SarifLogger.cs
lgolding: My preference would be to throw if a rule is listed once with its id and one or more times by keys. (Multiple keys are fine, of course, that's why `ruleKey` exists.)

michaelcfanning: ok, there is a subtlety here. if the user provides a result.RuleKey, this suggests that are professing to handle the uniqueness of the dictionary keys. if we consume rule.Id, then the logger probably should handle this automatically. agree?

PR #304, opened 2016-05-24
Use JSchema for validation; remove dependency on VS JSON libs

lgolding:
Also:

* Remove the Sarif.Validator project (a library which wrapped the VS JSON library) and the SarifValidator project (a command-line driver for Sarif.Validator). JSchema now supplies the equivalent of both of those.

@michaelcfanning


src/Sarif.FunctionalTests/SarifValidatorTests.cs
```diff
@@ -22,38 +23,47 @@ public class SarifValidatorTests
...
 
         [Theory]
         [MemberData(nameof(ConverterTestCases))]
         public void Converter_validation(string inputFile)
         {
-            IEnumerable<JsonError> errors = Validator.ValidateFile(inputFile, JsonSchemaFile);
+            string instanceText = File.ReadAllText(inputFile);
+            var validator = new Validator(_schema);
+
+            string[] errors = validator.Validate(instanceText);
```
lgolding: At present, the JSchema validator just returns an array of formatted error messages. When I modify the command line tool to emit SARIF, I'll probably need the validator to return an array of error objects -- possibly `Result` objects -- similar to what the VS-based validator did. (The VS-based validator returned an array of "`JsonError`" objects which the command-line tool turned into `Result` objects. But there's no reason the JSchema validator couldn't return `Result` objects directly.)

src/Tools/SarifValidator/App.config
```diff
@@ -1,30 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
lgolding: This was the VS-based command line tool.

src/Sarif.Validation/JsonError.cs
```diff
@@ -1,35 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT
```
lgolding: This was the VS-based validation library.

src/Sarif.FunctionalTests/Sarif.FunctionalTests.csproj
```diff
@@ -18,9 +18,9 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.props))\build.props" />
   <PropertyGroup>
     <!-- Override the TargetFrameworkVersion specified in build.props. This project must target
-         .NET 4.6 because the Sarif.Validator project, upon which this project depends, targets
+         .NET 4.5.2 because the Microsoft.Json.Schema package, upon which this project depends, targets
```
michaelcfanning: can't JSON.Schema compile against 4.5.1? The rest of SARIF does. Also, we set this property in build.props, do you don't need TargetFrameworkVersion here at all (except to override it as you've done)

michaelcfanning:
nice! +1 love seeing all the code going away. :)

src/Sarif.FunctionalTests/Sarif.FunctionalTests.csproj
lgolding: JSchema can almost certainly build against 4.5.1; I'll fix it.

I don't understand your second sentence. It seems to recapitulate the comment ("Override the TargetFrameworkVersion...").

lgolding:
Well, the code hasn't exactly gone away; it's just hiding ;-)

PR #305, opened 2016-05-24
Schema: require anyOf "ruleId" or "ruleKey"

lgolding:
@michaelcfanning


lgolding:
We can't do this. Not all tools produce a rule id.

PR #306, opened 2016-05-24
Don't require .NET 4.5.2 for functional tests

lgolding:
Upgrade to Json.Schema 0.37.0, which only requires .NET 4.5.1.

Also, fix a bug I introduced when I got rid of the Sarif.Validation project. That was the only project that was requesting the Json.Schema.ToDotNet package. As a result, ToDotNet wasn't getting pulled down at all, which caused the code generation step in the Sarif project build to fail.

To fix this, I modified the targets file which does the code generation to explicitly do a nuget restore on the required package. The only complication is that nuget restore requires the packages file to be called literally "packages.config", and there was already a packages.config in the Sarif project directory. Rather than adding ToDotNet to Sarif\packages.config (which would have been a bad idea, since that packages file expresses the build dependencies of the Sarif project), I made a separate packages.config file and pushed it into a subdirectory called ToDotNet. Then I moved the targets file down there, too.


src/Sarif/ToDotNet/ToDotNet.targets
```diff
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <NuGetPath>$(MSBuildProjectDirectory)\..\..\.nuget</NuGetPath>
     <JsonSchemaToDotNetPath>..\packages\Microsoft.Json.Schema.ToDotNet.0.36.0\tools\JsonSchemaToDotNet.exe</JsonSchemaToDotNetPath>
```
lgolding: Note that even though I upgraded Json.Schema to 0.37.0, I left Json.Schema.ToDotNet at 0.36.0 (I didn't bother to republish it). ToDotNet actually does still depend on .NET 4.5.2 (because of Roslyn), but we don't have a build-time dependency on ToDotNet; we just run the command line tool to do the code generation.

PR #307, opened 2016-05-25
Webskim locations

michaelcfanning:
@lgolding @rtaket 

I updated Webskim to produce complete SARIF regions.


src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -9,13 +9,20 @@ namespace Microsoft.Sarif.Viewer.Sarif
...
+        public static bool TryGetRule(this Run run, string ruleId, string ruleKey, out IRule rule)
         {
             rule = null;
 
             if (run != null && run.Rules != null)
             {
-                rule = run.Rules.Values.FirstOrDefault(r => r.Id == ruleId);
+                if (ruleKey != null)
+                {
+                    rule = run.Rules[ruleKey];
```
lgolding: This method is called `Try`, but it does a direct dictionary lookup, which might throw.

lgolding:
:+1: except for the bug it `RunExtensions` (need to use `TryGetValue` instead of doing a direct dictionary lookup).

PR #310, opened 2016-05-26
Fix #308: FormatForVisualStudio fails on relative URIs

### LINKED ISSUE #308: FormatForVisualStudio fails on relative URIs
`FormatForVisualStudio` contains this code:

```
                Uri uri = physicalLocation.Uri;
                string path = uri.IsFile ? uri.LocalPath : uri.ToString();
```
The `uri` property is allowed to be relative (combined with a `uriBaseId`). But `Uri.IsFile` throws an exception if `uri` is relative.

lgolding:
Also:
* Check in the OM change resulting from the schema change that removed `Run.AnalysisTargetUri`. Apparently these weren't checked in as part of the commit that changed the schema.
* Fix up the Sarif project file from a previous commit where I moved files around but didn't update the project file. As a result the new files weren't showing up in Solution Explorer, and the old file _was_ still showing up, with a yellow triangle.

@michaelcfanning


michaelcfanning:
+1, thanks for fixing this. would be great to have a non-file scheme absolute URI test, [http://contoso.com/file.html](http://contoso.com/file.html) or some such
manual scan of code suggests this should be fine, will flow through URI.ToString().

lgolding:
Added unit test, merging.

PR #314, opened 2016-05-26
Use new version of JSchema validator that produces SARIF

lgolding:
Upgrade to Json.Schema 0.38.0, in which the validator returns an array of SARIF `Result` objects rather than an array of formatted messages.

This induces a circular dependency in the FunctionalTests project (which, until now, has included the validation tests). FunctionalTests has a project reference to the version of Sarif.dll being built (say, 1.5.25). But FunctionalTests also has a NuGet reference to Json.Schema, which in turn was built by taking a NuGet reference to the *published* version of Sarif.Sdk (say 1.5.24). The two versions of Sarif.dll can't coexist in the same project.

To break this dependency, we move the validation tests into their own project. This project does *not* have a project dependency on Sarif.dll. Instead it has NuGet references both to Json.Schema, and to the matching version of Sarif.Sdk. (We need the explicit NuGet reference to Sarif.Sdk, even though Json.Schema's NuGet package declares a dependency on it, because the validation tests directly use SARIF types such as `Result`.)

It is ok to have the validator tests depend on an older version of Sarif.Sdk because these tests are not (directly) testing the behavior of the SARIF APIs. Rather, they are verifying that the files in the TestData directory conform to the most recent version of the SARIF schema. They happen to do so by creating SARIF `Result` objects. The tests don't care if the `Result` objects don't have the newest "shape"; they just care that the right number of them are produced.

And the reason for doing all of this is so that the SARIF validator command line tool (which is built around the validator object from Json.Schema) can create SARIF output by serializing the `Result` objects to a file.

@michaelcfanning

### DIFF
--- a/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
+++ b/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
@@ -1,0 +1,80 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <ProjectGuid>{A35E38C4-8189-43C0-BA09-8D05B8EF0791}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <RootNamespace>Microsoft.CodeAnalysis.Sarif.ValidationTests</RootNamespace>
+    <AssemblyName>Sarif.ValidationTests</AssemblyName>
+  </PropertyGroup>
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\TestKey.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.props))\build.props" />
+  <ItemGroup>
+    <Reference Include="Microsoft.Json.Schema, Version=0.38.0.0, Culture=neutral, PublicKeyToken=7aebd41ac4801eb9, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Json.Schema.0.38.0\lib\net451\Microsoft.Json.Schema.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\packages\Newtonsoft.Json.6.0.8\lib\net45\Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Sarif, Version=1.5.24.0, Culture=neutral, PublicKeyToken=06c70c7db0c6cd6b, processorArchitecture=MSIL">
+      <HintPath>..\packages\Sarif.Sdk.1.5.24-beta\lib\net45\Sarif.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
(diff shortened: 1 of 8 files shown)


src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -0,0 +1,80 @@
...
+    <Compile Include="SarifValidatorTests.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Target Name="CopyTestData" AfterTargets="Build">
```
lgolding: A little MSBuild magic, needed because the test data still lives in the FunctionalTests project. Building the FunctionalTests project copies the files to the FunctionalTests output directory, but we need them in the ValidationTests output directory as well.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -29,7 +30,7 @@ public SarifValidatorTests()
         {
             _jsonSchemaFilePath = Path.Combine(Environment.CurrentDirectory, JsonSchemaFile);
             string schemaText = File.ReadAllText(JsonSchemaFile);
-            _schema = SchemaReader.ReadSchema(schemaText);
+            _schema = SchemaReader.ReadSchema(schemaText, JsonSchemaFile);
```
lgolding: The new version of the `SchemaReader` takes then name of the schema file so it can include it in the `Result` objects it produces when there's an error reading the schema.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -39,7 +40,7 @@ public void Direct_Producer_validation(string inputFile)
             string instanceText = File.ReadAllText(inputFile);
             var validator = new Validator(_schema);
 
-            string[] errors = validator.Validate(instanceText);
+            Result[] errors = validator.Validate(instanceText, inputFile);
```
lgolding: And here's the whole point; the validator returns `Result` objects. It also takes the name of the input file so it can include those in the `Result` objects.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -110,13 +111,13 @@ public static IEnumerable<object[]> ConverterTestCases
...
 
-        private string FailureReason(string[] errors)
+        private string FailureReason(Result[] errors)
         {
             var sb = new StringBuilder("file should be valid, but the following errors were found:\n");
 
             foreach (var error in errors)
             {
-                sb.AppendLine(error);
+                sb.AppendLine(error.FormatForVisualStudio(RuleFactory.GetRuleFromRuleId(error.RuleId)));
```
lgolding: `RuleFactory` comes from Json.Schema.

src/Sarif.FunctionalTests/Properties/AssemblyInfo.cs
```diff
@@ -1,8 +1,5 @@
-﻿// /********************************************************
-// *                                                       *
-// *   Copyright (C) Microsoft. All rights reserved.       *
-// *                                                       *
-// ********************************************************/
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
lgolding: I just noticed this when I happened to choose this file to copy into the new ValidationTests project. It's not like I was going around looking for it.

michaelcfanning:
+1. interested to know offline to what degree your validator completes inspection when multiple issues exist.

lgolding:
As we discussed, the answer is as follows:

* If there are any syntax errors in the schema file, the validator reports the first one and stops.
* Otherwise, if there are any "logical" errors in the schema file (for example, "type" has an invalid value), the validator will report all of them, and then stop.
* Otherwise, if there are any syntax errors in the schema file, the validator reports the first one and stops.
* Otherwise, if there are any validation errors in the instance file, the validator reports all of them.

---
In reply to: [222168218](https://github.com/Microsoft/sarif-sdk/pull/314#issuecomment-222168218) [](ancestors = 222168218)

PR #316, opened 2016-05-28
Allow annotatedCodeLocation.id to be a positive integer or stringized positive integer

lgolding:
We change the schema to allow `annotatedCodeLocation.id` to be either a string or an integer. We add constraints to the schema to enure that a string value matches the pattern of a positive integer, and that an integer value is positive.

By ordering the permitted types as `[ "integer", "string" ]`, we cause the code generator to choose `int` rather than `string` as the type of the generated property.

To enable either type to be deserialized into the same int-valued property,we introduce the `AnnotatedCodeLocationIdConverter` (along with its unit tests), and we use a code gen hint to add the necessary `JsonConverter` attribute to the property.

The converter enforces the same constraints as the schema, so it is not possible to deserialize a SARIF file that violates those constraints, even if you haven't validated it first.

The unit tests of the converter revealed a problem: even though the converter rejects an integer value of 0, the unit test was failing (that is, the unit test incorrectly accepted an integer value of 0). The reason is that 0 is the default value of the property, and by default, Json.NET won't even call the converter if the incoming value is the default.

To fix this, I needed to override Json.NET's default default-value-handling by adding a `JsonProperty` attribute to the `Id` property, and specifying `DefaultHandling.IgnoreAndPopulate`. This value means that when serializing, a default value will be ignored, but when deserializing, the property will be explicitly populated -- in particular, this means that the converter will be called -- instead of being left with the default value that .NET gives it.

But this meant that the `Id` property needed _two_ attributes (`JsonConverter` and `JsonProperty`), and until now, the code generator only allowed you to specify one `AttributeHint` per property. So I enhanced the code generator to allow multiple `AttributeHints` per property, and updated the package files to refer to the new version (0.42.0).

There is still one problem: if the `id` value is missing altogether, the `Id` property will have its default value of 0, which is invalid. SDK clients can (and should) ignore a 0 value, but we should do better. In particular, this property should be of type `int?`, but the code generator doesn't support nullable property types. We can talk on Monday about when/if to make this change.

Also:
* Fix a bug where I had neglected to add the new Sarif.ValidationTests project to the solution.
* Enhance BuildAndTest.cmd to accept a `/config` option.
* Replace all hard-coded parameter names in `throw` statements with `nameof`. Two of them were actually wrong before.

@michaelcfanning


BuildAndTest.cmd
```diff
@@ -5,15 +5,25 @@ SETLOCAL
...
 @REM %~dp0.nuget\NuGet.exe update -self
 
+set Platform=Any CPU
+set Configuration=Release
+
+:NextArg
+if "%1" == "" goto :EndArgs
+if "%1" == "/config" (
+    if not "%2" == "Debug" if not "%2" == "Release" echo error: /config must be either Debug or Release && goto :ExitFailed
+    set Configuration=%2&& shift && shift && goto :NextArg
```
lgolding: The lack of white space after "`%2`" is intentional. Any white space becomes part of the value of `Configuration`.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -36,7 +36,9 @@ public SarifNodeKind SarifNodeKind
         /// An identifier for the location, unique within the scope of the code flow within which it occurs.
         /// </summary>
         [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
-        public string Id { get; set; }
+        [Newtonsoft.Json.JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.AnnotatedCodeLocationIdConverter))]
+        [Newtonsoft.Json.JsonProperty("id", DefaultValueHandling = Newtonsoft.Json.DefaultValueHandling.IgnoreAndPopulate)]
```
lgolding: Here are the two attributes needed to properly handle the `Id` property. We specify the converter, and we ensure that the converter is called even when the incoming value is the default (0). This is what required us to upgrade to a new JSchema version.

src/Sarif/CodeGenHints.json
```diff
@@ -9,6 +9,29 @@
...
+        "typeName": "Newtonsoft.Json.JsonProperty",
+        "arguments": [
+          "\"id\""
+        ],
+        "properties": {
+          "DefaultValueHandling": "Newtonsoft.Json.DefaultValueHandling.IgnoreAndPopulate"
+        }
+      }
+    }
+  ],
```
lgolding: Look, ma,two `AttributeHint`s!

src/Sarif/CodeGenHints.json
```diff
@@ -19,26 +42,26 @@
...
+    },
+    {
+      "kind": "PropertyModifiersHint",
+      "arguments": {
+        "modifiers": [
+          "internal",
+          "override"
+        ]
+      }
+    }
```
lgolding: Not sure why Git thinks anything has changed here.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -37,7 +37,9 @@
 
         "id": {
           "description": "An identifier for the location, unique within the scope of the code flow within which it occurs.",
-          "type": "string"
+          "type": [ "integer", "string" ],
+          "pattern": "^[1-9][0-9]*$",
+          "minimum": 1
```
lgolding: The reason for the whole thing.

src/Sarif/ToDotNet/packages.config
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.Json.Schema.ToDotNet" version="0.36.0" targetFramework="net45" />
+  <package id="Microsoft.Json.Schema.ToDotNet" version="0.41.0" targetFramework="net451" />
```
lgolding: Somewhere between 0.36.0 and 0.39.0 I fixed the TFM in this package.

src/Sarif/CodeGenHints.json
lgolding: (You can't see it in CodeFlow, but Git shows a diff here even though comparing both sides shows them to be the same.)

---
In reply to: [64995043](https://github.com/Microsoft/sarif-sdk/pull/316#discussion_r64995043) [](ancestors = 64995043)

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
lgolding: Here are the two attributes needed to properly handle the Id property. We specify the converter, and we ensure that the converter is called even when the incoming value is the default (0). This is what required us to upgrade to a new JSchema version.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -1,19 +1,20 @@
 // Copyright (c) Microsoft.  All Rights Reserved.
 // Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-using Microsoft.CodeAnalysis.Sarif.Readers;
 using System;
 using System.CodeDom.Compiler;
 using System.Collections.Generic;
 using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Newtonsoft.Json;
```
lgolding: The newest JSchema code generator orders System usings first.

src/Sarif/ToDotNet/packages.config
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.Json.Schema.ToDotNet" version="0.36.0" targetFramework="net45" />
+  <package id="Microsoft.Json.Schema.ToDotNet" version="0.42.0" targetFramework="net451" />
```
lgolding: I fixed the TFM in this package.

src/Sarif/ToolFormatConverter.cs
```diff
@@ -152,7 +152,7 @@ private static void CreateConverterRecord<T>(IDictionary<ToolFormat, Lazy<IToolF
         /// <returns>The converted PREfast log file in SARIF format.</returns>
         public static string ConvertPREfastToStandardFormat(string inputFileName)
         {
-            if (inputFileName == null) { throw new ArgumentNullException("inputStream"); };
+            if (inputFileName == null) { throw new ArgumentNullException(nameof(inputFileName)); };
```
lgolding: This one was wrong.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -40,12 +40,12 @@ public ResultTextMarker(IServiceProvider serviceProvider, Region region, string
...
             if (serviceProvider == null)
             {
-                throw new ArgumentNullException("serviceProvider");
+                throw new ArgumentNullException(nameof(serviceProvider));
             }
 
             if (region == null)
             {
-                throw new ArgumentNullException("location");
+                throw new ArgumentNullException(nameof(region));
```
lgolding: This one was wrong.

src/Sarif/Autogenerated/AnnotatedCodeLocation.cs
```diff
@@ -36,7 +37,9 @@ public SarifNodeKind SarifNodeKind
         /// An identifier for the location, unique within the scope of the code flow within which it occurs.
         /// </summary>
         [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
-        public string Id { get; set; }
+        [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.AnnotatedCodeLocationIdConverter))]
+        [JsonProperty("id", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
```
lgolding: Here are the attributes that make it work, provided by multiple `AttributeHint`s.

src/Sarif/Readers/AnnotatedCodeLocationIdConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+            // We have to look at the type of the object returned by reader.Value, rather
+            // than looking at objectType, because Newtonsoft does something a bit odd
+            // when the input is a string that parses as a valid integer, such as "1". In
+            // that case, objectType is Int32, but the object returned by reader.Value is
+            // a string!
+            Type valueType = value.GetType();
+
+            if (valueType == typeof(string))
+            {
+                var stringValue = (string)value;
```
michaelcfanning: you make lots of type tests and casts. why not do something like this:

var stringValue = value as string;
if (stringValue != null)

src/Sarif/Readers/AnnotatedCodeLocationIdConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+                if (!int.TryParse((string)value, out intValue))
+                {
+                    throw new ArgumentOutOfRangeException(
```
michaelcfanning: If you intend to throw here, why not just call int.Parse and let the CLR exceptions (which are excellent) bubble out?

src/Sarif/Readers/AnnotatedCodeLocationIdConverter.cs
lgolding: Better. Fixed.

---
In reply to: [65252840](https://github.com/Microsoft/sarif-sdk/pull/316#discussion_r65252840) [](ancestors = 65252840)

michaelcfanning:
+1. no nullable types in the API, please. :) there's nothing wrong with a 0 value indicating nothing explicit has been set. We can define a const for this NOT_INITIALIZED. you're getting your mind around the all caps const definitions, aren't you? :) would be helpful if the property setter for Id threw on accepting an invalid value, might not be feasible while we depend on codegen classes.

src/Sarif/Readers/AnnotatedCodeLocationIdConverter.cs
lgolding: The CLR exception doesn't say what I want. It just tells you that the thing you tried to parse isn't an integer. My error message (which I use again below, when I range-test the resulting integer) is:

    '{0}' is not a valid value for the id property of an annotatedCodeLocation object. The value must be a positive integer or the string representation of a positive integer.

This is fully informative, and applies whether the input was a malformed string ("x1") or an out-of-range integer (-1).

---
In reply to: [65252923](https://github.com/Microsoft/sarif-sdk/pull/316#discussion_r65252923) [](ancestors = 65252923)

PR #324, opened 2016-06-03
Validated if "result" parameter was null.

kschecht:
@jericahuang @lgolding

### DIFF
--- a/src/Sarif/ConsoleLogger.cs
+++ b/src/Sarif/ConsoleLogger.cs
@@ -70,6 +70,11 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         public void Log(IRule rule, Result result)
         {
+            if (result == null)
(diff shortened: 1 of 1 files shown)


src/Sarif/ConsoleLogger.cs
```diff
@@ -70,6 +70,11 @@ public void LogMessage(bool verbose, string message)
 
         public void Log(IRule rule, Result result)
         {
+            if (result == null)
+            {
+                throw new ArgumentNullException("result");
```
lgolding: Very good. You can use a C# 6.0 feature here:

```
throw new ArgumentNullException(nameof(result));
```

You can look up `nameof` to see why that's a good thing.

jericahuang:
👍

lgolding:
:+1:

PR #325, opened 2016-06-03
CA1804 Deletion fix

Rolstenhouse:


kschecht:
:+1:

lgolding:
What is the Code Analysis warning that's being fixed here?

src/Sarif/Converters/ClangAnalyzerConverter.cs
```diff
@@ -43,7 +43,6 @@ public void Convert(Stream input, IResultLogWriter output)
 
                 using (XmlReader xmlReader = XmlReader.Create(input, settings))
                 {
-                    XmlNodeType nodeType = xmlReader.MoveToContent();
```
lgolding: I'm concerned about this one. I understand that the return value from `MoveToContent` -- the variable `nodeType` -- is not used. But presumably the call to `MoveToContent` has side effects (changing the internal state of the reader). Do the tests pass when you remove this call?

src/Sarif/Converters/ClangAnalyzerConverter.cs
```diff
@@ -355,7 +354,7 @@ private void ReadPlistDictionary(XmlReader xmlReader, IList<Result> results)
                                     throw new InvalidDataException("Expected key value before dictionary data.");
                                 }
 
-                                string value = xmlReader.ReadElementContentAsString();
```
lgolding: See, here's an example where you removed the variable but kept the call, which makes sense.

src/Sarif/Converters/FortifyPathElement.cs
```diff
@@ -66,7 +66,6 @@ public static FortifyPathElement Parse(XmlReader xmlReader, FortifyStrings strin
                 throw xmlReader.CreateException(SdkResources.FortifyNotValidPathElement);
             }
 
-            int pathElementDepth = xmlReader.Depth;
```
lgolding: And this one obviously has no side effects, so it's fine.

lgolding:
This looks good except for the very first change. See my comment on that one.

lgolding:
:+1:

Rolstenhouse:
Added next commit before closing pull request, please check it out

src/Sarif/ExtensionMethods.cs
```diff
@@ -153,15 +153,15 @@ public static void Populate(this Region region, NewLineIndex newLineIndex)
             }
         }
 
-        public static string GetMessageText(this Result result, IRule rule, bool concise = false)
```
lgolding: The purpose is to allow the _caller_ to pass in `consise = true;`. I'll come see you about this one.

src/Sarif/ExtensionMethods.cs
```diff
@@ -153,15 +153,19 @@ public static void Populate(this Region region, NewLineIndex newLineIndex)
             }
         }
 
+        public static string GetMessageText(this Result result, IRule rule)
+        {
+            return GetMessageText(result, rule, concise: false);
+        }
+
         public static string GetMessageText(this Result result, IRule rule, bool concise = false)
```
lgolding: You added the overload which is good, but now you have to remove the default parameter:

```
public static string GetMessageText(this Result result, IRule rule, bool concise)
{
    ...
```

lgolding:
:+1:

src/Sarif/ExtensionMethods.cs
```diff
@@ -185,7 +185,19 @@ public static string GetMessageText(this Result result, IRule rule, bool concise
             {
                 throw new ArgumentNullException(nameof(rule));
             }
+            return GetMessageText(result, rule, concise: false);
```
lgolding: A subtlety: When you have an overload that is just a _pass-through_ to an overload with more arguments, do all the argument checking in the overload with the most arguments. That way, you only have to write the argument-checking code once. So remove this check... unless... do you get another warning if you remove this check?

Also, formatting nitpick: our convention is to add a blank line after a closing brace. Likewise after line 200.

PR #332, opened 2016-06-04
Validate dependencies

lgolding:
Upgrade to JSchema 0.43.0, which validates the JSON Schema "dependencies" property. Add appropriate property dependencies to the schema to take advantage of it.

@michaelcfanning

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -194,7 +194,11 @@
           }
         }
       },
-      "required": [ "uri", "replacements" ]
+
+      "required": [ "uri", "replacements" ],
+      "dependencies": {
+        "uriBaseId": [ "uri" ]
+      }
     },
     "file": {
       "description": "A single file. In some cases, this file might be nested within another file.",
@@ -261,6 +265,10 @@
             }
           }
         }
+      },
+
+      "dependencies": {
+        "uriBaseId": [ "uri" ]
       }
     },
     "fix": {
@@ -553,7 +561,7 @@
           "type": "string",
           "format": "uri"
         },
-        
+
         "uriBaseId": {
           "description": "A string that identifies the conceptual base for the 'uri' property (if it is relative), e.g.,'$(SolutionDir)' or '%SRCROOT%'.",
           "type": "string"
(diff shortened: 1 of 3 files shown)


lgolding:
@michaelcfanning ping on this PR

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -1050,7 +1062,12 @@
         }
       },
 
-      "required": ["fullyQualifiedLogicalName"]
+      "required": [ "fullyQualifiedLogicalName" ],
+      "dependencies": {
+        "uriBaseId": [ "uri" ],
```
michaelcfanning: just playing devil's advocate but could the uriBaseId itself be sufficient? what does it look like if the very root of the uriBaseId is, for example, a scan target?

lgolding: Not sure what you mean by "very root", but... keeping in mind that the value `uriBaseId` is an arbitrary string upon the meaning of which the producer and consumer of the log file agree... are you asking what would happen if the producer and consumer agreed that the meaning of, say `"target"` was the analysis target itself, so that the consumer would resolve `uriBaseId` to `file:///C:/bin/foo.dll"?

lgolding: If that's what you're asking, I would say "not now". It may be a fine idea, but I propose that we are locked on both the format and the semantics of v1.0.0. We can consider this in a future version.

michaelcfanning:
Imagine scan target is the root of the enlistment. d:\
This is equivalent to TFS '$', yes? the root. That value is appropriate for uriBaseId and other things might be rooted off it, e.g., $\foo.cs. But again, if the root itself, which has no path, comprises the URI, then do we actually need to require the URI? For that case, can it be emitted.

lgolding:
So in your scenario, a `baseUriId` property has a value like `$srcroot`, which the producer and consumer  agree means "the root of the enlistment" (`c:\src`, or TFS `$`, or whatever). And the tool is actually instructed to scan `c:\src`, and one or more result locations have an URI that is also `c:\src`. And _if_ we required `uri` to be present in that situation (which we currently do), it would have a value of `.`. And you're asking, why not omit it?

It's easy for me to remove the dependency `"uriBase": [ "uri" ]`. And it's easy for me to add a sentence to the spec that says "If the absolute URI of location would be the same as the value implied by the `uriBaseId` property, then the `uri` property may be absent; otherwise, the `uri` property must be present." And it's easy for the consumer to do the extra logic to check for `uri` being absent.

So the trade is to pay this small increase in complexity of spec and implementation, for the small convenience of not having to write `"uri": "."` in (what I think you'll agree is) this corner case. I don't think it's worth it.

michaelcfanning:
ok, thanks for thinking this through.

PR #333, opened 2016-06-06
CA1816: GC.SuppressFinalize() to Dispose().

kschecht:
@jericahuang @Rolstenhouse @lgolding @michaelcfanning

### DIFF
--- a/src/Sarif/HashUtilities.cs
+++ b/src/Sarif/HashUtilities.cs
@@ -34,7 +34,12 @@ namespace Microsoft.CodeAnalysis.Sarif
 
                         var sha256Cng = new SHA256Cng();
                         checksum = sha256Cng.ComputeHash(bufferedStream);
-                        sha256 = BitConverter.ToString(checksum).Replace("-", String.Empty);                            
+                        sha256 = BitConverter.ToString(checksum).Replace("-", String.Empty);
+
+                        // disposals
+                        md5Cng.Dispose();
+                        sha1Cng.Dispose();
+                        sha256Cng.Dispose();                        
                     }
                 }
             }
@@ -55,6 +60,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                         var sha = new SHA256Cng();
                         byte[] checksum = sha.ComputeHash(bufferedStream);
                         sha256Hash = BitConverter.ToString(checksum).Replace("-", String.Empty);
+
+                        sha.Dispose();
                     }
                 }
             }
@@ -76,6 +83,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                         var sha = new SHA1Cng();
                         byte[] checksum = sha.ComputeHash(bufferedStream);
                         sha1 = BitConverter.ToString(checksum).Replace("-", String.Empty);
+
+                        sha.Dispose();
                     }
                 }
             }
@@ -97,6 +106,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                         var sha = new MD5Cng();
                         byte[] checksum = sha.ComputeHash(bufferedStream);
                         md5 = BitConverter.ToString(checksum).Replace("-", String.Empty);
+
+                        sha.Dispose();
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -223,6 +223,9 @@ public void Dispose()
                 _issueLogJsonWriter.Dispose();
             }
             if (_textWriter != null) { _textWriter.Dispose(); }
+
+            GC.SuppressFinalize(_issueLogJsonWriter);
+            GC.SuppressFinalize(_textWriter);
```
lgolding: This isn't the correct fix. I'm curious, does the warning go away when you add this code?

To understand why this is wrong, look at the web page for CA1816: [https://msdn.microsoft.com/en-us/library/ms182269.aspx](https://msdn.microsoft.com/en-us/library/ms182269.aspx). The idea is that the Dispose method for an object `X` frees the resources held be `X`. That means it's no longer necessary for the garbage collector to free `X`'s resources by calling `X`'s "finalizer" method. So you _tell_ the garbage collector not to bother "finalizing" `X` by calling `GC.SuppressFinalize` _on X itself_. So the correct fix is:

```
GC.SuppressFinalize(this);
```

You don't have to call `GC.SuppressFinalize` on `X`'s resources (in this case, `_textWriter` and `_issueLogJsonWriter`) because you just called _their_ `Dispose` methods -- and they are responsible for freeing their own resources.

lgolding:
:clock1:

michaelcfanning:
FxCop rule logic probably is doing some analysis that minimizes false positives (likely it only looks for some presence of GC.Collect in the method body, without verifying that it operates against the this pointer).

lgolding:
@kschecht, @michaelcfanning isn't disagreeing with my comment, he's just explaining why the warning might have gone away -- if in fact it did. @michaelcfanning, @kschecht had not actually verified that his change suppressed the warning.

kschecht:
For the commit "CA2000: Disposals", there were 8 warnings that told me to dispose of elements before they went out of scope. However, the last two elements that I am told to dispose of follow the form "new FileStream(outputFilePath, FileMode.Create, ....)". Is it possible to remedy this warning? If so, how?

src/Sarif/HashUtilities.cs
```diff
@@ -34,7 +34,12 @@ public static void ComputeHashes(string fileName, out string md5, out string sha
 
                         var sha256Cng = new SHA256Cng();
                         checksum = sha256Cng.ComputeHash(bufferedStream);
-                        sha256 = BitConverter.ToString(checksum).Replace("-", String.Empty);                            
+                        sha256 = BitConverter.ToString(checksum).Replace("-", String.Empty);
+
+                        // disposals
+                        md5Cng.Dispose();
+                        sha1Cng.Dispose();
+                        sha256Cng.Dispose();
```
lgolding: As we've discussed: Use `using` statements to ensure that `Dispose` is called even if the intervening code throws an exception.

lgolding:
Add the `using` directives as we discussed, then merge this after @michaelcfanning have signed off on it. We'll figure out the `new FileStream` issue in a separate PR.

src/Sarif/HashUtilities.cs
```diff
@@ -18,23 +18,31 @@ public static void ComputeHashes(string fileName, out string md5, out string sha
...
+                                stream.Seek(0, SeekOrigin.Begin);
+                                bufferedStream.Seek(0, SeekOrigin.Begin);
+
+                                using (var sha256Cng = new SHA256Cng())
+                                {
+                                    checksum = sha256Cng.ComputeHash(bufferedStream);
+                                    sha256 = BitConverter.ToString(checksum).Replace("-", String.Empty);
+                                }
+                            }
+                        }
```
lgolding: You don't need multiple levels of nesting, because you only need each "Cng" object long enough to compute the checksum:

```
byte[] checksum;

using (var md5Cng = new MD5Cng())
{
   checksum = md5Cng.ComputeHash(bufferedStream);
   md5 = BitConverter.ToString(checksum).Replace("-", String.Empty);
}

stream.Seek(0, SeekOrigin.Begin);
bufferedStream.Seek(0, SeekOrigin.Begin);

using (var sha1Cng = new SHA1Chng())
{
    checksum = sha1Cng.ComputeHash(bufferedStream);
    sha1 = BitConverter.ToString(checksum).Replace("-", String.Empty);
}

// etc.
```

michaelcfanning:
+1

PR #334, opened 2016-06-06
CA1021 avoid out parameters

Rolstenhouse:
This introduced several spelling based warning which should be suppressed.

I'm expecting this to have plenty of fixes due.

### DIFF
--- a/src/Sarif/Core/FileData.cs
+++ b/src/Sarif/Core/FileData.cs
@@ -28,24 +28,24 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (computeHashes && uri.IsAbsoluteUri && uri.IsFile)
             {
-                string md5, sha1, sha256;
+                //string md5, sha1, sha256;
 
-                HashUtilities.ComputeHashes(uri.LocalPath, out md5, out sha1, out sha256);
+                HashData hashes = HashUtilities.ComputeHashes(uri.LocalPath);
                 fileData.Hashes = new List<Hash>
                         {
                             new Hash()
                             {
-                                Value = md5,
+                                Value = hashes.GetMD5,
                                 Algorithm = AlgorithmKind.MD5,
                             },
                             new Hash()
                             {
-                                Value = sha1,
+                                Value = hashes.GetSha1,
                                 Algorithm = AlgorithmKind.Sha1,
                             },
                             new Hash()
                             {
-                                Value = sha256,
+                                Value = hashes.GetSha256,
                                 Algorithm = AlgorithmKind.Sha256,
                             },
                         };
--- a/src/Sarif/Core/FileData.cs
+++ b/src/Sarif/Core/FileData.cs
@@ -28,8 +28,6 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (computeHashes && uri.IsAbsoluteUri && uri.IsFile)
             {
-                //string md5, sha1, sha256;
-
                 HashData hashes = HashUtilities.ComputeHashes(uri.LocalPath);
                 fileData.Hashes = new List<Hash>
                         {
--- a/src/Sarif/Core/FileData.cs
+++ b/src/Sarif/Core/FileData.cs
@@ -33,17 +33,17 @@ namespace Microsoft.CodeAnalysis.Sarif
                         {
                             new Hash()
                             {
-                                Value = hashes.GetMD5,
+                                Value = hashes.MD5,
                                 Algorithm = AlgorithmKind.MD5,
                             },
                             new Hash()
                             {
-                                Value = hashes.GetSha1,
+                                Value = hashes.Sha1,
                                 Algorithm = AlgorithmKind.Sha1,
                             },
                             new Hash()
                             {
-                                Value = hashes.GetSha256,
+                                Value = hashes.Sha256,
                                 Algorithm = AlgorithmKind.Sha256,
                             },
                         };
--- a/src/Sarif/HashData.cs
+++ b/src/Sarif/HashData.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 namespace Microsoft.CodeAnalysis.Sarif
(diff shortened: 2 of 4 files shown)


src/Sarif/Core/FileData.cs
```diff
@@ -28,24 +28,24 @@ public static FileData Create(Uri uri, bool computeHashes, out string fileDataKe
 
             if (computeHashes && uri.IsAbsoluteUri && uri.IsFile)
             {
-                string md5, sha1, sha256;
+                //string md5, sha1, sha256;
```
lgolding: Remove this now-commented-out line (and the blank line below it).

src/Sarif/Core/FileData.cs
```diff
@@ -28,24 +28,24 @@ public static FileData Create(Uri uri, bool computeHashes, out string fileDataKe
...
 
-                HashUtilities.ComputeHashes(uri.LocalPath, out md5, out sha1, out sha256);
+                HashData hashes = HashUtilities.ComputeHashes(uri.LocalPath);
```
lgolding: Yes, I like this approach.

src/Sarif/Core/FileData.cs
```diff
@@ -28,24 +28,24 @@ public static FileData Create(Uri uri, bool computeHashes, out string fileDataKe
...
                 fileData.Hashes = new List<Hash>
                         {
                             new Hash()
                             {
-                                Value = md5,
+                                Value = hashes.GetMD5,
```
lgolding: Naming convention: Property names don't begin with "Get". Just `hashes.MD5`, `hashes.Sha1`, _etc._

src/Sarif/HashUtilities.cs
```diff
@@ -7,11 +7,35 @@
...
+            this.md5 = md5;
+            this.sha1 = sha1;
+            this.sha256 = sha256;
+        }
+
+        public string GetMD5 { get { return md5; } }
+        public string GetSha1 { get { return sha1; } }
+        public string GetSha256 { get { return sha256; } }
+
+    }
```
lgolding: Formatting convention: blank line after closing brace. But...

we usually put each class in a separate file.

src/Sarif/HashUtilities.cs
```diff
@@ -7,11 +7,35 @@
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
+    public class HashData
+    {
+        string md5;
```
lgolding: Conventions:

* We explicitly specify the accessibility of each member, even if it is the default.
* We prefix instance field names with "`_`".

So:

```
    private string _md5;
```

But, our

src/Sarif/HashUtilities.cs
```diff
@@ -7,11 +7,35 @@
...
+        string sha1;
+        string sha256;
+
+        public HashData()
+        {
+            this.md5 = String.Empty;
```
lgolding: Conventions:

* We don't use "`this.`" because the naming convention tells us it's a field.
* We use the "type alias" rather than the class name for the built-in types ("`string`" rather than "`String`").

So:


```
    _md5 = string.Empty;

File: src/Sarif/HashUtilities.cs [hunk 7]
```diff
@@ -7,11 +7,35 @@
...
+            this.sha1 = String.Empty;
+            this.sha256 = String.Empty;
+        }
```
lgolding: I don't think we need to supply a default constructor. We don't have a scenario where we'd need to construct a `HashData` without already having in hand the arguments to the other constructor.

lgolding:
:clock1:

src/Sarif/HashUtilities.cs
lgolding: Ahh, I can show you a C# shortcut. If you have a simple property that doesn't do any computation -- it just sets or returns a value -- then you can use what's called an "auto-property", and you can dispense with the backing field:

```
public class HashData
{
    public HashData(string md5, string sha1, string sha256)
    {
        MD5 = md5;
        Sha1 = sha1;
        Sha256 = sha256;
    }

    public string MD5 { get; }
    public string Sha1 { get; }
    public string Sha256 { get; }
}

Much shorter!
```

src/Sarif/Core/FileData.cs
Rolstenhouse: Good catch.

src/Sarif/HashUtilities.cs
Rolstenhouse: Yes, and much cleaner as well. Will change this and place the other class in a separate file.

lgolding: The only place this class is constructed is in `FileData` (the other file, where you changed the arguments to the method call). There is no place that default-constructs a `HashData`.

Also notice that the properties of `HashData` are read-only ("get"), so if you default-constructed one, its properties would always be empty. What would such an object represent? If you don't have hash data available, you can just return a null instead of a `HashData`, and that's actually what `FileData` does.

This is a specific example of a general principle, called "YAGNI", which means "You Ain't Gonna Need It". We try not to write a single line of code until it's needed. Today, I can't see a need for the default constructor, so we don't write one.

src/Sarif/HashData.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
```
lgolding: Every file in this project has a copyright/license notice at the top. Copy/paste it from another file.

src/Sarif/HashData.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
```
lgolding: Not all of these `using` directives are needed. The ones that aren't will show up gray in VS; you can remove them.

michaelcfanning: In VS, you can right-click in the using sections and 'sort and remove unused namespaces'. Also, on seeing a grayed out namespace, you can invoke the auto-fixer and resolve this condition across the entire solution. +1 on this change, btw, nice work.

Rolstenhouse: Will remove.

Rolstenhouse:
bump

lgolding:
:+1:

PR #335, opened 2016-06-06
Resolved CA 2208 and CA 2202

jericahuang:
@lgolding @michaelcfanning @kschecht @Rolstenhouse

### DIFF
--- a/src/Sarif/Converters/FxCopConverter.cs
+++ b/src/Sarif/Converters/FxCopConverter.cs
@@ -580,23 +580,35 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         {
             XmlSchemaSet schemaSet = new XmlSchemaSet();
             Assembly assembly = typeof(FxCopLogReader).Assembly;
+            Stream stream = null;
 
-            using (var stream = assembly.GetManifestResourceStream(FxCopLogReader.FxCopReportSchema))
-            using (var reader = XmlReader.Create(stream))
+            try
             {
-                XmlSchema schema = XmlSchema.Read(reader, new ValidationEventHandler(ReportError));
-                schemaSet.Add(schema);
-            }
-
-            using (var sparseReader = SparseReader.CreateFromStream(_dispatchTable, input, schemaSet))
-            {
-                if (sparseReader.LocalName.Equals(SchemaStrings.ElementFxCopReport))
+                stream = assembly.GetManifestResourceStream(FxCopLogReader.FxCopReportSchema);
+                using (var reader = XmlReader.Create(stream))
                 {
-                    ReadFxCopReport(sparseReader, context);
+                    stream = null;
+                    XmlSchema schema = XmlSchema.Read(reader, new ValidationEventHandler(ReportError));
+                    schemaSet.Add(schema);
                 }
-                else
+
+                using (var sparseReader = SparseReader.CreateFromStream(_dispatchTable, input, schemaSet))
                 {
-                    throw new XmlException(String.Format(CultureInfo.InvariantCulture, "Invalid root element in FxCop log file: {0}", sparseReader.LocalName));
+                    if (sparseReader.LocalName.Equals(SchemaStrings.ElementFxCopReport))
+                    {
+                        ReadFxCopReport(sparseReader, context);
+                    }
+                    else
+                    {
+                        throw new XmlException(String.Format(CultureInfo.InvariantCulture, "Invalid root element in FxCop log file: {0}", sparseReader.LocalName));
+                    }
+                }
+            }
+            finally
+            {
+                if(stream != null)
+                {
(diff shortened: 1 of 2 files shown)


src/Sarif/PerLanguageOption.cs
```diff
@@ -48,7 +48,7 @@ public PerLanguageOption(string feature, string name, Func<T> defaultValue, stri
 
             if (string.IsNullOrWhiteSpace(name))
             {
-                throw new ArgumentException(nameof(name));
+                throw new ArgumentNullException(nameof(name));
```
lgolding: This is one of two possible approaches. I'll let @michaelcfanning decide which he prefers.

This FxCop warning alerts you to a common mistake that people make when constructing an `ArgumentException`. On the web page [https://msdn.microsoft.com/en-us/library/ms182347.aspx](https://msdn.microsoft.com/en-us/library/ms182347.aspx), the Rule Description section provides the details. The problem is that in the `ArgumentException` constructor overloads, the argument is a message string, not the name of the parameter.

So you have two choices:

1. Provide a message string, for example:

```
throw new ArgumentException("You must provide a non-empty string", nameof(name));
```

(except that you would use a resource string instead of a literal string for the `message` argument).

2. Use a more specific exception type, as you have done.

The only problem with #2 -- if it is a problem -- is that if `name` is a string consisting of only white space, say `"   "`, then you will get an `ArgumentNullException` even though the argument is not technically null.

Full disclosure, I usually write the code like this:

```
if (name == null)
{
    throw new ArgumentNullException(nameof(name));
}

if (name.IsNullOrWhiteSpace(name))
{
    throw new ArgumentException(Resources.ErrorStringMustNotBeEmpty, nameof(name));
}
```

... but I don't get too worked up if other people don't do that.

jericahuang:
@lgolding @michaelcfanning Can you also review my changes to FxCopConverter.cs in my latest commit? It was to resolve the CA 2202 errors (Do not dispose objects multiple times).

lgolding:
@jericahuang Sure thing. A suggestion: try to keep unrelated changes in separate PRs. You can do this by making each change in a branch. That allows each set of related changes to be merged in as soon as they are approved, without having to wait for approval of the unrelated changes.

src/Sarif/Converters/FxCopConverter.cs
```diff
@@ -580,23 +580,35 @@ public void Read(Context context, Stream input)
...
-            using (var reader = XmlReader.Create(stream))
+            try
             {
-                XmlSchema schema = XmlSchema.Read(reader, new ValidationEventHandler(ReportError));
-                schemaSet.Add(schema);
-            }
+                stream = assembly.GetManifestResourceStream(FxCopLogReader.FxCopReportSchema);
+                using (var reader = XmlReader.Create(stream))
+                {
+                    stream = null;
```
lgolding: Yes, this is exactly the fix that the help page for this rule (https://msdn.microsoft.com/en-us/library/ms182334.aspx) suggests.

lgolding:
:+1: from me, but please hold for @michaelcfanning to look at this one.

To continue to make progress, you'll need to make a branch from the commit _before_ you started this work. I'll show you how.

lgolding:
@michaelcfanning According to https://msdn.microsoft.com/en-us/library/ms182334.aspx:

> If the `IDisposable` resource of the nested inner using statement contains the resource of the outer using statement, the `Dispose` method of the nested resource releases the contained resource.

That suggests that this is the reason the rule is firing: because it sees the stream (the "outer resource") being passed to the reader (the "inner resource").

But according to https://msdn.microsoft.com/en-us/library/system.xml.xmlreadersettings.closeinput(v=vs.110).aspx, the default value of `XmlReaderSettings.CloseInput` is `false`. This means that if we set `stream = null`, then the stream won't be disposed at all -- because we won't do it, and the `XmlReader` won't do it either.

So I don't think this fix is right (or rather, it's not complete). To suppress this warning and still ensure the stream is disposed, we could add the following to @jericahuang's code:

```
var settings = new XmlReaderSettings { CloseInput = true; }
using (var reader = XmlReader.Create(stream, settings);
{
    settings = null;
    ....
```

src/Sarif/PerLanguageOption.cs
michaelcfanning: I'm fine with the proposed change and would actually prefer not to over-elaborate on exception handling in this instance (given that we do not enforce a similar bar elsewhere in the code).

michaelcfanning: +1

jericahuang:
@lgolding @michaelcfanning 

Following up from our discussion on CA2202, this instance involves a generic Stream (not an InputStream).

jericahuang:
Reverted changes for 2208
Ready to merge for 2202

PR #337, opened 2016-06-06
CA1026 Removing default arguments

Rolstenhouse:
Please ignore CA1021 commits. I did not file the pull request properly.

View line comments. Two warnings could not be initially resolved.

### DIFF
--- a/src/Sarif/PerLanguageOption.cs
+++ b/src/Sarif/PerLanguageOption.cs
@@ -39,7 +39,7 @@ namespace Microsoft.CodeAnalysis.Sarif
         /// </summary>
         public Func<T> DefaultValue { get; }
 
-        public PerLanguageOption(string feature, string name, Func<T> defaultValue, string description = null)
+        public PerLanguageOption(string feature, string name, Func<T> defaultValue)
         {
             if (string.IsNullOrWhiteSpace(feature))
             {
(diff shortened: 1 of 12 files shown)


src/Sarif/PerLanguageOption.cs
```diff
@@ -51,10 +51,28 @@ public PerLanguageOption(string feature, string name, Func<T> defaultValue, stri
...
-            this.Feature = feature;
-            this.Name = name;
-            this.DefaultValue = defaultValue;
-            this.Description = description;
+            Feature = feature;
+            Name = name;
+            DefaultValue = defaultValue;
+            Description = null;
+        }
+
```
Rolstenhouse: Possibility of creating a default class, but that overrides the getter and setter methods. Will research better way to do this.

src/Sarif/PropertyBag.cs
```diff
@@ -17,7 +17,7 @@ public class PropertyBagDictionary : TypedPropertyBagDictionary<object>
         internal const string DEFAULT_POLICY_NAME = "default";
 
         public PropertyBagDictionary() : base() { }
-
+        //CA1026 not fixed
```
Rolstenhouse: Unsure of how to handle this one.

PR #338, opened 2016-06-06
CA1026 Redo Pull Request files

Rolstenhouse:
Redone PR

### DIFF
--- a/src/Sarif/PerLanguageOption.cs
+++ b/src/Sarif/PerLanguageOption.cs
@@ -39,22 +39,9 @@ namespace Microsoft.CodeAnalysis.Sarif
         /// </summary>
         public Func<T> DefaultValue { get; }
 
-        public PerLanguageOption(string feature, string name, Func<T> defaultValue)
+        public PerLanguageOption(string feature, string name, Func<T> defaultValue) :
+            this(feature, name, defaultValue, description: null)
         {
-            if (string.IsNullOrWhiteSpace(feature))
-            {
-                throw new ArgumentNullException(nameof(feature));
-            }
-
-            if (string.IsNullOrWhiteSpace(name))
-            {
-                throw new ArgumentException(nameof(name));
-            }
-
-            Feature = feature;
-            Name = name;
-            DefaultValue = defaultValue;
-            Description = null;
         }
 
         public PerLanguageOption(string feature, string name, Func<T> defaultValue, string description)
--- a/src/Sarif/PerLanguageOption.cs
+++ b/src/Sarif/PerLanguageOption.cs
@@ -39,7 +39,7 @@ namespace Microsoft.CodeAnalysis.Sarif
         /// </summary>
         public Func<T> DefaultValue { get; }
 
-        public PerLanguageOption(string feature, string name, Func<T> defaultValue, string description = null)
+        public PerLanguageOption(string feature, string name, Func<T> defaultValue)
         {
             if (string.IsNullOrWhiteSpace(feature))
             {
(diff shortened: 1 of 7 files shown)


src/Sarif/PerLanguageOption.cs
```diff
@@ -51,10 +51,28 @@ public PerLanguageOption(string feature, string name, Func<T> defaultValue, stri
...
+
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                throw new ArgumentException(nameof(name));
+            }
+
+            Feature = feature;
+            Name = name;
+            DefaultValue = defaultValue;
+            Description = description;
```
Rolstenhouse: This bit is duplicated and is poor for code maintainability.

src/Sarif/PropertyBag.cs
```diff
@@ -17,7 +17,7 @@ public class PropertyBagDictionary : TypedPropertyBagDictionary<object>
         internal const string DEFAULT_POLICY_NAME = "default";
 
         public PropertyBagDictionary() : base() { }
-
+        //CA1026 Not sure how to fix
```
Rolstenhouse: Setting constructor null default arguments. Unsure of how to fix

src/Sarif/PerLanguageOption.cs
lgolding: This is true. That is why you should follow the usual pattern for overloads: The overload with fewer arguments should just call the overload with more arguments, passing default values for the additional arguments.

Now, in this case, the function is a constructor, so you need to use the C# syntax that allows a constructor to call another constructor in the same class:

```
public PerLanguageOption(string feature, string name, Func<T> defaultValue) :
    this(feature, name, defaultValue, null)
{
}

public PerLanguageOption(string feature, string name, Func<T> defaultValue, string description)
{
   ...
}
```

Rolstenhouse: Thank you. Was unsure of how to overload a constructor in c#

src/Sarif/PropertyBag.cs
```diff
@@ -16,7 +16,12 @@ public class PropertyBagDictionary : TypedPropertyBagDictionary<object>
...
-        public PropertyBagDictionary() : base() { }
+        public PropertyBagDictionary() : this(null) { }
+
+        public PropertyBagDictionary(PropertyBagDictionary initializer) :
+            this(initializer, null)
+        {
+        }
 
         public PropertyBagDictionary(
             PropertyBagDictionary initializer = null,
```
lgolding: Remove the argument initializer ("`= null`"). Remember that that's why we added the overloads: to ensure that all languages could have the convenience of the abbreviated argument lists (not all languages understand default parameters).

lgolding: Oh right, one more thing about this. If you have an constructor that takes no arguments, and a constructor that takes two arguments, _both of which have defaults_, and you write:

```
X x = new X();
```

... then the call is ambiguous. The compiler didn't know if you meant to call the no-argument constructor, or to call the two-argument constructor with the default arguments. I don't think the code as currently written will build. Did you try building it?

---
In reply to: [66078330](https://github.com/Microsoft/sarif-sdk/pull/338#discussion_r66078330) [](ancestors = 66078330)

lgolding:
:clock1:

michaelcfanning:
just my opinion, I believe there is a place in the world for default parameters. also not a big fan generally of 'don't use this feature' rules. so let's talk about whether to continue to fix occurrences of this rule. +1 on this change.

src/Sarif/PropertyBag.cs
Rolstenhouse: Code built successfully.

Rolstenhouse:
Merge

PR #339, opened 2016-06-06
CA1703: Suppress incorrect spelling.

kschecht:
@jericahuang @Rolstenhouse @lgolding @michaelcfanning 
I created a Global Suppression file to suppress warnings resulting from what it perceives as a misspelling of the word "Cpp", pending your approval.


lgolding:
> ��[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1703:ResourceStringsShouldBeSpelledCorrectly", MessageId = "cppcheck", Scope = "resource", Target = "Microsoft.CodeAnalysis.Sarif.SdkResources.resources")]

This is very strange. VS seems to have created this file with UTF-16 encoding (two bytes per Unicode character).

To fix this:

1. Open the file in VS
2. Select File, Save As.
    Do not change the file name.
3. Click the arrow next to the Save button, and the select Save with encoding
4. Choose encoding "Unicode (UTF-8 without signature) - Codepage 65001"
5. Click OK.

Then push the change to the PR.

---
Refers to: src/Sarif/GlobalSuppressions.cs:1 in 6bf0c9f. [](commit_id = 6bf0c9fdb0b78d36e955e564cbd31739fc1fa69d, deletion_comment = False)

kschecht:
Clarification of CA1704:
CA1704 appears to check for a similar error to CA1703, in that both give warnings for spelling errors, which is why I added this to the same branch instead of making a new one for it. All of my further changes not related to spelling errors will be added to a new branch.

src/Sarif/LineInfo.cs
```diff
@@ -86,6 +86,7 @@ public override string ToString()
         /// <param name="lhs">The left hand side.</param>
         /// <param name="rhs">The right hand side.</param>
         /// <returns>The result of the operation.</returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "rhs")]
```
lgolding: There's actually a .NET Framework Design Guidelines section on this one:

> **DO** use _left_ and _right_ for binary operator overload parameter names if there is no meaning to the parameters.

So rename `lhs` to `left`, `rhs` to `right`, and remove the suppression.

src/Sarif/Readers/SarifContractResolver.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Readers
 {
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif")]
```
lgolding: This will occur often, so add an assembly-level suppression to GlobalSuppressions.cs to allow this name everywhere (and remove the individual suppressions). Not sure of the syntax. It might be:

```
[assembly: SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif")]
```

... so you'll want to research it.

src/Sarif/RuntimeConditions.cs
```diff
@@ -30,6 +30,7 @@ public enum RuntimeConditions
         ExceptionInSkimmerInitialize = 0x02,
         ExceptionRaisedInSkimmerCanAnalyze = 0x04,
         ExceptionInSkimmerAnalyze = 0x08,
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Logfile")]
```
lgolding: In this case, change the identifier name to `ExceptionCreatingLogFile` (log file, two words) and remove the suppression.

src/Sarif/GlobalSuppressions.cs
```diff
@@ -0,0 +1,17 @@
...
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif", Scope = "namespace", Target = "Microsoft.CodeAnalysis.Sarif.Converters")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif", Scope = "namespace", Target = "Microsoft.CodeAnalysis.Sarif.Driver")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif", Scope = "namespace", Target = "Microsoft.CodeAnalysis.Sarif.Readers")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif", Scope = "namespace", Target = "Microsoft.CodeAnalysis.Sarif.Writers")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sarif")]
+// This file is used by Code Analysis to maintain SuppressMessage
```
lgolding: Oh! I see you already have a global suppression for "Sarif" (the `SuppressMessage` attribute has no `Target` property, so it applies to everything). You should be able to remove the individual suppressions.

src/Sarif/Readers/SarifContractResolver.cs
lgolding: I just looked at the GlobalSuppressions.cs file and you already have the global suppression for "Sarif". Try removing one of the individual suppressions and see if the warning is still suppressed. If so, get rid of all the individual suppressions for "Sarif". Try the lower-case "sarif" suppressions separately and add global suppression if needed.

---
In reply to: [66075163](https://github.com/Microsoft/sarif-sdk/pull/339#discussion_r66075163) [](ancestors = 66075163)

lgolding:
:clock1:

michaelcfanning:
Can we please use a custom dictionary to resolve these rather than suppressions?
https://msdn.microsoft.com/en-us/library/bb514188.aspx

PR #341, opened 2016-06-07
CA1305 String formatting

Rolstenhouse:
Current Culture: Information displayed to user
Invariant Culture: Information written to disk.

- Unsure of which to use where. Or if it would be better to supply a new culture for some

### DIFF
--- a/src/Sarif/ConsoleLogger.cs
+++ b/src/Sarif/ConsoleLogger.cs
@@ -264,12 +264,12 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (!string.IsNullOrEmpty(notification.Id))
             {
-                sb.AppendFormat(CultureInfo.CurrentCulture, "$" +notification.Id+": ");
+                sb.AppendFormat(CultureInfo.CurrentCulture, notification.Id + ": ");
             }
 
             if (!string.IsNullOrEmpty(notification.RuleId))
             {
-                sb.AppendFormat(CultureInfo.CurrentCulture, "$" + notification.RuleId + ": ");
+                sb.AppendFormat(CultureInfo.CurrentCulture, notification.RuleId + ": ");
             }
 
             sb.Append(notification.Message);
--- a/src/Sarif/ConsoleLogger.cs
+++ b/src/Sarif/ConsoleLogger.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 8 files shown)


src/Sarif/ConsoleLogger.cs
```diff
@@ -263,12 +264,12 @@ private string FormatNotificationMessage(Notification notification)
...
             if (!string.IsNullOrEmpty(notification.Id))
             {
-                sb.Append($" {notification.Id}: ");
+                sb.AppendFormat(CultureInfo.CurrentCulture, "$" +notification.Id+": ");
             }
 
             if (!string.IsNullOrEmpty(notification.RuleId))
             {
-                sb.Append($"{notification.RuleId}: ");
+                sb.AppendFormat(CultureInfo.CurrentCulture, "$" + notification.RuleId + ": ");
```
Rolstenhouse: Changing the way the string was generated fixed the error

lgolding:
Very nice. You correctly figured out that:

* You use `InvariantCulture` when constructing a string that's intended to be machine-readable, such as the contents of a SARIF file.
* You use `CurrentCulture` when constructing a string that's intended to be human-readable, such as console output. But...
* You use `InvariantCulture` in exception messages because they're intended for the developer.

(By the way, that last point is debatable, since so often exception messages end up being presented to the end user.)

:+1:

src/Sarif/ConsoleLogger.cs
michaelcfanning: You need to get rid of the appended "$" here. It used to broadcast an interpolated string. Or you need to reintroduce the interpolated string literal as the argument to the call (if permitted, I assumed you made a change due to a compilation issue??)

https://msdn.microsoft.com/en-us/library/dn961160.aspx

lgolding: Sorry, I missed this during my review. Michael is right, the "`$`" must be removed. To see what is was there for, do a web search for "c# interpolated strings".

In any case, you don't need to use `AppendFormat` at all. Just call


```
sb.Append(notification.RuleId);
```

... which avoids the whole question of which culture to use.

Rolstenhouse:
Scheduled for merge and close

src/Sarif/ConsoleLogger.cs
```diff
@@ -263,12 +264,12 @@ private string FormatNotificationMessage(Notification notification)
 
             if (!string.IsNullOrEmpty(notification.Id))
             {
-                sb.Append($" {notification.Id}: ");
+                sb.AppendFormat(CultureInfo.CurrentCulture, notification.Id + ": ");
```
lgolding: I think I commented on a previous version of this: you don't need to "format" at all here. You can just write

```
sb.Append(notification.Id);
```
... and avoid the whole issue of "culture".

Rolstenhouse: Is the appended semicolon unnecessary?

lgolding: Oh, I'm sorry! I didn't even see it. I take it all back.

PR #343, opened 2016-06-07
Convert FxCop MessageLevel to SARIF ResultLevel equivalent, where appropriate.

michaelcfanning:
@lgolding @jericahuang @kschecht @Rolstenhouse

### DIFF
--- a/src/Sarif/Converters/FxCopConverter.cs
+++ b/src/Sarif/Converters/FxCopConverter.cs
@@ -131,11 +131,73 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             result.Locations = new List<Location> { location };
 
-            TryAddProperty(result, context.Level, "Level");
+            bool mapsDirectlyToSarifName;
+
+            result.Level = ConvertFxCopLevelToResultLevel(context.Level, out mapsDirectlyToSarifName);
+
+            if (!mapsDirectlyToSarifName)
+            {
+                // We will not recapitulate FxCop MessageLevel names (such as 
+                // "Error" and "Warning") as a property. For names that differ
+                // (such as "CriticalWarning" and "Information"), we will also 
+                // include the FxCop-specific values in the property bag.
+                TryAddProperty(result, context.Level, "Level");
+            }
+
             TryAddProperty(result, context.Category, "Category");
             TryAddProperty(result, context.FixCategory, "FixCategory");
 
             return result;
+        }
+
+        private static ResultLevel ConvertFxCopLevelToResultLevel(string fxcopLevel, out bool mapsDirectlyToSarifName)
+        {
+            mapsDirectlyToSarifName = true;
+
+            // Values below derived from definition of FxCop MessageLevel enum
+            // Microsoft.VisualStudio.CodeAnalysis.Extensibility.MessageLevel
+
+            switch (fxcopLevel)
+            {
+                case "Error":
+                {
+                    return ResultLevel.Error;
+                }
+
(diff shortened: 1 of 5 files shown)


src/Sarif/Converters/FxCopConverter.cs
```diff
@@ -131,13 +131,75 @@ internal Result CreateResult(FxCopLogReader.Context context)
...
+
+            result.Level = ConvertFxCopLevelToResultLevel(context.Level, out mapsDirectlyToSarifName);
+
+            if (!mapsDirectlyToSarifName)
+            {
+                // We will not recapitulate FxCop MessageLevel names (such as 
+                // "Error" and "Warning") as a property. For names that differ
+                // (such as "CriticalWarning" and "Information"), we will also 
+                // include the FxCop-specific values in the property bag.
+                TryAddProperty(result, context.Level, "Level");
```
lgolding: Maybe call the property "`FxCopLevel`" to avoid confusion about why there are two properties (one a first-class property, one in the property bag) both named level, and to make it clear that this is "FxCop's notion of 'level'".

lgolding:
:+1: modulo a suggestion to rename the "Level" property in the property bag.

src/Sarif/Converters/FxCopConverter.cs
michaelcfanning: No, sir. :) By specification, converters should not be synthesizing property names. 'Level' is the original attribute name in the FxCop project XML that contains this data and we need to preserve that.

lgolding: Let me think about this. The intent of the guidance for converter authors in the SARIF spec is that converters should not make guesses about the appropriate values for elements of the SARIF format that are not directly represented in the tool's native output format. Here, the converter knows exactly what the FxCop "Level" property means, and the question is how best to present it to the user in the SARIF file.

I think your concern is that if one converter decides that "FxCopLevel" is the best name for this, and another converter decides that "NativeLevel" is best, then a client could not easily consume SARIF files produced by those two different converters. That's true. On the other hand, there's nothing in the spec that requires a converter to populate a "Level" property in the property bag _at all_, so you still have the problem of converters behaving differently.

All in all, I think your argument comes down to "let's not introduce _gratuitous_ inconsistencies"; that is, "let's not do anything in our converter that another converter implementation might do differently". And I guess that preserving the tool's property name comes closest to that goal, even though (as I said) there's no requirement for a converter to emit it at all.

So bottom line, sure, let's leave it as "Level".

michaelcfanning: I like this simple rule, which is crystal clear and avoids lots of issues: if a converter perceives value in some tool-specific results data for which there is no equivalent SARIF slot, they can choose to populate the property bag with the information. If that data is named by the tool (via XML element or attribute, for example), the original tool name should be used, and the converter should preserve casing and values whatever is stored.

lgolding: Good, I'll add that guidance to the spec.

lgolding: Hah, actually that guidance already exists in Annex C of the spec:

> If the input data includes information for which there is no SARIF equivalent, converters may use it to populate the various property bags and tag lists defined by the SARIF format, or they may simply omit it from the output. When populating a property bag with such information, converters should use a property name that matches the name of that piece of information in the native tool format, even if that name does not conform to the camelCase convention used in the rest of this specification. This makes it easier to match these properties with the source data in the native tool format.

PR #346, opened 2016-06-07
Conference

Rolstenhouse:
Fixes from out meeting and CA 1702. Build successful, test successful

@jinujoseph

### DIFF
--- a/src/Sarif/Converters/AndroidStudioConverter.cs
+++ b/src/Sarif/Converters/AndroidStudioConverter.cs
@@ -57,11 +57,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 IgnoreComments = true,
                 IgnoreProcessingInstructions = true,
                 NameTable = _nameTable,
-                DtdProcessing = DtdProcessing.Ignore
+                DtdProcessing = DtdProcessing.Ignore,
+                XmlResolver = null
             };
 
             ISet<Result> results;
-            using (XmlReader xmlReader = XmlReader.Create(input, settings))
+            using (XmlReader xmlReader = XmlReader.Create(input, settings)) 
             {
                 results = ProcessAndroidStudioLog(xmlReader);
             }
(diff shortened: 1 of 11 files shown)


src/Sarif/Converters/AndroidStudioConverter.cs
```diff
@@ -57,11 +57,12 @@ public override void Convert(Stream input, IResultLogWriter output)
...
                 IgnoreProcessingInstructions = true,
                 NameTable = _nameTable,
-                DtdProcessing = DtdProcessing.Ignore
+                DtdProcessing = DtdProcessing.Ignore,
+                XmlResolver = null
             };
 
             ISet<Result> results;
-            using (XmlReader xmlReader = XmlReader.Create(input, settings))
+            using (XmlReader xmlReader = XmlReader.Create(input, settings))
```
lgolding: Remove extra blank space at end of line.

src/Sarif/IOptionsProvider.cs
```diff
@@ -7,6 +7,8 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public interface IOptionsProvider
     {
+        //GetOptions as a method may have significant processing and should not be an attribute
```
lgolding: "... should not be a property".

As it happens, the `SuppressMessageAttribute` class has a property named `Justification` that you can/should use instead of a comment here:

```
[System.Diagnostics.CodeAnalysis.SuppressMessage(
    "Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate",
    Justification = "This method is expensive and so should not be a property.")]
```

lgolding:
:+1: except for two minor comments.

src/Sarif/IOptionsProvider.cs
Rolstenhouse: Fixed both

lgolding:
+2

Rolstenhouse:
Merge and close

PR #347, opened 2016-06-07
CA1801: ConvertToSchemaUri uses this.

kschecht:


src/Sarif.Sdk.sln
```diff
@@ -31,6 +31,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "PREfastXmlSarifConverter",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sarif.ValidationTests", "Sarif.ValidationTests\Sarif.ValidationTests.csproj", "{A35E38C4-8189-43C0-BA09-8D05B8EF0791}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SarifTests", "SarifTests\SarifTests.csproj", "{335D24C1-6242-4328-93B5-25E05AC90C89}"
+EndProject
```
lgolding: There is already a project `Sarif.UnitTests`. Please add your new tests to that project.

src/SarifTests/SarifUtilitiesTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.CodeAnalysis.Sarif;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.Sarif.Tests
+{
+    [TestClass()]
```
lgolding: If an attribute has no arguments, you can omit the parentheses:

```
[TestClass]
```

src/SarifTests/SarifUtilitiesTests.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class SarifUtilitiesTests
+    {
+        private const string V1_0_0 = "1.0.0";
+        private const string V1_0_0_BETA_5 = "1.0.0-beta.5";
+
+        [TestMethod()]
+        public void ConvertToSchemaUriTest1()
```
lgolding: Try to provide informative test names, so that somebody looking at the list of tests in the Test Explorer window knows what they do. In this case the names "test 1" and "test 2" don't provide any information.

Perhaps `ConvertToSchemaUriV100`, `ConvertToSchemaUriV100Beta5`.

src/SarifTests/SarifUtilitiesTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.Sarif.Tests
+{
+    [TestClass()]
+    public class SarifUtilitiesTests
+    {
+        private const string V1_0_0 = "1.0.0";
+        private const string V1_0_0_BETA_5 = "1.0.0-beta.5";
```
lgolding: You don't need to provide these private constants. `ConvertToSchemaUri` takes as its argument a member of the `SarifVersion` enumeration. See below for how to do this.

src/SarifTests/SarifUtilitiesTests.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        [TestMethod()]
+        public void ConvertToSchemaUriTest1()
+        {
+            V1_0_0.ConvertToSarifVersion().ConvertToSchemaUri();
```
lgolding: A unit test has three parts:

1. Prepare to perform some action (optional: in this case, there's no preparation to do).
2. Perform the action.
3. Verify that the action had the correct result.

So you've done #2 but not #3.

You need to _assert_ that the action you've taken had the correct result. For now, let's use the assertion library that comes with Visual Studio (I actually prefer another library, but this will get us started quickly):

```
Uri uri = SarifVersion.OneZeroZero.ConvertToSchemaUri();
Assert.AreEqual(uri.ToString(), "http://json.schemastore.org/sarif-1.0.0")
```

src/SarifTests/SarifUtilitiesTests.cs
lgolding: I don't think you'll need these. See my proposed code pattern below.

src/SarifTests/SarifUtilitiesTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        [TestMethod()]
+        public void ConvertToSchemaUriTest2()
+        {
+            V1_0_0_BETA_5.ConvertToSarifVersion().ConvertToSchemaUri();
+        }
```
lgolding: Now of course, as the product code stands today, this test will fail (once it's rewritten as above). That means you can't check these tests in as is.

You have two choices:

* Mark this test as Ignored, check in, and then immediately submit a separate PR for the product code change.

```
[TestMethod, Ignore]
```

* (Preferable) Make the product code change in the same PR where you introduce the failing test -- that will make the test pass.

src/Sarif/SarifUtilities.cs
```diff
@@ -56,7 +56,7 @@ public static string ConvertToText(this SarifVersion sarifVersion)
 
         public static Uri ConvertToSchemaUri(this SarifVersion sarifVersion)
         {
-            return new Uri("http://json.schemastore.org/sarif-" + V1_0_0, UriKind.Absolute);
+            return new Uri("http://json.schemastore.org/sarif-" + sarifVersion.ConvertToText(), UriKind.Absolute);
```
lgolding: Yes, this is the right fix.

src/Sarif.Sdk.sln
lgolding: I see that you've now placed your new tests in the existing project `Sarif.UnitTests`. But don't forget to remove the now-unnecessary proejct `SarifTests`.

src/SarifTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
lgolding: Again, don't forget to remove the whole SarifTests project and directory.

src/Sarif.UnitTests/ConvertToSchemaUriTests.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
lgolding: The tests in this file now look good.

lgolding: Just a couple of formatting issues:

* All our source files need a copyright/license comment at the top. Copy one from an existing file and paste it in here.
* We do sort the `using` directives into alphabetical order, but we place the `System.*` usings first. (I don't know why ;-)) In VS, you can just right-click anywhere in the using directives, and then select menu item Organize usings, Sort usings.
* Remove the extra blank line before the first test method.

kschecht: In regards to your second point, when I click Sort Usings, the usings remain in the order of Microsoft.VisualStudio.TestTools.UnitTesting first and System second. Is Visual Studio correct to maintain this order or should I manually move System to the top of the list?

lgolding: This is a VS option. Go to Tools, Options. In the Options dialog, in the tree view on the left, select the node Text Editor, C#, Advanced. In the right-hand pane, in the "Organize Usings" section, check the box for "Place 'System' directives first when sorting usings".

src/SarifTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Runtime.CompilerServices;
```
lgolding: These files are still present in your enlistment. You must delete the SarifTests directory.

src/Sarif/Sarif.ruleset
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for Sarif" Description="Code analysis rules for Sarif.csproj." ToolsVersion="14.0">
+  <IncludeAll Action="Warning" />
+  <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
+    <Rule Id="CA2104" Action="None" />
+  </Rules>
+</RuleSet>
```
lgolding: I see that you created a rule set, but the rule set is not mentioned in the project file Sarif.csproj (that is, I don't see any changes to Sarif.csproj).

You _could_ make the required changes by hand (and I could show you how), but I'm curious why it didn't happen automatically when you created the rules file. Did you create the rules file by going to the Sarif project properties page in VS, selecting the Code Analysis tab, clicking Open, and then unchecking the check box for this rule?

+ @michaelcfanning

lgolding:
Looks good! :+1:

PR #348, opened 2016-06-08
Ca1057

Rolstenhouse:
CA1057, CA1303

@jinujoseph


src/Sarif/Errors.cs
```diff
@@ -411,7 +411,7 @@ public static RuntimeConditions LogUnhandledEngineException(IAnalysisContext con
             string exceptionMessage = exception?.Message;
             if (!string.IsNullOrEmpty(exceptionMessage))
             {
-                message += " ('" + exceptionMessage + "')";
+                message += Properties.Resources.OpenParenthesis + exceptionMessage + Properties.Resources.CloseParenthesis;
```
lgolding: There is already a .resx file src\Sarif\SdkResources.resx. Please place any new strings there.

But I wouldn't create resource strings for the individual punctuation marks. What the code is doing here is formatting an additional element into the message, in the case where there is an exception to report. So I'd define a resource string `Resources.NotificationWithExceptionMessage` with the value `"{0} ({1})"`. Then I'd write:

```
message = string.Format(CultureInfo.InvariantCulture, Resources.NotificationWithExceptionMessage, message, exceptionMessage);
```

src/Sarif/RuleUtilities.cs
```diff
@@ -103,7 +104,7 @@ public static string NormalizeFormatId(string ruleId, string formatId)
                 throw new ArgumentNullException(nameof(formatId));
             }
 
-            if (!string.IsNullOrEmpty(ruleId) && formatId.StartsWith(ruleId + "_"))
+            if (!string.IsNullOrEmpty(ruleId) && string.Compare(formatId.Substring(0, ruleId.Length+1), ruleId+"_", StringComparison.Ordinal)==0)
```
lgolding: You don't need to introduce `string.Compare` just for the sake of including the comparison type. `StartsWith` has an overload that takes a comparison type, so you can write just `formatId.StartsWith(ruleId + "_", StringComparison.Ordinal)`.

src/Sarif/Sarif.csproj
```diff
@@ -39,6 +39,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="GlobalSuppressions.cs" />
```
lgolding: This PR didn't introduce a GlobalSuppression file (I think Kevin did that yesterday, right?). Do you have a merge problem? In any case, you won't be introducing a new .resx file, so you can just revert the changes to this project file.

Rolstenhouse:
Ready for merging

PR #349, opened 2016-06-08
Hotfix. Build broken

Rolstenhouse:


Rolstenhouse:
Not sure how this was put in the master branch, but obviously a bug.

lgolding:
:+1: 

The most likely answer is that somebody merged without running BuildAndTest.cmd first :-)

PR #351, opened 2016-06-08
CA3053 Fixes

Rolstenhouse:
Unsure of why global suppressions file still shows up as added. I just fetched and rebased @10:00


src/Sarif/Converters/ClangAnalyzerConverter.cs
```diff
@@ -35,9 +35,12 @@ public void Convert(Stream input, IResultLogWriter output)
...
             {
-                XmlReaderSettings settings = new XmlReaderSettings();
-                settings.IgnoreWhitespace = true;
-                settings.DtdProcessing = DtdProcessing.Ignore;
+                XmlReaderSettings settings = new XmlReaderSettings
+                {
+                    IgnoreWhitespace = true,
+                    DtdProcessing = DtdProcessing.Ignore,
+                    XmlResolver = null
+                };
```
lgolding: Good use of the C# property initialization syntax!

lgolding:
OK, so IIRC our conversation, the idea here is that you correctly fixed the vulnerability, but CA3053 has a bug and doesn't recognize the fix, so you have to suppress anyway. Right? If so...

:+1:

Rolstenhouse:
Merge and close.

PR #356, opened 2016-06-08
CA2227: Suppressed readonly warnings.

kschecht:
This rule wants me to make certain things read-only by removing their setters. However, these things need to be set as different values later in the program, so they cannot be read-only. For now, I have suppressed this rule. However, I realize that the documentation recommends never suppressing this rule since it is a serious security concern, so I am open to any better solutions to this issue you recommend.

### DIFF
--- a/src/Sarif/IAnalysisContext.cs
+++ b/src/Sarif/IAnalysisContext.cs
@@ -20,6 +20,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         IRule Rule { get; set; }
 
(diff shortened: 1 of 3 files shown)


michaelcfanning:
Well, this problem _can_ be a serious security concern but in many or most cases, is not. :) I think the suppressions are ok. I suggest you file an issue to go clean this up in a future change, we mostly have deferred this because the ROI doesn't add-up. Your time is important and it's time to move on to new topics. :)

+1

PR #357, opened 2016-06-10
CA1720: Renamed parameter "obj" to "value".

kschecht:

### DIFF
--- a/src/Sarif.Driver/ReferenceEqualityComparer`1.cs
+++ b/src/Sarif.Driver/ReferenceEqualityComparer`1.cs
@@ -33,13 +33,15 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             return object.ReferenceEquals(x, y);
         }
 
+        #pragma warning disable CA1720 // Identifier contains type name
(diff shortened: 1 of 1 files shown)


lgolding:
I'm not wild about this change for two (related) reasons:
1) The argument to `GetHashCode` really is (any) object, and the name emphasizes that.
2) The [MSDN documentation](https://msdn.microsoft.com/en-us/library/system.collections.iequalitycomparer.gethashcode(v=vs.110).aspx) for this method names the parameter `obj`.

I would suppress.

michaelcfanning:
I disagree. I personally don't like to see suppressions in source code when a sensible change could have avoided it. There's nothing wrong with 'value' as an argument. 'obj' clearly doesn't hit bar for .NET naming conventions, as an abbreviation (and plenty of mistakes along these lines exist in .NET.

If you decide to retain the name, would suggest moving the suppression to the global suppression file (can Roslyn even do this, though??)

michaelcfanning:
@kschecht since you've already responded to larry's PR feedback, feel free to merge. my disagreement is not weighty. :)

PR #361, opened 2016-06-15
Appveyor

Rolstenhouse:
Global suppression information will be handled by a merge


BeforeBuild.cmd
```diff
@@ -0,0 +1,35 @@
...
+
+@ECHO off
+SETLOCAL
+
+if NOT exist "GeneratedKey.snk" (
+sn -k GeneratedKey.snk
+)
+
+if "%ERRORLEVEL%" NEQ "0" (
+echo COMMAND sn -k FAILED
```
lgolding: Avoid ALL-CAPS in messages. It sounds like you're shouting :-) Instead:

```
echo Command "sn -k" failed.
```

... here and throughout.
```

File: BeforeBuild.cmd [hunk 2]
```diff
@@ -0,0 +1,35 @@
+::Configuration for AppVeyor build platform
+
+@ECHO off
+SETLOCAL
+
+if NOT exist "GeneratedKey.snk" (
+sn -k GeneratedKey.snk
```
lgolding: This runs on the AppVeyor server, right? That means that `sn.exe` has to be on the execution path... ah, I see that you set up the Visual Studio dev environment in the `.yml` file. Good.

BuildAndTest.cmd
```diff
@@ -5,6 +5,12 @@ SETLOCAL
 @REM create a nuget package for the SARIF SDK) so must opt-in
 @REM %~dp0.nuget\NuGet.exe update -self
 
+call BeforeBuild.cmd
```
lgolding: Not that it matters alot, but let's put this down below all those `echo` commands.. that is, let's put this new code in the same place you put the old code. It just makes the GitHub diff easier to read.

BeforeBuild.cmd
```diff
@@ -0,0 +1,35 @@
...
+if "%ERRORLEVEL%" NEQ "0" (
+echo NUGET RESTORE FAILED
+goto ExitFailed
+)
+
+goto Exit
+
+:ExitFailed
+@echo.
+@echo SCRIPT FAILED
```
rtaket: You might change this to "Script %~n0 failed".  %~n0 resolves to the file name of the script, so it should show up in the log as "Script BeforeBuild.cmd failed".

lgolding:
:+1: subject to the two small suggestions I made. The `.yml` file is very interesting!

rtaket:
👍 
... I'm not sure about the suppression and project file changes, but the batch and appveyor files look fine.

lgolding:
Oh, right, what _is_ up with the new GlobalSuppressions.cs file? Is that something that shouldn't be in your branch?

src/Sarif/Sarif.csproj
```diff
@@ -39,6 +39,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="GlobalSuppressions.cs" />
```
nguerrera: I don't know what these have to do with the PR or why GitHub is claiming the GlobalSuppressions are a binary file. Can you remove this?

nguerrera:
Nice, Looks good apart from:

1) You have merge conflicts, which need to be resolved before this can be merged.
2) You should clean up the history (squash). There's not much value in the 43 commits as they're currently split.

Feel free to grab me for help with both of those.

src/Sarif/Sarif.csproj
lgolding: It's claiming it's a binary file because VS apparently generates this file as UTF-16 and GitHub can't deal. @jericahuang has fixed this problem once before by re-saving the file as UTF-8 (no signature), and I think @kschecht has done the same. This file shouldn't be changed/added in @Rolstenhouse's branch at all, though.

You might have to just save the new files (which you know now work) off to the side, abandon this branch, check out master, and add the new files to master. Your branch seems to be messed up somehow (as we see by the huge number of commits in this change.

lgolding:
As I commented before, I think he should just move the changed files off to the side, abandon the AppVeyor branch, and apply these changes (except the GlobalSuppressions and project file changes) to master.

nguerrera:
I agree with @lgolding re: fresh branch. That's probably the simplest way to get this done. :)

src/Sarif/Sarif.csproj
nguerrera: (Complete aside: UTF-16 files are terrible. Is there a bug filed on the VS integration to use UTF-8?)

Rolstenhouse:
Will make new branch and add files to it. Will keep this pr open until other branch is made.

lgolding:
@Rolstenhouse Two things:

1. First, make sure the `master` branch your fork is up to date with the upstream:

```
git checkout master
git fetch upstream
git merge upstream/master
BuildAndTest.cmd # To make sure all is well
git push
```

2. Then you can either do the work in a new branch from the latest `master`, or you can do the work directly in `master`. There's nothing wrong with that, since you're working in the `master` from your fork, not the `master` in the upstream repo.

Rolstenhouse:
Closed and got rid of old files

PR #362, opened 2016-06-15
appveyor configuration

Rolstenhouse:
Final.

Also, one of the tests is failing. This test was added after fetch was completed and fails as a result. 

Failed                Microsoft.CodeAnalysis.Sarif.Readers.AnnotatedCodeLocationKindConverterTests.AnnotatedCodeLocationKind_AllMembers

EDIT: Test failing because actual returned object contains a value "essential": false, where the test does not seem to set anything for expected. @michaelcfanning EDIT2: set test to pass

### DIFF
--- a/BeforeBuild.cmd
+++ b/BeforeBuild.cmd
@@ -1,0 +1,35 @@
+::Configuration for AppVeyor build platform
+
+@ECHO off
+SETLOCAL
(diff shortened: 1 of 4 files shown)


nguerrera:
:+1: Thanks.

PR #364, opened 2016-06-16
Email and basic artifact

Rolstenhouse:
Email message needs to be written better. Also need to learn to squash commit history

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,9 +29,6 @@ test:
 artifacts:
   - path: '\bld\bin\*.nupkg'
     name: Nuget Packages
-    
-  - path: 'https://ci.appveyor.com/api/buildjobs/%APPVEYOR_JOB_ID%/log'
-    name: Logged file
 
 notifications:
   - provider: Email
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,7 @@ test:
 notifications:
   - provider: Email
     to:
+      - rolsthoorn12@gmail.com
       - $env:EMAIL
       - $env:PERSONAL
     subject: 'Sarif-SDK Build {{status}}'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,6 @@ init:
 
 branches:
   only:
-    - appveyor
     - master
 
 before_build:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,17 +27,12 @@ test:
     - '**\Sarif.ValidationTests.dll'
     
 artifacts:
-  - path: '**\bld\bin\*.nupkg'
-    name: Nuget Packages
-    
   - path: '**\bld\bin\*'
-    name: Bin path
 
 notifications:
   - provider: Email
     to:
-      - rolsthoorn12@gmail.com
-      - vsarcplat@microsoft.com
+#      - vsarcplat@microsoft.com
     subject: 'Sarif-SDK Build {{status}}'
     message: 'In reference to {{commitID}}. View log output <a href="{{buildURL}}">here</a> '
     on_build_status_changed: true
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -25,6 +24,17 @@ test:
...
+    
+artifacts:
+  - path: 'bld\bin'
+
+notifications:
+  - provider: Email
+    to:
+      - vsarcplat@microsoft.com
+    subject: 'Sarif-SDK Build {{status}}'
+    message: 'In reference to {{commitID}}. View output <a href="{{buildURL}}">here</a> '
```
lgolding: Is there any way to add the commit message?

appveyor.yml
```diff
@@ -25,6 +24,17 @@ test:
...
+    on_build_status_changed: true
```
lgolding: Does this mean I get an email when the build starts, and when it finishes?

appveyor.yml
```diff
@@ -25,6 +24,17 @@ test:
     - '**\Sarif.Driver.UnitTests.dll'
     - '**\Sarif.FunctionalTests.dll'
     - '**\Sarif.ValidationTests.dll'
+    
+artifacts:
+  - path: 'bld\bin'
```
lgolding: So I get that this means that the contents of the `bld\bin` directory are uploaded to the cloud after the build (and presumably the build status page will give a link to that location).

But why do we need to do that? @nguerrera ?

appveyor.yml
Rolstenhouse: Only emails when it finishes

lgolding: ok.

appveyor.yml
Rolstenhouse: Yes. Email message still needs to be cleaned up. But it's by adding {{commitMessage}} into a relevant part of the body

lgolding: Yes, according to https://www.appveyor.com/docs/notifications, {{commitMessage}} would do it. Let's add it.

appveyor.yml
Rolstenhouse: I thought this would be valuable for down the line when nuget packages could be obtained from the artifact. This is more of a proof of concept than anything

appveyor.yml
rtaket: It would be nice to add some combination of {{commitAuthorEmail}}, {{commitAuthor}}, {{commitDate}}, {{projectName}}, {{buildVersion}} too.

---
In reply to: [67393094](https://github.com/Microsoft/sarif-sdk/pull/364#discussion_r67393094) [](ancestors = 67393094)

appveyor.yml
rtaket: I wanted that (to drop the bld folder.) As Rob stated, the next steps for the build are to update the assemblyinfo versions and get the nuget package to build and deploy.

---
In reply to: [67393231](https://github.com/Microsoft/sarif-sdk/pull/364#discussion_r67393231) [](ancestors = 67393231)

lgolding: Let's just skip it. We public the NuGet packages to nuget.org, and I've never needed to get the Nuget package file for a version that we _haven't_ published to NuGet.org.

rtaket:
> init:

Just FYI, it looks like the appveyor build for this PR was automatically kicked off and has completed successfully.

---
Refers to: appveyor.yml:1 in 6e61c7d. [](commit_id = 6e61c7ddeec81de3464c3c4526cdbd1d0173159f, deletion_comment = False)

appveyor.yml
nguerrera: I don't think we'll be able to use appveyor to produce official packages because we won't have the right access to sign them w/ Microsoft key.

lgolding:
:+1: but let's add some more information to the email, and as I discussed with Ryley, we don't really need to persist the artifact.

PR #367, opened 2016-06-21
Move SARIF converters to a separate project

rtaket:
Move the converter files from the Sarif project to a new project named Sarif.Converters.

Note: There shouldn't be any functional changes. I just moved the files from one project to a new project.

Besides the moving of files to a new assembly, the other breaking change is a namespace rename. In a few files, I changed the namespace from Microsoft.CodeAnalysis.Sarif to Microsoft.CodeAnalysis.Sarif.Converters to match the project name.


src/Sarif.Converters/ConverterResources.resx
```diff
@@ -0,0 +1,195 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
rtaket: These strings were moved from src/Sarif/SdkResources.resx.

src/Sarif.Converters/Extensions.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
rtaket: These members were moved from src/Sarif/Extensions.cs.

michaelcfanning:
Ok, so work to load arbitrary additional converter plugins TBD. 

+1

PR #374, opened 2016-06-24
Support aCL.importance and aCL.step

lgolding:
Phases one and two of support for `annotatedCodeLocation.importance` and `annotatedCodeLocation.step`. Phase three will be to install "corrective optics" for Update 3 PREfast files.


src/Sarif/CodeGenHints.json
```diff
@@ -34,6 +34,30 @@
...
+    {
+      "kind": "AttributeHint",
+      "arguments": {
+        "typeName": "JsonProperty",
+        "namespaceName": "Newtonsoft.Json",
+        "arguments": [
+          "\"essential\""
+        ],
+        "properties": {
+          "DefaultValueHandling": "DefaultValueHandling.Ignore"
```
lgolding: This says "don't write it out if it has the default value". The problem is that if it _doesn't_ have the default value, it _will_ be written out. So a result with essential==true will be written out the same way, which we don't want. It might be better to just put a `JsonIgnore` attribute here (never read it in, never write it out). That would present a challenge for our "corrective optics". I'm tempted to leave this as is until we get to the corrective optics phase.

src/PREfastXmlSarifConverter/SarifFormat.h
```diff
@@ -118,9 +118,9 @@ class SarifAnnotatedCodeLocation
         m_values[L"kind"] = kind;
     }
 
-    void SetEssential(bool essential)
+    void SetImportance(const std::wstring &importance)
     {
-        m_values[L"essential"] = json::Value(essential);
+		m_values[L"importance"] = json::Value(importance.c_str());
```
michaelcfanning: we don't explicitly call .c_str() in other helpers, probably best to remain consistent.

src/Sarif.FunctionalTests/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
```diff
@@ -43,10 +43,10 @@
                     "startLine":5
                   }
                 },
-                "id":"1",
+                "step":0,
```
michaelcfanning: do we want the converted code to explicitly include the 'id' from the PREfast log? what about the native PREfast 'importance' property (which consists of either 'important' or 'essential, i think). we can include this data for completeness. arguably, since we can directly convert to an existing SARIF concept with no loss of fidelity it can be omitted (but you need to know the rules of conversion). thoughts?

michaelcfanning:
>         "name": "StaticDriverVerifier"

hm, looks like some of my changes didn't get in. I fixed this. will take a look...

---
Refers to: src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/iodpcroutine_bug1.tt.sarif:7 in f801075. [](commit_id = f8010752b4402ce809e21ee425745852a1e4753f, deletion_comment = False)

src/Sarif.Converters/FortifyConverter.cs
```diff
@@ -147,8 +147,17 @@ public static Result ConvertFortifyIssueToSarifIssue(FortifyIssue fortify)
...
+                            new AnnotatedCodeLocation
+                            {
+                                PhysicalLocation = source,
+                                Importance = AnnotatedCodeLocationImportance.Essential
+                            },
+
+                            new AnnotatedCodeLocation
+                            {
+                                PhysicalLocation = primaryOrSink,
+                                Importance = AnnotatedCodeLocationImportance.Essential
```
michaelcfanning: Same here. If a converter actually doesn't distinguish between multiple importance levels, each step should be the default, imo.

src/PREfastXmlSarifConverter/SarifFormat.h
lgolding: Fixed.

src/Sarif.FunctionalTests/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
lgolding: * _Re_ `"id"`: I don't think so. PREfast's `KEYEVENT/ID` corresponds directly to SARIF's `annotatedCodeLocation.step`, and I think it's entirely within the spirit of the spec's guidelines for converters to translate the one to the other.

* _Re_: `"importance"`: I think the code is right as it stands. PREfast's `KEYEVENT/IMPORTANCE` property (whose value is either `Essential` or `Full`) corresponds directly to SARIF's `annotatedCodeLocation.importance (with the mappings `Essential => essential` and `Full => normal`).

src/Sarif.Converters/FortifyConverter.cs
lgolding: I agree, but here I was just directly translating the "`Essential = true`" on the left-hand side of the diff to "`Importance = AnnotatedCodeLocationImportance.Essential``" on the right-hand side.

I think you're saying that the "`Essential = true`" shouldn't have been there in the first place, and I agree with that.

I'm going to write myself a note and take care of this in a separate checkin.

lgolding:
The one unit test failure ("redact command line") is known and unrelated to this work. Merging.

PR #387, opened 2016-06-28
Suppressed support. First sentence extraction refinement.

michaelcfanning:
@lgolding @rtaket


src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -133,6 +133,11 @@ public override bool TryGetValue(int index, string columnName, out object conten
                     var error = _errors[index];
                     content = !string.IsNullOrEmpty(error.Message);
                 }
+                else
```
lgolding: I would explicitly state the column here ("`else if (this-is-the-suppressions-column`") for two reasons:
1. in case we add more columns
2. for understandability

michaelcfanning:
Failed   SarifLogger_RedactedCommandLine
Test is failing on master. Will chase down. @rtaket  +1 in person.

src/Sarif/ExtensionMethods.cs
```diff
@@ -278,18 +281,42 @@ public static string GetFirstSentence(string text)
...
+                    }
+
+                    // If we encounter a line-break, we return all leading text.
                     case '\n':
                     case '\r':
-                    case '.':
                     {
                         if (withinQuotes || withinParentheses) { continue; }
-                        return text.Substring(0, length).TrimEnd('\r', '\n');
+                        return text.Substring(0, length).TrimEnd('\r', '\n', ' ', '.') + ".";
```
lgolding: Consider extracting a method `EnsureEndingPeriod` since this calculation occurs twice.

michaelcfanning:
Got this and agreed. Ryley also provide some notes (we don't handle tabs, for example). I will propose a follow-on change. Pushed this one in due to pending changes coming from Ryley and Kevin.

PR #391, opened 2016-06-28
Fix Appveyor SarifLogger_RedactedCommandLine test failure

rtaket:
It looks like AppVeyor uses a different command line to invoke the tests. This caused the SarifLogger_RedactedCommandLine test to fail.

### DIFF
--- a/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
+++ b/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
@@ -22,6 +22,12 @@ namespace Microsoft.CodeAnalysis.Sarif
         {
             var sb = new StringBuilder();
 
+            // On a developer's machine, the script BuildAndTest.cmd runs the tests with a particular command line. 
+            // Under AppVeyor, the appveyor.yml file simply specifies the names of the test assemblies, and AppVeyor 
+            // constructs and executes its own, different command line. So, based on our knowledge of each of those 
+            // command lines, we select a different token to redact in each of those cases.
+            //
+            //
             // Sample test execution command-line from within VS. We will redact the 'TestExecution' role data
             //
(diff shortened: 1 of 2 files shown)


src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -34,24 +33,38 @@ public void SarifLogger_RedactedCommandLine()
             // "C:\Program Files (x86\\Microsoft Visual Studio 14.0\Common7\IDE\QTAgent32_40.exe\" 
             // /agentKey a144e450-ac06-46d0-8365-c21ea7872d23 /hostProcessId 8024 /hostIpcPortName 
             // eqt -60284c64-6bc1-3ecc-fb5f-a484bb1a2475"
+            //
```
lgolding: It took me a while to figure out why this was needed, and I think a comment is in order, something like this:

> On a developer's machine, the script BuildAndTest.cmd runs the tests with a particular command line. Under AppVeyor, the appveyor.yml file simply specifies the names of the test assemblies, and AppVeyor constructs and executes its own, different command line. So, based on our knowledge of each of those command lines, we select a different token to redact in each of those cases.

lgolding: Having said that, this is an awful lot of trouble to test a simple string replacement routine. If your goal is to test the string replacement itself, you could just declare `SarifLogger.Redact` as `internal`. If your goal is to convince yourself that the `SarifLogger` ctor actually _calls_ `Redact`, then yeah, I guess you have to do it this way.

lgolding: While you're in here, could you fix the nit in `SarifLogger.CreateRun` where it redacts `run.Invocation.CommandLine` twice (on line 56 and 59 of SarifLogger.cs)?

lgolding:
:+1: with a few thoughts.

lgolding:
:+1:-ier

PR #392, opened 2016-06-28
FortifyConverter: Don't emit aCL.importance

lgolding:
This converter always emitted Importance=Essential (previously, Essential=true), even though the native tool output doesn't have any equivalent.


michaelcfanning:
+1

PR #404, opened 2016-06-30
Rule metadata.

kschecht:
@rtaket @michaelcfanning


src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
```diff
@@ -104,6 +109,7 @@
     <Compile Include="WeakEventHandlerManager.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="Data\ruleLookup.json" />
```
rtaket: Eventually we should just include this file in the package. I'll make that change after you commit.

kschecht:
Commit 6001591: @rtaket @michaelcfanning Fixes issue #412 
Added 5 point margins. Did not add margins between Tool and Invocation because there is already a line of blank space between them. Can modify either of these properties.

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -27,6 +29,20 @@ public static bool TryGetRule(this Run run, string ruleId, string ruleKey, out I
...
+            // No rule in log file, created "fake" rule
+            else
+            {
+                Byte[] ruleLookupBytes = Resource1.ruleLookup;
+                string ruleLookupText = Encoding.UTF8.GetString(ruleLookupBytes);
+                JObject metadata = JObject.Parse(ruleLookupText);
+
+                if (metadata[ruleId] != null)
+                {
+                    string ruleName = metadata[ruleId]["heading"].Value<string>();
```
rtaket: metadata[ruleId]["heading"] could be null. We should add a null check before using it.

rtaket: metadata[ruleId]["url"] could also be null. Add a null check before using it.

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -27,6 +29,20 @@ public static bool TryGetRule(this Run run, string ruleId, string ruleKey, out I
                     rule = run.Rules[ruleId];
                 }
             }
+            // No rule in log file, created "fake" rule
+            else
+            {
+                Byte[] ruleLookupBytes = Resource1.ruleLookup;
+                string ruleLookupText = Encoding.UTF8.GetString(ruleLookupBytes);
+                JObject metadata = JObject.Parse(ruleLookupText);
```
rtaket: Parsing the file for every defect/rule might be expensive. We should probably only read it from disk once (i.e. cache the data after reading the file.)

rtaket:
👍

PR #413, opened 2016-07-07
Support invocation.responseFiles

lgolding:
Also:
* Address sarif-standard/sarif-specification#173 by making the schema descriptions of the `result.level` and `notification.level` properties consistent.
* Remove a couple of invalid `"default"` properties which specified that
  the default value for a string-valued property was an empty object.
* Added `"additionalProperties": true` to `invocation.environmentVariables`,
  just for explicitness.


michaelcfanning:
:shipit:

PR #420, opened 2016-07-09
Github URI Reformatting and Prefast example

Rolstenhouse:


appveyor.yml
```diff
@@ -17,7 +17,12 @@ configuration:
 
 build:
   project: src/Everything.sln
-  
+
+before_test:
+  - ps: cl.exe /c /analyze /analyze:quiet- /W4 /analyze:autolog:ext sarif src\Sarif.FunctionalTests\ConverterTestData\PREfast\src\AdvancedAliasing\AA_IndexAlias.cpp; ls
```
lgolding: This is not a PowerShell script, so according to the [appveyor.yml Reference](https://www.appveyor.com/docs/appveyor-yml), it should either start with "`cmd: `", or omit the prefix (because that's the default).

lgolding: Is the "`ls`" at the end just so you could see in the build log that the file had been created?

appveyor.yml
```diff
@@ -17,7 +17,12 @@ configuration:
...
+  - ps: Push-AppVeyorArtifact aa_indexalias.sarif
+  - ps: .\jsonparse.ps1 "aa_indexalias.sarif"  $env:APPVEYOR_REPO_PROVIDER $env:APPVEYOR_PROJECT_SLUG
```
lgolding: There are two issues with the name "jsonparse.ps1":

1. General: The name isn't informative.

    The name of anything -- a class, a method, a file -- should specifically say what the thing is or does. This script replaces file URIs with GitHub URIs, and the name doesn't convey that (in fact, the name suggests that the script is something it is not: a JSON parser).

2. Specific: The name doesn't follow PowerShell conventions.

    There are [naming conventions](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx) for Powershell cmdlets, and although AFAIK those guidelines don't apply to the names of Powershell _script files_, I like to follow them anyway.  Powershell commands names have the form "Verb-Noun", and there is a list of [approved verbs](https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx).

In this case, the best fit among the approved verbs appears to be "Convert", so how about: `Convert-FileUrisToGitHubUris.ps1`?

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
+$location=$args[0]
+$repo=$args[1]
+$name=$args[2]
+
+$location
+$repo
+$name
+$repo.Title
```
lgolding: Now that you've debugged this, let's remove these debug prints.

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+$JSON = (Get-Content $location) -join "`n" | ConvertFrom-Json
```
lgolding: Naming:

1. Since all your other variable names are camelCase (which agrees with PowerShell [coding conventions](https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/36) that I've seen in various places, let's use that convention for this variable as well ("`$json`") and also for "`$uri`".

2. But this particular blob of JSON represents a SARIF log, so let's name it either `$sarifLog` or `$log`. (I'd prefer `$sarifLog` because that's the name of the class in the SARIF object model: [SarifLog.cs](https://github.com/Microsoft/sarif-sdk/blob/master/src/Sarif/Autogenerated/SarifLog.cs).

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+$base = "https://github.com/"
+if($repo.Title -eq "github")
+{
+  $base = "https://github.com/"
+}
+#Renaming the key value of file
```
lgolding: The `if` statement doesn't do anything. `$base` still ends up with the same value.

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+$URI = $JSON.runs
```
lgolding: 1. This variable doesn't contain a URI, so it's misnamed. But...

2. You don't actually use this variable before you assign it again. Is this some leftover debugging code?

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+$JSON.runs.files.PSObject.Properties | foreach-object {
+  $URIReplacement = $_.Name
```
lgolding: This apparently works, but it is a dangerous coding pattern. The problem is that you're modifying a collection _while you are iterating over it_. In C#, [you'd get an exception](http://stackoverflow.com/questions/16497028/how-add-or-remove-object-while-iterating-collection-in-c-sharp).

Now for all I know this is perfectly safe in PowerShell, but it sets off alarm bells. I suggest you reseach this to find out if it's a safe pattern in PowerShell.

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
+$location=$args[0]
+$repo=$args[1]
+$name=$args[2]
```
lgolding: "`$name`" is a very generic name; it could be the name of anything. I see that when this script is called, this argument comes from [`$env:APPVEYOR_PROJECT_SLUG`](https://www.appveyor.com/docs/environment-variables), which is the "username/repo" string.

So I suggest you name this "`$projectSlug`".

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+$location
+$repo
+$name
+$repo.Title
+$JSON = (Get-Content $location) -join "`n" | ConvertFrom-Json
+
+$base = "https://github.com/"
```
lgolding: Again with my preference for specific, informative names, I'd call this "`$gitHubBaseUri`".

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+{
+  $base = "https://github.com/"
+}
+#Renaming the key value of file
+$URI = $JSON.runs
+$JSON.runs.files.PSObject.Properties | foreach-object {
+  $URIReplacement = $_.Name
+  $content = $_.Value
+  $JSON.runs.files.PSObject.Properties.Remove($_.Name)
+  $URIReplacement = $base + $URIReplacement.SubString($URIReplacement.IndexOf($name))
```
lgolding: This doesn't look right. At this point in the code just before this line, the variable `$URIReplacement` has the value `$_`, which is the property name, which is something like `"file:///C:/code/foo.cs"`. And as we've seen, `$name` is really the "project slug", so it has a value like `"Microsoft/sarif-sdk"`.

So unless I'm misreading the code, the string $URIReplacement doesn't contain the string `$name`, so  `$URIReplacement.IndexOf($name)` will return -1.

I might be missing something.

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+  $JSON.runs.files | add-member -Name $URIReplacement -Value $content -MemberType NoteProperty
+}
+
+$URI = $JSON.runs.files
```
lgolding: `runs` is an array, each element of which has a `files` property, so I'm not sure why this works. It's possible PowerShell's JSON handling automatically gives you the first array element if you don't specify an index, so that `$sarifLog.runs.files` is equivalent to `$sarifLog.runs[0].files`. You might look that up.

But in any event, the script should handle arrays with more than one element, and SARIF has many nested arrays, so you'll need nested loops. I'll give an example below.

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+for($i = 0; $i -lt $URI.Count; $i++){
+  $JSON.runs.files[$i].uri = $base + $URI[$i].uri.SubString($URI[$i].uri.IndexOf($name))
+}
+
+# Rewriting all urls for the results locations
+$URI = $JSON.runs.results.locations
+for($i = 0;$i -lt $URI.Count; $i++)
+{
+   $JSON.runs.results.locations[$i].resultFile.uri = $base + $URI[$i].resultFile.uri.SubString($URI[$i].resultFile.uri.IndexOf($name))
+}
```
lgolding: Here's an example. This needs to be something like this:

```
$runs = $sarifLog.runs
for ($i = 0; $i < $runs.Count; ++$i)
{
    $results = $runs[$i].results
    if ($results -ne null)
    {
        for ($j = 0; $j < $results.Count; ++$j)
        {
            $locations = $results[$j].locations
            if ($locations -ne null)
            {
                for ($k = 0; $k < $locations.Count; ++$k)
                {
                    $location = $locations[$k]
                    if ($location.analysisTarget -ne null)
                    {
                        $location.analysisTarget.uri = // replacement logic
                    }

                    $if ($location.resultFile -ne null)
                    {
                        $location.resultFile.uri = // replacement logic
                    }
                }
            }
        }
    }
}
```

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+$URI = $JSON.runs.results.codeFlows.locations.physicalLocation
+for($i = 0;$i -lt $URI.Count; $i++)
+{
+  $JSON.runs.results.codeFlows.locations.physicalLocation[$i].uri = $base + $URI[$i].uri.SubString($URI[$i].uri.IndexOf($name))
+}
```
lgolding: And finally, note that you have repeated the same replacement logic in several places. I suggest you pull that logic out into a function, say `Rebase-Uri`, and call it everywhere, for example:

```
$location.analysisTarget.uri = Rebase-Uri $location.analysisTarget.uri $githubUriBase #... plus whatever other args it needs
```

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
...
+  $JSON.runs.results.stacks.frames[$i].uri = $base + $URI[$i].uri.SubString($URI[$i].uri.IndexOf($name))
+}
+
+$URI = $JSON.runs.results.fixes.fileChanges
+for($i = 0; $i -lt $URI.Count; $i++)
+{
+  $JSON.runs.results.fixes.fileChanges[$i].uri = $base + $URI[$i].uri.SubString($URI[$i].uri.IndexOf($name))
+}
+
+#Only works for one value
```
lgolding: Once you've written the loops it will work always.

lgolding:
:clock1:

jsonparse.ps1
rtaket: I think our replacement uri is being formed incorrectly.

I believe the current script will create a replacement uri similar to this: [https://github.com/sarif-sdk-8e0d2/src/Sarif/Extensions.cs](https://github.com/sarif-sdk-8e0d2/src/Sarif/Extensions.cs)
I don't think this url will resolve on github. 

I suspect this is the url you intended. It's the url to the commit version of the file on github.
[https://github.com/Microsoft/sarif-sdk/blob/8e0d243d1eb61cc726d638b909144a84576dc6a4/src/Sarif/Extensions.cs](https://github.com/Microsoft/sarif-sdk/blob/8e0d243d1eb61cc726d638b909144a84576dc6a4/src/Sarif/Extensions.cs)
I think this is the format using the appveyor environment variables.
[https://github.com/$env:APPVEYOR_REPO_NAME/blob/$env:APPVEYOR_REPO_COMMIT/src/Sarif/Extensions.cs](https://github.com/$env:APPVEYOR_REPO_NAME/blob/$env:APPVEYOR_REPO_COMMIT/src/Sarif/Extensions.cs)

That being said, we should consider using the  link to the raw file instead of the github url. I can see arguments for either, but this link will make it easier for consumers and, I think, more closely matches the intent of the link. 
[https://raw.githubusercontent.com/Microsoft/sarif-sdk/8e0d243d1eb61cc726d638b909144a84576dc6a4/src/Sarif/Extensions.cs](https://raw.githubusercontent.com/Microsoft/sarif-sdk/8e0d243d1eb61cc726d638b909144a84576dc6a4/src/Sarif/Extensions.cs)
This is the format using the appveyor environment variables.  
[https://raw.githubusercontent.com/$env:APPVEYOR_REPO_NAME/$env:APPVEYOR_REPO_COMMIT/src/Sarif/Extensions.cs](https://raw.githubusercontent.com/$env:APPVEYOR_REPO_NAME/$env:APPVEYOR_REPO_COMMIT/src/Sarif/Extensions.cs)


---
In reply to: [70169279](https://github.com/Microsoft/sarif-sdk/pull/420#discussion_r70169279) [](ancestors = 70169279)

jsonparse.ps1
```diff
@@ -0,0 +1,83 @@
+$location=$args[0]
```
rtaket: I think we should put this file in a separate directory. i.e. not in the root.

appveyor.yml
Rolstenhouse: Fixed

appveyor.yml
Rolstenhouse: Removed for clarity

jsonparse.ps1
Rolstenhouse: Cleaned

jsonparse.ps1
Rolstenhouse: Let's meet up and discuss this one

jsonparse.ps1
Rolstenhouse: Done

jsonparse.ps1
Rolstenhouse: Will move all files into an appveyor directory

appveyor/Push-PrefastResultsToAppveyor.ps1
```diff
@@ -0,0 +1,8 @@
+$Dir = get-childitem .\ -Recurse -Filter "*.prefast.sarif"
+
+for ($i=0; $i -lt $Dir.Count; $i++)
+{
+  $fileLocation = $Dir[$i].FullName
+  Push-AppVeyorArtifact $fileLocation
+  .\appveyor\Convert-FileUrisToGitHubUris.ps1 $fileLocation
```
lgolding: Are these two lines in the right order? Shouldn't we convert the file before we ask AppVeyor to save it?

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
+$location=$args[0]
```
lgolding: I suggest the name `$logFilePath` since "location" means something else in SARIF. (In any case, variable names with "path" are a common convention when the variable contains a file system path.)

appveyor/Push-PrefastResultsToAppveyor.ps1
Rolstenhouse: So right now, I'm pushing the results before conversion and after conversion. There is another push artifact script at the end of the Convert-Files script

appveyor/Convert-FileUrisToGitHubUris.ps1
Rolstenhouse: Yep. That's definitely more descriptive.

Do you have any advice for naming variables in general? I usually name things the with the first name that comes to mind, which is usually pretty generic

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+##$projectSlug='sarif-sdk-931m5'
+#$repoName='Rolstenhouse/sarif-sdk'
+#$repoCommit='f82ebd892917bd6157efc04b832a57964ec1f2c6'
+
+$githubBaseUri = ""
+
+if($repoProvider -eq "gitHub")
+{
+  $githubBaseUri = "https://raw.githubusercontent.com/"
+}
```
lgolding: If the provider isn't `github`, shouldn't we just error out? Otherwise, `Rebase-Uri` returns a string like `Rolstenhouse/sarif-sdk/f82...2c6/..." with no host name.

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+  {
+      return $pathName
+  }
+  $directoryInfo = New-Object IO.DirectoryInfo($pathName)
+
+  if ($directoryInfo.Parent -ne $null)
+  {
+      $parentPath = Get-CaseSensitivePath($directoryInfo.Parent.FullName)
+      $childPath = $directoryInfo.Parent.GetFileSystemInfos($directoryInfo.Name)[0].Name
+      return(Join-Path $parentPath $childpath -resolv)
```
lgolding: "`-resolv`" => "`-resolve`"

appveyor/Convert-FileUrisToGitHubUris.ps1
Rolstenhouse: Yep. It should just quit the rewrite process.

appveyor/Convert-FileUrisToGitHubUris.ps1
Rolstenhouse: Changed

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+}
+
+$sarifLog = (Get-Content $location) -join "`n" | ConvertFrom-Json
+
+for ($i = 0; $i -lt $sarifLog.runs.Count; $i++)
+{
+  #files
+  for($j = 0; $j -lt $sarifLog.runs[$i].files.Count; $j++)
+  {
+    $sarifLog.runs[$i].files[$j].uri = Rebase-Uri($sarifLog.runs[$i].files[$j].uri)
```
lgolding: No parentheses around the arguments to a PowerShell function. This is actually kind of important, and the only reason you get away with it here is that there's only one argument. The syntax is:

```
Do-Stuff arg1 arg2 ...
```

or in this case

```
Rebase-Uri $sarifLog.runs[$i].files[$j].uri
```

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+    #UNTESTED
+    $props = JSON.runs[$i].files[$j].PSObject.Properties
```
lgolding: "`JSON`" => "`$sarifLog`"

lgolding: Anyway I'm not sure of this whole block of code for iterating over the named properties of an object. You'll need to research this a little.

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+    {
+      $URIReplacement = $props.Name
+      $content = $props.Value
+      $sarifLog.runs.files.PSObject.Properties.Remove($props.Name)
+      $sarifLog.runs.files | add-member -Name Rebase-Uri($URIReplacement) -Value $content -MemberType NoteProperty
+    }
+  }
+
+  #results
+  for($j = 0; $j -lt $sarifLog.runs[$i].results.Count; $j++)
```
lgolding: This would need to be in a loop over the `runs`, indexed by `$i`. As the code stands, you do this only for one run, using whatever value `$i` happens to have at the moment.

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+  {
+    for($k = 0; $k -lt $sarifLog.runs[$i].results[$j].locations.Count; $k++)
+    {
```
lgolding: Yes, the rest of this is the right idea.

appveyor/Convert-FileUrisToGitHubUris.ps1
Rolstenhouse: Will research and provide more feedback. It worked properly in the previously committed code, but hasn't been tested yet

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  for($j=0; $j -lt $sarifLog.runs[$i].toolNotifications.Count; $j++){
+    $sarifLog.runs[$i].toolNotifications[$j].physicalLocation.uri = Rebase-Uri($sarifLog.runs[$i].toolNotifications[$j].physicalLocation.uri);
+  }
+}
+
+$newExtension = ".github.sarif"
+$writeLocation = [System.IO.Path]::GetFileNameWithoutExtension($location)+$newExtension
+$sarifLog | ConvertTo-Json -depth 999 | Out-File $writeLocation
```
lgolding: [`ConvertTo-Json`](https://technet.microsoft.com/en-us/library/hh849922.aspx?f=255&MSPPError=-2147217396) does allow you to pass the `-InputObject` parameter through the pipeline, as you do here. As a matter of style -- and this is just my preference, I haven't seen this suggested anywhere -- I think that for the first command in a pipeline, the code is clearer if you pass the object on the command line (making the command name the first word on the line). So:

```
ConvertTo-Json -Depth 999 $sarifLog | Out-File $writeLocation
```

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    for($k = 0; $k -lt $sarifLog.runs[$i].results[$j].stacks.Count; $k++)
+    {
+      $sarifLog.runs[$i].results[$j].stacks[$k].frames.uri = Rebase-Uri($sarifLog.runs[$i].results[$j].stacks[$k].frames.uri)
+    }
+
+    for($k = 0; $k -lt $sarifLog.runs[$i].results[$j].fixes.Count; $k++)
+    {
+      $sarifLog.runs[$i].results[$j].fixes[$k].fileChanges.uri = Rebase-Uri($sarifLog.runs[$i].results[$j].fixes[$k].fileChanges.uri)
```
lgolding: `fileChanges` is itself an array: a single "fix" might require you to modify one or more files. Each changed file is represented by a `fileChange` object. The `fileChange` object, in turn, consists of a `uri` saying which file is being changed, together with an array of `replacement` objects; each `replacement` object specifies a single contiguous range of bytes to modify within the file.

So you'll need a nested loop over the elements of the `fileChanges` array.

appveyor/Convert-FileUrisToGitHubUris.ps1
lgolding: I choose names that accurately, specifically describe what the thing is (for variables, classes, etc) or does (for methods). Good names are one of the most important aspects of code that is readable, and code has to be readable so that you can maintain or enhance it. It's totally worthwhile to spend a couple of minutes to come up with a good name. We can talk more offline.

appveyor.yml
```diff
@@ -28,7 +34,7 @@ test:
 notifications:
   - provider: Email
     to:
-     - vsarcplat@microsoft.com
+     - t-rools@microsoft.com
```
rtaket: >t-rools [](start = 7, length = 7)

Change the email address back to vsarcplat@microsoft.com.

appveyor.yml
```diff
@@ -11,13 +11,19 @@ before_build:
...
+
 configuration:
 #  - Debug
   - Release
 
 build:
   project: src/Everything.sln
-  
+
+before_test:
```
rtaket: >before_test [](start = 0, length = 11)

Would after_build be more appropriate than before_test? I assume that after_build runs even when the build fails. But before_test probably only runs when the build succeeds.

lgolding: Agreed, I meant to mention that. The important thing the code needs to convey here is "you can't do this until the build is done". As written, the code conveys "you can't run the tests before you do this" -- which isn't true.

lgolding: (And Ryley's point is correct -- beyond the question of code clarity, there really is a semantic difference.)

appveyor/Push-PrefastResultsToAppveyor.ps1
rtaket: I think we should add a header (using comment string) to both ps1 files which describes their functionality.

---
In reply to: [70728786](https://github.com/Microsoft/sarif-sdk/pull/420#discussion_r70728786) [](ancestors = 70728786)

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+
+if($repoProvider -eq "gitHub")
+{
+  $githubBaseUri = "https://raw.githubusercontent.com/"
+}
+
+function Rebase-Uri($originalURI)
+{
+  $caseSensitiveUri = Get-CaseSensitivePath($originalURI.SubString($originalURI.IndexOf("c:/")))
+  return $githubBaseUri + $repoName+"/"+$repoCommit+"/"+$caseSensitiveUri.SubString($caseSensitiveUri.IndexOf($projectSlug)+$projectSlug.Length+1)
```
rtaket: >$githubBaseUri [](start = 9, length = 14)

I think we should use a UriBuilder to form the URL string. I believe it will take care of the string formatting issues we're seeing (backslashes instead of forwardslashes, etc.)

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+$githubBaseUri = ""
+
+if($repoProvider -eq "gitHub")
+{
+  $githubBaseUri = "https://raw.githubusercontent.com/"
+}
+
+function Rebase-Uri($originalURI)
+{
+  $caseSensitiveUri = Get-CaseSensitivePath($originalURI.SubString($originalURI.IndexOf("c:/")))
```
rtaket: >IndexOf [](start = 80, length = 7)

$originalURI.IndexOf("c:/") is kind of strange for a few reasons. First, assuming the string is always a path, won't this always evaluate to zero when it succeeds? So I don't understand how it would affect the Substring() call. Second, are we sure the only drive we need to be concerned about is the C:\ drive? And if so, is the drive letter always going to be lower cased?

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+
+  if ($directoryInfo.Parent -ne $null)
+  {
+      $parentPath = Get-CaseSensitivePath($directoryInfo.Parent.FullName)
+      $childPath = $directoryInfo.Parent.GetFileSystemInfos($directoryInfo.Name)[0].Name
+      return(Join-Path $parentPath $childpath -resolv)
+  }
+  else
+  {
+      return $directoryInfo.Name
```
rtaket: I'd add a comment stating that the root string (most likely the drive letter) is not converted.

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,120 @@
...
+    $sarifLog.runs[$i].configurationNotifications[$j].physicalLocation.uri = Rebase-Uri($sarifLog.runs[$i].configurationNotifications[$j].physicalLocation.uri);
+  }
+
+  for($j=0; $j -lt $sarifLog.runs[$i].toolNotifications.Count; $j++){
+    $sarifLog.runs[$i].toolNotifications[$j].physicalLocation.uri = Rebase-Uri($sarifLog.runs[$i].toolNotifications[$j].physicalLocation.uri);
+  }
+}
+
+$newExtension = ".github.sarif"
+$writeLocation = [System.IO.Path]::GetFileNameWithoutExtension($location)+$newExtension
```
rtaket: >GetFileNameWithoutExtension [](start = 35, length = 27)

You can use Path.ChangeExtension to apply the new file extension.

appveyor/Convert-FileUrisToGitHubUris.ps1
lgolding: Ryley, the file URL will be of the form `file:///c:/path/file`, so he's stripping the leading `file:///`.

As for the C: drive, I noticed that myself, but I decided to let it pass because we "know" what the AppVeyor's VMs look like. But I wouldn't object to making this a little more resilient by using `"file:///".Length"` as the starting index.

rtaket: I see. Thanks for the uri string example. Makes sense now.

appveyor.yml
Rolstenhouse: That makes more sense so it was changed

appveyor/Convert-FileUrisToGitHubUris.ps1
Rolstenhouse: Later, everything before the repository name will be rewritten anyway. I think dealing with the edge case of having the repository name as the root with incorrect casing, will not occur

appveyor/Convert-FileUrisToGitHubUris.ps1
```diff
@@ -0,0 +1,112 @@
...
+
+$builder = New-Object System.UriBuilder
+
+if($repoProvider -eq "gitHub"){
+  $builder.Scheme = 'https'
+  $builder.Host = 'raw.githubusercontent.com'
+}
+else{
+  throw "Repository provider i.e. gitHub, bitbucket ... is not recognized"
+}
```
lgolding: 1. Don't mention bitbucket if the script doesn't support it.
2. Do tell the user how to fix the problem. For example:

```
throw "Unrecognized repository provider: $repoProvider. Set the APPVEYOR_REPO_PROVIDER environment variable to a supported provider. Supported providers are: github."
```

lgolding:
:+1: except for changing the exception message (see my comment on the `throw`).

PR #423, opened 2016-07-12
Add 'callee' and 'calleeKey' to annotatedCodeLocation object.

michaelcfanning:
@lgolding @rtaket 

ryley, if you can merge this change and regenerate the telemetry build, I'd appreciate it. i'm temporarily taking a build with 'callee' (but not the telemetry work) in order to make progress against SDV direct production.


lgolding:
@rtaket Cancel that. I already had a PR out for this schema change. I'll merge it as soon as it's through AppVeyor.

PR #428, opened 2016-07-14
Convert method + tests

jericahuang:


src/Sarif.Viewer.VisualStudio.UnitTests/CodeFlowToTreeConverterTests.cs
```diff
@@ -40,7 +116,35 @@ public void CanConvertCodeFlowToTree()
...
+                    {
+                        Kind = AnnotatedCodeLocationKind.Declaration
+                    },
+                }
+            };
+
+            CallTreeNode root = CodeFlowToTreeConverter.Convert(codeFlow);
+
+            root.Children.Count.Should().Be(3);
+            root.Children[0].Children.Count.Should().Be(0);
```
lgolding: Good looking set of tests!

src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -4,14 +4,57 @@
...
+            };
+
+            return root;
+        }
+
+        private static List<CallTreeNode> GetChildren(CodeFlow codeFlow, ref int i)
+        {
+            i++;
+            List<CallTreeNode> children = new List<CallTreeNode>();
+            while (i < codeFlow.Locations.Count )
```
lgolding: A couple of suggestions on structuring:

1. If you use a `switch`, you won't have to repeat the expression `codeFlow.Locations[i].Kind` in each branch.
2. I'd expect to see calls before returns -- obviously it doesn't matter, but it reads more naturally to me that way -- so I'd handle the `Call` case first.

The only problem is that this switch is enclosed within a `while` loop, and you need to break out of the `while` loop when you find a `CallReturn`. And since C# doesn't have a "multi-level break" statement, you'll need an auxilliary flag.

So, something like this:

```C#
bool foundCallReturn = false;
while (i < codeFlowLocations.Count && !foundCallReturn)
{
    switch (codeFlow.Locations[i].Kind)
    {
        case AnnotatedCodeLocationKind.Call:
            ....
            break;

        case AnnotatedCallLocation.CallReturn:
            ...
            foundCallReturn = true;
            break;

        default:
            ...
            break;
    }
}
```

lgolding:
:clock1:

lgolding:
Ok, that's my only comment. See how that restructuring works for you, otherwise good!

lgolding:
Oh, I do have one more comment, about the naming of the variable "`i`". Let's find a more communicative name. What it is, is the current index into the codeFlow. So maybe "`codeFlowIndex`", or even "`currentCodeFlowIndex`" [if you're not into the whole brevity thing](http://entertainment.time.com/2011/06/10/bueller-bueller-top-10-quotable-movies/slide/the-big-lebowski/).

src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -4,14 +4,61 @@
...
+using System.Collections.Generic;
 
 namespace Microsoft.Sarif.Viewer.VisualStudio
 {
     internal static class CodeFlowToTreeConverter
     {
         internal static CallTreeNode Convert(CodeFlow codeFlow)
         {
-            throw new NotImplementedException();
+           int i = -1;
```
lgolding: Let's rename `i` to `currentCodeFlowIndex` here, too.

lgolding:
:+1:  Please make that one more variable rename, wait for AppVeyor, then go ahead and merge.

PR #432, opened 2016-07-14
Display real call tree in viewer

lgolding:
Add a list of `CallTree` objects to the view model. Populate them by converting each code flow to a nested call tree using @jericahuang and @kschecht's `CodeFlowToTreeConverter` class. Display them in a new tab in the "code locations" window. 

As we've discussed, I was wrong to suggest that the `CodeFlowToTreeConverter` produce a single, dummy node containing all the top-level code flow locations. Instead, it now produces a list of the top-level nodes. I adjusted the converter and the tests accordingly.

This is _just enough_ code to demonstrate that we get real code flow data in a real tree view. There's no styling, no data templating, no scroll bars -- I'm just displaying the `AnnotatedCodeLocation.Kind` to show that it's real. The fun part comes next.

image

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Sarif/CodeFlow.extensions.cs
+++ b/src/Sarif.Viewer.VisualStudio/Sarif/CodeFlow.extensions.cs
@@ -1,14 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved. 
 // Licensed under the MIT license. See LICENSE file in the project root for full license information. 
 
+using System.Collections.Generic;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.Sarif.Viewer.Models;
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using Microsoft.Sarif.Viewer.VisualStudio;
 
 namespace Microsoft.Sarif.Viewer.Sarif
 {
@@ -37,6 +33,18 @@ namespace Microsoft.Sarif.Viewer.Sarif
             }
 
             return model;
+        }
+
+        public static CallTree ToCallTree(this CodeFlow codeFlow)
+        {
+            if (codeFlow == null || codeFlow.Locations.Count == 0)
+            {
+                return null;
+            }
+
+            List<CallTreeNode> topLevelNodes = CodeFlowToTreeConverter.Convert(codeFlow);
+
+            return new CallTree(topLevelNodes);
         }
     }
 }
--- a/src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
+++ b/src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
@@ -1,0 +1,26 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:local="clr-namespace:Microsoft.Sarif.Viewer.Themes"
+                    xmlns:shell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0">
+
+    <ResourceDictionary.MergedDictionaries>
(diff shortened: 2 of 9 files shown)


src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
```diff
@@ -0,0 +1,26 @@
...
+    </ResourceDictionary.MergedDictionaries>
+
+    <DataTemplate x:Key="CallTreeTemplate">
+        <ListView ItemsSource="{Binding}">
+            <ListView.ItemTemplate>
+                <DataTemplate>
+                    <TreeView ItemsSource="{Binding TopLevelNodes}">
+                        <TreeView.ItemTemplate>
+                            <HierarchicalDataTemplate DataType="x:Type local:CallTreeNode" ItemsSource="{Binding Children}">
+                                <TextBlock Text="{Binding Location.Kind}" />
```
lgolding: @jericahuang @kschecht Although it won't be obvious to you without some deeper understanding of WPF, this is the place where we'll put all the fancy data about each location. Here's where we'll decide what text to display, and what color the font and the background will be, to match SDV's display. And we'll have to figure out what how to display the "state data" for each location.

src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
```diff
@@ -153,6 +154,23 @@
                 <ContentControl ContentTemplate="{StaticResource AnnotatedCodeLocationCollectionDataGridTemplate}"
                                 Content="{Binding CodeFlows}" />
             </TabItem>
+            <TabItem Visibility="{Binding CallTrees, Converter={StaticResource collectionToVisibility0Converter}}">
```
lgolding: @jericahuang @kschecht Here's where we add the extra tab you see in the screen shot.

src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
```diff
@@ -153,6 +154,23 @@
...
+                <TabItem.Header>
+                    <DockPanel Margin="5 0 5 0">
+                        <TextBlock DockPanel.Dock="Left"
+                                   Text="Call Trees" />
```
lgolding: Here's where the title of the tab comes from.

src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
```diff
@@ -153,6 +154,23 @@
...
+                        <TextBlock DockPanel.Dock="Left"
+                                   Visibility="{Binding CallTrees, Converter={StaticResource collectionToVisibility1Converter}}">
+                            <Run Text=" (" />
+                            <Run Text="{Binding CallTrees, Mode=OneWay, Converter={StaticResource collectionToCountConverter}}" />
+                            <Run Text=")" />
+                        </TextBlock>
+                    </DockPanel>
+                </TabItem.Header>
+
+                <ContentControl ContentTemplate="{StaticResource CallTreeTemplate}"
```
lgolding: Here's where we display the actual call trees, using the "template" defined in CallTrees.xaml, above.

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlow.extensions.cs
```diff
@@ -38,5 +34,17 @@ public static AnnotatedCodeLocationCollection ToAnnotatedCodeLocationCollection(
 
             return model;
         }
+
+        public static CallTree ToCallTree(this CodeFlow codeFlow)
+        {
+            if (codeFlow == null || codeFlow.Locations.Count == 0)
```
michaelcfanning: >codeFlow == null || codeFlow.Locations [](start = 16, length = 38)

hm, does the following not work for this? also, what does our schema say about ensuring Locations is non-null? Does it make sense for it to be non-null but empty?

if (codeFlow?.Locations.Count == 0)

michaelcfanning:
:shipit:

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlow.extensions.cs
lgolding: Absolutely, I'll change it to use the null conditional operator.

The spec says that `locations` is required and must have at least one element. I think we need to be resilient against invalid SARIF, though.

PR #433, opened 2016-07-15
Display line number and callee in calltree nodes

lgolding:
This is the start of the converter to display the correct text in the call tree nodes.

What's missing from the SDV converter is that, on non-call nodes, they show code lines that don't actually appear as "snippets" in the SARIF file; they're obviously getting it from the actual source files.

Also note that most of the nodes don't have line numbers because the SARIF codeFlow doesn't have them.

Still no scroll bars or styling.

image


src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
```diff
@@ -64,6 +65,11 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>Resource1.resx</DependentUpon>
     </Compile>
+    <Compile Include="Resources.Designer.cs">
```
rtaket: Just FYI, there's already a resource file in the vsix, resource1.resx. Your file name is better though. Also, resource1.resx was intended to be temporary. We needed to include an external file in the vsix, but I couldn't find the correct vsixmanifest settings. As a workaround, we embedded the file in resource1.resx until I figured out the manifest. That was a long way of saying, we should keep your resources.resx file and I'll remove resource1.resx in a future commit.

rtaket:
:+1:

src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
lgolding: Yes, I noticed resource1.resx, and I was going to rename it in a future commit as soon as I figured out what it was for :-) But now I'll just leave it to you.

PR #436, opened 2016-07-15
Checking if Calls and CallReturns are balanced + tests

jericahuang:


lgolding:
Interesting idea! But there's nothing in the SARIF spec that requires calls and returns to be balanced.

By the spec, a code flow is a "sequence of program locations that specify a possible execution path through the code". We include them in the spec so that we can represent the execution path that lead to the problem.

So it's perfectly possible for the code flow to be:

1. f() called g()
2. g() called h()
3. h() called j()
4. BOOM!

However... take a look at my comment in the code, where I'll show you a way to use FluentAssertions to write tests where you expect an exception to be thrown.

src/Sarif.Viewer.VisualStudio.UnitTests/CodeFlowToTreeConverterTests.cs
```diff
@@ -173,5 +173,193 @@ public void CanConvertCodeFlowToTreeOnlyDeclarations()
...
+            catch (System.ArgumentException)
+            {
+                exceptionThrown = true;
+            }
+
+            if (!exceptionThrown)
+            {
+                throw new System.Exception();
+            }
+
```
lgolding: FluentAssertions has a [lovely facility](https://github.com/dennisdoomen/fluentassertions/wiki#exceptions) for this:

```C#
Action action = () => CodeFlowToTreeConverter.Convert(codeFlow); // C# lambda syntax

action.ShouldThrow<ArgumentException>();
```

lgolding:
Come to think of it, you can repurpose this PR to add a unit test that shows that your converter works properly with an unbalanced tree.

src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -27,14 +29,24 @@ private static List<CallTreeNode> GetChildren(CodeFlow codeFlow, ref int current
...
                             Location = codeFlow.Locations[currentCodeFlowIndex],
-                            Children = GetChildren(codeFlow, ref currentCodeFlowIndex)
+                            Children = GetChildren(codeFlow, ref currentCodeFlowIndex, ref unReturnedCalls)
                         });
                         break;
 
                     case AnnotatedCodeLocationKind.CallReturn:
+                        if (unReturnedCalls.Count == 0)
+                        {
+                            throw new System.ArgumentException("At least one AnnotatedCodeLocation Call in this CodeFlow is not paired with a CallReturn, causing an imbalanced tree.");
```
rtaket: InvalidDataException might be better. ... although as Larry stated, the SARIF is valid.

PR #437, opened 2016-07-15
Fix #409 Space between "File" and "Name"

### LINKED ISSUE #409: VSIX: Info Tab: The tool file name property name should have a space between File and Name.
Repro Steps
1. Open the attached sarif file
2. Double-click on any of the issues in the Error List.
3. Select the Info tab.

Expected: The file name property should read "File name:"
Actual: The file name property reads "FileName:"

See attached screenshot.

filename
[AttributeAnalysis.zip](https://github.com/Microsoft/sarif-sdk/files/343905/AttributeAnalysis.zip)

lgolding:
Also:

* Use sentence casing on data item labels.
* Use "level" instead of "severity".
* Don't put spaces before colons.
* Add some dummy data to display in the call tree tab at design time.


rtaket:
:+1:

PR #438, opened 2016-07-15
Fix #435: Remove non-existent Tool and Rule properties

### LINKED ISSUE #435: VSIX: Remove non-existent properties from Tool and Rule
Information.xaml refers to properties that don't exist in SARIF, like Tool.OwnerName and Rule.Version. Remove them from the XAML and from the underlying view model and model classes.

lgolding:
These properties aren't defined in SARIF:
* Rule.Version
* Rule.OwnerName
* Rule.FeedbackUri (but HelpUri does exist, so show that)
* Tool.HelpUri
* Tool.OwnerName
* Tool.FeedbackUri
* Tool.Description (although it's not a bad idea)

Also:

* Add some design-time dummy data to the Call Trees tab.
* Add a little margin below the tool name/version in the Info tab
* Display the actual HelpUri instead of the text "Help" for the hyperlink.

Here's what it looks like now, with dummy data inserted for Category and HelpUri:

image

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/RuleModel.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/RuleModel.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Sarif.Viewer.Models
         private string _id;
         private string _name;
         private string _category;
-        private string _severity;
+        private string _defaultLevel;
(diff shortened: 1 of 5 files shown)


rtaket:
:+1:

PR #440, opened 2016-07-15
Fix #378: Don't offer ".sarif.json" as an extension

### LINKED ISSUE #378: ConvertToSarif default output filename defaults to ".sarif"
"A file path to the converted SARIF log. Defaults to <input file name>.sarif."

In other analysis tool work, I've seen references that Sarif output files should be named ".sarif.json". Is there a preferred naming style for Sarif output files from tools?

lgolding:
The official recommended extension for SARIF files is ".sarif", not ".sarif.json", for reasons discussed in #378. So remove mentions of ".sarif.json" from the VS SARIF VSIX.

@bobschumaker FYI


michaelcfanning:
+1

PR #441, opened 2016-07-15
Tweak "tool" property descriptions

lgolding:
... for consistency with the spec. This addresses sarif-standard/sarif-spec#176, "Review SDK tool definition to ensure consistency with spec" (that work item really should've been in the sdk repo, but no matter).


michaelcfanning:
+1

PR #444, opened 2016-07-17
Add spec sample files to validation tests

lgolding:
The files in the new Sarif.FunctionalTests\SpecSamples directory are copied from Annex G of the SARIF spec. The idea is to ensure that the files in the spec are valid.

Also:
* DRY out the validation tests. With one small exception, the logic is the same for all the test file subdirectories. So just aggregate the list of files over all subdirectories, and run a single test over the aggregated list.

**NOTES**

1. The "comprehensive" sample file is not yet finished.

2. Look out for the corresponding PR that adds Annex G to the spec; I won't send that PR until the "comprehensive" sample is finished.

### DIFF
--- a/src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
+++ b/src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
@@ -39,8 +39,17 @@
         }
       },
       "files": {
-        "file:///user/builder/work/src/collections/list.cpp": {
-          "mimeType": "text/x-c"
+        "collections/list.cpp": {
+          "uri": "collections/list.cpp",
+          "uriBaseId": "SRCROOT",
+          "mimeType": "text/x-c",
+          "length": 980,
+          "hashes": [
+            {
+              "algorithm": "sha256",
+              "value": "b13ce2678a8807ba0765ab94a0ecd394f869bc81"
+            }
+          ]
         }
       },
       "logicalLocations": {
@@ -62,16 +71,47 @@
       "results": [
         {
           "ruleId": "C2001",
-          "message": "Variable \"n\" was used without being initialized.",
+          "formattedRuleMessage": {
+            "formatId": "default",
+            "arguments": [
+              "n"
+            ]
+          },
           "locations": [
             {
               "analysisTarget": {
-                "uri": "file:///user/builder/work/src/collections/list.cpp",
+                "uri": "collections/list.cpp",
                 "region": {
                   "startLine": 15
                 }
               },
               "fullyQualifiedLogicalName": "collections::list:add"
+            }
+          ],
+          "codeFlows": [
+            {
+              "message": "Code flow #1",
+              "locations": [
+                {
+                  "step": 0,
+                  "kind": "declaration",
+                  "importance": "essential",
+                  "fullyQualifiedLogicalName": "collections::list::add",
+                  "physicalLocation": {
+                    "uri": "collections/list.cpp",
+                    "uriBaseId": "SRCROOT",
+                    "region": {
+                      "startLine": 12,
+                      "startColumn": 5,
+                      "endLine": 12,
+                      "endColumn": 11,
+                      "offset": 115,
+                      "length": 6
+                    }
+                  },
+                  "snippet": "int n;"
+                }
+              ]
             }
           ]
         }
@@ -99,7 +139,14 @@
       "rules": {
         "C2001": {
           "id": "C2001",
-          "fullDescription": "A variable was used without being initialized. This can result in runtime errors such as null reference exceptions"
+          "name": "UninitializedVariable",
+          "shortDescription": "A variable was used without being initialized.",
+          "fullDescription": "A variable was used without being initialized. This can result in runtime errors such as null reference exceptions",
+          "messageFormats": {
+            "default": "Variable \"{0}\" was used without being initialized."
+          },
+          "defaultLevel": "error",
+          "helpUri": "http://www.example.com/tools/codescanner/rules/C2001.aspx"
         }
       }
     }
--- a/src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
+++ b/src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
@@ -1,0 +1,107 @@
+{
+  "version": "1.0.0",
+  "$schema": "http://json.schemastore.org/sarif-1.0.0",
+  "runs": [
+    {
+      "id": "BC650830-A9FE-44CB-8818-AD6C387279A0",
+      "stableId": "Nightly code scan",
+      "baselineId": "0A106451-C9B1-4309-A7EE-06988B95F723",
+      "automationId": "Build-14.0.1.2-Release-20160716-13:22:18",
+      "architecture": "x86",
+      "tool": {
+        "name": "CodeScanner",
+        "fullName": "CodeScanner 1.1 for Unix (en-US)",
+        "version": "2.1",
+        "semanticVersion": "2.1.0",
+        "fileVersion": "2.1.0.0",
+        "language": "en-US",
+        "sarifLoggerVersion": "1.25.0",
+        "properties": {
+          "copyright": "Copyright (c) 2016 by Example Corporation. All rights reserved."
+        }
+      },
+      "invocation": {
+        "commandLine": "CodeScanner @collections.rsp",
+        "responseFiles": {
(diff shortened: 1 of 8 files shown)


src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -19,8 +19,6 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public class SarifValidatorTests
     {
```
lgolding: It might be easier to read this file as new code, since it's so heavily refactored. The test for the "converter" directory and the "direct producer" directory were already very similar. The test for the new "spec samples" directory would have been identical to those for the direct producer directory. So aggregate the files for all three directories and write a single data-driven test.

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
```diff
@@ -0,0 +1,107 @@
...
+        },
+        "collections": {
+          "name": "collections",
+          "kind": "namespace"
+        }
+      },
+      "results": [
+        {
+          "ruleId": "C2001",
+          "message": "Variable \"n\" was used without being initialized.",
```
michaelcfanning: This almost looks like a line-break, \n due to escaping. :) suggest altering name slightly. if you use single quote, then the SARIF is more readable. for most internal tools, we use single quotes.

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
```diff
@@ -0,0 +1,107 @@
...
+                }
+              },
+              "fullyQualifiedLogicalName": "collections::list:add"
+            }
+          ]
+        }
+      ],
+      "configurationNotifications": [
+        {
+          "id": "UnknownRule",
```
michaelcfanning: How about ExampleRule, following your Example Corporation, above?

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
```diff
@@ -0,0 +1,55 @@
...
+  "version": "1.0.0",
+  "runs": [
+    {
+      "tool": {
+        "name": "CodeScanner",
+        "semanticVersion": "2.1.0"
+      },
+      "files": {
+        "file:///user/builder/work/src/collections/list.cpp": {
+          "mimeType": "text/x-c"
```
michaelcfanning: incorrect mime type

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
```diff
@@ -0,0 +1,55 @@
...
+        },
+        "collections": {
+          "name": "collections",
+          "kind": "namespace"
+        }
+      },
+      "results": [
+        {
+          "ruleId": "C2001",
+          "message": "Variable \"n\" was used without being initialized.",
```
michaelcfanning: so, recommended doesn't make use of formattedMessages, eh?

src/Sarif.FunctionalTests/SpecExamples/RuleMetadata.sarif
```diff
@@ -0,0 +1,17 @@
...
+  "runs": [
+    {
+      "tool": {
+        "name": "CodeScanner",
+        "semanticVersion": "2.1.0"
+      },
+      "rules": {
+        "C2001": {
+          "id": "C2001",
+          "fullDescription": "A variable was used without being initialized. This can result in runtime errors such as null reference exceptions"
```
michaelcfanning: we should build out this example to include more rules metadata, especially since this file is small

michaelcfanning: btw - binskim provides good examples that may be scrubbable

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
```diff
@@ -0,0 +1,55 @@
...
+          "locations": [
+            {
+              "analysisTarget": {
+                "uri": "file:///user/builder/work/src/collections/list.cpp",
+                "region": {
+                  "startLine": 15
```
michaelcfanning: is a startLine only recommended? would suggest adding more here. let's talk tomorrow about how you're defining recommended.

michaelcfanning:
+1

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
lgolding: I defined that in Annex G of the spec, which you haven't seen yet. It just means "properties that the spec says 'should' be present."

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
lgolding: Changed name to `count`. Kept double quotes.

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
lgolding: I wasn't using `UnknownRule` to mean "some arbitrary rule", I meant it in the sense of the associated message "Your command line mentioned a rule I don't know about."

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
lgolding: That's correct. [`result.formattedRuleMessage`](https://rawgit.com/sarif-standard/sarif-spec/master/Static%20Analysis%20Results%20Interchange%20Format%20(SARIF).html#rule-messageformats) is an optional property ("A _result_ object **may** contain a property named `formattedRuleMessage`...). Annex G, which you haven't seen yet uses "recommended" to mean "properties that the spec says 'should' be present."

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
lgolding: This is actually correct, or at least it is "a" correct answer, because:

* [IANA](http://www.iana.org/assignments/media-types/media-types.xhtml) doesn't define a MIME type for C++.

* [freeformatter.com](http://www.freeformatter.com/mime-types-list.html#mime-types-list) (which gets its list from "the 'mime.types' file of the Apache HTTPD [virtual private server](http://www.webhostingsearch.com/virtual-private-server.php)") gives `x-c` for "C source file" but does not mention C++, but...

* A [stackoverflow page](http://stackoverflow.com/questions/1735659/list-of-all-mimetypes-on-the-planet-mapped-to-file-extensions) links to the [latest version](http://svn.apache.org/viewvc/httpd/httpd/trunk/docs/conf/mime.types?view=markup) of the Apache mime.types file, which says:

    ```
    text/x-c					c cc cxx cpp h hh dic
    ```

src/Sarif.FunctionalTests/SpecExamples/RuleMetadata.sarif
lgolding: Sounds good. I made a note in the Annex to expand the rules metadata example. I'll update it here once the Annex is accepted.

src/Sarif.FunctionalTests/SpecExamples/Recommended.sarif
lgolding: Oh, and to answer your question: `startLine` is required for a text file (assuming the result has a region at all, which it need not). But my "minimal" example, which by design is the absolutely minimal valid SARIF file, doesn't have any results in it.

I wouldn't spend a lot of effort deciding exactly what goes in each example. The problem is nesting. A minimal file doesn't have to have any `result`s. If it does have a `result`, the `result` doesn't have to have any `location`s. If it does have a `location`, then the `physicalLocation` (that is, either the `analysisTarget` or the `resultFile`) doesn't have to have a `region`. If it does have a `region`, _and_ it's in a text file -- ahh, _then_ `startLine` is required. It isn't practical to show examples with each level of nesting.

While I'm on the topic of the difficulty of crafting these examples: I can't give a really "comprehensive" example, either, in the sense of #35 -- that is, an example that demonstrates every property -- because some properties are mutually exclusive. For example, a `result` can't have both a `message` and a `formattedRuleMessage` (I'll demonstrate `message` in the "recommended" example and `formattedRuleMessage` in the "comprehensive" example).

Bottom line, I consider this a "best effort" exercise, with the goal of showing implementers realistic examples. But it can't be exhaustive.

PR #447, opened 2016-07-18
Importance to Background Converter

jericahuang:


src/Sarif.Viewer.VisualStudio/Converters/ImportanceToBackgroundConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Windows.Data;
+
+namespace Microsoft.Sarif.Viewer.Converters
+{
+    class ImportanceToBackgroundConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var node = value as CallTreeNode;
+            return node.Location.Importance == AnnotatedCodeLocationImportance.Important ? Color.Yellow : Color.White;
```
rtaket: If value is not of type CallTreeNode, then node will be null and you'll raise a null reference exception on line 19. Verify node is not null before calling node.Location.

rtaket: Instead of white for the default background color, should we use transparent instead?

src/Sarif.Viewer.VisualStudio/Converters/ImportanceToBackgroundConverter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.CodeAnalysis.Sarif;
```
rtaket: Can you add the copyright/license comment to this file. Just copy it from one of the other files. 
Also, do the same for ImportanceToForegroundConverter.cs, if required.

src/Sarif.Viewer.VisualStudio.UnitTests/ImportanceToForegroundConverterTests .cs
```diff
@@ -0,0 +1,117 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
lgolding: I'm confused. Why does this file show up as new code, when @kschecht already merged his changes?

lgolding:
:+1:

src/Sarif.Viewer.VisualStudio.UnitTests/ImportanceToForegroundConverterTests .cs
jericahuang: Possibly because I filed this PR before @kschecht 's was merged and I did a fetch and rebase to fix merge conflicts?

PR #448, opened 2016-07-18
Exclamation point visibility

kschecht:


lgolding:
I'm going to merge this, but let's look at that "use strings in converters" issue this morning. I'm not convinced :)

PR #451, opened 2016-07-19
Importance converters return brushes, not strings

lgolding:
At first, the converters did not work because they produced `Color` values, but `TextBlock.Foreground` and `Background` are of type `Brush`. @rtaket suggested a fix where the converters would return a string, which works because WPF has a converter that can convert a string into a `Brush` (in this case, it creates a `SolidColorBrush`).

But really the fix should be for the converters to return `SolidColorBrush` objects, and that's what I do in this PR. I also simplify the converter logic.


rtaket:
:+1:

PR #453, opened 2016-07-20
Drop down .NET framework dependency to 4.5. Update release history. R…

michaelcfanning:
…evise working version of package/other builds. @lgolding @rtaket


lgolding:
:+1:

PR #456, opened 2016-07-20
Fix broken validation tests

lgolding:
The validation tests project must refer to Sarif.dll via the same NuGet package that JSchema (upon which it depends) was built against.

BuildAndTest.cmd now passes.

Also:
* Update version files

### DIFF
--- a/src/Sarif.Driver/VersionConstants.cs
+++ b/src/Sarif.Driver/VersionConstants.cs
@@ -4,8 +4,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 {                                                                              
     public static class VersionConstants                                       
     {                                                                          
-        public const string Prerelease = "";                       
-        public const string AssemblyVersion = "1.5.25";       
+        public const string Prerelease = "-developer";                       
+        public const string AssemblyVersion = "1.5.26";       
         public const string FileVersion = AssemblyVersion + ".0";              
         public const string Version = AssemblyVersion + Prerelease;            
     }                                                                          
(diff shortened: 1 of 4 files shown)


src/Sarif.Driver/VersionConstants.cs
```diff
@@ -4,8 +4,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 {                                                                              
     public static class VersionConstants                                       
     {                                                                          
-        public const string Prerelease = "";                       
-        public const string AssemblyVersion = "1.5.25";       
+        public const string Prerelease = "-developer";                       
+        public const string AssemblyVersion = "1.5.26";
```
lgolding: Since you've already published 1.5.26, shouldn't this be 1.5.27?

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -39,7 +39,11 @@
     <Reference Include="System.Data" />
     <Reference Include="System.Net.Http" />
     <Reference Include="System.Xml" />
-    <Reference Include="xunit.abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c, processorArchitecture=MSIL">
+    <Reference Include="Sarif, Version=1.5.25.0, Culture=neutral, PublicKeyToken=a567ef0185dba9aa, processorArchitecture=MSIL">
```
lgolding: This is disturbing. In the PR where you dropped down to v4.5.1 (PR #453), where you _removed_ the NuGet reference to the previous version of SARIF, the `PublicKeyToken` was `06c70c7db0c6cd6b`. Now it won't build unless I use `a567ef0185dba9aa`. Is it possible you built 1.5.25 with a different GeneratedKey.snk file?

PR #457, opened 2016-07-20
VSIX: Enable vertical scrolling for the tab items

rtaket:
Enabled vertical scrolling for the tab items
Fixed DockPanel.Dock=Bottom bug in the items panel
Call Trees treeview expands to fill the remaining whitespace. https://github.com/Microsoft/sarif-viewer-vsix/issues/12

Here's the screenshots with the scrollbars.

codeflowswithscroll
calltreewithscroll

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
+++ b/src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
@@ -9,8 +9,20 @@
     </ResourceDictionary.MergedDictionaries>
 
     <DataTemplate x:Key="CallTreeTemplate">
-        <ListView ItemsSource="{Binding}">
-            <ListView.ItemTemplate>
+        <ItemsControl ItemsSource="{Binding}">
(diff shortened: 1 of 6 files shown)


lgolding:
:+1:

michaelcfanning:
This is an improvement. We need to dispense with the yellow highlighting, which isn't appropriate to appear in a tool window. I suggest formatting each importance level (unimportant, important, and essential) as grayed out, normal and bolded, respectively. We need to stop inlining the excalamation point and red is a bad color from an accessibility standpoint. For this, suggest adding a new column to hold this information and to use either the VS profiling 'hotspot' icon to indicate essential or to use a simple blue circle with a white exclamation point (which should provide sufficient contrast to display the exclamation point, without unduly disrupting the overall color scheme).

michaelcfanning:
@jericahuang @Rolstenhouse @kschecht 

Adding explorers to raise visibility on feedback. It's great to see the infrastructure/styling improvements. When we get a bit further, we can run this work past an actual UX designer internally.

PR #460, opened 2016-07-21
VSIX: Call Tree: Bold essential nodes.

rtaket:
- Bold essential nodes.
- Do not highlight rows in treeview.
- Remove red bang for essential nodes.
- Expand essential nodes.

treenodestyle


rtaket:
I'm not sure I like this new style. Would appreciate opinions...

src/Sarif.Viewer.VisualStudio/Themes/CallTrees.xaml
```diff
@@ -31,22 +31,23 @@
...
+                        </TreeView.ItemContainerStyle>
+
                         <TreeView.ItemTemplate>
                             <HierarchicalDataTemplate DataType="x:Type local:CallTreeNode" ItemsSource="{Binding Children}">
-                                <StackPanel Orientation="Horizontal">
-                                    <TextBlock Text="! " FontWeight="Bold" Foreground="Red"
-                                               Style="{StaticResource callTreeTextStyle}"
-                                               Background="{Binding Location.Importance, 
-                                                            Converter={StaticResource importanceToBackgroundConverter}}"
-                                               Visibility="{Binding Location.Importance,
```
lgolding: If you're going to remove the background coloring, then remove the `importanceToBackgroundConverter` resource, the `ImportanceToBackgroundConverter` class, and its unit tests.

lgolding: Likewise if you're removing the red-bang, then remove the `importanceToEssentialMarkerVisibilityConverter` resource, the `ImportanceToEssentialMarkerVisibilityClass`, and its unit tests.

src/Sarif.Viewer.VisualStudio/Themes/DefaultStyles.xaml
```diff
@@ -49,6 +49,23 @@
                 Value="Courier New" />
         <Setter Property="FontSize"
                 Value="12" />
+        <Style.Triggers>
+            <DataTrigger Binding="{Binding Location.Importance}"
+                         Value="Essential">
+                <Setter Property="FontWeight"
+                        Value="Bold" />
```
lgolding: Huh, for some reason I never think of using style triggers for things like this. I would've written an `ImportanceToFontWeightConverter`. Your way is much better.

lgolding:
:+1: Subject to

1. @michaelcfanning weighing in on appearance.
2. Removal of now-unused resources, classes, and tests.

src/Sarif.Viewer.VisualStudio/Themes/DefaultStyles.xaml
lgolding: ... and for that matter you could get rid of the `ImportanceToForegroundConverter` and its unit tests, and use style triggers for that as well.

jericahuang:
I've also been experimenting with @michaelcfanning 's styling suggestions. Here's what the exclamation point in a blue circle would look like...
bluebang

rtaket:
Maybe we should steal the error list icons.

image

lgolding:
Good so far. You can also remove the `ImportanceToEssentialMarkerVisibility` class, and its tests, and the resource.

lgolding:
:+1: @michaelcfanning has yet to weigh in on styling. But with these changes, it's so trivial to update the styling that I'd say, go ahead and merge this, if only for the sake of getting rid of all that code.

PR #461, opened 2016-07-21
VSIX: Navigate from Call Tree to source code

rtaket:
There were a few changes required to make this work.
- Add an attached property so we can get the SelectedItem of the TreeView. (There's no built-in way to do this.)
- Update the CallTreeNode model to extend CodeLocationObject. CodeLocationObject is the type which enables all of the navigate to source logic.
- Wire up the CallTreeNodes to the path rebasing and document attaching features.

navigatefromcalltree


src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
```diff
@@ -1,14 +1,54 @@
...
+    public class CallTreeNode : CodeLocationObject
     {
-        public AnnotatedCodeLocation Location { get; set; }
+        private AnnotatedCodeLocation _location;
+
+        public AnnotatedCodeLocation Location
+        {
+            get
+            {
+                return this._location;
```
lgolding: Are we doing "`.this`" in this repo?

src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
```diff
@@ -1,14 +1,54 @@
...
+            }
+            set
+            {
+                this._location = value;
+
+                if (value != null && value.PhysicalLocation != null)
```
lgolding: Or use null-conditional operator:
```C#
if (value?.PhysicalLocation != null)
```

src/Sarif.Viewer.VisualStudio/TreeViewHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Microsoft.Sarif.Viewer
+{
+    /// <summary>
+    ///  The TreeViewHelper class enables binding to the SelectedItem of a TreeView.
+    /// </summary>
+    /// <remarks>
+    /// Original source code taken from http://stackoverflow.com/questions/7153813/wpf-mvvm-treeview-selecteditem.
+    /// </remarks>
+    public class TreeViewHelper
```
lgolding: static class

src/Sarif.Viewer.VisualStudio/TreeViewHelper.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) Microsoft. All rights reserved. 
+// Licensed under the MIT license. See LICENSE file in the project root for full license information. 
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
```
lgolding: Remove unnecessary usings.

lgolding:
ok, there's alot of plumbing here that I don't understand, but as far as I can understand it this looks ok, +/- a few nitpicks.

:+1:

src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
rtaket: We probably shouldn't use "this."s to be consistent. I'll remove them in the files in this PR. Likely there are other places in the sln where I've introduced "this."'s. I guess we can remove them as we touch the files.

src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
rtaket: Fixed

src/Sarif.Viewer.VisualStudio/TreeViewHelper.cs
rtaket: Fixed

src/Sarif.Viewer.VisualStudio/TreeViewHelper.cs
rtaket: Fixed

PR #462, opened 2016-07-21
Created Parent property for CallTreeNode

kschecht:
Created a Parent property for CallTreeNode to help with implementation of certain features, including F7/F8 step traversal (Microsoft/sarif-viewer-vsix#20).


src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -30,15 +30,17 @@ private static List<CallTreeNode> GetChildren(CodeFlow codeFlow, ref int current
                         children.Add(new CallTreeNode
                         {
                             Location = codeFlow.Locations[currentCodeFlowIndex],
-                            Children = GetChildren(codeFlow, ref currentCodeFlowIndex)
+                            Parent = parent
                         });
+                        children[children.Count - 1].Children = GetChildren(codeFlow, ref currentCodeFlowIndex, children[children.Count - 1]);
```
lgolding: I had to stare at this for a while to understand it: You need to pass the newly created node to _its own children_ as their parent, and the newly created node doesn't have a name yet, so you identify it as "the most recently added child". Let me think if this can be written more clearly...

lgolding: Ok, how about this:

```
                        var newNode = new CallTreeNode
                        {
                            Location = codeFlow.Locations[currentCodeFlowIndex],
                            Parent = parent
                        };
                        newNode.Children = GetChildren(codeFlow, ref currentCodeFlowIndex, newNode);
                        children.Add(newNode);
                        break;
```

lgolding:
A couple of general comments:

1. This is new functionality, so needs to affect the unit tests.
2. Although it is _plausible_ to me that F7/F8 will require the `Parent` property, it's not obvious to me. I suggest that you implement the actual feature in the same pull request, so we can see the use of the property you're adding.

In this case, my usual preference for small change sets is in tension with my desire to be sure we never add any unnecessary code.

kschecht:
Working on unit tests as we speak. I'll include your modifications to the code and then work on the F7/F8. The other reason I made this a separate PR was in case it would help @jericahuang with intelligent expand.

lgolding:
@kschecht That's a good thought, but @jericahuang can work on Expand all/Collapse all without that, right? I suggest that Jerica do each of those three (Expand all, Collapse all, intelligent expand) as separate PRs, since each of them is a valid user story on its own.

lgolding:
Ok, that all looks good. I'm going to stick to my suggestion that you include the actual F7/F8 functionality in this PR, to demonstrate that it really does require the `Parent` property.

lgolding:
Now just waiting on the unit tests, right?

rtaket:
@kschecht The merge we did last week was bad. I submitted a PR to your branch to fix the merge. You should review and merge it.

src/Sarif.Viewer.VisualStudio/Models/CallTree.cs
```diff
@@ -37,34 +42,184 @@ public ObservableCollection<CallTreeNode> TopLevelNodes
...
                 }
+            }
+        }
 
-                _selectedItem = value;
-                this.NotifyPropertyChanged(nameof(SelectedItem));
+        private static int GetIndexInCallTreeNodeList(IList<CallTreeNode> list, CallTreeNode givenNode)
+        {
+            int index = 0;
+            foreach (CallTreeNode listNode in list)
```
rtaket: >list [](start = 46, length = 4)

Just FYI, you'll raise a NullReferenceException if list is null. Raising the exception seems fine since it's an unexpected condition. But maybe we should validate the argument and throw an ArgumentNullException instead.

src/Sarif.Viewer.VisualStudio/Models/CallTree.cs
```diff
@@ -37,34 +42,184 @@ public ObservableCollection<CallTreeNode> TopLevelNodes
...
+                    index++;
+                }
+            }
 
-                // Navigate to the source of the selected node and highlight the region.
-                if (_selectedItem != null)
+            /* this exception should never be thrown, as this method should only be called when 
+             * givenNode is already known to be a member of list 
+             */
+            throw new System.Exception("Given node was not in the list.");
```
rtaket: >Exception [](start = 29, length = 9)

You should probably throw something more specific than System.Exception (so your callers don't need to catch System.Exception.) Maybe an IndexOutOfRangeException?

src/Sarif.Viewer.VisualStudio/Models/CallTree.cs
```diff
@@ -37,34 +42,184 @@ public ObservableCollection<CallTreeNode> TopLevelNodes
...
                 if (_selectedItem != null)
                 {
                     _selectedItem.ApplyDefaultSourceFileHighlighting();
                 }
+            }
+        }
 
-                _selectedItem = value;
-                this.NotifyPropertyChanged(nameof(SelectedItem));
+        private static int GetIndexInCallTreeNodeList(IList<CallTreeNode> list, CallTreeNode givenNode)
```
rtaket: >static [](start = 16, length = 6)

Does this method need to be static?

src/Sarif.Viewer.VisualStudio.UnitTests/CallTreeTraversalTests.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
```
rtaket: Sort using statements.

src/Sarif.Viewer.VisualStudio.UnitTests/CallTreeTraversalTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.Sarif.Viewer.Models;
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
```
rtaket: You should probably add a test where SelectedItem == null.

rtaket:
:+1:

PR #463, opened 2016-07-22
Fix Microsoft/sarif-viewer-vsix#27

lgolding:
Microsoft/sarif-viewer-vsix#27: Can't open SARIF file without any results

If you try to open a SARIF file with no results (because it only contains rule metadata), you get a null reference exception.

Also:
* If you opened a SARIF file with no results, it left the existing Error List entries in place. Don't do that; clear them.


rtaket:
👍

Issue #431, opened 2016-07-14
URI serialization should always produce a valid URL for relative file paths


michaelcfanning:
Currently does not. @nguerrera  has authored a helper for Roslyn we should steal for this.

lgolding:
Reopened to track the fix for handling dot-segments.

lgolding:
Finished off with PR #465.

PR #458, opened 2016-07-20
Fix #431: Properly serialize URIs

lgolding:
Borrow @nguerrera's URI serialization helper from Roslyn to ensure that URIs are correctly serialized to SARIF.

Update the test files to reflect the new (correct) serialization.

Also, fix problems with my implementation of the SDV test files into the functional tests:
* SDV's native output files have extension `.tt`, not  `.xml`, so the functional tests did not find any files to convert. So the SDV tests appeared to pass because there were no files.
* Then once we fixed that, it turned out that for one the `.tt` files (checkadddevice_bug1.tt), we didn't check in the reference file (checkadddevice_but1.tt.sarif).

Fixed those problems, and the SDV functional tests now actually run, and pass.

PR #465, opened 2016-07-24
Handle dot-segments when normalizing URIs

lgolding:
In PR #458, we fixed Issue #431, using @nguerrera's code from Roslyn to always serialize valid URIs (whether they were absolute or relative, and whether or not they required percent encoding).

@rtaket noticed a bug in the implementation, namely that it erroneously stripped leading "dot-segments" ("../", "./"). This change fixes that problem. It retains leading "../" segments (but it does drop any leading "./" segments, which are useless).

Add unit test for new functionality.

### DIFF
--- a/src/Sarif/UriHelper.cs
+++ b/src/Sarif/UriHelper.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
@@ -69,6 +71,16 @@ namespace Microsoft.CodeAnalysis.Sarif
                     !path.StartsWith(@"\", StringComparison.Ordinal))
                 {
                     validUri = validUri.Substring(1);
+
+                    // When the UriBuilder constructs an absolute URI, it strips any
+                    // leading "." and ".." segments ("dot-segments", as RFC 3986 calls
+                    // them). Glue them back on so we don't lose the relative path
+                    // information.
+                    string leadingDotSegments = GetLeadingDotSegments(path);
+                    if (!string.IsNullOrEmpty(leadingDotSegments))
+                    {
+                        validUri = leadingDotSegments + validUri;
+                    }
                 }
             }
             else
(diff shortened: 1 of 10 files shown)


src/Sarif/UriHelper.cs
```diff
@@ -69,6 +71,16 @@ public static string MakeValidUri(string path)
...
+
+                    // When the UriBuilder constructs an absolute URI, it strips any
+                    // leading "." and ".." segments ("dot-segments", as RFC 3986 calls
+                    // them). Glue them back on so we don't lose the relative path
+                    // information.
+                    string leadingDotSegments = GetLeadingDotSegments(path);
+                    if (!string.IsNullOrEmpty(leadingDotSegments))
+                    {
+                        validUri = leadingDotSegments + validUri;
+                    }
```
nguerrera: Sorry that I missed the first code review on this. 

I'm not seeing the explanation anywhere for why this code path (even before this change) for the relative case has a different implementation from what I used in Roslyn. I tried to do it without explicit string manipulation. I assume you found something wrong with that even before the dot-segment issue? Please file a bug against Roslyn for anything that it does not handle correctly vs. your changes.

lgolding: Sorry, the comment above was intended to explain this code path. Without this code, the "URI" `..\..\..\foo\bar\baz.c` serializes as `foo/bar/baz.c`, losing the "dot-segments" (the relative path information). With this code, that URI serializes as `../../../foo/bar/baz.c` (and the new unit test verifies that). I'll add that example to the comment.

We only found that because SDV emits paths of that form. If Roslyn never does, there's no bug. Is it possible for Roslyn to hit this?

nguerrera: The exposure in Roslyn to the non-absolute path is mostly theoretical since we don't expose the errorlogger except as a switch to the command-line compiler, which always provides absolute paths. Handling relative paths was a defense against future use of the error logger against arbitrary SourceText.Path values (which are possible if you do the analysis via the API and not the command line).

My question wasn't about this bug but why this code path for relative case was different than what I wrote even before this change. Did you fix another bug in porting my code here originally? What was it?

I think we should sync up the helpers to be exactly the same. My head hurts trying to understand what's different and if that's OK. I will also file a bug against corefx to get a way for System.Uri to just do this for us, because getting this right with worakarounds is proving to be quite complicated!

lgolding: Yes, IIRC there was already a bug in the relative path case: that `foo\bar\baz.c` serialized as `/foo/bar/baz.c`. So I added the logic to remove the leading slash -- unless the _original_ path started with a slash.

Again, since you never encountered relative paths, you wouldn't have seen this.

I think syncing the helpers is a good idea, and I think the version I have here is good (and has a good set of tests).

nguerrera: Interesting. I'm surprised because there is a test for `Relative Additional\Location` becoming `Relative%20Adittional/Location`. I would expect that test to fail and output `/Relative%20Additional/Location`

A quick interactive test does show that I am broken for the ../ case, though.

https://github.com/dotnet/roslyn/pull/11390/commits/c66413016d28dde42a1cd6cc669b3c4d47835deb

lgolding: Well, I'm probably wrong about that second issue. I just know I reworked the logic a bit in a way I fondly believed made it a bit simpler :-)

src/Sarif/UriHelper.cs
```diff
@@ -78,5 +90,36 @@ public static string MakeValidUri(string path)
...
+                    path = path.Substring(2);
+                }
+                else if (s_twoDotPattern.IsMatch(path))
+                {
+                    path = path.Substring(3);
+                    sb.Append("../");
+                }
+                else
+                {
+                    moreDotSegments = false;
```
rtaket: I think we miss the scenario where the input path is just a dot segment (".."). 
... However, I don't think there's a valid SARIF scenario for this because ".." would indicate a path, and I think all the SARIF URIs are files?

rtaket:
:+1:

src/Sarif/UriHelper.cs
lgolding: Fixed. Added special case for "end-with-or-only two-dot-segments".

PR #467, opened 2016-07-25
Fix viewer#35: Exception when location lacks region

lgolding:
If you double-click on an issue which contains any `physicalLocation` object that lacks a `region` property -- whether that be an `analysisTarget`, a `resultFile`, or any occurrence of an `annotatedCodeLocation` -- then the viewer throws an exception.

The exception occurs when the viewer tries to mark the source file line corresponding to the `physicalLocation`, because it assumes that every `physicalLocation` has a `region`.

NOTE: This is one of what I expect to be several issues that will be revealed as I try to view the new "comprehensive" sample file in the viewer.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
@@ -285,23 +285,23 @@ namespace Microsoft.Sarif.Viewer
 
         internal void RemoveMarkers()
         {
-            LineMarker.RemoveMarker();
+            LineMarker?.RemoveMarker();
 
             foreach (AnnotatedCodeLocationModel location in this.Locations)
             {
-                location.LineMarker.RemoveMarker();
+                location.LineMarker?.RemoveMarker();
             }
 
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -399,62 +399,62 @@ internal void RemapFilePath(string originalPath, string remappedPath)
 
         internal void AttachToDocument(string documentName, long docCookie, IVsWindowFrame pFrame)
         {
-            LineMarker.AttachToDocument(documentName, (long)docCookie, pFrame);
+            LineMarker.AttachToDocument(documentName, docCookie, pFrame);
```
rtaket: Do we not need the LineMarker null check here?
Same question for line 433.

rtaket:
:+1:  ... except for the one question/comment about the null check.

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
lgolding: You're right. Good catch. I misread that as a static method call. Will fix, then merge.

PR #468, opened 2016-07-25
Fix viewer#34: Stack frames don't show line or column numbers

lgolding:
The property path was wrong for the line number (was `Region.StartLine`, changed to `Line`), and the visibility converters were wrong for both the line number and the column number (were `StringToVisibilityConverter`, changed to `Int32ToVisibilityConverter`).

Before:

image

After:

image


rtaket:
:+1:

PR #473, opened 2016-07-26
Update SDV converter

michaelcfanning:
This converter change brings the built-in conversion very close to direct production. @rtaket @lgolding

### DIFF
--- a/src/Sarif.Converters/StaticDriverVerifierConverter.cs
+++ b/src/Sarif.Converters/StaticDriverVerifierConverter.cs
@@ -6,17 +6,20 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
     internal class StaticDriverVerifierConverter : ToolFileConverterBase
     {
         private StringBuilder _sb;
+        private Stack<string> _callers;
 
         /// <summary>Initializes a new instance of the <see cref="StaticDriverVerifierConverter"/> class.</summary>
         public StaticDriverVerifierConverter()
         {
             _sb = new StringBuilder();
+            _callers = new Stack<string>();
         }
 
         /// <summary>
@@ -111,23 +114,22 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             if (int.TryParse(tokens[STEP], out step))
             {
-                bool displayed = true;
-
                 // If we find a numeric value as the first token,
                 // this is a general step. We don't actually consume
                 // the step, as it is a 0-indexed value and SARIF
                 // specifies 1-indexed.
 
+                Uri uri = null;
                 string uriText = tokens[URI].Trim('"');
 
-                if (uriText.Equals("?", StringComparison.Ordinal))
+                if (!uriText.Equals("?", StringComparison.Ordinal))
                 {
-                    // If we have not literal location, then we are processing
-                    // an informational step, which we will not persist
-                    displayed = false;
+                    if (File.Exists(uriText))
+                    {
+                        uriText = Path.GetFullPath(uriText);
+                    }
+                    uri = new Uri(uriText, UriKind.RelativeOrAbsolute);
                 }
-
-                var uri = new Uri(tokens[URI].Trim('"'), UriKind.RelativeOrAbsolute);
 
                 // We assume a valid line here. This code will throw if not.
                 int line = int.Parse(tokens[LINE]);
@@ -144,15 +146,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
                 AnnotatedCodeLocationKind kind = ConvertToAnnotatedCodeLocationKind(sdvKind.Trim());
 
-                string message = BuildMessageFromState(tokens[STATE]);
-
                 var annotatedCodeLocation = new AnnotatedCodeLocation
                 {
                     Kind = kind,
                     Step = step,
-                    Importance = displayed ? AnnotatedCodeLocationImportance.Important : AnnotatedCodeLocationImportance.Unimportant,
-                    Message = message,
-                    PhysicalLocation = displayed ? new PhysicalLocation
+                    Importance = AnnotatedCodeLocationImportance.Unimportant,
+                    PhysicalLocation = (uri != null) ? new PhysicalLocation
                     {
                         Uri = uri,
                         Region = new Region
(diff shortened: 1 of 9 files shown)


src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/checkadddevice_bug1.tt.sarif
```diff
@@ -32,154 +32,105 @@
...
-                  "importance": "unimportant",
-                  "properties": {
-                    "State": "^====Auto====="
-                  }
+                  "importance": "unimportant"
                 },
                 {
                   "step": 6,
-                  "message": "====Auto=====",
+                  "fullyQualifiedLogicalName": "main",
```
lgolding: Is this right? In Step 0, `OS` called `main`. In Step 2, `main` called `_sdv_init4`. Now in Step 6 we are returning from `_sdv_init4` to `main`, so `fullyQualifiedLogicalName` -- which refers to the code that's currently executing -- should be `_sdv_init4`.

Now we know we're going to add some more properties to `annotatedCallLocation`, such as a property that specifies the target of the return when the `kind` is `callReturn` (see Issue sarif-standard/sarif-spec#249, which I filed this morning). That's the place that "`main`" would go.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -228,14 +265,25 @@ private void ProcessLine(string logFileLine, Result result)
             }
         }
 
-        private string BuildMessageFromState(string state)
+        private bool IsHarnessOrRulesFiles(string fileName)
         {
-            if (string.IsNullOrEmpty(state) || state == "^")
+            return fileName.EndsWith(".slic") || fileName.EndsWith("sdv-harness.c");
```
rtaket: Do these need to be case insensitive? Esp the file extension since that could be user code (for a custom rule.)

rtaket:
:+1:

src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/checkadddevice_bug1.tt.sarif
michaelcfanning: Good question. The callReturn currently returns the fullyQualifiedName of the caller. i.e., where we are returning control to. You could argue that this value should contain the callee name, do you think this makes the most sense?

call instance, FQN == caller, Callee property == callee
call return for the above, FQN == callee? or caller?

src/Sarif.Converters/StaticDriverVerifierConverter.cs
michaelcfanning: I use the internal SDV logic for this. I will change it in both places (SDV and converter) in my next change.

src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/checkadddevice_bug1.tt.sarif
lgolding: Here's what the spec says:

> An `annotatedCodeLocation` object may contain a property named `fullyQualifiedLogicalName` whose value is a string containing the fully qualified name of the method to which this `annotatedCodeLocation` refers.

Now "refers" isn't very specific. I can tell you what I had in mind when I wrote those words (which might not be what you had in mind). My intent was that FQLN should be the name of the method _in which this line of code occurs_.

So:

```C
void a()
{
    b();   // Line 3
}

void b()
{
    return;   // Line 8
}
```

**Step 0**
* physLoc.region.startLine = 3
* FQLN = a
* kind = call
* callee = b

**Step 1**
* physLoc.region.startLine = 8
* FQLN = b
* kind = callReturn
* returnTarget = a  // Assuming such a thing existed

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -6,17 +6,20 @@
...
+        private Stack<string> _callers;
 
         /// <summary>Initializes a new instance of the <see cref="StaticDriverVerifierConverter"/> class.</summary>
         public StaticDriverVerifierConverter()
         {
             _sb = new StringBuilder();
+            _callers = new Stack<string>();
         }
 
         /// <summary>
```
lgolding: Also I read over the `ProcessSdvDefectStream` method again. Instead of reading bytes from a stream until EOF, and looking for newlines, you could make a `StreamReader` out of the `input` stream, and then just call `ReadLine` on it.

lgolding: Looking back over `ProcessLine` now, I see the comment "SARIF specified 1-indexed [step]". But the spec says:

> The value of the `step` property shall be an integer whose value is the 0-based sequence number of the location within the code flow, that is, it shall be 0 for the first location, 1 for the second, and so on.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -111,24 +114,23 @@ private void ProcessLine(string logFileLine, Result result)
 
             if (int.TryParse(tokens[STEP], out step))
             {
-                bool displayed = true;
-
                 // If we find a numeric value as the first token,
                 // this is a general step. We don't actually consume
                 // the step, as it is a 0-indexed value and SARIF
                 // specifies 1-indexed.
```
lgolding: The spec says:

> The value of the `step` property shall be an integer whose value is the 0-based sequence number of the location within the code flow, that is, it shall be 0 for the first location, 1 for the second, and so on.

src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/checkadddevice_bug1.tt.sarif
michaelcfanning: We both interpret the meaning of FQLN the same way. But I consider the origin and the return of the call being in the caller code. After all, why would they be inconsistent (why is the call FQLN the caller but the return is the callee?).  We create a holder node at the call site, it makes sense for the return to be matched at the same tree level.

One thing you might be thinking of is that we don't have an explicit mechanism for identifying where a function exits. We previously had functionEnter and functionReturn for these matched pairs (and not I think I see the argument for returning them). One thing about both call and callReturn and functionEnter and functionExit is that it's not 100% clear whether we should mandate they are provided in matched pairs.

michaelcfanning: ok, happily, I was not correct, functionEnter and functionExit survived. So, call and callReturn refer to the location of a call site. functionEnter and functionExit are the matched pair of locations within a callee. FQLN should be identical for each pair.

btw - the spec does not make it clear whether we require these to be emitted as matched pairs or not.

lgolding: Re: FQLN for `callReturn`: OK, I agree with your argument that since we also have `functionEnter` and `functionExit`, then `functionExit` should refer to the actual `return` statement in the callee. That in turn implies that `callReturn` refers to a location in the caller.

So there are two questions:

1. In a .tt file, does a "Return" line (for example, `269 "?" 0 true ^====Auto=====^s=1 Return`) really mean a `return` statement in the code (in which case, SARIF should represent it with `kind = functionExit`), or does it really mean the next location in the caller (in which case SARIF should represent it with `kind = callReturn`)? I will ask Jakob.

2. When the viewer builds the tree, it _indents_ the `callReturn` aCLs under the caller. By your argument, the `callReturn` should pop back to the same indentation level as the caller, right?

Re: Call/callReturn pairs: the spec does not require them to occur in matched pairs. Since a codeFlow is a possible path through the code (generally used to specify a code path leading to the discovery of a problem), it's perfectly fine for the code path to be:

- a calls b
- b calls c
- c calls d
- BOOM

PR #475, opened 2016-07-26
SDV converter PR feedback.

michaelcfanning:
@lgolding @rtaket

### DIFF
--- a/src/Sarif.Converters/StaticDriverVerifierConverter.cs
+++ b/src/Sarif.Converters/StaticDriverVerifierConverter.cs
@@ -78,7 +78,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             using (var reader = new StreamReader(input))
             {
-                ProcessLine(reader.ReadLine(), result);
+                string line;
+
+                while (!string.IsNullOrEmpty(line = reader.ReadLine()))
+                {
+                    ProcessLine(line, result);
+                }
             }
 
             return result;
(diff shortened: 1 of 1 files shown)


src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -66,24 +67,18 @@ private Result ProcessSdvDefectStream(Stream input)
...
 
             result.Locations = new List<Location>();
-
-            result.CodeFlows = new List<CodeFlow>();        
-            result.CodeFlows.Add(new CodeFlow
-            {
-                Locations = new List<AnnotatedCodeLocation>()
-            });
-
-            var sb = new StringBuilder();
```
lgolding: Since you're no longer using `StringBuilder`, you probably don't need the `using` directive for `System.Text`.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -66,24 +67,18 @@ private Result ProcessSdvDefectStream(Stream input)
...
-                if (current == '\n')
+                new CodeFlow
                 {
-                    ProcessLine(sb.ToString(), result);
-                    sb.Clear();
+                    Locations = new List<AnnotatedCodeLocation>()
                 }
+            };
+
+            using (var reader = new StreamReader(input))
```
lgolding: >reader [](start = 23, length = 6)

You need a loop over lines.

lgolding:
:clock1:

PR #478, opened 2016-07-27
VSIX: Restore UI focus after closing the FileOpenDialog

rtaket:
Displaying the FileOpenDialog caused the TreeView to lose focus. A downstream effect was that any tree node that needed file rebaselining would lose focus/selection after the FileOpenDialog was dismissed. The fix was to restore focus to the appropriate UI element after the FileOpenDialog is dismissed.


michaelcfanning:
+1

PR #479, opened 2016-07-28
Create non-beta SDK driver package.

michaelcfanning:
@rtaket @lgolding 

We need a non-beta driver package. This requires us to consume CommandLineParser as a checked in binary (because we can't include a beta package dependency in our nuspec).

All this is in support of creating a non-beta BinSkim . @tikiwanMS, fyi, that will be done tomorrow.


lgolding:
:+1:

lgolding:
Consider adding a README to the `references` directory explaining why we have FCIBs.

src/Nuget/Sarif.Driver.nuspec
```diff
@@ -15,21 +15,18 @@
...
     </dependencies>
   </metadata>
   <files>
     <file src="..\..\..\..\src\ReleaseHistory.md" target="lib\ReleaseHistory.md" />
     <file src="..\..\..\..\src\Sarif\Schemata\Sarif.schema.json" target="lib\Sarif.schema.json" />
-    
-    <file src='Sarif.Driver.dll' target='lib\net40\Sarif.Driver.dll'/>
-    <file src='Sarif.Driver.pdb' target='lib\net40\Sarif.Driver.pdb'/>
 
+    <file src='Sarif.Driver.dll' target='lib\net45\CommandLine.dll'/>
```
rtaket: Should this src attribute be src='CommandLine.dll' ?

src/Sarif.Driver/Sarif.Driver.csproj
```diff
@@ -28,8 +28,8 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
   <ItemGroup>
     <Reference Include="CommandLine, Version=2.0.275.0, Culture=neutral, PublicKeyToken=de6f01bd326f8c32, processorArchitecture=MSIL">
-      <HintPath>..\packages\CommandLineParser.2.0.275-beta\lib\net45\CommandLine.dll</HintPath>
-      <Private>True</Private>
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\references\CommandLine.dll</HintPath>
```
rtaket: I'm just curious why the reference in the csproj file had to change. Can't the csproj file and the nuspec both include the CommandLine.dll from the package?

src/Nuget/Sarif.Driver.nuspec
rtaket: The target directory should probably be the same as the csproj reference (so we test and deploy the same binaries.) So, I think this should either point to the references directory or the csproj should continue to use the package.

michaelcfanning:
will act on all this feedback, thanks

michaelcfanning:
@rtaket does the nuspec look right to you now?

rtaket:
:+1:  ... except for the issue discussed in IM.

PR #494, opened 2016-08-16
VSIX: Add verbosity filter to the Code Flow tab

rtaket:
- Add verbosity filter to the Code Flow tab
- Add expand and collapse buttons to Code Flow tab


rtaket:
@michaelcfanning

michaelcfanning:
+1

PR #496, opened 2016-08-17
Provide optional storage for contents of file in log.

michaelcfanning:
@lgolding @rtaket 

Provide for completely encapsulated SARIF logs (i.e., results plus encoded contents).


lgolding:
:+1: 

I filed sarif-standard/sarif-spec#260, "Add `file.contents` property."

PR #499, opened 2016-08-21
Fix #497: Modify PREfast converter per recent aCL changes

### LINKED ISSUE #497: Modify PREfast converter per recent annotatedCodeLocation changes
This work was partially done in June (commits f200c04e and ca4144cb). But

1. The converter didn't actually emit the `step` property for `SFA`s without a `KEYEVENT`.
2. It didn't emit the `keyEventId` property into the property bag the way PREfast does during direct production.
3. It used a 0-based `step` property.
4. It didn't set `"importance":"unimportant"` for `SFA`s without a `KEYEVENT`.
5. It didn't check to see if the checker actually used `KEYEVENT`s.

lgolding:
This work was partially done in June (commits f200c04e and ca4144cb). But:

1. The converter didn't actually emit the `step` property for `SFA`s without a `KEYEVENT`.
2. It didn't emit the `keyEventId` property into the property bag the way PREfast does during direct production.
3. It used a 0-based `step` property.
4. It didn't set `"importance":"unimportant"` for `SFA`s without a `KEYEVENT`.
5. It didn't check to see if the checker actually used `KEYEVENT`s.

Note: The implementation differs from the direct production implementation in PREfast in that the converter doesn't bother to emit the default value `"importance": "important"` in those cases where it would be appropriate. We should talk about whether to change PREfast to do the same.

### DIFF
--- a/src/PREfastXmlSarifConverter/PREfastXmlSarifConverter.cpp
+++ b/src/PREfastXmlSarifConverter/PREfastXmlSarifConverter.cpp
@@ -408,6 +408,26 @@ extern "C"
     }
 }
 
+namespace
+{
+    // Return a value indicating whether the specified code path uses key events.
+    //
+    // None of the checkers built on EspXtension or the old Esp stack support key
+    // events. This includes important ones like the buffer overrun checker and
+    // the concurrency checker.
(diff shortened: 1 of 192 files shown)


michaelcfanning:
:shipit:

PR #501, opened 2016-08-23
Implement smart validation in SarifCli tool

lgolding:
We implement the framework for the validator and its functional tests, and we implement the following rules:

* `AnnotatedCodeLocationEssentialIsObsolete`
* `AnnotatedCodeLocationIdIsObsolete`
* `DoNotUseFriendlyNameAsRuleId`
* `EndColumnMustNotBeLessThanStartColumn`
* `EndLineMustNotBeLessThanStartLine`
* `EndTimeMustBeAfterStartTime`
* `HashAlgorithmsMustBeUnique`
* `ImportanceMustAppearOnlyInCodeFlowLocations`
* `MessagesShouldEndWithPeriod`
* `StepMustAppearOnlyInCodeFlowLocations`
* `StepValuesMustFormOneBasedSequence`
* `UrisMustBeValid`
* `UseAbsolutePathsForNestedUriFragments`

Subsequent PRs will implement more rules.

The validator is built on the `Sarif.Driver` framework. The validator performs schema validation in addition to the "smart validation" rules.

The smart validation rules get their location information as follows:

1. Construct a JSON pointer (such as `"/runs/0/rules/SV0001"`) to represent the logical location of the result.
2. Evaluate that pointer against a `JToken` that represents the entire document, producing a `JToken` that represents the element specified by the pointer.
3. Extract the location information from the `JToken`.

We use the JSON Pointer implementation from `Microsoft.Json.Pointer`, which is obtained from NuGet along with `Microsoft.Json.Schema`. This is the first use of the JSON Pointer implementation.

Although a skeleton project for the validator already existed, you should read this as completely new code.

NOTE 1: I ran into the circular dependency thing we've seen before, where a project in sarif-sdk (in this case, the new `SarifCli` and `SarifCli.FunctionalTest` projects) depends on JSchema, which in turn depends on a different version of sarif-sdk. The comments on commit bf4b015 explain how I fixed it.

NOTE 2: I also took the opportunity to fix our dependency on the foreign checked in binary Microsoft.Json.Schema.dll in the `src\References` directory. Commits 046c98f, d6de920, and c1e2261 removed the FCIB dependencies, replaced them with  NuGet dependencies on Microsoft.Json.Schema and Microsoft.Json.Pointer, and removed the `src\References` directory.

NOTE 3: I took the opportunity to upgrade our dependency on Microsoft.Json.Schema.ToDotNet from version 0.42.0 to the latest 0.45.0 (commit 00a6e30). As a result, all the files in the `Autogenerated` directory were modified (the version specified in their `GeneratedCode` attributes changed from `0.42.0` to `0.45.0`.


src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdDoesNotMatchFriendlyName.sarif
```diff
@@ -0,0 +1,25 @@
+{
```
lgolding: This test file does _not_ violate the rule.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdDoesNotMatchFriendlyName_Expected.sarif
```diff
@@ -0,0 +1,47 @@
+{
```
lgolding: The validator produces this output file for the log file that does _not_ violate the rule. There are no `result`s.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdDoesNotMatchFriendlyName_Expected.sarif
```diff
@@ -0,0 +1,47 @@
...
+        "semanticVersion": "1.5.28",
+        "sarifLoggerVersion": "1.5.28.0",
+        "language": "en-US",
+        "properties": {
+          "Comments": "Object model and utilities for reading and writing files in the SARIF format.",
+          "CompanyName": "Microsoft Corporation",
+          "ProductName": "Microsoft SARIF SDK"
+        }
+      },
+      "results": [],
```
lgolding: No results.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName.sarif
```diff
@@ -0,0 +1,25 @@
+{
```
lgolding: This test file _does_ violate the rule.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName_Expected.sarif
```diff
@@ -0,0 +1,83 @@
+{
```
lgolding: The validator produces this output file for the log file that _does_ violate the rule. There is one `result`.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName_Expected.sarif
```diff
@@ -0,0 +1,83 @@
...
+              "ImportantRule"
+            ]
+          },
+          "locations": [
+            {
+              "analysisTarget": {
+                "uri": "file:///G:/Code/sarif-sdk/bld/bin/SarifCli.FunctionalTests/AnyCPU_Release/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName.sarif",
+                "region": {
+                  "startLine": 13,
+                  "startColumn": 27
```
lgolding: The `result` has the correct location.

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName_Expected.sarif
```diff
@@ -0,0 +1,83 @@
...
+        }
+      ],
+      "rules": {
+        "SV0001": {
+          "id": "SV0001",
+          "name": "DoNotUseFriendlyNameAsRuleId",
+          "shortDescription": "Do not use the same string for a rule's id and name properties. The id property ...",
+          "fullDescription": "Do not use the same string for a rule's id and name properties. The id property must be a stable, opaque identifer such as \"SV0001\". The name property should be a string that is understandable to an end user, such as \"DoNotUserFriendlyNameAsRuleId\".",
+          "messageFormats": {
+            "DefaultFormatId": "In \"{0}\", the name and id properties of rule \"{1}\" are the same."
```
lgolding: This is IMO an unfortunate feature of the framework: you must leave a placeholder for the target file name, even though it's mentioned everywhere you might look at the result: in the Error List, in the `result` element of the raw log file, and on the console, resulting in this repetition:

> `G:\Code\sarif-sdk\src\SarifCli.FunctionalTests\TestData\DoNotUseFriendlyNameAsRuleId\RuleIdMatchesFriendlyName.sarif(13,27): warning SV0001: In "RuleIdMatchesFriendlyName.sarif", the name and id properties of rule "ImportantRule" are the same.`

This makes sense in BinSkim, where the message all look like "foo.exe is a 64-bit imange with...". But for other tools I don't think it makes sense. This seems to be a bit of tail-wagging-dog, where a convenience feature for BinSkim influenced the framework.

I think BinSkim should just specify {0} manually, and the framework shouldn't provide it.

Or at the very least, if the framework insists on providing the target file name as `{0}`, it shouldn't have an assertion that insists that every parameter be mentioned in the message.

appveyor.yml
```diff
@@ -31,6 +31,7 @@ test:
     - '**\Sarif.FunctionalTests.dll'
     - '**\Sarif.ValidationTests.dll'
     - '**\Sarif.Viewer.VisualStudio.UnitTests.dll'
+    - '**\Microsoft.CodeAnalysis.SarifCli.FunctionalTests.dll'
```
michaelcfanning: >Microsoft.CodeAnalysis [](start = 10, length = 22)

Why is this naming convention different? Should be SarifCli.FunctionalTests.dll.

src/Sarif/Writers/MimeType.cs
```diff
@@ -73,6 +73,8 @@ public static string DetermineFromFileExtension(Uri fileUri)
         public static readonly string Directory = "application/x-directory";
         /// <summary>The MIME type used for CSharp files.</summary>
         public static readonly string CSharp = "text/x-csharp";
+        /// <summary>The MIME type for SARIF files.</summary>
+        public static readonly string Sarif = "text/x-sarif";
```
michaelcfanning: >text/x-sarif [](start = 47, length = 12)

so we wouldn't just use 'application/json' as our mime type?

src/SarifCli/AnalyzeCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            catch (JsonSyntaxException ex)
+            {
+                ReportResult(ex.Result, logger);
+            }
+            catch (SchemaValidationException ex)
+            {
+                ReportInvalidSchemaErrors(ex, schemaFilePath, logger);
+            }
+            catch (Exception ex)
```
michaelcfanning: >Exception ex) [](start = 19, length = 13)

do you need this? doesn't the framework catch unhandled exceptions?

src/SarifCli/AnalyzeCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        private static void ReportResult(
+            Result result,
+            IAnalysisLogger logger)
+        {
+            Rule rule = Json.Schema.Sarif.RuleFactory.GetRuleFromRuleId(result.RuleId);
+            logger.Log(rule, result);
+        }
+
+        private static void LogToolNotification(
```
michaelcfanning: hm, this stuff looks generally useful. should it be in the core framework?

src/SarifCli/AnalyzeOptions.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Cli
+{
+    [Verb("validate", HelpText = "Validate a SARIF file against the schema and against additional correctness rules.")]
```
michaelcfanning: >validate [](start = 11, length = 8)

you only have one arg, a path to a schema, why use a named arg? suggest a positional arg instead. also, your named args in general look exceeding verbose. this one could simply be --schema for example. imo

src/SarifCli/JPointerStringExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Cli
+{
+    // TODO: These methods belong in the Microsoft.Json.Pointer library.
```
michaelcfanning: >TODO [](start = 7, length = 4)

open an issue and reference its path in this comment?

src/SarifCli/AnalyzeCommand.cs
lgolding: Yes. Actually I copied it from BinSkim (which is the model I used for this whole framework-based tool), and seeing we've used it in two places, yes, we should move it into the framework. I'd do that in a separate PR, though.

---
In reply to: [77238560](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77238560) [](ancestors = 77238560)

src/SarifCli/SarifValidationContext.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                return Path.GetExtension(TargetUri.LocalPath).Equals(".sarif", StringComparison.OrdinalIgnoreCase);
+            }
+        }
+
+        public IAnalysisLogger Logger { get; set; }
+
+        public string MimeType
+        {
+            get { return "text/x-sarif"; }
```
michaelcfanning: >"text/x-sarif" [](start = 25, length = 14)

DRY this out by referencing a constant elsewhere (such as used by mime detection utility)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Cli.Rules
+{
+    public class DoNotUseFriendlyNameAsRuleId : SarifValidationSkimmerBase
+    {
+        public override string FullDescription => RuleResources.SV0001_DoNotUseFriendlyNameAsRuleIdDescription;
+
+        /// <summary>
+        /// SV0001
```
michaelcfanning: >SV0001 [](start = 12, length = 6)

What about SARIFXXX as a convention? Has SARIF in the name and allows for 1000 validations (surely we won't need more). Just asking

src/SarifCli/JPointerStringExtensions.cs
lgolding: Done. Filed [https://github.com/Microsoft/sarif-sdk/issues/512](https://github.com/Microsoft/sarif-sdk/issues/512), "Move JPointerExtensions from SarifCli into Microsoft.Json.Pointer library", and added comment.

---
In reply to: [77238851](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77238851) [](ancestors = 77238851)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    nameof(RuleResources.SV0001_DefaultFormatId)
+                };
+            }
+        }
+
+        protected override void Analyze(Rule rule, string rulePointer)
+        {
+            if (rule.Id != null
+                && rule.Name != null
```
michaelcfanning: >&& [](start = 16, length = 2)

If you leave these operators on the preceding line, your conditional clauses will line up nicely

if (rule.Id != null && 
     rule.Id ....

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+
+        protected override void Analyze(Rule rule, string rulePointer)
+        {
+            if (rule.Id != null
+                && rule.Name != null 
+                && rule.Id.Equals(rule.Name, StringComparison.OrdinalIgnoreCase))
+            {
+                LogResult(
```
michaelcfanning: >LogResult [](start = 16, length = 9)

Up to you. But I always add a source comment when firing messages that shows the complete string that's emitted. Very useful for understanding rule function. See binskim for examples of this discipline.

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        private void Visit(CodeFlow codeFlow, string codeFlowPointer)
+        {
+            if (codeFlow.Locations != null)
+            {
+                AnnotatedCodeLocation[] annotatedCodeLocations = codeFlow.Locations.ToArray();
+                string annotatedCodeLocationsPointer = codeFlowPointer.AtProperty(SarifPropertyName.Locations);
+
+                for (int iAnnotatedCodeLocation = 0; iAnnotatedCodeLocation < annotatedCodeLocations.Length; ++iAnnotatedCodeLocation)
```
michaelcfanning: >iAnnotatedCodeLocation [](start = 25, length = 22)

what's wrong with 'index' or simply 'i' here? :) the verbosity in this case doesn't seem helpful

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        protected virtual void Analyze(StackFrame frame, string framePointer)
+        {
+        }
+
+        protected void LogResult(ResultLevel level, string jPointer, string formatId, params string[] args)
+        {
+            Region region = GetRegionFromJPointer(jPointer);
+
+            // All messages start with "In {file}, at {jPointer}, ...". Prepend the jPointer to the args.
```
michaelcfanning: >All messages start with "In {file}, at {jPointer}, [](start = 15, length = 50)

Suggest we don't do this. Instead, use the VS formatting helpers to build the standard message format. When redirected to the output window in VS, your messages, on double-click, should jump to the relevant location. Try integrating ModernCop in VS as an external tool for an example.

src/SarifCli/Rules/UseAbsolutePathsForNestedFileUriFragments.cs
```diff
@@ -0,0 +1,113 @@
...
+                if (UriHasNonAbsoluteFragment(fileUri))
+                {
+                    LogResult(
+                        ResultLevel.Warning,
+                        filePointer,
+                        nameof(RuleResources.SV0002_DefaultFormatId),
+                        fileUri.OriginalString);
+                }
+            }
+            catch
```
michaelcfanning: >catch [](start = 12, length = 5)

please get rid of this general catch handler. i.e., discover specific exceptions for malformed URIs and let all other exceptions through

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
lgolding: Originally, I didn't have the visiting code so nicely factored, so I had multiple loops (say, a loop over results, with a nested loop over locations). So I used the verbose index names for (what I fondly believed to be) clarity. I'll shorten them.

---
In reply to: [77240006](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77240006) [](ancestors = 77240006)

src/SarifCli.FunctionalTests/TestData/UseAbsolutePathsForNestedFileUriFragments/NestedFileUriFragmentIsRelativeInCodeFlowLocationUri_Expected.sarif
```diff
@@ -0,0 +1,84 @@
...
+        }
+      ],
+      "rules": {
+        "SV0002": {
+          "id": "SV0002",
+          "name": "UseAbsolutePathsForNestedFileUriFragments",
+          "shortDescription": "When specifying the URI of a nested file with a URI fragment, express the fragment as an absolute path.",
+          "fullDescription": "When specifying the URI of a nested file with a URI fragment, express the fragment as an absolute path. That is, begin the with a forward slash, for example \"file:///C:/bin/MyApp.zip#/images/grape.jpg\".",
+          "messageFormats": {
+            "DefaultFormatId": "In \"{0}\", at \"{1}\", the fragment portion of the URI \"{2}\", which specifies the location of a nested file with respect to its parent, does not begin with a forward slash."
```
michaelcfanning: >In \"{0}\", at \"{1}\", [](start = 32, length = 23)

see other note. console logger should already emit file/line locations, don't recap here.

michaelcfanning:
:shipit:

lgolding:
Per our discussion, I'll merge this and address the feedback in a separate PR>

src/SarifCli.FunctionalTests/TestData/UseAbsolutePathsForNestedFileUriFragments/NestedFileUriFragmentIsRelativeInCodeFlowLocationUri_Expected.sarif
lgolding: That's actually not what's happening. There are several things to say here:

1. The code actually _is_ using the Visual Studio formatting helper. It calls into the framework's `ConsoleLogger`, which ultimately calls `FormatForVisualStudio`. So the results produced by the validator _are_ in the correct format, and you _can_ navigate to them.

2. The "{0}" here is indeed the file name, repeated. It's there because, as the framework stands today, it _always_ prepends the file name to the list of message arguments. I complained about this in another comment on this code review (please take a look at that comment). This is really a bug in the framework, and I'm going to file a bug and fix it. BinSkim (which relies on this "feature") will need to be updated so that if it wants the file name in the text of every message -- which apparently it does -- it puts it there itself.

3. The "{1}" here is not the line/column location. It is the JSON pointer location, which specifies the logical path through the JSON to the error. The message would read, for example, "In 'foo.sarif', at '/runs/0/results/0/locations/0/analysisTarget/uri', the fragment portion of the URI 'bar#baz'... blah blah". Once I fix the bug in the framework, it will just read "At /runs/0... ".

---
In reply to: [77240794](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77240794) [](ancestors = 77240794)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
lgolding: I did consider SARIF, but I wasn't aware of a precedent for a 5-character prefix. Sure, I'll change it.

---
In reply to: [77239380](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77239380) [](ancestors = 77239380)

appveyor.yml
lgolding: Fixed.

---
In reply to: [76704201](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r76704201) [](ancestors = 76704201)

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
lgolding: Fixed.

---
In reply to: [77240771](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77240771) [](ancestors = 77240771,77240006)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
lgolding: Fixed.

---
In reply to: [77239564](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77239564) [](ancestors = 77239564)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
lgolding: Fixed.

---
In reply to: [77245086](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77245086) [](ancestors = 77245086,77239380)

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
lgolding: I'm going to fix the framework so it doesn't automatically prepend the file name. I'll upgrade to the new framework, and modify my messages so they all start with "{jPointer}: ...". Then, since I _do_ in fact use the framework's logging facilities, which _do_ use the VS formatting helper, the messages will end up looking like this:

C:\Path\To\log.sarif(12,25): /runs/0/results/0/locations/0/analysisTarget/uri: Something is dreadfully wrong with the uri "foobles".

This will continue to work with the VS error list window (as it does today).

---
In reply to: [77240384](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77240384) [](ancestors = 77240384)

src/SarifCli/JPointerStringExtensions.cs
lgolding: Fixed.

---
In reply to: [77239499](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77239499) [](ancestors = 77239499,77238851)

src/SarifCli/AnalyzeCommand.cs
lgolding: Fixed.

---
In reply to: [77238913](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77238913) [](ancestors = 77238913,77238560)

src/SarifCli.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/RuleIdMatchesFriendlyName_Expected.sarif
lgolding: Fixed the SDK to not automatically prepend the target path, upgraded SarifCli to the latest SDK, and adjusted the SarifCli resource strings accordingly.

---
In reply to: [75972569](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r75972569) [](ancestors = 75972569)

src/SarifCli/Rules/SarifValidationSkimmerBase.cs
lgolding: Fixed the SDK to not automatically prepend the target path, upgraded SarifCli to the latest SDK, and adjusted the SarifCli resource strings accordingly.

---
In reply to: [77263861](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77263861) [](ancestors = 77263861,77240384)

src/Sarif/Writers/MimeType.cs
lgolding: "application" is definitely right. I'd like us to officially register SARIF with IANA, probably as "application/sarif+json" (similar to the dozens of existing MIME types "application/foo+xml"). I'll change it to "application/json" for now.

---
In reply to: [76704516](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r76704516) [](ancestors = 76704516)

src/SarifCli/AnalyzeCommand.cs
lgolding: You're right. Fixed. Removed the handler.

---
In reply to: [77238313](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77238313) [](ancestors = 77238313)

src/Sarif/Writers/MimeType.cs
lgolding: Fixed: Set MIME type to "application/sarif+json".

---
In reply to: [77440108](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77440108) [](ancestors = 77440108,76704516)

src/SarifCli/SarifValidationContext.cs
lgolding: We can't yet DRY out SarifCli's usage of this literal string -- that is, we can't have SarifCli refer to the new value of `MimeType.Sarif` -- until the next time we upgrade SarifCli to consume the latest version of the SDK. Remember, SarifCli consumes the SDK from NuGet, not as a P2P dependency.

---
In reply to: [77239179](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77239179) [](ancestors = 77239179)

src/SarifCli/Rules/DoNotUseFriendlyNameAsRuleId.cs
lgolding: Not wild about this because you have to remember to update the comment if you modify the resource string. Won't fix.

---
In reply to: [77239668](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77239668) [](ancestors = 77239668)

src/SarifCli/Rules/UseAbsolutePathsForNestedFileUriFragments.cs
lgolding: Fixed.

---
In reply to: [77240531](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77240531) [](ancestors = 77240531)

src/SarifCli/AnalyzeOptions.cs
lgolding: It's not the case that there's only one arg. Remember that this derives from `AnalyzeOptionsBase`, from which I get (among others) `--output`. Most important, `AnalyzeOptionsBase` gives me the unnamed argument list `TargetFileSpecifiers` -- the input files -- so I can't use an unnamed arg for the schema.

I will, however, change the long-form argument name to `json-schema`. I won't make it just `schema` because that wouldn't match the short form `j` (and I can't use `s`  for the short form because it's already used for `statistics`).

---
In reply to: [77238716](https://github.com/Microsoft/sarif-sdk/pull/501#discussion_r77238716) [](ancestors = 77238716)

PR #508, opened 2016-08-29
Fix #507: Add IAnalyzeOptions .NoConsoleOutput

### LINKED ISSUE #507: Add IAnalyzeOptions .NoConsoleOutput
Add a Boolean property `NoConsoleOutput` to `IAnalyzeOptions` and `AnalyzeOptionsBase`. When set, it will prevent the `AggregatingLogger` from including a `ConsoleLogger` (in the same way -- but with the opposite logical sense -- that the `Statistics` property causes the `AggregatingLogger` to include a `StatisticsLogger`.

The reason for doing this is that functional tests can set this property on the options object, so the console output from the tests will be clean instead of including the (expected) error output.

lgolding:
Add a Boolean property `NoConsoleOutput` to `IAnalyzeOptions` and `AnalyzeOptionsBase`. When set, it will prevent the `AggregatingLogger` from including a `ConsoleLogger` (in the same way -- but with the opposite logical sense -- that the `Statistics` property causes the `AggregatingLogger` to include a `StatisticsLogger`).

The reason for doing this is that functional tests can set this property on the options object, so the console output from the tests will be clean instead of including the (expected) error output.

With this change, the output from the functional tests looks like this:

```
xUnit.net Console Runner (32-bit .NET 4.0.30319.42000)
  Discovering: Sarif.Driver.UnitTests
  Discovered:  Sarif.Driver.UnitTests
  Starting:    Sarif.Driver.UnitTests

# valid targets: 1
# invalid targets: 0
Time elapsed: 00:00:00.0258590
  Finished:    Sarif.Driver.UnitTests
```

(The remaining console output is from the statistics logger.)

NOTE: My motivation for doing this is that the SarifCli.FunctionalTests were also producing console output which I didn't want to see. If you really want to keep seeing the console output from the Sarif.Driver unit tests, I can revert the changes to AnalyzeCommandTests, but keep the rest of the changes. That will let me suppress the console output from my SarifCli.FunctionalTests. But I do think that that test output should be clean.

### DIFF
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -39,10 +39,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         public string ConfigurationFilePath { get; set; }
 
         [Option(
-            'n',
-            "no-console-output",
+            'q',
+            "quiet",
             HelpText = "Do not log results to the console.")]
-        public bool NoConsoleOutput { get; set; }
+        public bool Quiet { get; set; }
 
         [Option(
             's',
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -39,6 +39,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         public string ConfigurationFilePath { get; set; }
 
         [Option(
+            'n',
+            "no-console-output",
+            HelpText = "Do not log results to the console.")]
+        public bool NoConsoleOutput { get; set; }
+
+        [Option(
             's',
             "statistics",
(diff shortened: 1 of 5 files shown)


src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -39,6 +39,12 @@ public class AnalyzeOptionsBase : IAnalyzeOptions
         public string ConfigurationFilePath { get; set; }
 
         [Option(
+            'n',
```
michaelcfanning: Is 'q' a better choice here (for 'quiet')?

michaelcfanning:
+1 Suggested using 'q' as a shorthand for no console output (i.e., it is a 'quiet' mode). Take it or leave it, though. :)

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
lgolding: Yes, much better.

PR #518, opened 2016-09-01
Add AnalyzeCommandBase.LogToolNotification

lgolding:
This was originally in SarifCli, but it's generally useful. After we publish a new version of the driver with this change, we'll upgrade SarifCli to use it (in a separate commit), and remove SarifCli's copy of the method.

NOTE: There are other two seemingly-generally-useful methods in SarifCli's derived `AnalyzeCommand` class: `LogResult` and `LogResults`. But they're SARIF-specific, so they don't belong in the framework. With a little more work they could be refactored so the common parts could be put into the framework, but I won't do that now.


michaelcfanning:
+1

PR #522, opened 2016-09-02
Fix #506, Fix #521

### LINKED ISSUE #506: Driver framework produces binary analysis-specific message
When a skimmer throws an exception that the framework catches, the framework produces a message like this:

errorERR998.ExceptionInAnalyze : SV0003 : An exception was raised analyzing 'G:\Code\sarif-sdk\src\SarifCli.FunctionalTests\TestData\UrisMustBeValid\InvalidUriInFileChange.sarif' for check 'UrisMustBeValid' (which has been disabled). The exception may have resulted from a problem related to parsing image metadata and not specific to the rule, however.

That bit about "parsing image metadata" doesn't apply to all checkers.

### LINKED ISSUE #521: Framework insists on repeating file name in {0} of formatted messages
So for example, you get this:

```
G:\Code\sarif-sdk\src\SarifCli.FunctionalTests\TestData\EndColumnMustNotBeLessThanStartColumn\
EndColumnLessThanStartColumnInRelatedLocation.sarif(31,34): error SARIF013: In
"EndColumnLessThanStartColumnInRelatedLocation.sarif", at
"/runs/0/results/0/relatedLocations/0/physicalLocation/region/endColumn",
the value of the "endColumn" property is 1, which is less than the value of the "startColumn"
property, which is 2.
```

The framework appears to have been implemented this way for the sake of BinSkim, which really does want to mention the name of the binary at the start of each message, for example:

```
Foo.dll is not marked with DYNAMICBASE.
```

But for other tools, it doesn't make sense. Remove the SDK code that automatically prepends the file name as `{0}`. When BinSkim upgrades to a version of the SDK with this change, it will have to be updated to explicitly specify the file name in each list of formatted message arguments.

@michaelcfanning FYI

lgolding:
506: Driver framework produces binary analysis-specific message
521: Framework insists on repeating file name in {0} of formatted messages

521: The framework was implemented this way for the sake of BinSkim, which really does want to mention the name of the binary at the start of each message, for example:

    Foo.dll is not marked with DYNAMICBASE.

But for other tools, it doesn't always make sense. Remove the SDK code that automatically prepends the file name as `{0}`. When BinSkim upgrades to a version of the SDK with this change, it will have to be updated to explicitly specify the file name in each list of formatted message arguments.

### DIFF
--- a/src/Sarif.Driver.UnitTests/Sarif.Driver.UnitTests.csproj
+++ b/src/Sarif.Driver.UnitTests/Sarif.Driver.UnitTests.csproj
@@ -74,6 +74,7 @@
     <Compile Include="Sdk\FormatForVisualStudioTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="SarifHelpers.cs" />
+    <Compile Include="Sdk\RuleUtilitiesTests.cs" />
     <Compile Include="Sdk\SarifLoggerTests.cs" />
     <Compile Include="TestAnalysisContext.cs" />
     <Compile Include="TestAnalyzeCommand.cs" />
--- a/src/Sarif.Driver.UnitTests/Sdk/RuleUtilitiesTests.cs
+++ b/src/Sarif.Driver.UnitTests/Sdk/RuleUtilitiesTests.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Driver;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Driver.Sdk
+{
+    public class RuleUtilitiesTests
+    {
+        [Fact]
+        public void BuildResult_BuildsExpectedResult()
+        {
+            // Arrange
+            const string FormatId = "Default";
+            const string RuleId = "TST0001";
+            string[] Arguments = new string[] { "42", "54" };
+
+            var context = new TestAnalysisContext
+            {
+                TargetUri = new System.Uri("file:///c:/src/file.c"),
+                Rule = new Rule
+                {
+                    Id = RuleId,
+                    MessageFormats = new Dictionary<string, string>
+                    {
+                        [FormatId] = "Expected {0} but got {1}."
+                    }
+                }
+            };
+
+            var region = new Region
+            {
+                StartLine = 42
+            };
+
+            // Act.
+            Result result = RuleUtilities.BuildResult(
+                ResultLevel.Error,
+                context,
+                region,
+                FormatId,
+                Arguments);
+
+            // Assert.
+            result.RuleId.Should().Be(RuleId);
+
+            result.FormattedRuleMessage.FormatId.Should().Be(FormatId);
+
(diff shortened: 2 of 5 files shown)


src/Sarif.Driver.UnitTests/Sdk/RuleUtilitiesTests.cs
```diff
@@ -0,0 +1,61 @@
...
+                region,
+                FormatId,
+                Arguments);
+
+            // Assert.
+            result.RuleId.Should().Be(RuleId);
+
+            result.FormattedRuleMessage.FormatId.Should().Be(FormatId);
+
+            result.FormattedRuleMessage.Arguments.Count.Should().Be(Arguments.Length);
```
lgolding: Here's where I show that `BuildResult` no longer prepends the extra argument.

src/Sarif.Driver.UnitTests/Sarif.Driver.UnitTests.csproj
```diff
@@ -74,6 +74,7 @@
     <Compile Include="Sdk\FormatForVisualStudioTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="SarifHelpers.cs" />
+    <Compile Include="Sdk\RuleUtilitiesTests.cs" />
```
lgolding: Added a unit test file for the code I modified.

src/Sarif/RuleUtilities.cs
```diff
@@ -24,33 +22,22 @@ public static Result BuildResult(ResultLevel level, IAnalysisContext context, Re
...
-            result.RuleId = context.Rule.Id;
-
-            if (!string.IsNullOrEmpty(targetPath))
+            Result result = new Result
             {
-                // In the event of an analysis target, we always provide
-                // the local path to this item as the 0th argument
-                messageArguments = new string[arguments.Length + 1];
-                messageArguments[0] = Path.GetFileName(targetPath);
-                arguments.CopyTo(messageArguments, 1);
```
lgolding: Don't do that!

src/Sarif/RuleUtilities.cs
```diff
@@ -24,33 +22,22 @@ public static Result BuildResult(ResultLevel level, IAnalysisContext context, Re
                 throw new ArgumentNullException(nameof(arguments));
             }
 
-            string[] messageArguments = arguments;
```
lgolding: No need for the temp, since we're not going to modify `arguments`. Just use `arguments` directly.

src/Sarif/RuleUtilities.cs
```diff
@@ -24,33 +22,22 @@ public static Result BuildResult(ResultLevel level, IAnalysisContext context, Re
...
+                {
+                    FormatId = formatId,
+                    Arguments = arguments
+                },
 
-            result.Level = level;
+                Level = level
+            };
 
+            string targetPath = context.TargetUri?.LocalPath;
```
lgolding: Since we no longer use `targetPath` to do argument-munging logic, move its declaration down to its point of use.

src/Sarif/RuleUtilities.cs
```diff
@@ -24,33 +22,22 @@ public static Result BuildResult(ResultLevel level, IAnalysisContext context, Re
...
-                Arguments = messageArguments
-            };
+                FormattedRuleMessage = new FormattedRuleMessage()
+                {
+                    FormatId = formatId,
+                    Arguments = arguments
+                },
 
-            result.Level = level;
+                Level = level
```
lgolding: Replaced multiple assignments with property initialization syntax.

michaelcfanning:
:shipit:

PR #525, opened 2016-09-03
Add 'configuration' property to 'rule' instances (with possible value…

michaelcfanning:
…s, unknown, enabled, disabled). @lgolding


SetCurrentVersion.cmd
```diff
@@ -1,9 +1,9 @@
 set MAJOR_PREVIOUS=1
 set MINOR_PREVIOUS=5
-set PATCH_PREVIOUS=30
+set PATCH_PREVIOUS=31
 set PRERELEASE_PREVIOUS=
 
 set MAJOR=1
 set MINOR=5
-set PATCH=31
+set PATCH=32
```
lgolding: You won't need to make this change since I've already merged it.

src/ReleaseHistory.md
```diff
@@ -69,9 +69,13 @@
...
 
 ## **v1.5.29** [Driver](https://www.nuget.org/packages/Sarif.Driver/1.5.29) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/1.5.29)
-
 * Add `--quiet` option to suppress console output.
 
 ## **v1.5.30** [Driver](https://www.nuget.org/packages/Sarif.Driver/1.5.30) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/1.5.30)
+* Add static helper method `AnalyzeCommandBase.LogToolNotification`.
+
+## **v1.5.31** [Driver](https://www.nuget.org/packages/Sarif.Driver/1.5.31) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/1.5.31)
+* Minor updates to console output
```
lgolding: You'll want to take my merged change for 1.5.31. It's actually a breaking change.

src/Sarif/CodeGenHints.json
```diff
@@ -326,6 +326,21 @@
...
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "RuleConfiguration",
+        "description": "Values specifying the configuration of a rule.",
+        "memberNames": [
+          "Unknown",
+          "Enabled",
+          "Disabled"
+        ],
+        "allowMemberCountMismatch": true
```
lgolding: You don't need `allowMemberCountMismatch`. There are the same number of `memberNames` here are there are `enum` values in the schema.

src/Sarif/Autogenerated/IRule.cs
```diff
@@ -40,6 +40,11 @@ public partial interface IRule
         IDictionary<string, string> MessageFormats { get; }
 
         /// <summary>
+        /// A value specifying the configuration state of the rule.
```
lgolding: >configuration state of the rule [](start = 35, length = 31)

I understand why the member name is called `Configuration`. But I think the comment should be more explicit "A value specifying whether the rule is enabled."

lgolding:
:shipit:

PR #533, opened 2016-09-20
Proposal for annotatedCodeLocation 'annotations' member.

michaelcfanning:
Larry, does this look right per previous discussion? Note that I've proposed that an annotation can include an array of locations, so that users can easily duplicate messages to multiple locations.

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -127,6 +127,16 @@
           "type": "string"
         },
 
+        "annotations": {
+          "description": "A set of messages relevant to the current annotated code location.",
+          "type": "array",
+          "minItems": 1,
+          "uniqueItems": true,
+          "items": {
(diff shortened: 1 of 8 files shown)


michaelcfanning:
@lgolding

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -146,1050 +156,1072 @@
...
+            "description": "The region within the file where the result was detected.",
+            "$ref": "#/definitions/region"
+          }
         },
 
-        "length": {
-          "description": "The length of the region in bytes or characters.",
-          "type": "integer",
-          "minimum": 0
+        "required": [ "uri" ],
```
lgolding: `uri` can no longer be a required property of `physicalLocation`, since there is a circumstance where it's not required. The smart validator will need to handle this.

lgolding:
:+1: except for one very important comment about `physicalLocation.uri`.

PR #535, opened 2016-09-21
FInal 1.32 schema/other changes

michaelcfanning:


michaelcfanning:
@lgolding FYI, isolated final schema changes for 1.32. After this, I will build and submit the updated generated files. Producing this to verify the uri != required update.

PR #545, opened 2016-11-26
Semmle converter

lgolding:
This is a first cut at the Semme converter. It uses the `TextFieldParser` class from `Microsoft.VisualBasic.FileIO` to parse the CVS file. It emits `level`, `message`, `uri`, and `region` for each `result`. Note that Semmle defines a severity "`recommended`" which doesn't correspond to any member of our `ResultLevel` enumeration, so for now I've mapped it to `ResultLevel.Note`.

Semmle doesn't have a "rule id". What we think of as rules, they implement as "queries", and queries have "names", but they don't have anything like our opaque rule id.


src/Sarif.Converters/SemmleConverter.cs
```diff
@@ -11,8 +15,189 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
...
+            if (output == null)
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            _toolNotifications = new List<Notification>();
+
+            var results = GetResultsFromStream(input);
+
+            var tool = new Tool
```
michaelcfanning: >tool [](start = 16, length = 4)

I see that we don't have any other tool data we can extract from the output file. Too bad. Can we confirm the tool name with Semmle? I think they actually advertise this functionality as 'Semmle QL' in their product portfolio. Semmle itself being the company name.

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv
```diff
@@ -0,0 +1,80 @@
...
+"Incorrect lazy initialization of a static field","Initializing a static field without synchronization can be problematic in a multi-threaded context.","warning","Incorrect lazy initialization of static field [[""SKIP_CHAR_BUFFER""|""relative://src/main/java/org/apache/commons/io/IOUtils.java:169:6517:16""]]: Missing synchronization.","src/main/java/org/apache/commons/io/IOUtils.java","commons-io/revision-2016-November-08--14-46-36/src/main/java/org/apache/commons/io/IOUtils.java","2793","13","2793","57"
+"Busy wait","Calling 'Thread.sleep' to control thread interaction is less effective than waiting for a notification and may also result in race conditions. Merely synchronizing over shared variables in a loop to control thread interaction may waste system resources and cause performance problems.","warning","Prefer wait/notify or java.util.concurrent to communicate between threads.","src/main/java/org/apache/commons/io/FileUtils.java","commons-io/revision-2016-November-08--14-46-36/src/main/java/org/apache/commons/io/FileUtils.java","1708","21","1708","59"
```
michaelcfanning: >'Thread.sleep' [](start = 21, length = 14)

Note Semmle uses single apostrophes to enquote contents as I also prefer. :P Couldn't resist, sorry.

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+                  "endLine": 2718,
+                  "endColumn": 57
+                }
+              }
+            }
+          ]
+        },
+        {
+          "level": "warning",
+          "message": "Incorrect lazy initialization of static field [[\"SKIP_CHAR_BUFFER\"|\"relative://src/main/java/org/apache/commons/io/IOUtils.java:169:6517:16\"]]: Missing synchronization.",
```
michaelcfanning: >[[\"SKIP_CHAR_BUFFER\"|\"relative://src/main/java/org/apache/commons/io/IOUtils.java:169:6517:16\"]] [](start = 68, length = 100)

looks like Semmle messages can encode links within the messages, an interesting concept we don't provide. The SARIF analog here would be to convert this location into a related location. Probably the right thing to do is to grab the section of the message that is effectively the text string for the link, here, "SKIP_CHAR_BUFFER" and to convert it into a standard message, stuffed into related location. This suggestion only makes sense if this Semmle convention is restricted to provide links to declared symbols (we should ask about whether they use this linking strategy for arbitrary/other purposes)

"message": "\"SKIP_CHAR_BUFFER\" is declared here.",

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+          "locations": [
+            {
+              "resultFile": {
```
michaelcfanning: >resultFile [](start = 15, length = 10)

two months away from locking down the spec, i see that 'resultFile' is a bad name. :P it reads as if it were the output file for results.

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+                  "endLine": 270,
+                  "endColumn": 81
+                }
+              }
+            }
+          ]
+        },
+        {
+          "level": "note",
+          "message": "Method doCopyFile ignores exceptional return value of File.setLastModified.",
```
michaelcfanning: >doCopyFile [](start = 29, length = 10)

Semmle missed opportunity to both enquote symbol name and to provide link to declaration location. :) Someone should report out to them.

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+                  "endLine": 1430,
+                  "endColumn": 58
+                }
+              }
+            }
+          ]
+        },
+        {
+          "level": "warning",
+          "message": "This field is in a serializable class,  but is not serializable itself because CharSequence is not serializable.",
```
michaelcfanning: >   [](start = 60, length = 2)

Semmle has two spaces following the (strictly speaking unnecessary, I think ) comma in this sentence. Are these rules part of their open-source offering?

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+                  "endLine": 172,
+                  "endColumn": 59
+                }
+              }
+            }
+          ]
+        },
+        {
+          "level": "note",
+          "message": "toUpperCase without locale parameter.",
```
michaelcfanning: >toUpperCase [](start = 22, length = 11)

too brief message, missing quotes for symbol name

src/Sarif.Converters/SemmleConverter.cs
```diff
@@ -11,8 +15,189 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
...
+        {
+            switch (semmleSeverity)
+            {
+                case "error":
+                    return ResultLevel.Error;
+
+                case "warning":
+                    return ResultLevel.Warning;
+
+                case "recommendation":
```
michaelcfanning: >recommendation [](start = 22, length = 14)

OK, got it. Recommendation is, in fact, a note.

src/Sarif.Converters/SemmleConverter.cs
```diff
@@ -11,8 +15,189 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
...
+            return results.ToArray();
+        }
+
+        private Result ParseResult()
+        {
+            string[] fields = _parser.ReadFields();
+            return new Result
+            {
+                Level = ResultLevelFromSemmleSeverity(GetString(fields, FieldIndex.Severity)),
+                Message = fields[(int)FieldIndex.Message],
```
michaelcfanning: >Message [](start = 16, length = 7)

Need to parse message here for encoded link in order to create related locations (and scrub link data from the message, we have no encoding strategy here). Probably, this scenario should be included in the SARIF todo around whether to allow for inlined formatting/etc. of messages (for bolding, encoding links, etc.)

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -15,6 +15,7 @@
   </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.props))\build.props" />
   <ItemGroup>
+    <Reference Include="Microsoft.VisualBasic" />
```
michaelcfanning: ><Reference Include="Microsoft.VisualBasic" /> [](start = 4, length = 45)

why did this get added?

michaelcfanning: oh, i see, for CSV parsing. had no idea. i wonder how VB's apparently general purpose type handles enquoting of strings that contain the delimiter (a comma). Does it have some special understanding of this scenario? Anyway, no real note here.

---
In reply to: [90078642](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90078642) [](ancestors = 90078642)

michaelcfanning:
:clock1:

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv
```diff
@@ -0,0 +1,80 @@
...
+"Commented-out code","Commented-out code makes the remaining code more difficult to read.","warning","This comment appears to contain commented-out code.","src/main/java/org/apache/commons/io/input/BOMInputStream.java","commons-io/revision-2016-November-08--14-46-36/src/main/java/org/apache/commons/io/input/BOMInputStream.java","290","9","293","54"
+"Confusing overloading of methods","Overloaded methods that have the same number of parameters, where each pair of corresponding parameter types is convertible by casting or autoboxing, may be confusing.","recommendation","Method IOUtils.toInputStream(..) could be confused with overloaded method [[""toInputStream""|""relative://src/main/java/org/apache/commons/io/IOUtils.java:1521:59539:13""]], since dispatch depends on static types.","src/main/java/org/apache/commons/io/IOUtils.java","commons-io/revision-2016-November-08--14-46-36/src/main/java/org/apache/commons/io/IOUtils.java","1472","31","1472","43"
```
michaelcfanning: >relative: [](start = 319, length = 9)

Interesting convention to resolve a problem we are well aware of. :) Looking forward to seeing how you handle this in the converter.

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+            }
+          ]
+        },
+        {
+          "level": "warning",
+          "message": "Lines of code in files too high (1168). It is a warning for the value to be higher or equal to 1000.",
+          "locations": [
+            {
+              "resultFile": {
+                "uri": "src/main/java/org/apache/commons/io/FileUtils.java",
```
michaelcfanning: >src/main/java/org/apache/commons/io/FileUtils.jav [](start = 24, length = 49)

I think these URIs should be prefixed with a slash. 

See section 4.2
[http://www.ietf.org/rfc/rfc3986.txt](http://www.ietf.org/rfc/rfc3986.txt)

src/Sarif.Converters/SemmleConverter.cs
lgolding: Will send mail to Semmle with this and any other questions I see as I go through your comments.

src/Sarif.Converters/SemmleConverter.cs
lgolding: I looked for and didn't find a work item for message formatting, either in the sarif-sdk repo or in the sarif-spec repo. I filed sarif-standard/sarif-spec#275, "Should we allow formatting in messages?".

But the issue of parsing out links from the Semmle output, creating related locations for them, and scrubbing them, is a separate issue, for two reasons:

1. Semmle's embedded link format is unlikely to match whatever format(s) we decide to support for formatted messages:

```
Incorrect lazy initialization of static field [["SKIP_BYTE_BUFFER"|"relative://src/main/java/org/apache/commons/io/IOUtils.java:170:6561:16"]]: Missing synchronization.
```

2. In any case, the act of creating related locations and scrubbing these escape sequences is a separate work item. I filed #546, "Semmle: Create related locations from embedded links in messages".

src/Sarif.Converters/SemmleConverter.cs
lgolding: Will send mail to Semmle with this and any other questions I see as I go through your comments.

---
In reply to: [90072851](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90072851) [](ancestors = 90072851)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv
lgolding: I never objected to single quotes within double quotes (which is what we have here). :P

---
In reply to: [90073074](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90073074) [](ancestors = 90073074)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
lgolding: I'll add to my mail to them, the question of whether there are restrictions on the links they embed.

---
In reply to: [90075185](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90075185) [](ancestors = 90075185)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
lgolding: The changes resulting from the standards process might merit a major version change to the spec, in which case we'd be free to change this.

---
In reply to: [90076011](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90076011) [](ancestors = 90076011)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+          "level": "warning",
+          "message": "Avoid using thread priorities. The effect is unpredictable and not portable.",
+          "locations": [
+            {
+              "resultFile": {
+                "uri": "src/main/java/org/apache/commons/io/FileCleaningTracker.java",
+                "region": {
+                  "startLine": 211,
+                  "startColumn": 13,
+                  "endLine": 211,
```
lgolding: It was an easy change, so I fixed this.

---
In reply to: [90075704](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90075704) [](ancestors = 90075704)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
lgolding: I agree. Not actionable by the converter, of course -- the message comes straight from the Semmle file.

---
In reply to: [90077039](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90077039) [](ancestors = 90077039)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
```diff
@@ -0,0 +1,1353 @@
...
+  "$schema": "http://json.schemastore.org/sarif-1.0.0",
+  "version": "1.0.0",
+  "runs": [
+    {
+      "tool": {
+        "name": "Semmle"
+      },
+      "results": [
+        {
+          "level": "warning",
```
lgolding: Fixed to not emit "warning". I don't think we have default handling _per se_ in the SDK. I think each producer is deciding whether to emit default values.

---
In reply to: [90075506](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90075506) [](ancestors = 90075506)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
lgolding: I assume you're asking because we'd offer to fix it if they were. I'll ask them.

---
In reply to: [90076792](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90076792) [](ancestors = 90076792)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv.sarif
lgolding: I'm not terribly comfortable pushing back on their typographic conventions. But I'll mention the opportunity to add a related location.

---
In reply to: [90076419](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90076419) [](ancestors = 90076419)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv
lgolding: What problem, and what convention? Are you talking about the colon-separated region components 1521:59539:13 in the bracketed link? Not sure what you think the converter should do here. If we do the work to extract these links (I filed a work item for that), then we would parse these out when we created the related location. Although really, to make that safe, they ought to specify a fragment: '.../commons/io/IOUtils.java#1521:59539:13'.

---
In reply to: [90079504](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90079504) [](ancestors = 90079504)

src/Sarif.FunctionalTests/ConverterTestData/Semmle/results.csv
lgolding: Oh, I see your point: There are scenarios where message strings appear within syntactically required double-quotes (like the ones that enclose fields that might contain commas in CSV files), and in those scenarios, the path of least resistance is to use single quotes in the message string. But you can usually avoid the problem by escaping. For example, in CSV, you can escape double quotes by doubling them: "This is ""quoted,"" you see." In XML, you can use the &quot; entity. And if you're using a library that produces your output format (CSV, XML or whatever), then it should take care of the escaping for you. So I'm still not convinced that this is a reason to use non-American-standard English.

---
In reply to: [90761992](https://github.com/Microsoft/sarif-sdk/pull/545#discussion_r90761992) [](ancestors = 90761992,90073074)

Issue #552, opened 2017-01-02
Framework emits invalid attribute correlationId


lgolding:
By the spec, it should be `automationId`.

PR #554, opened 2017-01-03
Fortify FPR converter

lgolding:
This is a first cut at a converter for the Fortify FPR format.

NOTES:

1. We emit the command line arguments, but at present we hard-code the tool executable name. The executable is actually present in the file, under a `Property` element:

    ```
    <Property>
        <name>com.fortify.SCAExecutablePath</name>
        <value>C:Program FilesHP_FortifyHP_Fortify_SCA_and_Apps_4.30binsourceanalyzer.exe</value>
    </Property>
    ```

    ... but it's missing the backslashes, which appears to be a Fortify bug.

2. This is the first converter that populates what SARIF refers to as the `automationId` property. This led to the discovery of #552, "Framework emits invalid attribute correlationId" -- the SDK still used the old name `correlationId` for this property (contrary to both the spec and the schema), so I fixed that.

### DIFF
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
@@ -122,7 +123,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     {
                         ParseUuid();
                     }
-                    if (AtStartOfNonEmpty(_strings.Build))
+                    // Note: CreatedTS is an empty element (it has only attributes),
+                    // so we can't call AtStartOfNonEmpty here.
+                    else if (AtStartOf(_strings.CreatedTimestamp))
+                    {
+                        ParseCreatedTimestamp();
+                    }
+                    else if (AtStartOfNonEmpty(_strings.Build))
                     {
                         ParseBuild();
                     }
@@ -153,6 +160,20 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         private void ParseUuid()
         {
             _runId = _reader.ReadElementContentAsString();
+        }
+
+        private void ParseCreatedTimestamp()
+        {
+            string date = _reader.GetAttribute(_strings.DateAttribute);
+            string time = _reader.GetAttribute(_strings.TimeAttribute);
+            if (!string.IsNullOrEmpty(date) && !string.IsNullOrEmpty(time))
+            {
+                string dateTime = date + "T" + time;
+                _invocation.StartTime = DateTime.Parse(dateTime, CultureInfo.InvariantCulture);
+            }
+
+            // Step past the empty element.
+            _reader.Read();
         }
 
         private void ParseBuild()
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -20,6 +20,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private XmlReader _reader;
         private Invocation _invocation;
+        private string _runId;
         private string _automationId;
         private List<Result> _results = new List<Result>();
         private List<Notification> _toolNotifications = new List<Notification>();
@@ -66,7 +67,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             ParseFprFile(input);
             AddMessagesToResults();
 
-            output.Initialize(id: null, automationId: _automationId);
+            output.Initialize(id: _runId, automationId: _automationId);
 
             output.WriteTool(tool);
             output.WriteInvocation(_invocation);
@@ -117,6 +118,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             {
                 while (_reader.Read())
                 {
+                    if (AtStartOfNonEmpty(_strings.Uuid))
+                    {
+                        ParseUuid();
+                    }
                     if (AtStartOfNonEmpty(_strings.Build))
                     {
                         ParseBuild();
@@ -143,6 +148,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     }
                 }
             }
+        }
+
+        private void ParseUuid()
+        {
+            _runId = _reader.ReadElementContentAsString();
         }
 
         private void ParseBuild()
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -74,6 +74,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             ParseFprFile(input);
             AddMessagesToResults();
             AddSnippetsToResults();
+            AddSnippetsToAnnotatedCodeLocations();
 
             output.Initialize(id: _runId, automationId: _automationId);
 
@@ -585,6 +586,31 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     _snippetIdToSnippetTextDictionary.TryGetValue(snippetId, out snippetText))
                 {
                     result.Snippet = snippetText;
+                }
+            }
+        }
+
+        private void AddSnippetsToAnnotatedCodeLocations()
+        {
+            foreach (Result result in _results)
+            {
+                if (result.CodeFlows != null)
+                {
+                    foreach (CodeFlow codeFlow in result.CodeFlows)
+                    {
+                        if (codeFlow.Locations != null)
+                        {
+                            foreach (AnnotatedCodeLocation acl in codeFlow.Locations)
+                            {
+                                string snippetId, snippetText;
+                                if (_aclToSnippetIdDictionary.TryGetValue(acl, out snippetId) &&
+                                    _snippetIdToSnippetTextDictionary.TryGetValue(snippetId, out snippetText))
+                                {
+                                    acl.Snippet = snippetText;
+                                }
+                            }
+                        }
+                    }
                 }
             }
         }
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Text;
 using System.Xml;
 
 namespace Microsoft.CodeAnalysis.Sarif.Converters
@@ -13,6 +14,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     {
         private readonly NameTable _nameTable;
         private readonly FortifyFprStrings _strings;
+        private Invocation _invocation;
 
         /// <summary>Initializes a new instance of the <see cref="FortifyFprConverter"/> class.</summary>
         public FortifyFprConverter()
@@ -45,11 +47,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 Name = "Fortify"
             };
 
+            _invocation = new Invocation();
+
             ParseFprFile(input);
 
             output.Initialize(id: null, correlationId: null);
 
             output.WriteTool(tool);
+            output.WriteInvocation(_invocation);
 
             output.OpenResults();
             output.CloseResults();
@@ -96,6 +101,19 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private void ParseCommandLineArguments(XmlReader reader)
         {
+            var sb = new StringBuilder();
+            reader.MoveToElement();
+            reader.Read();
+            while (Ref.Equal(reader.LocalName, _strings.Argument))
+            {
+                string argument = reader.ReadElementContentAsString();
+                if (sb.Length > 0)
+                    sb.Append(' ');
+                sb.Append(argument);
+                reader.MoveToElement();
+            }
+
+            _invocation.CommandLine = sb.ToString();
         }
     }
 }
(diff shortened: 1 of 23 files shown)


src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,509 @@
...
+            switch (fileType)
+            {
+                case "xml":
+                    return "text/xml";
+
+                case "tsql":
+                    return "text/x-sql";
+
+                default:
+                    return "unknown/" + fileType;
```
lgolding: >unknown [](start = 28, length = 7)

We need to get the full list of values that the `File@type` attribute can have and decide what MIME type to emit for each. Some values, such as `"msil"`, don't have defined MIME types as far as I could find.

src/Sarif.Converters/ToolFormat.cs
```diff
@@ -14,8 +14,10 @@ public enum ToolFormat
         ClangAnalyzer,
         /// <summary>CppCheck's file format.</summary>
         CppCheck,
-        /// <summary>Fortify's file format.</summary>
+        /// <summary>Fortify's report file format.</summary>
         Fortify,
+        /// <summary>Fortify's FPR file format.</summary>
+        FortifyFpr,
```
michaelcfanning: >FortifyFpr [](start = 8, length = 10)

I wonder if the FPR shouldn't be given the core 'Fortify' value and the Fortify report could be granted 'FortifyReport'. The FPR is the really comprehensive thing and, generally, what we would prefer to convert.

michaelcfanning: Yes, this would be a breaking change to utility users but we don't have a large user base for this app. :)

---
In reply to: [94533957](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94533957) [](ancestors = 94533957)

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -364,7 +364,7 @@ private void EnsureInitialized()
         {
             if (_writeConditions == Conditions.None)
             {
-                Initialize(id : Guid.NewGuid().ToString(), correlationId: null);
+                Initialize(id : Guid.NewGuid().ToString(), automationId: null);
```
michaelcfanning: >automationId [](start = 59, length = 12)

my final 'fine-tooth comb' review of spec vs. sdk missed this oversight. tsk. :)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+        private const string FortifyExecutable = "[REMOVED]insourceanalyzer.exe";
+
+        private readonly NameTable _nameTable;
+        private readonly FortifyFprStrings _strings;
+
+        private XmlReader _reader;
+        private Invocation _invocation;
+        private string _runId;
+        private string _automationId;
+        private List<Result> _results = new List<Result>();
```
michaelcfanning: > = new List<Result>() [](start = 37, length = 21)

can we put all this instance member initialization in the constructor. I prefer this mostly so that debugging doesn't jump outside of methods.

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+        private void ParseUuid()
+        {
+            _runId = _reader.ReadElementContentAsString();
+        }
+
+        private void ParseCreatedTimestamp()
+        {
+            string date = _reader.GetAttribute(_strings.DateAttribute);
+            string time = _reader.GetAttribute(_strings.TimeAttribute);
+            if (!string.IsNullOrEmpty(date) && !string.IsNullOrEmpty(time))
```
michaelcfanning: >string [](start = 17, length = 6)

Most code in the SDK uses the non-alias actual .NET names for static methods, this is the standard. So, prefer String.Format. NIT

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+        private void ParseFile()
+        {
+            string fileType = _reader.GetAttribute(_strings.TypeAttribute);
+            int length = 0;
+            string sizeAttribute = _reader.GetAttribute(_strings.SizeAttribute);
+            if (sizeAttribute != null)
+            {
+                if (!int.TryParse(sizeAttribute, out length))
+                {
+                    length = 0;
```
michaelcfanning: >length = 0 [](start = 20, length = 10)

this assignment is unnecessary, s/be already set to 0 if TryParse fails.

michaelcfanning: Is there no error we should be raising here for this case, btw?

---
In reply to: [94534761](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94534761) [](ancestors = 94534761)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+                    });
+            }
+        }
+
+        private string FprTypeToMimeType(string fileType)
+        {
+            switch (fileType)
+            {
+                case "xml":
+                    return "text/xml";
```
michaelcfanning: >text/xml [](start = 28, length = 8)

Can't we use the MimeType helper to get these values? may need to add tsql. The MimeType helper (in \Writers\) should accept the fileName directly. Note that this helper assumes a binary mime type when encountering something unknown.

src/Sarif.Converters/FortifyFprConverter.cs
michaelcfanning: hm, should look at this. we should pinvoke to the OS helper that returns mime type and see what it says about a managed assembly.

---
In reply to: [94462728](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94462728) [](ancestors = 94462728)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+                    {
+                        Step = step++,
+                        PhysicalLocation = physLoc
+                    };
+
+                    // Remember the id of the snippet associated with this location.
+                    // We'll use it to fill the snippet text when we read the Snippets element later on.
+                    if (!string.IsNullOrEmpty(snippetId))
+                    {
+                        _aclToSnippetIdDictionary.Add(acl, snippetId);
```
michaelcfanning: >Add [](start = 50, length = 3)

why use Add here? are you doing this explicitly to indicate you expect acl will never already present as a key? if so, i support that nuance. :)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+
+        private PhysicalLocation ParsePhysicalLocationFromSourceInfo()
+        {
+            string path = _reader.GetAttribute(_strings.PathAttribute);
+
+            int startLine;
+            string lineAttr = _reader.GetAttribute(_strings.LineAttribute);
+            if (!int.TryParse(lineAttr, out startLine))
+            {
+                startLine = 0;
```
michaelcfanning: >startLine = 0 [](start = 16, length = 13)

more unnecessary assignment. I suggest you either add error handling here or, if there's a reason not to, just strand the TryParse call on its own:

TryParse(lineAtt, outStartLine)

with no if+conditional

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+            int endColumn;
+            string colEndAttr = _reader.GetAttribute(_strings.ColEndAttribute);
+            if (!int.TryParse(colEndAttr, out endColumn))
+            {
+                endColumn = 0;
+            }
+
+            return new PhysicalLocation
+            {
+                Uri = new Uri(path, UriKind.Relative),
```
michaelcfanning: >Uri = new Uri(path, UriKind.Relative) [](start = 16, length = 37)

I begin to think it is a good practice to emit all paths as relative to an analysis root. the question is whether you should emit the baseUriId in cases where you only have a single root. this question exists for the Semmle converter as well. i tend to think it's fine to omit it.

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -0,0 +1,631 @@
...
+                Region = new Region
+                {
+                    StartLine = startLine,
+                    StartColumn = startColumn,
+                    EndColumn = endColumn
```
michaelcfanning: >EndColumn = endColumn [](start = 20, length = 21)

hm, so a fortify result doesn't ever span multiple lines? probably the reporting is different for these cases and you haven't encountered an occurrence. we should look at more samples (and generate some of our own).

michaelcfanning:
For this converter and every other one, I think we should consider a new option that will inject the base64 encoded file contents, if we can locate them. Just a heads up on another proposed change. Semmle packages up all source files and so can ensure that this data is available whenever it produces a results file.

michaelcfanning:
This converter is an excellent start. We need to check in some samples. I would like to examine the emitted SARIF and to look at the FPRs more closely. I have hundreds of FPRs we can look at and can publish these to you (did I do that already)? Also, we should use a Fortify instance to generate useful samples that cover the range of cases (and which stretch the converter in some ways like attempting to produce a result that spans multiple files). We have a license and can produce these samples easily, obviously. I will give this a second look tomorrow. Nothing significant to comment on in this code so far. :)

michaelcfanning:
:shipit:

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Done.

---
In reply to: [94534318](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94534318) [](ancestors = 94534318)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Re: assignment: fixed. Thanks, I didn't realize that the value of the out parameter was defined/documented if the parse failed. It is ([https://msdn.microsoft.com/en-us/library/f02979c7(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/f02979c7(v=vs.110).aspx))

---
In reply to: [94534761](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94534761) [](ancestors = 94534761)

lgolding: Re: error reporting: I'll look at what other converters do when the input file is invalid. One possibility is to emit toolNotifications. But then the toolNotifications would be a mix of notifications from the original scanner (the Errors/Error elements in the FPR file) and from the converter.

---
In reply to: [94534782](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94534782) [](ancestors = 94534782,94534761)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Fixed everywhere.

---
In reply to: [94534692](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94534692) [](ancestors = 94534692)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Fixed. I now use this pattern:
```
            int length = 0;
            string sizeAttribute = _reader.GetAttribute(_strings.SizeAttribute);
            if (sizeAttribute != null)
            {
                int.TryParse(sizeAttribute, out length);
            }
```

---
In reply to: [94535783](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94535783) [](ancestors = 94535783)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Yes, in the only samples I've looked at, the SourceLine element has only "line", "colStart", and "colEnd" attributes. From the fact that the attribute is _not_ named "lineStart", I suspect that we won't find any samples with a "lineEnd" attribute.

---
In reply to: [94535925](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94535925) [](ancestors = 94535925)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: Yes, I did think about that, and decided that there was no point in emitting a baseUriId if there was only a single root. I think baseUriId should be used only when you have multiple roots (SRCROOT, BINROOT, OBJROOT...)

---
In reply to: [94535867](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94535867) [](ancestors = 94535867)

lgolding:
Yes, you sent me a file share with many samples (that's what I used to write the converter), but you asked me not to check those in. Sure, if you can run the tool and produce some check-in-able samples, that would be great (</Lumbergh>).

---
In reply to: [270305913](https://github.com/Microsoft/sarif-sdk/pull/554#issuecomment-270305913) [](ancestors = 270305913)

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: No, I didn't think about the subtle difference between `d[key] = val` and `d.Add(key, val)` (namely, that the latter throws if `key` is already present in `d`). In this case, since `acl` is a newly created object, and since all the SARIF classes have reference equality semantics, it's guaranteed not to exist in the dictionary, so it doesn't matter which I use.

---
In reply to: [94535642](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94535642) [](ancestors = 94535642)

src/Sarif.Converters/ToolFormat.cs
lgolding: I don't object to the change, but let's do it in a separate PR. The diffs are hard to examine otherwise.

---
In reply to: [94533983](https://github.com/Microsoft/sarif-sdk/pull/554#discussion_r94533983) [](ancestors = 94533983,94533957)

PR #555, opened 2017-01-11
Extract multiple related locations from the messages in a Semmle csv file

rtaket:
The VS viewer change is essentially unrelated. However, I couldn't open the converted Semmle file in VS without it. I can split it into another PR if you guys so desire.


michaelcfanning:
:shipit:

PR #558, opened 2017-01-26
Update GetFirstSentence to handle sentences with ellipsis

rtaket:


src/Sarif/ExtensionMethods.cs
```diff
@@ -297,7 +298,14 @@ public static string GetFirstSentence(string text)
 
                     case '.':
                     {
-                        if (withinQuotes || withinParentheses) { continue; }
+                        if (withinQuotes || withinParentheses || withinEllipsis) { continue; }
+                        if (length < text.Length && text[length] == '.')
```
michaelcfanning: This code would regard a series of four dots as an ellipsis. But I'm not sure the situation warrants any additional complexity.

michaelcfanning: LGTM

PR #560, opened 2017-02-15
Use Semmle query name as the SARIF rule id.

rtaket:
Fix bugs preventing the call tree from displaying.

### DIFF
--- a/src/Sarif.Converters/SemmleConverter.cs
+++ b/src/Sarif.Converters/SemmleConverter.cs
@@ -114,6 +114,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             Region region = MakeRegion(fields);
             var result = new Result
             {
+                RuleId = fields[(int)FieldIndex.QueryName],
                 Message = normalizedMessage,
                 Locations = new Location[]
                 {
--- a/src/Sarif.FunctionalTests/ConverterTestData/SemmleQL/results.csv.sarif
+++ b/src/Sarif.FunctionalTests/ConverterTestData/SemmleQL/results.csv.sarif
@@ -8,6 +8,7 @@
       },
       "results": [
         {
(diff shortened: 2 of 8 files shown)


src/Sarif.Converters/SemmleConverter.cs
```diff
@@ -114,6 +114,7 @@ private Result ParseResult()
             Region region = MakeRegion(fields);
             var result = new Result
             {
+                RuleId = fields[(int)FieldIndex.QueryName],
```
lgolding: Does the query name meet the requirements for a SARIF rule id? "... a stable, opaque identifier for the rule" Isn't it legal for the author of a Semmle query to rename it?

lgolding:
:shipit:

rtaket:
@Larry Golding<mailto:Laurence.Golding@microsoft.com>, this doesn’t meet the requirements for a SARIF rule id. As you mentioned, the rule author can change the query name anytime.

Maybe there’s a better solution? The issue identified by the Semmle team is that the query name is missing in the SARIF file. So I thought the correct place for the query name was the rule ID. However, as you pointed out, the query name doesn’t meet the requirements of the SARIF rule id. Any suggestions? I could prefix the message with the query name (and leave the rule ID blank.)

lgolding:
It depends on what the customer is trying to accomplish. The name of the issue was "Use Semmle query name as the SARIF rule id", but it sounds like Semmle doesn't care _where_ the query name appears, just that it appear somewhere. Prepending the rule id to the message, as you suggested, is one option. Another would be to add a "queryName" property to the result object's property bag. You could ask them which they'd prefer.

---
In reply to: [282825800](https://github.com/Microsoft/sarif-sdk/pull/560#issuecomment-282825800) [](ancestors = 282825800)

PR #561, opened 2017-02-21
Fix #559: Control Invocation property logging

### LINKED ISSUE #559: Provide control over which Invocation properties are logged
Provide a means for a tool to specify which properties of the `Invocation` object to populate. For security, the default should be for the logger to populate only the `StartTime` and `EndTime` properties.

lgolding:
By default, log only the `StartTime` and `EndTime` properties. Allow the tool user to specify additional properties to log by means of a command line option.


lgolding:
@michaelcfanning ping

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -168,7 +170,30 @@ private bool ValidateFile(TContext context, string filePath, bool? shouldExist)
...
         }
 
-        internal AggregatingLogger InitializeLogger(IAnalyzeOptions analyzeOptions)
+        private static bool ValidateInvocationPropertiesToLog(TContext context, IEnumerable<string> propertiesToLog)
+        {
+            bool succeeded = true;
+
+            if (propertiesToLog != null)
+            {
+                List<string> validPropertyNames = typeof(Invocation).GetProperties(BindingFlags.Public | BindingFlags.Instance)
```
michaelcfanning: You could have used a dictionary here initialized with a case-insensitive comparer, and skipped all the ToUpper calls. Just create the enumerable list as you're doing and pass that to the Dictionary ctor.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -69,5 +69,12 @@ public class AnalyzeOptionsBase : IAnalyzeOptions
             Separator = ';',
             HelpText = "Path to plug-in that will be invoked against all targets in the analysis set.")]
         public IEnumerable<string> PluginFilePaths { get; set; }
+
+        [Option(
+            'i',
+            "invocation-properties",
+            Separator = ';',
+            HelpText = "Properties of the Invocation object to log. NOTE: StartTime and EndTime are always logged.")]
```
michaelcfanning: when we support deterministic output, starttime and endtime will not appear. no call to action, just noting.

src/Sarif/Core/Invocation.cs
```diff
@@ -4,20 +4,40 @@
...
+            {
+                invocation.WorkingDirectory = Environment.CurrentDirectory;
+            }
+
+            if (invocation.ShouldLog(nameof(CommandLine)))
+            {
+                invocation.CommandLine = Environment.CommandLine;
+            }
 
             if (emitMachineEnvironment)
```
michaelcfanning: hm, should emitting of machine environment be handled by the property name control instead as a simplification?

src/Sarif/Errors.cs
```diff
@@ -201,6 +202,26 @@ public static void LogExceptionAccessingFile(IAnalysisContext context, string fi
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // '{0}' is not a property of the Invocation object.
+            context.Logger.LogConfigurationNotification(
+                CreateNotification(
+                    context.TargetUri,
+                    ERR997_InvalidInvocationPropertyName,
+                    NotificationLevel.Error,
+                    null,
```
michaelcfanning: provide name of parameter here for the embedded literals (and all subsequent args, of course, which is required)

michaelcfanning:
+1 left a few minor points. We could consider removing the machine environment argument from the command-line, but not strictly required. The whole bundle of relevant stuff here (machine env, items to redact, opt-in properties) becomes a bit strange/subtle. next time we put the hood up on this area, should have a broad talk.

PR #565, opened 2017-05-15
Update VSIX to build in VS2015 and work in VS2015 and V2017

rtaket:
Add references to packages required to build in VS215
Add InstallationTargets for VS2015 and VS2017
Update native projects to import the custom props file after the Microsoft props file so the default values are overwritten
Fix the rebasing logic so it splits paths on directory names (the old logic went char by char, so it was possible to rebase to nonexistent paths


michaelcfanning:
:shipit:

PR #569, opened 2017-05-19
Add response file support to SDK driver framework. Update binskim to …

michaelcfanning:
…permit all binskim analysis to pass. Bump version. @lgolding


src/Sarif.Driver.UnitTests/Sdk/EntryPointUtilitiesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+using FluentAssertions;
+
+using Moq;
```
lgolding: Extra blank line. Here we have three different third-party usings, and only one is separated.

src/Sarif.Driver.UnitTests/Sdk/EntryPointUtilitiesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            var args = new[] { "/a", "@" + ResponseFileName, "/f" };
+
+            string[] result = EntryPointUtilities.GenerateArguments(args, mockFileSystem, MakeMockEnvironmentVariables());
+
+            result.Length.Should().Be(6);
+            result.Should().ContainInOrder("/a", "/b", "/c:val", "/d", "/e", "/f");
+        }
+
+        [Fact]
+        public void EnryPointUtilities_GenerateArguments_IgnoresNonexistentResponseFile()
```
lgolding: >IgnoresNonexistentResponseFile [](start = 57, length = 30)

I'd expect that to be an error.

src/Sarif.Driver.UnitTests/Sdk/EntryPointUtilitiesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            var args = new[] { "@" + responseFileNameArgument };
+
+            string[] result = EntryPointUtilities.GenerateArguments(args, mockFileSystem, mockEnvironmentVariables);
+
+            result.Length.Should().Be(3);
+            result.Should().ContainInOrder("a", "b", "c");
+        }
+
+        private static IFileSystem MakeMockFileSystem(string fileName, string[] fileContents)
```
lgolding: >MakeMockFileSystem [](start = 35, length = 18)

This and MakeMockEnvironmentVariables are generally useful. I'd suggest breaking them into a separate file with static class MakeMock and methods FileSystem and EnvironmentVariables (=> MakeMock.FileSystem). Or maybe MockFactory.MakeFileSystem.

src/Sarif.Driver.UnitTests/Sdk/EntryPointUtilitiesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            var mock = new Mock<IFileSystem>(MockBehavior.Strict);
+            mock.Setup(fs => fs.FileExists(It.IsAny<string>())).Returns((string s) => s.Equals(fileName));
```
lgolding: >) [](start = 103, length = 1)

StringComparison.OrdinalIgnoreCase

src/Sarif.Driver.UnitTests/Sdk/EntryPointUtilitiesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            return mock.Object;
+        }
+
+        private static IEnvironmentVariables MakeMockEnvironmentVariables()
+        {
+            return MakeMockEnvironmentVariables(new Dictionary<string, string>());
+        }
+
+        // This regex is only approximate (environment variable names can contain
+        // non-"word" characters), but good enough for this tests.
```
lgolding: >this [](start = 55, length = 4)

this => these

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
...
+            string[] args,
+            IFileSystem fileSystem,
+            IEnvironmentVariables environmentVariables)
+        {
+            List<string> expandedArguments = new List<string>();
+
+            foreach (string argument in args)
+            {
+                if (argument[0] != '@' &&
+                    (argument.Length == 1 || argument[1] != '@'))
```
lgolding: >) [](start = 63, length = 1)

Is this an escaping mechanism? Is the idea that an argument "@@something" is treated as a literal argument ("don't interpret the leading '@' as introducing a response file name")? In that case I'd expect the argument to "@@something" to be interpreted as the literal string "@something" (and there needs to be a test for this logic).

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    expandedArguments.Add(argument);
+                    continue;
+                }
+
+                string responseFile = argument.Trim('"').Substring(1);
```
lgolding: >  [](start = 0, length = 1)

Consider breaking some of this logic into helper functions, so that this method reduces to (for example):
```
foreach (string argument in args)
{
    if (IsResponseFile(argument))
    {
        AddResponseFileContents(ResponseFileName(argument), expandedArguments);
    }
    else
    {
        argument = UnescapeAtSign(argument);
        expandedArguments.Add(argument);
    }
}
```

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
...
+                try
+                {
+                    responseFile = environmentVariables.ExpandEnvironmentVariables(responseFile);
+                    responseFile = fileSystem.GetFullPath(responseFile);
+
+                    if (!fileSystem.FileExists(responseFile))
+                        responseFile = null;
```
lgolding: >; [](start = 43, length = 1)

See my comment in the unit tests file (which I reviewed first, that's why the comment is there). I think this is an error condition.

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+                catch (IOException)
+                {
+                    responseFile = null;
```
lgolding: >responseFile [](start = 20, length = 12)

Likewise here. Maybe we couldn't open the response file because it was locked or we don't have permission. The tool runs and returns 0, with only the console warning to show for it.

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
...
+        private static void ExpandResponseFile(string[] responseFileLines, List<string> expandedArguments)
+        {
+            // BUG : This function does not allow response files to escape quotes like they can be escaped
+            // on the command line. Suggest following CommandLineToArgVW rules here:
+            // http://msdn.microsoft.com/en-us/library/windows/desktop/bb776391.aspx
+            // 
+            // * 2n backslashes followed by a quotation mark produce n backslashes followed by a quotation mark.
+            // * (2n) + 1 backslashes followed by a quotation mark again produce n backslashes followed by a quotation mark.
+            // * n backslashes not followed by a quotation mark simply produce n backslashes.
+            //
```
lgolding: >// [](start = 12, length = 2)

Why not just P/Invoke CommandLineToArgVW here instead of writing a partial implementation of its logic?

lgolding:
:shipit:

lgolding:
:clock1:

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
```diff
@@ -0,0 +1,111 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
lgolding: (I'd put this comment on the PR description but apparently that's not supported...) -- The description says you bumped the version, but I don't see that.

src/Sarif.Driver/Sdk/EntryPointUtilities.cs
michaelcfanning: No escaping mechanism. I am not aware of a reasonable scenario where @@ would be an expected prefix for an argument value

---
In reply to: [117518574](https://github.com/Microsoft/sarif-sdk/pull/569#discussion_r117518574) [](ancestors = 117518574)

lgolding: As we discussed: the original author did intend it as an escape mechanism, to allow a literal argument beginning with "@", but you're going to rip that out.

---
In reply to: [117542682](https://github.com/Microsoft/sarif-sdk/pull/569#discussion_r117542682) [](ancestors = 117542682,117518574)

PR #570, opened 2017-05-19
Add a --rich-return-code argument to core command

michaelcfanning:
@lgolding for Binskim request, some drivers want to return a 'rich' return code that consists of observed runtime conditions.


src/build.props
```diff
@@ -42,7 +42,7 @@
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" Condition="'$(DisableBinSkimTarget)'!='true'"/>
   <PropertyGroup>
     <BinSkimVerboseOutput>false</BinSkimVerboseOutput>
-    <BinSkimExePath>$(MsBuildThisFileDirectory)packages\Microsoft.CodeAnalysis.BinSkim.1.3.9\tools\x64\BinSkim.exe</BinSkimExePath>
+    <BinSkimExePath>$(MsBuildThisFileDirectory)packages\Microsoft.CodeAnalysis.BinSkim.1.4.0\tools\x64\BinSkim.exe</BinSkimExePath>
   </PropertyGroup>
```
lgolding: I never noticed this before: why does the SARIF SDK have a dependency on (or at least, refer to) a particular tool built from the SDK?

lgolding:
:shipit:

src/build.props
michaelcfanning: SARIF SDK runs binskim as part of its production build!

---
In reply to: [117521614](https://github.com/Microsoft/sarif-sdk/pull/570#discussion_r117521614) [](ancestors = 117521614)

PR #578, opened 2017-06-06
Change enum ToolFormat to a set of const strings

lgolding:
This is the first step in enabling third-party converter extensibility.


src/Sarif.Converters/Extensions.cs
```diff
@@ -9,6 +9,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
     public static class Extensions
     {
+        // Compare tool format strings with appropriate comparison type.
+        public static bool MatchesToolFormat(this string toolFormat, string other)
+        {
+            return toolFormat.Equals(other, StringComparison.OrdinalIgnoreCase);
```
michaelcfanning: >StringComparison.OrdinalIgnoreCase [](start = 44, length = 34)

well, fine, this is ok. but we probably should add a test that ensures no string declared in ToolFormat differs from another only by case.

src/Sarif.Converters/ToolFormat.cs
```diff
@@ -4,27 +4,36 @@
 namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
     /// <summary>Values referring to a source format to convert to the static analysis results interchange format.</summary>
-    public enum ToolFormat
+    public static class ToolFormat
```
michaelcfanning: >ToolFormat [](start = 24, length = 10)

as per other mail, suggest adding a simple test that puts all ToolFormat members into a case-insensitive dictionary or set, ensuring no member is already present. would use reflection for this to keep the test maintenance free as type changes.

michaelcfanning:
:shipit:

lgolding:
Abandoned. I was having trouble getting this to build after merging with recent changes. So I just updated my master branch and redid the changes on top of it.

PR #580, opened 2017-06-07
Enable ConvertToSarif to get converters from plugin assemblies

lgolding:
In the course of doing this, we factor all logic for instantiating a converter out of the `ToolFormatConverter` class and into a new `ConverterFactory` class.

Note that we do not use MEF. We don't need it, because the end user tells us on the command line, by way of the `-tool` option, exactly which converter class he wants.

Notes:
- The `ConverterFactory` is a static class. At present, it's not necessary to mock it (we can use the real class in unit tests, as you'll see), so it's fine as a static class.
- We insist that converter classes have a default ctor. In the book _Dependency Injection in .NET_, Mark Seeman refers to this as an anti-pattern, because it makes the classes harder to unit test, because you can't inject dependencies into their constructors.

Also:
- Add Sarif.Converters.UnitTests project to Everything.sln (it was only in Sarif.Sdk.sln).
- Change several uses of MSTest's `[ExpectedException]` attribute to Fluent Assertion's `action.ShouldThrow<T>`. This has a couple of advantages, including that you can make assertions about the resulting exception.
- Remove one copy of the file ToolFormatConverterTests.cs, which existed with identical contents in both the Sarif.UnitTests project and the Sarif.Converters.UnitTests project.
- Remove a few spurious semicolons.


src/Everything.sln
```diff
@@ -37,6 +37,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SarifCli", "SarifCli\SarifC
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SarifCli.FunctionalTests", "SarifCli.FunctionalTests\SarifCli.FunctionalTests.csproj", "{859920F3-DB09-4C1D-9EAD-DFB1D7A03A46}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sarif.Converters.UnitTests", "Sarif.Converters.UnitTests\Sarif.Converters.UnitTests.csproj", "{EDF102ED-3959-40A2-9FB1-CBD6719F77BF}"
```
lgolding: This project was included in Sarif.Sdk.sln but omitted from Everything.sln.

src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
```diff
@@ -17,73 +20,82 @@ public class ToolFormatConverterTests
...
         [TestMethod]
-        [ExpectedException(typeof(ArgumentException))]
         public void ToolFormatConverter_ConvertToStandardFormat_DirectorySpecifiedAsDestination()
         {
             string input = Path.GetTempFileName();
             string directory = Environment.CurrentDirectory;
-            _converter.ConvertToStandardFormat(ToolFormat.AndroidStudio, input, directory);
+            Action action = () => _converter.ConvertToStandardFormat(ToolFormat.AndroidStudio, input, directory);
+
+            action.ShouldThrow<ArgumentException>();
```
lgolding: Several tests now use `ShouldThrow<T>` instead of `[ExpectedException]`. It's more readable, more flexible (for example, you can say `action.ShouldThrow<T>().WithMessage("*Something*");` to assert that the exception message mentions "Something"), and improves code coverage in your unit test code (which is important for reasons we can discuss).

src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
```diff
@@ -109,5 +121,245 @@ public void ToolFormatConverter_TruncatesOutputFileInOverwriteMode()
                 Assert.AreEqual(expectedOutput, actualOutput);
             }
         }
+
```
lgolding: New tests of the plugin functionality start here.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    internal static class ConverterFactory
+    {
+        private static readonly IDictionary<string, Lazy<ToolFileConverterBase>> BuiltInConverters = CreateBuiltInConverters();
```
lgolding: Moved from ToolFormatConverter.cs, and renamed.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        internal static ToolFileConverterBase CreateConverter(string toolFormat, string pluginAssemblyPath)
+        {
+            return string.IsNullOrWhiteSpace(pluginAssemblyPath)
+                ? CreateBuiltInConverter(toolFormat)
+                : CreateConverterFromPlugin(toolFormat, pluginAssemblyPath);
+        }
+
+        private static ToolFileConverterBase CreateBuiltInConverter(string toolFormat)
+        {
```
lgolding: Moved from ToolFormatConverter.cs and extracted into a method.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,132 @@
...
+                    pluginAssemblyPath,
+                    toolFormat,
+                    typeof(ToolFileConverterBase).FullName);
+
+                throw new ArgumentException(message, nameof(pluginAssemblyPath));
+            }
+
+            return converter;
+        }
+
```
lgolding: Everything from here on down is moved from ToolFormatConverter.cs, with some renamings.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,132 @@
...
+        private static ToolFileConverterBase CreateBuiltInConverter(string toolFormat)
+        {
+            Lazy<ToolFileConverterBase> converter;
+            return BuiltInConverters.TryGetValue(toolFormat, out converter)
+                ? converter.Value
+                : null;
+        }
+
+        private static ToolFileConverterBase CreateConverterFromPlugin(string toolFormat, string pluginAssemblyPath)
+        {
```
lgolding: This method is the heart of the PR.

src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
```diff
@@ -109,5 +121,245 @@ public void ToolFormatConverter_TruncatesOutputFileInOverwriteMode()
...
+                action.ShouldThrow<ArgumentException>()
+                    .Where(ex =>
+                        ex.Message.Contains(pluginAssemblyPath)
+                        && ex.Message.Contains(ToolName));
+            }
+        }
+
+        [TestMethod]
+        public void ToolFormatConverter_FindsConverterInPluginAssembly()
```
michaelcfanning: >ToolFormatConverter_FindsConverterInPluginAssembly [](start = 20, length = 50)

File access tests generally require additional test cases to be complete. What about UnauthorizedAccessException, for example, if someone provides a path to a dll where access isn't granted? Also BadImageFormatException is another weird one that doesn't extend IOException. Speaking of which, do you have a test with illegal characters in path to produce an IOException?? Looking now. :)

michaelcfanning: No. :) So, note: provide tests that provoke IOException, BadImageFormatException, and UnauthorizedAccessException.

---
In reply to: [120773153](https://github.com/Microsoft/sarif-sdk/pull/580#discussion_r120773153) [](ancestors = 120773153)

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+            // reside in any namespace.
+            string converterTypeName = toolFormat + "Converter";
+
+            Assembly pluginAssembly = Assembly.LoadFile(pluginAssemblyPath);
+            Type[] pluginTypes = pluginAssembly
+                .GetTypes()
+                .Where(t => t.IsPublic && t.Name.Equals(converterTypeName, StringComparison.Ordinal))
+                .ToArray();
+
+            if (pluginTypes.Length == 0)
```
michaelcfanning: >pluginTypes [](start = 16, length = 11)

I like your converter identification mechanism.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -0,0 +1,131 @@
...
+                    pluginAssemblyPath);
+
+                throw new ArgumentException(message, nameof(pluginAssemblyPath));
+            }
+
+            // Convention: The converter type name is derived from the tool name. It can
+            // reside in any namespace.
+            string converterTypeName = toolFormat + "Converter";
+
+            Assembly pluginAssembly = Assembly.LoadFile(pluginAssemblyPath);
```
michaelcfanning: >LoadFile [](start = 47, length = 8)

You will run into problems here with plugins loaded in directory locations upstream, generally outside the driver exe install location. Converter dependencies may not be loaded properly from those locations. Make sense? So, someone has authored a converter that depends on a previously unloaded assembly alongside it. Won't be found. Resolution here is to enable a handler for unresolved assembly references and to use the directory of any loaded plug ins to attempt to resolve. ModernCop has some code for handling this i can point you to.

michaelcfanning:
:shipit:

PR #582, opened 2017-06-08
Plugin design improvements

lgolding:
This PR has two distinct changes (sorry):

1. Improve the design for the factories that create converter objects. The factories now form a Chain of Responsibility; we take the result from the first factory in the chain capable of creating the desired converter.

Among other things, this means that if you ask for a converter for a built-in tool format, and you also specify a plugin on the command line, it will work (it will find the built-in converter). Formerly, this would fail, because the `PluginConverterFactory` would fail to create the converter and throw an exception.

2. We take advantage of the Chain of Responsibility to support multiple plugin assemblies on the command line.

Also:
- Improve the design of the `PluginConverterFactory` by factoring all the "converter type location" logic into a single method, `IsConverterClassForToolFormat`.
- Add a check that the class is not abstract.

NOTE: This PR does _not_ implement the feedback from #580 -- I'll do that next.

Breaking: Per feedback, we removed the multiple plugin support, and added unit tests for the consistency and completeness of the `BuiltInConverterFactory`.


src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
```diff
@@ -286,12 +301,65 @@ public void ToolFormatConverter_FindsConverterInPluginAssembly()
...
                     ToolFormatConversionOptions.None,
-                    pluginAssemblyPath);
+                    pluginAssemblyPaths);
+
+                action.ShouldNotThrow();
+            }
+        }
+
+        [TestMethod]
+        public void ToolFormatConverter_FindsBuiltInConverterEvenIfPluginIsSpecified()
```
lgolding: This is the test that would previously have failed.

src/Sarif.Converters/ConverterFactory.cs
```diff
@@ -1,131 +1,32 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
```
lgolding: This class formerly combined the logic for plugins and built-in converters. We factor the logic into two new classes, `PluginConverterFactory` and `BuiltinConverterFactory`. The `ConverterFactory` class remains behind as an abstraction that implements the Chain of Responsibility. See the right-hand version; treat is as new code.

src/Sarif.Converters/ConverterResources.resx
```diff
@@ -201,19 +201,10 @@
   <data name="SemmleUnknownSeverity" xml:space="preserve">
     <value>The severity "{0}" does not correspond to a SARIF result level.</value>
   </data>
-  <data name="ErrorMissingConverterType" xml:space="preserve">
-    <value>The converter plugin assembly "{0}" does not contain a public class "{1}".</value>
```
lgolding: Remove several errors: when these things happen, the plugin factory now just returns null and delegates to the next factory in the chain.

src/ConvertToSarif/ConvertOptions.cs
```diff
@@ -43,9 +43,10 @@ internal class ConvertOptions
...
         [Option(
             'a',
-            "plugin-assembly-path",
+            "plugin-assembly-paths",
             Default = null,
-            HelpText = "Path to plugin assembly containing converter types.")]
-        public string PluginAssemblyPath { get; internal set; }
+            HelpText = "Comma-separated list of paths to plugin assemblies containing converter types.",
+            Separator = ',')]
+        public string[] PluginAssemblyPaths { get; internal set; }
```
michaelcfanning: >PluginAssemblyPaths [](start = 24, length = 19)

I thought about asking for this capability but then thought, why would we need it? is it so a developer can specify a bundle of converters and then simply change the tool type and trust it will be found? I tend to think a better model is to maintain a precise coupling between the converter and its containing assembly. Tool late now, perhaps, but my opinion right now is that this is all unnecessary machinery. Do you have a different valuable scenario in mind that I'm missing?

src/Sarif.Converters/BuiltInConverterFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+            Lazy<ToolFileConverterBase> converter;
+            return BuiltInConverters.TryGetValue(toolFormat, out converter)
+                ? converter.Value
+                : null;
+        }
+
+        private static Dictionary<string, Lazy<ToolFileConverterBase>> CreateBuiltInConverters()
+        {
+            var result = new Dictionary<string, Lazy<ToolFileConverterBase>>();
+            // TODO: Reflect to get them all.
```
michaelcfanning: >// TODO: Reflect to get them all. [](start = 12, length = 33)

Disagree. Breaks my first principle for for use reflection, if you can get away with not using it, don't use it. What I'd suggest instead is that we either create a #if DEBUG block that asserts we understand and have hard-coded for all existing built-in converters, or that we alter a unit test that fires if reflection doesn't match what is hard-coded. If you weren't using generics, it'd be easy to create an array of ToolFormat values below that unit-tests could compare with reflected values. I think you'd have to rework your API to accomplish this. Most straightforward way would be to have an api that understands ToolFormat and which has a switch statement, each case of which is away of the type of the converter to create. This might actually be nice in any case, since the way you've done it allows for decoupling between the converter type and the toolformat value...

michaelcfanning:
:clock1:

michaelcfanning:
Convince me we need this change. :)

src/Sarif.Converters/BuiltInConverterFactory.cs
lgolding: As we discussed: added unit tests for this.

---
In reply to: [120986166](https://github.com/Microsoft/sarif-sdk/pull/582#discussion_r120986166) [](ancestors = 120986166)

src/Sarif.Converters.UnitTests/BuiltinConverterFactoryTests.cs
```diff
@@ -0,0 +1,57 @@
...
+                Assert.IsTrue(
+                    BuiltInConverterFactory.BuiltInConverters.ContainsKey(toolFormat),
+                    $"There is no built-in converter for the {toolFormat} tool format, or the converter exists but is not registered in {factoryName}.");
+            }
+        }
+
+        [TestMethod]
+        public void BuiltInConverterFactory_HasMatchingConverterTypeNamesForAllRegisteredToolFormats()
+        {
+            foreach (string toolFormat in Utilities.GetToolFormats().Except(new[] { ToolFormat.None }))
```
lgolding: >.Except(new[] { ToolFormat.None }) [](start = 68, length = 34)

"Except" not needed; it's built into `Utilities.GetToolFormats`.

lgolding:
"This change" actually had several aspects; we agreed to keep all of them except multiple plugin support.

---
In reply to: [307213355](https://github.com/Microsoft/sarif-sdk/pull/582#issuecomment-307213355) [](ancestors = 307213355)

src/ConvertToSarif/ConvertOptions.cs
lgolding: As we discussed, removed this capability.

---
In reply to: [120984734](https://github.com/Microsoft/sarif-sdk/pull/582#discussion_r120984734) [](ancestors = 120984734)

src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     /// <summary>
     /// Converts a log file from the Semmle format to the SARIF format.
     /// </summary>
-    internal class SemmleConverter : ToolFileConverterBase
+    internal class SemmleQlConverter : ToolFileConverterBase
```
michaelcfanning: >SemmleQlConverter [](start = 19, length = 17)

Your casing is wrong. SemmleQL is correct. Two-letter acronyms are all caps

src/Sarif.Converters/ToolFormat.cs
```diff
@@ -31,7 +31,7 @@ public static class ToolFormat
         public const string PREfast = nameof(PREfast);
 
         /// <summary>Semmle's file format.</summary>
-        public const string SemmleQL = nameof(SemmleQL);
+        public const string SemmleQl = nameof(SemmleQl);
```
michaelcfanning: >SemmleQl [](start = 46, length = 8)

bad rename, revert. :)

src/Sarif.Converters/SemmleQlConverter.cs
lgolding: Aargh. I could have _sworn_ that wasn't the case, but I just [looked it up](https://msdn.microsoft.com/en-us/library/ms229043(v=vs.100).aspx) and you're right. Fixed.

---
In reply to: [121228572](https://github.com/Microsoft/sarif-sdk/pull/582#discussion_r121228572) [](ancestors = 121228572)

src/Sarif.Converters/ToolFormat.cs
lgolding: Fixed.

---
In reply to: [121228603](https://github.com/Microsoft/sarif-sdk/pull/582#discussion_r121228603) [](ancestors = 121228603)

PR #583, opened 2017-06-10
Update RuntimeConditions definition. Ship 1.5.47

michaelcfanning:
@lgolding 

If we set the 'command line did not parse' bit to 1, both for the 'rich' return case and the non-rich return case, this consistency is valuable, particularly if a user misenters the command-line verb. When this happens, it is difficult for us to know whether the intended verb even supports the --rich-return-code. We can check ourselves for this arg via System.Environment.CommandLine but the presence of --rich-return-code itself could be a mistake. 

What we do now is set the standard error result (1) to the same 'rich' return code bit 0x1. For this case, the behavior is the same no matter whether --rich-return-code was intended or correct.

This subtlety was uncovered while troubleshooting a Binskim issue.


src/Sarif/RuntimeConditions.cs
```diff
@@ -26,10 +26,15 @@ public enum RuntimeConditions : uint
...
-        ExceptionInstantiatingSkimmers = 0x01,
-        ExceptionInSkimmerInitialize = 0x02,
-        ExceptionRaisedInSkimmerCanAnalyze = 0x04,
-        ExceptionInSkimmerAnalyze = 0x08,
+        // InvalidCommandLineOption is a useful bit to 
+        // use as value 0x1. This allows for a consistent 
+        // return value in cases where a command-line argument
+        // for --rich-return-code can't be detected because it 
+        // fails to parse. 
+        InvalidCommandLineOption = 0x1,
```
lgolding: >InvalidCommandLineOption [](start = 8, length = 24)

The argument makes sense. Are you comfortable you're not breaking anyone who might be relying on a rich return code where 0x01 meant "ExceptionInstantiatingSkimmers"?

Wildly unlikely, and I'll approve -- just wanted to make sure you considered it.

lgolding:
:shipit:

PR #585, opened 2017-06-12
Fix #584: GetProperty<T> should work for string properties.

### LINKED ISSUE #584: GetProperty<T> should work for string properties
As designed, `PropertyBagHolder.GetProperty<T>` throws an exception if you call it with a property name whose value is a string. The exception message tells you to call the non-generic method `PropertyBagHolder.GetProperty` instead. But there are cases where you don't know a property's type ahead of time (for example, when displaying the properties in a UI). There's no way to ask for a property's type ahead of time so that you can call the right `GetProperty` method, and even if there were, you shouldn't have to do that.

So remove the restriction that `GetProperty<T>` doesn't work on string properties.

lgolding:
As designed, `PropertyBagHolder.GetProperty<T>` throws an exception if you call it with a property name whose value is a string. The exception message tells you to call the non-generic method `PropertyBagHolder.GetProperty` instead. But there are cases where you don't know a property's type ahead of time (for example, when displaying the properties in a UI). There's no way to ask for a property's type ahead of time so that you can call the right `GetProperty` method, and even if there were, you shouldn't have to do that.

So remove the restriction that `GetProperty<T>` doesn't work on string properties.

### DIFF
--- a/src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs
+++ b/src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs
@@ -142,6 +142,16 @@ namespace Microsoft.CodeAnalysis.Sarif.Core
         }
 
         [TestMethod]
+        public void PropertyBagHolder_GetPropertyOfT_WorksForStringProperties()
+        {
+            var inputObject = new TestClass();
+            inputObject.SetProperty(PropertyName, "x");
+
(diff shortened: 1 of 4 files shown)


lgolding:

lgolding:

michaelcfanning:
:shipit:

PR #591, opened 2017-08-21
Resolve PR notes and add fix for embedded file:// references in Semmle csv

michaelcfanning:
PR notes + fix for embedded file:// references in Semmle CSV. @lgolding

### DIFF
--- a/src/Sarif.Converters/SemmleQlConverter.cs
+++ b/src/Sarif.Converters/SemmleQlConverter.cs
@@ -16,7 +16,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     /// </summary>
     public class SemmleQLConverter : ToolFileConverterBase
     {
-        // Semmle logs are CSV files with inconsistent contents
+        // Semmle logs are CSV files
         private static readonly string[] s_delimiters = new[] { "," };
 
         // The fields are as follows:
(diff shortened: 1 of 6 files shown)


src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -187,10 +187,26 @@ private IList<AnnotatedCodeLocation> NormalizeRawMessage(string rawMessage, out
                     tokens = location.Split(':');
 
                     relatedLocations = relatedLocations ?? new List<AnnotatedCodeLocation>();
+                    PhysicalLocation physicalLocation;
 
-                    var relatedLocation = new AnnotatedCodeLocation
+                    if (tokens[0].Equals("file", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // special case for file paths
```
lgolding: >s [](start = 27, length = 1)

S

lgolding: >s [](start = 53, length = 1)

s.

lgolding:
>                     tokenText = tokens[0];

Naming: IMO if this were named `linkText` the comment would be unnecessary.

---
Refers to: src/Sarif.Converters/SemmleQlConverter.cs:184 in d719883. [](commit_id = d7198834a0e788368e6b649b8509e188a5ee0fb5, deletion_comment = False)

lgolding:
>                     string[] tokens = embeddedLink.Split(new char[] { '\"' }, StringSplitOptions.RemoveEmptyEntries);

I'm not wild about this (existing) code for a couple of reasons.

1. There are libraries to parse CSV. This code, which splits on double-quotes, and then has to take account of the fact that between every column value there is a comma separator (which is why we look at `tokens[0]` and `tokens[2]`, but skip `tokens[1]`, which is the comma) is not pretty (and I'm not even sure it always works).

2. I don't like the reuse of the variable `tokens`. It took me a minute to realize that it first held the quote-separated pieces, and then the colon-separated pieces.

No action. Just venting.

---
Refers to: src/Sarif.Converters/SemmleQlConverter.cs:181 in d719883. [](commit_id = d7198834a0e788368e6b649b8509e188a5ee0fb5, deletion_comment = False)

src/Sarif.Converters/SemmleQlConverter.cs
michaelcfanning: gotcha. created a new var 'locationTokens'. will defer the CSV parsing, we'll do that later.

---
In reply to: [134355305](https://github.com/Microsoft/sarif-sdk/pull/591#discussion_r134355305) [](ancestors = 134355305)

michaelcfanning:
>                     tokenText = tokens[0];

why introduce a new var? changed it to just reassign to 'embeddedLinksText' a nicely named thing.

---
In reply to: [323879490](https://github.com/Microsoft/sarif-sdk/pull/591#issuecomment-323879490) [](ancestors = 323879490)

---
Refers to: src/Sarif.Converters/SemmleQlConverter.cs:184 in d719883. [](commit_id = d7198834a0e788368e6b649b8509e188a5ee0fb5, deletion_comment = False)

src/Sarif.Converters/SemmleQlConverter.cs
lgolding: Consider naming the first one `csvTokens`. For symmetry. It feels wrong to have one generic name and one specific name.

---
In reply to: [134357737](https://github.com/Microsoft/sarif-sdk/pull/591#discussion_r134357737) [](ancestors = 134357737,134355305)

lgolding:
>                     tokenText = tokens[0];

Because then again you're using one variable to mean two different things, and the reader has to keep track of what it means over time.

---
In reply to: [323882240](https://github.com/Microsoft/sarif-sdk/pull/591#issuecomment-323882240) [](ancestors = 323882240,323879490)

---
Refers to: src/Sarif.Converters/SemmleQlConverter.cs:184 in d719883. [](commit_id = d7198834a0e788368e6b649b8509e188a5ee0fb5, deletion_comment = False)

PR #595, opened 2017-08-24
Fix #597: Crash when selecting binary files.

### LINKED ISSUE #597: SARIF VSIX crashes when navigating to a result found in a binary file
1. Open attached file binskim.sarif in Visual Studio 2017: Tools, Open Static Analysis Log File, Open SARIF File.
2. Double-click any SARIF result in error list.

Expected: It shouldn't crash. We need to decide whether it
Silently does nothing.
Displays a message saying you can't navigate to a location in a binary file.
Opens the file in the binary editor and places the cursor at the appropriate offset.
Something else?
Actual: It crashes.

vinaykapadia:


src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
+{
+    public class SarifErrorListItemTests
+    {
+        [Fact]
+        public void SarifErrorListItem_LineMarker_Region_StartLine()
+        {
+            new SarifViewerPackage();
```
lgolding: >SarifViewerPackage [](start = 16, length = 18)

Is this something you saw in other unit tests? What does it do? It must have a side effect, since you're discarding the resulting object. Does the test work without this line?

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
+{
+    public class SarifErrorListItemTests
+    {
+        [Fact]
+        public void SarifErrorListItem_LineMarker_Region_StartLine()
```
lgolding: >LineMarker_Region_StartLine [](start = 39, length = 27)

Our convention is to have test names describe what the test demonstrates. This name, in contrast, describes the input data ("Test what happens when a region has a start line").

I suggest:
```
SarifErrorListItem_WhenRegionHasStartLine_HasLineMarker
SarifErrorListItem_WhenRegionHasNoStartLine_HasNoLineMarker
SarifErrorListItem_WhenRegionIsAbsent_HasNoLineMarker
```

lgolding:
In VSTS, I would have marked this PR as "Approved with suggestions" -- it's good code and I'd be fine with you making the changes and merging without further review.

But GitHub PRs don't give me that option. I'm marking it "Code not ready" just to make absolutely sure you see that I made some comments.

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
vinaykapadia: The constructor sets a static variable in the class, which is used in the method I'm testing. Without it, it throws a null reference exception.

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
vinaykapadia: Will change these.

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
lgolding: In that case:
1. Introduce a parameterless constructor:
```
public SarifErrorListItem()
{
    new SarifViewerPackage();
}
```
xUnit will create a new instance to run each test, so effectively this becomes "common test startup code".
2. Add a comment to the ctor explaining why this is needed.

---
In reply to: [135291088](https://github.com/Microsoft/sarif-sdk/pull/595#discussion_r135291088) [](ancestors = 135291088)

vinaykapadia: The SarifErrorListItem class already has a parameterless constructor. My issue is that I don't want to declare a "new SarifViewerPackage()" in the actual project code, since I don't want to overwrite anything that is done elsewhere. I only need it to initialize one field for these tests (and that field is not even used during the test, its just read and set into a member of another class). Not sure how we should proceed.

---
In reply to: [135293512](https://github.com/Microsoft/sarif-sdk/pull/595#discussion_r135293512) [](ancestors = 135293512,135291088)

lgolding: My mistake. I meant to add a parameterless ctor for `SarifErrorListItemTests`!

---
In reply to: [135297993](https://github.com/Microsoft/sarif-sdk/pull/595#discussion_r135297993) [](ancestors = 135297993,135293512,135291088)

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
+{
+    public class SarifErrorListItemTests
+    {
+        /// <summary>
+        /// Constructor required to set a static member of the SarifViewerPackage class.
```
lgolding: MikeFan has a pet peeve about doc comments :-) Just make this a regular comment.

lgolding: Also I don't think we need to comment the ctor itself; just the `new` line -- how about this, a single comment on the `new` line:

Creating a SarifViewerPackage object has the side effect of setting a static variable in that class, without which the subsequent tests will fail. In production code, the package object is always created before any SarifErrorListItem objects.

---
In reply to: [135302513](https://github.com/Microsoft/sarif-sdk/pull/595#discussion_r135302513) [](ancestors = 135302513)

PR #596, opened 2017-08-24
Improved VSIX Performance for Large Files

jeff-bacon:
File name / project name pairs are now cached to prevent multiple lookups which was causing poor performance.


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -76,11 +76,8 @@ private static void ProcessSarifLog(SarifLog sarifLog, string logFilePath)
 
         private ErrorListService()
         {
-            this.documentToLineIndexMap = new Dictionary<string, NewLineIndex>();
```
lgolding: >documentToLineIndexMap [](start = 17, length = 22)

Unused, I see. OK.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Sarif.Viewer
+{
+    public sealed class ProjectNameLookup
+    {
+        private readonly static ProjectNameLookup _instance = new ProjectNameLookup();
+        private Dictionary<string, string> _projectNames = new Dictionary<string, string>();
+
+        static ProjectNameLookup() { }
+        private ProjectNameLookup() { }
```
lgolding: Not needed.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public static ProjectNameLookup Instance => _instance;
+
+        public void SetName(string fileName)
+        {
+            if (_projectNames.ContainsKey(fileName))
+                return;
```
lgolding: >return [](start = 16, length = 6)

Our convention, as is the case with many MS teams, is always to use the braces:
```
if (condition)
{
    return;
}
```

Same with the if/else below.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            var project = SarifViewerPackage.Dte.Solution.FindProjectItem(fileName);
+            if (project?.Properties != null && project?.ContainingProject != null)
```
lgolding: >Properties [](start = 25, length = 10)

Why is it necessary to test `Properties`? You don't seem to use it in the subsequent code.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Sarif.Viewer
+{
+    public sealed class ProjectNameLookup
+    {
+        private readonly static ProjectNameLookup _instance = new ProjectNameLookup();
+        private Dictionary<string, string> _projectNames = new Dictionary<string, string>();
```
lgolding: >_projectNames [](start = 43, length = 13)

We don't use the leading underscore convention (actually I like that convention, but I can't have everything I want...)

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        static ProjectNameLookup() { }
+        private ProjectNameLookup() { }
+
+        public static ProjectNameLookup Instance => _instance;
```
lgolding: >Instance [](start = 40, length = 8)

Do we need to worry about thread safety? Do we have to do this?

```
private readonly thisLock = new object();
...
public static ProjectNameLookup Instance
{
    get
    {
        lock(thisLock)
        {
            return instance;
        }
    }
}
```

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Sarif.Viewer
+{
+    public sealed class ProjectNameLookup
```
lgolding: >ProjectNameLookup [](start = 24, length = 17)

I'm not crazy about this name, since class names are usually nouns, and this doesn't look like a noun.

What do you think of `ProjectNameCache`?

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
jeff-bacon: This check (project.Properties != null) existed in the previous code.  I wasn't sure of repercussions of removing, so I left it in.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -129,10 +120,6 @@ public override bool TryGetValue(int index, string columnName, out object conten
                     {
                         url = error.HelpLink;
                     }
-                    else
-                    {
-                        //url = string.Format("http://www.bing.com/search?q={0} {1}", _errors[index].Provider.Name, _errors[index].ErrorCode);
```
lgolding: >//url [](start = 24, length = 5)

Thanks for the cleanup!

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
lgolding: Please remove it. If it breaks, we'll figure out why, and, at the very least, we'll be able to add a comment that explains _why_ this is needed.

---
In reply to: [135284322](https://github.com/Microsoft/sarif-sdk/pull/596#discussion_r135284322) [](ancestors = 135284322)

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
jeff-bacon: I believe the private constructor must be there to instantiate the singleton.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
jeff-bacon: Easy fix!  I'll get that put in.

lgolding:
Before you merge, please change the title of the PR. The title is currently "Users/v jebac/vsix performance". Change to something descriptive like "Improve VSIX performance on large files".

The PR title ends up in the history, so we want to make it as informative as possible.

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
lgolding: I was referring to the `static` constructor.

The `private` constructor is indeed part of the Singleton pattern. It's not there to allow the Singleton to be instantiated; it's there to _prevent_ anyone from creating additional instances by writing `new Foo()`.

---
In reply to: [135286787](https://github.com/Microsoft/sarif-sdk/pull/596#discussion_r135286787) [](ancestors = 135286787)

jeff-bacon: I was going for the fourth implementation found here: [http://csharpindepth.com/Articles/General/Singleton.aspx](http://csharpindepth.com/Articles/General/Singleton.aspx).  Classes with a static constructor apparently get their type initializers invoked only once per app domain which prevents us from having to use a null check or lock to ensure thread safety.  More information on this can be found here:  [http://csharpindepth.com/Articles/General/BeforeFieldInit.aspx](http://csharpindepth.com/Articles/General/BeforeFieldInit.aspx).  Do you think we should change this to use a lock instead?

---
In reply to: [135287508](https://github.com/Microsoft/sarif-sdk/pull/596#discussion_r135287508) [](ancestors = 135287508,135286787)

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
jeff-bacon: What happens if two threads try to enter SetName around the same time?  Would it run the FindProjectItem method numerous times for the same fileName?  If so, would that be cheaper performance wise than taking out a lock every time we call the method?  What are your thoughts?

---
In reply to: [135165342](https://github.com/Microsoft/sarif-sdk/pull/596#discussion_r135165342) [](ancestors = 135165342)

src/Sarif.Viewer.VisualStudio/ProjectNameLookup.cs
lgolding: I did not know about this! No, keep what you have, it is fine.

---
In reply to: [135301008](https://github.com/Microsoft/sarif-sdk/pull/596#discussion_r135301008) [](ancestors = 135301008,135287508,135286787)

PR #600, opened 2017-08-28
Enable Errors Without File Location

jeff-bacon:


src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
```diff
@@ -125,9 +125,7 @@ public void AddErrors(IEnumerable<SarifErrorListItem> errors)
 
             CleanAllErrors();
 
-            var cleanErrors = errors.Where(e => e != null && !string.IsNullOrEmpty(e.FileName));
```
jeff-bacon: Errors will never be null and we want to keep errors without a FileName

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -414,7 +414,7 @@ public ResultTextMarker LineMarker
 
         internal void RemapFilePath(string originalPath, string remappedPath)
         {
-            if (FileName != null && FileName.Equals(originalPath, StringComparison.OrdinalIgnoreCase))
```
jeff-bacon: FileName can no longer be null.

src/Sarif.Viewer.VisualStudio/Sarif/Result.extensions.cs
```diff
@@ -42,7 +42,7 @@ public static string GetPrimaryTargetFile(this Result result)
         {
             if (result == null || result.Locations == null || result.Locations.Count == 0)
             {
-                return null;
+                return string.Empty;
```
jeff-bacon: This change was required because the FileName is used as a dictionary key.  Previously this didn't matter because errors without a location were never used in this dictionary.

lgolding: IMO it should still return `null` if `result` is `null`. The error list won't contain a `null`, so we don't have to worry about this being used as a key.

---
In reply to: [135596586](https://github.com/Microsoft/sarif-sdk/pull/600#discussion_r135596586) [](ancestors = 135596586)

PR #603, opened 2017-08-28
VSIX Unit Test STA Error

jeff-bacon:


src/Sarif.Viewer.VisualStudio/SarifToolWindow.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Sarif.Viewer
     /// </para>
     /// </remarks>
     [Guid("ab561bcc-e01d-4781-8c2e-95a9170bfdd5")]
-    public class SarifToolWindow : ToolWindowPane
+    public class SarifToolWindow : ToolWindowPane, IToolWindow
```
jeff-bacon: This is being called by other classes that are being used in UnitTests.  Implement IToolWindow to enable this class to be mocked.

src/Sarif.Viewer.VisualStudio/Models/CallTree.cs
```diff
@@ -15,10 +15,12 @@ public class CallTree : NotifyPropertyChangedObject
         DelegateCommand<TreeView> _selectPreviousCommand;
         DelegateCommand<TreeView> _selectNextCommand;
 
+        private IToolWindow toolWindow;
         private ObservableCollection<CallTreeNode> _topLevelNodes;
 
-        public CallTree(IList<CallTreeNode> topLevelNodes)
+        public CallTree(IList<CallTreeNode> topLevelNodes, IToolWindow toolWindow)
```
jeff-bacon: Take a IToolWindow instead of using the concrete class SarifToolWindow below

lgolding:
OMG, _so_ good!

PR #608, opened 2017-08-28
Fix #604: highlighting multiple runs

### LINKED ISSUE #604: File with Multiple Runs does not Highlight Properly
Open SARIF file with multiple runs.
Only the last run will have errors that properly highlight.

Error is in ErrorListService.cs line 97. Should add to collection, not overwrite.

vinaykapadia:


lgolding:
Code is fine; need unit tests.

michaelcfanning:
product change LGTM 
+1 tests

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
```diff
@@ -29,20 +29,20 @@ class SarifTableDataSource : ITableDataSource
         private SarifTableDataSource()
         {           
             var compositionService = ServiceProvider.GlobalProvider.GetService(typeof(SComponentModel)) as IComponentModel;
-            compositionService.DefaultCompositionService.SatisfyImportsOnce(this);
+            compositionService?.DefaultCompositionService.SatisfyImportsOnce(this);
```
lgolding: >compositionService [](start = 12, length = 18)

Rather than repeatedly testing the result, how about:
```
// The composition service will only be null in unit tests.
if (compositionService != null)
{
    // Do everything else.
}
```

PR #611, opened 2017-08-30
Save file contents from SARIF file.

vinaykapadia:


src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Text;
+
+namespace Microsoft.Sarif.Viewer.Models
+{
+    public class FileDetailsModel
+    {
+        /// <summary>
```
lgolding: >summary [](start = 13, length = 7)

Sorry Vinay, mikefan hates doc comments because he finds them less readable. Please use regular comments. (I don't actually agree with him :-)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -12,6 +12,7 @@
 
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.CodeAnalysis.Sarif.Writers;
+using Microsoft.Sarif.Viewer.Models;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.Win32;
 using Microsoft.Sarif.Viewer.Sarif;
```
michaelcfanning: >Sarif [](start = 16, length = 5)

sort and remove unused namespaces on this file

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -87,6 +90,23 @@ private void WriteRunToErrorList(Run run, string logFilePath, Solution solution)
...
+            if (run.Files != null)
+            {
+                foreach (var file in run.Files)
+                {
+                    var fileHash = file.Value.Hashes?.First(x => x.Algorithm == AlgorithmKind.Sha256)?.Value;
+                    var contents = file.Value.Contents;
+                    var key = new Uri(file.Key);
+                    if (fileHash != null && contents != null)
+                    {
+                        var fileDetails = new FileDetailsModel(fileHash, file.Value.Contents);
```
michaelcfanning: >FileDetailsModel [](start = 46, length = 16)

why doesn't this class accept a FileData instance in the ctor? Seems like breaking apart members will just leave us in a position to keep updating this code.

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
vinaykapadia: heh, forgot about that. will fix.

michaelcfanning: As typical w/someone who disagrees with my position, Larry fails to represent it in a nuanced way. :)

Doc comments are less readable than straight comments. And so, I avoid them on non-public things, like this private field. I also don't believe in useless comments (this one here is actually good). 

My rule is this: if you can't demonstrate that a doc comment is viewable outside reading the direct source file (i.e., if you can't get to it via IntelliSense) don't use it. Use a standard comment. Take all the energy you save from authoring structured XML to make the comment as good as possible.,


---
In reply to: [136137221](https://github.com/Microsoft/sarif-sdk/pull/611#discussion_r136137221) [](ancestors = 136137221)

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Contents of file. May or may not be Base64 encoded.
+        /// </summary>
+        private string _contents;
+
+        /// <summary>
+        /// Tells whether or not the _contents are Base64 encoded.
+        /// </summary>
+        private bool _encoded;
```
michaelcfanning: >_encoded [](start = 21, length = 8)

if you name this bool 'isBase64Encoded' you won't need to comment it.

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            Hash = hash;
+            _contents = contents;
+            _encoded = true;
+        }
+
+        /// <summary>
+        /// SHA256 hash for file.
+        /// </summary>
+        public string Hash { get; }
```
michaelcfanning: >Hash [](start = 22, length = 4)

If you name this Sha256Hash, you don't need a comment

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Returns the decoded file contents.
+        /// </summary>
+        public string Contents
+        {
+            get
+            {
+                // If the contents are encoded, decode them.
+                if (_encoded)
+                {
+                    DecodeContents();
```
michaelcfanning: >DecodeContents [](start = 20, length = 14)

Decoding is a non-trivial operation and so typically wouldn't be a good candidate for a property accessor. If you accessed this member in a debug watch window, for example, you will change the state of the object. Ouch.

A GetContents method would therefore be preferable.

Now, you could also consider decoding on instantiating this object, and save a lot of bother. If that has adverse performance impact, however, not a good idea (but if in practical use we can't defer decoding costs, we should consider it).

If my suggestion to prefer a GetContents method causes issues in WPF binding, that's a good reason to ignore my comment, so feel free to push back.

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -0,0 +1,62 @@
...
+                return _contents;
+            }
+        }
+
+        /// <summary>
+        /// Decodes the contents.
+        /// </summary>
+        private void DecodeContents()
+        {
+            var data = Convert.FromBase64String(_contents);
```
michaelcfanning: >var [](start = 12, length = 3)

Please specify this data type explicitly. We typically only use var when the type is actually detectable by reading code. Devs who don't know the return type of this method will have to go look it up. Is is a string? Is it a byte[]? it's reasonable to wonder.

src/Sarif.Viewer.VisualStudio.UnitTests/TestUtilities.cs
```diff
@@ -0,0 +1,24 @@
+﻿
+using Microsoft.CodeAnalysis.Sarif;
```
michaelcfanning: need copyright here and in other added files

src/Sarif.Viewer.VisualStudio.UnitTests/TestUtilities.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Sarif.Viewer.ErrorList;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
+{
+    public static class TestUtilities
+    {
+        public static void InitializeTestEnvironment()
+        {
+            // Creating a SarifViewerPackage object has the side effect of setting a static
```
michaelcfanning: >Creating a SarifViewerPackage [](start = 15, length = 29)

would it be less expensive to just expose those static items from the package and force init? or are complicated things constructed that would make this difficult?

src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/MultipleRunsPerSarifTests.cs
```diff
@@ -75,7 +73,7 @@ public ErrorListTests()
                 }
             };
 
-            ErrorListService.ProcessSarifLog(testLog, "TestPath.sarif", null);
+            TestUtilities.InitializeTestEnvironment(testLog);
```
michaelcfanning: >InitializeTestEnvironment [](start = 26, length = 25)

Seems like this helper should accept the file path arg?

vinaykapadia: Not sure. I don't actually use a file called "TestPath.sarif", the only reason that's there is because the ProcessSarifLog takes it in. I don't use the value in any tests. Is there value in having this specified, or should I just change it to "" instead?

michaelcfanning: investigate for future. sounds like it isn't needed at all. 

---
In reply to: [136143640](https://github.com/Microsoft/sarif-sdk/pull/611#discussion_r136143640) [](ancestors = 136143640)

PR #612, opened 2017-08-30
Persist file contents

michaelcfanning:
I plan to ship after adding this functionality, so that we can modify some standard tools that emit SARIF. I will also take a follow-on change to provide this option to the conversion tool. That's a non-trivial change that will give us significantly more test files for the viewer.

@lgolding @vinaykapadia @jeff-bacon


src/Sarif/Core/FileData.cs
```diff
@@ -42,9 +66,18 @@ public static FileData Create(Uri uri, bool computeHashes)
                                 Algorithm = AlgorithmKind.Sha256,
                             },
                         };
+                }
+
             }
+            catch (Exception e) when (e is IOException || e is UnauthorizedAccessException) { }
```
michaelcfanning: >UnauthorizedAccessException [](start = 63, length = 27)

Someone tell me how to mock File.ReadAllText to raise this exception. It's hard to contrive this case from .NET, need an AppDomain that both creates a file, then cleans it up later.

src/Sarif/Core/FileData.cs
```diff
@@ -12,19 +15,40 @@ namespace Microsoft.CodeAnalysis.Sarif
...
-        public static FileData Create(Uri uri, bool computeHashes)
+        public static FileData Create(Uri uri, SarifWriters.LoggingOptions loggingOptions)
         {
             if (uri == null) { throw new ArgumentNullException(nameof(uri)); }
 
             var fileData = new FileData()
             {
-                MimeType = SarifWriters.MimeType.DetermineFromFileExtension(uri)
+                MimeType = SarifWriters.MimeType.DetermineFromFileExtension(uri),
+                Uri = uri
```
michaelcfanning: >Uri [](start = 16, length = 3)

ouch, we never set the URI before. not sure yet why this wasn't a problem. Unit testing FTW

src/Sarif/Core/FileData.cs
lgolding: As we discussed: See `MakeMockFileSystem`

---
In reply to: [136142197](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136142197) [](ancestors = 136142197)

src/Sarif/ExtensionMethods.cs
```diff
@@ -6,11 +6,19 @@
...
 using System.IO;
+using System.Xml;
+
+using Microsoft.CodeAnalysis.Sarif.Writers;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        public static bool IsSet(this LoggingOptions loggingOptions, LoggingOptions otherLoggingOptions)
```
lgolding: >IsSet [](start = 27, length = 5)

I think `Includes` reads better:
```
if (loggingOptions.Include(LoggingOptions.PrettyPrint)
{
     ...
}
```

src/Sarif/ExtensionMethods.cs
```diff
@@ -336,5 +344,105 @@ public static string GetFirstSentence(string text)
             }
             return text.TrimEnd('.') + ".";
         }
+
+        /// <summary>Retrieves a property value if it exists, or null.</summary>
```
lgolding: >Retrieves [](start = 21, length = 9)

Dude! You have repeatedly busted my chops about doc comments ("Don't use"). I'm happy to have you use them. Please don't change this.

src/Sarif/ExtensionMethods.cs
```diff
@@ -336,5 +344,105 @@ public static string GetFirstSentence(string text)
...
+        /// <param name="properties">A properties object from which the property shall be
+        /// retrieved, or null.</param>
+        /// <param name="key">The property name / key.</param>
+        /// <returns>
+        /// If <paramref name="properties"/> is not null and an entry for the supplied key exists, the
+        /// value associated with that key; otherwise, null.
+        /// </returns>
+        internal static string PropertyValue(this Dictionary<string, string> properties, string key)
```
lgolding: >PropertyValue [](start = 31, length = 13)

Having trouble seeing the value of a wrapper around `TryGetValue`. Perhaps it will be clear when I see it used...

src/ReleaseHistory.md
```diff
@@ -145,3 +145,7 @@
 ## **v1.5.47** [Driver](https://www.nuget.org/packages/Sarif.Driver/1.5.47) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/1.5.47)
  * Enable converter plugins
  * Adjust RuntimeConditions enum so that 'command line parse' error is 0x1. 
+
+## **v1.6.0** [Driver](https://www.nuget.org/packages/Sarif.Driver/1.6.0) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/1.6.0)
+ * Enable persistence of base64-encoded file contents via SarifLogger.
+ * Rename AnalyzeOptions.ComputeTargetsHash to ComputeFileHashes
```
lgolding: And:

* Fix bug in Semme file conversion.

src/Sarif/ExtensionMethods.cs
```diff
@@ -336,5 +344,105 @@ public static string GetFirstSentence(string text)
...
+        /// </summary>
+        /// <param name="array">The array to check.</param>
+        /// <param name="startIndex">The start index in the array to check.</param>
+        /// <param name="target">Target string to look for in the array.</param>
+        /// <returns>
+        /// true if the range [<paramref name="startIndex"/>, <paramref name="startIndex"/> +
+        /// <paramref name="target"/><c>.Length</c>) is equal to
+        /// <paramref name="target"/>. If the range is undefined in the bounds of the array, false.
+        /// </returns>
+        internal static bool ArrayMatches(char[] array, int startIndex, string target)
```
lgolding: >char[] array [](start = 42, length = 12)

This one isn't an extension method; intentional?

src/Sarif/ExtensionMethods.cs
lgolding: I see that you combined two files, Extensions.cs and ExtensionMethods.cs. That's fine.

Are there _any other changes_ in the migrated code that I should look at?

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -81,5 +81,10 @@ public abstract class AnalyzeOptionsBase
             "rich-return-code",
             HelpText = "Emit a 'rich' return code consisting of a bitfield of conditions (as opposed to 0 or 1 indicating success or failure.")]
         public bool RichReturnCode { get; set; }
+
+        [Option(
+            "persist-file-contents",
```
lgolding: `'f'` is available for the short name.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -81,5 +81,10 @@ public abstract class AnalyzeOptionsBase
...
+            HelpText = "Persist a base64-encoded version of any referenced file to the log.")]
```
lgolding: >any referenced file [](start = 60, length = 19)

"all referenced files"

lgolding: >version [](start = 49, length = 7)

"representation"

src/Sarif.UnitTests/ExtensionsTests.cs
```diff
@@ -38,31 +38,31 @@ public void Extensions_PropertValue_SetValue()
         [TestMethod]
         public void Extensions_IsNewline_CarriageReturn()
         {
-            Assert.IsTrue(Extensions.IsNewline('\r'));
+            Assert.IsTrue(ExtensionMethods.IsNewline('\r'));
```
lgolding: >ExtensionMethods.IsNewline('\r') [](start = 26, length = 32)

Urk! Make this an actual extension method, so you can write

if (ch.IsNewline())
{
    .
}
```

File: src/ReleaseHistory.md (hunk 6 above)
Comment (michaelcfanning) [reply src/ReleaseHistory.md]: fixed

---
In reply to: [136144420](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136144420) [](ancestors = 136144420)

File: src/Sarif.UnitTests/ExtensionsTests.cs (hunk 10 above)
Comment (lgolding) [inline src/Sarif.UnitTests/ExtensionsTests.cs]: >Assert.IsTrue [](start = 12, length = 13)

Since you're in this file, let's take the opportunity to introduce FluentAssertions (just the file you're touching).

File: src/Sarif/ExtensionMethods.cs (hunk 5 above)
Comment (michaelcfanning) [reply src/Sarif/ExtensionMethods.cs]: no

---
In reply to: [136150453](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136150453) [](ancestors = 136150453)

File: src/Sarif.UnitTests/ExtensionsTests.cs (hunk 10 above)
Comment (lgolding) [reply src/Sarif.UnitTests/ExtensionsTests.cs]:
```
'\r'.IsNewline().Should().BeTrue();
```

How nice is that?

---
In reply to: [136153377](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136153377) [](ancestors = 136153377)

Comment (michaelcfanning) [reply src/Sarif.UnitTests/ExtensionsTests.cs]: what the heck, why wasn't this an extension method?? :)

---
In reply to: [136153212](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136153212) [](ancestors = 136153212)

**Comment by lgolding** [comment]
>     [TestClass]

I see that you added xUnit to the NuGet packages for this project, but you're still using VS Test attributes.

Assuming you go the rest of the way and convert completely to xUnit, don't forget to modify BuildAndTest.cmd to run this assembly with xunit.runner.console instead of mstest.exe.

Actually you can do both, running xunit.runner.console will run the xUnit tests, and mstest.exe will find the VS tests, if you don't want to change them all at once.

---
Refers to: src/Sarif.UnitTests/ExtensionsTests.cs:13 in c37bf43. [](commit_id = c37bf43d4f4b9a4aa86f665b51b07107e6140818, deletion_comment = False)

**Comment by michaelcfanning** [comment]
>     [TestClass]

right there's a mix. good call to make sure script works.

we will shortly convert everything to xunit.

---
In reply to: [326080766](https://github.com/Microsoft/sarif-sdk/pull/612#issuecomment-326080766) [](ancestors = 326080766)

---
Refers to: src/Sarif.UnitTests/ExtensionsTests.cs:13 in c37bf43. [](commit_id = c37bf43d4f4b9a4aa86f665b51b07107e6140818, deletion_comment = False)

File: src/Sarif.UnitTests/Core/PropertyBagHolderTests.cs [hunk 11]
```diff
@@ -9,7 +9,7 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
-namespace Microsoft.CodeAnalysis.Sarif.Core
+namespace Microsoft.CodeAnalysis.Sarif
```
lgolding: >Sarif [](start = 33, length = 5)

PropertyBagHolder is in Core.

IMO the convention should be that tests for classes in namespace X.Y.Z should be in namespace X.Y.Z.Tests, but I don't know if we've done that.

TL;DR, why bother removing this?

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.CodeAnalysis.Sarif.Writers;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class FileDataTests
+    {
+        [Fact]
+        public void FileData_Create_NullUri()
```
lgolding: >FileData_Create_NullUri [](start = 20, length = 23)

I want to advocate for a naming convention where a test name states what it tests. Then the list of tests reads like a specification. (That's exactly what BDD does: you write a list of tests that together constitute a specification).

So for example:
```
FileData_Create_WithNullUri_ThrowsArgumentNullException
```

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            Action fileDataCreation = () => { FileData.Create(null, LoggingOptions.None); };
```
lgolding: >fileDataCreation [](start = 19, length = 16)

Nothing really wrong with this. I tend to just always use the name `action` in this context.

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            fileDataCreation.ShouldThrow<ArgumentNullException>();
+        }
+
+        [Fact]
+        public void FileData_ComputeHashes()
+        {
+            string filePath = Path.GetTempFileName();
```
lgolding: >GetTempFileName [](start = 35, length = 15)

I agree with getting this in so Vinay can make progress. Please right away go back and use the mock file system here, in a separate PR.

src/Sarif.UnitTests/Core/FileDataTests.cs
michaelcfanning: ok. you win. let's update a standards doc somewhere and scrub later.

---
In reply to: [136156139](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136156139) [](ancestors = 136156139)

src/Sarif.UnitTests/Core/FileDataTests.cs
michaelcfanning: like it

---
In reply to: [136156396](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136156396) [](ancestors = 136156396)

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                foreach (Hash hash in fileData.Hashes)
+                {
+                    switch (hash.Algorithm)
+                    {
+                        case AlgorithmKind.MD5: { hash.Value.Should().Be(hashes.MD5); break; }
+                        case AlgorithmKind.Sha1: { hash.Value.Should().Be(hashes.Sha1); break; }
+                        case AlgorithmKind.Sha256: { hash.Value.Should().Be(hashes.Sha256); break; }
+                    }
+                }
```
lgolding: >} [](start = 16, length = 1)

Test that the array of hashes actually does contain one of each algorithm.

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            string filePath = Path.GetTempFileName();
+            string fileContents = Guid.NewGuid().ToString();
+            Uri uri = new Uri(filePath);
+
+            try
+            {
+                File.WriteAllText(filePath, fileContents);
+                FileData fileData = FileData.Create(uri, LoggingOptions.ComputeFileHashes);
+                fileData.Uri.Should().Be(uri.ToString());
```
lgolding: >; [](start = 56, length = 1)

For symmetry with next test, assert that the file contents were _not_ persisted.

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+            finally
+            {
+                if (File.Exists(filePath)) { File.Delete(filePath); }
+            }
+        }
+
+        [Fact]
+        public void FileData_PersistFileContents()
```
lgolding: Need a test for binary files. The spec says those are just plain base64 encoded. The idea was that the tool creating the log file presumably knows whether it's analyzing a source file or a binary file.

I think what has to happen is that `FileData.Create` takes an additional argument specifying whether the file is text or binary. But how will we know? We can assume binary unless we have a MIME type?

lgolding: Oh! I just looked at the product code (I always look at the tests first :-)) -- and you have the MIME type available.

---
In reply to: [136158851](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136158851) [](ancestors = 136158851)

src/Sarif/Core/FileData.cs
lgolding: Probably wasn't a problem because in most cases the URI recaps the dictionary key.

---
In reply to: [136142944](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136142944) [](ancestors = 136142944)

src/Sarif.UnitTests/Core/FileDataTests.cs
michaelcfanning: you are now proposing a useful update to the Create implementation, deferring. I will do this next, add mocks, and also convert everything to xunit

---
In reply to: [136158851](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136158851) [](ancestors = 136158851)

michaelcfanning: actually probably should fix now, since we're releasing.

---
In reply to: [136160339](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136160339) [](ancestors = 136160339,136158851)

lgolding: This isn't a "useful update". The current behavior violates the spec.

---
In reply to: [136160339](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136160339) [](ancestors = 136160339,136158851)

michaelcfanning: right, precisely why i reversed my position. see this update.

---
In reply to: [136160846](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136160846) [](ancestors = 136160846,136160339,136158851)

src/Sarif.UnitTests/Core/FileDataTests.cs
lgolding: The `default` doesn't entirely fix the problem. What if the array has three Sha256 hashes in it? The test will pass.

---
In reply to: [136157410](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136157410) [](ancestors = 136157410)

src/Sarif/Core/FileData.cs
```diff
@@ -12,19 +15,44 @@ namespace Microsoft.CodeAnalysis.Sarif
...
             {
-                HashData hashes = HashUtilities.ComputeHashes(uri.LocalPath);
-                fileData.Hashes = new List<Hash>
+                if (!uri.IsAbsoluteUri || !uri.IsFile || !File.Exists(uri.LocalPath))
+                {
+                    return fileData;
+                }
+
+                string filePath = uri.LocalPath;
+                bool encodeAsUtf8 = (fileData.MimeType != SarifWriters.MimeType.Binary);
```
lgolding: >fileData.MimeType != SarifWriters.MimeType.Binar [](start = 37, length = 48)

There are many MIME types other than Binary that are non-text. For example, application/zip.

lgolding: This is a very hard problem.

---
In reply to: [136173115](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136173115) [](ancestors = 136173115)

lgolding: Well ok, first of all, this variable is no longer used, so omit this line.

---
In reply to: [136173223](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136173223) [](ancestors = 136173223,136173115)

src/Sarif/Core/FileData.cs
```diff
@@ -42,9 +70,37 @@ public static FileData Create(Uri uri, bool computeHashes)
...
+            }
+            else
+            {
+                fileContents = File.ReadAllBytes(filePath);
+            }
+
+            return Convert.ToBase64String(fileContents);
+        }
+
+        private static Encoding DetectEncoding(string filePath)
```
lgolding: >DetectEncoding [](start = 32, length = 14)

Nobody calls this.

lgolding: I think you intended to say "if inputEncoding not supplied, then detect encoding"

---
In reply to: [136173600](https://github.com/Microsoft/sarif-sdk/pull/612#discussion_r136173600) [](ancestors = 136173600)

src/Sarif/Core/FileData.cs
```diff
@@ -42,9 +70,37 @@ public static FileData Create(Uri uri, bool computeHashes)
...
+            catch (Exception e) when (e is IOException || e is UnauthorizedAccessException) { }
 
             return fileData;
         }
+
+        private static string EncodeFileContents(string filePath, string mimeType, Encoding inputFileEncoding)
+        {
+            byte[] fileContents;
+
+            if (mimeType != SarifWriters.MimeType.Binary)
```
lgolding: >mimeType != SarifWriters.MimeType.Binary [](start = 16, length = 40)

This is the heart of the problem. We can't reliably detect "text-ness" from MIME type.

PR #613, opened 2017-08-31
Lazy Loading of File Contents

vinaykapadia:


src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -11,35 +11,30 @@ namespace Microsoft.Sarif.Viewer.Models
     public class FileDetailsModel
     {
         // Contents of file. May or may not be Base64 encoded.
-        private string _contents;
+        private string _rawContents;
         
-        private bool _isBase64Encoded;
+        private readonly Lazy<string> _decodedContents;
```
vinaykapadia: Lazy<> implementation is by default thread safe. My implementation was not.

src/Sarif.Viewer.VisualStudio/Models/FileDetailsModel.cs
```diff
@@ -11,35 +11,30 @@ namespace Microsoft.Sarif.Viewer.Models
     public class FileDetailsModel
     {
         // Contents of file. May or may not be Base64 encoded.
```
lgolding: >May or may not be Base64 encoded. [](start = 29, length = 33)

This is no longer true. "Base64-encoded contents of file."

PR #614, opened 2017-09-01
Add file contents persistence option to converters.

michaelcfanning:
@lgolding @vinaykapadia @jeff-bacon


src/Sarif/IFileSystem.cs
```diff
@@ -37,6 +39,17 @@ public interface IFileSystem
...
         /// <summary>
+        /// Opens a binary file, reads all contents into a byte array, and then closes the file.
+        /// </summary>
+        /// <param name="path">
+        /// The file to open for reading.
+        /// </param>
+        /// <returns>
+        /// A byte array containing the contents of the file
+        /// </returns>
+        byte[] ReadAllBytes(string path);
```
michaelcfanning: What nonsense. We apply this to the interface. Then we apply it to the default concrete implementation. It's already available for direct calls to the actual .NET API. None of these API are particularly abstruse in any case. 

Just a bunch of busy work that consumes vertical space. Bleagh.

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -34,7 +33,7 @@ public void FileData_ComputeHashes()
             {
                 File.WriteAllText(filePath, fileContents);
                 FileData fileData = FileData.Create(uri, LoggingOptions.ComputeFileHashes);
-                fileData.Uri.Should().Be(uri.ToString());
+                fileData.Uri.Should().BeNull();
```
michaelcfanning: >fileData.Uri [](start = 16, length = 12)

file.Uri is actually something like file.NestedUri or file.RelativeUri (for cases when an item is relative to some container file).

src/Sarif/IFileSystem.cs
lgolding: MMMV

---
In reply to: [136618018](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136618018) [](ancestors = 136618018)

jeff-bacon: Is this in reference to the comments or the addition of the new method signature?  

---
In reply to: [136618018](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136618018) [](ancestors = 136618018)

lgolding: The comments.

---
In reply to: [136618958](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136618958) [](ancestors = 136618958,136618018)

SetCurrentVersion.cmd
```diff
@@ -1,6 +1,6 @@
 set MAJOR_PREVIOUS=1
 set MINOR_PREVIOUS=5
-set PATCH_PREVIOUS=47
+set PATCH_PREVIOUS=46
```
lgolding: Prev should be 1.6.0, and the new version should be 1.6.1 (really IMO it s.b. 1.7.0 -- this is significant new functionality).

src/ConvertToSarif/ConvertOptions.cs
```diff
@@ -46,5 +46,11 @@ internal class ConvertOptions
             "plugin-assembly-path",
             HelpText = "Path to plugin assembly containing converter types.")]
         public string PluginAssemblyPath { get; internal set; }
+
+
+        [Option(
```
lgolding: >( [](start = 15, length = 1)

Hadn't you already added this option in a previous PR, and didn't you use `'f'` as the short name, per my comment? Between this an my comment on SetCurrentVersion.cmd, I wonder if the branch you're working in is synced to latest changes.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -157,15 +157,17 @@ private SarifLogger(TextWriter textWriter, LoggingOptions loggingOptions)
             }
         }
 
-        public bool ComputeFileHashes { get { return (_loggingOptions & LoggingOptions.ComputeFileHashes) == LoggingOptions.ComputeFileHashes; } }
+        public bool ComputeFileHashes { get { return _loggingOptions.Includes(LoggingOptions.ComputeFileHashes); } }
```
lgolding: >Includes [](start = 69, length = 8)

Yes, this is all from a previous PR as well.

src/ConvertToSarif/ConvertOptions.cs
lgolding: >( [](start = 15, length = 1)

Ah, unlike in AnalyzeOptionsBase, `'f'` is already taken here. ok.

src/ConvertToSarif/Program.cs
```diff
@@ -25,18 +26,23 @@ public static int RunConvertFile(ConvertOptions convertOptions)
...
                 if (convertOptions.PrettyPrint)
                 {
-                    toolFormatConversionOptions |= ToolFormatConversionOptions.PrettyPrint;
+                    loggingOptions |= LoggingOptions.PrettyPrint;
                 };
 
                 if (convertOptions.Force)
                 {
-                    toolFormatConversionOptions |= ToolFormatConversionOptions.OverwriteExistingOutputFile;
+                    loggingOptions |= LoggingOptions.OverwriteExistingOutputFile;
```
lgolding: >OverwriteExistingOutputFile [](start = 53, length = 27)

Why not "Force"? That's pretty conventional for meaning "overwrite existing file".

src/Sarif.Converters/FileInfoFactory.cs
```diff
@@ -9,13 +9,15 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
...
     {
-        private readonly Dictionary<string, FileData> _fileInfoDictionary;
+        private readonly LoggingOptions _loggingOptions;
         private readonly Func<string, string> _mimeTypeClassifier;
+        private readonly Dictionary<string, FileData> _fileInfoDictionary;
 
-        internal FileInfoFactory(Func<string, string> mimeTypeClassifier)
-        {
+        internal FileInfoFactory(Func<string, string> mimeTypeClassifier, LoggingOptions loggingOptions)
+        {
```
lgolding: >             [](start = 9, length = 12)

Remove trailing whitespace.

src/Sarif.Converters/FileInfoFactory.cs
```diff
@@ -89,15 +91,17 @@ private void AddFile(PhysicalLocation physicalLocation)
                 filePath = uri.LocalPath;
             }
 
+            FileData fileData = FileData.Create(
```
lgolding: >FileData [](start = 12, length = 8)

We should consider whether `Foo.Create` is sufficiently explicit to allow us to use `var` here.

src/Sarif/Core/FileData.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// </summary>
     public partial class FileData : ISarifNode
     {
+        internal static IFileSystem _fileSystem = new FileSystem();
```
lgolding: >internal static IFileSystem _fileSystem = new FileSystem() [](start = 8, length = 58)

I suggest that you use constructor injection, as follows (well, it's not exactly "constructor" injection, but close enough):
```
public static FileData Create(TheArgs theArgs)
{
    Create(theArgs, new FileSystem());
}

// For testing only:
internal static FileData Create(TheArgs theArgs, IFileSystem fileSystem)
{
    // the real code.
}
```

src/Sarif/Core/FileData.cs
```diff
@@ -25,7 +27,6 @@ public static FileData Create(Uri uri, SarifWriters.LoggingOptions loggingOption
             var fileData = new FileData()
             {
                 MimeType = mimeType,
-                Uri = uri
```
lgolding: >Uri = uri [](start = 16, length = 9)

Why is this removed?

src/Sarif.Converters/ToolFormatConverter.cs
```diff
@@ -60,44 +60,19 @@ public class ToolFormatConverter
...
             }
         }
 
-        /// <summary>Converts a tool log file to the SARIF format.</summary>
-        /// <param name="toolFormat">The tool format of the input file.</param>
-        /// <param name="inputFileName">The input log file name.</param>
-        /// <param name="outputFileName">The name of the file to which the resulting SARIF log shall be
-        /// written. This cannot be a directory.</param>
-        /// <remarks>This overload should be used only from test code.</remarks>
-        public void ConvertToStandardFormat(
```
lgolding: >ConvertToStandardFormat [](start = 20, length = 23)

Yay for deletion

BuildAndTest.cmd
```diff
@@ -81,6 +81,15 @@ goto ExitFailed
 @REM Run all tests
 SET PASSED=true
 
+mstest /testContainer:bld\bin\Sarif.Converters.UnitTests\AnyCPU_%Configuration%\Sarif.Converters.UnitTests.dll
```
lgolding: >mstest [](start = 0, length = 6)

We were never running these before?? Good catch.

src/Sarif/IFileSystem.cs
michaelcfanning: we have a true compromise. no one is happy. everyone has agreed to drop doc comments against internal/private items. i have agreed to maintain/provide them for public.


---
In reply to: [136619704](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136619704) [](ancestors = 136619704,136618958,136618018)

src/ConvertToSarif/Program.cs
michaelcfanning: 'Force' could be a star wars reference, that's why. 'OverwriteExistingOutputFile' is perfectly descriptive. Think of it as an XML doc comment that's burned RIGHT INTO THE NAME, hee,

---
In reply to: [136626110](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136626110) [](ancestors = 136626110)

BuildAndTest.cmd
michaelcfanning: Yep. Neglected this change when we refactored the converters.

---
In reply to: [136628913](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136628913) [](ancestors = 136628913)

src/Sarif.Converters/ToolFormatConverter.cs
michaelcfanning: Yay! Death to unnecessary overload

---
In reply to: [136628767](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136628767) [](ancestors = 136628767)

src/Sarif/Core/FileData.cs
michaelcfanning: Per other comment, this is the wrong data for this member (which should be a relative uri only for nested items)

---
In reply to: [136628150](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136628150) [](ancestors = 136628150)

src/Sarif/Core/FileData.cs
michaelcfanning: why should I instantiate the file system for each call? Also, you approach will require mucking with every API that utilizes the file system object. my approach modifies a common internal member that arbitrary API could use.

---
In reply to: [136627741](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136627741) [](ancestors = 136627741)

src/Sarif.Converters/FileInfoFactory.cs
michaelcfanning: i don't care about having 'var' here. The value of 'var' in my opinion is 1) reduce line column count, 2) allows declarations to line up nicely in multi-var decls.

in this case, we have a single variable and FileData is not an egregiously long type name

---
In reply to: [136626728](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136626728) [](ancestors = 136626728)

src/Sarif/Core/FileData.cs
lgolding: Actually, no:
----
5.15.2:
If the file object represents a top-level file, then the uri property may be present. If present, it shall be equal to the name of the property within run.files (§5.12.9) whose value is this file object. If absent, it shall be interpreted as having that same value.

If the file object represents a nested file whose location relative to its parent can be expressed only by means of a path, then the uri property shall be present, and its value shall be a valid relative URI expressing that path.
----
In other words, for a top-level file, you are allowed to repeat the key as the `uri` property, but you don't have to. So, you can certainly justify this as "not needed, saves space". But it's not the case that this is always a "nested" URI.

---
In reply to: [136632876](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136632876) [](ancestors = 136632876,136628150)

src/Sarif/Core/FileData.cs
lgolding: Because global data is evil. You could imagine getting a test wrong by neglecting-to-set-up-a-new-mock-and-assign-it-to-the-static; then your test would accidentally use the mock from some previous test -- and who could tell which one? -- and your test would fail unpredictably.

With injection, you guarantee each test to use its own mock.

---
In reply to: [136633228](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136633228) [](ancestors = 136633228,136627741)

src/Sarif.UnitTests/Core/FileDataTests.cs
lgolding: Not necessarily. See my other comment.

---
In reply to: [136618490](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136618490) [](ancestors = 136618490)

src/Sarif/Core/FileData.cs
lgolding: As we discussed, better pattern is:
```
public static FileData Create(TheArgs theArgs, IFileSystem fileSystem = null)
{
    if (fileSystem == null) fileSystem = new FileSystem();
}
```

---
In reply to: [136634455](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136634455) [](ancestors = 136634455,136633228,136627741)

michaelcfanning: beautiful, like it

---
In reply to: [136639877](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136639877) [](ancestors = 136639877,136634455,136633228,136627741)

src/Sarif/Core/FileData.cs
michaelcfanning: Oops. ok, will resolve

---
In reply to: [136633850](https://github.com/Microsoft/sarif-sdk/pull/614#discussion_r136633850) [](ancestors = 136633850,136632876,136628150)

PR #616, opened 2017-09-07
Create a SARIF 'multitool' that applies arbitrary transformations to input files

michaelcfanning:
Change absorbs log file conversion and SARIF validation. Complete roll-out of validation currently blocked by Jschema SARIF dependency. Would appreciate getting the review started, this work is getting unblocked in parallel. 

@lgolding @jeff-bacon @vinaykapadia


lgolding:
> [assembly: AssemblyTitle("SARIF Conversion Tool")]

Title and Description are now out of date.

---
Refers to: src/Sarif.Multitool/Properties/AssemblyInfo.cs:6 in b3df05e. [](commit_id = b3df05e00d2c61efb885d332debd707d38dbfb44, deletion_comment = False)

src/ConvertToSarif/ConvertOptions.cs
```diff
@@ -3,28 +3,20 @@
 
 using CommandLine;
 
-namespace Microsoft.CodeAnalysis.Sarif.ConvertToSarif
```
lgolding: >ConvertToSarif [](start = 39, length = 14)

I don't agree with pushing the classes in this tool up into the base SARIF namespace.

src/Sarif.Multitool/MultitoolOptionsBase.cs
```diff
@@ -3,28 +3,20 @@
...
 {
-    [Verb("convert", HelpText = "Convert a tool output log to SARIF format.")]
-    internal class ConvertOptions
+    public class MultitoolOptionsBase
     {
         [Value(0,
-               MetaName = "<inputLogFile>",
-               HelpText = "A file path to a tool log file that should be converted to the SARIF format.",
+               MetaName = "<inputFile>",
+               HelpText = "A path to a file to operate against.",
```
lgolding: >operate against [](start = 47, length = 15)

"process"?

src/Sarif.Multitool/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>The entry point for the SARIF file format conversion utility.</summary>
+        /// <param name="args">Arguments passed in from the tool's command line.</param>
+        /// <returns>0 on success; nonzero on failure.</returns>
+        public static int Main(string[] args)
+        {
+            return Parser.Default.ParseArguments<
+                ConvertOptions, 
+                RewriteOptions>(args)
+              .MapResult(
```
lgolding: >MapResult [](start = 15, length = 9)

I believe, because of the particular form of the lambdas, you can express this more simply as:
```
.MapResult(
    ConvertCommand.Run,
    RewriteCommand.Run,
    errs => 1
```

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (!String.IsNullOrEmpty(rewriteOptions.OutputFilePath))
+            {
+                return rewriteOptions.OutputFilePath;
+            }
+
+            string extension = Path.GetExtension(filePath);
+
+            // For an input file named MyFile.sarif, returns MyFile.transformed.sarif
+            if (extension.Equals(".sarif", StringComparison.OrdinalIgnoreCase))
```
lgolding: >) [](start = 78, length = 1)

Consider:
```
private const string SarifExtension = ".sarif";
private const string TransformedExtension = ".transformed" + SarifExtension;
...
return blahBlah + TransformedExtension;
```

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                return Path.GetFileNameWithoutExtension(filePath) + "transformed.sarif";
```
lgolding: >"t [](start = 68, length = 2)

Lacks period before the extension. Moot if you adopt suggestion above.

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                LoggingOptions loggingOptions = rewriteOptions.ConvertToLoggingOptions();
+
+                SarifLog reformattedLog = new ReformattingVisitor(loggingOptions).VisitSarifLog(actualLog);
+
+                File.WriteAllText(fileName, JsonConvert.SerializeObject(reformattedLog, settings));
+            }
+            catch(Exception ex)
+            {
+                Console.WriteLine(ex.ToString());
```
lgolding: >ToString() [](start = 37, length = 10)

Not necessary. `Console.WriteLine` calls `ToString` on its arguments.

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                Formatting formatting = rewriteOptions.PrettyPrint ? Formatting.Indented : Formatting.None;
+
+                JsonSerializerSettings settings = new JsonSerializerSettings()
+                {
+                    ContractResolver = SarifContractResolver.Instance,
+                    Formatting = Formatting.Indented
+                };
+
+                string sarifText = File.ReadAllText(rewriteOptions.InputFilePath);
```
lgolding: >File.ReadAllText [](start = 35, length = 16)

If you do this, you can unit test even your top level code:
```
public static int Run(RewriteOptions rewriteOptions)
{
    return Run(rewriteOptions, new FileSystem());
}

internal static int Run(
    RewriteOptions rewriteOptions,
    IFileSystem fileSystem)
{
    ...
    string sarifText = fileSystem.ReadAllText(...);
}
```

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+                SarifLog actualLog = JsonConvert.DeserializeObject<SarifLog>(sarifText, settings);
```
lgolding: >actualLog [](start = 25, length = 9)

Consider: `inputLog`.

I usually reserve names `actualFoo` for unit tests, where I compare them to `expectedFoo`.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -0,0 +1,55 @@
...
+                            node.Files[key] = VisitDictionaryValueNullChecked(key, value);
+                        }
+                    }
+                }
+            }
+
+            return node;
+        }
+
+        public FileData VisitDictionaryValueNullChecked(string key, FileData node)
```
lgolding: >public [](start = 8, length = 6)

Does this need to be public?

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            bool workToDo = false;
+
+            workToDo |= node.Hashes == null && _loggingOptions.Includes(LoggingOptions.ComputeFileHashes);
```
lgolding: >| [](start = 21, length = 1)

Probably works, but you really want `||=` here since you're dealing with Booleans, not bit fields.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class ReformattingVisitor : SarifRewritingVisitor
```
lgolding: >ReformattingVisitor [](start = 17, length = 19)

Isn't this really something like `FileDataCompletingVisitor`?

src/Sarif.Multitool/Program.cs
michaelcfanning: no, 'parameterization cannot be inferred from usage'

src/Sarif.Multitool/RewriteCommand.cs
michaelcfanning: Right, but we should avoid calling Console.WriteLine(object) due to unexpected issues around culture-specific behaviors of ToString() calls. Same with String.Concat.

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: not at all. |= is clearly defined in the C# spec and this usage is acceptable

src/Sarif.Multitool/RewriteCommand.cs
lgolding: I see. ToString doesn't actually help here (still culture sensitive). BUT, there are classes that offer a ToString(CultureInfo) overload, giving you the opportunity to use InvariantCulture. If you get in the habit of always using ToString, you'll at least check to see if the CultureInfo overload is available. Right?

---
In reply to: [137909535](https://github.com/Microsoft/sarif-sdk/pull/616#discussion_r137909535) [](ancestors = 137909535)

src/Sarif.Multitool/Program.cs
lgolding: bummer.

---
In reply to: [137909245](https://github.com/Microsoft/sarif-sdk/pull/616#discussion_r137909245) [](ancestors = 137909245)

src/Sarif/Visitors/ReformattingVisitor.cs
lgolding: Ah. TIL: "The | operator performs a bitwise logical OR operation on integral operands and logical OR on bool operands."

Thanks!

---
In reply to: [137909791](https://github.com/Microsoft/sarif-sdk/pull/616#discussion_r137909791) [](ancestors = 137909791)

PR #618, opened 2017-09-08
Display Embedded File Contents

vinaykapadia:


src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -212,6 +212,9 @@ public int OnQueryCloseSolution(object pUnkReserved, ref int pfCancel)
 
         public int OnBeforeCloseSolution(object pUnkReserved)
         {
+            // When closing solution (or closing VS), remove the temporary folder.
+            CleanupTemporaryFiles();
```
lgolding: >Cleanup [](start = 12, length = 7)

As a verb, "clean up" is two words, so: `CleanUpTemporaryFiles`. Or maybe just `RemoveTemporaryFiles`.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +242,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
         }
 
+        // Contents are embedded in SARIF. Create a file from these contents.
+        private string CreateFileFromContents(string fileName)
+        {
+            var fileData = FileDetails[fileName];
+
+            // Get temporary path.
+            string path = Path.GetTempPath();
+            path = Path.Combine(path, "SARIF-Viewer");
```
lgolding: >SARIF-Viewer [](start = 39, length = 12)

```
private const string TemporaryFileDirectoryName = "SarifViewer";
```

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +242,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
 
             string rebaselinedFile;
 
-            if (Uri.IsWellFormedUriString(originalFilename, UriKind.Absolute))
+            // File contents embedded in SARIF.
```
lgolding: >// [](start = 12, length = 2)

As written, this comment appears to apply to the whole `if`/`else` statement. Since it only applies to the `if` block, I'd place it in there:
```
if (FileDetails.ContainsKey(originalFilename))
{
    // File contents are embedded in the SARIF log file.
    rebaselinedFile = CreateFileFromContents(originalFilename);
}
else
{
    ...
}
```

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
             return true;
         }
 
+        // Contents are embedded in SARIF. Create a file from these contents.
+        private string CreateFileFromContents(string fileName)
+        {
+            var fileData = FileDetails[fileName];
+
+            // Get temporary path.
+            string path = Path.GetTempPath();
```
lgolding: It's confusing that `path` means two different things:
1. The path to the temporary directory.
2. The full path to the stored file.

I suggest:
1. Declare an instance variable `temporaryFileDirectoryPath`. Initialize it in the ctor, using the logic shown here.
2. Then declare `path` here (not initialized yet).
3. Then:
```
if (fileName.StartsWith(temporaryFileDirectoryPath))
{
    path = fileName;
}
{
    ...
}
```

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+                // leading slash, so that we can relate the file path to an
+                // arbitrary remapped disk location.
+                string filePath = fileName;
+                filePath = filePath.Substring(Path.GetPathRoot(filePath).Length);
+
+                path = Path.Combine(path, fileData.Sha256Hash);
+                path = Path.Combine(path, filePath);
+            }
+
+            string directory = new FileInfo(path).Directory.FullName;
```
lgolding: >new FileInfo(path).Directory.FullName; [](start = 31, length = 38)

Simpler:
```
string = Path.GetDirectoryName(path);
```
[MSDN](https://msdn.microsoft.com/en-us/library/system.io.path.getdirectoryname(v=vs.110).aspx)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+
+            if (!Directory.Exists(directory))
```
lgolding: >!Directory.Exists(directory) [](start = 16, length = 28)

No need to test. It's safe to call `CreateDirectory` if it already exists.

[MSDN](https://msdn.microsoft.com/en-us/library/54a0at6s(v=vs.110).aspx)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+            {
+                Directory.CreateDirectory(directory);
+            }
+
+            if (!File.Exists(path))
+            {
+                string contents = fileData.GetContents();
+                File.WriteAllText(path, contents);
```
lgolding: >WriteAllText [](start = 21, length = 12)

What happens to binary files?

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+                // File should be readonly, because it is embedded.
```
lgolding: >File should be readonly, because it is embedded [](start = 19, length = 47)

Is there also a way to mark the editing buffer as readonly? As the code stands, I can type in it, and Save As it somewhere else. Is that ok? Thoughts?

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+            }
+            // Else we have to create a location under the temp path.
+            else
+            {
+                // If the file path is absolute, or if it is relative to an
+                // unknown root, we need to strip the drive letter and/or
+                // leading slash, so that we can relate the file path to an
+                // arbitrary remapped disk location.
+                string filePath = fileName;
+                filePath = filePath.Substring(Path.GetPathRoot(filePath).Length);
```
lgolding: This is the part of the code I had the most trouble with. I can't match the comment "If the file path is absolute, or if it is relative to an unknown root..." up to this piece of logic. (

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -239,24 +243,89 @@ public bool TryRebaselineCurrentSarifError(string uriBaseId, string originalFile
...
+
+                path = Path.Combine(path, fileData.Sha256Hash);
+                path = Path.Combine(path, filePath);
```
lgolding: "Good news, everyone!" Since .NET 4, there's a 3-argument overload:
```
path = Path.Combine(path, fileData.Sha256Hash, filePath);
```

Also: worth a comment explaining how you know `Sha256Hash` is guaranteed to be present.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
vinaykapadia: Nothing, we don't even get this far with binary files. You can't open an error in a binary file.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
vinaykapadia: That's one of our requirements. For now, you can't show an embedded file that doesn't have a hash (and as per Michael's thoughts, in the future we'll calculate the hash if it's not present).

lgolding: Yes, but where is that requirement enforced? That's what the concerned reader will want to know. It's not obvious from examining the code locally.

---
In reply to: [137899640](https://github.com/Microsoft/sarif-sdk/pull/618#discussion_r137899640) [](ancestors = 137899640)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
lgolding: On the day we add "open in binary editor", life will be great until the first time (months or years later) when somebody embeds a binary in the SARIF file. I don't know how to handle this: at this point, do we even know if the file was binary?

---
In reply to: [137898777](https://github.com/Microsoft/sarif-sdk/pull/618#discussion_r137898777) [](ancestors = 137898777)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
vinaykapadia: As we discussed: The improved name `finalPath` makes the code clear.

---
In reply to: [137895686](https://github.com/Microsoft/sarif-sdk/pull/618#discussion_r137895686) [](ancestors = 137895686)

PR #624, opened 2017-09-25
NET Core Port

jeff-bacon:


src/Sarif/HashUtilities.cs
```diff
@@ -20,7 +20,7 @@ public static HashData ComputeHashes(string fileName)
                         string md5, sha1, sha256;
                         byte[] checksum;
 
-                        using (var md5Cng = new MD5Cng())
+                        using (var md5Cng = MD5.Create())
```
jeff-bacon: *Cng algorithms no longer exist in NET Core

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -14,7 +14,10 @@
   </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.props))\build.props" />
   <ItemGroup>
-    <Reference Include="Microsoft.VisualBasic" />
+    <Reference Include="CsvHelper, Version=2.0.0.0, Culture=neutral, PublicKeyToken=8c4959082be5c823, processorArchitecture=MSIL">
```
jeff-bacon: Replaced CSV parser with NET Core compatible.

src/Sarif.Viewer.VisualStudio.UnitTests/Sarif.Viewer.VisualStudio.UnitTests.csproj
```diff
@@ -89,8 +89,9 @@
       <Name>Sarif.Viewer.VisualStudio</Name>
     </ProjectReference>
     <ProjectReference Include="..\Sarif\Sarif.csproj">
-      <Project>{CC9B247E-7103-4BF7-BDAB-6E625B3680A8}</Project>
+      <Project>{c5fc2683-47cf-447f-b5ea-9fde0fd78754}</Project>
       <Name>Sarif</Name>
+      <AdditionalProperties>TargetFramework=net452</AdditionalProperties>
```
jeff-bacon: Since this project wasn't replaced with a NET Core project, specify which version output of Sarif it should use.

src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
```diff
@@ -374,15 +383,6 @@
       <Generator>MSBuild:Compile</Generator>
     </Page>
   </ItemGroup>
-  <ItemGroup />
-  <PropertyGroup>
```
jeff-bacon: There was an issue with PREfastXmlSarifConverter.dll not being available for copy at the correct time.  Removed this and now included a .dll in a /lib folder

.gitignore
```diff
@@ -218,5 +218,4 @@ GeneratedArtifacts/
 _Pvt_Extensions/
 ModelManifest.xml
 
-PREfastXmlSarifConverter.dll
-
+obj.core/
```
jeff-bacon: This needs to be removed.  This directory no longer exists.

BuildAndTest.cmd
```diff
@@ -57,79 +57,76 @@ echo         public const string Version = AssemblyVersion + Prerelease;
...
 
-
 call BeforeBuild.cmd
 
 if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
-msbuild /verbosity:minimal /target:rebuild src\Everything.sln /p:"Configuration=%Configuration%" /p:"Platform=Any CPU" /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=true" /p:"BinSkimVerboseOutput=true"
+msbuild /verbosity:minimal /target:rebuild src\Everything.sln /p:"Configuration=%Configuration%" /p:"Platform=Any CPU" /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
```
jeff-bacon: This needs to be set back to 'RunBinSkim=true'

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -1,92 +1,48 @@
...
+
   <ItemGroup>
-    <ProjectReference Include="..\Sarif\Sarif.csproj">
-      <Project>{cc9b247e-7103-4bf7-bdab-6e625b3680a8}</Project>
-      <Name>Sarif</Name>
-    </ProjectReference>
+    <PackageReference Include="CsvHelper" Version="2.16.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+    <PackageReference Include="System.IO.Compression" Version="4.3.0" />
+    <PackageReference Condition="'$(TargetFramework)' == 'net452'" Include="System.Collections.Immutable" Version="1.1.37" />
```
jeff-bacon: The version of Sarif.Converters used in the SARIF Viewer needs this specific version of System.Collections.Immutable because the Visual Studio Shell expects it.

BeforeBuild.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
cfaucon: General question/comment: in addition to the unit tests and this port successfully building, what other testing has been performed to verify everything passes sanity?

jeff-bacon: Mostly Unit Tests passing and I've tested installing the SARIF Viewer and opening SARIF files within it.  I've also converted a few files to SARIF with the converter but this was before it was replaced with Multitool.

BuildAndTest.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
```
michaelcfanning: wow. you ported the whole build file. :)

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -52,7 +59,13 @@ public void SarifLogger_RedactedCommandLine()
                 string pathToExe = Path.GetDirectoryName(Assembly.GetCallingAssembly().Location);
                 string commandLine = Environment.CommandLine;
 
-                if (pathToExe.IndexOf(@"\Extensions", StringComparison.OrdinalIgnoreCase) != -1)
+                if (commandLine.Contains("testhost.dll") || commandLine.Contains("\\xunit.console"))
+                {
+                    var index = commandLine.LastIndexOf("\\");
```
michaelcfanning: >var [](start = 20, length = 3)

don't use var to hold assignment from method calls, prefer the explicit type (so that readers know what it is). Fix this issue everywhere. :)

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -52,7 +59,13 @@ public void SarifLogger_RedactedCommandLine()
                 string pathToExe = Path.GetDirectoryName(Assembly.GetCallingAssembly().Location);
                 string commandLine = Environment.CommandLine;
 
-                if (pathToExe.IndexOf(@"\Extensions", StringComparison.OrdinalIgnoreCase) != -1)
+                if (commandLine.Contains("testhost.dll") || commandLine.Contains("\\xunit.console"))
```
michaelcfanning: >testhost.dll [](start = 42, length = 12)

is it ok that this isn't case-sensitive?

src/Sarif.UnitTests/EqualityComparerTests.cs
```diff
@@ -33,14 +32,14 @@ public void EqualityComparer_ComputesTheSameHashCodeForDistinctEquivalentObjects
                 }
             };
 
-            Assert.AreNotSame(result1, result2);
+            Assert.NotSame(result1, result2);
```
michaelcfanning: >Assert.NotSame [](start = 12, length = 14)

later we could consider a fluent assertions scrub, not worth it right now of course.

BeforeBuild.sh
jeff-bacon: Mostly Unit Tests passing and I've tested installing the SARIF Viewer and opening SARIF files within it.  I've also converted a few files to SARIF with the converter but this was before it was replaced with Multitool.

---
In reply to: [141489911](https://github.com/Microsoft/sarif-sdk/pull/624#discussion_r141489911) [](ancestors = 141489911)

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
jeff-bacon: I don't think Environment.CommandLine would return inconsistent values.  Since this is a test case, it's probably better to be safe than sorry though.

---
In reply to: [141766792](https://github.com/Microsoft/sarif-sdk/pull/624#discussion_r141766792) [](ancestors = 141766792)

src/Sarif.UnitTests/EqualityComparerTests.cs
jeff-bacon: I was actually going to do this but I felt this PR was big enough already.  It's a nice to have for in the future though.

---
In reply to: [141766980](https://github.com/Microsoft/sarif-sdk/pull/624#discussion_r141766980) [](ancestors = 141766980)

PR #625, opened 2017-09-26
Adding merge verb.

vinaykapadia:


src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+        private static SarifLog ReadFile(string filePath)
+        {
+            JsonSerializerSettings settings = new JsonSerializerSettings()
+            {
+                ContractResolver = SarifContractResolver.Instance
+            };
+
+            string logText = File.ReadAllText(filePath);
+            
+            SarifLog log = JsonConvert.DeserializeObject<SarifLog>(logText, settings);
```
jeff-bacon: Variable isn't used.  Return JsonConvert.DeserializeObject

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    internal static class MergeCommand
+    {
+        public static int Run(MergeOptions mergeOptions)
+        {
+            // Get all SARIF files listed in options.
+            IEnumerable<string> sarifFiles = from path in mergeOptions.Files
```
jeff-bacon: This may be easier to read if it wasn't linq.  Also, each variable being set with these linq statements could be in their own methods.  This would also eliminate the comments above each variable.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+				    }
+			    }
+		    }
+	    }
+
+	    private static IEnumerable<string> GetSarifFiles(IEnumerable<string> files)
+	    {
+		    foreach (var path in files)
+		    {
+			    var lastBlackslashPos = path.LastIndexOf('\\') + 1;
```
cfaucon: This will explode if the file string is actually a directory and ends with a '/'.  Add protection here and log a user actionable error.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+			    var directory = path.Substring(0, lastBlackslashPos);
+			    var filename = path.Substring(lastBlackslashPos, path.Length - lastBlackslashPos);
+			    foreach (var file in Directory.GetFiles(directory, filename))
+			    {
+				    if (file.EndsWith(".sarif", StringComparison.InvariantCultureIgnoreCase))
```
cfaucon: Is this a const somewhere so that the file format extension can be / is manage in a single point of code.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+		    }
+	    }
+
+	    private static IEnumerable<string> GetSarifFiles(IEnumerable<string> files)
+	    {
+		    foreach (var path in files)
+		    {
+			    var lastBlackslashPos = path.LastIndexOf('\\') + 1;
+			    var directory = path.Substring(0, lastBlackslashPos);
+			    var filename = path.Substring(lastBlackslashPos, path.Length - lastBlackslashPos);
```
cfaucon: Consider using Path. help functions or FileInfo class methods instead of substring manipulation.

src/Sarif.Multitool/MergeOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    [Verb("merge", HelpText = "Merge multiple SARIF files into one.")]
+    internal class MergeOptions : MultitoolOptionsBase
+    {
+        [Value(0,
+            MetaName = "<files>",
+            HelpText = "Files to process (wildcards ? and * allowed).",
+            Required = false)]
+        public IList<string> Files { get; internal set; }
+    }
```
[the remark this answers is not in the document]
jeff-bacon: This question has been brought up by Vinay.  I think it's a good idea and wouldn't take too much work.

---
In reply to: [141466086](https://github.com/Microsoft/sarif-sdk/pull/625#discussion_r141466086) [](ancestors = 141466086)

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+    internal static class MergeCommand
+    {
+        public static int Run(MergeOptions mergeOptions)
+        {
+            var sarifFiles = GetSarifFiles(mergeOptions.Files);
+
+	        var allRuns = GetAllRuns(sarifFiles);
+
+	        // Build one SarifLog with all the Runs.
+			SarifLog combinedLog = new SarifLog(SarifVersion.OneZeroZero.ConvertToSchemaUri(),
```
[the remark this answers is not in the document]
vinaykapadia: It's a web browser issue. Looks fine in VS.

---
In reply to: [141463923](https://github.com/Microsoft/sarif-sdk/pull/625#discussion_r141463923) [](ancestors = 141463923)

src/Sarif.Multitool/MergeCommand.cs
vinaykapadia: Not that I can find. The string ".sarif" does show up multiple times in the code, so this is something we should be doing. I think we should get this PR done then add a work item for that though.

---
In reply to: [141469083](https://github.com/Microsoft/sarif-sdk/pull/625#discussion_r141469083) [](ancestors = 141469083)

src/Sarif.Multitool/MergeOptions.cs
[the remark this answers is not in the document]
vinaykapadia: Added.

---
In reply to: [141471181](https://github.com/Microsoft/sarif-sdk/pull/625#discussion_r141471181) [](ancestors = 141471181,141466086)

PR #626, opened 2017-09-27
Fixed FxCop Issues

jeff-bacon:
Fixed issues where FxCop files without a location would not open in the viewer.  Also fixed an issue where some relative URIs were causing failure in opening FxCop files.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -15,6 +15,7 @@ using Microsoft.CodeAnalysis.Sarif.Readers;
 using Microsoft.CodeAnalysis.Sarif.Writers;
 using Microsoft.Sarif.Viewer.Models;
 using Newtonsoft.Json;
+using Microsoft.Sarif.Viewer.Sarif;
 
 namespace Microsoft.Sarif.Viewer.ErrorList
 {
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -87,25 +88,9 @@ private ErrorListService()
         private void WriteRunToErrorList(Run run, string logFilePath, Solution solution)
         {
             List<SarifErrorListItem> sarifErrors = new List<SarifErrorListItem>();
-
             var projectNameCache = new ProjectNameCache(solution);
 
-            if (run.Files != null)
```
jeff-bacon: Extracted this into it's own method

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -120,7 +105,36 @@ private void WriteRunToErrorList(Run run, string logFilePath, Solution solution)
...
+        {
+            if (files == null)
+            {
+                return;
+            }
+
+            foreach (var file in files)
+            {
+                Uri key;
+                var isValid = Uri.TryCreate(file.Key, UriKind.RelativeOrAbsolute, out key);
```
jeff-bacon: Ensure this is a valid relative or absolute Uri before continuing.

src/Sarif.Viewer.VisualStudio/Sarif/Result.extensions.cs
```diff
@@ -60,10 +60,12 @@ public static string GetPrimaryTargetFile(this Result result)
             {
                 return primaryLocation.AnalysisTarget.Uri.ToPath();
             }
-            else
+            else if (primaryLocation.FullyQualifiedLogicalName != null)
```
jeff-bacon: Some files have no ResultFile, AnalysisTarget or FullyQualifiedLogicalName.  In this case, return an empty string for the file name.  Must return some value other than null because this is used as a key in a dictionary.

PR #628, opened 2017-09-29
Populate files object from physical locations on request to embed fil…

michaelcfanning:
…e contents/hashes in a rewrite operation.

@cfaucon @jeff-bacon @vinaykapadia


jeff-bacon:
>         public string InputFilePath { get; internal set; }

'SingleFileOptionsBase.InputFilePath' hides inherited member 'MultitoolOptionsBase.InputFilePath'. Use the new keyword if hiding was intended.

---
Refers to: src/Sarif.Multitool/MultitoolOptionsBase.cs:13 in ec1af0b. [](commit_id = ec1af0b86f1b1a5d4fa4a063f63c85b12eaa2b22, deletion_comment = False)

vinaykapadia:
>         public string InputFilePath { get; internal set; }

Michael, I removed this from here and added it to the SingleFileOptionsBase so that any option that uses only a single file can inherit from that (which in turn inherits from this). If there is an option that takes in multiple files (or doesn't take in a file at all) then this will not work for that.

---
Refers to: src/Sarif.Multitool/MultitoolOptionsBase.cs:13 in ec1af0b. [](commit_id = ec1af0b86f1b1a5d4fa4a063f63c85b12eaa2b22, deletion_comment = False)

PR #633, opened 2017-10-16
Added TSLint Converter and unit tests.

boAndron:
- Added TSLintConverter and integrated it into the Converters project
- Minimal unit testing (Note: Added dependency on Moq (via nuget))

### DIFF
--- a/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
+++ b/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
@@ -25,7 +25,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 Failure = "failure.test.value",
                 Name = "name.test.value",
                 RuleName = "ruleName.test.value",
-                RuleSeverity = "ruleSeverity.test.value"
+                RuleSeverity = "WARN"
             };
 
             TSLintLogFix testFix = new TSLintLogFix()
@@ -64,7 +64,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             {
                 RuleId = "ruleName.test.value",
                 Message = "failure.test.value",
-                Level = ResultLevel.NotApplicable
+                Level = ResultLevel.Warning
             };
 
             Region region = new Region()
@@ -165,7 +165,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         [Fact]
         public void TSLintConverter_CreateResult_WhenInputIsNull_ThrowsArgumentNullException()
         {
-            TSLintConverter converter = new TSLintConverter();
+            var converter = new TSLintConverter();
 
             Action action = () => converter.CreateResult(null);
             action.ShouldThrow<ArgumentNullException>();
@@ -174,14 +174,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         [Fact]
         public void TSLintConverter_CreateResult_CreatesExpectedResult()
         {
-            TSLintConverter converter = new TSLintConverter();
+            var converter = new TSLintConverter();
             TSLintLog tSLintLog = CreateTestTSLintLog();
 
             Result actualResult = converter.CreateResult(tSLintLog[0]);
             Result expectedResult = CreateTestResult();
 
-            Assert.True(Result.ValueComparer.Equals(actualResult, expectedResult));
+            Result.ValueComparer.Equals(actualResult, expectedResult).Should().BeTrue();
         }
-
     }
 }
--- a/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
+++ b/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
@@ -1,0 +1,162 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+using Moq;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    public class TSLintConverterTests
+    {
+        public TSLintLog CreateTestTSLintLog()
+        {
+            TSLintLog tsLintLog = new TSLintLog();
+
+            TSLintLogEntry testEntry = new TSLintLogEntry()
+            {
+                Failure = "failure.test.value",
+                Name = "name.test.value",
+                RuleName = "ruleName.test.value",
+                RuleSeverity = "ruleSeverity.test.value"
+            };
+
+            TSLintLogFix testFix = new TSLintLogFix()
+            {
+                InnerLength = 5,
+                InnerStart = 10,
+                InnerText = "fix.innerText.test.value"
+            };
+
+            testEntry.Fixes = new List<TSLintLogFix>() { testFix };
+
+            TSLintLogPosition testStartPos = new TSLintLogPosition()
+            {
+                Character = 1,
+                Line = 2,
+                Position = 3
+            };
+            testEntry.StartPosition = testStartPos;
+
+            TSLintLogPosition testEndPos = new TSLintLogPosition()
+            {
+                Character = 11,
+                Line = 12,
+                Position = 13
+            };
+            testEntry.EndPosition = testEndPos;
+
+            tsLintLog.Add(testEntry);
+
+            return tsLintLog;
+        }
+
+        // This is necessary because Microsoft.CodeAnalysis.Sarif.Result doesn't implement IEquatable
+        // Must be kept up to date with any changes to TSLintConverter.CreateResult
+        public bool CompareResultToTestData(Result result)
+        {
+            if (!result.RuleId.Equals("ruleName.test.value")) return false;
+            if (!result.Message.Equals("failure.test.value")) return false;
+
+            if (result.Level != ResultLevel.NotApplicable) return false;
+
+            if (result.Locations.Count != 1) return false;
+
+            Region region = result.Locations[0].AnalysisTarget.Region;
+            if (region.StartLine != 3) return false;
+            if (region.StartColumn != 2) return false;
+            if (region.EndLine != 13) return false;
+            if (region.EndColumn != 12) return false;
+
+            if (region.Offset != 3) return false;
+            if (region.Length != 11) return false;
+
+            if (!result.Locations[0].AnalysisTarget.Uri.OriginalString.Equals("name.test.value")) return false;
+
+            if (result.Fixes.Count != 1) return false;
+            if (result.Fixes[0].FileChanges.Count != 1) return false;
+            if (result.Fixes[0].FileChanges[0].Replacements.Count != 1) return false;
+            Replacement replacement = result.Fixes[0].FileChanges[0].Replacements[0];
+
+            if (replacement.Offset != 10) return false;
+            if (replacement.DeletedLength != 5) return false;
+            if (!replacement.InsertedBytes.Equals("fix.innerText.test.value")) return false;
+
+            return true;
+        }
+
+        //[Fact]
+        //[Trait("Category", "E2E")]
+        //public void SimpleE2ETest_ManualValidation()
+        //{
+        //    ToolFormatConverter converter = new ToolFormatConverter();
+        //    converter.ConvertToStandardFormat(ToolFormat.TSLint, @"E:\data\TSLint\TSLint.Results.json", @"E:\data\TSLint\TSLint.Results.sarif");
+        //}
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintConverterTest_Convert_S2S_NullInput()
+        {
+            TSLintConverter converter = new TSLintConverter();
+
+            Mock<IResultLogWriter> mockWriter = new Mock<IResultLogWriter>();
+
+            Assert.Throws<ArgumentNullException>(() => converter.Convert(null, mockWriter.Object, LoggingOptions.None));
+            Assert.Throws<ArgumentNullException>(() => converter.Convert(new MemoryStream(), null, LoggingOptions.None));
+        }
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintConverterTest_Convert_S2S_ValidInput()
+        {
+            Mock<ITSLintLoader> mockLoader = new Mock<ITSLintLoader>();
+
+            mockLoader.Setup(loader => loader.ReadLog(It.IsAny<Stream>())).Returns(CreateTestTSLintLog());
+
+            Mock<IResultLogWriter> mockWriter = new Mock<IResultLogWriter>();
+            mockWriter.Setup(writer => writer.Initialize(It.IsAny<string>(), It.IsAny<string>()));
+            mockWriter.Setup(writer => writer.WriteTool(It.IsAny<Tool>()));
+            mockWriter.Setup(writer => writer.WriteFiles(It.IsAny<IDictionary<string, FileData>>()));
+            mockWriter.Setup(writer => writer.OpenResults());
+            mockWriter.Setup(writer => writer.CloseResults());
+            mockWriter.Setup(writer => writer.WriteResults(It.IsAny<List<Result>>()));
+
+            TSLintConverter converter = new TSLintConverter(mockLoader.Object);
+
+            converter.Convert(new MemoryStream(), mockWriter.Object, LoggingOptions.None);
+
+            mockLoader.Verify(loader => loader.ReadLog(It.IsAny<Stream>()), Times.Once);
+            mockWriter.Verify(writer => writer.Initialize(It.IsAny<string>(), It.IsAny<string>()), Times.Once);
+            mockWriter.Verify(writer => writer.WriteTool(It.IsAny<Tool>()), Times.Once);
+            mockWriter.Verify(writer => writer.WriteFiles(It.IsAny<IDictionary<string, FileData>>()), Times.Once);
+            mockWriter.Verify(writer => writer.OpenResults(), Times.Once);
+            mockWriter.Verify(writer => writer.CloseResults(), Times.Once);
+            mockWriter.Verify(writer => writer.WriteResults(It.IsAny<List<Result>>()), Times.Once);
+        }
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintConverterTest_HelperMethods_NullInput()
+        {
+            TSLintConverter converter = new TSLintConverter();
+
+            Assert.Throws<ArgumentNullException>(() => converter.CreateResult(null));
+        }
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintConverterTest_HelperMethods_ValidInput()
+        {
+            TSLintConverter converter = new TSLintConverter();
+            TSLintLog tSLintLog = CreateTestTSLintLog();
+
+            Result actualResult = converter.CreateResult(tSLintLog[0]);
+
+            Assert.True(CompareResultToTestData(actualResult));
+        }
+
+    }
+}
--- a/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
+++ b/src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
@@ -6,14 +6,17 @@ using System.IO;
 using System.Collections.Generic;
 using System.Text;
 using Microsoft.CodeAnalysis.Sarif.Writers;
+using Microsoft.CodeAnalysis.Sarif.Converters.TSLintObjectModel;
 using Moq;
 using Xunit;
+using FluentAssertions;
+
 
 namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
     public class TSLintConverterTests
     {
-        public TSLintLog CreateTestTSLintLog()
+        private TSLintLog CreateTestTSLintLog()
         {
             TSLintLog tsLintLog = new TSLintLog();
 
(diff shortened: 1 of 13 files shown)


src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                IList<Replacement> replacements = new List<Replacement>();
+
+                foreach (TSLintLogFix fix in tslEntry.Fixes)
+                {
+                    Replacement replacement = new Replacement()
+                    {
+                        Offset = fix.InnerStart,
+                        DeletedLength = fix.InnerLength,
+                        InsertedBytes = fix.InnerText
```
boAndron: I didn't get a clarification on this part in the initial CR. Is InsertedBytes supposed to be pre-base64encoded, or does the writer take care of that? I can easily add
```var plainTextBytes = System.Text.Encoding.UTF8.GetBytes(fix.InnerText);```
```replacement.InsertedBytes = System.Convert.ToBase64String(plainTextBytes);```

File: src/Sarif.Converters.UnitTests/TSLintConverterTests.cs [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+            return true;
+        }
+
+        //[Fact]
+        //[Trait("Category", "E2E")]
+        //public void SimpleE2ETest_ManualValidation()
+        //{
+        //    ToolFormatConverter converter = new ToolFormatConverter();
+        //    converter.ConvertToStandardFormat(ToolFormat.TSLint, @"E:\data\TSLint\TSLint.Results.json", @"E:\data\TSLint\TSLint.Results.sarif");
+        //}
```
boAndron: Commented out (manual verification). I can remove it if you prefer.

src/Sarif.Converters/TSLintConverter.cs
lgolding: Hi Bogdan! The SARIF spec [says](https://rawgit.com/sarif-standard/sarif-spec/master/Static%20Analysis%20Results%20Interchange%20Format%20(SARIF).html#replacement-insertedbytes) that the inserted bytes must be Base64-encoded:

> If the file into which the bytes are to be inserted is a binary file, the value of the `insertedBytes` string shall be the MIME Base64 encoding of the byte sequence to be inserted.
>
> If the file into which the bytes are to be inserted is a text file, the characters to be inserted shall first be encoded in UTF-8. The value of the `insertedBytes` string shall be the MIME Base64 encoding of the resulting UTF-8 byte sequence.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+            var results = new List<Result>();
+            foreach(TSLintLogEntry entry in tsLintLog)
+            {
+                results.Add(CreateResult(entry));
+            }
+
+            var fileInfoFactory = new FileInfoFactory(MimeType.DetermineFromFileExtension, loggingOptions);
+            Dictionary<string, FileData> fileDictionary = fileInfoFactory.Create(results);
+            
+            if (fileDictionary != null && fileDictionary.Any())
```
jeff-bacon: Can simply by using null propagation.  
fileDictionary?.Any()

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Json;
+using System.IO;
+
+[assembly: InternalsVisibleTo("Sarif.Driver.UnitTests.dll,PublicKey=0024000004800000940000000602000000240000525341310004000001000100433fbf156abe971" +
```
lgolding: In our code base, we conventionally place `InternalsVisibleTo` attributes in the file `<projectDirectory>\Properties\AssemblyInfo.cs`. That file already exists in the `Sarif.Converters` project, and it already grants access to `Sarif.Driver.UnitTests.dll`, so this attribute should not be necessary.

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly XmlObjectSerializer Serializer;
+
+        public TSLintLoader()
+        {
+            Serializer = new DataContractJsonSerializer(typeof(TSLintLog));
+        }
+
+        public TSLintLoader(XmlObjectSerializer serializer)
+        {
+            Serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
```
lgolding: Oooohhh, C# 7.0 `throw` expressions! I like. I'm gonna start doing that.

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// Necessary because the TSLint json contains multiple patterns for fix, i.e.
+        /// "fix":{"innerStart":4429,"innerLength":0,"innerText":"\r\n"},
+        /// "fix":[{"innerStart":4429,"innerLength":0,"innerText":"\r\n"}],
+        /// and "fix":[{"innerStart":4429,"innerLength":0,"innerText":"\r\n"},{"innerStart":4429,"innerLength":0,"innerText":"\r\n"}],
+        /// 
+        /// We need to normalize it so that every instance of fix is a list.
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns></returns>
+        internal Stream NormalizeTSLintFixFormat(Stream stream)
```
lgolding: This can be `private`.

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            stream = stream ?? throw new ArgumentNullException(nameof(stream));
+
+            string contents;
+            using (StreamReader reader = new StreamReader(stream))
+            {
+                contents = reader.ReadToEnd();
+            }
+
+            Regex regex = new Regex("(\"fix\":{[^}]+},)");
```
lgolding: Do you need the outer grouping parens in the regex? Wouldn't this work?
```
Regex regex = new Regex("\"fix\":{[^}]+},");
```

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -0,0 +1,75 @@
...
+            
+            string[] tokens = regex.Split(contents);
+
+            for (int i = 0; i < tokens.Length; i++) 
+            {
+                if (tokens[i].Contains("\"fix\":{"))
+                {
+                    tokens[i] = tokens[i].Replace("\"fix\":{", "\"fix\":[{").Replace("},", "}],");
+                }
+            }
```
lgolding: Maybe I'm badly misreading this code, but I don't think it works. `Regex.Split` returns an array of substrings that _excludes_ the delimiting regex. So if you have

```
string s = "123fix4fixfix5678";
var regex = new Regex("fix");
string[] substrings = regex.Split(s);
```
then `substrings` will contain
```
[ "123", "4", "", "5678" ]
```
... and none of the substrings will contain `"fix"`.

src/Sarif.Converters/TSLintLoader.cs
lgolding: The same comment applies to TSLintConverter.cs.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+    public class TSLintConverter : ToolFileConverterBase
+    {
+        private readonly ITSLintLoader loader;
+        
+        public TSLintConverter()
+        {
+            loader = new TSLintLoader();
+        }
+
+        public TSLintConverter(ITSLintLoader loader)
```
lgolding: This ctor is intended only for use in unit tests, so make it `internal`, and add a comment to that effect.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+            var results = new List<Result>();
+            foreach(TSLintLogEntry entry in tsLintLog)
+            {
+                results.Add(CreateResult(entry));
+            }
+
+            var fileInfoFactory = new FileInfoFactory(MimeType.DetermineFromFileExtension, loggingOptions);
+            Dictionary<string, FileData> fileDictionary = fileInfoFactory.Create(results);
+            
+            if (fileDictionary?.Any() == true)
```
lgolding: `== true` is not needed.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                output.WriteFiles(fileDictionary);
+            }
+
+            output.OpenResults();
+            output.WriteResults(results);
+            output.CloseResults();
+        }
+
+        internal Result CreateResult(TSLintLogEntry tslEntry)
```
lgolding: We generally avoid abbreviations in this code base. Since there is no other "entry" with which this one might be confused, just call it `entry`.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+                case "ERROR":
+                    result.Level = ResultLevel.Error;
+                    break;
+                case "WARN":
+                case "WARNING":
+                    result.Level = ResultLevel.Warning;
+                    break;
+                case "DEFAULT":
+                    result.Level = ResultLevel.Default;
+                    break;
```
lgolding: What are the semantics when a TSLint log entry has a `RuleSeverity` of `"DEFAULT"`? It might not match what SARIF's `ResultLevel.Default` means.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+                default:
+                    result.Level = ResultLevel.NotApplicable;
+                    break;
```
lgolding: What are the semantics when a TSLint log entry has no `RuleSeverity`? Does that really mean that the rule is "not applicable"?

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+            int length = tslEntry.EndPosition.Position - tslEntry.StartPosition.Position + 1;
+            region.Length = length > 0 ? length : 0;
+
+            Uri analysisTargetUri = new Uri(tslEntry.Name, UriKind.Relative);
+
+            PhysicalLocation analysisTarget = new PhysicalLocation(uri: analysisTargetUri, uriBaseId: null, region: region);
+            Location location = new Location();
+            location.AnalysisTarget = analysisTarget;
+
+            result.Locations.Add(location);
```
lgolding: Rather than creating an empty list, creating a Location object, and adding it to the list, you could write:
```
result.Locations = new List<Location> { location };
```

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            result.Fixes = new List<Fix>();
+
+            if (tslEntry.Fixes?.Any() == true)
```
lgolding: `== true` is not needed.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                IList<Replacement> replacements = new List<Replacement>();
+
+                foreach (TSLintLogFix fix in tslEntry.Fixes)
```
lgolding: To understand if this code is right, I need to understand the semantics of `Fixes`. Is the idea that, to correct the problem identified by the log entry, you must make _all the replacements specified by the Fixes_? In other words, does a TSLint "fix" have the same semantics as a SARIF "replacement"?

src/Sarif.Converters/TSLintLoader.cs
lgolding: Ok, I'm wrong the code does work. But it relies on what I think is an obscure feature of [`Regex.Split`](https://msdn.microsoft.com/en-us/library/8yttk7sy(v=vs.110).aspx):

> If capturing parentheses are used in a Regex.Split expression, any captured text is included in the resulting string array. For example, if you split the string "plum-pear" on a hyphen placed within capturing parentheses, the returned array includes a string element that contains the hyphen.

So, disregard my PR comment on Line 58 above, but please add a source code comment there to explain why this code works.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -45,4 +45,8 @@
     <ProjectReference Include="..\Sarif\Sarif.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="TSLintObjectModel\" />
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
boAndron: It isn't empty, it contains the Object Model classes.

src/Sarif.Converters/TSLintConverter.cs
boAndron: It is. Cannot implicitly convert from bool? to bool.

src/Sarif.Converters/TSLintConverter.cs
boAndron: I have no idea, I couldn't find a spec. I can remove this case if you prefer.

src/Sarif.Converters/TSLintConverter.cs
boAndron: I took NotApplicable to mean that the RuleSeverity property is not applicable, not the rule. Like I mentioned before, I couldn't find a spec for the TSLint log format, and all the logs I had had only "ERROR" as a level. If you prefer a different value, or no value as the default, let me know.

src/Sarif.Converters/TSLintConverter.cs
boAndron: See above. Doesn't compile without it because tslEntry.Fixes?.Any() is a ternary value (true, false, or null)

src/Sarif.Converters/TSLintConverter.cs
boAndron: Couldn't find a spec, so I inferred from the logs I had. Semantics seem to fit. 
- innerStart aligns with startPosition.position (which is a 0-initiated offset into the file). So it fits for Offset
- inner length appears a 0 when something is to be inserted or as a positive integer when something is to be deleted or replaced, i.e.
```"failure": "Consecutive blank lines are forbidden",
    "fix": [
      {
        "innerStart": 78,
        "innerLength": 2,
        "innerText": ""
      }
    ],```
or
```"failure": "Forbidden 'var' keyword, use 'let' or 'const' instead",
    "fix": {
      "innerStart": 210,
      "innerLength": 3,
      "innerText": "let"
    },```
- so innerLenght fits DeletedLength
- InnerText and InnerBytes seem to differ only in encoding, which I've fixed
Let me know if you still have concerns about this

Comment (boAndron) [reply src/Sarif.Converters/TSLintConverter.cs]: Sorry for the bad formatting, GitHub seems to have different markdown rules.
For the first example "fix" the failure is "failure": "Consecutive blank lines are forbidden",

File: src/Sarif.Converters/TSLintLoader.cs (hunk 6 above)
Comment (boAndron) [reply src/Sarif.Converters/TSLintLoader.cs]: Exposed for test.

File: src/Sarif.Converters/TSLintConverter.cs (hunk 1 above)
Comment (lgolding) [reply src/Sarif.Converters/TSLintConverter.cs]: I see this is fixed.

File: src/Sarif.Converters.UnitTests/TSLintConverterTests.cs (hunk 2 above)
Comment (lgolding) [reply src/Sarif.Converters.UnitTests/TSLintConverterTests.cs]: I see this is fixed.

File: src/Sarif.Converters/TSLintConverter.cs (hunk 3 above)
Comment (lgolding) [reply src/Sarif.Converters/TSLintConverter.cs]: I see this is fixed.

File: src/Sarif.Converters.UnitTests/TSLintConverterTests.cs [hunk 18]
```diff
@@ -0,0 +1,162 @@
...
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+using Moq;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    public class TSLintConverterTests
+    {
+        public TSLintLog CreateTestTSLintLog()
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            testEntry.EndPosition = testEndPos;
+
+            tsLintLog.Add(testEntry);
+
+            return tsLintLog;
+        }
+
+        // This is necessary because Microsoft.CodeAnalysis.Sarif.Result doesn't implement IEquatable
+        // Must be kept up to date with any changes to TSLintConverter.CreateResult
+        public bool CompareResultToTestData(Result result)
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        //[Fact]
+        //[Trait("Category", "E2E")]
+        //public void SimpleE2ETest_ManualValidation()
+        //{
+        //    ToolFormatConverter converter = new ToolFormatConverter();
+        //    converter.ConvertToStandardFormat(ToolFormat.TSLint, @"E:\data\TSLint\TSLint.Results.json", @"E:\data\TSLint\TSLint.Results.sarif");
+        //}
+
+        [Fact]
+        [Trait("Category", "Unit")]
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        public void TSLintConverterTest_Convert_S2S_NullInput()
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            TSLintConverter converter = new TSLintConverter();
```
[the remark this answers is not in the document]
lgolding: I see this is fixed. (I also fixed it in a couple more places.)

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            Mock<IResultLogWriter> mockWriter = new Mock<IResultLogWriter>();
+
+            Assert.Throws<ArgumentNullException>(() => converter.Convert(null, mockWriter.Object, LoggingOptions.None));
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        public void TSLintConverterTest_HelperMethods_NullInput()
+        {
+            TSLintConverter converter = new TSLintConverter();
+
+            Assert.Throws<ArgumentNullException>(() => converter.CreateResult(null));
+        }
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintConverterTest_HelperMethods_ValidInput()
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters.UnitTests/TSLintLoaderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            Assert.Throws<ArgumentNullException>(() => loader.ReadLog(null));
+        }
+
+        [Fact]
+        [Trait("Category", "Unit")]
+        public void TSLintLoaderTest_ReadLog_ValidInput()
+        {
+            Mock<XmlObjectSerializer> mockSerializer = new Mock<XmlObjectSerializer>();
+
+            string input = "[{\"endPosition\":{\"character\":1,\"line\":113,\"position\":4429},\"failure\":\"file should end with a newline\",\"fix\":{\"innerStart\":4429,\"innerLength\":0,\"innerText\":\"\r\n\"},\"name\":\"SecureApp/js/index.d.ts\",\"ruleName\":\"eofline\",\"ruleSeverity\":\"ERROR\",\"startPosition\":{\"character\":1,\"line\":113,\"position\":4429}}]";
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters/ITSLintLoader.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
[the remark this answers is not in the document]
lgolding: I see this is fixed. I pushed a change to put `System` usings first, following the convention in this code base.

src/Sarif.Converters/Sarif.Converters.csproj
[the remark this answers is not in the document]
lgolding: Oh, I see, this is one of those new-fangled project files that doesn't mention all the source files by name.

src/Sarif.Converters/TSLintObjectModel/TSLintLog.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
```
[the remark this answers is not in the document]
lgolding: I see this is fixed.

src/Sarif.Converters/TSLintLoader.cs
lgolding: I pushed a change to fix this.

src/Sarif.Converters/TSLintLoader.cs
lgolding: ok.

src/Sarif.Converters/TSLintLoader.cs
lgolding: I withdraw this comment; see below.

src/Sarif.Converters/TSLintLoader.cs
lgolding: I pushed a change to add the comment.

src/Sarif.Converters/TSLintConverter.cs
lgolding: I see this is fixed.

src/Sarif.Converters/TSLintConverter.cs
lgolding: You are correct.

src/Sarif.Converters/TSLintConverter.cs
lgolding: I see this is fixed.

src/Sarif.Converters/TSLintConverter.cs
lgolding: @boAndron Lacking any other information, I propose to use `ResultLevel.Note` for any TSLint `RuleSeverity` other than `"ERROR"`, `"WARN"`, or `"WARNING"`. In SARIF, that means "a purely informational log entry".

@michaelcfanning I'm going to push (and merge) that change, but feel free to chime in if you disagree.

src/Sarif.Converters/TSLintConverter.cs
lgolding: See comment above. I'm going to use `ResultLevel.Note` for this.

src/Sarif.Converters/TSLintConverter.cs
lgolding: I see this is fixed.

src/Sarif.Converters/TSLintConverter.cs
lgolding: You are correct.

src/Sarif.Converters/TSLintConverter.cs
lgolding: I think you are right about this.

Issue #637, opened 2017-10-17
Simplify key handling


lgolding:
We have been using a complicated key handling scheme where:

1. We check in a test key (`TestKey.snk`), which never changes, and which we use to sign only test assemblies.
2. Product assemblies grant `InternalsVisibleTo` test assemblies signed with `TestKey.snk`.
3. The build process generates a key file, `GeneratedKey.snk`, which is _not_ checked in, and which is used to sign product assemblies.

This scheme has the advantage that the product key is never exposed, but the disadvantage that versions of the SDK built by different developers have different keys.

We have decided that checking in the key is ok, because
- If you build the SDK yourself, you know what you have.
- If you get the SDK from NuGet.org, again you know what you have.

... and the only danger is if you acquire a copy of the SDK from somewhere else.

Other Microsoft projects, for example https://github.com/OfficeDev/Open-XML-Package-Editor-Power-Tool-for-Visual-Studio, take the simpler approach.

lgolding:
Closed in PR #638.

PR #638, opened 2017-10-17
#637: Simplify key handling

lgolding:
We have been using a complicated key handling scheme where:

1. We check in a test key (`TestKey.snk`), which never changes, and which we use to sign only test assemblies.
2. Product assemblies grant `InternalsVisibleTo` to test assemblies signed with `TestKey.snk`.
3. The build process generates a key file, `GeneratedKey.snk`, which is _not_ checked in, and which is used to sign product assemblies.

This scheme has the advantage that the product key is never exposed, but the disadvantage that versions of the SDK built by different developers have different keys.

We have decided that checking in the key is ok, because
- If you build the SDK yourself, you know what you have.
- If you get the SDK from NuGet.org, again you know what you have.

... and the only danger is if you acquire a copy of the SDK from somewhere else.

Other Microsoft projects, for example https://github.com/OfficeDev/Open-XML-Package-Editor-Power-Tool-for-Visual-Studio, take the simpler approach.


jeff-bacon:

michaelcfanning:
>   <PropertyGroup>

there appears to be a lot of duplication between this props file and build.props. can't we have each import a common core file?

---
Refers to: src/netcorebuild.props:2 in 421f9a8. [](commit_id = 421f9a8162ee1daabdb2d1794ef083ff1193f802, deletion_comment = False)

michaelcfanning:
I don't like where you've put key.snk. A \Build\ directory is confusing because we already have \bld\

Can't this file live at the root of source? How many files are there in this directory? One? Note that the SLN file is at the root of source, and it's a build file

lgolding:
@michaelcfanning I will push the (now) two files in the `src\build` directory up into `src`.

I will DRY out the .props file as well.

jeff-bacon:
>   <PropertyGroup>

This can and should be done.  I created this file and didn't have a good understanding of .props files at the time.  

---
In reply to: [338044891](https://github.com/Microsoft/sarif-sdk/pull/638#issuecomment-338044891) [](ancestors = 338044891)

---
Refers to: src/netcorebuild.props:2 in 421f9a8. [](commit_id = 421f9a8162ee1daabdb2d1794ef083ff1193f802, deletion_comment = False)

lgolding:
@jeff-bacon no problem, stand by for these changes...

jeff-bacon:
GeneratedKey.snk was being generated in the root directory where the BuildAndTest.cmd lives.  Maybe this would be a good location for it?

---
In reply to: [338045289](https://github.com/Microsoft/sarif-sdk/pull/638#issuecomment-338045289) [](ancestors = 338045289)

PR #639, opened 2017-10-19
Bundle PREfastXmlSarifConverter.dll with Sarif.Converters NuGet Package

jeff-bacon:
Updated version.  Fixed issue where PREfastXmlSarifConverter.dll was not being bundled with the Sarif.Converters nuget package.


BuildAndTest.cmd
```diff
@@ -64,60 +64,63 @@ if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
-msbuild /verbosity:minimal /target:rebuild src\Everything.sln /p:"Configuration=%Configuration%" /p:"Platform=Any CPU" /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
+msbuild /verbosity:minimal /target:rebuild src\Everything.sln /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
+set Platform=AnyCPU
```
jeff-bacon: Have to reset platform to 'AnyCPU' without a space after MSBUILD because MSBuild will remove the space from 'Any CPU' but dotnet pack does not.  This was causing the code to build twice during BuildAndTest.  MSBuild would be built to 'AnyCPU_Release' and dotnet pack and dotnet xunit would look for their source in 'Any CPU_Release'

jeff-bacon: Also, passing 'AnyCPU' without a space, to MSBuild, results in an error.  This is why it has to start off as 'Any CPU' and change to 'AnyCPU' after MSBuild

---
In reply to: [145827547](https://github.com/Microsoft/sarif-sdk/pull/639#discussion_r145827547) [](ancestors = 145827547)

BuildAndTest.cmd
```diff
@@ -64,60 +64,63 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+dotnet pack .\src\Sarif.Multitool\Sarif.Multitool.csproj %PackOptions%
 
 if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
 @REM Run all tests
 
-pushd .\src\Sarif.Converters.UnitTests && dotnet xunit && popd
+pushd .\src\Sarif.Converters.UnitTests && dotnet xunit -configuration Release && popd
```
jeff-bacon: Forgot to add -nobuild

BuildAndTest.cmd
lgolding: That's horrible ;-)

BuildAndTest.cmd
```diff
@@ -64,60 +64,63 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+msbuild /verbosity:minimal /target:rebuild src\Everything.sln /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
+set Platform=AnyCPU
 
 if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
 @REM Build Nuget packages
-dotnet pack .\src\Sarif\Sarif.csproj /p:PackageVersion=%Version%
+set PackOptions=--configuration %Configuration% --no-build /p:PackageVersion=%Version% /p:Platform=%Platform%
```
lgolding: Yay!

BuildAndTest.cmd
```diff
@@ -64,60 +64,63 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+dotnet pack .\src\Sarif.Multitool\Sarif.Multitool.csproj %PackOptions%
 
 if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
 @REM Run all tests
 
-pushd .\src\Sarif.Converters.UnitTests && dotnet xunit && popd
+pushd .\src\Sarif.Converters.UnitTests && dotnet xunit -nobuild -configuration Release && popd
```
lgolding: Yay `nobuild`!

SetCurrentVersion.cmd
```diff
@@ -1,9 +1,9 @@
...
-set MINOR_PREVIOUS=5
-set PATCH_PREVIOUS=46
+set MINOR_PREVIOUS=6
+set PATCH_PREVIOUS=0
 set PRERELEASE_PREVIOUS=
 
-set MAJOR=1
-set MINOR=6
+set MAJOR=2
+set MINOR=0
```
lgolding: Curious why this rates a major version change.

jeff-bacon: Good question.  This probably should have been done with the original NET Core merge but I overlooked it.  I believe that was a big enough change to warrant a major version change.  What do you think?

---
In reply to: [145850825](https://github.com/Microsoft/sarif-sdk/pull/639#discussion_r145850825) [](ancestors = 145850825)

BuildAndTest.cmd
jeff-bacon: It's so bad.  To make it worse, the %Platform% and %Configuration% properties override any platform and configuration property for MSBuild even if you hard code them in the msbuild parameters.  This is why I was able to remove some of the msbuild parameters.  I was confused when trying to figure that one out.

---
In reply to: [145850618](https://github.com/Microsoft/sarif-sdk/pull/639#discussion_r145850618) [](ancestors = 145850618)

SetCurrentVersion.cmd
lgolding: Sure, let's do it now.

PR #646, opened 2017-10-20
Fix #645: Viewer doesn't show ruleId unless rule metadata is present

### LINKED ISSUE #645: Viewer doesn't show ruleId unless rule metadata is present
Repro steps:

1. Open a SARIF log file that doesn't include rule metadata.

Expected: the `ruleId` from the result appears in the Code column of the Error List Window.

Actual: the Code column is blank.

### LINKED ISSUE #647: Viewer NRE when result lacks message/formattedRuleMessage but rule metadata is present
If a `result` lacks both the `message` and `formattedRuleMessage` properties, but a `rule` object corresponding to `result.ruleId` is present, the Viewer throws `NullReferenceException`.

lgolding:
Repro steps:

1. Open a SARIF log file that doesn't include rule metadata.

Expected: the `ruleId` from the result appears in the Code column of the Error List Window.

Actual: the Code column is blank.

Also:
- Fix #647: Viewer NRE when result lacks message/formattedRuleMessage but rule metadata is present
- Add `TSLint` to `-t` option help text.
- Add `launchSettings.json` (the file that contains the command line used by VS 2017 when you F5 a project) to `.gitignore`.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Sarif.Viewer
             }
 
             Tool = run.Tool.ToToolModel();
-            Rule = rule?.ToRuleModel(result.RuleId);
+            Rule = rule.ToRuleModel(result.RuleId);
             Invocation = run.Invocation.ToInvocationModel();
 
             if (String.IsNullOrWhiteSpace(run.Id))
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -209,7 +209,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             {
                 text = string.Empty;    // Ensure that it's not null.
 
-                if (rule != null)
+                if (rule != null && result.FormattedRuleMessage != null)
                 {
                     string formatId = result.FormattedRuleMessage.FormatId;
                     string messageFormat;
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@ bld/
 # Visual Studio 2017 browsing database files
 .BROWSE.VC.DB-*
 
+# Visual Studio 2017 Application/Debug properties
+launchSettings.json
+
 # VS Code configuration
 .vscode/
 
--- a/src/Sarif.Multitool/ConvertOptions.cs
+++ b/src/Sarif.Multitool/ConvertOptions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
         [Option(
             't',
             "tool",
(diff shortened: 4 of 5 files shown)


src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -59,7 +59,7 @@ public SarifErrorListItem(Run run, Result result, string logFilePath, ProjectNam
             }
 
             Tool = run.Tool.ToToolModel();
-            Rule = rule?.ToRuleModel(result.RuleId);
+            Rule = rule.ToRuleModel(result.RuleId);
```
lgolding: This is what fixes the bug. When `rule` is `null`, the `ToRuleModel` extension method synthesizes a `RuleModel` object whose `id` is the argument passed in (`result.RuleId`). But as the code stood, the extension method was never called when `rule` was `null`, and the `RuleModel` object ended up as `null`.

src/Sarif/ExtensionMethods.cs
```diff
@@ -209,7 +209,7 @@ public static string GetMessageText(this Result result, IRule rule, bool concise
             {
                 text = string.Empty;    // Ensure that it's not null.
 
-                if (rule != null)
+                if (rule != null && result.FormattedRuleMessage != null)
```
lgolding: An unrelated bug: you could get an NRE if there _was_ a rule object, but `result` did not have a `FormattedRuleMessage` property.

*sigh* I should add a test for this one too...

lgolding: Ok, I filed #646 for this, and added the unit test.

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
michaelcfanning: Why is it ok for an extension method to behave this way on being passed a null item? That seems to break the principle of least surprise.



---
In reply to: [146093187](https://github.com/Microsoft/sarif-sdk/pull/646#discussion_r146093187) [](ancestors = 146093187)

src/Sarif/ExtensionMethods.cs
jeff-bacon: Looks good

---
In reply to: [146093210](https://github.com/Microsoft/sarif-sdk/pull/646#discussion_r146093210) [](ancestors = 146093210)

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
jeff-bacon: Should `ToRuleModel` run if rule is null?  It sounds like it's possible but very confusing considering how this differs from how methods of the class itself function.

---
In reply to: [146093187](https://github.com/Microsoft/sarif-sdk/pull/646#discussion_r146093187) [](ancestors = 146093187)

lgolding: @michaelcfanning I agree, `ToRuleModel` is oddly implemented. This is the smallest change that fixes the bug.

Issue #651, opened 2017-10-21
SARIF schema incorrectly says that fix.description is required


lgolding:
The spec says that a `fix` object **SHOULD** (not **SHALL**) contain a description, so the schema should not mark `fix.description` as `required`.

@michaelcfanning FYI. I discovered this while writing the TSLint converter functional tests.

lgolding:
Fixed by PR #660.

Issue #649, opened 2017-10-21
TSLint converter emits arrays containing 0 fixes


lgolding:
Given this TSLint log entry:
```
{
  "endPosition": {
    "character":11,
    "line":8,
    "position":97
  },
  "failure":"Forbidden reference to reserved keyword: any",
  "name":"C:/a2/_work/6/s/SecureApp/ts/test-data/NoReservedKeywords/NoReservedKeywordsTestInput-any.ts",
  "ruleName":"no-reserved-keywords",
  "ruleSeverity":"ERROR",
  "startPosition":{
    "character":8,
    "line":8,
    "position":94
  }
}
```
the TSLint converter emits this SARIF result:
```
{
  "ruleId": "no-reserved-keywords",
  "level": "error",
  "message": "Forbidden reference to reserved keyword: any",
  "locations": [
     {
      "analysisTarget": {
        "uri": "file:///C:/a2/_work/6/s/SecureApp/ts/test-data/NoReservedKeywords/NoReservedKeywordsTestInput-any.ts",
        "region": {
          "startLine": 9,
          "startColumn": 9,
          "endLine": 9,
          "endColumn": 12,
          "offset": 94,
          "length": 4
        }
      }
    }
  ],
  "fixes": []
}
```
Note the empty `fixes` array at the end. But this is invalid SARIF (the `fixes` array must contain at least one element) so the validation tests fail.

If there are no fixes, the converter must refrain from emitting the `fixes` array.

lgolding:
The root cause is that the SARIF schema marks the `fix.description` property as `required`. That means the SARIF writer will emit this property even if a `null` value is supplied (as the TSLint converter does).

The schema is incorrect. The spec says that a `fix` object **SHOULD** (not **SHALL**) contain a `description` property. 

The real fix for this is to update the schema. But I don't want to do that in the same PR as the one that adds the TSLint converter functional tests. So for now I'm going to change the converter to emit `""` for `fix.description`.

Issue #648, opened 2017-10-21
TSLint converter doesn't handle all "fix" cases


lgolding:
The TSLint converter uses a regex in an attempt to ensure that all `fix` objects are wrapped in arrays (see method `TSLoader.NormalizeTSLintFixFormat`). For example, it converts
```
    "fix": {
      "innerStart": 169728,
      "innerLength": 0,
      "innerText": " "
    },
```
to
```
    "fix": [
      {
      "innerStart": 169728,
      "innerLength": 0,
      "innerText": " "
      }
    ],
```

But there are two problems:

1. The regex ignores the possibility that the `fix` object might have this structure:
```
    "fix": {
      "innerRuleName": "no-consecutive-blank-lines",
      "innerReplacements": [
        {
          "innerStart": 467,
          "innerLength": 2,
          "innerText": ""
        }
      ]
    },
```
This causes the converter to fail on the file TSLint-03.json.

**NOTE** I can't attach a `.json` file to a GitHub issue -- ask me for the failing files if you need to see them.

2. The regex (which attempts to match the `{` following `"fix":` with the next `}`) does not account for the fact that the `innerText` element might contain braces, for example:
```
    "fix": {
      "innerStart": 597,
      "innerLength": 3,
      "innerText": "{}"
    },
```
This causes the converter to fail on the file TSLint-06.json.

In any case, [JSON is not a regular language](https://cstheory.stackexchange.com/questions/3987/is-json-a-regular-language) so attempting to parse it with a regex is doomed to fail. I don't know if the subset of JSON emitted by TSLint is regular.

PR #652, opened 2017-10-22
Fix #644: Add functional tests for TSLint converter

### LINKED ISSUE #644: Add functional tests for TSLint converter

### LINKED ISSUE #650: TSLint converter emits null fix description
Given this TSLint log entry:
```
{
  "endPosition": {
    "character": 1,
    "line": 5,
    "position": 79
  },
  "failure": "Consecutive blank lines are forbidden",
  "fix": [
    {
      "innerStart": 78,
      "innerLength": 2,
      "innerText": ""
    }
  ],
  "name": "SecureApp/ts/test-data/NoCookies/NoCookiesFailingTestInput.ts",
  "ruleName": "no-consecutive-blank-lines",
  "ruleSeverity": "ERROR",
  "startPosition": {
    "character": 0,
    "line": 5,
    "position": 78
  }
}
```
the TSLint converter emits this SARIF result:
```
{
  "ruleId": "no-consecutive-blank-lines",
  "level": "error",
  "message": "Consecutive blank lines are forbidden",
  "locations": [
    {
      "analysisTarget": {
        "uri": "SecureApp/ts/test-data/NoCookies/NoCookiesFailingTestInput.ts",
        "region": {
          "startLine": 6,
          "startColumn": 1,
          "endLine": 6,
          "endColumn": 2,
          "offset": 78,
          "length": 2
        }
      }
    }
  ],
  "fixes": [
    {
      "description": null,
      "fileChanges": [
        {
          "uri": "SecureApp/ts/test-data/NoCookies/NoCookiesFailingTestInput.ts",
          "replacements": [
            {
              "offset": 78,
              "deletedLength": 2,
              "insertedBytes": ""
            }
          ]
        }
      ]
    }
  ]
},
```
Note the `null` value for `fixes[0].description`. This is not valid SARIF: `fix.description` must be a `string`. Since TSLint does not provide a fix description, the converter must refrain from emitting one. (It could provide an empty string `""` for the value, but there's no point to that.

lgolding:
In the course of implementing these tests, I found that the TSLint converter could not convert two of the 11 test files that Matt and Bogdan supplied. I filed #648, "TSLint converter doesn't handle all 'fix' cases", to track this.

From the 9 of 11 files that the converter _could_ convert, it turned out that the converter was emitting invalid SARIF for some of them, causing the validation tests to fail. I filed #649, "TSLint converter emits arrays containing 0 fixes", and #650, "TSLint converter emits null fix description", and fixed them as part of this PR.

The root cause of #650 is actually a bug in the SARIF schema. It incorrectly marks the property `fix.description` as `required`. I filed #651, "SARIF schema incorrectly says that fix.description is required" to track this. I'll fix that next, but I didn't want to do that as part of this PR. For now, I fixed #650 by having the converter emit the empty string `""` instead of `null` for `fix.description`. This in turn caused one of the converter unit tests to fail, so I adjusted the unit test to match the current behavior. When I fix the schema, I'll revert those changes.

**UPDATE** I pushed another change to fix part of #648. (#648 is really two bugs and I might go back and split it up.) The part I fixed was that `TSLintLoader` now uses Newtonsoft.Json, rather than a regex replacement, to wrap individual `fix` objects in an array. This allows us to add the functional test data file `TSLint-06.json` (which previously failed due to this bug) into the PR.

We still need to fix the other part of #648 (the part related to `innerReplacement`). We'll do that in a separate PR. When we do that, we'll be able to add the functional test data file `TSLint-03.json` (which fails due to that bug).


src/Sarif.FunctionalTests/Sarif.FunctionalTests.csproj
```diff
@@ -629,6 +629,66 @@
...
+    </None>
+    <None Include="ConverterTestData\TSLint\TSLint-01.json.sarif">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="ConverterTestData\TSLint\TSLint-02.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="ConverterTestData\TSLint\TSLint-02.json.sarif">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
lgolding: Note that `TSLint-03.json` is missing. It fails due to the other half of #648 (the part I haven't fixed yet).

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -33,55 +36,107 @@ public TSLintLog ReadLog(Stream input)
...
+        private void NormalizeEntry(JObject entry)
+        {
+            JProperty fixProperty = entry.Properties().SingleOrDefault(p => p.Name.Equals("fix"));
+            if (fixProperty != null)
+            {
+                NormalizeFixProperty(fixProperty);
+            }
         }
 
+        private static void NormalizeFixProperty(JProperty fixProperty)
```
boAndron: Cool. This seems a lot safer and more elegant than the Regex fix.

michaelcfanning:
>           "mimeType": "text/typescript"

why don't these test files have embedded files in them? should they?

---
Refers to: src/Sarif.FunctionalTests/ConverterTestData/TSLint/TSLint-08.json.sarif:11 in b375e3a. [](commit_id = b375e3ab97810658398ed104754bd07dcdc8b4d5, deletion_comment = False)

PR #655, opened 2017-10-23
appveyor build fix

jeff-bacon:


src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -11,7 +11,7 @@
 using System.Xml;
 using Microsoft.CodeAnalysis.Sarif.Writers;
 
-namespace Microsoft.CodeAnalysis.Sarif.Converters
+namespace Microsoft.CodeAnalysis.Sarif.Converters
```
lgolding: Looks like you accidentally added a trailing space on this line.

jeff-bacon: I don't believe AppVeyor fires a build unless you change a code file.  Config files and scripts don't fire the CI build last I checked.

PR #659, opened 2017-10-28
Fix #648: TSLint converter doesn't handle innerReplacements

### LINKED ISSUE #648: TSLint converter doesn't handle all "fix" cases

lgolding:
Issue #648 actually reported two distinct issues, and I should have filed them separately. This PR addresses item #&#8203;1 within this issue: `TSLintLoader` does not handle log files in which the `"fix"` object contains an array `"innerReplacements"`.

Item #&#8203;2 within this issue (`TSLintLoader` improperly parsed `"fix"` objects whose `"innerText"` was `"{}"`) was fixed in a previous PR. Therefore this PR closes #648 entirely.

With this fix, we can add the one functional test data file which was previously omitted because this bug caused it to fail: `TSLint-03.json`.

This concludes the addition of TSLint support to the SARIF SDK.


src/Sarif.Converters.UnitTests/TSLintLoaderTests.cs
```diff
@@ -160,6 +160,73 @@ public void TSLintLoader_NormalizeLog_WrapsSingleFixInArray()
...
+                {
+                    ""endPosition"": {
+                        ""character"": 4,
+                        ""line"": 56,
+                        ""position"": 1876
+                    },
+                    ""failure"": ""trailing whitespace"",
+                    ""fix"": {
+                        ""innerRuleName"": ""no-trailing-whitespace"",
+                        ""innerReplacements"": [
```
lgolding: This is the construct that the original TSLint converter didn't handle.

src/Sarif.Converters/TSLintLoader.cs
```diff
@@ -73,8 +73,26 @@ public TSLintLog ReadLog(Stream input)
         // "fix":[{"innerStart":4429,"innerLength":0,"innerText":"\r\n"}]
         // "fix":[{"innerStart":4429,"innerLength":0,"innerText":"\r\n"},{"innerStart":4429,"innerLength":0,"innerText":"\r\n"}]
         //
+        // The following pattern also occurs, although the most recent version of the TSLint
+        // source code does not appear to support it:
+        //
+        // "fix": {
+        //   "innerRuleName": "no-trailing-whitespace",
```
boAndron: I might be wrong, but from the code below it looks like we're throwing innerRuleName away. Is that intended?

lgolding: @boAndron Yes, we don't have a slot for that.

boAndron: Got it, thanks!

Issue #658, opened 2017-10-25
Build process no longer regenerates object model


lgolding:
The SARIF SDK build process is supposed to regenerate the SARIF object model classes from the schema. But due to changes made to the project file `Sarif.csproj` in the course of the .NET Core work (commit 0ef89c6), this no longer happens. The reason is that this commit removed the line

    <Import Project="ToDotNet\ToDotNet.targets" />

from the project file. This is the line that runs the object model generation tool.

I don't know if this line was removed by mistake, or if it did it not work in .NET Core. Either way, we need it back.

lgolding:
Fixed by PR #660.

PR #660, opened 2017-10-29
Fix #631, #651, #658: Schema changes and build fixes

### LINKED ISSUE #631: Add run.properties to schema
This bug was originally filed against the SARIF specification: https://github.com/oasis-tcs/sarif-spec/issues/32. The spec is correct; it's the schema (which is part of the SARIF SDK) that's wrong. So we're closing the spec bug and tracking this issue here, in the sarif-sdk repo (where we will definitely fix it).

@michaelcfanning @mj1856

lgolding:
The primary purpose of this PR is to make two fixes to the JSON schema:

#631: The spec says that the `run` object has a property bag, but the schema omitted it.
#651: The spec says that a `fileChange` object **SHOULD** have a `description` property, but the schema required it.

The SARIF SDK's build process is supposed to regenerate the SARIF object model classes from the schema on every build. But it turned out that this was no longer happening:

#658: Build process no longer regenerates object model

... so the changes to the schema had no effect. The necessary `.targets` file had been removed from `Sarif.csproj` in the course of the .NET Core work, so I restored it.

But then it turned out that the `.targets` file as written did not work well with the new `.csproj` file format. The reason is that new `.csproj` files have a "default compilation" feature (see [Additions to csproj format for .NET Core](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj)), which doesn't work properly when your build dynamically adds files to the `Compile` item list. You can disable the default compilation feature, but I chose another course. Instead of running `ToDotNet.targets` in the course of the normal build of the `Sarif` project, I modified `BeforeBuild.cmd` to build the necessary target before building the solution. That ensures that all the source files are in place before `Sarif.csproj` is built, which in turn ensures that the default compilation feature works properly.


BuildAndTest.cmd
```diff
@@ -64,36 +64,37 @@ if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
-msbuild /verbosity:minimal /target:rebuild src\Everything.sln /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
-set Platform=AnyCPU
+msbuild /verbosity:minimal /target:BuildAndInjectObjectModel src\Sarif\Sarif.csproj /fileloggerparameters:Verbosity=detailed
+if "%ERRORLEVEL%" NEQ "0" (
```
lgolding: Here's where we introduce the separate build step I mentioned in the PR description. We rebuild the SARIF object model from the schema before attempting to compile the `Sarif` project.

BuildAndTest.cmd
```diff
@@ -64,36 +64,37 @@ if "%ERRORLEVEL%" NEQ "0" (
...
+goto ExitFailed
+)
 
+msbuild /verbosity:minimal /target:rebuild src\Everything.sln /filelogger /fileloggerparameters:Verbosity=detailed /p:"RunBinSkim=false" /p:"BinSkimVerboseOutput=true"
 if "%ERRORLEVEL%" NEQ "0" (
 goto ExitFailed
 )
 
+set Platform=AnyCPU
```
lgolding: Minor, unrelated change: check the result of the `msbuild` command immediately after the command finishes. Don't set the environment variable until after that.

src/Sarif/Autogenerated/Fix.cs
```diff
@@ -34,7 +34,7 @@ public SarifNodeKind SarifNodeKind
         /// <summary>
         /// A string that describes the proposed fix, enabling viewers to present a proposed change to an end user.
         /// </summary>
-        [DataMember(Name = "description", IsRequired = true)]
+        [DataMember(Name = "description", IsRequired = false, EmitDefaultValue = false)]
```
lgolding: The change to this autogenerated file is the result of changing the schema to not require the `fix.description` property.

src/Sarif/Autogenerated/Run.cs
```diff
@@ -5,6 +5,7 @@
 using System.CodeDom.Compiler;
 using System.Collections.Generic;
 using System.Runtime.Serialization;
+using Microsoft.CodeAnalysis.Sarif.Readers;
```
lgolding: The changes to the autogenerated files `Run.cs` and `RunEqualityComparer.cs` are the result of changing the schema to give the `run` object a property bag.

src/Sarif/CodeGenHints.json
```diff
@@ -365,6 +365,16 @@
       "kind": "DictionaryHint"
     }
   ],
+  "run": [
```
lgolding: The JSON schema-to-.NET code generator accepts (in addition to the schema file) a "hint" file that can modify the resulting generated code. In this case, we add a hint to the `run` object which tells it that the generated `Run` class should be declared to derive from `PropertyBagHolder`. See `Run.cs` for the result.

src/Sarif/Sarif.csproj
```diff
@@ -59,4 +59,5 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <Import Project="ToDotNet\ToDotNet.targets" />
```
lgolding: Restore the missing `.targets` file.

src/Sarif/ToDotNet/ToDotNet.targets
```diff
@@ -4,11 +4,6 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- Inject object model building. Must be after Microsoft.CSharp.targets is imported. -->
-    <CompileDependsOn>BuildAndInjectObjectModel;$(CompileDependsOn)</CompileDependsOn>
-  </PropertyGroup>
```
lgolding: We no longer build this target in the course of the compilation, so this isn't necessary.

src/Sarif/ToDotNet/ToDotNet.targets
```diff
@@ -57,12 +52,5 @@
 
     <Copy SourceFiles="@(_NotYetGeneratedFiles)"		
             DestinationFolder="$(_ObjectModelOutputDirectory)" />
-
-    <ItemGroup>
-      <!-- FileWrites are the set of files that get deleted on clean. -->
-      <FileWrites Include="$(_ObjectModelOutputDirectory)\*.cs" />
-      <!-- Compile are the set of files that go into the C# build. -->
-      <Compile Include="$(_ObjectModelOutputDirectory)\*.cs" />
-    </ItemGroup>
```
lgolding: Again, since we no longer build this target in the course of the compilation, there's no need to dynamically add `Compile` items. It was that dynamic addition of `Compile` items that wasn't working properly with the new-style `.csproj` files.

BeforeBuild.cmd
```diff
@@ -10,7 +10,14 @@ SETLOCAL
 dotnet restore src\Everything.sln --packages src\packages
 
 if "%ERRORLEVEL%" NEQ "0" (
-echo nuget restore failed
+echo nuget restore failed.
+goto ExitFailed
+)
+
+:: Generate the SARIF object model classes from the SARIF JSON schema.
+msbuild /verbosity:minimal /target:BuildAndInjectObjectModel src\Sarif\Sarif.csproj /fileloggerparameters:Verbosity=detailed
```
lgolding: Note: we do this in `BeforeBuild.cmd` rather than in `BuildAndTest.cmd` to ensure that this happens in AppVeyor as well as on the desktop.

BeforeBuild.sh
```diff
@@ -13,7 +13,16 @@ dotnet restore src/Sarif.Sdk.sln
 
 if [ $? != 0 ]
 then
-    echo nuget restore failed
+    echo nuget restore failed.
+    Exit
+    return $?
+fi
+
+dotnet msbuild /verbosity:minimal /target:BuildAndInjectObjectModel src/Sarif/Sarif.csproj /fileloggerparameters:Verbosity=detailed
```
lgolding: @michaelcfanning @jeff-bacon This will only work if the JSON Schema to .NET Code Generation Tool has been ported to .NET Core. Has it been?

jeff-bacon: Linux is still not supported for SARIF NET Core.  jschema has yet to be ported to net core because of it's dependency on Microsoft.CodeAnalysis.  

---
In reply to: [147596572](https://github.com/Microsoft/sarif-sdk/pull/660#discussion_r147596572) [](ancestors = 147596572)

src/Sarif/ToDotNet/ToDotNet.targets
jeff-bacon: This is interesting  bit of information to know.  Good catch.

---
In reply to: [147594338](https://github.com/Microsoft/sarif-sdk/pull/660#discussion_r147594338) [](ancestors = 147594338)

lgolding:
@jeff-bacon @michaelcfanning I commented out the SARIF OM generation on Linux before merging this.

PR #662, opened 2017-10-30
Fix JSon String Comparisions with JToken Deep Comparision

komal436:
Comparing JSon file outputs in string format doesn't work if the
generated output is not prettified.

Compare the Json Token Deeply by value.

Signed-off-by: Komal Seelam <komal436@gmail.com>


lgolding:
@komal436 Komal, would you please do the following:

1. File a bug explaining exactly what the failure case is.
2. Write a unit test that fails without your fix, but passes when you apply your fix.

komal436:
@lgolding Larry, 
1. Bug #665  is created to track the sarif log comparision.
2. I have added the tests to validate the Sarif Log comparison in the Sarif.FunctionalTests. Writing the sarif log in a string with proper indent was not easy.

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -121,29 +167,14 @@ private void RunConverter(StringBuilder sb, string toolFormat, string inputFileN
 
             string expectedSarif = File.ReadAllText(expectedFileName);
             string actualSarif = File.ReadAllText(generatedFileName);
-
-            if (expectedSarif == actualSarif)
+
```
lgolding: Remove white space from blank line.

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -103,6 +130,25 @@ private static string FormatFailureReason(StringBuilder sb, string toolName)
             return sb.ToString();
         }
 
+        private bool CanGeneratedSarifBeDeserializedToExpectedSarif(string generatedSarif, string expectedSarif)
```
lgolding: A couple of comments about this name:

1. This function is a predicate, a statement of something we expect to be true. The thing we expect to be true is that "generated SARIF can be deserialized to expected SARIF", so the function name should be `GeneratedSarifCanBeDeserializedToExpectedSarif`.

Note that this makes the code read smoothly in English:
```
if (GeneratedSarifCanBeDeserializedToExpectedSarif(actualSarif, expectedSarif)
{
    ...
```

lgolding: This method can be `static`.

lgolding: As a rule I like to put the higher-level functions first, followed by the functions that they use. So IMO we should move this function down below `RunConverter`.

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -121,29 +167,14 @@ private void RunConverter(StringBuilder sb, string toolFormat, string inputFileN
...
-                {
-                    return;
-                }
-                else
-                {
-                    File.WriteAllText(generatedFileName, actualSarif);
-                }
+                return;
+            }
+            else
```
lgolding: I know the code was already structure this way, but while you're in here, could you please fix this:

There is no need for the `else` clause. The code can be more simply written:
```
    if (!AreEquivalentSarifLogs(actualSarif, expectedSarif)
    {
        File.WriteAllText(generatedFileName, actualSarif);

        string errorMessage = "The output of the {0} converter did not match for input {1}.";
        sb.AppendLine(string.Format(CultureInfo.CurrentCulture, errorMessage, toolFormat, inputFileName));
        sb.AppendLine("Check differences with:");
        sb.AppendLine(GenerateDiffCommand(expectedFileName, generatedFileName));
    }
}
```

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -77,6 +78,32 @@ public void PREfastConverter_EndToEnd()
...
+        public void ValidateDeserializationAndNotPrettifiedSarifStringComparisionLogic()
+        {
+            string sarifComparatorFileFolder = "TestSarifComparator/NotPrettified";
+            string testDirectory = Path.Combine(TestDirectory, sarifComparatorFileFolder);
+            string generatedSarif = File.ReadAllText(Directory.GetFiles(testDirectory, "generated.sarif")[0]);
+            string expectedSarif = File.ReadAllText(Directory.GetFiles(testDirectory, "expected.sarif")[0]);
+            CanGeneratedSarifBeDeserializedToExpectedSarif(generatedSarif, expectedSarif).Should().Be(true);
+        }
+
+        [Fact]
```
lgolding: Remove this empty test.

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -77,6 +78,32 @@ public void PREfastConverter_EndToEnd()
             BatchRunConverter(ToolFormat.PREfast);
         }
 
+        [Fact]
+        public void ValidateDeserializationAndPrettifiedSarifStringComparisionLogic()
```
lgolding: I appreciate your thinking about this, but I'm uncomfortable with this. In general, it's not necessary to test your test code (because where would you stop? would you test the code that tests your test code as well?) And in this particular case, you have to go to a lot of trouble -- four new test files and two new tests.

I would remove these tests and the new test files, and then I'd just do a quick hand verification of your logic as follows:

1. Run `BuildAndTest.cmd` -- this ensures that all the tests you expect to pass, do continue to pass.
2. Modify one of the functional test .sarif files -- say `ConverterTestData\AndroidStudio\AndroidStudio_issueLog1_raw.xml.sarif` -- by modifying the white space. Adding a single extra space should suffice.
3. Rerun the tests. It should still pass even though the string contents of the `.xml.sarif` file and the `.xml.sarif.actual` file will be different.
4. Undo your change to the test file.

komal436: Sure. I will follow the TDD process. I have removed the test files and manually validate the changes.

src/Sarif.FunctionalTests/SarifConverterTests.cs
komal436: Removed the tests and manually validated the changes.

src/Sarif.FunctionalTests/SarifConverterTests.cs
komal436: Fixed.

src/Sarif.FunctionalTests/SarifConverterTests.cs
komal436: Fixed.

src/Sarif.FunctionalTests/SarifConverterTests.cs
komal436: Fixed

Issue #623, opened 2017-09-21
Path rebase fails if path mentioned in SARIF file differs from resolved path only in drive letter


lgolding:
When you double-click on a result in a file that doesn't exist on your disk at the specified location, the viewer prompts with an Open File dialog. When you select the file in the correct location, the viewer should open it. But it doesn't.

Repro steps:

1. Unzip this archive to location C:\ (the root of the C: drive), which will create folder C:\EmbeddedContentDemo.

[EmbeddedContentDemo.zip](https://github.com/Microsoft/sarif-sdk/files/1322347/EmbeddedContentDemo.zip)

2. Open VS 2017 with the latest VSIX viewer installed.

3. Select menu item **Tools**, **Open Static Analysis Log File**, **Open SARIF file**. The "Open Static Analysis Results Interchange Format (SARIF) file" dialog opens.

4. Select file C:\EmbeddedContentDemo\2-IncorrectPaths.sarif, and then click **Open**. A single result appears in the Error List window.

5. Double-click the result. The "Locate missing file" dialog opens.

6. Select file C:\EmbeddedContentDemo\src\FileWithResult.cpp, and then click **Open**.

Expected: The viewer should open the selected file and highlight the line with the result.

Actual: An exception occurs:

![exception](https://user-images.githubusercontent.com/11762653/30712790-0ebf2004-9ec2-11e7-819d-09a926ad21d9.png)

Here is the ActivityLog file mentioned in the exception (renamed to .xml.txt because you can't attach an XML file to an issue):

[ActivityLog.xml.txt](https://github.com/Microsoft/sarif-sdk/files/1322375/ActivityLog.xml.txt)

The relevant part is at the end:
```
  <entry>
    <record>650</record>
    <time>2017/09/21 18:41:51.379</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.ArgumentOutOfRangeException: StartIndex cannot be less than zero.
Parameter name: startIndex
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Microsoft.Sarif.Viewer.CodeAnalysisResultManager.GetRebaselinedFileName(String uriBaseId, String fileName) in E:\src\sarif-sdk\src\Sarif.Viewer.VisualStudio\CodeAnalysisResultManager.cs:line 443
[...]

PR #670, opened 2017-11-08
Refactor: Inject dependencies into CodeAnalysisResultsManager

lgolding:
This is Step 1 of my plan for fixing Bug #623, and getting `CodeAnalysisResultManager` under test in the process. There will be four PRs, as follows:

1.	Inject dependencies into `CodeAnalysisResultManager`. <= THIS PR
2.	Write unit tests for `CodeAnalysisResultManager.GetRebaselinedFileName`.
3.	Fix Bug #623 (and add unit test).
4.	Perform further refactoring on `CodeAnalysisResultManager.GetRebaselinedFileName`.

The idea behind Step 4 is that after Step 3, we’ll still have this really nasty method, but now we’ll be able to safely refactor it.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/CommonDialogFactory.cs
+++ b/src/Sarif.Viewer.VisualStudio/CommonDialogFactory.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sarif.Viewer
+{
+    public class CommonDialogFactory : ICommonDialogFactory
+    {
+        public IOpenFileDialog CreateOpenFileDialog()
+        {
+            return new RealOpenFileDialog();
+        }
+    }
+}
--- a/src/Sarif.Viewer.VisualStudio/RealKeyboard.cs
+++ b/src/Sarif.Viewer.VisualStudio/RealKeyboard.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Windows;
+using System.Windows.Input;
+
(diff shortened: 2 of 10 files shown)


src/Sarif.Viewer.VisualStudio/CommonDialogFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sarif.Viewer
+{
+    public class CommonDialogFactory : ICommonDialogFactory
+    {
+        public IOpenFileDialog CreateOpenFileDialog()
+        {
+            return new RealOpenFileDialog();
```
lgolding: Note on naming: I call the dialog wrapper class `RealOpenFileDialog`, rather than just `OpenFileDialog`, because there's a class `OpenFileDialog` in the `Microsoft.Win32` namespace. OTOH, I call the factory class `CommonDialogFactory`, rather than `RealCommonDialogFactory`, because there's no existing class for the name to conflict with.

src/Sarif.Viewer.VisualStudio/RealKeyboard.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Windows;
+using System.Windows.Input;
+
+namespace Microsoft.Sarif.Viewer
+{
+    internal class RealKeyboard : IKeyboard
```
lgolding: I call this class `RealKeyboard` rather than just `Keyboard` because there's already a `Keyboard` class in the `System.Windows.Input` namespace, and having another class with the same name might cause confusion. OTOH, there is no existing "`IKeyboard`" interface, so that name is available for use without confusion.

PR #674, opened 2017-11-10
Delete jpeg 6b src

michaelcfanning:
Delete checked in jpeg6b source content.

@lgolding @jeff-bacon @vinaykapadia


lgolding:
LGTM

PR #676, opened 2017-11-13
Refactor: Inject delegate to prompt for rebaselined path

lgolding:
We use a "no-mocks" approach to unit testing the `CodeAnalysisResultsManager` class. Rather than injecting mocks for the `Keyboard` and `OpenFileDialog` objects, we inject a delegate for the method that prompts the user for the rebaselined path.

This greatly simplifies the test code (removing 36 lines), and also eliminates six classes from the test project. There are two down sides:

1. The method `CodeAnalysisResultManager.PromptForResolvedPath` is no longer unit testable (because, in unit tests, that entire method is replaced with the test delegate).

IMO, this is an acceptable tradeoff, because that method is simple enough to verify by inspection, but I'll want to hear your thoughts.

OTOH, notice that in this same PR I have made a couple of small changes to `PromptForResolvedPath`. The resulting code is simpler (which is important if we want to believe we can verify it by inspection), but without the unit tests we can't automatically verify the change.

2. This design "re-couples" the `CodeAnalysisResultsManager` class to the `System.Windows.Input.Keyboard` and `Microsoft.Win32.OpenFileDialog` classes (note that the `using` directives for those two namespaces have re-appeared).

I really like the idea of simplifying the test code and getting rid of six classes, but let me know what you think.


vinaykapadia:
I'm always happier with more streamlined code. Having a smaller method not be unit testable while making the rest of the class easier to test sounds like a good tradeoff to me.

jeff-bacon:
The previous design seemed a little overly complex just for the sake of unit testing.  This achieves the same thing with much less.  I like it.

lgolding:
@vinaykapadia That's a nice, concise way of expressing the trade-off. Thanks.

@jeff-bacon You're right. I rarely reach for the "no-mocks" tool in my toolkit. I'll try to use it more.

@michaelcfanning @rtaket I mergeed this.

PR #680, opened 2017-11-15
Fix #679: Region tooltip does not display region values

### LINKED ISSUE #679: Region tooltip does not display region values
Repro steps:

1. Save the attached file `NoRegionToolTipValues.sarif.txt`, removing the `.txt` extension so that the extension of the saved file is `.sarif`.

[NoRegionToolTipValues.sarif.txt](https://github.com/Microsoft/sarif-sdk/files/1472018/NoRegionToolTipValues.sarif.txt)

2. Open the file in Visual Studio. The Error List Window will show a single issue with Code `JS3085`.

3. Click on the issue. The SARIF Explorer window will display information about the issue.

4. Hover over the region information "`(1, 30)`" in the header of the SARIF Explorer window.

Expected: The tool tip should read "Line 1 Col 30".

Actual: The tool tip reads "Line Col", as shown in this screen shot:

noregiontooltipvalues

lgolding:
There were two distinct problems:

1. As stated in the bug, the `ToolTip` was attempting to use the `DataContext` from the `DockPanel` to which it was attached. But a `ToolTip` is not a part of the visual tree of the object to which it is attached, so this doesn't work. The Stack Overflow item https://stackoverflow.com/questions/2212171/wpf-tooltip-binding shows a technique for working around this.

2. Even with the `ToolTips`'s data context correctly set, the data was still not displayed, because the properties to which the UI was bound were not public.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
+++ b/src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
@@ -3,10 +3,6 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-             xmlns:local="clr-namespace:Microsoft.Sarif.Viewer.Views"
-             xmlns:viewer="clr-namespace:Microsoft.Sarif.Viewer"
-             xmlns:shell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
-             xmlns:models="clr-namespace:Microsoft.Sarif.Viewer.Models"
              xmlns:viewModels="clr-namespace:Microsoft.Sarif.Viewer.ViewModels"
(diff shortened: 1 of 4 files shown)


src/Sarif.Viewer.VisualStudio/Views/CodeLocations.xaml
```diff
@@ -77,19 +73,22 @@
...
-                                    <Run Text="{Binding Region.StartLine, Mode=OneWay}" />
-                                </TextBlock>
-                                <TextBlock Visibility="{Binding Region.StartColumn, Converter={StaticResource int32ToVisibilityConverter}}">
-                                    <Run Text=" Col" />
-                                    <Run Text="{Binding Region.StartColumn, Mode=OneWay}" />
-                                </TextBlock>
-                            </WrapPanel>
+                            <ToolTip
+                                DataContext="{Binding PlacementTarget.DataContext, RelativeSource={RelativeSource Self}}">
+                                <WrapPanel Orientation="Horizontal">
```
lgolding: The property `FrameworkElement.ToolTip` is of type `object` -- it can be anything at all. But by wrapping the `WrapPanel` in a `ToolTip` object, we can make use of the property `ToolTip.PacementTarget` -- which specifies the UI element to which the tooltip is attached -- to access the `DockPanel`'s `DataContext`.

Issue #581, opened 2017-06-08
Improve plugin handling


lgolding:
Address Michael's comments on #580.

While we're at it, fix a piece of logic that seemed right at the time, but now I think is wrong: As the code stands, if you specify `-plugin-assembly-path`, but you specify a `-tool` for one of the _built-in_ converters, it won't find the built-in converter. It will look in the plugin assembly, not find the converter, and give up.

We should implement the converter search as Chain of Responsibility: Give the plugin first crack at it, then try the built-ins.

This means that the plugin search and the built-in search should both implement an `IConverterFactory` interface, as should the current `ConverterFactory` class which chains the two of them together.

lgolding:
The "Chain of Responsibility" was implemented in PR #582. We haven't yet addressed Michael's comments on PR #580:

1. Add tests for `IOException`, `BadImageFormatException`, and `UnauthorizedAccessException` when failing to load plug-in.

2. Resolve any dependencies of a plug-in on assemblies that are _not_ in the directory of the executing binary.

PR #685, opened 2017-11-18
Partial #581: Add unit test for plugin load failures

lgolding:
When I introduced the plugin model for converters (PR #580), @michaelcfanning had a couple of suggestions, which I captured in Issue #581. This PR addresses the first of them: we add a unit test demonstrating that when the plugin assembly cannot be loaded, the `ToolFormatConverter` throws the expected exception.

Michael actually suggested that I add tests for `IOException`, `BadImageFormatException`, and `UnauthorizedAccessException`, but as you'll see, one test suffices.

To make this scenario testable, we again employ the "no-mocks" technique of injecting into the `ToolFormatConverter` a delegate that loads the plugin assembly. By default, that delegate is `Assembly.LoadFile`, but test code can supply a different delegate which (for example) throws an exception.

Also:
- Two unit tests in `ToolFormatConverterTests` were using XUnit assertions. Change them to use FluentAssertions.


src/Sarif.Converters/PluginConverterFactory.cs
```diff
@@ -32,7 +39,7 @@ public override ToolFileConverterBase CreateConverterCore(string toolFormat)
                 throw new ArgumentException(message, nameof(this.pluginAssemblyPath));
             }
 
-            Assembly pluginAssembly = Assembly.LoadFile(this.pluginAssemblyPath);
+            Assembly pluginAssembly = this.assemblyLoadFileDelegate(this.pluginAssemblyPath);
```
jeff-bacon: >this. [](start = 68, length = 5)

Should the 'this' keyword be present outside of the constructor for this variable?

src/Sarif.Converters/ToolFormatConverter.cs
```diff
@@ -100,12 +107,12 @@ public class ToolFormatConverter
...
         // factory capable of creating it.
         // This method is internal, rather than private, for test purposes.
-        internal static ConverterFactory CreateConverterFactory(string pluginAssemblyPath)
+        internal ConverterFactory CreateConverterFactory(string pluginAssemblyPath)
         {
             ConverterFactory factory = new BuiltInConverterFactory();
             if (!string.IsNullOrWhiteSpace(pluginAssemblyPath))
             {
-                factory = new PluginConverterFactory(pluginAssemblyPath)
+                factory = new PluginConverterFactory(pluginAssemblyPath, this.assemblyLoadFileDelegate)
```
jeff-bacon: >this. [](start = 73, length = 5)

Should the 'this' keyword be present outside of the constructor for this variable?

src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
```diff
@@ -144,6 +145,37 @@ public void ToolFormatConverter_FailsIfPluginAssemblyDoesNotExist()
...
+                string inputFilePath = tempDir.Write("input.txt", string.Empty);
+                string outputFilePath = tempDir.Combine("output.txt");
+
+                // Unlike all the other tests, which default-construct the converter, this test
+                // provides the converter with a delegate which simulates the failure to load the
+                // plugin assembly.
+                const string Message = "Something went dreadfully wrong.";
+                AssemblyLoadFileDelegate assemblyLoadFileDelegate = path => throw new BadImageFormatException(Message);
+
+                ToolFormatConverter converter = new ToolFormatConverter(assemblyLoadFileDelegate);
```
jeff-bacon: >ToolFormatConverter [](start = 16, length = 19)

`var` can be used here

src/Sarif.Converters/PluginConverterFactory.cs
lgolding: In brief, yes. Our convention is to use `this.` for instance fields (check with Michael to confirm). The SARIF SDK code base does not follow this convention consistently.

src/Sarif.Converters/ToolFormatConverter.cs
lgolding: See my reply above.

src/Sarif.Converters.UnitTests/ToolFormatConverterTests.cs
lgolding: Fixed.

src/Sarif.Converters/PluginConverterFactory.cs
jeff-bacon: Ah.  Seems like a good replacement for `_`.  

---
In reply to: [152131690](https://github.com/Microsoft/sarif-sdk/pull/685#discussion_r152131690) [](ancestors = 152131690)

PR #691, opened 2017-11-21
Fix #581: Resolve dependencies of plug-ins

### LINKED ISSUE #581: Improve plugin handling

lgolding:
This is the second of two changes to address suggestions made by @michealfanning on the pull request where I introduced the pluging model for converters (PR #580). The first change was PR #685.

In this change, we address a problem that occurs when you run the `Sarif.Multitool convert` command with the `--plugin-assembly-path` option, and supply a plugin which

1. is not in the same directory as the conversion tool executable, and
2. itself depends on assemblies that are present in the plugin's directory, but are _not_ present in the conversion tool directory.

In that case, when the conversion tool loaded the plugin, the CLR would attempt to resolve the plugin's dependencies, and it would fail because those dependencies are not on the assembly search path.

The article [Best Practices for Assembly Loading](https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading) explains the problem.

The problem was that we were loading the plugin assembly with `Assembly.LoadFile`. As a result, the assembly was not loaded into any "load context", and its dependencies were not automatically found.

The fix is to load the plugin assembly with `Assembly.LoadFrom`. This causes the plugin to be loaded into the "load from context", and causes the plugin's dependencies to be automatically loaded from the same context.

### DIFF
--- a/src/Sarif.Converters/PluginConverterFactory.cs
+++ b/src/Sarif.Converters/PluginConverterFactory.cs
@@ -9,7 +9,7 @@ using System.Reflection;
 
 namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
-    public delegate Assembly AssemblyLoadFileDelegate(string path);
+    public delegate Assembly AssemblyLoadFromDelegate(string path);
 
     // Factory class for creating a converter from a specified plug-in assembly.
     internal class PluginConverterFactory : ConverterFactory
@@ -17,13 +17,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         // This field is internal, rather than private, for test purposes.
         internal readonly string pluginAssemblyPath;
 
-        private readonly AssemblyLoadFileDelegate assemblyLoadFileDelegate;
+        private readonly AssemblyLoadFromDelegate assemblyLoadFromDelegate;
 
         internal PluginConverterFactory(
             string pluginAssemblyPath,
-            AssemblyLoadFileDelegate assemblyLoadFileDelegate = null)
+            AssemblyLoadFromDelegate assemblyLoadFromDelegate = null)
         {
-            this.assemblyLoadFileDelegate = assemblyLoadFileDelegate ?? Assembly.LoadFile;
+            this.assemblyLoadFromDelegate = assemblyLoadFromDelegate ?? Assembly.LoadFrom;
             this.pluginAssemblyPath = pluginAssemblyPath;
         }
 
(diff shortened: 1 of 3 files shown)


src/Sarif.Converters/PluginConverterFactory.cs
```diff
@@ -9,21 +9,21 @@
...
-        private readonly AssemblyLoadFileDelegate assemblyLoadFileDelegate;
+        private readonly AssemblyLoadFromDelegate assemblyLoadFromDelegate;
 
         internal PluginConverterFactory(
             string pluginAssemblyPath,
-            AssemblyLoadFileDelegate assemblyLoadFileDelegate = null)
+            AssemblyLoadFromDelegate assemblyLoadFromDelegate = null)
         {
-            this.assemblyLoadFileDelegate = assemblyLoadFileDelegate ?? Assembly.LoadFile;
+            this.assemblyLoadFromDelegate = assemblyLoadFromDelegate ?? Assembly.LoadFrom;
```
lgolding: The change from `Assembly.LoadFile` to `Assembly.LoadFrom` is the fix. Everything else is just variable and type renames.

michaelcfanning:
Assembly.LoadFrom is a dangerous API. It is dangerous for exactly the reason it is expedient to fix this issue: it provokes automatic assembly probing/loading from the specified directory.

All this points to a general insecurity in the SDK, I think, presuming that users can specify an arbitrary code path to load plug-ins. Ideally, plug-ins should drop into a privileged location, such as an application install point or a child directory. LoadFrom (or LoadFile with sufficient handling for unresolved assembly references) can provoke load of unexpected dependencies. Historically, this has been a serious issue for native security (preventing 'dll preloading' attacks, see below). I'm actually not sure whether a pinvoke dependency to a native file in a plug-in directory could cause this sort of compromise (it would be interesting to check). 

Not really sure how to proceed here. The more handling that's added to LoadFile, the more we approximate the insecurities of LoadFrom. I'm not super comfortable simply dropping back to LoadFrom due to implications.

https://support.microsoft.com/en-us/help/2389418/secure-loading-of-libraries-to-prevent-dll-preloading-attacks

michaelcfanning:
I've been thinking about this. Both LoadFile and LoadFrom can cause other issues, related to load context (as described in the link you've provided). This can be an issue in our scenario, because plug-ins need to reference application binaries. LoadFrom will actually load those dependencies (if they are alongside a plug-in assembly) into a different context. Type comparisons in driver code may fail as a result.

Anyway. The cleanest way to implement all this is to follow the article's suggestion. i.e., require plug-ins to be installed to a fusion probe-able location. If these plug-ins should always be available, we could simply use MEF to instantiate them. 

We can either go ahead with LoadFrom and try to address this later. Or can we suspend this discussion for now until we contrive a final plan. Thoughts?

PR #692, opened 2017-11-21
Fixes & suppressions for FxCop violations

EasyRhinoMSFT:


src/Sarif/PropertiesDictionary.cs
```diff
@@ -199,6 +200,7 @@ public void LoadFromXml(string filePath)
             }
         }
 
+        [SuppressMessage("Microsoft.Security.Xml", "CA3053:UseXmlSecureResolver")]
```
michaelcfanning: >[SuppressMessage("Microsoft.Security.Xml", "CA3053:UseXmlSecureResolver")] [](start = 8, length = 74)

this should be in a gloabl suppressions cs because it is a false positive. valid issues we are excluding are applied to the scan target.

PR #695, opened 2017-11-27
Compliance updates

EasyRhinoMSFT:
Remove 3rd-party code
Refactor reference code
Update comments


michaelcfanning:
>             // http://referencesource.microsoft.com/#System.Xml/Xml/System/Xml/XmlConvert.cs#1069

This is the bad reference.

---
Refers to: src/Sarif.Converters/SparseReader.cs:240 in d72c07e. [](commit_id = d72c07e97ab5274fc68677d5a39b2bcfb8c80451, deletion_comment = False)

Issue #712, opened 2017-12-12
Add symbols & source NuGet packages to build


EasyRhinoMSFT:

michaelcfanning:
Fixed

PR #715, opened 2017-12-13
Fix #711 #712 NuGet package generation

### LINKED ISSUE #711: Drop NuGet packages in bld\bin\nuget folder
Packages are currently dropped in bld\bin\<project>\<arch>

EasyRhinoMSFT:
Change output location to bld\bin\NuGet
Add generation of symbols & source packages


michaelcfanning:
sorry for the confusion, I made these changes and more in a separate PR. I should have indicated I was doing this. :) btw - when you pass --include-symbols on dotnet pack, it generates two packages, the shipping one and the one that contains symbols, so you don't need to pack twice.

PR #716, opened 2017-12-13
Various package generation fixes.

michaelcfanning:
@jeff-bacon @EasyRhinoMSFT @lgolding


BuildAndTest.cmd
```diff
@@ -20,7 +20,8 @@ echo Unrecognized option "%1" && goto :ExitFailed
 
 @REM Remove existing build data
 rd /s /q bld
-md bld\bin\nuget
+set NUGET_OUTPUT_DIR=..\..\bld\bin\nuget
+md %NUGET_OUTPUT_DIR%
```
EasyRhinoMSFT: Doesn't this result in the folders being created two levels up? The current dir is the repo. This md isn't needed actually, because the output folder will be created by the pack operation.

michaelcfanning: Fixed, thanks.

PR #717, opened 2017-12-13
Fix #708 PREfast converter NullRefException

### LINKED ISSUE #708: PREfast Conversion Failing When 'PATH' Element Missing
If the PATH element is missing, the converter will fail with a NullReferenceException.

EasyRhinoMSFT:


src/Sarif.Converters/PREFastConverter.cs
```diff
@@ -117,8 +117,11 @@ private void SetProbability(Defect defect, Result result)
 
         private void GenerateCodeFlows(Defect defect, Result result)
         {
-            if (defect.Path?.SFAs.Any() == false)
+            var sfas = defect?.Path?.SFAs;
+            if (sfas == null || sfas.Any() == false)
+            {
                 return;
+            }
```
michaelcfanning: can this be simplified to this?

if (!sfas?.Any())
{
    return;
}

EasyRhinoMSFT: No, because the type is bool?, which can't be used in a boolean expression. You actually get a syntax error that mentions an explicit cast. If you do that, you get a nullref. If you do "sfas?Any() == false", the safe nav check fails you out of the if and you don't return out. So the explicit null check is necessary.

PR #729, opened 2017-12-20
AppInsights telemetry

EasyRhinoMSFT:


src/Sarif.Viewer.VisualStudio/App.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" />
+  </startup>
+  <appSettings>
+    <add key="TelemetryInstrumentationKey_Debug" value="80a368bb-f285-46eb-bb20-3e424593ec56"/>
```
michaelcfanning: >80a368bb-f285-46eb-bb20-3e424593ec56 [](start = 56, length = 36)

Do we want the same GUID for debug vs. release? The nice thing about two distinct guids is that we can debug/test against one and not compromise release telemetry.

EasyRhinoMSFT: Yes that is the one outstanding to-do. I can't create a new instance so I'll need you to do that for me.

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.ApplicationInsights;
```
michaelcfanning: >Microsoft.ApplicationInsights; [](start = 6, length = 30)

using System directives should come first. you might need to change your VS settings.

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Linq;
+
+namespace Microsoft.Sarif.Viewer
+{
+    /// <summary>
+    /// Wrapper class for Application Insights client.
+    /// </summary>
+    internal static class TelemetryProvider
+    {
+        private static TelemetryClient appInsightsClient;
```
michaelcfanning: >appInsightsClient [](start = 39, length = 17)

please use s_ as a prefix for all private static members.

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        /// <param name="toolFormat">The tool format of the menu item.</param>
+        public static void WriteMenuCommandEvent(string toolFormat)
+        {
+            if (string.IsNullOrWhiteSpace(toolFormat))
+            {
+                throw new ArgumentNullException(nameof(toolFormat));
+            }
+
+            WriteEvent(TelemetryEvent.LogFileOpenedByMenuCommand, "Format".KeyWithValue(toolFormat == ToolFormat.None ? "SARIF" : toolFormat));
```
michaelcfanning: >ToolFormat.None [](start = 102, length = 15)

Is ToolFormat.None really a valid value to apss here? Or are we masking some other issue if this code path is exercised?

michaelcfanning:
@komal436 @jeff-bacon

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
michaelcfanning: >"Format" [](start = 66, length = 8)

what's with this pattern? is there an extension method somewhere that adds KeyWithValue to the string type? 

I'm not sure I like this., the issue is that we're adding a bolt-on method to a primitive type that's used pervasively in scenarios where KeyWithValue doesn't make sense to apply. 

I'll keep my eyes open for the extension method. if it's defined within a narrow namespace, could be ok.

EasyRhinoMSFT: Yes it's in the new file String.extensions.cs

EasyRhinoMSFT: Yeah if you look at OpenLogFileCommands.cs line 168, you see that the tool format is left at None for sarif files.

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
EasyRhinoMSFT: Cool, didn't know about that option.

src/Sarif.Viewer.VisualStudio/String.extensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Sarif.Viewer
+{
+    static class StringExtensions
```
michaelcfanning: >static [](start = 4, length = 6)

please provide accessibility modifier (internal) for this class

michaelcfanning:
> // Copyright (c) Microsoft. All rights reserved.

Can you please name this file StringExtensions.cs, which matches the type name?

---
Refers to: src/Sarif.Viewer.VisualStudio/String.extensions.cs:1 in e467a26. [](commit_id = e467a2658377cb292a9da5772af822a4893192f6, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/String.extensions.cs
michaelcfanning: >StringExtensions [](start = 17, length = 16)

Do me a favor, zap this class, actually, and go add a static helper to TelemetryEvent or TelemetryProvider that does this construction.

problems here (all small)
1) new type with a single method
2) extension method that applies a method to a commonly used type where the method is only applicable in a small number of scenarios where that type is used

alternately, we could have put this extension method in a sub-namespace that would restrict use in contexts where we opt into the telemetry. that would entail putting the telemetry types in their own namespace.

src/Sarif.Viewer.VisualStudio/String.extensions.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public static KeyValuePair<string, string> KeyWithValue(this string key, string value)
```
michaelcfanning: >KeyValuePair [](start = 22, length = 12)

where are the unit tests for this admittedly simple method?

pretty easy to test. :)

src/Sarif.Viewer.VisualStudio/TelemetryProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Sarif.Viewer
+{
+    /// <summary>
+    /// Wrapper class for Application Insights client.
+    /// </summary>
+    internal static class TelemetryProvider
```
michaelcfanning: >TelemetryProvider [](start = 26, length = 17)

tests for this class and other?

you can defer this work. suggest opening issues for it. we do want to get the telemetry into the viewer release if possible

src/Sarif.Viewer.VisualStudio/String.extensions.cs
EasyRhinoMSFT: I'd rather just yank it and do KVP construction directly.

michaelcfanning:
Chris, have you verified the telemetry is functioning? 

---
In reply to: [353115509](https://github.com/Microsoft/sarif-sdk/pull/729#issuecomment-353115509) [](ancestors = 353115509)

src/Sarif.Viewer.VisualStudio/App.config
michaelcfanning: ok, open an issue on it for chris faucon

---
In reply to: [158073364](https://github.com/Microsoft/sarif-sdk/pull/729#discussion_r158073364) [](ancestors = 158073364)

PR #734, opened 2018-01-03
Fix multitool package

EasyRhinoMSFT:
Build the Sarif.Multitool standalone NuGet package using a nuspec manifest so we can include all of the dependent reference assemblies.

### DIFF
--- a/BuildPackages.cmd
+++ b/BuildPackages.cmd
@@ -14,12 +14,11 @@ call :BuildNuGetPackage Sarif.Converters %Version% || goto :ExitFailed
 ::Build pre-release packages
 call :BuildNuGetPackage Sarif.Driver    %Version%-beta || goto :ExitFailed
 
-::Build Multitool with dependencies included
-::call :BuildNuGetPackage Sarif.Multitool %Version%-beta || goto :ExitFailed
+::Build Multitool package from nuspec
 echo .
 echo Building Sarif.Multitool package...
-echo .nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties id=Sarif.Multitool;configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\bld\bin -OutputDirectory .\bld\bin\Nuget
-.nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties id=Sarif.Multitool;configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\bld\bin -OutputDirectory .\bld\bin\Nuget
+echo .nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\ -OutputDirectory .\bld\bin\Nuget
+.nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\ -OutputDirectory .\bld\bin\Nuget
 if "%ERRORLEVEL%" NEQ "0" (echo Sarif.Multitool NuGet package creation FAILED.)
 Exit /B %ERRORLEVEL%
 
--- a/BuildPackages.cmd
+++ b/BuildPackages.cmd
@@ -18,8 +18,8 @@ call :BuildNuGetPackage Sarif.Driver    %Version%-beta || goto :ExitFailed
 ::call :BuildNuGetPackage Sarif.Multitool %Version%-beta || goto :ExitFailed
 echo .
 echo Building Sarif.Multitool package...
-echo .nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties id=Sarif.Multitool;configuration=%Configuration%;version=%Version%-beta -Verbosity Quiet -BasePath .\bld\bin -OutputDirectory .\bld\bin\Nuget
-.nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties id=Sarif.Multitool;configuration=%Configuration%;version=%Version%-beta -Verbosity Quiet -BasePath .\bld\bin -OutputDirectory .\bld\bin\Nuget
+echo .nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties id=Sarif.Multitool;configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\bld\bin -OutputDirectory .\bld\bin\Nuget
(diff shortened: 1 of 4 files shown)


src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -14,20 +14,7 @@
...
-    <Authors>Microsoft Corporation</Authors>
-    <Description>SARIF-based classes and utilities for building static analysis drivers. Includes the SARIF SDK.</Description>
-    <PackageReleaseNotes>Version $(Version) of the .NET SARIF SDK (for SARIF v1.0.0)</PackageReleaseNotes>
-    <Copyright>Copyright Microsoft 2015</Copyright>
-    <PackageLicenseUrl>https://github.com/Microsoft/sarif-sdk/blob/master/LICENSE</PackageLicenseUrl>
-    <PackageProjectUrl>https://github.com/microsoft/sarif-sdk</PackageProjectUrl>
-    <PackageTags>command line driver utilities static analysis sarif framework utils driver.utilities</PackageTags>
-    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-  </PropertyGroup>
+  <!--NuGet metadata located in src\Nuget\Sarif.Multitool.nuspec-->
```
lgolding: I don't think this comment is necessary. I don't think anybody reading the project file will wonder "Where did the NuGet metadata go?" because they won't know it was here in the first place. And they won't be surprised to see a .nuspec file on the file system, and they'll know to look there for the metadata.

src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0"?>
+<package>
+  <metadata>
+    <id>Sarif.Multitool</id>
+    <version>$version$</version>
+    <title>Microsoft SARIF Multitool (includes SARIF SDK)</title>
+    <authors>Microsoft Corporation</authors>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>SARIF-based classes and utilities for building static analysis drivers. Includes the SARIF SDK.</description>
```
lgolding: This comment is wrong (copy/pasted from the SDK).

src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -0,0 +1,33 @@
...
+    <releaseNotes>Version $version$ of the .NET SARIF Multitool (for SARIF v1.0.0)</releaseNotes>
+    <copyright>Copyright Microsoft 2015</copyright>
+    <licenseUrl>https://github.com/Microsoft/sarif-sdk/blob/master/LICENSE</licenseUrl>
+    <projectUrl>https://github.com/microsoft/sarif-sdk</projectUrl>
+    <tags>command line driver utilities static analysis sarif framework utils driver.utilities</tags>
```
lgolding: These tags are also copy/pasted from the SDK. Not all of them are appropriate.

BuildPackages.cmd
```diff
@@ -13,7 +13,14 @@ call :BuildNuGetPackage Sarif.Converters %Version% || goto :ExitFailed
 
 ::Build pre-release packages
 call :BuildNuGetPackage Sarif.Driver    %Version%-beta || goto :ExitFailed
-call :BuildNuGetPackage Sarif.Multitool %Version%-beta || goto :ExitFailed
+
+::Build Multitool package from nuspec
+echo .
+echo Building Sarif.Multitool package...
+echo .nuget\NuGet.exe pack .\src\Nuget\Sarif.Multitool.nuspec -Symbols -Properties configuration=%Configuration%;version=%Version% -Suffix beta -Verbosity Quiet -BasePath .\ -OutputDirectory .\bld\bin\Nuget
```
lgolding: I'm always uncomfortable with a hard-coded command line echo. It's so easy to get out of sync. Do we need it?

BuildPackages.cmd
```diff
@@ -13,7 +13,14 @@ call :BuildNuGetPackage Sarif.Converters %Version% || goto :ExitFailed
 
 ::Build pre-release packages
 call :BuildNuGetPackage Sarif.Driver    %Version%-beta || goto :ExitFailed
-call :BuildNuGetPackage Sarif.Multitool %Version%-beta || goto :ExitFailed
+
+::Build Multitool package from nuspec
+echo .
+echo Building Sarif.Multitool package...
```
lgolding: Can `:BuildNuGetPackage` not do this? Why do we need an explicit command here?

BuildPackages.cmd
EasyRhinoMSFT: I'm just following the existing practice. It's kind of nice to see the command line that was executed.

src/Sarif.Multitool/Sarif.Multitool.csproj
EasyRhinoMSFT: Removed

src/Nuget/Sarif.Multitool.nuspec
EasyRhinoMSFT: These were moved from the csproj. Do you have better values I can use?

BuildPackages.cmd
EasyRhinoMSFT: Are you saying we should change the package build approach over to nuspecs for all of the projects?

lgolding: Sorry, I didn't understand your change until I looked more carefully at the original file. I now understand that whereas before, we built all four packages by calling the `:BuildNuGetPackage` routine (which invoked `dotnet pack`), now you have modified one of them to use `NuGet.exe` instead, so that you can specify a .nuspec file.

I'm not wild about the resulting structure of the script. We build three packages by calling a subroutine, and we build the fourth with "inline" code. Perhaps instead you could define a second subroutine `:BuildNuGetPackageFromNuspec`, and call that one to build the Multitool package?

I wasn't suggesting that you use .nuspecs for all the packages -- although now that you mention it, it _would_ be nice to build all the packages in the same way.

BuildPackages.cmd
lgolding: I agree it's nice to see the command line. Another way to to that, though, would be to turn `echo` on just before you issue the command line, and turn it off afterwards. That would guarantee that you saw exactly what was issued, and reduce the amount of duplicate code in the script.

src/Nuget/Sarif.Multitool.nuspec
lgolding: How about: `Multi-purpose tool for analyzing and manipulating SARIF files`

src/Nuget/Sarif.Multitool.nuspec
lgolding: How about `SARIF command line static analysis`

BuildPackages.cmd
EasyRhinoMSFT: Sounds good. I'll also file an issue to move to nuspecs (the current best practice approach).

BuildPackages.cmd
EasyRhinoMSFT: OK will remove it

src/Nuget/Sarif.Multitool.nuspec
EasyRhinoMSFT: Done

src/Nuget/Sarif.Multitool.nuspec
EasyRhinoMSFT: Done

PR #737, opened 2018-01-04
Remove two obsolete attributes from AnnotatedCodeLocation.

lgolding:
The `Id` and `Essential` properties on `AnnotatedCodeLocation` were previously marked obsolete, and are now removed. They were replaced (even in SARIF 1.0) with `Step` and `Importance`.

I think you'll find it easiest to review this PR in this order:

1. See the two properties removed from Sarif.schema.json.
2. Then, in CodeGenHints.json, note the removal of the "code generation hints" for those two properties.
3. Then note that one of those hints decorated the `id` property with an attribute that specified a JSON converter. That converter and its unit tests are no longer necessary.
4. In AnnotatedCodeLocation.extensions.cs (part of the SARIF Viewer VSIX), in the method which constructed the view model for AnnotatedCodeLocation, it is no longer necessary to handle the case of files that specified `Essential` rather than `Importance`.
5. Finally, you can glance at the generated object model files AnnotatedCodeLocation.cs and AnnotatedCodeLocationEqualityComparer.cs.
  
 Also:
- Enable AppVeyor builds on the sarif-v2 branch (see appveyor.yml).


lgolding:
Thanks all.

PR #740, opened 2018-01-05
Add TC-approved rule.help to the schema.

lgolding:
The SARIF TC approved the addition of a `help` property to the `rule` object. This property contains help text for the rule, and is useful when there is no online documentation for the rule (for example, when an analysis tool allows developers to author their own rules, for which there is no online documentation).

The primary change is in Sarif.schema.json. That leads to changes in three of the SARIF OM files that the build process auto-generates from the schema. Finally, a few source files need to be updated to take account of the new property.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
+++ b/src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
@@ -74,9 +74,13 @@ namespace Microsoft.Sarif.Viewer.Sarif
                         helpUri = new Uri(RuleMetadata[ruleId]["url"].Value<string>());
                     }
 
+                    // PREfast rules don't have or need a "help" property; they all have online documentation.
+                    // The rule.help property is only useful in scenarios where there's no online documentation.
+                    string help = null;
+
                     if (ruleName != null || helpUri != null)
                     {
-                        rule = new Rule(ruleId, ruleName, null, null, null, RuleConfiguration.Unknown, ResultLevel.Warning, helpUri, null);
+                        rule = new Rule(ruleId, ruleName, null, null, null, RuleConfiguration.Unknown, ResultLevel.Warning, helpUri, help, null);
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


src/Sarif/Autogenerated/IRule.cs
```diff
@@ -53,5 +53,10 @@ public partial interface IRule
         /// A URI where the primary documentation for the rule can be found.
         /// </summary>
         Uri HelpUri { get; }
+
+        /// <summary>
+        /// Provides the primary documentation for the rule, useful when there is no on-line documentation.
```
michaelcfanning: >on- [](start = 85, length = 3)

'online' no hyphen necessary

src/Sarif/Autogenerated/Rule.cs
```diff
@@ -80,6 +80,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "helpUri", IsRequired = false, EmitDefaultValue = false)]
         public Uri HelpUri { get; set; }
 
+        /// <summary>
+        /// Provides the primary documentation for the rule, useful when there is no on-line documentation.
```
michaelcfanning: >on-line [](start = 85, length = 7)

ditto

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -74,9 +74,13 @@ public static bool TryGetRule(this Run run, string ruleId, string ruleKey, out I
                         helpUri = new Uri(RuleMetadata[ruleId]["url"].Value<string>());
                     }
 
+                    // PREfast rules don't have or need a "help" property; they all have online documentation.
+                    // The rule.help property is only useful in scenarios where there's no online documentation.
+                    string help = null;
+
                     if (ruleName != null || helpUri != null)
                     {
-                        rule = new Rule(ruleId, ruleName, null, null, null, RuleConfiguration.Unknown, ResultLevel.Warning, helpUri, null);
```
michaelcfanning: >null [](start = 58, length = 4)

do me a favor, at the first null literal, use a named arg instead (which will require you to name all remaining args).

make this a multiline statement. After these two things, you won't need the meaningless local initialized to null in order to make this code readable.

lgolding:
Thanks. I'll make all the changes you suggested.

src/Sarif/Autogenerated/Rule.cs
lgolding: (Both of those comments come from the schema, so only one change is needed.)

PR #741, opened 2018-01-06
Add v1.0.0 of the schema to the sarif-v2 branch.

lgolding:
This will be used by the upgrade tool. The tool will first check if the log file is a valid v1.0.0 log file before attempting to upgrade it to v2.0.0.


src/Sarif/Schemata/Sarif-1.0.0.schema.json
```diff
@@ -0,0 +1,1250 @@
+{
```
lgolding: This file is copied without change from Sarif.schema.json in the master branch to Sarif-1.0.0.schema.json in the sarif-v2 branch.

PR #745, opened 2018-01-15
Implement schema changes to support rich messages

lgolding:
This PR implements the schema changes that address oasis-tcs/sarif-spec#33, as described in the document [Schema changes between SARIF v1 and v2](https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/Schema%20changes%20between%20SARIF%20v1%20and%20v2.md). Several properties are added, and others are renamed for clarity.

This commit does _not_ include logic to populate the `rule.richMessageTemplates` property. The `SkimmerBase` class declares that property, and initializes it to an empty dictionary, but it does not yet populate the dictionary from the application resources. I will do that in a subsequent PR. The reason I'm not doing it here is that we first need to decide on a naming convention for the rich template resource strings.


src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -18,37 +18,57 @@ public SkimmerBase()
...
+        }
+
+        private Dictionary<string, string> InitializeMessageTemplates()
+        {
+            return RuleUtilities.BuildDictionary(ResourceManager, TemplateIds, Id);
+        }
+
+        private Dictionary<string, string> InitializeRichMessageTemplates()
+        {
+            return new Dictionary<string, string>();
```
lgolding: This is the part that's not implemented. See the difference between this method and `InitializeMessageTemplates`. `RuleUtilities.BuildDictionary` relies on a naming convention for resource ids to to populate the `MessageTemplates` dictionary. We need to decide on a similar resource convention for populating the `RichMessageTemplates` dictionary. It might be as simple as `<rule id>_rich_<template id>`. For example, BinSkim has a resource id `BA2005_Error_Bad_Module`. Perhaps the rich text equivalent can be simply `BA2005_rich_Error_Bad_Module`.

michaelcfanning:
>           "description": "A string that describes the condition that was encountered.",

we should not this is a plain text message here

---
Refers to: src/Sarif/Schemata/Sarif.schema.json:602 in 89d37f3. [](commit_id = 89d37f3403b1ed9c9b504f6fecec8ea60beea59f, deletion_comment = False)

src/Sarif.Driver/Sdk/SkimmerBase.cs
michaelcfanning: ok. can we get an issue opened on this?

---
In reply to: [161602933](https://github.com/Microsoft/sarif-sdk/pull/745#discussion_r161602933) [](ancestors = 161602933)

lgolding: @michaelcfanning I opened #746:

> In `SkimmerBase`, the `MessageTemplates` property is lazily initialized by a call to `InitializeMessageTemplates`, which calls `RuleUtilities.BuildDictionary` to populate the property by reading resource strings whose names follow a certain convention.

> At present, there is no similar mechanism for populating `SkimmerBase.RichMessageTemplates`; that property is always an empty dictionary. We need to define a convention for the resource strings which define rich message templates, then implement the logic (filling in the body of `SkimmerBase.InitializeRichMessageTemplates`, which currently does nothing), and write unit tests for it.

PR #752, opened 2018-01-17
Fix #746: Populate RichMessageTemplates in SkimmerBase

### LINKED ISSUE #746: Populate RichMessageTemplates in SkimmerBase
In `SkimmerBase`, the `MessageTemplates` property is lazily initialized by a call to `InitializeMessageTemplates`, which calls `RuleUtilities.BuildDictionary` to populate the property by reading resource strings whose names follow a certain convention.

At present, there is no similar mechanism for populating `SkimmerBase.RichMessageTemplates`; that property is always an empty dictionary. We need to define a convention for the resource strings which define rich message templates, then implement the logic (filling in the body of `SkimmerBase.InitializeRichMessageTemplates`, which currently does nothing), and write unit tests for it.

lgolding:
In `SkimmerBase`, the `MessageTemplates` property is lazily initialized by a call to `InitializeMessageTemplates`, which calls `RuleUtilities.BuildDictionary` to populate the property by reading resource strings whose names follow a certain convention.

At present, there is no similar mechanism for populating `SkimmerBase.RichMessageTemplates`; that property is always an empty dictionary. We need to define a convention for the resource strings which define rich message templates, then implement the logic (filling in the body of `SkimmerBase.InitializeRichMessageTemplates`, which currently does nothing), and write unit tests for it.

That's what this PR does. We introduce the convention that a skimmer should initialize its `RichTemplateResourceIds` property with a list of resource string names of the form `<ruleId>_Rich_<templateId>`. The code in `SkimmerBase.RichMessageTemplates.get` will strip off the prefix and use `<templateId>` as key into the dictionary (with the resource string as the value, just as for `MessageTemplates`).


src/Sarif.Driver.UnitTests/TestAnalysisContext.cs
```diff
@@ -1,32 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
-    public class TestAnalysisContext : IAnalysisContext
+    public class TestAnalysisContext : AnalysisContext
```
lgolding: There's no need to derive directly from `IAnalysisContext` and reimplement all the members. `AnalysisContext` already does that.

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -55,12 +57,12 @@ public SkimmerBase()
 
         private Dictionary<string, string> InitializeMessageTemplates()
         {
-            return RuleUtilities.BuildDictionary(ResourceManager, TemplateIds, Id);
+            return RuleUtilities.BuildDictionary(ResourceManager, TemplateResourceIds, ruleId: Id);
```
lgolding: Add the parameter name for clarity, since on first reading its not obvious what this is the "id" of.

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -23,7 +23,9 @@ public SkimmerBase()
 
         abstract protected ResourceManager ResourceManager { get; }
 
-        abstract protected IEnumerable<string> TemplateIds { get; }
+        abstract protected IEnumerable<string> TemplateResourceIds { get; }
```
lgolding: IMO this is a clearer name. These are not in fact the SARIF template ids; they are resource ids of the form `<ruleId>_<templateId>`, from which the template ids are extracted.

michaelcfanning:
Suggest using 'Markdown' as your marker term in resource strings rather than 'Rich' due to the non-obvious meaning of 'rich' in this context. 'Markdown' will be perfectly clear, we expect to populate it with markdown.

src/Sarif.Driver/Sdk/SkimmerBase.cs
michaelcfanning: >TemplateResourceIds [](start = 47, length = 19)

Maybe 'TemplateResourceNames' is even better?

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -23,7 +23,9 @@ public SkimmerBase()
...
+
+        virtual protected IEnumerable<string> RichTemplateResourceIds => new List<string>();
```
michaelcfanning: >RichTemplateResourceIds [](start = 46, length = 23)

See other comment. I like Markdown better. So MarkdownTemplateResourceNames. 

What do you think?

src/Sarif/RuleUtilities.cs
```diff
@@ -105,6 +106,12 @@ public static string NormalizeTemplateId(string ruleId, string templateId)
             {
                 templateId = templateId.Substring(ruleId.Length + 1);
             }
+
+            if (!string.IsNullOrEmpty(prefix) && templateId.StartsWith(prefix + "_", StringComparison.Ordinal))
+            {
+                templateId = templateId.Substring(prefix.Length + 1);
```
michaelcfanning: >templateId.Substring(prefix.Length + 1 [](start = 29, length = 38)

you said in the description that the prefix would be inserted after the rule id, but here, you permit the string to be prefixed with the prefix. 

need to be consistent. if we follow your change description, we don't need this new conditional, which is nice

src/Sarif.Driver.UnitTests/TestAnalysisContext.cs
michaelcfanning: nice

---
In reply to: [161965854](https://github.com/Microsoft/sarif-sdk/pull/752#discussion_r161965854) [](ancestors = 161965854)

src/Sarif/RuleUtilities.cs
```diff
@@ -105,6 +106,12 @@ public static string NormalizeTemplateId(string ruleId, string templateId)
             {
                 templateId = templateId.Substring(ruleId.Length + 1);
             }
+
+            if (!string.IsNullOrEmpty(prefix) && templateId.StartsWith(prefix + "_", StringComparison.Ordinal))
```
michaelcfanning: >string [](start = 17, length = 6)

looking at your tests, i don't believe this conditional is ever hit. prove me wrong in code coverage? or delete this conditional if i am proven right.

src/Sarif.Driver/Sdk/SkimmerBase.cs
lgolding: The spec uses the term “rich” everywhere to describe messages with formatting information. It starts right in the terminology section, where we define

**rich text message**
    message which contains formatting information such as Markdown formatting characters

… and continues throughout the spec, in the spec language and in all property names such as `result.richMessage`. I think that it’s important for the SDK to use the same language as the spec. Thoughts? (Warning: I feel strongly about this, so I’ll need to be convinced to change it 😊)

src/Sarif/RuleUtilities.cs
lgolding: The method `NormalizeTemplateId` first strips the `<ruleId>` (if present), and then strips the `<prefix>` (if present). So any of these will work (in the sense that when `NormalizeTemplateId` is finished), only `<templateId>` will remain:

- `<ruleId>_<prefix>_<templateId>`
- `<ruleId>_<templateId>`
- `<prefix>_<templateId>`
- `<templateId>`

Those last two would be useful when the same message is shared by multiple rules (skimmers).

src/Sarif/RuleUtilities.cs
lgolding: @michaelcfanning: The conditional is indeed hit, and both the true and false cases are exercised, as I just demonstrated to myself by

1. commenting it out and watching my unit test fail on this assertion with "The given key was not present in the dictionary"
    ```
     skimmer.RichMessageTemplates["Pass"]
        .Should().Be(SkimmerBaseTestResources.TST0001_Rich_Pass);
    ```
2. setting a breakpoint and watching it hit both the `true` and `false` branches when expected.

src/Sarif.Driver/Sdk/SkimmerBase.cs
lgolding: @michaelcfanning  You are right. In the API[ `ResourceManager.GetString(string)`](https://msdn.microsoft.com/en-us/library/d17ax2xk(v=vs.110).aspx), the name of the parameter is indeed `name` instead of `id`. Fixed.

PR #753, opened 2018-01-18
Update version 1.7.2, new CommandLineParser, remove beta flag, update Multitool packaging

EasyRhinoMSFT:


lgolding:
:+1:

Question -- are we now publishing separate NuGet packages for each target platform?

EasyRhin0:
No, each package contains bits for all three.

EasyRhinoMSFT:
Larry, I just realized why you’re asking this: you’re looking at the three calls to CreatePublishPackage, one for each framework. What “dotnet publish” does is drop everything needed for the nuget package into a folder, including all nuget package references. Later we use “dotnet pack” with a .nuspec file to create the actual .nupkg package with the contents of the publish folder.

PR #760, opened 2018-01-30
Updates for v1.7.3

EasyRhinoMSFT:


src/ReleaseHistory.md
```diff
@@ -168,4 +168,8 @@
  * Update Multitool nuget package build
  * Enable "pretty print" .sarif formatting via --pretty argument
  * Code sign 3rd party dependency assemblies (CommandLineParser, CsvHelper, Newtonsoft.Json)
- * Remove -beta flag from Driver and Multitool packages
\ No newline at end of file
+ * Remove -beta flag from Driver and Multitool packages
+ 
+ ## **v1.7.3** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/1.7.3) | [Driver](https://www.nuget.org/packages/Sarif.Driver/1.7.3) | [Converters](https://www.nuget.org/packages/Sarif.Converters/1.7.3) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/1.7.3)
+ * Make SupportedPlatform a first class concept for skimmers
```
Evmaus-MS: I'm not sure if we were calling this out as a documented feature yet--just check with @michaelcfanning

EasyRhinoMSFT: Confirmed with Michael that this is OK.

PR #778, opened 2018-02-08
Fix #773: Adjust byte offset when BOM is present

### LINKED ISSUE #773: Viewer: fix offsets are wrong when source file has BOM

EasyRhinoMSFT:


src/Sarif.Viewer.VisualStudio/Models/FixModel.cs
```diff
@@ -165,7 +161,16 @@ private bool TryFixFile(string filePath, IEnumerable<ReplacementModel> replaceme
 
                 // Delete/Insert the bytes for each replacement.
                 List<byte> bytes = File.ReadAllBytes(filePath).ToList();
+
                 int offsetDelta = 0;
+
+                // Account for the BOM if it's present
+                // Don't remove it because that may be undesirable to the user
+                if (bytes.Count > 2 && bytes[0] == 0xEF && bytes[1] == 0xBB && bytes[2] == 0xBF)
```
michaelcfanning: >bytes.Count [](start = 20, length = 11)

I'm not sure this fix s/be required. isn't it the responsibility of the analysis tool to provide a correct offset, which includes being aware of the byte order mark? Hm.

EasyRhinoMSFT: Definitely, but we're the ones who end up with the broken functionality when they don't. :) I guess the real problem is that we won't know if the BOM has been taken into account.

michaelcfanning: right. disregarding the BOM makes complete sense for text edits, otherwise, you end up with non-obvious column offsets for your first line. the question is whether it makes sense to disregard it for offsets in other contexts. imagine a checker that had a fixer to strip a BOM. if we disregard it, then you couldn't write this rule. :) just thought it'd be funny to mention that silly example.

let me double-check the test case we generated. maybe what actually happened here is some files got out of sync.

michaelcfanning: ok, this change is wrong according to the SARIF spec, which says that replacements should be computed as an offset of bytes in the file. this places the burden of accounting for the BOM on the SARIF producer. I don't know whether I think this is right or not. I'll open an issue on the SARIF repo to ask about it.

EasyRhinoMSFT: OK, then this change can be abandoned. The same change is in the offset fix, I'll remove it.

PR #788, opened 2018-02-20
Fix #786: Show error list items for config & tool notifications

### LINKED ISSUE #786: Viewer doesn't display catastrophic tool and configuration notifications.
Repro:

open the following file. we get all the informational messages but don't see the errors that indicate the PDBs were stripped. Basically, we need to process configurationNotifications and toolNotifications, as we do the results.

https://github.com/Microsoft/binskim/blob/master/src/Test.FunctionalTests.BinSkim.Driver/BaselineTestsData/Expected/Native_x86_VS2017_15.5.4_PdbStripped.dll.sarif

EasyRhinoMSFT:
Excluding tool notifications where level=="info", such as "run started" etc. If you disagree, testify!

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
@@ -122,10 +122,8 @@ namespace Microsoft.Sarif.Viewer
             Message = notification.Message;
             ShortMessage = notification.Message;
             LogFilePath = logFilePath;
+            FileName = notification.PhysicalLocation?.Uri.LocalPath ?? run.Tool.FullName;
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -114,11 +114,37 @@ public SarifErrorListItem(Run run, Result result, string logFilePath, ProjectNam
                 }
             }
         }
+        public SarifErrorListItem(Run run, Notification notification, string logFilePath, ProjectNameCache projectNameCache) : this()
```
michaelcfanning: >SarifErrorListItem [](start = 15, length = 18)

NIT: introduce line space before this ctor

PR #791, opened 2018-02-21
Provide mechanism for disabling checks

michaelcfanning:
@Evmaus-MS 

btw - I haven't reviewed this myself yet. So feel free to wait until that occurs.


src/Sarif/Warnings.cs
```diff
@@ -10,8 +10,9 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class Warnings
     {
-        public const string Wrn997InvalidTarget = "WRN997.InvalidTarget";
-        public const string Wrn998NotSupportedPlatform = "WRN998.NotSupportedPlatform";
+        public const string Wrn997InvalidTarget        = "WRN997";
+        public const string Wrn998NotSupportedPlatform = "WRN998";
+        public const string Wrn999RuleDisabled         = "WRN999";
```
Evmaus-MS: So there's a convention here and in Errors.cs that WRN998.(explanation)/ERR998.(Explanation) are used for rule-disabling errors/warnings. 

Similarly, (WRN/ERR)997 are configuration/target errors (missing pdb, invalid target, cannot write log, etc.).

I think that model is a little better than just addressing these sequentially as we assign them.

src/Sarif/RuntimeConditions.cs
```diff
@@ -49,12 +49,19 @@ public enum RuntimeConditions : uint
...
         // Non-fatal conditions
+        RuleWasExplicitlyDisabled = 0x04000000,
         RuleCannotRunOnPlatform   = 0x08000000,
         RuleNotApplicableToTarget = 0x10000000,
         TargetNotValidToAnalyze   = 0x20000000,
         OneOrMoreWarningsFired    = 0x40000000,
         OneOrMoreErrorsFired      = 0x80000000,
 
-        Nonfatal = 0xF8000000
+        // 0xFC000000
```
Evmaus-MS: We'll need to update our BinSkim documentation.

We also should consider if we just want 0xFFC or similar to be our 'non-fatal exit code' space, since when we change this it may affect teams integrating the tool.

src/Sarif.Driver/Sdk/RuleEnabledState.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
Evmaus-MS: Nit: Copyright

src/Sarif.Driver/Sdk/RuleEnabledState.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif.Driver.Sdk
```
Evmaus-MS: Nit:  Namespace should just be Driver

src/Sarif.Driver/Sdk/DefaultDriverOptions.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Driver.Sdk
```
Evmaus-MS: Nit:  Namespace should probably just be Microsoft.CodeAnalysis.Sarif.Driver

src/Sarif.Driver/Sdk/ExportConfigurationCommandBase.cs
```diff
@@ -20,17 +23,41 @@ public override int Run(ExportConfigurationOptions exportOptions)
...
+
+                    string ruleOptionsKey = rule.Id + "." + rule.Name + ".Options";
+
+                    if (!allOptions.TryGetValue(ruleOptionsKey, out objectResult))
+                    {
+                        objectResult = allOptions[ruleOptionsKey] = new PropertiesDictionary();
+                    }
+                    properties = (PropertiesDictionary)objectResult;
+
+                    foreach (IOption option in DefaultDriverOptions.Instance.GetOptions())
```
Evmaus-MS: Design:  So, this particular implementation seems to mean that consumers of the driver SDK cannot provide more default cross rule options, correct?  Do we want to enable that?  

I know the Driver+SDK+BinSkim are pretty tightly coupled, but this sort of thing will increase that coupling (esp. if down the line we want to add more "cross rule" configuration options for BinSkim's sake).

Just something to think about; I think this is fine for now, but we may want to add a comment in the DefaultDriverOptions or etc. about reducing this coupling in the future rather than adding more options there.

src/Sarif.Driver/Sdk/RuleEnabledState.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// analysis was disabled.
+        /// </summary>
+        Disabled = 0x1,
+
+        /// <summary>
+        /// User has reduced all signal from a rule to warning. Warnings
+        /// should always be persisted to reports but may not block
+        /// engineering processes.
+        /// </summary>
+        Warning = 0x2,
```
Evmaus-MS: Q:  Do Warning/Error do anything at the moment, or will the rules or logger need to all be altered to consume this setting correctly?

src/Sarif/Warnings.cs
michaelcfanning: warning WRN999 : Rule 'BA2010' was explicitly disabled by the user.
error ERR997.ExceptionLoadingPdb : BA2006 : 'MixedMode_x86_VS2013_MissingPdb.dll' was not evaluated for check 'BuildWithSecureTools'.

See output above for why I made the change. I think I should further qualify the ids as you say, then maybe we can consider changing the console output to strip the friendly sub-name (since the output text makes things clear).


---
In reply to: [169829092](https://github.com/Microsoft/sarif-sdk/pull/791#discussion_r169829092) [](ancestors = 169829092)

src/Sarif.Driver/Sdk/RuleEnabledState.cs
michaelcfanning: This file lives in the Sdk  directory, so by convention s/be in that namespace. ?

---
In reply to: [169829799](https://github.com/Microsoft/sarif-sdk/pull/791#discussion_r169829799) [](ancestors = 169829799)

src/Sarif.Driver/Sdk/RuleEnabledState.cs
michaelcfanning: That's right, future work item. 

---
In reply to: [169831826](https://github.com/Microsoft/sarif-sdk/pull/791#discussion_r169831826) [](ancestors = 169831826)

src/Sarif.Driver/Sdk/DefaultDriverOptions.cs
michaelcfanning: I see now that the code is in an inconsistent condition. I went ahead and removed all reference to the SDK namespace.

---
In reply to: [169829946](https://github.com/Microsoft/sarif-sdk/pull/791#discussion_r169829946) [](ancestors = 169829946)

src/Sarif.Driver/Sdk/ExportConfigurationCommandBase.cs
michaelcfanning: Yes, right now other consumers cant enable this sort of thing, would need to be elevated to the SDK incrementally.,

Agreed, the coupling is clear and unfortunate. It's not a great way to evolve the driver framework, lots of assumptions that are relevant to binskim only are baked in...

---
In reply to: [169830521](https://github.com/Microsoft/sarif-sdk/pull/791#discussion_r169830521) [](ancestors = 169830521)

src/Sarif/Warnings.cs
Evmaus-MS: Makes sense.  My main thinking is that we should be consistent.

Maybe WRN998.RuleDisabled:  '(x)' was explicitly disabled by the user and WRN.998.RuleDisabled: '(x)' cannot execute on this platform?

PR #793, opened 2018-02-23
Default Config File for Driver

Evmaus-MS:
Add a default behavior of pulling from a configuration file beside the tool (default.xml).  This will allow organizations to create a custom default configuration and repackage a tool without requiring all of their users to pass a --config argument.

@michaelcfanning -- This still needs unit tests, I was going to wait until your current PR is merged and then pull it into this branch so that I can use the additional work you did around unit testing the configuration routine there.


src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -277,8 +286,17 @@ protected virtual void InitializeConfiguration(TOptions options, TContext contex
...
+            {
+                return;
+            }
+
+            if (String.IsNullOrEmpty(options.ConfigurationFilePath) &&
+                    File.Exists(DefaultConfigurationPath))
+            {
+                options.ConfigurationFilePath = DefaultConfigurationPath;
+            }
+            else
```
michaelcfanning: NIT, considering reversing this conditional to that exit occurs here and then the positive code path happens as follow-on execution. i.e., 

if (!String.IsNullOrEmpty(options.blah) ||
    !File.Exists(blah)
{
  return;
}

options.ConfigurationFilePath = DefaultConfdigblah..

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -48,8 +48,7 @@ public abstract class AnalyzeOptionsBase
         [Option(
             'c',
             "config",
-            Default = "default",
-            HelpText = "Path to policy file that will be used to configure analysis. Passing value of 'default' (or omitting the argument entirely) will configure BinSkim to use built-in settings.")]
+            HelpText = "Path to policy file that will be used to configure analysis. We default to 'default.xml' beside the main tool; passing value of 'default' or removing that file will configure BinSkim to use built-in settings.")]
```
michaelcfanning: >default.xml [](start = 100, length = 11)

default.configuration.xml

PR #796, opened 2018-02-28
Fix #706 UI strings + Viewer main panel styles

### LINKED ISSUE #706: Viewer: move hard-coded UI strings to resources
To enable localization

EasyRhinoMSFT:


src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
```diff
@@ -0,0 +1,62 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:system="clr-namespace:System;assembly=mscorlib">
+
+    <!-- SARIF Viewer Control -->
+    <system:String x:Key="SarifViewerControl_DialogHeader_Separator_Text_String">|</system:String>
```
lgolding: Does this need to be localized?

src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
```diff
@@ -0,0 +1,62 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:system="clr-namespace:System;assembly=mscorlib">
+
+    <!-- SARIF Viewer Control -->
```
lgolding: What do you think of the idea of pushing the string resources for each category down into the corresponding resource file, rather then mixing them all into a single StringResources.xaml file? Like, the Stacks-related strings go into StacksStyles.xaml? Alternatively, separate them into separate files like StacksStrings.xaml, so that Stacks.xaml would include both StacksStyles and StacksStrings?

src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
EasyRhinoMSFT: I guess not, I just don't like having literals in markup.

src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
EasyRhinoMSFT: The idea is that the strings file can be sent to a localization service, and they return the same file for each language. Isolating the strings makes this process clean and simple.

lgolding: @EasyRhinoMSFT  Ok. I leave it to you to make the right trade-off here between separation of concerns, on the one hand, and ease of localization, on the other. _If_ it were just as easy for the localizers to handle a zip file containing several separate resource files as it is for them to deal with a single resource file, I'd like my proposal better. But you know the process and I don't, so do whatever it right.

src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
lgolding: @EasyRhinoMSFT In your response to my other comment, you cited ease of localization in support of your position, which is reasonable. But then here, why should we make the localizers look at this and wonder how they're supposed to translate "`|`"? :-) IMO we should just take that off their plates.

EasyRhinoMSFT: Fair enough

src/Sarif.Viewer.VisualStudio/Views/StringResources.xaml
EasyRhinoMSFT: Split the difference, created separate files for each.

src/Sarif.Viewer.VisualStudio/Views/CallTreesStringResources.xaml
```diff
@@ -0,0 +1,12 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:system="clr-namespace:System;assembly=mscorlib">
+
+    <system:String x:Key="CallTrees_ToolBarButtonExpand_Text_String">+</system:String>
```
lgolding: Oh, here are some more strings that probably shouldn't be localized. Sorry!

PR #800, opened 2018-03-07
Rebase URI/Sarif log processing work

Evmaus-MS:
This adds two things--a visitor that allows us to rebase all of the URIs in a file (making them relative to some variable stored in the Run's properties, and storing the appropriate variable in the Run's properties), and a set of generic classes that allow us to take a generic action on a set of files (in form of a class that will perform a Map/Fold on a set of files)--this is the early stages for enabling users to define custom multitool actions that are largely configuration driven.


Evmaus-MS:
Closing this for now.  More work is needed to go from relative paths back to absolute paths.

PR #803, opened 2018-03-13
Sarif file processing

Evmaus-MS:
Adds uri rebase and absolute uri tasks to the multitool, using a new pipeline model for actions on sarif files (that will eventually allow serializability of the pipeline/potentially extensibility--eventually allowing us to drive the multitool  or other transformations from data).


src/Sarif/SdkResources.resx
```diff
@@ -238,7 +238,7 @@
     <value>Rule '{0}' was disabled as it cannot run on the current platform '{1}'.  It can only run on '{2}'.</value>
   </data>
   <data name="WRN999_RuleExplicitlyDisabled" xml:space="preserve">
-    <value>Rule '{0}' was explicitly disabled by the user. As result, this too run cannot be used to for compliance or other auditing processes that require a comprehensive analysis.</value>
+    <value>Rule '{0}' was explicitly disabled by the user. As result, this tool run cannot be used to for compliance or other auditing processes that require a comprehensive analysis.</value>
```
michaelcfanning: >to for [](start = 99, length = 6)

'to for' s/be 'for'

src/Sarif/Processors/Generic/GenericFoldAction.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class GenericFoldAction<T> : IFoldAction<T> where T : new()
+    {
+        Func<T, T, T> _action;
+
+        public GenericFoldAction(Func<T, T, T> action)
+        {
+            _action = action;
+        }
+
+        public T Fold(IEnumerable<T> list, T accumulator)
```
michaelcfanning: >list [](start = 37, length = 4)

propose 'collection' as a name here, given type of parameter NIT

src/Sarif/Processors/Generic/IGenericAction.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif.Processors
+{
+    public interface IGenericAction<T>
```
michaelcfanning: >IGenericAction [](start = 21, length = 14)

Generic as a term here is a bit confusing for this generic type.  IActionWrapper. IBasicAction, IGeneralAction?

src/Sarif/Processors/Log/SarifLogActions.cs
```diff
@@ -0,0 +1,20 @@
...
+    // Mapping Actions
+    public enum SarifLogAction
+    {
+        None = 0,
+        RebaseUri,
+        MakeUrisAbsolute,
+        Merge,
+        // Future work...
+        Sort,
+        MakeDeterministic,
```
michaelcfanning: >MakeDeterministic [](start = 8, length = 17)

other actions, embed file contents, remove file contents, populate files table, provide file hashes.

note that sorting is implied by making the file deterministic, but not vice versa. Might need 'StripNondeterministicPathInformation', and then 'MakeDeterministic' == Sort | StripNondeterministicPathInformation

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -0,0 +1,124 @@
...
+        
+        /// <summary>
+        /// Two ActionTuples are equal if they take the same actions with the same parameters.
+        /// </summary>
+        public override bool Equals(object obj)
+        {
+            ActionTuple other = obj as ActionTuple;
+            if (other == null)
+            {
+                return base.Equals(obj);
```
michaelcfanning: > base.Equals(obj) [](start = 22, length = 17)

can this call ever return true?

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            if(this.Action != other.Action)
```
michaelcfanning: >if(t [](start = 12, length = 4)

NIT space issues

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                return false;
+            }
+
+            if (this.Parameters.Length != other.Parameters.Length)
```
michaelcfanning: >this.Parameters [](start = 16, length = 15)

is this.Parameters always guaranteed to be non-null for this instance and the one we're comparing to? If not we will crash

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -0,0 +1,124 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Processors
+{
+    public class ActionTuple
```
michaelcfanning: >ActionTuple [](start = 17, length = 11)

put this class in its own file?

src/Sarif/Processors/Log/SarifLogStageFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Processors
+{
+    public class SarifLogProcessorFactory
+    {
+        public static IGenericAction<SarifLog> GetActionStage(SarifLogAction action, params string[] args)
+        {
+            switch (action)
+            {
+                case SarifLogAction.None:
+                    return new GenericMappingAction<SarifLog>(a => a);
```
michaelcfanning: >return new GenericMappingAction<SarifLog>(a => a); [](start = 20, length = 50)

insert braces in these cases is my style suggestion. :)

michaelcfanning: >return new GenericMappingAction [](start = 20, length = 31)

why is this action state valid? just curious

src/Sarif/Processors/Log/SarifLogStageFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+                case SarifLogAction.RebaseUri:
+                    return new GenericMappingAction<SarifLog>(log =>
+                    {
+                        RebaseUriVisitor visitor = new RebaseUriVisitor(args[0], new Uri(args[1]));
+                        return visitor.VisitSarifLog(log);
+                    });
+                case SarifLogAction.Merge:
+                    return new GenericFoldAction<SarifLog>(mergeFunction);
+                case SarifLogAction.Sort:
+                    throw new NotImplementedException("Future work, not yet supported.");
```
michaelcfanning: >"Future work, not yet supported." [](start = 54, length = 33)

don't need a message here in my opinion, as the exception type communicates everything that the text string does

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
+﻿// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT        
+// license. See LICENSE file in the project root for full license information. 
+
+using Microsoft.CodeAnalysis.Sarif.Readers;
```
michaelcfanning: >Microsoft [](start = 6, length = 9)

NIT namespaces not ordered properly, system directive s/be first

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+            Debug.Assert(_baseUri.IsAbsoluteUri);
+        }
+
+        public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
+        {
+            PhysicalLocation newNode = base.VisitPhysicalLocation(node);
+            
+            if (string.IsNullOrEmpty(newNode.UriBaseId))
+            {
+                if (_baseUri.IsBaseOf(newNode.Uri) && newNode.Uri.IsAbsoluteUri)
```
michaelcfanning: >_baseUri.IsBaseOf(newNode.Uri) && newNode.Uri.IsAbsoluteUri [](start = 20, length = 59)

do we get slightly better perf if we switch these conditions?

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+            return newNode;
+        }
+
+        public override Run VisitRun(Run node)
+        {
+            Run newRun = base.VisitRun(node);
+
+            if (newRun.Files != null)
+            {
+                FixFiles(newRun);
```
michaelcfanning: >ixFiles(newRun); [](start = 17, length = 16)

why didn't you do this in an override of VisitFiles? Curious

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+                node.Properties = new Dictionary<string, SerializedPropertyInfo>();
+            }
+
+            // If the dictionary doesn't exist, we should add it to the properties.  If it does, we should add/update the existing dictionary.
+            Dictionary<string, Uri> baseUriDictionary = new Dictionary<string, Uri>();
+            if (node.Properties.ContainsKey(BaseUriDictionaryName))
+            {
+                if (!TryDeserializePropertyDictionary(node.Properties[BaseUriDictionaryName], out baseUriDictionary))
+                {
+                    // If for some reason we don't have a valid dictionary in the originalUriBaseIds, we move it to another location.
```
michaelcfanning: >If for some reason  [](start = 23, length = 19)

ouch. we have to preserve bad data produced due to previously existing bug

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+                // If the old uri is absolute and we need to rebase it, we should.
+                if (Uri.TryCreate(key, UriKind.Absolute, out oldUri) && oldUri.IsAbsoluteUri && _baseUri.IsBaseOf(oldUri))
+                {
+                    Uri newUri = _baseUri.MakeRelativeUri(oldUri);
+
+                    // Ensure the filedata reflects the correct base URI details.
+                    if (data != null)
+                    {
+                        if (data.ParentKey != null)
+                        {
```
michaelcfanning: consider eliminating nesting and improving readability by doing something like

if (data == null || data.ParentKey == null) { continue; }

I actually will put this sort of thing on the same line as above.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+                            Uri parentUri;
+                            if (Uri.TryCreate(data.ParentKey, UriKind.Absolute, out parentUri))
+                            {
+                                if (oldUri.IsAbsoluteUri && _baseUri.IsBaseOf(parentUri))
```
michaelcfanning: >oldUri.IsAbsoluteUri && _baseUri.IsBaseOf [](start = 36, length = 41)

look, in this case you put the comparison to isAbsoluteUri first

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+                                    data.ParentKey = _baseUri.MakeRelativeUri(new Uri(data.ParentKey)).ToString();
+                                }
+                            }
+                        }
+
+                        data.Uri = newUri;
+                        data.UriBaseId = _baseName;
+                    }
+
+                    if(newDictionary.ContainsKey(newUri.ToString()))
```
michaelcfanning: space. shall we configure stypecop or code formatter via automation? good idea for avoiding my trivial PR notes :)

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+                    {
+                        throw new InvalidOperationException("Cannot rebase this file, as two URIs will collide in the file dictionary.");
+                    }
+                    newKey = newUri.ToString();
```
michaelcfanning: >newKey = newUri.ToString() [](start = 20, length = 26)

this code makes me realize we could attempt to normalize URL keys to avoid issues where URLs differ in rendering but aren't distinct when properly encoded or otherwise normalized. maybe this is a new action?

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -0,0 +1,187 @@
...
+        internal static bool TryDeserializePropertyDictionary(SerializedPropertyInfo serializedProperty, out Dictionary<string, Uri> dictionary)
+        {
+            try
+            {
+                dictionary = JsonConvert.DeserializeObject<Dictionary<string, Uri>>(serializedProperty.SerializedValue, _settings);
+
+                return true;
+            }
+            // Didn't deserialize correctly
+            catch (Exception)
```
michaelcfanning: >Exception [](start = 19, length = 9)

please catch a more specific exception, i.e., the one that's actually generated when we hit our known serialization bug

src/Sarif.Multitool/AbsoluteUriCommand.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis.Sarif.Processors;
```
michaelcfanning: >Microsoft [](start = 6, length = 9)

these namespaces not ordered properly

src/Sarif/Visitors/RebaseUriVisitor.cs
Evmaus-MS: VisitFileData (the only option) does not allow us to change the keys as well, and keeping this all together seemed cleaner.

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -0,0 +1,207 @@
...
+            newRun.Properties.Should().ContainKey(RebaseUriVisitor.BaseUriDictionaryName);
+            
+            newRun.Files.Keys.Where(k => k.StartsWith(@"C:\src\")).Should().BeEmpty();
+        }
+
+        [Fact]
+        public void RebaseUriVisitor_VisitRun_DoesNotPatchFileDictionaryKeysWhenNotABaseUri()
+        {
+            // Slightly roundabout.  We want to randomly test this, but we also want to be able to repeat this if the test fails.
+            int randomSeed = (new Random()).Next();
```
michaelcfanning: if you provide randomization, it might be helpful to ensure that random seed is persisted to the log file. This seed may be required in order to reproduce a specific failure.

src/Sarif.UnitTests/Visitors/MakeUriAbsoluteVisitorTest.cs
```diff
@@ -0,0 +1,183 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class MakeUriAbsoluteVisitorTest
+    {
+        private Run GenerateRunForTest(Dictionary<string, Uri> uriBaseIdMapping)
+        {
+            Run run = new Run();
+            run.Files = new Dictionary<string, FileData>()
+            {
+                {"src/file1.cs", new FileData(){Uri=new Uri("src/file1.cs", UriKind.Relative), UriBaseId="%TEST1%",ParentKey=null } },
```
michaelcfanning: >{U [](start = 47, length = 2)

NIt how about some spaces after these braces?

src/Sarif.UnitTests/Visitors/MakeUriAbsoluteVisitorTest.cs
```diff
@@ -0,0 +1,183 @@
...
+
+                run.Properties[RebaseUriVisitor.BaseUriDictionaryName] = RebaseUriVisitor.ReserializePropertyDictionary(uriBaseIdMapping);
+            }
+            return run;
+        }
+
+        [Fact]
+        public void MakeUriAbsoluteVisitor_VisitPhysicalLocation_SetsAbsoluteURI()
+        {
+            PhysicalLocation location = new PhysicalLocation() {UriBaseId="%TEST%", Uri=new Uri("src/file.cs", UriKind.Relative), };
```
michaelcfanning: >UriBaseId [](start = 64, length = 9)

space

src/Sarif.UnitTests/Processors/Log/RebaseUriStageTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(5)]
+        public void RewriteUri_RewritesAllFiles(int fileCount)
+        {
+            // Slightly roundabout.  We want to randomly test this, but we also want to be able to repeat this if the test fails.
+            int randomSeed = (new Random()).Next();
+            Random random = new Random(randomSeed);
```
michaelcfanning: put all this construction in a helper, then you can add logging there.

src/Sarif.UnitTests/Processors/GenericTests/GenericProcessorTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using System.Text;
+using Xunit;
+using FluentAssertions;
+
+namespace Microsoft.CodeAnalysis.Sarif.Processors
+{
+    public class GenericProcessorTests
+    {
+        private List<int> GenerateRandomIntList()
```
michaelcfanning: >GenerateRandomIntList [](start = 26, length = 21)

Create some API use tests against your extension methods? [NOTE: this comment doesn't apply to this file but to a test class that doesn't exist yet :)]

src/Sarif.UnitTests/Processors/Log/RebaseUriStageTests.cs
michaelcfanning: or maybe just move this into the RandomSarifLogGenerator, i.e., have the Generate method just create the Random objects and log the seed value. 

this suggestion might not be a good one, depending on whether it would require plumbing through some output/logging object.

---
In reply to: [175206341](https://github.com/Microsoft/sarif-sdk/pull/803#discussion_r175206341) [](ancestors = 175206341)

src/Sarif.Multitool/RebaseUriOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+            )]
+        public string BasePath { get; internal set; }
+
+        [Option(
+            'n',
+            "base-path-name",
+            Required = true,
+            HelpText = "Variable to use for the base path (e.x. 'SRCROOT')"
+            )]
+        public string BasePathName { get; internal set; }
```
michaelcfanning: >BasePathName [](start = 22, length = 12)

You should call this BasePathToken, I think. BasePathPlaceholder also fine. But SARIF prefers token for this concept. 

the actual name of the relevant thing in SARIF is uriBaseId, maybe we should change this to uriBaseToken

src/Sarif.Multitool/RebaseUriOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using CommandLine;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    [Verb("rebaseuri", HelpText = "Rebase the URIs in one or more sarif files.")]
+    internal class RebaseUriOptions : MultitoolOptionsBase
+    {
+        [Value(0,
+            MetaName = "<files>",
```
michaelcfanning: >MetaName [](start = 12, length = 8)

we need this argument to be similar to other commands that provide references to files, can we look at that? check out the way the driver framework declares these for the Analyze base command

src/Sarif.Multitool/RebaseUriCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    internal static class RebaseUriCommand
+    {
+        public static int Run(RebaseUriOptions rebaseOptions)
+        {
+            Uri baseUri;
+            if (!Uri.TryCreate(rebaseOptions.BasePath, UriKind.Absolute, out baseUri))
+            {
+                Console.WriteLine($"BasePath {rebaseOptions.BasePath} was not an absolute URI.  It must be.");
+                return -1;
```
michaelcfanning: > return -1; [](start = 15, length = 11)

your result handling is diverging from other commands. check out the RewriteCommand. I don't mind altering the convention in that command but let's make sure all of them stay consistent.

The RewriteCommand would handle this like so, all exits are accomplished by throwing an exception. the core command wraps catch(Exception ex), writes ex.ToString() to the console, and returns 1 to indicate an error.

src/Sarif.Multitool/MultitoolFileHelpers.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                ContractResolver = SarifContractResolver.Instance
+            };
+
+            string logText = File.ReadAllText(filePath);
+
+            return JsonConvert.DeserializeObject<SarifLog>(logText, settings);
+        }
+
+        public static void WriteSarifFile(SarifLog sarifFile, string outputName, Formatting formatting)
```
michaelcfanning: >WriteSarifFile [](start = 27, length = 14)

this looks like a useful extension off SarifLog, or even a direct method on it.

src/Sarif.Multitool/MultitoolFileHelpers.cs
```diff
@@ -0,0 +1,38 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Helper functions for the multitool.
+    /// </summary>
+    static class MultitoolFileHelpers
+    {
+        public static SarifLog ReadSarifFile(string filePath)
```
michaelcfanning: >ReadSarifFile [](start = 31, length = 13)

we could provide an extension method or direct "LoadFrom" method where a SarifLog hydrates itself based on a file path.

michaelcfanning:
>         public static int Run(MergeOptions mergeOptions)

whoops! merge command needs to wrap all this in a catch(Exception) handler and return a non-zero result for failure. See RewriteCommand pattern.

---
Refers to: src/Sarif.Multitool/MergeCommand.cs:15 in 2d75fff. [](commit_id = 2d75fffd75142ec4810b0a82f309fa029d04a117, deletion_comment = False)

src/Sarif.Multitool/AbsoluteUriOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    [Verb("absoluteuri", HelpText = "Turn all relative Uris into absolute URIs (to be used after rebaseUri is run)")]
+    internal class AbsoluteUriOptions : MultitoolOptionsBase
+    {
+        [Value(0,
+            MetaName = "<files>",
+            HelpText = "Files to process (wildcards ? and * allowed).",
+            Required = false)]
+        public IList<string> Files { get; internal set; }
```
michaelcfanning: >Files [](start = 29, length = 5)

Please rationalize all file argument command across driver framework and this code.

src/Sarif.UnitTests/Visitors/MakeUriAbsoluteVisitorTest.cs
Evmaus-MS: Resolved.

src/Sarif.UnitTests/Processors/Log/RebaseUriStageTests.cs
Evmaus-MS: Done.

src/Sarif.UnitTests/Visitors/MakeUriAbsoluteVisitorTest.cs
Evmaus-MS: Done

src/Sarif.UnitTests/Processors/GenericTests/GenericProcessorTests.cs
Evmaus-MS: Added a few.

src/Sarif.Multitool/RebaseUriOptions.cs
Evmaus-MS: Renamed to BasePathToken.

src/Sarif.Multitool/RebaseUriOptions.cs
Evmaus-MS: Moved to a MultipleToolFileOptionsBase that handles this properly (with a helper function that works like the driver's).

PR #809, opened 2018-03-28
Fix #765: Related location hyperlinks in error list

### LINKED ISSUE #765: Viewer: support inline links, rather than separate "related links" view
Mend places the link inline in the result message, which is a slightly nicer UI that providing “related locations” in a separate pane, as the SARIF viewer does.

EasyRhin0:


EasyRhin0:
The "Replace dummy Foo method with ToString override" change is to address this voodoo: https://stackoverflow.com/questions/46084270/adding-removing-methods-causes-exception
Have no idea what could cause that behavior. (See for yourself by simply removing ToString.)

PR #811, opened 2018-03-30
Restrict download prompt to HTTP/HTTPS, use MessageDialog to prompt

EasyRhin0:


src/Sarif.Viewer.VisualStudio/Uri.extensions.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Sarif.Viewer.Sarif
     static class UriExtensions
     {
         // The acceptable URI schemes
-        static List<string> s_schemes = new List<string>() { Uri.UriSchemeHttp, Uri.UriSchemeHttps };
+        static List<string> s_httpSchemes = new List<string>() { Uri.UriSchemeHttp, Uri.UriSchemeHttps };
```
michaelcfanning: >List<string>  [](start = 15, length = 13)

you can use a hashset<string> here, pass the ordinal string comparer and then call Contains directly off it.

Issue #84, opened 2016-01-22
ERR0998 should call out the id of the rule that failed


rtaket:
Since ERR0998 is raised for rule exceptions, the rule id of the offending rule should be explicitly stated in the ERR0998 result node.  Currently, the rule name is listed in the format args, but the rule id does not exist.

michaelcfanning:
This should be addressed by the new toolNotifications object in the SARIF standard, that provide a ruleId property.

PR #814, opened 2018-03-30
Add attachment object to schema

michaelcfanning:
@EasyRhin0 

@lgolding FYI

### DIFF
--- a/src/Sarif/Schemata/Sarif.schema.json
+++ b/src/Sarif/Schemata/Sarif.schema.json
@@ -34,7 +34,7 @@
       "additionalProperties": false,
       "type": "object",
       "properties": {
-        
+
         "region": {
           "description": "The region of an analysis tool's log file that was transformed into the result object.",
           "$ref": "#/definitions/region"
@@ -205,6 +205,25 @@
       "required": [ "message", "locations" ]
     },
 
+    "attachment": {
+      "description": "A file relevant to a tool invocation or to a result.",
+      "type": "object",
+      "properties": {
+
+        "description": {
+          "description": "A message describing the role played by the attachment.",
+          "type": "string"
+        },
+
+        "fileLocation": {
+          "description": "The location of the attachment.",
+          "$ref": "#/definitions/fileLocation"
+        }
+      },
+
+      "required": [ "fileLocation" ]
+    },
+
     "codeFlow": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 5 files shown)


src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -205,6 +205,25 @@
       "required": [ "message", "locations" ]
     },
 
+    "attachment": {
+      "description": "A file relevant to a tool invocation or to a result.",
+      "type": "object",
+      "properties": {
```
lgolding: Add the following to the description of the `attachment` object:
```
"additionalProperties": false
```

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -205,6 +205,25 @@
...
+
+        "description": {
+          "description": "A message describing the role played by the attachment.",
+          "type": "string"
```
lgolding: The type of the description property should be a `message` object, so instead of
```
"type": "string"
```
you need to write
```
"$ref": "#/definitions/message"
```

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -416,6 +435,26 @@
...
+
+        "uriBaseId": {
+          "description": "A string which indirectly specifies the absolute URI with respect to which a relative URI in the \"uri\" property is interpreted.",
+          "type": "string"
+        }
+      },
+
+      "required": [ "uri" ]
+    },
     "fix": {
```
lgolding: @michaelcfanning Thank you for alphabetizing :-) And yes, in this intermediate state, you need `fileLocation` for `attachments` but it wouldn't be used anywhere else.

I suggest you do the changes in chronological order, since later changes often depend on earlier changes, as you saw here.

lgolding: @michaelcfanning @EasyRhin0 Also thanks for copying the definition, with its oh-so-carefully-crafted text :-), from the copy of the schema in the OASIS repo.

michaelcfanning:
@lgolding , all this content was copied from your presumed final version of the SARIF JSON. why does that document not contain your suggestions above?

lgolding:
@michaelcfanning @EasyRhin0 You found bugs in my schema (remember, I said you might :-)).

So yes, I will make those changes in the OASIS schema. **BUT!** You can't actually declare `attachment.description` as `message` yet, because you won't have a `message` object until you do #84.

lgolding:
@EasyRhin0 Did you notice this merge failed due to merge conflict?

PR #820, opened 2018-04-02
Change result.fingerprint to string dictionary

EasyRhin0:
And update unit tests + assets

### DIFF
--- a/src/Sarif.Converters/FortifyConverter.cs
+++ b/src/Sarif.Converters/FortifyConverter.cs
@@ -95,8 +95,19 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             if (!string.IsNullOrWhiteSpace(fortify.InstanceId))
             {
-                result.ToolFingerprintContributions = new Dictionary<string, string>();
-                result.ToolFingerprintContributions.Add("InstanceId", fortify.InstanceId);
+                if (result.ToolFingerprintContributions == null)
+                {
+                    result.ToolFingerprintContributions = new Dictionary<string, string>();
+                }
+
+                if (result.ToolFingerprintContributions.ContainsKey("InstanceId"))
+                {
+                    result.ToolFingerprintContributions["InstanceId"] = fortify.InstanceId;
+                }
+                else
+                {
+                    result.ToolFingerprintContributions.Add("InstanceId", fortify.InstanceId);
+                }
             }
 
             List<string> messageComponents = new List<string>();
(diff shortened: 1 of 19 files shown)


src/Sarif.Converters/FortifyConverter.cs
```diff
@@ -92,7 +92,13 @@ public static Result ConvertFortifyIssueToSarifIssue(FortifyIssue fortify)
         {
             var result = new Result();
             result.RuleId = fortify.Category;
-            result.ToolFingerprintContribution = fortify.InstanceId;
+
+            if (!string.IsNullOrWhiteSpace(fortify.InstanceId))
+            {
+                result.ToolFingerprintContributions = new Dictionary<string, string>();
```
lgolding: Are we sure that `result.ToolFingerprintContributions` doesn't already have a value at this point? And are we sure that if it does exist, it doesn't already contain an entry for `InstanceId`?

src/Sarif/CodeGenHints.json
```diff
@@ -256,6 +256,11 @@
       }
     }
   ],
+  "Result.ToolFingerprintContributions": [
+    {
+      "kind": "DictionaryHint"
+    }
+  ],
```
lgolding: Yay!

src/Sarif.Converters/FortifyConverter.cs
EasyRhin0: I am not sure. I will add some checks.

EasyRhin0: Actually yes we do know, because the parameterless Result constructor is being called, and the fingerprints property isn't initialized. Still, I think doing the checks is a good idea.

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -85,8 +85,19 @@ internal Result CreateResult(FxCopLogReader.Context context)
...
+                }
+
+                if (result.ToolFingerprintContributions.ContainsKey("UniqueId"))
+                {
+                    result.ToolFingerprintContributions["UniqueId"] = uniqueId;
+                }
+                else
+                {
+                    result.ToolFingerprintContributions.Add("UniqueId", uniqueId);
+                }
```
lgolding: This is fine. After this PR, consider adding a helper method
```
static void SafeAddDictionaryElement(Dictionary<TKey, TValue> dict, TKey key, TValue value)
```
... to some `Utilities` class to encapsulate this repeated logic. Perhaps in the `SarifUtilities` class.

PR #821, opened 2018-04-02
Add dictionary helper method

EasyRhin0:


src/Sarif/SarifUtilities.cs
```diff
@@ -124,5 +124,17 @@ public static string FormatMessage(this Exception exception)
             // System.InvalidOperationException: Operation is not valid due to the current state of the object.
             return exception.ToString().Split(new string[] { Environment.NewLine }, StringSplitOptions.None)[0];
         }
+
+        public static void AddOrUpdateDictionaryEntry<TKey, TValue>(IDictionary<TKey, TValue> dictionary, TKey key, TValue val)
+        {
```
lgolding: IMO the check for existence and the initial dictionary creation should also be in this helper.

EasyRhin0: OK. I left it out because the property type is IDictionary, but I'll just cast the param.

lgolding: @EasyRhin0  I'd rather change the property type of the parameter. _If_ you agree that this method can do the creation, it's appropriate for the parameter to specify the concrete type of the object it's creating.

PR #822, opened 2018-04-02
Issue #15: Document how converters should provide notifications

EasyRhin0:


src/Sarif.Driver.UnitTests/SarifHelpers.cs
```diff
@@ -23,17 +23,17 @@ internal static class SarifHelpers
                 resultAction(result);
             }
 
-            if (run.ToolNotifications != null)
+            if (run.Invocation.ToolNotifications != null)
```
lgolding: Any need to check for `Invocation` non-null? How about `run.Invocation?.ToolNotifications`?

src/Sarif.Multitool.FunctionalTests/SarifMultitoolTestBase.cs
```diff
@@ -57,13 +57,13 @@ private static void SelectiveCompare(SarifLog actualLog, SarifLog expectedLog)
 
             SelectiveCompare(actualResults, expectedResults);
 
-            Notification[] actualConfigurationNotifications = SafeListToArray(actualLog.Runs[0].ConfigurationNotifications);
-            Notification[] expectedConfigurationNotifications = SafeListToArray(expectedLog.Runs[0].ConfigurationNotifications);
+            Notification[] actualConfigurationNotifications = SafeListToArray(actualLog.Runs[0].Invocation.ConfigurationNotifications);
```
lgolding: Again make sure `Invocation` exists. (It's test code, so maybe you actually do know that for sure.)

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -466,14 +456,14 @@ public bool ShouldLog(ResultLevel level)
 
         public void LogToolNotification(Notification notification)
         {
-            _run.ToolNotifications = _run.ToolNotifications ?? new List<Notification>();
-            _run.ToolNotifications.Add(notification);
+            _run.Invocation.ToolNotifications = _run.Invocation.ToolNotifications ?? new List<Notification>();
```
lgolding: Create `Invocation` if it doesn't exist? Or are we sure it exists at this point?

PR #826, opened 2018-04-04
Move location anaysis target to result

EasyRhin0:


src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -97,10 +97,10 @@ internal Result CreateResult(TSLintLogEntry entry)
 
             Uri analysisTargetUri = new Uri(entry.Name, UriKind.Relative);
```
lgolding: Let's rename `analysisTargetUri` to `uri`. We're now reserving "analysis target" to mean "the thing I was asked to scan that led me to discover a result in some other file."

src/Sarif.FunctionalTests/ConverterTestData/AndroidStudio/AndroidStudio_issueLog5_raw.xml.sarif
```diff
@@ -53,10 +53,6 @@
           "message": "Typo: In word 'userguide'",
           "locations": [
             {
-              "analysisTarget": {
-                "uri": "gradle.properties",
-                "uriBaseId": "$PROJECT_DIR$"
-              },
```
lgolding: Yes! Since they're both the same, we don't need to populate `result.analysisTarget`.

src/Sarif.FunctionalTests/ConverterTestData/AndroidStudio/fileEmpty.xml.sarif
```diff
@@ -24,14 +24,14 @@
       },
       "results": [
         {
+          "analysisTarget": {
+            "uri": "northwindtraders-droid-2/src/main/java/com/northwindtraders/droid/model/Model.java",
+            "uriBaseId": "$PROJECT_DIR$"
+          },
```
lgolding: No: If a `location` object has only one `physicalLocation`, populate `location.physicalLocation`, not `result.analysisTarget`.

src/Sarif.FunctionalTests/ConverterTestData/FxCop/FxCopExampleReport2.xml.sarif
```diff
@@ -55,14 +55,14 @@
           }
         },
         {
+          "analysisTarget": {
```
lgolding: As above: `location.physicalLocation`, not `result.analysisTarget`.

lgolding: Do this here and throughout the FxCop test files. This is related to the code change we just discussed on the phone.

src/Sarif.FunctionalTests/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
```diff
@@ -9,22 +9,7 @@
...
-            {
-              "value": "6727693E48A1C8ED14251D17BC8A16D162DF879C",
-              "algorithm": "sha1"
-            },
-            {
-              "value": "B4B0A1F7894E935F34F0DE81407DF9676C4A894FBCB0AD2BF635BB09F430B5C4",
-              "algorithm": "sha256"
-            }
-          ]
+          "mimeType": "text/x-cpp"
```
lgolding: Do you know why the embedded contents and the hashes went away? That shouldn't have happened.

lgolding: I suspect it has to do with you not having updated the PREfast converter. If the locations in the PREfast SARIF files aren't correctly populated, it probably can't locate the files to fill in this information.

src/Sarif.FunctionalTests/ConverterTestData/PREfast/MissingPathElement.xml.sarif
```diff
@@ -0,0 +1,44 @@
+{
```
lgolding: What's the purpose of this new file?

src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/cancelspinlock_bug1.tt.sarif
```diff
@@ -229,7 +229,7 @@
                   "kind": "assignment",
                   "importance": "unimportant",
                   "properties": {
-                    "currentDataValues": "^&sdv_ControllerIrp=8520",
+                    "currentDataValues": "^\u0026sdv_ControllerIrp=8520",
```
lgolding: This is valid JSON -- `\u0026` is indeed the `"&"` character, and any character _may_ be escaped in a JSON string -- but this character does not _have_ to be escaped. Do you have any idea why this crept in? If I go further down the review, will I find that we upgraded to a new version of Newtonsoft.Json that does this conversion for some reason?

src/Sarif.FunctionalTests/ConverterTestData/StaticDriverVerifier/cancelspinlock_bug1.tt.sarif
```diff
@@ -2051,7 +2051,7 @@
                   "kind": "assignment",
                   "importance": "unimportant",
                   "properties": {
-                    "currentDataValues": "^pirp->IoStatus.Information=72",
+                    "currentDataValues": "^pirp-\u003eIoStatus.Information=72",
```
lgolding: Again, `\u003e` is `">"`, but why?

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
```diff
@@ -78,6 +78,9 @@
       },
       "results": [
         {
+		  "analysisTarget": {
+			"uri": "file:///home/buildAgent/src/collections/list.cpp"
+		  },
```
lgolding: Yes! (Nit: indentation) This (list.cpp _vs_ list.h) is the canonical example for this feature.

src/Sarif.FunctionalTests/UpdateBaselines.ps1
```diff
@@ -38,13 +48,13 @@ function Build-Baselines($toolName)
 		
         # Actually run the converter
         Remove-Item $outputTemp -ErrorAction SilentlyContinue
-        & Write-Host "$utility convert --input ""$input"" --tool $toolName --output ""$outputTemp"" "
-        &$utility convert --input "$input" --tool $toolName --output "$outputTemp" --pretty --persist-file-contents
+        & Write-Host "$utility convert ""$input"" --tool $toolName --output ""$outputTemp"" -p --persist-file-contents"
+        &$utility convert "$input" --tool $toolName --output "$outputTemp" -p --persist-file-contents
```
lgolding: Maybe _this_ has something to do with the reason the file contents went away in the PREfast examples.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -191,8 +191,7 @@ public void SarifLogger_ScrapesFilesFromResult()
                         {
                             new Location
                             {
-                                AnalysisTarget = new PhysicalLocation {  Uri = new Uri(@"file:///file0.cpp")},
```
lgolding: Shouldn't `AnalysisTarget` have moved up to the `Result` object?

src/Sarif/Baseline/DefaultBaseline/LocationBaselineEquals.cs
```diff
@@ -13,12 +13,7 @@ public bool Equals(Location x, Location y)
         {
             if (!object.ReferenceEquals(x, y))
             {
-                // Target and Result file should match.
```
lgolding: Shouldn't we have moved the analysisTarget-related code into ResultBaselineEquals.cs?

**CAUTION** I don't know anything about the "Baseline" directory. I don't even know if the code here is used. Please look in the history, find the owner, and decide what's the right thing to do with this code.

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -861,6 +856,11 @@
       "type": "object",
       "properties": {
 
+        "analysisTarget": {
+          "description": "Identifies the file that the analysis tool was instructed to scan. This need not be the same as the file where the result actually occurred.",
+          "$ref": "#/definitions/fileLocation"
+        },
+
```
lgolding: Let's move this down to right above the `locations` property, since they're related.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -278,11 +278,6 @@ private void CaptureFilesInResult(Result result)
             {
                 foreach (Location location in result.Locations)
                 {
-                    if (location.AnalysisTarget != null)
```
lgolding: Don't we want to "capture" `result.AnalysisTarget` if it exists? Perhaps _that's_ why the file contents were missing.

See, now I have three different theories about that :-)

src/Sarif/Baseline/DefaultBaseline/LocationBaselineEquals.cs
EasyRhin0: This is all new work from Everett that just went in last week. I'm adding a new equality class for FileLocation so that check can be added to ResultBaselineEquals.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
EasyRhin0: Yep

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -149,32 +149,32 @@ public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
             string file = problem.File;
             if (!String.IsNullOrEmpty(file))
             {
-                location.PhysicalLocation = new PhysicalLocation
-                {
-                    Region = problem.Line <= 0 ? null : Extensions.CreateRegion(problem.Line)
-                };
+                location.PhysicalLocation = new PhysicalLocation();
+                location.PhysicalLocation.Region = problem.Line <= 0 ? null : Extensions.CreateRegion(problem.Line);
```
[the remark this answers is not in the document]
EasyRhin0: Refactoring/mind-changing collateral damage :)

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -109,26 +109,32 @@ internal Result CreateResult(FxCopLogReader.Context context)
...
+                location.PhysicalLocation = new PhysicalLocation
                 {
-                    result.AnalysisTarget = new FileLocation
-                    {
-                        Uri = uri
-                    };
-                }
+                    Uri = new Uri(sourceFile, UriKind.RelativeOrAbsolute),
+                    Region = context.Line == null ? null : Extensions.CreateRegion(context.Line.Value)
+                };
```
lgolding: Gorgeous!

src/Sarif/Baseline/DefaultBaseline/FileLocationBaselineEquals.cs
```diff
@@ -0,0 +1,42 @@
...
+        public bool Equals(FileLocation x, FileLocation y)
+        {
+            if (!object.ReferenceEquals(x, y))
+            {
+                if (x == null || y == null)
+                {
+                    return false;
+                }
+
+                // Only compare URI (so file path/relative file path).  UriBaseId may change run over run.
```
lgolding: `uriBaseId` won't change. It will always be a token like `"SRCROOT"`. The absolute URI to which that token refers might change, but the token won't change.

lgolding:
@EasyRhin0 It worked in the previous commit, right? (The "attachments" change.) So it has to be due to something in this change.

EasyRhin0:
I didn’t actually use multitool for that, I just did a find and replace since resultFile is unique. This repros in master. It might be PREfast-specific, because I just converted an FxCop log and the files did get embedded.

lgolding:
@EasyRhin0 In that case, the behavior of the PREfast converter has changed. Please file a bug "PREfast converter no longer embeds content" and note that you ran the converter in the Multitool.

Also: Does the main branch have the same issue with the change to the Unicode escaping behavior in strings? That is, what used to be `"&"` is now `"\u0026"`? If so, file a bug for that too.

I think after that I still had a couple more comments on the review -- please run through all the comments and let me know what still needs to be addressed. Then we can get this in and move on.

EasyRhin0:
OK @lgolding I believe all feedback has been addressed.

PR #832, opened 2018-04-06
Remove analysisToolLogFileContents, change result.conversionProvenance

EasyRhin0:
No changes to converter test logs, so we have a unit test gap.


src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
```diff
@@ -171,7 +171,6 @@
           ],
           "relatedLocations": [
             {
-              "message": "\"count\" was declared here.",
```
lgolding: Changing the type of `result.relatedLocations` from `annotatedCodeLocation[]` to `location[]` is indeed part of the larger location redesign (oasis-tcs/sarif-spec#130). We didn't originally envision that change as part of this PR, but it's fine if you do it.

But to do that, you have to give the `location` object the necessary features of the `annotatedCodeLocation` object. That means it needs a `message` property. Now if you look at the [Schema changes](https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/Schema%20changes%20between%20SARIF%20v1%20and%20v2.md) . document, you'll see that `location.message` should be of type `message`, not `string`. But we haven't yet introduced the `message` object; that's part of oasis-tcs/sarif-spec#84 (localization). So declare `location.message` as `string` for now.

lgolding: Ah, I just saw your comment acknowledging the unit test gap. Our plan of record is now to keep the unit tests up to date as we go, so that needs to be part of this PR.

lgolding:
@EasyRhin0 @michaelcfanning I'm having second thoughts about the need to add unit tests for SARIF features that no direct producer or converter yet exercises.

This PR changes the type of `result.conversionProvenance` from `annotatedCodeLocation[]` (that type is now gone) to `physicalLocation[]`. What is there to test? No converter populates that property. No direct converter populates it either (and even if one did, we're not updating the direct producers yet). So the only test we could write would be one that created an in-memory `SarifLog` object that included `Result.ConversionProvenance`, serialize it, and examine the resulting string to ensure it serialized correctly.

Does it make sense to write such a test? Remember that at build time we generate the SARIF object model classes from the JSON schema, and then at runtime, Newtonsoft.Json serializes the result. So if a unit test like this were to fail, it would mean one of three things:

1. We made a mistake in the JSON schema.
2. Our code generator has a bug.
3. Newtonsoft.Json's serializer has a bug.

I suppose the first of these is plausible. But is it really worth writing tests for each schema change?

The floor is yours.

EasyRhin0:
Not sure whose floor, but it seems unnecessary until there is code that actually uses it. If we are concerned about third-parties, we could file bugs and get to this later. I don't think it makes sense to add that work on top of everything we have to do to complete v2.

lgolding:
@EasyRhin0 @michaelcfanning Yes, go ahead and merge. Then let's chat about the next step. No point blocking progress on this question.

Issue #90, opened 2016-01-27
We need documentation on how to use the Sarif.Sdk to produce Sarif files


davidknise:
We took a crack at integrating Sarif into two of our prototyping tools and were able to do so, but through trial and error.  Some sample code in the README file for generating a Sarif output file using the SDK would be helpful!

michaelcfanning:
Yes. The SDK needs development here and it would be helpful if the spec provided some guidance on core concepts to express. You have the VS Code extension to validate against the current schema, don't you? I can provide you an experimental VSIX that puts SARIF into the error list, also helpful for debugging problems. @lgolding  and I have discussed creating a SARIF language service in VS that provides more exhaustive validation. Finally, we need to augment our JSON schema to express constructs such as oneOf that are currently absent (an artifact of the automated process we use to produce the schema, which is not sufficiently expressive to cover all JSON schema validation mechanisms). @lgolding  for visibility.

Can you send your sample files to Larry and me?

michaelcfanning:
We now have samples

PR #835, opened 2018-04-08
#90 introduce file location object

EasyRhin0:


src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
```diff
@@ -33,8 +33,10 @@ public void SemmleQLConverter_SimpleCsv()
           ""locations"": [
             {
               ""physicalLocation"": {
-                ""uri"": ""RelativePath"",
-                ""uriBaseId"": ""$srcroot"",
+                ""fileLocation"": {
+                  ""uri"": ""RelativePath"",
+                  ""uriBaseId"": ""$srcroot""
```
lgolding: Yes! That's the deal.

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -151,14 +151,15 @@ public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
             {
                 location.PhysicalLocation = new PhysicalLocation
                 {
+                    FileLocation = new FileLocation(),
```
lgolding: This will lead to the same problem we had in the other review (`"fileLocation": {}`) if `uri` is `null`. Is `uri` guaranteed non-`null` here? If not, only populate `FileLocation` if `uri` is non-`null`.

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
```diff
@@ -22,9 +22,11 @@
...
       "invocation": {
         "commandLine": "CodeScanner @collections.rsp",
-        "responseFiles": {
-          "collections.rsp": "-input src/collections/*.cpp -log out/collections.sarif -rules all -disable C9999"
-        },
+        "responseFiles": [
+          {
+            "uri": "collections.rsp"
+          }
+        ],
```
lgolding: It doesn't look like you added the response file to the `run.files` dictionary, so now the response file contents don't appear in the log file. Please see the sample JSON I sent in email.

src/Sarif.Viewer.VisualStudio.UnitTests/Converters/CallTreeNodeToTextConverterTests.cs
```diff
@@ -148,7 +147,6 @@ public void CallTreeNodeToTextConverter_HandlesMessage()
                         },
                         PhysicalLocation = new PhysicalLocation
                         {
-                            Uri = new System.Uri(sourceFile),
```
lgolding: Why did this disappear, instead of being wrapped in a `FileLocation`?

src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
```diff
@@ -32,16 +32,20 @@ public CodeFlowLocation Location
...
-                    // are used to navigate to the source location and highlight the line.
-                    Uri uri = value.Location.PhysicalLocation.Uri;
-                    if (uri != null)
+                    // If the backing CodeFlowLocation has a PhysicalLocation, set the 
+                    // Region property. If it has a FileLocation, set the FilePath.
+                    // The FilePath and Region properties are used to navigate to the
+                    // source location and highlight the line.
+                    Region = value.Location.PhysicalLocation.Region;
+
+                    if (value.Location.PhysicalLocation.FileLocation != null)
```
lgolding: You can collapse this to
```
Uri uri = value.Location.PhysicalLocation.FileLocation?.Uri;
if (uri != null)
{
    FilePath = uri.ToPath();
}
```

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlowLocationExtensions.cs
```diff
@@ -19,11 +19,14 @@ public static Microsoft.Sarif.Viewer.Models.CodeFlowLocationModel ToAnnotatedCod
                 {
                     model.Region = codeFlowLocation.Location.PhysicalLocation.Region;
 
-                    Uri uri = codeFlowLocation.Location.PhysicalLocation.Uri;
-                    if (uri != null)
+                    if (codeFlowLocation.Location.PhysicalLocation.FileLocation != null)
                     {
-                        model.FilePath = uri.ToPath();
-                        model.UriBaseId = codeFlowLocation.Location.PhysicalLocation.UriBaseId;
+                        Uri uri = codeFlowLocation.Location.PhysicalLocation.FileLocation.Uri;
```
lgolding: As in the previous comment, you can collapse this to a single test with `FileLocation?.Uri`.

src/Sarif/Baseline/DefaultBaseline/PhysicalLocationBaselineEquals.cs
```diff
@@ -17,8 +17,8 @@ public bool Equals(PhysicalLocation x, PhysicalLocation y)
                     return false;
                 }
 
-                // Only compare URI (so file path/relative file path).  UriBaseId and Region may change run over run.
-                if (x.Uri != y.Uri)
+                // Only compare URI (so file path/relative file path). UriBaseId and Region may change run over run.
```
lgolding: No. As in a previous review: `UriBaseId` is a token like `SRCROOT`. The absolute URI the token refers to might change over time, but that token won't.

Why would `Region` change?

src/Sarif.Converters/AndroidStudioConverter.cs
EasyRhin0: Yes it will be non-null

src/Sarif.FunctionalTests/SpecExamples/Comprehensive.sarif
EasyRhin0: So what I have right now is that file.contents is a string. What you gave me was an object, so I figured it was a change we hadn't gotten to yet.

lgolding: Sorry, I can't keep straight in my head the exact order of changes in the [Schema changes](https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/Schema%20changes%20between%20SARIF%20v1%20and%20v2.md) document :-)

So the JSON should be:
```
{                                                  # A run object.
  "files": {                                       # run.files is a dictionary whose property values are file objects.
    "collections.rsp": {                           # A file object
      "contents": "-input src/collections/*.cpp -log out/collections.sarif -rules all -disable C9999".
    }
  },

  "invocation": {
    "commandLine": "CodeScanner @collections.rsp",
    "responseFiles": [                             # An array of fileLocation objects.
      {                                            # A fileLocation object.
        "uri": "collections.rsp"
      }
    ]
  }
}
```

src/Sarif.Viewer.VisualStudio.UnitTests/Converters/CallTreeNodeToTextConverterTests.cs
EasyRhin0: When you set the CallTreeNode.Location, there is a code path that ends up needing the IServiceProvider. I thought it best to just remove that property, since it's not involved in the test.

lgolding: Sure, if it's not needed for the test, this is fine.

src/Sarif/Baseline/DefaultBaseline/PhysicalLocationBaselineEquals.cs
EasyRhin0: Sorry I thought I had done this correctly this time...

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlowLocationExtensions.cs
EasyRhin0: Fixed

src/Sarif.Viewer.VisualStudio/Models/CallTreeNode.cs
EasyRhin0: Fixed

PR #837, opened 2018-04-09
Sarif v1 preliminary

michaelcfanning:
@lgolding @Evmaus-MS @EasyRhin0 

I have added a copy of the V1 OM inside the \VersionOne directory. All v1 types live in a new namespace (suffixed with VersionOne). Every type name from v1 is also suffixed with VersionOne. [Larry, I did this mostly by modifying jschema, just need a bit of clean-up to add it as a feature]

In addition to the new types, I added two empty visitors intended to transform v1 -> current and vice versa. I authored these types simply to demonstrate that the API can coexist in a common type. I look forward to authoring these transformers (v1 -> current is most critical) unless someone else picks it up. :)

NOTE 1 of 3: it is likely a few more files from the SDK need to come over, such as the contract resolver or anything else necessary for serialization or deserialization. This should be the only content we bring over. Once you get from disk to a V1 instance, we should insist on transforming through our visitor and then people can utilize the standard V2 helpers, then drop back down to V1 through the second visitor, if required.

NOTE 2 of 3: it is not clear that a visitor based transformation is the best approach given the structural changes between v1 and v2; I propose to start here but there may be a more sensible approach.

NOTE 3 of 3: Larry, it occurs to me that the transformation from V2 back to V1 should be done in a way such that a subsequent V1 -> V2 transformation would be completely lossless. We can do this by consistently storing things in various property bags. This approach could actually be formalized as part of the V2 spec (here's how to completely express all V2 concepts if you need to narrow them through the V1 format for some compat reason)


michaelcfanning:
>     public partial interface IRule : IPropertyBagHolder

This file was previously named IRule.cs.cs, so I fixed it

---
Refers to: src/Sarif/Core/IRule.cs:6 in 65de46f. [](commit_id = 65de46fe86fdaf97a127ab54c26286d3f903366a, deletion_comment = False)

michaelcfanning:
@lgolding, you could very well be correct, and these helpers aren't necessary. the SDK certainly builds without them. i did the work initially simply because I thought it best to bring over everything integrated with each generated type, allowing all OM-directed usage to be ported seamlessly.

in retrospect, i probably could have deferred the work, contingent on need. it may even be that we should remove it, we'll puzzle on it. merging for now.

lgolding:
@michaelcfanning As we ponder that, one consideration is that by providing those helpers, we suggest that the `VersionOne` types are more capable and useful than we intend them to be. By providing a bare object model, we discourage use of the v1 OM, consistend with your note:

> Once you get from disk to a V1 instance, we should insist on transforming through our visitor and then people can utilize the standard V2 helpers, then drop back down to V1 through the second visitor, if required.

lgolding:
@michaelcfanning Re: your NOTE 3 (use of property bags). I agree with the use of property bags for v1 -> v2 conversion. For example, the property bag on a v2 `threadFlow` object could contain v1's `kind`-dependent properties (`parameters`, `values`, _etc_) that we removed in v2. That would allow round-tripping.

But I don't think the v2 spec should or even can specify a use of property bags for converting a "native" v2 file to v1. It works ok for some features. For example, you could copy v2's `run.graphs` property into v1's `run.properties["graphs"]`. But `codeFlows` is another matter. A v2 `codeFlow` is an array of `threadFlows`, each of which contains an array of code flow locations. I don't know how to map that back to v1's "flat" code flow structure (even using the v1 `codeFlow`'s property bag).

And I don't see why we'd want to do any of that. The only purpose I see in v2 -> v1 conversion is round-tripping. IMO the message should be that if you want to take advantage of v2's advanced features, you should use the v2 SDK and v2-enabled tools such as the v2 viewer.

Thoughts?

michaelcfanning:
@lgolding I understand your point. But I can't resist pointing out that a v1 codeflow can easily be converted into a v2 construct. the v2 annotated code location object contains a thread id, and so it should be possible to separate individual thread flows (if they are intermingled) into the v2 array of thread flows model. In doing so, we apply a synchronization index which simply maps to the index of the original v1 codeflow array. we could perform a similar trick squashing v2 code flows into a single array (ordered by the synchronization index).

The round-tripping itself is pretty useful for testing. If you have a v1 SARIF, transform to v2, and back to v1, the two v1 files should be literally equal. I'm tempted to author the v1->v2 and v2->v1 visitors in lock-step going over each transformation in (tested) detail. 

Another reason that v2->v1 transform may be help (at least internally at Microsoft) is because various results management systems may continue to use v1 as a lingua franca until they themselves have time to learn how to parse/update to the new version. This would allow them to accept an updated tool that only supports v2 (as BinSkim shortly will, for example).

A v2 -> v1 transformer would also allow a tool like Binskim the ability to just go ahead and produce a v1 file. It would emit the standard v2, run the visitor and emit the result. Even more helpful, as in this case the results management system wouldn't need to inject the SARIF multitool to process v2 files (they'd just need to throw a new switch on the tool to emit the old version).

So, at this point, I've convinced myself that providing the v2->v1 transformation is valuable to ease onboarding and adoption. :)

Issue #105, opened 2016-02-05
URI serialization should be done with a Json.NET converter


lgolding:
... so that customers don't have to discover the obscure helper method `CreateUriForJsonSerialization`.

michaelcfanning:
Fixed

PR #840, opened 2018-04-10
29 rule config 102 run invocation array

EasyRhin0:


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -117,20 +117,20 @@ private int WriteRunToErrorList(Run run, string logFilePath, Solution solution)
                 }
             }
 
-            if (run.Invocation != null)
+            if (run.Invocations != null && run.Invocations.Count > 0)
```
lgolding: If this code is populating the error list with the notifications, then you should loop over all the invocations, not just the first. It's true that there's no code yet to populate multiple invocations, but we should be able to _consume_ them.

src/Sarif/CodeGenHints.json
```diff
@@ -314,40 +314,6 @@
...
-          "NotApplicable",
-          "Pass",
-          "Note",
-          "Warning",
-          "Error"
-        ],
-        "allowMemberCountMismatch": true
-      }
-    }
-  ],
```
lgolding: This should not go away. Instead, it should become the type of the new `RuleConfiguration.DefaultLevel` property (which, without this declaration, would be a string).

src/Sarif/Schemata/Sarif.schema.json
```diff
@@ -1124,6 +1117,41 @@
...
+      "properties": {
+
+        "enabled": {
+          "description": "Specifies whether the rule will be evaluated during the scan.",
+          "type": "boolean"
+        },
+
+        "defaultLevel": {
+          "description": "Specifies the default severity level of the result.",
+          "enum": [ "note", "warning", "error", "open" ]
```
lgolding: The new `ruleConfiguration.defaultLevel` property does have a default value, just as the old `rule.defaultLevel` had. The thing that needs to go away is the default value for the `result.level` property. You can call me if you want to understand why.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -175,11 +180,15 @@ public void Dispose()
             {
                 _issueLogJsonWriter.CloseResults();
 
-                if (_run != null &&
-                    _run.Invocation != null &&
-                    _run.Invocation.StartTime != new DateTime())
+                if (_run?.Invocations != null)
                 {
-                    _run.Invocation.EndTime = DateTime.UtcNow;
+                    foreach (var invocation in _run.Invocations)
```
lgolding: This is a tough one. I don't know what the right answer is, but this isn't it :-)

The reason `invocations` is an array is that some tools perform an analysis by executing a sequence of programs (rather than just one program). Each program in the sequence needs its own invocation. And of course, each has its own start time.

Just populate the _first_start time for now.

src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -662,10 +662,14 @@ public void AnalyzeCommand_EndToEndAnalysisWithExplicitlyDisabledRules()
                 // Three notifications. One for each disabled rule. And an error
                 // notification that all rules have been disabled
                 configurationNotificationCount.Should().Be(4);
-                run.Invocation.ConfigurationNotifications.Where((notification) => notification.Level == NotificationLevel.Error).Count().Should().Be(1);
-                run.Invocation.ConfigurationNotifications.Where((notification) => notification.Level == NotificationLevel.Warning).Count().Should().Be(3);
 
-                run.Invocation.ConfigurationNotifications.Where((notification) => notification.Id == Warnings.Wrn999_RuleExplicitlyDisabled).Count().Should().Be(3);
+                if (run.Invocations != null && run.Invocations.Count > 0)
```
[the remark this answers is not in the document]
EasyRhin0: Fixed

src/Sarif.Driver.UnitTests/TestRuleBase.cs
```diff
@@ -39,13 +39,7 @@ public virtual SupportedPlatform SupportedPlatforms
...
         internal override IDictionary<string, SerializedPropertyInfo> Properties { get; set; }
 
-        public RuleConfiguration Configuration
-        {
-            get
-            {
-                return RuleConfiguration.Enabled;
-            }
-        }
+        public RuleConfiguration Configuration { get { return new RuleConfiguration(); } }
```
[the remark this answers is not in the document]
EasyRhin0: Fixed

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
EasyRhin0: Right, obvious. Fixed.

src/Sarif/CodeGenHints.json
EasyRhin0: Fixed

src/Sarif/Writers/SarifLogger.cs
EasyRhin0: Done

src/Sarif/Schemata/Sarif.schema.json
EasyRhin0: So you want me to do #105 now?

lgolding: The intent of my comment was that you _not_ remove the default from `ruleConfiguration.defaultLevel`. I forgot that removing the default from `result.level` was a separate issue.

Let's _not_ do #105 yet. If it were just a matter of a one-line change in the schema, I'd say go for it -- but it might have test impact that I'd rather deal with separately.

PR #843, opened 2018-04-11
Remove Viewer projects from Everything.sln & build

EasyRhin0:


EasyRhin0:
This is in preparation for moving the Viewer to its own repo.

EasyRhin0:
Still have to remove tests from RunUnitTests.ps1 and the BuildAndTest,cmd

lgolding:
@EasyRhin0 Yes. My point is, there's no reason to edit Everything.sln. Just get rid of it.

EasyRhin0:
Oh sorry, I misread.

EasyRhin0:
It was some kind of weird problem when we were adding .NET Core support, don't remember exactly.

PR #846, opened 2018-04-13
Change SDK project refs to nuget packages

EasyRhin0:
We need to update the Newtonsoft package ref in the SDK to v11 the next time we publish. Having the Viewer reference a newer version causes a very weird and inexplicable exception.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/packages.config
+++ b/src/Sarif.Viewer.VisualStudio/packages.config
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="CommandLineParser" version="2.2.1" targetFramework="net461" />
   <package id="CsvHelper" version="7.1.0" targetFramework="net461" />
   <package id="EnvDTE" version="8.0.2" targetFramework="net461" />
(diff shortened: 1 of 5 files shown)


EasyRhin0:
One note: this project uses the same strong-name key as the SDK. I'm not sure if that is desirable, acceptable, unacceptable, etc.

src/Sarif.Viewer.VisualStudio/packages.config
```diff
@@ -8,6 +9,7 @@
   <package id="EnvDTE90a" version="9.0.2" targetFramework="net461" />
   <package id="Microsoft.ApplicationInsights" version="2.5.1" targetFramework="net461" />
   <package id="Microsoft.Build.Framework" version="15.6.82" targetFramework="net461" />
+  <package id="Microsoft.CodeAnalysis.BinSkim" version="1.4.5" targetFramework="net461" />
```
michaelcfanning: > <package id="Microsoft.CodeAnalysis.BinSkim" v [](start = 1, length = 47)

why is this required? didn't we shut down the binskim analysis?

EasyRhin0: It's a dependency of the Sarif.Sdk package. I'll file an issue for this and another nuget package fix we need.

PR #853, opened 2018-04-22
Add support for embedded hyperlinks in message strings

EasyRhin0:
One minor issue: the regex won't match a link that contains a backslash anywhere in the link text

### DIFF
--- a/src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
+++ b/src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
@@ -86,7 +86,7 @@ namespace Microsoft.CodeAnalysis.Sarif
       ""results"": [
         {
           ""message"": {
-            ""text"": ""Call to Equals() comparing incomparable types[IComparable] and [ClientAttributeValue]""
+            ""text"": ""Call to Equals() comparing incomparable types[IComparable](1) and [ClientAttributeValue](1)""
           },
           ""locations"": [
             {
@@ -107,6 +107,7 @@ namespace Microsoft.CodeAnalysis.Sarif
           ""relatedLocations"": [
             {
               ""physicalLocation"": {
+                ""id"": 1,
                 ""fileLocation"": {
                   ""uri"": ""file:///C:/Windows/Company.NET/Framework/v2.0.50727/mscorlib.dll""
                 },
@@ -115,6 +116,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             },
             {
               ""physicalLocation"": {
+                ""id"": 2,
                 ""fileLocation"": {
                   ""uri"": ""/ClientClient/Company.ResourceManagement.ObjectModel/ClientAttributeValue.cs"",
                   ""uriBaseId"": ""$srcroot""
@@ -128,6 +130,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             },
             {
               ""physicalLocation"": {
+                ""id"": 3,
                 ""fileLocation"": {
                   ""uri"": ""/ClientClient/Company.ResourceManagement.ObjectModel/ClientAttributeValue_ISerializable.cs"",
                   ""uriBaseId"": ""$srcroot""
--- a/src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
+++ b/src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
@@ -86,7 +86,7 @@ namespace Microsoft.CodeAnalysis.Sarif
       ""results"": [
         {
           ""message"": {
-            ""text"": ""Call to Equals() comparing incomparable types[IComparable](1) and [ClientAttributeValue](1)""
+            ""text"": ""Call to Equals() comparing incomparable types[IComparable](1) and [ClientAttributeValue](2)""
           },
           ""locations"": [
             {
--- a/src/Sarif.Converters/SemmleQlConverter.cs
+++ b/src/Sarif.Converters/SemmleQlConverter.cs
@@ -104,7 +104,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                         results.Add(ParseResult(row));
                     }
                 }
-
             }
 
             return results.ToArray();
@@ -151,14 +150,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         {
             // The rawMessage contains embedded related locations. We need to extract the related locations and reformat the rawMessage embedded links wrapped in [brackets].
             // Example rawMessage
-            //     po (coming from [["hbm"|"relative://code/.../brushapi.cxx:176:4882:3"],["hbm"|"relative://code/.../ntgdi.c:1873:50899:3"],["hbm"|"relative://code/.../ntgdi.c:5783:154466:3"]]) may not have been checked for validity before call to vSync.
+            //     po (coming from [["hbm"|"relative://code/.../file1.cxx:176:4882:3"],["hbm"|"relative://code/.../file2.c:1873:50899:3"],["hbm"|"relative://code/.../file2.c:5783:154466:3"]]) may not have been checked for validity before call to vSync.
             // Example normalizedMessage, where 'id' is the related location id to link to
             //   Note: the first link in the message links to the first related location in the list, the second link to the second, etc.
             //     po (coming from [hbm](id)) may not have been checked for validity before call to vSync.
             // Example relatedLocations
-            //     relative://code/.../brushapi.cxx:176:4882:3
-            //     relative://code/.../ntgdi.c:1873:50899:3
-            //     relative://code/.../ntgdi.c:5783:154466:3
+            //     relative://code/.../file1.cxx:176:4882:3
+            //     relative://code/.../file2.c:1873:50899:3
+            //     relative://code/.../file2.c:5783:154466:3
             List<Location> relatedLocations = null;
 
             var sb = new StringBuilder();
@@ -175,13 +174,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 index = rawMessage.IndexOf("]]");
 
                 // embeddedLinksText contains the text for one set of embedded links except for the leading '[[' and trailing ']]'
-                // "hbm"|"relative:/code/.../brushapi.cxx:176:4882:3"],["hbm"|"relative://code/.../ntgdi.c:1873:50899:3"],["hbm"|"relative://code/.../ntgdi.c:5783:154466:3"
+                // "hbm"|"relative:/code/.../file1.cxx:176:4882:3"],["hbm"|"relative://code/.../file2.c:1873:50899:3"],["hbm"|"relative://code/.../file2.c:5783:154466:3"
                 string embeddedLinksText = rawMessage.Substring(0, index - 1);
 
                 // embeddedLinks splits the set of embedded links into invividual links
-                // 1.  "hbm"|"relative://code/.../brushapi.cxx:176:4882:3"
-                // 2.  "hbm"|"relative://code/.../ntgdi.c:1873:50899:3"
-                // 3.  "hbm"|"relative://code/.../ntgdi.c:5783:154466:3"
+                // 1.  "hbm"|"relative://code/.../file1.cxx:176:4882:3"
+                // 2.  "hbm"|"relative://code/.../file2.c:1873:50899:3"
+                // 3.  "hbm"|"relative://code/.../file2.c:5783:154466:3"
 
                 string[] embeddedLinks = embeddedLinksText.Split(new string[] { "],[" }, StringSplitOptions.None);
 
--- a/src/Sarif.Converters/SemmleQlConverter.cs
+++ b/src/Sarif.Converters/SemmleQlConverter.cs
@@ -151,19 +151,20 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         {
             // The rawMessage contains embedded related locations. We need to extract the related locations and reformat the rawMessage embedded links wrapped in [brackets].
             // Example rawMessage
-            //     po (coming from [["hbm"|"relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3"]]) may not have been checked for validity before call to vSync.
-            // Example normalizedMessage
-            //     po (coming from [hbm]) may not have been checked for validity before call to vSync.
+            //     po (coming from [["hbm"|"relative://code/.../brushapi.cxx:176:4882:3"],["hbm"|"relative://code/.../ntgdi.c:1873:50899:3"],["hbm"|"relative://code/.../ntgdi.c:5783:154466:3"]]) may not have been checked for validity before call to vSync.
+            // Example normalizedMessage, where 'id' is the related location id to link to
+            //   Note: the first link in the message links to the first related location in the list, the second link to the second, etc.
+            //     po (coming from [hbm](id)) may not have been checked for validity before call to vSync.
             // Example relatedLocations
-            //     relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3
-            //     relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3
-            //     relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3
+            //     relative://code/.../brushapi.cxx:176:4882:3
+            //     relative://code/.../ntgdi.c:1873:50899:3
+            //     relative://code/.../ntgdi.c:5783:154466:3
             List<Location> relatedLocations = null;
-            normalizedMessage = String.Empty;
 
             var sb = new StringBuilder();
 
             int count = 0;
+            int linkIndex = 0;
             int index = rawMessage.IndexOf("[[");
             while (index > -1)
             {
@@ -174,13 +175,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 index = rawMessage.IndexOf("]]");
 
                 // embeddedLinksText contains the text for one set of embedded links except for the leading '[[' and trailing ']]'
-                // "hbm"|"relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3"
+                // "hbm"|"relative:/code/.../brushapi.cxx:176:4882:3"],["hbm"|"relative://code/.../ntgdi.c:1873:50899:3"],["hbm"|"relative://code/.../ntgdi.c:5783:154466:3"
                 string embeddedLinksText = rawMessage.Substring(0, index - 1);
 
                 // embeddedLinks splits the set of embedded links into invividual links
-                // 1.  "hbm"|"relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3"
-                // 2.  "hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3"
-                // 3.  "hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3"
+                // 1.  "hbm"|"relative://code/.../brushapi.cxx:176:4882:3"
+                // 2.  "hbm"|"relative://code/.../ntgdi.c:1873:50899:3"
+                // 3.  "hbm"|"relative://code/.../ntgdi.c:5783:154466:3"
 
                 string[] embeddedLinks = embeddedLinksText.Split(new string[] { "],[" }, StringSplitOptions.None);
 
@@ -244,8 +245,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 }
 
                 // Re-add the text portion of the link in brackets with the location id in parens, e.g. [link text](id)
-                // Use the id of the first related location
-                sb.Append($"[{embeddedLinksText}]({relatedLocations[relatedLocations.Count - count].PhysicalLocation.Id})");
+                sb.Append($"[{embeddedLinksText}]({relatedLocations[linkIndex++].PhysicalLocation.Id})");
 
                 rawMessage = rawMessage.Substring(index + "]]".Length);
                 index = rawMessage.IndexOf("[[");
--- a/src/Sarif.Converters/SemmleQlConverter.cs
+++ b/src/Sarif.Converters/SemmleQlConverter.cs
@@ -163,6 +163,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             var sb = new StringBuilder();
 
+            int count = 0;
             int index = rawMessage.IndexOf("[[");
             while (index > -1)
             {
@@ -202,6 +203,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                         // "IComparable"|"file://C:/Windows/Microsoft.NET/Framework/v2.0.50727/mscorlib.dll:0:0:0:0"
                         physicalLocation = new PhysicalLocation
                         {
+                            Id = ++count,
                             FileLocation = new FileLocation
                             {
                                 Uri = new Uri($"{locationTokens[0]}:{locationTokens[1]}:{locationTokens[2]}", UriKind.Absolute)
@@ -218,6 +220,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     {
                         physicalLocation = new PhysicalLocation
                         {
+                            Id = ++count,
                             FileLocation = new FileLocation
                             {
                                 Uri = new Uri(locationTokens[1].Substring(1), UriKind.Relative),
@@ -240,8 +243,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     relatedLocations.Add(relatedLocation);
                 }
 
-                // Re-add the text portion of the link in brackets.
-                sb.Append($"[{embeddedLinksText}]"); 
+                // Re-add the text portion of the link in brackets with the location id in parens, e.g. [link text](id)
+                // Use the id of the first related location
+                sb.Append($"[{embeddedLinksText}]({relatedLocations[relatedLocations.Count - count].PhysicalLocation.Id})");
 
                 rawMessage = rawMessage.Substring(index + "]]".Length);
                 index = rawMessage.IndexOf("[[");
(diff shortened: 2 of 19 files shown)


src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
```diff
@@ -86,7 +86,7 @@ public void SemmleQLConvert_EmbeddedLocations()
       ""results"": [
         {
           ""message"": {
-            ""text"": ""Call to Equals() comparing incomparable types[IComparable] and [ClientAttributeValue]""
+            ""text"": ""Call to Equals() comparing incomparable types[IComparable](1) and [ClientAttributeValue](1)""
```
lgolding: The second link should be `(2)`. But... does it make sense to link into a DLL? The user will click on it and then what?

EasyRhin0: Yeah I don't know how to handle this

EasyRhin0: Maybe we need to wait until they add (complete) support for this to the Semmle log format.

lgolding: Well, if that hyperlink is part of the original message in the Semmle output file, then there's nothing you can do. Let me know if it is and I'll mention it to Luke and Michael.

What about that wrong index on the second link?

EasyRhin0: They provide a list of related links. See line 177 in the converter for an example.

Looking at the index now.

lgolding: That example looks wrong:
```
            // The rawMessage contains embedded related locations. We need to extract the related locations and reformat the rawMessage embedded links wrapped in [brackets].
            // Example rawMessage
            //     po (coming from [["hbm"|"relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3"],["hbm"|"relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3"]]) may not have been checked for validity before call to vSync.
            // Example normalizedMessage
            //     po (coming from [hbm]) may not have been checked for validity before call to vSync.
```
1. It doesn't show the link numbers.
2. The original message has three related locations, but the resulting message only shows one.

EasyRhin0: 1. Right, this is what I mean by complete support. I believe Luke said that was coming soon, don't know which version though.

src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -163,6 +163,7 @@ private IList<Location> NormalizeRawMessage(string rawMessage, out string normal
```
lgolding: Not shown in this review, but on Line 162, the assignment to `normalizedMessage` isn't necessary. It's assigned at the very end of the message.

Removing this does not introduce a latent error because if in future we add an "early out" path that avoids that assignment, it won't compile.

src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
EasyRhin0: 2. Yes that's why I was linking to the first location in the list. You are right, I need to change it to use location 2 for the second link.

src/Sarif.Converters/SemmleQlConverter.cs
EasyRhin0: Done

src/Sarif.Converters.UnitTests/SemmleQLConverterTests.cs
EasyRhin0: Fixed

lgolding:
@EasyRhin0 To clarify:

1. Before deciding whether to create a new `physicalLocation`, you have to check both the file location + the region. If the region is different, it's a new physical location. So in fact, the likelihood of finding duplicates is low -- but you do have to check. The `PhysicalLocationEqualityComparer` will do the correct check for you.

2. That check will be an O(N^2) unless you do something clever like keeping a hash table of the physical location objects you've seen so far. Check with @michaelcfanning to see if it's worth doing now. I think it is because SARIF logs can have hundreds of thousands of results.

3. I haven't done the actual "viewer" part of the review yet -- the snapshot, the `BindableTextBlock`, and so forth. So I will have more comments.

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                new Run(" jumps over the lazy dog.")
+            };
+
+            var actual = SdkUiUtilities.GetInlinesForErrorMessage(message);
+
+            actual.Count.Should().Be(expected.Count);
+
+            actual[0].Should().BeOfType(expected[0].GetType());
+            (actual[0] as Run).Text.Should().Be((expected[0] as Run).Text);
+        }
```
lgolding: Loop over all of them, don't just check the first one.

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlowLocation.extensions.cs
```diff
@@ -17,13 +17,16 @@ public static CodeFlowLocationModel ToCodeFlowLocationModel(this CodeFlowLocatio
             {
                 if (codeFlowLocation.Location.PhysicalLocation != null)
```
lgolding: You can DRY out this code by having `CodeFlowLocationExtensions.ToCodeFlowLocationModel` just call `LocationExtensions.ToCodeFlowLocationModel`:
```
public static CodeFlowLocationModel ToCodeFlowLocationModel(this CodeFlowLocation codeFlowLocation)
{
    return codeFlowLocation?.Location != null
        ? codeFlowLocation.Location.ToCodeFlowLocationModel()
        : new CodeFlowLocationModel();
}
```

src/Sarif.Viewer.VisualStudio/Controls/BindableTextBlock.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class BindableTextBlock : TextBlock
+    {
+        public ObservableCollection<Inline> InlineList
+        {
+            get { return (ObservableCollection<Inline>)GetValue(InlineListProperty); }
+            set { SetValue(InlineListProperty, value); }
+        }
+
+        public static readonly DependencyProperty InlineListProperty =
```
lgolding: Our convention: put `static` properties first.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -23,7 +23,6 @@ namespace Microsoft.Sarif.Viewer.ErrorList
     internal class SarifSnapshot : TableEntriesSnapshotBase, IWpfTableEntriesSnapshot
     {
         private readonly List<SarifErrorListItem> _errors;
```
lgolding: While we're in here: This class has a get-only property named `Errors` that is never assigned. It can't be needed, right?

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -1080,5 +1086,82 @@ internal static bool IsShellInCommandLineMode(System.IServiceProvider servicePro
...
+
+        /// <summary>
+        /// Builds a set of Inline elements from the specified message, optionally with embedded hyperlinks.
+        /// </summary>
+        /// <param name="message">The message to process.</param>
+        /// <param name="createHyperlinks">True to generate embedded hyperlinks; otherwise, false.</param>
+        /// <param name="data">A data object to include in each hyperlink's Tag property.</param>
+        /// <param name="clickHandler">A delegate for the Hyperlink.Click event.</param>
+        /// <returns>A collection of Inline elements that represent the specified message.</returns>
+        internal static List<Inline> GetInlinesForErrorMessage(string message, object data, RoutedEventHandler clickHandler)
```
lgolding: Unless you foresee a case where a caller would pass something other than the index, IMO the parameter `object data` should be `int index`. From the point of view of the caller, that's what they're passing. The fact that internally, this message makes use of this data by storing it in an object-valued tag is of no concern to the caller.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -188,26 +163,31 @@ private void ErrorListInlineLink_Click(object sender, RoutedEventArgs e)
 
             if (hyperLink != null)
             {
-                Tuple<int, int> indices = hyperLink.Tag as Tuple<int, int>;
+                Tuple<object, int> data = hyperLink.Tag as Tuple<object, int>;
```
lgolding: To expand on a comment I made elsewhere: Why store this index as an `object`? As the code stands you have to convert the integer index to `object` on the way in, and then call `Convert.ToInt32` on the way out. Why not go with `int` all the way?

lgolding: It's worth a comment that the first element of the tuple is the index of this item in the error list, and the second is the index of the hyperlink target in the result's `RelatedLocations` array.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -188,26 +163,31 @@ private void ErrorListInlineLink_Click(object sender, RoutedEventArgs e)
...
-                    SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
-                    SarifViewerPackage.SarifToolWindow.Control.DataContext = sarifResult;
-                }
-                else
+                if (location != null)
                 {
-                    SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
-                }
+                    // Set the current sarif error in the manager so we track code locations.
+                    CodeAnalysisResultManager.Instance.CurrentSarifError = sarifResult;
```
lgolding: Suggest name change `CurrentSarifResult`.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -188,26 +163,31 @@ private void ErrorListInlineLink_Click(object sender, RoutedEventArgs e)
...
+
+                    if (sarifResult.HasDetails)
+                    {
+                        // Setting the DataContext to be null first forces the TabControl to select the appropriate tab.
+                        SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
+                        SarifViewerPackage.SarifToolWindow.Control.DataContext = sarifResult;
+                    }
+                    else
+                    {
+                        SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
```
lgolding: Since you do this in both branches, hoist it out of the conditional; then you don't need the `else` branch.

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -30,6 +33,9 @@ public static class SdkUiUtilities
         private static readonly Guid s_appIdUsesIsolatedCLR = new System.Guid("{074a44d3-1d9d-406c-9f91-c2a4982b1974}");
         internal static readonly Guid s_enviornmentThemeCategory = new System.Guid("624ed9c3-bdfd-41fa-96c3-7c824ea32e3d");
 
+        // Embedded link format: [link text](n) where n is a positive integer
+        private static string s_embeddedLinkRegexPattern = @"\[(?<link>[^\\\]]+)\]\((?<index>\d+)\)";
```
lgolding: `private const string EmbeddedLinkPattern;`

(I don't think you need "regex" in the name; it's clearly a regex pattern.)

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -30,6 +33,9 @@ public static class SdkUiUtilities
         private static readonly Guid s_appIdUsesIsolatedCLR = new System.Guid("{074a44d3-1d9d-406c-9f91-c2a4982b1974}");
         internal static readonly Guid s_enviornmentThemeCategory = new System.Guid("624ed9c3-bdfd-41fa-96c3-7c824ea32e3d");
 
+        // Embedded link format: [link text](n) where n is a positive integer
```
lgolding: Spec says `n` is non-negative, not positive (that is, it can be 0).

src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -151,19 +151,20 @@ private IList<Location> NormalizeRawMessage(string rawMessage, out string normal
...
+            // Example normalizedMessage, where 'id' is the related location id to link to
+            //   Note: the first link in the message links to the first related location in the list, the second link to the second, etc.
+            //     po (coming from [hbm](id)) may not have been checked for validity before call to vSync.
             // Example relatedLocations
-            //     relative://windows/Core/ntgdi/gre/brushapi.cxx:176:4882:3
-            //     relative://windows/Core/ntgdi/gre/windows/ntgdi.c:1873:50899:3
-            //     relative://windows/Core/ntgdi/gre/windows/ntgdi.c:5783:154466:3
+            //     relative://code/.../brushapi.cxx:176:4882:3
+            //     relative://code/.../ntgdi.c:1873:50899:3
+            //     relative://code/.../ntgdi.c:5783:154466:3
```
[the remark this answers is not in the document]
EasyRhin0: Done

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
EasyRhin0: Oops copy/paste

src/Sarif.Viewer.VisualStudio/Sarif/CodeFlowLocation.extensions.cs
EasyRhin0: Good spotting

src/Sarif.Viewer.VisualStudio/Controls/BindableTextBlock.cs
EasyRhin0: Check

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
```diff
@@ -110,13 +85,13 @@ public override bool TryGetValue(int index, string columnName, out object conten
...
                 {
-                    content = _errors[index].ShortMessage;
+                    content = SdkUiUtilities.UnescapeBrackets(_errors[index].ShortMessage);
                 }
                 else if (columnName == StandardTableKeyNames.FullText)
                 {
                     if (!string.IsNullOrEmpty(_errors[index].Message) && _errors[index].Message.Trim() != _errors[index].ShortMessage.Trim())
                     {
-                        content = _errors[index].Message;
+                        content = SdkUiUtilities.UnescapeBrackets(_errors[index].Message);
```
michaelcfanning: >SdkUiUtilities [](start = 34, length = 14)

NIT casing here should be SdkUIUtilities (see UI vs Ui)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
EasyRhin0: It's set in the ctor

EasyRhin0: OH sorry, misread. Will remove

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
EasyRhin0: I made it an object because the consumer of it will be the caller's click handler, so this way they can stash whatever they need in there. (The Tuple is actually what's stored in the Tag.)
I'm not married to it though...

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -175,6 +176,15 @@ public string FileName
         [Browsable(false)]
         public string Message { get; set; }
 
+        [Browsable(false)]
+        public ObservableCollection<XamlDoc.Inline> SanitizedMessageInlines
```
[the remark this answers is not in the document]
EasyRhin0: I guess I was thinking about the fact that the string has been cleaned. Not needed, will change.

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -175,6 +176,15 @@ public string FileName
         [Browsable(false)]
         public string Message { get; set; }
```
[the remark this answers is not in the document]
EasyRhin0: Yes, it's used in a bunch of other places. Inlines are actual UI controls, while Message is a string.

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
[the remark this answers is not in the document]
EasyRhin0: Agreed. (should have been "ErrorListMessage" anyway)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
EasyRhin0: Done

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
EasyRhin0: Fixed

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
EasyRhin0: Fixed

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSnapshot.cs
EasyRhin0: Fixed. SdkUIUtilitiesTests too

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
EasyRhin0: Changed

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        [Fact]
+        public void SarifSnapshot_GetMessageEmbeddedLinkInlines_ZeroLinksEscapedBrackets()
+        {
+            string message = @"The quick \[brown fox\] jumps over the lazy dog.";
```
lgolding: This renders the first test superfluous.

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                new Run(" jumps over the "),
+                link2
+            };
+
+            var actual = SdkUiUtilities.GetInlinesForErrorMessage(message, data: 1, clickHandler: Hyperlink_Click);
+
+            actual.Count.Should().Be(expected.Count);
+
+            actual[0].Should().BeOfType(expected[0].GetType());
+            (actual[0] as Run).Text.Should().Be((expected[0] as Run).Text);
```
lgolding: Consider: For these two lines, introduce a helper
```
private static void VerifyTextRun(Inline expected, Inline actual)
```
... so this would be
```
VerifyTextRun(actual[0], expected[0]);
```

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            actual[1].Should().BeOfType(expected[1].GetType());
+            (actual[1] as Hyperlink).Inlines.Count.Should().Be((expected[1] as Hyperlink).Inlines.Count);
+            ((actual[1] as Hyperlink).Inlines.FirstInline as Run).Text.Should().Be(((expected[1] as Hyperlink).Inlines.FirstInline as Run).Text);
+            Tuple<object, int> tag1Actual = (actual[1] as Hyperlink).Tag as Tuple<object, int>;
```
lgolding: Likewise, these three lines could be replaced with a helper
```
private static void VerifyHyperlink(Inline expected, Inline actual)
```
and called as
```
VerifyHyperlink(expected[1], actual[1]);
```

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+            var actual = SdkUiUtilities.GetInlinesForErrorMessage(message, data: 1, clickHandler: Hyperlink_Click);
+
+            actual.Count.Should().Be(expected.Count);
+
+            actual[0].Should().BeOfType(expected[0].GetType());
+            (actual[0] as Run).Text.Should().Be((expected[0] as Run).Text);
+
+            actual[1].Should().BeOfType(expected[1].GetType());
+            (actual[1] as Hyperlink).Inlines.Count.Should().Be((expected[1] as Hyperlink).Inlines.Count);
+            ((actual[1] as Hyperlink).Inlines.FirstInline as Run).Text.Should().Be(((expected[1] as Hyperlink).Inlines.FirstInline as Run).Text);
```
lgolding: Likewise, these three lines could be replaced with a helper
```
private static void VerifyHyperlink(Inline expected, Inline actual)
```
and called as
```
VerifyHyperlink(expected[1], actual[1]);
```

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
lgolding: A couple of comments about the regex:

1. If you use the [`IgnorePatternWhitespace` option](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netframework-4.7.1#System_Text_RegularExpressions_RegexOptions_IgnorePatternWhitespace), you can make this more understandable by splitting it up into multiple lines and adding comments:
```
private const string EmbeddedLinkPattern = @"
    \[               # An actual left bracket
    (?<link>
      [^\\\]]+       # One or more of anything that isn't a backslash and isn't a right bracket
    )
    \]               # An actual right bracket
    \(               # An actual left paren
    (?<index>
      \d+
    )
    \)               # An actual right paren
    ";
```

2. Having done that, I can now understand it well enough to ask: Why can't the link text contain a backslash?

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -1080,5 +1086,82 @@ internal static bool IsShellInCommandLineMode(System.IServiceProvider servicePro
...
+
+            return inlines;
+        }
+
+        /// <summary>
+        /// Removes escape backslashes that were used to suppress embedded linking.
+        /// </summary>
+        /// <param name="s">The string to be processed.</param>
+        /// <returns></returns>
+        internal static string UnescapeBrackets(string s)
```
lgolding: As I said in a comment on the tests, I don't think you need to test this alone, so this can be private.

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -1080,5 +1086,82 @@ internal static bool IsShellInCommandLineMode(System.IServiceProvider servicePro
...
+
+                    start = match.Index + match.Length;
+                }
+            }
+
+            if (start < message.Length)
+            {
+                // Add the plain text segment after the last link
+                inlines.Add(new Run(UnescapeBrackets(message.Substring(start))));
+            }
```
lgolding: The logic looks good!

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -1080,5 +1086,82 @@ internal static bool IsShellInCommandLineMode(System.IServiceProvider servicePro
 
             return ((bool)isInCommandLineModeAsObject);
         }
+
+        /// <summary>
+        /// Builds a set of Inline elements from the specified message, without embedded hyperlinks.
+        /// </summary>
+        /// <param name="message">The message to process.</param>
+        /// <returns>A collection of Inline elements that represent the specified message.</returns>
+        internal static List<Inline> GetInlinesForErrorMessage(string message)
```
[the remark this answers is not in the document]
EasyRhin0: This is the one you call if you don't want hyperlinks

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
EasyRhin0: Removed

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
```diff
@@ -0,0 +1,197 @@
...
+            Tuple<object, int> tag1Actual = (actual[1] as Hyperlink).Tag as Tuple<object, int>;
+            Tuple<object, int> tag1Expected = (actual[1] as Hyperlink).Tag as Tuple<object, int>;
+            tag1Actual.Item1.Should().Be(tag1Expected.Item1);
+            tag1Actual.Item2.Should().Be(tag1Expected.Item2);
```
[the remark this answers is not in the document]
EasyRhin0: Yeah I did it this way.

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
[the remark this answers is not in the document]
lgolding: When is that? When you're populating the string-valued `Message` property? The one you told me is still used somewhere? (I'm curious where.)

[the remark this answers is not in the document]
EasyRhin0: In SarifErrorListItem.MessageInlines. It's displayed at the top of the tool window. That's the BindableTextBlock.

src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
EasyRhin0: Removed

lgolding:
@EasyRhin0 Approved! I'm still curious about the regex: why can't the link include a backslash?

EasyRhin0:
I just haven’t been able to figure it out. It works for the syntactically correct pattern, so I figured it was good enough until I have some time to work on it again.

PR #857, opened 2018-04-23
Schema changes 46, 98, 107, 108, 120

EasyRhin0:


src/Sarif/Autogenerated/FileData.cs
```diff
@@ -56,6 +56,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "length", IsRequired = false, EmitDefaultValue = false)]
         public int Length { get; set; }
 
+        /// <summary>
+        /// The role or roles played by the file in the analysis.
+        /// </summary>
+        [DataMember(Name = "roles", IsRequired = false, EmitDefaultValue = false)]
+        public IList<string> Roles { get; set; }
```
lgolding: These should be a `FileRoles` enum, by way of an `EnumHint` in CodeGenHints.json.

src/Sarif/Core/FileData.cs
```diff
@@ -26,7 +26,7 @@ public partial class FileData : ISarifNode
 
             mimeType = mimeType ?? SarifWriters.MimeType.DetermineFromFileExtension(uri);
             fileSystem = fileSystem ?? new FileSystem();
-            encoding = encoding ?? Encoding.UTF8;
+            encoding = encoding ?? System.Text.Encoding.UTF8;
```
lgolding: There is no default value any more.

EasyRhin0: Fixed

src/Sarif/Autogenerated/FileData.cs
EasyRhin0: Fixed

src/Sarif/Core/FileData.cs
```diff
@@ -92,7 +92,7 @@ private static string EncodeFileContents(IFileSystem fileSystem, string filePath
 
             if (mimeType != SarifWriters.MimeType.Binary)
             {
-                fileContents = Encoding.UTF8.GetBytes(fileSystem.ReadAllText(filePath, inputFileEncoding));
+                fileContents = System.Text.Encoding.UTF8.GetBytes(fileSystem.ReadAllText(filePath, inputFileEncoding));
```
lgolding: There are three bugs here:
1. Now that there's no default, you can't get the file contents as text if you don't know the `inputFileEncoding`. In that case, the best you can to is populate `fileContents.binary` by reading the raw bytes.
2. On line 102, you always Base64-encode the resulting string. But -- and this is a change in the spec that you wouldn't be aware of -- the intention of `fileContents.text` is to be human-readable text, so the spec now says *not* to Base64 encode the contents of `fileContents.text.
3. If you don't have an `encoding` on the `file` object, fall back to `run.defaultFileEncoding`.

So the procedure is:
```
get the encoding from the file or from the run (they might both be absent).
read the bytes from the file
if the file is MIME binary OR you don't have an encoding OR the encoding isn't one that .NET supports
    base64 encode the bytes
   put the resulting string in the Binary property
else
   decode the bytes into a string using the encoding
   put the resulting string in the Text property
```

src/Sarif/Autogenerated/FileData.cs
```diff
@@ -56,6 +56,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "length", IsRequired = false, EmitDefaultValue = false)]
         public int Length { get; set; }
 
+        /// <summary>
+        /// The role or roles played by the file in the analysis.
+        /// </summary>
+        [DataMember(Name = "roles", IsRequired = false, EmitDefaultValue = false)]
+        public FileRole Roles { get; set; }
```
lgolding: Sorry I wasn't clear. `file.roles` is an array, each of whose elements is a member of the `FileRoles` enum. So it should generate `public IList<FileRole> Roles; { get; set; }`. Did you change the schema to make this singular?

src/Sarif/Autogenerated/FileRole.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Values specifying the roles played by the file in the analysis.
+    /// </summary>
+    [Flags]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.49.0.0")]
+    public enum FileRole
+    {
+        Default,
```
lgolding: There is no "default" role. Modify the code hint to name the "0" enum value "`None`". I know there's a knob for that.

lgolding: Ah, there it is, it's `"zeroValueName"`.

EasyRhin0: Fixed

src/Sarif/Autogenerated/FileData.cs
EasyRhin0: Oh I thought it was a Flags enum

lgolding: I'm sorry, you're right!

lgolding: We need tests for serializing/deserializing this property. I suppose just four tests: (serializing, deserializing) x (one enum value, multiple enum values) = 4.

src/Sarif/Core/FileData.cs
EasyRhin0: How can I get the fileData.encoding when I'm in the process of creating the fileData in the first place?

lgolding: The caller of `FileData.Create` passes `encoding` as an argument. That gets passed to `EncodeFileContents` in the `inputFileEncoding` parameter. The only trick is that the caller now needs to pass `defaultFileEncoding` (which it should have gotten from the `Run` object) to the `Create` method as well.

EasyRhin0: OK currently I don't see any callers providing an encoding param except one unit test. The only call where I see the opportunity to get the Run.DefaultFileEncoding is in SarifLogger.CaptureFile.

src/Sarif/Autogenerated/FileData.cs
EasyRhin0: It looks like there is a problem with the way flags enums are generated. Because the members are not given explicit values, they get the sequential defaults {0, 1, 2, 3, 4, 5 ...} instead of the bitfield power of 2 values.

lgolding: As we discussed: merge as is (with two tests ignored), fix JSchema code gen, update SDK to new JSchema, re-enable tests.

src/Sarif/Core/FileData.cs
EasyRhin0: So now we are returning a string from EncodeFileContents that the caller puts into either Text or Binary based on conditions within EncodeFileContents. Since the caller has no way of knowing those conditions, I was thinking of changing EncodeFileContents to non-static "SetFileContents" and creating the FileContent property in there. Thoughts?

EasyRhin0: Or static EncodeFileContents could return a FileContent instance

EasyRhin0: Definitely the second one...

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -72,7 +73,7 @@ public void FileData_PersistFileContentsBinary()
                 fileData.Hashes.Should().BeNull();
 
                 string encodedFileContents = Convert.ToBase64String(File.ReadAllBytes(filePath));
-                fileData.Contents.Text.Should().Be(encodedFileContents);
+                fileData.Contents.Binary.Should().Be(encodedFileContents);
```
lgolding: `fileData.Contents.Text.Should().BeNull();`

EasyRhin0: Ah yup

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -356,7 +367,16 @@ private void CaptureFile(Uri uri)
...
 
-            _run.Files[fileDataKey] = FileData.Create(uri, _loggingOptions);
+
+            Encoding encoding = null;
+
+            try
+            {
+                encoding = Encoding.GetEncoding(_run.DefaultFileEncoding);
+            }
+            catch (ArgumentException) { }
```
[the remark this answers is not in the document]
lgolding: @EasyRhin0 Also, what does `Encoding.GetEncoding` do it the argument is `null`? If if throws `ArgumentNullException`, that's fine, because `ArgumentException` will catch it.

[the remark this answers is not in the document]
EasyRhin0: It does throw ArgumentNullException for any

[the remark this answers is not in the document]
EasyRhin0: Test added

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -186,6 +186,51 @@ public void SarifLogger_WritesFileData()
             sarifLog.Runs[0].Files[fileDataKey].Hashes[2].Value.Should().Be("0953D7B3ADA7FED683680D2107EE517A9DBEC2D0AF7594A91F058D104B7A2AEB");
         }
 
+        [Fact]
+        public void SarifLogger_WritesFileDataWithUnrecognizedEncoding()
+        {
+            var sb = new StringBuilder();
+            string file;
```
lgolding: Declare at point of use.

EasyRhin0: It's used outside the using block, down on line 225.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -186,6 +186,51 @@ public void SarifLogger_WritesFileData()
...
+                }
+            }
+
+            string logText = sb.ToString();
+
+            string fileDataKey = new Uri(file).AbsoluteUri;
+            byte[] fileBytes = Encoding.Default.GetBytes(fileText);
+
+            var sarifLog = JsonConvert.DeserializeObject<SarifLog>(logText);
+            sarifLog.Runs[0].Files[fileDataKey].MimeType.Should().Be(MimeType.CSharp);
```
lgolding: You could DRY this out with
```
FileData fileData = sarifLog.Runs[0].Files[fileDataKey];
```

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
lgolding: I missed that.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
EasyRhin0: Done

PR #862, opened 2018-04-26
Transform v1 -> v2 run.logicalLocations

EasyRhin0:


src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -106,5 +106,64 @@ public void SarifTransformerTests_ToCurrent_MinimumWithTwoRuns()
...
+            SarifLogVersionOne v1Log = GetSarifLogVersionOne(v1LogText);
+            var transformer = new SarifVersionOneToCurrentVisitor();
+            transformer.VisitSarifLogVersionOne(v1Log);
+
+            SarifLog v2Log = transformer.SarifLog;
+
+            v2Log.Runs.Should().NotBeNull();
+            v2Log.Runs.Count.Should().Be(1);
+
+            v2Log.Runs[0].LogicalLocations.Should().NotBeNull();
```
lgolding: Could be shorter with temp variable `IList<LogicalLocation> logicalLocations = v2Log.Runs[0].LogicalLocations;`

PR #863, opened 2018-04-26
Sdk sample app

EasyRhin0:


src/Sarif.Sdk.Sample.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
lgolding: Consider placing this whole thing into a `samples` directory. Since we are likely to have more than one sample, consider placing this one into a subdirectory. That would require you to come up with a name more specific than `Sdk.Sample`, which I think would be a good thing.

Having said all that, if your goal here is just to get this checked in ASAP, then at least put it as-is under a `src\samples` directory -- we can rename the sample, and subdivide the directory, later, if needed.

lgolding: Ok, looking at what the sample does, how about `Sarif.Sdk.ReadWriteSample`, in directory `src\samples\readWrite`?

src/Sarif.Sdk.Sample/Program.cs
```diff
@@ -0,0 +1,383 @@
...
+                    errors => 1);
+
+            return result;
+        }
+
+        /// <summary>
+        /// Loads a SARIF log file from disk and deserializes it into a code object.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns></returns>
```
lgolding: If you're not going to populate an XML comment element, just omit it. IMO.

src/Sarif.Sdk.Sample/Program.cs
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        /// <summary>
+        /// Constructs a sample SARIF log and writes it to the specified location.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        static int CreateSarifLogFile(CreateOptions options)
+        {
+            // We'll use this source file for several defect results -- the SampleSourceFiles folder should be at the same level as the project folder
```
lgolding: Nit: line too long (150 columns).

src/Sarif.Sdk.Sample.sln
EasyRhin0: Done

src/Sarif.Sdk.Sample/Program.cs
EasyRhin0: Fixed

src/Sarif.Sdk.Sample/Program.cs
EasyRhin0: Fixed

PR #864, opened 2018-04-26
Transform v1 -> v2 run.files

EasyRhin0:


src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -168,5 +170,93 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
...
+                        ""name"": ""CodeScanner"",
+                        ""semanticVersion"": ""2.1.0""
+                      },
+                      ""results"": [
+                      ]
+                    }
+                  ]
+                }";
+
+            SarifLogVersionOne v1Log = GetSarifLogVersionOne(v1LogText);
```
lgolding: I suggest you DRY this out as follows:
```
SarifLog v2Log = TransformVersionOneToCurrent(v1LogText);

...

private static SarifLog TransformVersionOneToCurrent(string v1LogText)
{
    SarifLogVersionOne v1Log = GetSarifLogVersionOne(v1LogText);
    var transformer = new SarifVersionOneToCurrentVisitor();
    transformer.VisitSarifLogVersionOne(v1Log);

    return transformer.SarifLog;
}


```

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -168,5 +170,93 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
...
+            var transformer = new SarifVersionOneToCurrentVisitor();
+            transformer.VisitSarifLogVersionOne(v1Log);
+
+            SarifLog v2Log = transformer.SarifLog;
+
+            v2Log.Runs.Should().NotBeNull();
+            v2Log.Runs.Count.Should().Be(1);
+
+            Dictionary<string, FileData> files = v2Log.Runs[0].Files as Dictionary<string, FileData>;
```
lgolding: Cast is not needed. `Files` is already `IDictionary<string, FileData>`. Your test code shouldn't need a concrete `Dictionary`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -26,6 +27,59 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne no
...
+                if (node.Uri != null)
+                {
+                    fileData.FileLocation = new FileLocation
+                    {
+                        Uri = node.Uri,
+                        UriBaseId = node.UriBaseId
+                    };
+                }
+
+                if (!string.IsNullOrWhiteSpace(node.Contents))
```
lgolding: At first glance, I'd ask "why shouldn't I copy the contents even if they're all whitespace?". Perhaps the answer is that all-whitespace is never a valid MIME Base64 string.

Check on that. If all-whitespace _is_ valid, change this to a simple `null` check. If not, add a comment explaining why all-whitespace is bad.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -26,6 +27,59 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne no
...
+                {
+                    fileData.Contents = new FileContent
+                    {
+                        Binary = node.Contents
```
lgolding: It's true that a v1 log file doesn't explicitly tell you if the file is text or binary, so "always populate `Binary`" might be a reasonable approach. But you do have the MIME type. So perhaps you could populate `Text` as well for a selection of common textual MIME types. The v1 spec said that `File.Contents` contains the MIME-Base64 encoding of the UTF-8 file contents (if the file is in another encoding, the producer had to transcode it to UTF-8 and then MIME-Base64 encode it).

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -26,6 +27,59 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne no
...
+                {
+                    fileData.Hashes = new List<Hash>();
+
+                    foreach (HashVersionOne hash in node.Hashes)
+                    {
+                        fileData.Hashes.Add(TransformHashVersionOne(hash));
+                    }
+                }
+
+                fileData.Tags.UnionWith(node.Tags);
```
lgolding: Why `UnionWith`? `Tags` is guaranteed empty at this point. Why not just assign?

I had this question on a usage of `UnionWith` in a previous review, but I forgot to comment on it.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: It return empty string

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Tags is get-only

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Do you mean that the MIME Base64 encoding of any all-whitespace string is the empty string?

EasyRhin0: That's correct

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Oh!

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -43,6 +97,22 @@ public LogicalLocation TransformLogicalLocationVersionOne(LogicalLocationVersion
...
 
+        public Hash TransformHashVersionOne(HashVersionOne node)
+        {
+            Hash hash = null;
+
+            if (node != null)
+            {
+                hash = new Hash
+                {
+                    Algorithm = (AlgorithmKind)Enum.Parse(typeof(AlgorithmKind), node.Algorithm.ToString()),
```
lgolding: 1. In v2, `hash.algorithm` is a `string`. I think you didn't remove the `EnumHint` from `Hash.Algorithm`. Once you've done that, it's just `Algorithm = node.Algorithm.ToString();`
2. **BUT**! You can't just copy the string, because the spec says that if your algorithm is one of those that appears in the IANA registry of [Hash function textual names](https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xhtml), then you have to spell it like IANA spells it. For example, we have `AlgorithmKind.Sha256`, but IANA spells it `"sha-256"` (note the hyphen). Don't worry, there are only 7 hash functions in the IANA registry.

3. Although the spec says that `hash.algorithm` is case-insensitive, I suggest you pull the values to lower case for consistency with IANA.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Ah, the heck with it. v1 is just less rich than v2. Let me ping @michaelcfanning to see if he thinks this is worth it.

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -168,5 +170,93 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
...
+                      },
+                      ""tool"": {
+                        ""name"": ""CodeScanner"",
+                        ""semanticVersion"": ""2.1.0""
+                      },
+                      ""results"": [
+                      ]
+                    }
+                  ]
+                }";
```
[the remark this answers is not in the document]

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -102,9 +182,9 @@ public override ToolVersionOne VisitToolVersionOne(ToolVersionOne node)
                     Version = node.Version
                 };
 
-                foreach (string propertyName in node.PropertyNames)
+                if (tool.Properties == null)
                 {
-                    tool.PropertyNames.Add(propertyName);
+                    tool.Properties = new Dictionary<string, SerializedPropertyInfo>();
```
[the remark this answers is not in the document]
EasyRhin0: Already checked

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: I had the nagging sense that this wasn't an enum anymore...

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -168,5 +170,93 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
             logicalLocations["collections"].Kind.Should().Be("namespace");
             logicalLocations["collections"].ParentKey.Should().BeNull();
         }
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_OneRunWithFiles()
+        {
+            string content = "The quick brown fox jumps over the lazy dog";
+            byte[] bytes = Encoding.ASCII.GetBytes(content);
```
lgolding: The spec says to use UTF-8. It doesn't matter here, but your resident SARIF experts ;-) will trip over the use of `ASCII`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Done, not sure if you'll like the way I did it though...

src/Sarif.UnitTests/SarifTransformerTests.cs
EasyRhin0: OK

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -168,5 +170,93 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
...
+
+            FileData fileData = null;
+
+            fileData = files["file:///home/list.txt"];
+            fileData.MimeType.Should().Be("text/plain");
+            fileData.Length.Should().Be(43);
+            fileData.FileLocation.Should().BeNull();
+            fileData.Contents.Should().NotBeNull();
+            fileData.Contents.Binary.Should().Be("VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw==");
+            Encoding.ASCII.GetString(Convert.FromBase64String(fileData.Contents.Binary)).Should().Be(content);
```
lgolding: I don't see the point of this test. You synthesized that Base64 string from your original text, so of course it will come back the same. Really all you're testing is that you synthesized the string correctly. But why does it matter? All that matters is that the string in the v2 file matches the string in the v1 file.

EasyRhin0: Yeah dumb

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -194,7 +189,10 @@ public void SarifTransformerTests_ToCurrent_OneRunWithFiles()
                           ]
                         },
                         ""file:///home/buildAgent/bin/app.zip"": {
-                          ""mimeType"": ""application/zip""
+                          ""mimeType"": ""application/zip"",
+                          ""properties"": {
+                            ""my_key"": ""some value""
+                          }
```
lgolding: But now you need to test that the property bag was copied.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: I see you made the change but you didn't change the spellings. It needs to be `"sha-1"`, "`sha-256"` to match the IANA registry (says the spec).

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -44,9 +44,9 @@ public void FileData_ComputeHashes()
                 {
                     switch (hash.Algorithm)
                     {
-                        case AlgorithmKind.MD5: { hash.Value.Should().Be(hashes.MD5); break; }
-                        case AlgorithmKind.Sha1: { hash.Value.Should().Be(hashes.Sha1); break; }
-                        case AlgorithmKind.Sha256: { hash.Value.Should().Be(hashes.Sha256); break; }
+                        case "md5": { hash.Value.Should().Be(hashes.MD5); break; }
+                        case "sha1": { hash.Value.Should().Be(hashes.Sha1); break; }
+                        case "sha256": { hash.Value.Should().Be(hashes.Sha256); break; }
```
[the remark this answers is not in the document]
EasyRhin0: Fixed

src/Sarif.UnitTests/SarifTransformerTests.cs
EasyRhin0: Lines 243-44?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -11,16 +11,57 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
...
+        private static HashSet<string> s_TextMimeTypes = new HashSet<string>()
+        {
+            "application/ecmascript",
+            "application/java-archive",
+            "application/javascript",
+            "application/java-serialized-object",
+            "application/java-vm",
+            "application/json",
+            "application/onenote",
+            "application/pdf",
```
lgolding: Some of these are not text format. Just keep the ones that we know are text:
- ecmascript
- javascript
- rss+xml
- rtf
- typescript
- x-csh
- xhtml+xml
- xml
- x-sh

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -11,16 +11,57 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
...
-                                                  "text/x-c",
-                                                  "text/x-h",
-                                                  "text/x-csharp",
-                                                  "text/x-java-source",
-                                                  "text/javascript",
-                                                  "text/ecmascript",
-                                                  "text/html",
-                                                  "text/xml" };
-            
+        #region Text MIM# types
```
lgolding: Typo
```
#region Text MIME types
```

EasyRhin0: Oops :p

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

lgolding: @EasyRhin0 I'm so sorry! `application\json` is text -- given our context, I don't know how I missed that!

PR #869, opened 2018-04-29
Transform v1 -> v2 run.invocations

EasyRhin0:


src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -13,14 +13,15 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public class SarifTransformerTests
     {
-        private static SarifLogVersionOne GetSarifLogVersionOne(string logText)
+        private static JsonSerializerSettings s_jsonSettings = new JsonSerializerSettings
```
lgolding: `readonly`

EasyRhin0: Fixed

src/Sarif.UnitTests/SarifTransformerTests.cs
```diff
@@ -365,5 +366,84 @@ public void SarifTransformerTests_ToCurrent_OneRunWitRules()
...
+        {
+            string v1LogText =
+              @"{
+                  ""version"": ""1.0.0"",
+                  ""runs"": [
+                    {
+                      ""invocation"": {
+                        ""commandLine"": ""CodeScanner @collections.rsp"",
+                        ""responseFiles"": {
+                          ""collections.rsp"": ""-input src/collections/*.cpp -log out/collections.sarif -rules all -disable C9999""
```
lgolding: This isn't right: `responseFiles` is now an array of `fileLocation` objects, each of which specifies the location of a response file. To embed the response file contents, you make a corresponding entry in the `run.files` dictionary, and populate `file.contents.text`.

I'm surprise this even works... oh, I see what you did! You abused `uriBaseId` to hold the contents. Umm, no :-) :-)

lgolding: (Actually, although my diagnosis turns out to be right, my comment above was premature. I forgot that this is a v1 file :-))

lgolding:
```
{                        # A run object
  "files": {
    "collections.rsp": {
      "contents": {
        "text": "-input src/collections/*.cpp -log out/collections.sarif -rules all -disable C9999"
      }
    }
  }

  "invocations": [
    {
      "responseFiles": [
        {
          "uri": "collections.rsp"
        }
      ]
    }
  ]
}

Comment (EasyRhin0) [reply src/Sarif.UnitTests/SarifTransformerTests.cs]: Well this json is the v1 version which I took from Comprehensive.sarif. In the transfer coe, I did use UriBaseId for the contents based on my interpretation of this statement in 3.13.4:
A SARIF producer MAY embed the contents of a response file in the SARIF log file by mentioning the response file in run.files (§3.11.11), and providing a value for the file.contents property (§3.16.8).
So I have apparently misinterpreted that!

Comment (lgolding) [reply src/Sarif.UnitTests/SarifTransformerTests.cs]: The "MAY" means that you don't have to embed the contents at all. But if you do, the go in `file.contents.text` as in my example. No worries.

Comment (EasyRhin0) [reply src/Sarif.UnitTests/SarifTransformerTests.cs]: Gotcha

File: src/Sarif.UnitTests/SarifTransformerTests.cs (hunk 1 above)
Comment (lgolding) [reply src/Sarif.UnitTests/SarifTransformerTests.cs]: I didn't mean to say it shouldn't be `static`, I meant it should also be `readonly` (`static readonly`).

File: src/Sarif.UnitTests/SarifTransformerTests.cs [hunk 3]
```diff
@@ -365,5 +366,94 @@ public void SarifTransformerTests_ToCurrent_OneRunWitRules()
...
+
+            Run run = v2Log.Runs[0];
+
+            run.Invocations.Should().NotBeNull();
+            run.Invocations.Count.Should().Be(1);
+
+            Invocation invocation = run.Invocations[0];
+
+            invocation.CommandLine.Should().Be("CodeScanner @collections.rsp");
+            invocation.Arguments.Should().NotBeNull();
```
lgolding: @EasyRhin0 @michaelcfanning I don't think our converter should populate `invocation.arguments` for now. The reason `arguments` exists _in addition to_ `commandLine` is so consumers don't have to parse the command line, which requires knowing and implementing the quoting and escaping rules on the SARIF producer's platform. It's alot more complicated than just breaking on white space. Unless the converter
1. knows the platform on which the producer ran, and
2. implements a command line parser for that platform

it can't reliably populate `arguments`.
SARIF doesn't define an "originatingPlatform" property, but a converter could get that information out of band, for example on its command line. So in future our converter could be enhanced with parsing for popular platforms and shells. But IMO we shouldn't take that on now.

src/Sarif.UnitTests/SarifTransformerTests.cs
EasyRhin0: oh OK :)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -295,25 +388,18 @@ public override RunVersionOne VisitRunVersionOne(RunVersionOne node)
...
-
-                SarifLog.Runs.Add(run);
-
-                VisitToolVersionOne(node.Tool);
             }
 
             return null;
         }
 
-        public override ResultVersionOne VisitResultVersionOne(ResultVersionOne node)
```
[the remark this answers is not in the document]
EasyRhin0: Yeah and I'm finding the TransferXxxVersionOne model better anyway

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -295,25 +388,18 @@ public override RunVersionOne VisitRunVersionOne(RunVersionOne node)
...
+        public Tool TransformToolVersionOne(ToolVersionOne node)
```
lgolding: @EasyRhin0 @michaelcfanning Ok, I see what you're doing here. Rather than overriding the visitor-provided tool-visiting method, you provide your own, which returns the type you need (`Tool`) rather than `ToolVersionOne`. And I see you did that in your `TransformInvocationOne` method above, as well. This all makes sense.

From these two examples, we gain an insight: we don't need to base our code on the auto-generated v1 visitor at all! It  would be a small change to your existing code to create a class that accepts a v1 OM, and walks down it using methods that return the appropriate v2 objects.

Remember, the reason we code-gen'd the visitor classes was so that the visitors would always be up to date when the schema changed. But the v1 schema is frozen! So we can hard-code our converter class to the existing, frozen schema.

Using the code-gen'd visitor was a good way to get started, but now that you've demonstrated a better way to code this, we should go all in on it.

By the way, remember at the very beginning when you recognized that the code-gen'd visitor returned a v1 log object -- and so you had to hack around it by synthesizing a v2 log object in a property, and then pulling out the property at the end? This small redesign would avoid having to do that.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
...
                     Architecture = node.Architecture,
-                    BaselineId = node.BaselineId,
                     AutomationId = node.AutomationId,
+                    BaselineId = node.BaselineId,
                     Id = node.Id,
                     Properties = node.Properties,
                     Results = new List<Result>(),
-                    StableId = node.StableId
+                    StableId = node.StableId,
+                    Tool = TransformToolVersionOne(node.Tool)
```
lgolding: @EasyRhin0 @michaelcfanning Much better! See my mega-comment below about what this change really means.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
             return rule;
         }
 
+        private List<FileLocation> CreateFileLocationsFromDictionary(IDictionary<string, string> dictionary, Run run)
+        {
+            List<FileLocation> result = null;
+
+            if (dictionary != null)
+            {
+                result = new List<FileLocation>();
```
lgolding: Name this `fileLocations` -- a better name in general, since "result" is so generic, but especially so in this context, where `result` has a very different meaning!

EasyRhin0: Yep meant to do that...

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Perfect

EasyRhin0: THIS will be the next PR :)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
...
+
+                    if (run != null && !string.IsNullOrWhiteSpace(dictionary[key]))
+                    {
+                        // We have contents, so mention this file in run.files
+                        if (run.Files == null)
+                        {
+                            run.Files = new Dictionary<string, FileData>();
+                        }
+
+                        if (!run.Files.ContainsKey(key))
```
lgolding: Should be:
```
if (!run.Files.Contains(key))
{
    run.Files.Add(key, new FileData());
}

File responseFile = run.Files[key];  // Now guaranteed to exist.
responseFile.Contents = new FileContent
{
    Text = dictionary[key];
};
```

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
             return rule;
         }
 
+        private List<FileLocation> CreateFileLocationsFromDictionary(IDictionary<string, string> dictionary, Run run)
```
lgolding: Now that I understand what this method is doing, I think a better name is `TransformResponseFiles`, and the first parameter might be named `responseFileToContentsDictionary`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
...
+                    EnvironmentVariables = node.EnvironmentVariables,
+                    Machine = node.Machine,
+                    ProcessId = node.ProcessId,
+                    Properties = node.Properties,
+                    ResponseFiles = CreateFileLocationsFromDictionary(node.ResponseFiles, SarifLog.Runs.Last()),
+                    StartTime = node.StartTime,
+                    WorkingDirectory = node.WorkingDirectory
+                };
+
+                if (!string.IsNullOrWhiteSpace(node.FileName))
```
lgolding: Remove this test. `invocation.commandLine` is allowed to exist even if `invocation.fileName` does not.

Those properties exist for two different reasons. `invocation.fileName`  (which is the improved name `executableLocation` in v2) is meant to allow the consumer to know exactly where the tool was executed from (like, it might help you discover that you'd executed an old version from some directory earlier in `PATH`). It doesn't have to be there, so don't couple these together.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
...
+                {
+                    if (!string.IsNullOrWhiteSpace(node.CommandLine))
+                    {
+                        string exeName = Path.GetFileName(node.FileName);
+                        int startIndex = node.CommandLine.IndexOf(exeName) + exeName.Length;
```
lgolding: This is wrong. `fileName` might be `"C:\Program Files (x86)\CodeScanner\scanner.exe"`, but `commandLine` might be "`scanner *.cs`" (if `scanner.exe` were on your path). That's why (as I mentioned above) `fileName` exists at all.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -243,21 +243,106 @@ public Rule TransformRuleVersionOne(RuleVersionOne node)
...
+                        string[] args = node.CommandLine.Substring(startIndex).Split(new char[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
lgolding: And finally, as we discussed: don't even _try_ to parse this :-)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Ok ok, I was thinking we wouldn't do anything if it already existed. Changed.

lgolding: It's unlikely, but it might exist with some other properties set. Another thing to consider: With this change, I can see at a glance that the code is correct. Before the change, I have to convince myself that the `file` object _could not have_ existed before. It might be true, but with the changed code, I don't have to think about it.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Sorry, I misspoke. I meant "don't condition your parsing of `commandLine` on whether this _other_ property `fileName` is present". But that turned out to be moot, because you weren't going to parse the command line itself.

You _do_ have to test `fileName` before trying to make a `Uri` out of it. Sorry again.

PR #870, opened 2018-04-29
Result level open

michaelcfanning:
@lgolding @EasyRhin0 

Add tests for ShouldLog behavior. Also add tests that pushes all ResultLevel enum value to this helper so that it doesn't break in future if we add a new enum member.

### DIFF
--- a/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
+++ b/src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
@@ -107,7 +107,23 @@ namespace Microsoft.CodeAnalysis.Sarif
 
 
         [Theory]
-        [InlineData(ResultLevel.Error, true, false)]
+        // These values are emitted both verbose and non-verbose
+        [InlineData(ResultLevel.Error, true, true)]
+        [InlineData(ResultLevel.Error, false, true)]
+        [InlineData(ResultLevel.Warning, true, true)]
+        [InlineData(ResultLevel.Warning, false, true)]
+        [InlineData(ResultLevel.Default, true, true)]
+        [InlineData(ResultLevel.Default, false, true)]
+
+        // These result levels only emitted in verbose logging mode
+        [InlineData(ResultLevel.NotApplicable, true, true)]
+        [InlineData(ResultLevel.NotApplicable, false, false)]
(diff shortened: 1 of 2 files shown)


src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -104,6 +104,54 @@ public void SarifLogger_RedactedCommandLine()
...
+            LoggingOptions loggingOptions = LoggingOptions.Verbose;
+            var sb = new StringBuilder();
+            var logger = new SarifLogger(new StringWriter(sb), loggingOptions);
+
+            foreach (object resultLevelObject in Enum.GetValues(typeof(ResultLevel)))
+            {
+                // The point of this test is that every defined enum value
+                // should pass a call to ShouldLog and will not raise an 
+                // exception because the enum value isn't recognized
+                logger.ShouldLog((ResultLevel)resultLevelObject);
```
lgolding: I _love_ tests like this! Good comment, too. And good test name.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -104,6 +104,54 @@ public void SarifLogger_RedactedCommandLine()
...
+        }
+
+        [Fact]
+        public void SarifLogger_ShouldLogRecognizesAllResultLevels()
+        {
+            LoggingOptions loggingOptions = LoggingOptions.Verbose;
+            var sb = new StringBuilder();
+            var logger = new SarifLogger(new StringWriter(sb), loggingOptions);
+
+            foreach (object resultLevelObject in Enum.GetValues(typeof(ResultLevel)))
```
lgolding: You could avoid the cast below using Linq like this:
```
foreach (ResultLevel resultLevel in Enum.GetValues(typeof(ResultLevel)).Cast<ResultLevel>())
```

michaelcfanning: Sure, at the expense of a longer line of code and use of an obscure extension method. I'm not personally sure that this sort of incremental elaboration of C# and its framework is always the right way to proceed.

PR #876, opened 2018-05-01
Transform v2 -> v1 minimum + logicalLocations + tests

EasyRhin0:


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            return JsonConvert.DeserializeObject<SarifLog>(logText, s_v2JsonSettings);
+        }
+
+        private static SarifLogVersionOne TransformCurrentToVersionOne(string v2LogText)
+        {
+            SarifLog v2Log = GetSarifLog(v2LogText);
+            var transformer = new SarifCurrentToVersionOneVisitor();
+            transformer.VisitSarifLog(v2Log);
+
+            return SarifCurrentToVersionOneVisitor.SarifLogVersionOne;
```
lgolding: I should have mentioned this in the last PR. The `SarifLogVersionOne` property is the one thing you _don't_ want to be static. Consider a multithreaded application that creates two transformers on two threads. The application tells one of the transformers to transform v2LogOne, and the other to transform v2LogTwo. Both transformer objects use the same set of static methods to build up two _different_ output objects.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""results"": []
+    }
+  ]
+}";
+
+            v1LogText.Should().Be(v1LogExpectedText);
+        }
```
lgolding: DRY this out with a helper method:
```
private static void VerifyCurrentToVersionOneTransformation(v2LogText, v1LogExpectedText)
{
    SarifLogVersionOne v1Log = TransformCurrentToVersionOne(v2LogText);
    string v1LogText = JsonConvert.SerializeObject(v1Log, s_v1JsonSettings);
    v1LogText.Should().Be(v1LogExpectedText);
}
```

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Yeah understood. The trouble there is that I have two spots that refer to it directly, so I'll have to refactor.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Done

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public static class SarifUtilitiesVersionOne
+    {
+        private const string V1_0_0 = "1.0.0";
+        private const string V1_0_0_BETA_5 = "1.0.0-beta.5";
```
lgolding: This will never be used. You'll never want to downgrad a SARIF v2 file to `1.0.0-beta.5`, and we don't claim to be able to upgrade `1.0.0-beta.5` files to v2.

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public static string ConvertToText(this SarifVersionVersionOne sarifVersion)
+        {
+            switch (sarifVersion)
+            {
+                case SarifVersionVersionOne.OneZeroZeroBetaFive: { return V1_0_0_BETA_5; }
```
lgolding: I would throw `ArgumentException` if you are asked to convert anything except `V1_0_0`.

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
EasyRhin0: Ah yep

EasyRhin0: Done

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
EasyRhin0: Done

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: So I'm not sure how to make this work: in CreateInvocation, I'm passing the last run we created to CreateResponseFilesList in case there are responseFiles that need to be added to the run's files collection. I'm not coming up with any slick ideas on how to get at that run instance, short of passing it all the way through as a parameter. Thoughts?

EasyRhin0:
Ah crap, don't look yet.

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private const string V1_0_0 = "1.0.0";
+
+        public static string ConvertToText(this SarifVersionVersionOne sarifVersion)
+        {
+            switch (sarifVersion)
+            {
+                case SarifVersionVersionOne.OneZeroZero: { return V1_0_0; }
+            }
+            throw new ArgumentException("Unsupported version", nameof(sarifVersion));
```
lgolding: I'd put this as the `default` case of the `switch`, to demonstrate that you've covered all cases. Some analyzers will complain if your `switch` doesn't cover every case.

EasyRhin0: ok

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -0,0 +1,202 @@
...
+                          ""kind"": ""namespace""
+                        }
+                      },
+                      ""results"": [
+                      ]
+                    }
+                  ]
+                }";
+
+            string v1LogExpectedText =
```
lgolding: As a matter of course, I `const` strings that aren't expected to change (and then follow the capitalization convention for `const`):
```
const string V2LogText = @"...";
```

lgolding: It's just one more way the compiler can help you out a little, at almost no cost to you.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            ContractResolver = SarifContractResolver.Instance,
+            Formatting = Formatting.Indented
+        };
+
+        private static SarifLog GetSarifLog(string logText)
+        {
+            return JsonConvert.DeserializeObject<SarifLog>(logText, s_v2JsonSettings);
+        }
+
+        private static SarifLogVersionOne TransformCurrentToVersionOne(string V2LogText)
```
lgolding: I'm sorry, I meant only to capitalize the variables that you explicitly declared `const`. The capitalization convention for parameters is `camelCase`.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            SarifLogVersionOne v1Log = TransformCurrentToVersionOne(V2LogText);
+            string v1LogText = JsonConvert.SerializeObject(v1Log, s_v1JsonSettings);
+            v1LogText.Should().Be(v1LogExpectedText);
+        }
+
+        [Fact]
+        public void SarifTransformerTests_ToVersionOne_Minimum()
+        {
+            const string V2LogText =
```
lgolding: Yes.

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
```diff
@@ -0,0 +1,29 @@
...
+        private const string V1_0_0 = "1.0.0";
+
+        public static string ConvertToText(this SarifVersionVersionOne sarifVersion)
+        {
+            switch (sarifVersion)
+            {
+                case SarifVersionVersionOne.OneZeroZero:
+                    return V1_0_0;
+                default:
+                    throw new ArgumentException("Unsupported version", nameof(sarifVersion));
```
lgolding: Yes.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Ah find&replace

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
lgolding: Nit: I'd say "Unsupported SARIF version".

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -273,9 +272,10 @@ private static IList<FileLocation> CreateResponseFilesList(IDictionary<string, s
 
         public static Invocation CreateInvocation(InvocationVersionOne v1Invocation,
                                            IEnumerable<NotificationVersionOne> v1ToolNotifications,
-                                           IEnumerable<NotificationVersionOne> v1ConfigurationNotifications)
+                                           IEnumerable<NotificationVersionOne> v1ConfigurationNotifications,
+                                           Run run)
```
lgolding: You know, if you're passing the run, you don't need to pass the rest of the parameters; you can get them all from the run.

EasyRhin0: SIGH yes

lgolding: It's the way of things. As you clean code up, you discover further refinements. We're almost there now!

EasyRhin0: Oh actually no, all that stuff is from the v1 run.

lgolding: Oh, right!

src/Sarif/VersionOne/SarifUtilitiesVersionOne.cs
EasyRhin0: Fixed

PR #877, opened 2018-05-01
Transform v2 -> v1 files + rules + tests

EasyRhin0:


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +198,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+      ""results"": []
+    }
+  ]
+}";
+
+            VerifyCurrentToVersionOneTransformation(V2LogText, V1LogExpectedText);
+        }
+
+        [Fact]
+        public void SarifTransformerTests_ToVersionOne_OneRunWitRules()
```
lgolding: "Wit" => "With"

EasyRhin0: No this is me being street ;)

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public static class SarifTransformerUtilities
+    {
+        #region Text MIME types
```
lgolding: I've seen these types defined elsewhere. Don't duplicate them.

lgolding: Oh hah I see you just moved them. GOod.

EasyRhin0: This is a move

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +198,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+      },
+      ""invocations"": [
+        null
+      ],
+      ""files"": {
+        ""file:///home/list.txt"": {
+          ""length"": 43,
+          ""mimeType"": ""text/plain"",
+          ""contents"": {
+            ""text"": ""VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw=="",
```
lgolding: The `text` property is intended to be readable.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +185,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""files"": {
+        ""file:///home/list.txt"": {
+          ""length"": 43,
+          ""mimeType"": ""text/plain"",
+          ""contents"": ""The quick brown fox jumps over the lazy dog"",
```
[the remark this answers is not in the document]
EasyRhin0: So when I transform, should I decode it back to text and put the string in the Text property?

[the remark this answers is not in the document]
EasyRhin0: Ah never mind, I'm getting myself all turned around.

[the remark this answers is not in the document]
EasyRhin0: ok here is a question for the v1 -> v2 transformer: if contents is set, and the mime type is a text one, do I assume UTF8 and convert to a readable string that goes in fileContent.text?
byte[] data = Convert.FromBase64String(encodedString);
string decodedString = Encoding.UTF8.GetString(data);

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +185,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""invocations"": [
```
lgolding: This isn't valid SARIF. If `invocations` exists, it is an array of one or more unique `invocation` objects. `null` is not a valid array entry.

SARIF does allow an empty array (see §3.6.1) in the same way it allows an empty object (§3.5), so you can do whichever you like here.

EasyRhin0: Yes I fixed this in the upcoming commit

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +185,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+          ""mimeType"": ""text/plain"",
+          ""contents"": {
+            ""text"": ""The quick brown fox jumps over the lazy dog"",
+            ""binary"": ""VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw==""
+          },
+          ""hashes"": [
+            {
+              ""value"": ""d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592"",
+              ""algorithm"": ""sha-256""
+            }
```
lgolding: Include the v2 `file` properties that didn't exist in v1: `encoding` and `roles`. Our informal email-based functional spec for the current-to-v1 converter says that they should go into the v1 `file` object's property bag.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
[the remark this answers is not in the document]
lgolding: @EasyRhin0 So, to be clear, you want to put the value of v2's `file.contents.binary` here.

The hard question is what to do when the v2 `file.contents` has only a `text` property. I think the answer is:
- If you know `file.encoding`, and it's one that .NET supports, then encode the string with that encoding and populate `file.contents` with the MIME Base64 encoded version of the result.
- Otherwise, add a property `textContents` to the v1 `file`'s property bag.
- In any case, if v2 `file.encoding` is present, add `encoding` to the v1 `file`'s property bag.

And of course we need tests for all of these. I would use separate test methods to make the sample log files shorter and to clarify the intent of each test.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -198,5 +185,208 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithLogicalLocations()
...
+        },
+        ""C2003"": {
+          ""id"": ""C2003"",
+          ""name"": ""Rule C2003"",
+          ""shortDescription"": ""Rules were meant to be broken."",
+          ""fullDescription"": ""Rent internal rebellion competence biography photograph."",
+          ""configuration"": ""disabled"",
+          ""defaultLevel"": ""pass""
+        }
+      }
```
lgolding: As in the case of the `file` object, add a test that shows that _new_ v2 `rule` properties go in the v1 `rule`'s property bag. For example, `helpLocation` and `configuration.parameters`. Note that the latter is itself a property bag.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -78,6 +164,56 @@ public static LogicalLocationVersionOne CreateLogicalLocationVersionOne(LogicalL
...
+                }
+
+                if (v2Rule.Name != null)
+                {
+                    rule.Name = v2Rule.Name.Text;
+                }
+
+                if (v2Rule.FullDescription != null)
+                {
+                    rule.FullDescription = v2Rule.FullDescription.Text;
```
lgolding: @michaelcfanning @EasyRhin0 The v2 `rule.FullDescription` property is a `message` object. That means in might contain both a plain text `text` property and a Markdown-based `richText` property. It might also have an `arguments` property. And it might even specify the message by means of its "resource identifier" properties `messageId` and `richMessageId`.

Downgrading a v2 `message` object in a way that is round-trippable back to v1 is difficult! For now, please do this:

1. Copy `v2Rule.FullDescription.RichText` to `rule.RichDescription`. At least that's an improvement.
2. File an issue "Design and implement v2 -> v1 -> v2 message round-tripping".
3. Create a GitHub issue label `converter` and label this issue with it.

... because I think we're going to find many issues like this, where we need to make progress on the easy cases without losing sight of the corner cases we'll have to deal with.

EasyRhin0: Doesn't seem to be a rule.RichDescription property.

lgolding: You're right. That was crazy talk. I stand by #&#xfeff;2 and #&#xfeff;3, though :-)

michaelcfanning:
@lgolding @EasyRhin0 

Why is this still open? i.e., why aren't we doing this work incrementally in the sarif-v2 branch?

lgolding:
@michaelcfanning This was the PR where we tripped over the fact that we hadn't updated the schema (for example, `logicalLocation.decoratedName` didn't yet exist. So @EasyRhin0 had to detour to do some schema updates.

EasyRhin0:
Actually this stalled because of the whole round-tripping discussion. I switched back to v1->v2 at that point, and that's when we hit the schema update thing.

michaelcfanning:
How do we unblock progress here? I don't know what the 'schema update' thing refers to, maybe there's something in email on it?

michaelcfanning:
OK.

lgolding:
@EasyRhin0 @michaelcfanning

One of the comments above mentions v2 properties that don't exist in v1. I suggest we take the approach we took in the v1 => v2 converter, and stash the entire v2 run in the v1 run's property bag.

EasyRhin0:
Agreed.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -121,13 +122,13 @@ public void SarifTransformerTests_ToCurrent_MinimumWithTwoRuns()
...
 
-            v2LogText.Should().Be(v2LogExpectedText);
+            VerifyVersionOneToCurrentTransformation(V1LogText, V2LogExpectedText);
         }
 
         [Fact]
         public void SarifTransformerTests_ToCurrent_MinimumWithPropertyAndTags()
         {
-            string v1LogText =
+            string V1LogText =
```
lgolding: `const`

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -78,7 +101,75 @@ public static LogicalLocationVersionOne CreateLogicalLocationVersionOne(LogicalL
...
+        {
+            RunVersionOne run = null;
+
+            if (v2Run != null)
+            {
+                string serializedV1Run;
+
+                if (v2Run.TryGetProperty("sarifv1/run", out serializedV1Run))
+                {
+                    run = JsonConvert.DeserializeObject<RunVersionOne>(serializedV1Run, Utilities.JsonSettingsV1);
```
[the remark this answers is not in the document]
EasyRhin0: Yes I'm aware of that.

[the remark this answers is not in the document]
EasyRhin0: Added.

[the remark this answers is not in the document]
EasyRhin0: Actually I've made some other changes in that file so I'll wait to commit this one.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -63,13 +67,13 @@ public void SarifTransformerTests_ToCurrent_Minimum()
...
 
-            v2LogText.Should().Be(v2LogExpectedText);
+            VerifyVersionOneToCurrentTransformation(V1LogText, V2LogExpectedText);
         }
 
         [Fact]
         public void SarifTransformerTests_ToCurrent_MinimumWithTwoRuns()
         {
-            string v1LogText =
+            string V1LogText =
```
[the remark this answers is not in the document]

PR #881, opened 2018-05-08
Add utilities class, param name cleanup, tests to new model

EasyRhin0:


src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -0,0 +1,155 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public static class SarifTransformerUtilities
```
lgolding: I love how we've hoisted these things twice, to more and more "common" location :-)

EasyRhin0: Time sometimes brings clarity...

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -225,49 +235,56 @@ public void SarifTransformerTests_ToCurrent_OneRunWithFiles()
...
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""files"": {
+        ""file:///home/list.txt"": {
+          ""length"": 43,
+          ""mimeType"": ""text/plain"",
+          ""contents"": {
+            ""text"": ""VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw=="",
```
lgolding: The MIME type tells us that this is a text file. In that case, the v1 spec tells us that v1 `file.contents` is a string containing a MIME-64 encoding of the UTF-8-transcoded contents of the file.

So we should populate v2 `file.contents.text` by decoding v1 `file.contents`. The result should be readable.

We don't know the original file's encoding, so we can't deduce the binary contents of the original file. So we should _not_ populate v2 `file.contents.binary`.

I could've sworn I did a review of that logic. Am I high?

EasyRhin0: No you are probably right. I definitely made the change, let me go find it in my other branch.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -225,49 +235,56 @@ public void SarifTransformerTests_ToCurrent_OneRunWithFiles()
...
+          ""hashes"": [
+            {
+              ""value"": ""d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592"",
+              ""algorithm"": ""sha-256""
+            }
+          ]
+        },
+        ""file:///home/buildAgent/bin/app.zip"": {
+          ""mimeType"": ""application/zip"",
+          ""contents"": {},
```
lgolding: This corresponds to your previous assertion `fileData.Contents.Should().NotBeNull()`. But why _isn't_ it `null`? Why are we emitting an empty object?

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -317,60 +334,65 @@ public void SarifTransformerTests_ToCurrent_OneRunWitRules()
...
+            }
+          },
+          ""C2002"": {
+            ""id"": ""C2002"",
+            ""fullDescription"": {
+              ""text"": ""Catfish season continuous hen lamb include dose copy grant.""
+            },
+            ""configuration"": {
+              ""enabled"": true,
+              ""defaultLevel"": 2
```
lgolding: The value should be the string `"error"`. Are we missing an `EnumHint` in CodeGenHints.json?

lgolding: And see how valuable it was to change the test to this form!

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -317,60 +334,65 @@ public void SarifTransformerTests_ToCurrent_OneRunWitRules()
...
+            ""name"": {
+              ""text"": ""Rule C2003""
+            },
+            ""shortDescription"": {
+              ""text"": ""Rules were meant to be broken.""
+            },
+            ""fullDescription"": {
+              ""text"": ""Rent internal rebellion competence biography photograph.""
+            },
+            ""configuration"": {}
```
lgolding: Why emit empty object?

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Fixed. Missing ContractResolver case for RuleConfigurationDefaultLevel

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: It's not actually empty. Enabled is false, and DefaultLevel is Note (0). So I guess those values lead to no output?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -9,218 +9,177 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
...
                     {
-                        Uri = node.Uri,
-                        UriBaseId = node.UriBaseId
+                        Uri = v1FileData.Uri,
+                        UriBaseId = v1FileData.UriBaseId
                     };
                 }
 
-                fileData.Contents = new FileContent
+                if (!string.IsNullOrWhiteSpace(v1FileData.Contents))
```
lgolding: Didn't we settle on just checking for `null` here, because the empty string `""` is valid? And I think I said something like "Don't bother checking for an all-white-space string. There are lots of strings that aren't valid MIME Base64, like ")(*^%&*(", and we're not checking for those".

EasyRhin0: Yup, fixed

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Makes sense. I was doing the check but was looking for the wrong values.

EasyRhin0:
Yeah already caught that

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -9,218 +9,177 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
...
-                    MessageStrings = node.MessageFormats,
-                    Properties = node.Properties
+                    Id = v1Rule.Id,
+                    MessageStrings = v1Rule.MessageFormats,
+                    Properties = v1Rule.Properties
                 };
 
-                if (node.Configuration != RuleConfigurationVersionOne.Unknown &&
-                    node.DefaultLevel != ResultLevelVersionOne.Default)
+                if (v1Rule.Configuration == RuleConfigurationVersionOne.Enabled &&
```
lgolding: This needs to be `||`, not `&&`. You want to emit the object if _either_ of the properties is non-default.

EasyRhin0: Grr, sorry

EasyRhin0: So I'm getting an empty value when defaultLevel is "note" (converted from "pass")

lgolding: You're saying that when you have a v1 file with `rule.defaultLevel == "note"`, and you convert it to v2, that the v2 file does _not_ contain a `rule.configuration` object? That behavior doesn't match the code (or at least, it doesn't match the code as it should be written):
```
if (v1Rule.Configuration == RuleConfigurationVersionOne.Enabled ||
    v1Rule.DefaultLevel != ResultLevelVersionOne.Warning)
{
    // Create the Configuration object.
}
```
How sure are you of this? Like, you saved the file, and rebuilt, and ran the test, and you got that result?

EasyRhin0: When the v1 value is "pass", it gets converted to v2 "note" and nothing is emitted. I'm watching the value get set in the debugger.

lgolding: When the debugger hits the conditional, does the conditional return `true`? Do you get into the code block that creates the `RuleConfiguration` object?

I'm gonna ask you to investigate this one.

EasyRhin0: OK I wrote a whole big thing but I'll give you the abridged version: whatever enum member has value 0 will not be emitted. Even the default ("warning") emits as long as it's not the first member.  I'm happy to explain all of the experiments I performed.

This isn't caught by the current test because the values in the OM are correct (big advantage of comparing json!),

lgolding: Ok. It seems to me that we want to always emit the `ruleConfiguration.defaultLevel` property, even if it is the "0" value.

Is there a codegen hint that says "don't emit the `EmitDefaultValue` attribute"?

lgolding: Sorry, I mean: "if we are going to emit the `ruleConfiguration` object _at all_, we should always emit its `defaultLevel` property". We should still test for the "all defaults" case so that we can decide not to emit the  `ruleConfiguration` object at all.

EasyRhin0: Right. I think all we need is to add a "zeroValueName": "Default" hint. Result.Level has this.

lgolding: Ah, good.

EasyRhin0: I used "None" instead of "Default", that ok?

lgolding: Sure, ok, `None`.

EasyRhin0: Yeah that was fixed after my last push. Current code:
RuleConfigurationDefaultLevel level = SarifTransformerUtilities.CreateRuleConfigurationDefaultLevel(v1Rule.DefaultLevel);

if (v1Rule.Configuration == RuleConfigurationVersionOne.Enabled ||
    level != RuleConfigurationDefaultLevel.Warning)
{
    rule.Configuration = new RuleConfiguration
    {
        DefaultLevel = level,
        Enabled = v1Rule.Configuration == RuleConfigurationVersionOne.Enabled
    };
}

lgolding: Ok.

PR #882, opened 2018-05-09
Stack + StackFrame + tests

EasyRhin0:
Notes:
- Sucked a Viewer fix up from master, you can ignore
- SarifVersionOneToCurrentVisitor.cs is a bit messy because I alphabetized the Create methods. Just look at CreateStack and CreateStackFrame, plus the unit test

### DIFF
--- a/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
+++ b/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
@@ -714,5 +714,141 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             v2LogText.Should().Be(v2LogExpectedText);
         }
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
+        {
+            string v1LogText =
+              @"{
+                  ""version"": ""1.0.0"",
+                  ""runs"": [
+                    {
+                      ""configurationNotifications"": [
+                        {
+                          ""id"": ""UnknownRule"",
+                          ""ruleId"": ""ABC0001"",
+                          ""level"": ""warning"",
+                          ""message"": ""Could not disable rule \""ABC0001\"" because there is no rule with that id.""
+                        }
+                      ],
+                      ""toolNotifications"": [
+                        {
+                          ""id"": ""CTN0001"",
+                          ""level"": ""note"",
+                          ""message"": ""Run started."",
+                          ""exception"": {
+                            ""kind"": ""ExecutionEngine.RuleFailureException"",
+                            ""message"": ""Unhandled exception during rule evaluation."",
+                            ""stack"": {
+                              ""message"": ""This is the stack messasge."",
+                              ""frames"": [
+                                {
+                                  ""message"": ""Exception thrown"",
+                                  ""module"": ""RuleLibrary"",
+                                  ""threadId"": 52,
+                                  ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Evaluate"",
+                                  ""address"": 10092852
+                                },
+                                {
+                                  ""module"": ""ExecutionEngine"",
+                                  ""threadId"": 52,
+                                  ""fullyQualifiedLogicalName"": ""ExecutionEngine.Engine.EvaluateRule"",
+                                  ""address"": 10073356,
+                                  ""offset"": 10475
+                                }
+                              ]
+                            },
+                            ""innerExceptions"": [
+                              {
+                                ""kind"": ""System.ArgumentException"",
+                                ""message"": ""length is < 0""
+                              }
+                            ]
+                          }
+                        }
+                      ],
+                      ""tool"": {
+                        ""name"": ""CodeScanner"",
+                        ""semanticVersion"": ""2.1.0""
+                      },
+                      ""results"": [
+                      ]
+                    }
+                  ]
+                }";
+
+            SarifLog v2Log = TransformVersionOneToCurrent(v1LogText);
+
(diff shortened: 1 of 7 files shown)


lgolding:
@EasyRhin0 Thanks for the review notes.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -710,5 +693,141 @@ public void SarifTransformerTests_ToCurrent_OneRunWithNotificationsButNoInvocati
...
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
+        {
+            string v1LogText =
+              @"{
+                  ""version"": ""1.0.0"",
+                  ""runs"": [
+                    {
+                      ""configurationNotifications"": [
```
lgolding: Since this PR is just about stacks and stack frames, let's omit `configurationNotifications`.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -710,5 +693,141 @@ public void SarifTransformerTests_ToCurrent_OneRunWithNotificationsButNoInvocati
...
+                          ""ruleId"": ""ABC0001"",
+                          ""level"": ""warning"",
+                          ""message"": ""Could not disable rule \""ABC0001\"" because there is no rule with that id.""
+                        }
+                      ],
+                      ""toolNotifications"": [
+                        {
+                          ""id"": ""CTN0001"",
+                          ""level"": ""note"",
+                          ""message"": ""Run started."",
```
lgolding: Let's make the example more realistic:
```
"level": "error",
"message": "Unhandled exception."
```

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -710,5 +693,141 @@ public void SarifTransformerTests_ToCurrent_OneRunWithNotificationsButNoInvocati
...
+                  ""message"": {
+                    ""text"": ""This is the stack messasge.""
+                  },
+                  ""frames"": [
+                    {
+                      ""module"": ""RuleLibrary"",
+                      ""threadId"": 52,
+                      ""address"": 10092852,
+                      ""properties"": {
+                        ""sarifv1/message"": ""Exception thrown""
```
lgolding: You noticed that the v2 `stackFrame` object doesn't have a `message` property. But it does have a `location` property, which has a `message`, so write:
```
{
  "module": "RuleLibrary",
  ...
  "location": {
    "message": {
      "text": "Exception thrown."
    }
  }
}
```

EasyRhin0: Ah ok. I haven't done location yet (next batch).

lgolding: Ok. How are we going to remember to fix this when you do the `location` PR?

lgolding: A sticky note :-)

EasyRhin0: Yeah I have a method that has worked so far :)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -430,7 +439,76 @@ public override RunVersionOne VisitRunVersionOne(RunVersionOne v1Run)
...
+                    foreach (StackFrameVersionOne v1StackFrame in v1Stack.Frames)
+                    {
+                        stack.Frames.Add(CreateStackFrame(v1StackFrame));
+                    }
+                }
+
+                if (v1Stack.Tags.Count > 0)
+                {
+                    stack.Tags.UnionWith(v1Stack.Tags);
+                }
```
lgolding: I hadn't caught this before. tags are persisted in the property bag:
```
{
  "properties": {
    "foo": "bar",
    "tags": [ "1", "2" ]
  }
}
```
So I would expect that when you copy the property bag from `v1Stack` to `stack`, the tags come along for the ride, and `stack.Tags` should automatically access them. So is this necessary?

Let's add a test for copying tags, and convince ourselves that it passes even if we remove this code (and similar code in other methods of this class).

EasyRhin0: Don't remember why I did it that way...done.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -118,6 +118,102 @@ public void SarifTransformerTests_ToCurrent_MinimumWithTwoRuns()
...
+      ""results"": []
+    }
+  ]
+}";
+
+            v2LogText.Should().Be(v2LogExpectedText);
+        }
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_MinimumWithTagsOnly()
```
lgolding: I don't see the added value of this test unless we're worried that tags only come over if some _other_ property also comes over. For me the first test is enough.

EasyRhin0: Fine by me. I tend to err on the side of excess with tests

PR #897, opened 2018-05-16
Transform stackFrame.location, create logicalLocations

EasyRhin0:
Not sure if you'll like this. :)
I needed a way to determine the FQN delimiter so I made an array of common ones. See GetLogicalLocationName


src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -226,7 +224,6 @@ public void SarifTransformerTests_ToCurrent_OneRunWithLogicalLocations()
           ""kind"": ""type""
         },
         ""collections"": {
-          ""name"": ""collections"",
```
lgolding: Why did this test not fail before your last merge?

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -84,6 +87,23 @@ public static class SarifTransformerUtilities
             { "sha-512", AlgorithmKindVersionOne.Sha512 }
         };
 
+        public static IList<TTo> TransformList<TFrom, TTo>(IList<TFrom> fromList, Func<TFrom, TTo> convertFunc)
```
lgolding: I think you just re-invented LINQ `Aggregate` :-)

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: That change affected the converters, where this one is the transformer. I thought of something we need to check though, which is that the logger is doing this correctly.

src/Sarif/Visitors/SarifTransformerUtilities.cs
EasyRhin0: Yeah but mine handles a null list. Couldn't think of a better way.

EasyRhin0: Ah I can use ?
I'll change this...

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -828,11 +839,47 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
...
+                        ""physicalLocation"": {
+                          ""fileLocation"": {
+                            ""uri"": ""file:///C:/src/main.cs""
+                          },
+                          ""region"": {
+                            ""startLine"": 15,
+                            ""startColumn"": 9
+                          }
+                        },
+                        ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Evaluate-0"",
```
lgolding: How does the transformer know that this isn't the same `Rules.SecureHashAlgorithm.Evaluate` that's already in the dictionary? That is, why did it decide to disambiguate. (Don't answer yet; this is a note-to-self. I'll keep reading.)

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -828,11 +839,47 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
...
                       ""address"": 10092852
                     },
                     {
+                      ""location"": {
+                        ""physicalLocation"": {
+                          ""fileLocation"": {
+                            ""uri"": ""file:///C:/src/main.cs""
+                          }
+                        },
+                        ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Register""
```
lgolding: Whereas this one it decided _not_ to disambiguate. Perhaps it's the `kind` on `Evaluate` that didn't match. If the transformer knows what `kind` is, though, it should populate it. We'll see...

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -17,6 +18,8 @@ public static class SarifTransformerUtilities
             { SarifVersion.TwoZeroZero, "sarifv2" }
         };
 
+        public static readonly string[] FullyQualifiedNameDelimiters = { ".", "/", "\\", "::" };
```
lgolding: I'm ok with this for now. Ultimately we might want a command line argument to override this (in which case I'd rename this to `DefaultFullyQualifiedNameDelimiters`. You might do that rename anyway, and add a comment suggesting the possibility of overriding it.

EasyRhin0: Good idea.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -29,28 +34,19 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne v1
...
-        public static ExceptionData CreateExceptionData(ExceptionDataVersionOne v1ExceptionData)
+       internal ExceptionData CreateExceptionData(ExceptionDataVersionOne v1ExceptionData)
         {
             ExceptionData exceptionData = null;
 
             if (v1ExceptionData != null)
             {
                 exceptionData = new ExceptionData
                 {
+                    InnerExceptions = v1ExceptionData.InnerExceptions?.Select(e => CreateExceptionData(e)).ToList(),
```
lgolding: Hah! You're becoming the LINQ whisperer. Pro tip: If the right-hand side of a lambda is a single call to a function whose signature matches the lambda's signature, you can just say its name:
```
InnerExceptions = v1ExceptionData.InnerExceptions?.Select(CreateExceptionData).ToList()
```

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -29,28 +34,19 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne v1
             return null;
         }
 
-        public static ExceptionData CreateExceptionData(ExceptionDataVersionOne v1ExceptionData)
```
lgolding: Yay! for reduced visibility. And I assume it's no longer `static` because somewhere down the call stack, we'll run into somebody that needs `_run`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Ah! cool

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -60,14 +56,15 @@ public static ExceptionData CreateExceptionData(ExceptionDataVersionOne v1Except
...
-        public static FileData CreateFileData(FileDataVersionOne v1FileData)
+        internal FileData CreateFileData(FileDataVersionOne v1FileData)
         {
             FileData fileData = null;
 
             if (v1FileData != null)
             {
                 fileData = new FileData
                 {
+                    Hashes = v1FileData.Hashes?.Select(h => CreateHash(h)).ToList(),
```
lgolding: Again, you can say
```
Hashes = v1FileData.Hashes?.Select(CreateHash).ToList()

File: src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs [hunk 9]
```diff
@@ -134,14 +137,13 @@ public static Hash CreateHash(HashVersionOne v1Hash)
             return hash;
         }
 
-        public static Invocation CreateInvocation(InvocationVersionOne v1Invocation,
-                                           IEnumerable<NotificationVersionOne> v1ToolNotifications,
-                                           IEnumerable<NotificationVersionOne> v1ConfigurationNotifications,
-                                           Run run)
```
lgolding: Right, we've given up on passing `run` all the way down.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -134,14 +137,13 @@ public static Hash CreateHash(HashVersionOne v1Hash)
...
+        internal Invocation CreateInvocation(InvocationVersionOne v1Invocation,
+                                             IList<NotificationVersionOne> v1ToolNotifications,
+                                             IList<NotificationVersionOne> v1ConfigurationNotifications)
         {
-            Invocation invocation = CreateInvocation(v1Invocation, run);
-            IList<Notification> toolNotifications = CreateNotificationsList(v1ToolNotifications);
-            IList<Notification> configurationNotifications = CreateNotificationsList(v1ConfigurationNotifications);
+            Invocation invocation = CreateInvocation(v1Invocation);
+            IList<Notification> toolNotifications = v1ToolNotifications?.Select(n => CreateNotification(n)).ToList();
```
lgolding: ... and again :-)

EasyRhin0: Yeah I'll get them all :)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -189,7 +191,59 @@ public static Invocation CreateInvocation(InvocationVersionOne v1Invocation, Run
...
+            {
+                location = new Location
+                {
+                    FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
+                    PhysicalLocation = CreatePhysicalLocation(v1Location.ResultFile),
+                    Properties = v1Location.Properties
+                };
+
+                location.SetProperty($"{FromPropertyBagPrefix}/analysisTarget", v1Location.AnalysisTarget);
+                location.SetProperty($"{FromPropertyBagPrefix}/decoratedName", v1Location.DecoratedName);
```
lgolding: Note to self: find out where `decoratedName` gets put in its new home in the `logicalLocation` object. And since it does have a home there, do we have to put it in the property bag? A v2 => v1 transformer would know where to find it.

lgolding: Likewise, `analysisTarget` goes on the `result`.

EasyRhin0: Yeah I have been of the mind to stash things like this that will be a chore to reassemble.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -189,7 +191,59 @@ public static Invocation CreateInvocation(InvocationVersionOne v1Invocation, Run
             return invocation;
         }
 
-        public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1LogicalLocation)
+        internal Location CreateLocation(LocationVersionOne v1Location)
```
lgolding: As far as I can see nobody is calling this yet (at least not from this file). Can we wait to add it?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifTransformerUtilities.cs
EasyRhin0: Renamed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Yeah I have been working down the line and then moving code out to a waiting area. Missed this one

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -189,7 +191,59 @@ public static Invocation CreateInvocation(InvocationVersionOne v1Invocation, Run
...
+                };
+
+                location.SetProperty($"{FromPropertyBagPrefix}/analysisTarget", v1Location.AnalysisTarget);
+                location.SetProperty($"{FromPropertyBagPrefix}/decoratedName", v1Location.DecoratedName);
+            }
+
+            return location;
+        }
+
+        internal Location CreateLocation(string fullyQualifiedLogicalName,
```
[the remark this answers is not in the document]
EasyRhin0: OK

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -189,7 +191,59 @@ public static Invocation CreateInvocation(InvocationVersionOne v1Invocation, Run
...
+            {
+                Message = CreateMessage(message)
+            };
+
+            if (string.IsNullOrWhiteSpace(logicalLocationKey) &&
+                !string.IsNullOrWhiteSpace(fullyQualifiedLogicalName))
+            {
+                LogicalLocation logicalLocation = CreateLogicalLocation(fullyQualifiedLogicalName);
+                location.FullyQualifiedLogicalName = AddLogicalLocation(logicalLocation);
+            }
```
lgolding: Bug: If they are both non-null, then `location.FQLN` should get `logicalLocationKey`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: That is a reasonable design decision, and worth a comment.

However -- is there any point in this if we decide (_did_ we decide?) to just paste the entire v1 run into the v2 run's property bag?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
-            List<Notification> notifications = null;
+            var logicalLocation = new LogicalLocation
+            {
+                DecoratedName = decoratedName,
+                FullyQualifiedName = fullyQualifiedLogicalName,
+                Name = GetLogicalLocationName(fullyQualifiedLogicalName),
+                ParentKey = parentKey
+            };
+
+            return logicalLocation;
```
lgolding: or just
```
return new LogicalLocation
{
    ...
};
```

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Yeah I've gone round and round on that subject

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
-                notifications = new List<Notification>();
+                _currentRun.LogicalLocations = new Dictionary<string, LogicalLocation>();
+            }
+
+            string fullyQualifiedName = logicalLocation.FullyQualifiedName;
+            string logicalLocationKey = logicalLocation.FullyQualifiedName;
+            int disambiguator = 0;
 
-                foreach (NotificationVersionOne notification in v1Notifications)
+            // We need to do the comparison here too
```
lgolding: What does that mean. Apparently you're doing it somewhere else, but it's not immediately apparent where. Also, it's not clear _why_ you need to do the comparison in two places.

EasyRhin0: Sorry, that was a note for me, meant to remove it.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
+                logLoc.Name = GetLogicalLocationName(fullyQualifiedName);
+                logLoc.FullyQualifiedName = logicalLocation.FullyQualifiedName;
+
+                if (logicalLocation.ValueEquals(logLoc))
                 {
-                    notifications.Add(CreateNotification(notification));
+                    break;
                 }
+
+                logicalLocationKey = logicalLocation.FullyQualifiedName + "-" + disambiguator.ToString(CultureInfo.InvariantCulture);
```
lgolding: I've seen this code before. How about DRYing it out into `Utilities.CreateDisambiguatedName(string baseName, int index)`? Then if we change the naming convention, we only have to change it in one place.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
+        internal string GetLogicalLocationName(string fullyQualifiedLogicalName)
+        {
+            if (string.IsNullOrWhiteSpace(fullyQualifiedLogicalName))
+            {
+                throw new ArgumentNullException(nameof(fullyQualifiedLogicalName));
+            }
+
+            return fullyQualifiedLogicalName.Split(Utilities.FullyQualifiedNameDelimiters,
+                                                   StringSplitOptions.RemoveEmptyEntries).Last();
+        }
```
lgolding: Again, I feel like I've seen this code before. I think there's some DRYing out to be done between the converters and the transformer. If you agree, let's merge what we have, then do a refactoring PR to clean this up.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Yup

lgolding: BTW, I'd suggest this encapsulation even if the code only occurred once, for the sake of clarity.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Sounds good.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
+
+        internal void FixUpAllLogicalLocations()
+        {
+            foreach (string key in _currentRun.LogicalLocations.Keys)
+            {
+                FixUpLogicalLocation(key);
+            }
+        }
+
+        internal void FixUpLogicalLocation(string key)
```
lgolding: Consider: replacing/renaming the "FixUp" methods as follows:
```
private RemoveRedundantLogicalLocationProperties()
{
    foreach (KeyValuePair<string, LogicalLocation logicalLocation> entry in _currentRun.LogicalLocation)
    {
        RemoveRedundantProperties(entry.Key, entry.Value);
    }
}

private void RemoveRedundantProperties(string key, LogicalLocation logicalLocation)
{
    ...
}
```

lgolding: Mostly it's a naming thing. I didn't want to have to read further down the page to know what was being "fixed up". Then once I knew, the renaming and refactoring suggested itself.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -206,24 +260,105 @@ public static LogicalLocation CreateLogicalLocation(LogicalLocationVersionOne v1
...
+            {
+                logicalLocation.Name = null;
+            }
+        }
+
+        internal Message CreateMessage(string text)
+        {
+            Message message = null;
+
+            if (!string.IsNullOrWhiteSpace(text))
```
lgolding: Tough call, but it somebody took the trouble to set `text` to `"   \t "`, I'd just go with it.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -295,20 +423,77 @@ private static IList<FileLocation> CreateResponseFilesList(IDictionary<string, s
...
                 rule = new Rule
                 {
+                    FullDescription = CreateMessage(v1Rule.FullDescription),
+                    HelpLocation = CreateFileLocation(v1Rule.HelpUri, null),
                     Id = v1Rule.Id,
                     MessageStrings = v1Rule.MessageFormats,
-                    Properties = v1Rule.Properties
+                    Name = CreateMessage(v1Rule.Name),
+                    Properties = v1Rule.Properties,
+                    ShortDescription = CreateMessage(v1Rule.ShortDescription)
```
lgolding: Ah yes, now with all the machinery in place, this becomes easy.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Done.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Done. In SarifTransformerUtilities for now but I'll probably move this somewhere more global as part of the refactoring PR

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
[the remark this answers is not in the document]

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: (This code is out for now, other comment where you pointed out that nothing is calling it)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -224,11 +208,17 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
-                LogicalLocation logicalLocation = CreateLogicalLocation(fullyQualifiedLogicalName);
-                location.FullyQualifiedLogicalName = AddLogicalLocation(logicalLocation);
+                if (string.IsNullOrWhiteSpace(logicalLocationKey))
+                {
+                    LogicalLocation logicalLocation = CreateLogicalLocation(fullyQualifiedLogicalName);
+                    location.FullyQualifiedLogicalName = AddLogicalLocation(logicalLocation);
+                }
+                else
+                {
+                    location.FullyQualifiedLogicalName = logicalLocationKey;
```
lgolding: Bug: In this branch, you don't add the `logicalLocation` object to the dictionary. You might need to refactor `AddLogicalLocation` to separate out the disambiguating logic from the add-to-dictionary logic,

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -316,19 +303,19 @@ internal string GetLogicalLocationName(string fullyQualifiedLogicalName)
...
                                                    StringSplitOptions.RemoveEmptyEntries).Last();
         }
 
-        internal void FixUpAllLogicalLocations()
+        internal void RemoveRedundantLogicalLocationProperties()
         {
             foreach (string key in _currentRun.LogicalLocations.Keys)
             {
-                FixUpLogicalLocation(key);
+                RemoveRedundantProperties(key);
```
lgolding: Should I be concerned that `RemoveRedundantProperties` is called twice, once from `RemoveRedundantLogicalLocationProperties` and once from `AddLogicalLocation`?

EasyRhin0: No. The call to RemoveRedundantLogicalLocationProperties is made after the v1 logicalLocation dictionary is imported, to ensure the properties that should be null are. The second call is made for each new l.L.

lgolding: Got it.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: I am out of gas, need a glass of wine. I'll finish this later, but don't wait up.

lgolding: No problem. Great progress, almost there.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: @EasyRhin0 Chris, I think this is still an open issue, right? We're not yet putting `decoratedName` or `analysisTarget` in their new homes.

EasyRhin0: v1 stackFrame doesn't have decoratedName. But I did raise an issue about the logicalLocation pointed to by the key property, see my comment from a few minutes ago.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Ah, you are right: Since this overload of `CreateLocation` was called with properties extracted directly from a v1 stack frame, there is already a dictionary entry with the disambiguated key.

That is _so_ worth a comment!

Yes, in this case, we should update the existing v2 dictionary entry with the additional properties (`name`, `fullyQualifiedLogicalName`) -- because in this case, they must both be present, because neither one can possibly match the key. So it's still a bug.

Let me make sure I understand what you mean by this:
> And since we don't get kind nor decoratedName properties from the stackFrame, we will disambiguate if either has a value.

I think you mean that a v1 `logicalLocation` will never equal a v2 `logicalLocation` that has either a `kind` or a `decoratedName`, because a v1 `logicalLocation` doesn't have those properties.

That's true, but is that the right behavior? Suppose you're about to add a `logicalLocation` that matches an existing v2 `logicalLocation` _except_ for `kind` and/or `decoratedName`. _Should_ we assume this is a different logical location? I don't think so. I see that you have a test that verifies this exact behavior, and you'll recall I commented on it, expressing this concern.

EasyRhin0: Ok that's what I was thinking. The statement you ask about is just the condition that made me realize there is a need to pull the existing lL and update it.

lgolding: Ok, so there are two necessary changes:
1. Allow a v1 `logicalLocation` to match a v2 `logicalLocation` that contains `kind` or `decoratedName`. That means you can't just transform the v1 `logicalLocation` to v2 and then use the equality comparer. You'll have to compare a few properties "by hand" -- and boy, will that need a comment.
2. If you find a disambiguated key in the dictionary (the `else` clause), make sure to populate `name` and `fullyQualifiedName` in the `logicalLocation` identified by that key.

lgolding: 3. Of course that will break a test, so you'll need to fix the test.

EasyRhin0: OK. (v1 logicalLocation does have kind, so decoratedName is the only issue)

lgolding: You are right. I was inaccurate. v1 `logicalLocation`s do have a `kind`. **BUT!** a `logicalLocation` synthesized from a v1 `stackFrame` -- which is the scenario that this method covers -- will never match a v2 `logicalLocation` with a `kind`, because v1 `stackFrame` doesn't have a `kind`.

We need to fix that and allow a `logicalLocation` synthesized from a v1 `stackFrame` to match an existing v2 `logicalLocation` with a `kind` (or a `decoratedName`). That is my bullet point #&#xFEFF;1 above. Then my #&#xFEFF;2 and #&#xFEFF;3 follow immediately.

EasyRhin0: Check my thinking please: if the stackFrame has a logicalLocationKey, I can just update FQN and Name for that existing item, right? Otherwise, if I have a FQLN, do the create & add with the manual property comparison.

lgolding: Correct. And we still need to do my #&#xFEFF1, #&#xFEFF2, and #&#xFEFF3 above.

lgolding: Sorry, your last sentence implies my #&#xFEFF;1, #&#xFEFF;2, and #&#xFEFF;3. We are on the same page.

EasyRhin0: OK great, thanks

EasyRhin0:
Appveyor is screwed up.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: @EasyRhin0 It looks like we still have the bug. It shouldn't be necessary to disambiguate `Rules.SecureHashAlgorithmRule.Evaluate` -- because the v1 `logicalLocation` with that key contained only `"name": "Evaluate"`, which is correct, and `"kind": "some kind"`, which should not participate in the comparison.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -828,11 +852,61 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
...
                       ""address"": 10092852
                     },
                     {
+                      ""location"": {
+                        ""physicalLocation"": {
+                          ""fileLocation"": {
+                            ""uri"": ""file:///C:/src/main.cs""
+                          }
+                        },
+                        ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Register-0""
```
lgolding: Again, it shouldn't be necessary to disambiguate this, because the existing v1 `logicalLocation` has only a `name`, which is correct.

EasyRhin0:
???
NuGet.Protocol.Core.Types.FatalProtocolException: The local source 'C:\projects\sarif-sdk\bld\bin\nuget' doesn't exist.

lgolding:
No idea about the `FatalProtocolException`. Reviewing your latest changes now.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -851,6 +925,23 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
...
+          ""kind"": ""some kind""
+        },
+        ""Rules.SecureHashAlgorithmRule.Register"": {
+          ""name"": ""Register""
+        },
+        ""ExecutionEngine.Engine.FooBar"": {
+          ""name"": ""EvaluateRule"",
+          ""fullyQualifiedName"": ""ExecutionEngine.Engine.EvaluateRule"",
+          ""kind"": ""another kind""
+        },
```
lgolding: Bug: These `name` and `fullyQualifiedName` values belong in the entry whose key is `ExecutionEngine.Engine.EvaluateRule"`.

lgolding: ... and the `name` for this entry should be `"FooBar"`.

EasyRhin0: This is for the stackFrame with the logicalLocationKey property set?

EasyRhin0: If this is wrong, can you check the last v1 stackFrame and tell me if it's valid

lgolding: Ok. That last stack frame, with
```
"fullyQualifiedLogicalName": "ExecutionEngine.Engine.EvaluateRule"
"logicalLocationKey": "ExecutionEngine.Engine.FooBar"
```
is valid. I didn't notice where this `logicalLocation` came from.

I suggest you change `"FooBar"` to `"Evaluate-1"`. Because of the unrelated name, I didn't recognize that this was the "disambiguation" part of the test.

EasyRhin0: So it's not meant to be the disambiguation test, it's meant to test the presence of logicalLocationKey. (I need to add a disambig. case)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -189,7 +190,51 @@ public static Invocation CreateInvocation(InvocationVersionOne v1Invocation, Run
...
+            };
+
+            LogicalLocation logicalLocation;
+
+            if (!string.IsNullOrWhiteSpace(logicalLocationKey) &&
+                _currentRun.LogicalLocations.TryGetValue(logicalLocationKey, out logicalLocation))
+            {
+                logicalLocation.FullyQualifiedName = fullyQualifiedLogicalName;
+                logicalLocation.Name = GetLogicalLocationName(fullyQualifiedLogicalName);
+                location.FullyQualifiedLogicalName = logicalLocationKey;
```
lgolding: Yes, this looks right.

PR #900, opened 2018-05-18
Transform result

EasyRhin0:
Since a result can have multiple locations but only one analysisTarget, I am just grabbing the analysisTarget from location[0].

### DIFF
--- a/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
+++ b/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
@@ -37,8 +37,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                         ""name"": ""CodeScanner"",
                         ""semanticVersion"": ""2.1.0""
                       },
-                      ""results"": [
-                      ]
+                      ""results"": []
                     }
                   ]
                 }";
@@ -76,8 +75,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                         ""name"": ""CodeScanner"",
                         ""semanticVersion"": ""2.1.0""
                       },
-                      ""results"": [
-                      ]
+                      ""results"": []
                     },
                     {
                       ""tool"": {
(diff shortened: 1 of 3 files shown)


lgolding:
That is correct.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -953,5 +951,168 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
...
+                }
+              },
+              ""fullyQualifiedLogicalName"": ""collections::list::add""
+            }
+          ],
+          ""suppressionStates"": [""suppressedExternally""],
+          ""baselineState"": ""existing"",
+          ""properties"": {
+            ""sarifv1/formattedRuleMessage"": {""formatId"":""default"",""arguments"":[""ptr""]},
+            ""sarifv1/snippet"": ""add_core(ptr, offset, val);\n    return;""
```
lgolding: You can put this in `region.snippet` of type `fileContents`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -114,6 +114,22 @@ internal FileLocation CreateFileLocation(Uri uri, string uriBaseId)
...
+            if (v1PhysicalLocation?.Uri != null)
+            {
+                fileLocation = new FileLocation
+                {
+                    Uri = v1PhysicalLocation.Uri,
+                    UriBaseId = v1PhysicalLocation.UriBaseId
+                };
+            }
+
+            return fileLocation;
```
lgolding: Doesn't this exist somewhere else?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -190,6 +206,28 @@ internal Invocation CreateInvocation(InvocationVersionOne v1Invocation)
...
+                {
+                    FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
+                    PhysicalLocation = CreatePhysicalLocation(v1Location.ResultFile),
+                    Properties = v1Location.Properties
+                };
+
+                if (_currentRun.LogicalLocations.ContainsKey(location.FullyQualifiedLogicalName))
+                {
+                    _currentRun.LogicalLocations[location.FullyQualifiedLogicalName].DecoratedName = v1Location.DecoratedName;
+                }
```
lgolding: If the entry doesn't exist, we should create it, right?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: There is an overload that takes the uri and uribaseid

lgolding: ok.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: When you do that, let's make the properties of the region consistent with the length of the snippet.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: The logicalLocations dictionary has been imported at this point, wouldn't that definitely include it?

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: or locations[0] maybe

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Not necessarily.  `run.logicalLocations` is optional (in both v1 and v2). Now, there's nowhere else to put `decoratedName`, so if you have `decoratedName`, you have to make sure `run.logicalLocations` is present. Come to think of it, you need to check for the existence of the dictionary before adding to it. And I suspect you'll be doing that check-for-dictionary-and-create-if-necessary-and-then-add-entry multiple times, so you might want to factor it out.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Put the snippet in `locations[0].physicalLocation.region.snippet`, and make sure the properties of the `region` in your test log are consistent with the text of the `snippet`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Yeah I already have AddLogicalLocation which should work here.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -547,7 +553,19 @@ internal Result CreateResult(ResultVersionOne v1Result)
...
+                    {
+                        region.Snippet = new FileContent
+                        {
+                            Text = v1Result.Snippet
+                        };
+
+                        region.Length = v1Result.Snippet.Length;
+                        region.StartColumn = 1;
+                        region.EndColumn = region.Length + 1;
+                    }
```
lgolding: I'm sorry, I was unclear. I meant: "For the sake of the people who read the v1 sample file, make the region properties in the v1 sample file consistent with the snippet in the v1 sample file." When it comes to actually _copying_ the properties from v1 to v2, just take what they give you. If the v1 file was inconsistent, tough luck. (In fact, you _have_ to just copy the properties, for the sake of round-tripping (unless and until we adopt the "stuff the whole v1 run in the v2 run's property bag" trick).

EasyRhin0: ok sorry, should have known that :p

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -467,6 +511,63 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+                {
+                    // Only the formattedRuleMessage.formatId property is used above
+                    result.SetProperty($"{FromPropertyBagPrefix}/formattedRuleMessage", v1Result.FormattedRuleMessage);
+                }
+
+                if (!string.IsNullOrWhiteSpace(v1Result.Snippet))
+                {
+                    Region region = result.Locations?[0]?.PhysicalLocation?.Region;
+
+                    if (region != null)
```
lgolding: We have to store the snippet. If the region isn't there, create it. Likewise for its ancestors the physicalLocation object, the location object, and the locations array.

PR #901, opened 2018-05-21
Transform v1 fix, fileChange, replacement

EasyRhin0:
Please check on how I'm transforming arguments that correspond to rule messages, I'm a bit confused about how that works.

### DIFF
--- a/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
+++ b/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
@@ -1061,6 +1061,11 @@ namespace Microsoft.CodeAnalysis.Sarif
         {
           ""ruleId"": ""C2001"",
           ""level"": ""error"",
+          ""message"": {
+            ""arguments"": [
+              ""ptr""
+            ]
+          },
           ""ruleMessageId"": ""default"",
           ""analysisTarget"": {
             ""uri"": ""file:///home/buildAgent/src/collections/list.cpp""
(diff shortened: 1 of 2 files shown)


src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1061,6 +1061,11 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
         {
           ""ruleId"": ""C2001"",
           ""level"": ""error"",
+          ""message"": {
+            ""arguments"": [
+              ""ptr""
+            ]
+          },
```
lgolding: This is correct. The spec now explicitly addresses this case in : §3.19.7, "ruleMessageId property":

>If the message string identified by `ruleId` and `ruleMessageId` includes placeholders (§3.9.4), then `result.message.arguments` (§3.19.6, §3.9.11) **SHALL** contain the replacement values for the placeholders. In this situation, `result.message` will contain _only_ the `arguments` property.
>
> EXAMPLE 2: In this example, the message string identified by `ruleId` and `ruleMessageId` has a single placeholder `"{0}"`. `message.arguments holds` the replacement value `"counter"`.
```
{                                # A run object (§3.11).
  "results": [
    {                            # A result object (§3.19).
      "ruleId": "CA2101",
      "ruleMessageId": "default",
      "message": {
        "arguments": [
          "counter"
        ]
      }
    }
  ],

  "resources": {
    "rules": {
      "CA2101": {                # A rule object
        "messageStrings": {
          "default": "Variable \"{0}\" is uninitialized."
        }
      }
    }
  }
}
```

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                          ],
+                          ""fixes"": [
+                            {
+                              ""description"": ""Wrap attribute values in single quotes."",
+                              ""fileChanges"": [
+                                {
+                                  ""uri"": ""http://localhost:34420/HtmlFixes.html"",
+                                  ""replacements"": [
+                                    {
+                                      ""offset"": 623,
```
lgolding: For consistency: This should be 720., and the second offset should be 725.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                          ""formattedRuleMessage"": {
+                            ""formatId"": ""default"",
+                            ""arguments"": [
+                              ""shape""
+                            ]
+                          },
+                          ""suppressionStates"": [ ""suppressedExternally"" ],
+                          ""baselineState"": ""existing"",
+                          ""level"": ""error"",
+                          ""snippet"": ""<area alt=\""Here is some text\"" coords=\""10 20 20\"" href=\""moon.html\"" shape=circle xweb:fixindex=\""0\"" />"",
```
lgolding: For the sake of brevity, I suggest that the example log files include only those properties that are either (a) required, (b) necessary to demonstrate the feature, or (c) helpful for understanding. In this example, `suppressionState`, `baselineState`, and `level` are optional, so I'd omit them, `snippet` helps clarify the example, so I'd keep it even though it's optional.

Other candidates for omission:
- `tool.semanticVersion` (used to be required, but is now optional)
- `run.files`

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                            {
+                              ""analysisTarget"": {
+                                ""uri"": ""http://localhost:34420/HtmlFixes.html""
+                              },
+                              ""resultFile"": {
+                                ""uri"": ""http://localhost:34420/HtmlFixes.html"",
+                                ""region"": {
+                                  ""startLine"": 24,
+                                  ""startColumn"": 4,
+                                  ""endColumn"": 38,
```
lgolding: For consistency: `"endColumn": 7, "length": 3.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                          ],
+                          ""fixes"": [
+                            {
+                              ""description"": ""Convert tag name to lowercase."",
+                              ""fileChanges"": [
+                                {
+                                  ""uri"": ""http://localhost:34420/HtmlFixes.html"",
+                                  ""replacements"": [
+                                    {
+                                      ""offset"": 804,
```
lgolding: For consistency: `"offset": 803`.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Great thank you.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
 
             v2LogText.Should().Be(v2LogExpectedText);
         }
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
+        {
+            string v1LogText =
+              @"{
```
lgolding: Nit: In the v2Log below, you push the sample log against the left margin. I like that better. (Saves file size, marginally easier to read, definitely easier to type).

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                                }
+                              ]
+                            }
+                          ]
+		                }
+                      ],
+                      ""rules"": {
+                        ""WEB1079.AttributeValueIsNotQuoted"": {
+                          ""id"": ""WEB1079"",
+                          ""shortDescription"": ""The attribute value is not quoted."",
```
lgolding: Another example of an optional property that aids understanding of the sample -- good.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: This is actually real data. It highlights the entire <DIV ...></DIV>

EasyRhin0: I can still change it of course

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Let's just fix this one for now. I'd rather not lengthen the PR. Let's do it opportunistically.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Oh! Sorry. Leave it.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Yes, now that you've explained that the region includes the whole element, this makes sense.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: OK

EasyRhin0:
I just realized I didn't add the call to CreateFix. Kind of important for this change. :p

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                  ""length"": 5,
+                  ""snippet"": {
+                    ""text"": ""<area alt=\""Here is some text\"" coords=\""10 20 20\"" href=\""moon.html\"" shape=circle xweb:fixindex=\""0\"" />""
+                  }
+                }
+              }
+            }
+          ],
+          ""suppressionStates"": [""suppressedExternally""],
+          ""baselineState"": ""existing"",
```
lgolding: Bug: The fixes are not emitted at all!

lgolding:
I will wait for your next push to review the logic.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,282 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+        {
+          ""ruleId"": ""WEB1079"",
+          ""level"": ""error"",
+          ""message"": {
+            ""arguments"": [
+              ""shape""
+            ]
+          },
+          ""ruleMessageId"": ""default"",
+          ""analysisTarget"": {
```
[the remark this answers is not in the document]

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Will keep this in mind going forward.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,325 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+          },
+          {
+            ""ruleId"": ""WEB1066"",
+            ""formattedRuleMessage"": {
+              ""formatId"": ""default"",
+              ""arguments"": [
+                ""DIV""
+              ]
+            },
+            ""suppressionStates"": [ ""suppressedExternally"" ],
```
lgolding: Remove these three optional attributes.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1114,5 +1119,325 @@ public void SarifTransformerTests_ToCurrent_BasicResult()
...
+                ""text"": ""Wrap attribute value in double quotes.""
+              },
+              ""fileChanges"": [
+                {
+                  ""fileLocation"": {
+                    ""uri"": ""http://localhost:34420/HtmlFixes.html""
+                  },
+                  ""replacements"": [
+                    {
+                      ""offset"": 720,
```
lgolding: In v2, a `replacement` object looks like this (§3.40, replacement object):
```
"fileChanges": [
  {
    "deletedRegion": {
      "offset": 12,
      "length": 5
    },
    "insertedContent": {
      "binary": "ZXhhbXBsZQ=="
    }
  },
  {
    "deletedRegion": {
      "offset": 20,
      "length": 3
    }
  },
  {
    "deletedRegion": {
      "offset": 312,
      "length": 0
    },
    "insertedContent": {
      "binary": "ZXhhbXBsZQ=="
    }
  }
]
```
Are we not up to date on that schema change?

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Fixed, sorry missed the second set

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Must not be.

EasyRhin0: Confirmed. Want me to include those changes here?

EasyRhin0: Or I can do them separately and merge.

lgolding: Let's do this, in sequence:
1) I'll review this the way it is, and we'll merge it.
2) Separate PR: Update the schema and modify the fix transformation logic.

lgolding: Oh, and
    3. Look for other missing schema changes. I could've sworn we were up to date.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -519,21 +565,34 @@ internal Result CreateResult(ResultVersionOne v1Result)
             {
                 result = new Result
                 {
-                    AnalysisTarget = CreateFileLocation(v1Result.Locations?[0]?.AnalysisTarget),
```
lgolding: Bug: The fix here is not to remove `AnalysisTarget`. It's to populate it _only if it differs from the result file_.

lgolding: Oh! I see you did that later. ok.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -519,21 +565,34 @@ internal Result CreateResult(ResultVersionOne v1Result)
...
-                    RuleMessageId = v1Result.FormattedRuleMessage.FormatId,
                     Stacks = v1Result.Stacks?.Select(CreateStack).ToList(),
                     SuppressionStates = Utilities.CreateSuppressionStates(v1Result.SuppressionStates),
                 };
 
-                // TODO: transfer formatted rule messages
+                if (v1Result.Locations?[0]?.AnalysisTarget?.Uri != v1Result.Locations?[0]?.ResultFile?.Uri)
+                {
+                    result.AnalysisTarget = CreateFileLocation(v1Result.Locations[0].AnalysisTarget);
+                }
```
lgolding: This is worth a comment: "The spec says that analysisTarget is required only if it differs from the result file."

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -415,6 +444,23 @@ internal Notification CreateNotification(NotificationVersionOne v1Notification)
...
+        {
+            Replacement replacement = null;
+
+            if (v1Replacement != null)
+            {
+                replacement = new Replacement
+                {
+                    DeletedLength = v1Replacement.DeletedLength,
+                    InsertedBytes = v1Replacement.InsertedBytes,
+                    Offset = v1Replacement.Offset
```
lgolding: In the next PR, this is the only part of the converter logic that will need to change.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Right, it's handled conditionally down after the result creation

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: OK

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Sounds good. I'll take 5 minutes to look back and see if it was ever changed,

lgolding:
(oh, right, and removing that second set of optional properties)

EasyRhin0:
Yup got those too. Off to lunch now...

PR #903, opened 2018-05-22
Schema change #97 replacement object

EasyRhin0:


src/Sarif.Viewer.VisualStudio/Sarif/Replacement.extensions.cs
```diff
@@ -19,15 +19,18 @@ public static ReplacementModel ToReplacementModel(this Replacement replacement)
...
+            if (!string.IsNullOrWhiteSpace(replacement.InsertedContent?.Text))
             {
-                model.InsertedString = Encoding.UTF8.GetString(Convert.FromBase64String(replacement.InsertedBytes));
-                model.InsertedBytes = Encoding.UTF8.GetBytes(model.InsertedString);
+                model.InsertedString = replacement.InsertedContent.Text;
+            }
+            else if (replacement.InsertedContent?.Binary != null)
+            {
+                model.InsertedBytes = Convert.FromBase64String(replacement.InsertedContent.Binary);
+                model.InsertedString = Encoding.UTF8.GetString(model.InsertedBytes);
```
lgolding: If the content is expressed in binary, there's no guarantee that decoding it into a string is sensible. And if it is, you don't know the encoding. If you wanted to be fancy you could check for the file's encoding (I don't know if you have access to `file.encoding` here), but otherwise, you can't populate `InsertedString` (and the rest of the Viewer needs to be ok with it being absent).

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1314,12 +1314,22 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
                   },
                   ""replacements"": [
                     {
-                      ""offset"": 720,
-                      ""insertedBytes"": ""Jw==""
+                      ""deletedRegion"": {
+                        ""offset"": 720,
+                        ""length"": 1
```
lgolding: Given the contents of the v1 `replacement` object, the length of the region is 0. The region expresses an "insertion point".

lgolding: ... and yet the test is passing, so something else must be wrong (presumably it would fail if you changed this "1" to the (correct) value of 0).

src/Sarif.Viewer.VisualStudio/Sarif/Replacement.extensions.cs
EasyRhin0: Understood, fixed.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -450,12 +450,28 @@ internal Replacement CreateReplacement(ReplacementVersionOne v1Replacement)
...
+                }
+
+                if (v1Replacement.DeletedLength > 0 || insertedLength > 0)
+                {
+                    replacement.DeletedRegion = new Region
+                    {
+                        Length = insertedLength > 0 ? insertedLength : v1Replacement.DeletedLength,
+                        Offset = v1Replacement.Offset
+                    };
+                }
```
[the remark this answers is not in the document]
EasyRhin0: Got it. Was misunderstanding the consumer logic, thanks for clarifying.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -451,27 +451,20 @@ internal Replacement CreateReplacement(ReplacementVersionOne v1Replacement)
...
                 {
-                    replacement.DeletedRegion = new Region
-                    {
-                        Length = insertedLength > 0 ? insertedLength : v1Replacement.DeletedLength,
-                        Offset = v1Replacement.Offset
-                    };
-                }
+                    Length = v1Replacement.DeletedLength,
+                    Offset = v1Replacement.Offset
+                };
```
lgolding: Yes.

PR #904, opened 2018-05-24
Transform code flows

EasyRhin0:
Process result.relatedLocations
Remove result.snippet

### DIFF
--- a/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
+++ b/src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
@@ -1011,6 +1011,19 @@ namespace Microsoft.CodeAnalysis.Sarif
                               ""fullyQualifiedLogicalName"": ""collections::list::add"",
                               ""decoratedName"": ""?add@list@collections@@QAEXH@Z""
                             }
+                          ],
+                          ""relatedLocations"": [
+                            {
+                              ""message"": ""\""count\"" was declared here."",
+                              ""physicalLocation"": {
+                                ""uri"": ""file:///home/buildAgent/src/collections/list.h"",
+                                ""region"": {
+                                  ""startLine"": 8,
+                                  ""startColumn"": 5
+                                }
+                              },
+                              ""fullyQualifiedLogicalName"": ""collections::list:add""
+                            }
                           ]
 		                }
                       ],
@@ -1088,6 +1101,23 @@ namespace Microsoft.CodeAnalysis.Sarif
                 }
               },
               ""fullyQualifiedLogicalName"": ""collections::list::add""
+            }
+          ],
+          ""relatedLocations"": [
+            {
+              ""physicalLocation"": {
+                ""fileLocation"": {
+                  ""uri"": ""file:///home/buildAgent/src/collections/list.h""
+                },
+                ""region"": {
+                  ""startLine"": 8,
+                  ""startColumn"": 5
+                }
+              },
+              ""fullyQualifiedLogicalName"": ""collections::list:add"",
+              ""message"": {
+                ""text"": ""\""count\"" was declared here.""
+              }
             }
           ],
           ""suppressionStates"": [""suppressedExternally""],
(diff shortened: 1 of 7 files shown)


EasyRhin0:
Notes:
I'm putting v1 AnnotatedCodeLocationKind in the new codeFlowLocation.kind property
I'm stashing the "essential" property even though it's obsolete

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+            string v1LogText =
+@"{
+  ""version"": ""1.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""logicalLocations"": {
```
lgolding: I don't think we need v1 `logicalLocations` to test v1 => v2 code flows.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+        },
+        ""collections"": {
+          ""name"": ""collections"",
+          ""kind"": ""namespace""
+        }
+      },
+      ""results"": [
+        {
+          ""ruleId"": ""C2001"",
+          ""formattedRuleMessage"": {
```
lgolding: We don't need rule metadata to test code flows. Just use `result.message`.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: The locations have fullyQualifiedLogicalNames

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: OK

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+                              ""text"": ""int *ptr;""
+                            }
+                          }
+                        },
+                        ""fullyQualifiedLogicalName"": ""collections::list:add"",
+                        ""message"": {
+                          ""text"": ""Variable \""ptr\"" declared.""
+                        }
+                      },
+                      ""kind"": ""Declaration"",
```
lgolding: Even though the v2 `codeFlowLocation` has a property `kind`, it doesn't mean the same thing as the v1 `annotatedCodeLocation.kind`. I can explain if you're interested. Put this in the property bag.

EasyRhin0: OK

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -192,5 +192,17 @@ public static BaselineState CreateBaselineState(BaselineStateVersionOne v1Baseli
...
+        public static CodeFlowLocationImportance CreateCodeFlowLocationImportance(AnnotatedCodeLocationImportanceVersionOne v1AnnotatedCodeLocationImportance)
+        {
+            switch (v1AnnotatedCodeLocationImportance)
+            {
+                case AnnotatedCodeLocationImportanceVersionOne.Essential:
+                    return CodeFlowLocationImportance.Essential;
+                case AnnotatedCodeLocationImportanceVersionOne.Important:
+                    return CodeFlowLocationImportance.Important;
+                default:
+                    return CodeFlowLocationImportance.Unimportant;
```
lgolding: The default for v1's `annotatedCodeLocation.importance` property was `important`.

(The same is true in v2, but that's neither here nor there, since you're processing a v1 value here.)

EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -33,6 +35,85 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne v1
...
+                };
+
+                if (v1CodeFlow.Locations != null && v1CodeFlow.Locations.Count > 0)
+                {
+                    _codeFlowLocationNestingLevel = 0;
+
+                    if (v1CodeFlow.Locations[0].Step == 0)
+                    {
+                        // If the steps are zero-based, add 1 to comply with the v2 spec
+                        _codeFlowLocationStepAdjustment = 1;
```
lgolding: Nice! 😄 

But (I hate to be a killjoy 😄 ): Bug: This means that `step` will not round-trip. If it was 0-based on the way in from v1 to v2, you'll have forgotten about that when you convert back to v1. You need to put the v1 `step` in the property  bag.

... as usual, unless and until we get tired of worrying about this stuff and stick the whole v1 run in the v2 run's property bag.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Sure, but why does v1 `logicalLocations` have to exist? You'll create v2 `logicalLocations` when you notice the FQLNs in the code flow.

lgolding: The spec says that _if_ `run.logicalLocations` exists, _then_ `location.FQLN` must match one of its property names. But the spec does not say that `run.logicalLocations` has to exist.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -33,6 +35,85 @@ public override SarifLogVersionOne VisitSarifLogVersionOne(SarifLogVersionOne v1
...
+                };
+
+                if (v1AnnotatedCodeLocation.Kind == AnnotatedCodeLocationKindVersionOne.Call)
+                {
+                    _codeFlowLocationNestingLevel++;
+                }
+                else if (v1AnnotatedCodeLocation.Kind == AnnotatedCodeLocationKindVersionOne.CallReturn)
+                {
+                    _codeFlowLocationNestingLevel--;
+                }
```
lgolding: Good.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -263,6 +344,40 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
+        internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCodeLocation)
+        {
+            Location location = null;
+
+            if (v1AnnotatedCodeLocation != null)
+            {
+                location = new Location
+                {
+                    Annotations = v1AnnotatedCodeLocation.Annotations?.SelectMany(a => a.Locations, (a, pl) => CreateRegion(pl)).ToList(),
+                    FullyQualifiedLogicalName = v1AnnotatedCodeLocation.LogicalLocationKey ?? v1AnnotatedCodeLocation.FullyQualifiedLogicalName,
```
lgolding: Good.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+                      ""module"": ""platform"",
+                      ""importance"": ""essential"",
+                      ""properties"": {
+                        ""sarifv1/annotations"": null,
+                        ""sarifv1/essential"": false,
+                        ""sarifv1/fullyQualifiedLogicalName"": ""collections::list:add"",
+                        ""sarifv1/id"": 0,
+                        ""sarifv1/logicalLocationKey"": null,
+                        ""sarifv1/taintKind"": 0,
```
lgolding: Bug: This won't round-trip. v1 `taintKind` was a string with enumerated values `"source"`, `"sink"`, `"sanitizer"`.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+                        ""sarifv1/target"": null,
```
lgolding: We need to test that this works for non-null values of `target`, `targetKey`, and `values`.

lgolding: ... unless and until we put the whole run in the property bag! I'm going to ping Michael on this.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,336 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+                        ""sarifv1/targetKey"": null,
+                        ""sarifv1/threadId"": 52,
+                        ""sarifv1/values"": null
```
lgolding: Bug: You need to bag v1  `annotatedCodeLocation.state` and `targetLocation` as well.

Until Michael lets us off the hook 😄.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: I'm sorry, I'm sure we went through this once before, but doesn't the presence of a FQLN imply a corresponding v1 logicalLocation?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -263,6 +344,40 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
 
+        internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCodeLocation)
+        {
+            Location location = null;
+
+            if (v1AnnotatedCodeLocation != null)
+            {
+                location = new Location
+                {
+                    Annotations = v1AnnotatedCodeLocation.Annotations?.SelectMany(a => a.Locations, (a, pl) => CreateRegion(pl)).ToList(),
```
lgolding: Oh, wow, very nice. I had to look it up to remind myself that v1 `location.annotations` is of type `region[]`.

But there's a bug. The v1 `annotation`  object contained a message, and a set of locations to which the message applies. It was a terrible design for reasons I can go into offline if you like.

The point is, in the new design -- where the `region` object has a `message` property -- that v1 `annotatedCodeLocation.message` needs to go on each of the v2 `region`s you create. As the code stands, the message is lost.

We need to enhance the tests to detect this (by adding a couple of v1 `annotations` to one of the v1 `annotatedCodeLocation`s.

EasyRhin0: Gotcha

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: No. If you want to tell the consumer more about the logical location whose FQLN is `location.FQLN`, you can create the `run.logicalLocations` dictionary,  add a property whose name is the value of `FQLN`, and set the property value to a `logicalLocation` object with the additional information. But you didn't have to do it in v1, and you still don't have to do it in v2.

lgolding: (Don't worry about asking the same question multiple times. I'm notorious for doing that -- just ask Michael 😄.)

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: ... or, don't bother doing that now, since we're not going to do this piecemeal bagging any more.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: State transfers to v2

EasyRhin0: No v1 targetLocation property

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: Ah right. We need a test for that.

lgolding: The "ah right, we need a test" was about `state`.

As to `targetLocation`, [are you sure](https://rawgit.com/sarif-standard/sarif-spec/master/Static%20Analysis%20Results%20Interchange%20Format%20(SARIF).html#annotatedcodelocation-target)?

EasyRhin0: [Yes](https://github.com/oasis-tcs/sarif-spec/raw/master/Documents/WorkingDrafts/sarif-v1.0-wd01.docx)
[Schema](https://github.com/Microsoft/sarif-sdk/blob/sarif-v2/src/Sarif/Schemata/Sarif-1.0.0.schema.json)

lgolding: Well, ok, that was a bug in the v1 schema. It didn't implement the v1 spec.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Clear, thanks

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,337 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+                      ""module"": ""platform"",
+                      ""importance"": ""essential"",
+                      ""properties"": {
+                        ""sarifv1/annotations"": null,
+                        ""sarifv1/essential"": false,
+                        ""sarifv1/fullyQualifiedLogicalName"": ""collections::list::add"",
+                        ""sarifv1/id"": 0,
+                        ""sarifv1/kind"": ""Declaration"",
+                        ""sarifv1/logicalLocationKey"": null,
+                        ""sarifv1/taintKind"": ""Sink"",
```
lgolding: Bug: The value in the v1 file was `"sink"`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -263,6 +346,43 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
 
+        internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCodeLocation)
+        {
+            Location location = null;
+
+            if (v1AnnotatedCodeLocation != null)
+            {
+                location = new Location
+                {
+                    Annotations = v1AnnotatedCodeLocation.Annotations?.SelectMany(a => a.Locations, (a, pl) => CreateRegion(pl, a.Message)).ToList(),
```
lgolding: There is a problem here caused by the poor design of the v1 `annotation` object. You have a `location` object that refers to a file, say _file1.c_. The v1 `annotation` object contained an array of `physicalLocation` objects -- which meant that an annotation on _file1.c_  could refer to a location in _file2.c_. That makes no sense. That's why we changed the design and got rid of the `annotation` object. In v2, annotations can only refer to regions in the location's own file.

So what do you do if you encounter a v1 file that makes use of this senseless feature (as your test example does)? The best you can do is
1. Copy the whole set of v1 annotations to the property bag. You're doing that.
2. Populate the v2 `location.annotations` property with _only those annotations that refer to the file specified by_ `location.physicalLocation`.

You'll never encounter this; AFAIK no existing v1 producer uses annotations at all, still less in this senseless manner. But the code may as well be prepared for it.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1453,5 +1483,337 @@ public void SarifTransformerTests_ToCurrent_TwoResultsWithFixes()
...
+  ""version"": ""1.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""results"": [
+        {
+          ""message"": ""C2001"",
```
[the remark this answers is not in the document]

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Changed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

PR #906, opened 2018-05-29
Transform v2 -> v1 invocation & notification

EasyRhin0:
Plus exceptionData, physicalLocation, and test cleanup


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -32,6 +32,45 @@ private static void VerifyCurrentToVersionOneTransformation(string v2LogText, st
...
+  ""version"": ""2.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""results"": [],
+      ""properties"": {
+        ""sarifv1/run"": ""{\""tool\"":{\""name\"":\""CodeScanner\"",\""semanticVersion\"":\""2.1.0\""},\""results\"":[]}""
```
lgolding: Without having yet read the code, it appears that when you persist a v1 `run` to the v2 `run`'s property bag, you compress out all the white space. When you extract the v1 `run` into its own log file, you pretty-print it, using `"\r\n"` as the newline sequence, and two-space indentation.

Another possible behavior would be to retain the original file's white space. The disadvantages of that approach are:
1. Increased v2 log file size.
2. Possible inconsistency between the `run` (which might use one whitespace convention) and the containing log file (which might use a different convention):
```
{
  "$schema": "",
  "version": "1.0.0",
  "runs": [                    # The v2-to-v1 converter generates everything up to here...
    {                              # ... but the extracted v1 run object, starting here, might use a different white space convention.
    }
  ]
}
```

All in all, I think you made the right choice, even though it doesn't preserve whitespace. But I'd like @michaelcfanning  to weigh in.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -389,11 +427,315 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithRules()
...
       }
     }
   ]
 }";
+
+            VerifyCurrentToVersionOneTransformation(V2LogText, V1LogExpectedText);
+        }
+
+        [Fact]
+        public void SarifTransformerTests_ToVersionOne_OneRunWithNotificationsButNoInvocations()
```
lgolding: It's not a great test name, because it _does_ contain an `invocation` object. But IMO we don't need this test. The previous test covered the ground.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -84,24 +103,136 @@ public HashVersionOne CreateHash(Hash v2Hash)
...
+                };
+
+                if (v2Invocation.ConfigurationNotifications != null)
+                {
+                    if (_currentRun.ConfigurationNotifications == null)
+                    {
+                        _currentRun.ConfigurationNotifications = new List<NotificationVersionOne>();
+                    }
+
+                    List<NotificationVersionOne> notifications = v2Invocation.ConfigurationNotifications.Select(CreateNotification).ToList();
```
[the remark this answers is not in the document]
EasyRhin0: Right, fixed.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -140,36 +271,54 @@ public RunVersionOne CreateRun(Run v2Run)
 
                 if (v2Run.TryGetProperty("sarifv1/run", out serializedV1Run))
                 {
-                    run = JsonConvert.DeserializeObject<RunVersionOne>(serializedV1Run, Utilities.JsonSettingsV1);
+                    run = JsonConvert.DeserializeObject<RunVersionOne>(Regex.Unescape(serializedV1Run), Utilities.JsonSettingsV1Compact);
```
lgolding: Bug: The call to`Regex.Unescape` isn't right.

You put the run into the property bag by calling `JsonConvert.SerializeObject`, which escapes the set of characters that need to be escaped in a JSON string. Then you take it out of the property bag by calling `Regex.Unescape`, which unescapes the set of characters that need to be escaped in a regex.

But those are two different sets of characters.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -140,36 +271,54 @@ public RunVersionOne CreateRun(Run v2Run)
...
                 }
             }
 
             return run;
         }
 
-        public ToolVersionOne CreateTool(Tool v2Tool)
+        internal void RemoveResponseFilesFromFileDictionary()
+        {
+            if (_currentRun.Invocation?.ResponseFiles != null)
```
lgolding: This is one of those things that is almost always ok, but could conceivably not be. There's no harm in leaving the file dictionary entries for the response files, in case it's needed for some other reason.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Since it's double-encoded, I don't know how to get back to a valid json string. Using Regex.Unescape was the approach recommended by others who were doing the same thing.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: OK. I did this because they are not there going the other direction.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Could you explain what you mean by "double-encoded"?

EasyRhin0: "Escaped" is what I really mean. Without the Unescape call, the quotes are still escaped with backslashes in the string obtained by TryGetProperty..

lgolding: We need to think more about this. By my reasoning above (different sets of escaped characters), this approach is just wrong, and you're unlucky that it works most of the time (unlucky, because it won't fail for months or years, and then we'll struggle to figure out why).

Code that is _almost_ correct is more dangerous than code that's obviously wrong and fails immediately ;-)

EasyRhin0: Yes, for this reason I spent a good deal of time trying to find the right approach. I'm all ears :-)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Removed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Changed to this:
    RunVersionOne run = null;
    if (v2Run.TryGetProperty<RunVersionOne>("sarifv1/run", out run))
    {
        return run;
    }

Results in this exception:
Newtonsoft.Json.JsonSerializationException : Error converting value "{"tool":{"name":"CodeScanner","semanticVersion":"2.1.0"},"results":[]}" to type 'Microsoft.CodeAnalysis.Sarif.VersionOne.RunVersionOne'. Path '', line 1, position 84.
---- System.ArgumentException : Could not cast or convert from System.String to Microsoft.CodeAnalysis.Sarif.VersionOne.RunVersionOne.

lgolding: I wouldn't expect that to work, because your property value is a string, not a `run` object.

EasyRhin0: I guess I could store the object itself?
      "properties": {
        "sarifv2/run": {"tool":{"name":"AssetScanner","semanticVersion":"1.7.2"},"results":[]}
      }

lgolding: You mean, as opposed to storing a serialized version of the object? That might work. Why did we originally decide to store the serialized string? It was probably to preserve whitespace on round-tripping. But if we're not doing that anyway, then this might work.

EasyRhin0: It just seemed natural to store it serialized.
This seems to be working. The only thing I'm seeing is enums are emitted as their integer value rather than their string name.

EasyRhin0: But they deserialize correctly

michaelcfanning: Here's my stupid suggestion, base64 encode the string. Does it really matter to provide a 'readable' embedded JSON file in the property bag? Probably not. This definitely sidesteps all the horrible escaping issues implied by what we're attempting here.

lgolding: @EasyRhin0 @michaelcfanning This exploration looks very promising! I'll have to sign off for the evening in a few minutes, but I'll look in tomorrow and see where we landed. Good work!

lgolding: (BTW, you are absolutely right that it doesn't matter how enum values are stored in the property bag, as long as they round-trip correctly.)

lgolding: And I'm intrigued by Michael's base64 idea. It does indeed side-step all the escaping/unescaping nonsense -- but at a cost of doubling (is that right?) the size of the bagged property. Really I like Chris's "store the object, not the serialized string" approach best. Its only downsize is not preserving whitespace, which I've given up on :-)

EasyRhin0:
Looking into some weird behavior in four tests. In VS, they fail when I run all tests, but then succeed when I re-run. The difference is output of enums ("error" vs. 2), and the datetime thing we saw a few months back (.000Z).

Issue #907, opened 2018-05-31
Tranformer APIs need refactoring


EasyRhin0:
Need a factory and a common interface between SarifLog and SarifLogVersionOne. We should also think about whether the visitor model is the best choice.

EasyRhin0:
And update multitool accordingly

PR #908, opened 2018-05-31
Add transform verb to multitool + cleanup

EasyRhin0:
The transformer API model is yucky; filed issue #907 to refactor.

### DIFF
--- a/src/Sarif.Multitool/AbsoluteUriCommand.cs
+++ b/src/Sarif.Multitool/AbsoluteUriCommand.cs
@@ -33,7 +33,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
             }
             catch (Exception ex)
             {
-                Console.WriteLine(ex.ToString());
+                Console.WriteLine(ex);
                 return 1;
             }
 
--- a/src/Sarif.Multitool/AbsoluteUriCommand.cs
+++ b/src/Sarif.Multitool/AbsoluteUriCommand.cs
@@ -4,8 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using Microsoft.CodeAnalysis.Sarif.Driver;
 using Microsoft.CodeAnalysis.Sarif.Processors;
+using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
@@ -25,8 +25,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                     // Write output to file.
                     string outputName = sarifLog.GetOutputFileName(absoluteUriOptions);
                     var formatting = absoluteUriOptions.PrettyPrint
-                        ? Newtonsoft.Json.Formatting.Indented
-                        : Newtonsoft.Json.Formatting.None;
+                        ? Formatting.Indented
+                        : Formatting.None;
 
                     MultitoolFileHelpers.WriteSarifFile(sarifLog.Log, outputName, formatting);
                 }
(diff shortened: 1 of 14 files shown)


src/Sarif.Multitool/CommandUtilities.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    internal static class CommandUtilities
+    {
+        internal static string GetTransformedOutputFileName(SingleFileOptionsBase options)
```
lgolding: Note to self: See if there are unit tests for this method.

src/Sarif.Multitool/AbsoluteUriCommand.cs
```diff
@@ -46,7 +46,7 @@ private static IEnumerable<AbsoluteUriFile> GetSarifFiles(AbsoluteUriOptions abs
             HashSet<string> fileNames = MultitoolFileHelpers.CreateTargetsSet(absoluteUriOptions.TargetFileSpecifiers, absoluteUriOptions.Recurse);
             foreach (var file in fileNames)
             {
-                yield return new AbsoluteUriFile() { FileName = file, Log = MultitoolFileHelpers.ReadSarifFile(file) };
+                yield return new AbsoluteUriFile() { FileName = file, Log = MultitoolFileHelpers.ReadSarifFile<SarifLog>(file) };
```
lgolding: I see that in MultitoolFileHelper.cs, you changed the API so that the generic parameter in the call to `JsonConvert.DeserializeObject`, which was formerly encapsulated, is now exposed to the caller. What is the advantage of this?

lgolding: Never mind, I see that it's so you can use it for `SarifLogVersionOne` as well.

src/Sarif.Multitool/Program.cs
```diff
@@ -1,11 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
```
lgolding: Thanks for cleanup!

src/Sarif.Multitool/CommandUtilities.cs
lgolding: Ok, I see there are none. Please add.

lgolding: Oh... this is an existing method that moved from one file to another. Never mind.

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+                    SarifLog actualLog = MultitoolFileHelpers.ReadSarifFile<SarifLog>(transformOptions.InputFilePath);
+                    var visitor = new SarifCurrentToVersionOneVisitor();
+                    visitor.VisitSarifLog(actualLog);
+
+                    MultitoolFileHelpers.WriteSarifFile(visitor.SarifLogVersionOne, fileName, formatting);
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex.ToString());
```
lgolding: `ToString` is not needed.

EasyRhin0: Fixed everywhere

src/Sarif.Multitool/CommandUtilities.cs
EasyRhin0: I'll bug this and use it to fill downtime...

EasyRhin0: #909

PR #911, opened 2018-05-31
Transform v2 -> v1 stack, stackFrame, region

EasyRhin0:
Please check the logicalLocations (aka the cold trail from hell), I'm not confident there.

### DIFF
--- a/src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
+++ b/src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
@@ -854,7 +854,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                   ""uri"": ""file:///C:/src/main.cs"",
                   ""module"": ""RuleLibrary"",
                   ""threadId"": 52,
-                  ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Register-0"",
+                  ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Register"",
+                  ""logicalLocationKey"": ""Rules.SecureHashAlgorithmRule.Register-0"",
                   ""address"": 1002485
                 },
                 {
@@ -869,7 +870,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                   ""uri"": ""file:///C:/src/foobar.cs"",
                   ""module"": ""ExecutionEngine"",
                   ""threadId"": 52,
-                  ""fullyQualifiedLogicalName"": ""ExecutionEngine.Engine.FooBar"",
+                  ""fullyQualifiedLogicalName"": ""ExecutionEngine.Engine.EvaluateRule"",
+                  ""logicalLocationKey"": ""ExecutionEngine.Engine.FooBar"",
                   ""address"": 10073356,
                   ""offset"": 10475
                 }
--- a/src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
+++ b/src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
@@ -667,5 +667,231 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             VerifyCurrentToVersionOneTransformation(V2LogText, V1LogExpectedText);
         }
+
+        [Fact]
+        public void SarifTransformerTests_ToVersionOne_NotificationExceptionWithStack()
+        {
+            const string V2LogText =
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""invocations"": [
+        {
+          ""toolNotifications"": [
+            {
+              ""id"": ""CTN0001"",
+              ""message"": {
+                ""text"": ""Unhandled exception.""
+              },
+              ""level"": ""error"",
+              ""exception"": {
+                ""kind"": ""ExecutionEngine.RuleFailureException"",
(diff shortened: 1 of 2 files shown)


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -667,5 +667,231 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithInvocationAndNotificati
...
+                  ""frames"": [
+                    {
+                      ""location"": {
+                        ""physicalLocation"": {
+                          ""fileLocation"": {
+                            ""uri"": ""file:///C:/src/main.cs""
+                          },
+                          ""region"": {
+                            ""startLine"": 15,
+                            ""startColumn"": 9
```
lgolding: @michaelcfanning @EasyRhin0 

Chris, you have no action for now, but be aware:

At TC #&#xFEFF;18 on May 30th, we approved changes to the default values of missing region properties. In particular, if `endColumn` is missing, in v1 that meant `endColumn` == `startColumn` (an insertion point). But in v2, `endColumn` defaults to “one character past EOL (excluding the newline sequence”).

The reason for this change is that one of the TC members argued persuasively that the most natural interpretation of the region `{ "startLine": 3 }` is "The whole of line 3" and not "An insertion point at the beginning of line 3."

So when you get around to making the schema changes for oasis-tcs/sarif-spec#93, you'll need to adjust the transformations as well.

EasyRhin0: OK thanks

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -667,5 +667,231 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithInvocationAndNotificati
...
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-1.0.0"",
+  ""version"": ""1.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
+      },
+      ""invocation"": {},
```
lgolding: Don't emit empty object.

EasyRhin0: This happens when there are notifications but no other properties set on the invocation.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -667,5 +667,231 @@ public void SarifTransformerTests_ToVersionOne_OneRunWithInvocationAndNotificati
...
+                  ""module"": ""RuleLibrary"",
+                  ""threadId"": 52,
+                  ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Evaluate"",
+                  ""address"": 10092852
+                },
+                {
+                  ""uri"": ""file:///C:/src/main.cs"",
+                  ""module"": ""RuleLibrary"",
+                  ""threadId"": 52,
+                  ""fullyQualifiedLogicalName"": ""Rules.SecureHashAlgorithmRule.Register-0"",
```
lgolding: In v2, the `stackFrame.fullyQualifiedLogicalName` property is always the key into the files dictionary (if the files dictionary exists). So you have to do this:
```
If the files dictionary exists
    AND it has a key that matches v2 stackFrame.fullyQualifiedLogicalName
    AND the file object specified by that key has a fullyQualifiedName property
THEN
    set v1 stackFrame.fullyQualifiedLogicalName to v2 file.fullyQualifiedName
    set v1 stackFrame.logicalLocationKey to v2 stackFrame.fullyQualifiedLogicalName
ELSE
    set v1 stackFrame.fullyQualifiedLogicalName to v2 stackFrame.fullyQualifiedLogicalName
```

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -204,6 +206,43 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(PhysicalLocation v2Ph
...
+                                      v2Region.EndColumn,
+                                      v2Region.EndLine,
+                                      v2Region.Length,
+                                      v2Region.Offset);
+            }
+
+            return region;
+        }
+
+        internal RegionVersionOne CreateRegion(int startColumn, int startLine, int endColumn = 0, int endLine = 0, int length = 0, int offset = 0)
```
lgolding: Why do you need this overload?

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Maybe. I'd have to check all of the strings for null/whitespace, and the processId == 0, and the datetimes == datetime.minvalue. Would that be deterministic?

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Copied from the other transformer, where I do need it. Not needed here yet, will remove.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
lgolding: Yes, but now that you spell it out, not worth the trouble. After all, if we're copying an entire run to the property bag, I've got now reason to gripe about this. Let it go.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Fixed. This actually clarifies it quite a bit, thanks

PR #913, opened 2018-06-01
Transform v2 -> v1 result, fileChange, fix, replacement

EasyRhin0:


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -895,5 +895,268 @@ public void SarifTransformerTests_ToVersionOne_NotificationExceptionWithStack()
...
+  ""version"": ""1.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner""
+      },
+      ""results"": [
+        {
+          ""ruleId"": ""WEB1079"",
+          ""fixes"": [
```
lgolding: The v1 equivalent of v2's `result.message` + `result.ruleMessageId` would be `result.formattedRuleMessage`, which I don't see here.

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -176,7 +162,22 @@ public static ResultLevel CreateResultLevel(ResultLevelVersionOne v1ResultLevel)
             }
         }
 
-        public static ResultLevelVersionOne CreateResultLevel(ResultLevel v2ResultLevel)
+        public static ResultLevelVersionOne CreateResultLevelVersionOne(RuleConfigurationDefaultLevel v2DefaultLevel)
```
lgolding: Just curious why you moved it down in the file. Alphabetizing methods?

EasyRhin0: Yep

EasyRhin0: Err, I like having the v2 version above the v1 version

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -231,6 +264,28 @@ internal RegionVersionOne CreateRegion(Region v2Region)
...
+            ReplacementVersionOne replacement = null;
+
+            if (v2Replacement != null)
+            {
+                replacement = new ReplacementVersionOne();
+
+                if (v2Replacement.InsertedContent != null)
+                {
+                    replacement.InsertedBytes = v2Replacement.InsertedContent.Text != null ?
+                                                    SarifUtilities.GetUtf8Base64String(v2Replacement.InsertedContent.Text) :
```
lgolding: This is a problem because we don't know the encoding of the file being modified, so we don't know if it's ok to stick a bunch of UTF-8 bytes into it.

I know we've run into this before -- do you remember where, and what we did about it?

EasyRhin0: I think it was in one of the schema updates, I'll snoop around.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -908,6 +908,34 @@ public void SarifTransformerTests_ToVersionOne_TwoResultsWithFixes()
...
         ""name"": ""CodeScanner""
       },
+      ""files"": {
+        ""http://localhost:34420/HtmlFixes.html"": {
+          ""length"": 43,
+          ""enoding"": ""utf-8"",
+          ""mimeType"": ""text/plain"",
+          ""contents"": {
+            ""text"": ""The quick brown fox jumps over the lazy dog"",
+            ""binary"": ""VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw==""
```
lgolding: `"binary"` property is not needed for this test.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -264,19 +275,36 @@ internal RegionVersionOne CreateRegion(Region v2Region)
...
 
-                if (v2Replacement.InsertedContent != null)
+                if (insertedContent != null)
                 {
-                    replacement.InsertedBytes = v2Replacement.InsertedContent.Text != null ?
-                                                    SarifUtilities.GetUtf8Base64String(v2Replacement.InsertedContent.Text) :
-                                                    v2Replacement.InsertedContent.Binary;
+                    if (insertedContent.Text != null)
+                    {
+                        FileData fileData = _currentV2Run.Files[uri.OriginalString];
```
lgolding: There might not be an entry for this file in the `run.files` dictionary, and if there is, it might not have an `encoding` property. The logic needs to be:
```
IF the run.files dictionary contains an entry for this file
    AND the entry contains an encoding property
    AND the encoding is known to .NET
THEN
    Compute the inserted bytes
ELSE
    throw
END
```

lgolding: And given that you might not know the encoding at all, I propose `UnknownEncodingException` instead of `UnsupportedEncodingException`. Your call.

lgolding: Really, this calls for four unit tests, one for each case:
1. There is no dictionary entry.
2. There is a dictionary entry but it has no encoding.
3. It has an encoding but .NET doesn't recognize it.
4. It has a recognized encoding.

You test currently covers #&#xFEFF3; and #&#xFEFF4;.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: I agree that an exception is appropriate, and I agree with your deferring the reporting. But I don't want to lose track of it; could you file an issue?

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -51,6 +53,23 @@ internal ExceptionDataVersionOne CreateExceptionData(ExceptionData v2ExceptionDa
...
 
+        internal FileChangeVersionOne CreateFileChange(FileChange v2FileChange)
+        {
+            FileChangeVersionOne fileChange = null;
+
+            if (v2FileChange != null)
+            {
+                fileChange = new FileChangeVersionOne
+                {
+                    Replacements = v2FileChange.Replacements?.Select(r => CreateReplacement(r, v2FileChange.FileLocation?.Uri)).ToList(),
```
lgolding: As the code stands, you would recompute the encoding for each replacement, even though the encoding is the same for all replacements in the file. This suggests that `CreateReplacement` isn't logically the right place for that computation. Instead, I'd do the following:

- Introduce a helper that implements the logic I described in a previous comment. Note that it swallows the exception for an unrecognized encoding and just returns `null`, because we don't yet know if we will _need_ the encoding.
```
private static Encoding GetTextEncoding(Uri uri)
{
    Encoding encoding = null;

    FileData fileData;
    if (uri != null
        && currentV2Run.Files.TryGetValue(uri.OriginalString, out fileData)
        && fileData.Encoding != null)
    {
        try {
            encoding = Encoding.GetEncoding(fileData.Encoding);
        }
        catch (ArgumentException)
        {
        }
    }

    return encoding
}
```
- Call `GetTextEncoding` from at the top of `CreateFileChange`:
```
    Encoding encoding = GetTextEncoding(v2FileChange.FileLocation?.Uri);
```
- Pass the _encoding_, not the URI, to `CreateReplacement` (note that that's the only thing `CreateReplacement` actually needs):
```
internal ReplacementVersionOne CreateReplacement(v2Replacement, Encoding encoding)
{
    ...
}
```
- Within `CreateReplacement`, if you find that you need the encoding but it's `null`, _then_ throw the `UnknownEncodingException`.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -231,6 +275,45 @@ internal RegionVersionOne CreateRegion(Region v2Region)
...
+            ReplacementVersionOne replacement = null;
+
+            if (v2Replacement != null)
+            {
+                replacement = new ReplacementVersionOne();
+                FileContent insertedContent = v2Replacement.InsertedContent;
+
+                if (insertedContent != null)
+                {
+                    if (insertedContent.Text != null)
```
lgolding: Bug: As the code stands, you will try (and possibly fail) to use the inserted text, even if the binary version is available to save you.

The logic should be:
```
IF insertedContent != null THEN
    IF the binary version exists
        Use it
    ELSE
        IF encoding is available
            Convert the text and use the result.
        ELSE
            throw
        END
    END
END
```

src/Sarif/Visitors/UnsupportedEncodingException.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    internal class UnsupportedEncodingException : Exception
+    {
+        public string EncodingName { get; private set; }
```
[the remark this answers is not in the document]
EasyRhin0: Done

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Much better. The only sacrifice is that the unknown encoding name is only available within the helper method.

EasyRhin0: Have a refactor that should fix this...

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -59,9 +59,11 @@ internal FileChangeVersionOne CreateFileChange(FileChange v2FileChange)
 
             if (v2FileChange != null)
             {
+                string encodingName = GetFileEncodingName(v2FileChange.FileLocation?.Uri);
```
lgolding: I understand the trade-off you're making. You call `Encoding.GetEncoding(encodingName)` on every replacement, rather than doing it once, so you have `encodingName` available to throw in the exception.

As an alternative, you could call `GetEncoding` once, and pass _both_ `encoding` and `encodingName` into `CreateReplacement`.  Your call.

EasyRhin0: Yeah good point, I don't like the repetition. What do you think about going back to the CreateReplacement(r, encoding) model, catching our UnknownEncodingException in CreateFileChange, setting the encoding name, and re-throwing?

lgolding: I think that's fine as long as there's a comment explaining why you're doing the catch and rethrow.

PR #927, opened 2018-06-20
Issue 93 regions

EasyRhin0:
Be gentle.


EasyRhin0:
I'm going to factor the FileData/Encoding/TextReader part into a utility method

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -941,7 +941,8 @@ public void SarifTransformerTests_ToCurrent_NotificationExceptionWithStack()
                           },
                           ""region"": {
                             ""startLine"": 15,
-                            ""startColumn"": 9
+                            ""startColumn"": 9,
+                            ""endColumn"": 9
```
lgolding: Yes, that's right. In v1, if `endColumn` was absent, it meant an insertion point before `startColumn`. To express that in v2, you set `endColumn` == `startColumn`.

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -3,7 +3,6 @@
 
 using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
```
lgolding: Thank you for the cleanup :-)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -106,7 +106,8 @@ public Region GetSourceLocation()
         {
             Region sourceLocation = new Region()
             {
-                Offset = m_region.Offset,
+                ByteOffset = m_region.ByteOffset,
+                CharOffset = m_region.CharOffset,
```
lgolding: I doubt you need this line. Before the change, this method was able to do everything it needed to do with `Offset` (which meant an offset measured in bytes). So it should still be able to do everything it needs to do with `ByteOffset`, which means the same thing that `Offset` used to mean.

src/Sarif.Viewer.VisualStudio/ViewModels/ViewModelLocator.cs
```diff
@@ -64,31 +64,31 @@ private static SarifErrorListItem GetDesignTimeViewModel1()
...
                 FilePath = @"D:\GitHub\NuGet.Services.Metadata\src\Ng\Catalog2Dnx.cs",
-                Region = new Region(11, 1, 11, 2, 0, 0, snippet: null, message: null),
+                Region = new Region(11, 1, 11, 2, 0, 0, 0, 0, snippet: null, message: null),
             });
 
             viewModel.Locations.Add(new Models.CodeFlowLocationModel()
             {
                 FilePath = @"D:\GitHub\NuGet.Services.Metadata\src\Ng\Catalog2Dnx.cs",
-                Region = new Region(12, 1, 12, 2, 0, 0, snippet: null, message: null),
+                Region = new Region(12, 1, 12, 2,0, 0, 0, 0, snippet: null, message: null),
```
lgolding: Missing blank space before "0", here and below.

EasyRhin0: Fixed

src/Sarif/ConsoleLogger.cs
```diff
@@ -186,7 +186,8 @@ private void WriteToConsole(ResultLevel level, Uri uri, Region region, string ru
             if (region != null)
             {
                 // TODO 
-                if (region.Offset > 0 ||
+                if (region.CharOffset > 0 ||
+                    region.ByteOffset > 0 ||
                     region.StartColumn == 0)
```
lgolding: This change is correct in that it is consistent with the existing logic. The existing logic says "If you don't give me a start column, or if you _do_ give me an offset, then I don't know how to format the region string". I don't think you should throw, though. I'd just produce an empty string.

EasyRhin0: Done

src/Sarif/ExtensionMethods.cs
```diff
@@ -144,53 +144,6 @@ public static string FormatForVisualStudio(this ResultLevel level)
             }
         }
 
-        /// <summary>
-        /// Completely populate all Region property members. Missing data
-        /// is computed based on the values that are already present.
-        /// </summary>
-        /// <param name="region"></param>
-        /// <param name="newLineIndex"></param>
-        public static void Populate(this Region region, NewLineIndex newLineIndex)
```
lgolding: Why is this removed? Was it never used?

EasyRhin0: Not used

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -282,7 +283,7 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(PhysicalLocation v2Ph
             {
                 physicalLocation = new PhysicalLocationVersionOne
                 {
-                    Region = CreateRegion(v2PhysicalLocation.Region),
+                    Region = CreateRegion(v2PhysicalLocation.Region, v2PhysicalLocation.FileLocation?.Uri),
```
lgolding: Right, you might need the file contents.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -291,31 +292,166 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(PhysicalLocation v2Ph
...
             RegionVersionOne region = null;
 
             if (v2Region != null && (v2Region.StartColumn > 0 ||
                                      v2Region.StartLine > 0 ||
                                      v2Region.EndColumn > 0 ||
                                      v2Region.EndLine > 0 ||
-                                     v2Region.Length > 0 ||
-                                     v2Region.Offset > 0))
+                                     v2Region.ByteLength > 0 ||
+                                     v2Region.ByteOffset > 0 ||
```
lgolding: This was a bug in the original code. `ByteOffset` can validly be 0 if the region starts at the beginning of the file. `ByteLength` can be 0 if the region represents an insertion point.  Similarly for the `Char` properties.

So... the original logic, which attempted to use the property values to decide whether to create a region, was just wrong. If there is a `v2Region` object, you have to convert it, even if all its properties are 0 -- that just means "an insertion point at the beginning of the file".

So this should just be:
```
if (v2Region != null)
{
    region = new RegionVersionOne()
    {
        ...
    };
}
```

EasyRhin0: Good, fixed.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -291,31 +292,166 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(PhysicalLocation v2Ph
...
                     EndColumn = v2Region.EndColumn,
                     EndLine = v2Region.EndLine,
-                    Length = v2Region.Length,
-                    Offset = v2Region.Offset,
-                    StartColumn = v2Region.StartColumn,
+                    StartColumn = v2Region.StartColumn > 0 ? v2Region.StartColumn : 1,
                     StartLine = v2Region.StartLine
                 };
+
+                if ((v2Region.CharLength > 0 || v2Region.CharOffset > 0) ||
```
lgolding: Again, you can't make any decision based on the value of `CharOffset`. It can validly be 0.

lgolding: I'm going to need to think about this. I think there are valid cases that are not covered. I'll get back to you...

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
EasyRhin0: Was wavering on this. Fixed.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: I sent mail about this. Also, I've looked at the logic below now, and it's not quite right. The problem is that you treat the char/offset case and the line/column case separately. That is, the logic embodies the notion that a text region is specified _either_ by charOffset/charLength _or_ by a combination of line/column properties. But the spec says:

> The start of a text region **MAY** be specified by a combination of `startLine` and `startColumn`, or by `charOffset`, or both.
>
> The end of a text region **MAY** be specified by a combination of `endLine` and `endColumn`, or by `charLength`, or both.

So you might specify the start by `charOffset` and the end by `endLine`, `endColumn`. Or you might specify the start by `startLine`, `startColumn` and the end by `charLength`. Or any combination.

EasyRhin0: Yeah this is why I was asking for advice on the conditional flow the other day. I need to meditate...

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1108,9 +1108,7 @@ public void SarifTransformerTests_ToVersionOne_ResultLocations()
                 ""region"": {
                   ""startLine"": 20,
                   ""startColumn"": 69,
-                  ""endColumn"": 74,
-                  ""offset"": 720,
-                  ""length"": 5
+                  ""endColumn"": 74
```
lgolding: You may as well copy `offset` and `length` to `byteOffset` and `byteLength`.

lgolding: The fact that the "v2" log above includes `offset` and `length`, rather than `byteOffset` and `byteLength`, explains why this v1 location is missing its `offset` and `length` properties. When you fix the v2 log above, the missing properties should reappear. It's worth asking why you're not seeing an exception when you deserialize the incorrect v2 log.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1164,10 +1158,7 @@ public void SarifTransformerTests_ToVersionOne_ResultLocations()
                   ""message"": ""Foxes, quick and brown"",
                   ""locations"": [
                     {
-                      ""region"": {
-                        ""offset"": 1345,
-                        ""length"": 22
-                      }
+                      ""region"": {}
```
lgolding: Likewise, you shouldn't be getting an empty region here. It should just have `byteOffset` and `byteLength`.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1152,9 +1148,7 @@ public void SarifTransformerTests_ToVersionOne_ResultLocations()
                 ""region"": {
                   ""startLine"": 56,
                   ""startColumn"": 8,
-                  ""endColumn"": 78,
-                  ""offset"": 1124,
-                  ""length"": 22
+                  ""endColumn"": 78
```
lgolding: See line 1143 above. I can't comment that line in GitHub because it's not part of the change, but... part of Issue #93 is that `snippet` became a property of `region`.

Please do that in a separate PR -- this one is already complicated enough. But do make a note so it doesn't fall through the cracks!

lgolding: No WAIT! The move of `snippet` to `region` was part of oasis-tcs/sarif-spec/130, "Fix the location object". This is just another example, where we are silently deserializing an invalid v2 log file.

Please file the following issue: v2 log file with extra properties is silently deserialized.

Once you fix that, all the broken tests will pop to the surface, and you'll need to fix them as part of the same PR.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Didn't notice this.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
             {
                 physicalLocation = new PhysicalLocationVersionOne
                 {
-                    Region = CreateRegion(v2Region)
+                    Region = CreateRegion(v2Region, null)
```
lgolding: I suggest you name the `null` argument value, `uri: null`. I tripped over this because at first I thought the argument was an encoding.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        // The start of the region is described by character offset
+
+                        if (v2Region.CharOffset == 0)
+                        {
+                            region.Offset = 0;
+                        }
+                        else
+                        {
+                            // Try to get the byte offset using the file encoding and contents
+                            region.Offset = GetRegionByteOffset(v2Region, uri);
```
lgolding: You don't need to pass the whole region; the only thing it uses is charOffset. So I suggest you rename the method: `ConvertCharOffsetToByteOffset(v2Region.CharOffset, uri)`

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: 1143 is v1 json
The issue still exists of course so I'll bug it.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: ... but see my next comment...

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: #933

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Good call, fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        }
+                    }
+
+                    if (v2Region.CharLength > 0)
+                    {
+                        // The end of the region is described by character length
+                        // Try to get the byte length using the file encoding and contents
+                        region.Length = GetRegionByteLength(v2Region, uri);
```
lgolding: Ok, in this case your method needs both an offset and a length. So you _could_ call it `ConvertCharOffsetAndLengthToByteLength(int charOffset, int charLength, Uri uri)`. But perhaps you think that's just too long a name, and you want to keep `GetRegionByteLength`. In that case, you might argue that for symmetry, you should keep the name `GetRegionByteOffset` for the other method.

Your call.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Yup, changed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: #NamesAreMyPassion

EasyRhin0: Yeah but I need the start props here too

lgolding: Thanks for offline explanation. Yes, this method should stay as `GetRegionByteLength`.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        // IT MEANS "THE REST OF THE StartLine". BUT IF CHARLENGTH IS
+                        // PRESENT BUT 0, IT MEANS "INSERTION POINT". AND WE CAN'T TELL
+                        // THE DIFFERENCE.
+                        // TODO: Issue #932
+
+                        // Assume it's an insertion point
+                        if (v2Region.StartLine > 0)
+                        {
+                            region.EndLine = region.StartLine;
+                            region.EndColumn = region.StartColumn;
```
lgolding: This works only if the start of the v2 region was specified by `startLine`/`startColumn` If the start of the v2 region was specified by `charOffset`, then you never populated `region.startLine` and `region.startColumn`.

But good news! The v1 spec says:
> If none of endLine, endColumn, or length is present, the length of the region shall be considered to be 0.

So, if you want an insertion point, just set `region.EndLine = region.EndColumn = region.Length = 0`. You don't need the conditonal on `v2Region.StartLine` being present. And then you don't need the helper method `GetRegionStart`.

EasyRhin0: Ah gotcha

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                if (reader != null)
+                {
+                    reader.Dispose();
+                }
+            }
+        }
+
+        private int GetRegionByteOffset(Region v2Region, Uri uri)
+        {
+            int result = 0;
```
lgolding: Ultra-nit: Since "result" is such a common SARIF term, name this value `byteOffset`. In fact, I think that's a better (more explicit) name even if we weren't talking about SARIF.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Agreed. Fixed everywhere

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+
+            return result;
+        }
+
+        private int GetRegionByteLength(Region v2Region, Uri uri)
+        {
+            int result = 0;
+            TextReader reader = null;
+
+            try
```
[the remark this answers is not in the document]
EasyRhin0: Duh. I think I had it in my head that using only worked with a ctor

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+            {
+                Encoding encoding;
+                reader = GetFileTextReader(uri, out encoding);
+
+                if (reader != null)
+                {
+                    char[] buffer = new char[v2Region.CharOffset];
+
+                    // Read everything up to charOffset
+                    if (reader.ReadBlock(buffer, 0, buffer.Length) > 0)
```
lgolding: Interesting. I didn't know about `ReadBlock` and I had to look up [the difference](https://stackoverflow.com/questions/10497575/what-is-the-difference-between-read-and-readblock-calls-on-textreader) between it and `Read`. I agree, `ReadBlock` is appropriate here.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+            {
+                Encoding encoding;
+                reader = GetFileTextReader(uri, out encoding);
+
+                if (v2Region.StartLine > 0) // Use line and column 
+                {                    
+                    string sourceLine = null;
+
+                    // Read down to the line before startLine
+                    for (int i = 1; i < v2Region.StartLine; sourceLine = reader.ReadLine(), i++) { }
```
lgolding: You need to check for EOF in case there aren't that many lines.

lgolding: Also: Why did you use `ReadBlock` in the other method but `Read` here?

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+
+                    // Read to startColumn
+                    char[] buffer = new char[v2Region.StartColumn];
+                    reader.Read(buffer, 0, buffer.Length);
```
lgolding: Again, there might not be that many columns.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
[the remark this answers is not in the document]

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+
+                    // Read the next charLength characters
+                    buffer = new char[v2Region.CharLength];
+                    reader.Read(buffer, 0, buffer.Length);
+
+                    result = encoding.GetByteCount(buffer);
+                }
+                else // Use charOffset
+                {
```
lgolding: You have to read (skip) the the first `charOffset` characters, and then read the next `charLength` characters. As the code stands, you're just reading  the first `charOffset` characters.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+            int result = 0;
+            TextReader reader = null;
+
+            try
+            {
+                Encoding encoding;
+                reader = GetFileTextReader(uri, out encoding);
+                string sourceLine = null;
+
+                // Read down to the line before endLine
```
lgolding: Bug: You need the length of the `endLine`'th line, not the one before it.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                FileData fileData;
+                if (_currentV2Run.Files.TryGetValue(uri.OriginalString, out fileData))
+                {
+                    // Determine the encoding
+                    string encodingName = null;
+
+                    if (fileData.Contents?.Text != null)
+                    {
+                        // Embedded text shall be UTF-8 encoded
+                        encoding = Encoding.UTF8;
```
lgolding: The fact that the `file` object contains a `contents.text` property, together with the fact that text in a SARIF file is always encoded in UTF-8, does _not_ mean that the file's encoding is UTF-8. The spec explains (§3.2.2, fileContents.text property):
> If the external file is a text file, a `fileContent` object **SHOULD** contain a property named `text` whose value is a string containing the relevant text. Since SARIF log files are encoded in UTF-8 ([RFC3629]; see §3.1), this means that if the external file is a text file in any encoding other than UTF-8, the SARIF producer **SHALL** transcode the text to UTF-8 before assigning it to the text property...

So you see, the contents of `file.contents.text` might have been transcoded. You have to rely on the `file.encoding` property. If it's absent (and `run.defaultFileEncoding` is missing), you're out of luck, because it has no default value (see §3.17.9, file.encoding property):
> If the `file` object represents a text file and this property is absent, it **SHALL** default to the value of the `defaultFileEncoding` property (§3.11.18) of the containing `run` object (§3.11), if that property is present; otherwise, the file’s encoding **SHALL** be taken to be unknown.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +378,313 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                    }
+
+                    if (encoding != null)
+                    {
+                        try
+                        {
+                            if (fileData.Contents != null)
+                            {
+                                // Embedded text file content
+                                string content = fileData.Contents.Text ?? SarifUtilities.DecodeBase64String(fileData.Contents.Binary, encoding);
```
lgolding: They might both be null. The condition on Line 650 should be:
```
if (fileData.Contents != null && (fileData.Contents.Text != null || fileData.Contents.Binary != null))
```

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -630,25 +634,46 @@ internal PhysicalLocation CreatePhysicalLocation(PhysicalLocationVersionOne v1Ph
...
+
+                if (startIsTextBased && endIsTextBased && v1Region.EndColumn == 0)
+                {
+                    region.EndColumn = v1Region.StartColumn;
+                }
+                else if (startIsTextBased)
+                {
+                    region.CharLength = GetRegionCharLength(v1Region, v1PhysicalLocation.Uri);
+                }
+                else
```
lgolding: This was a bug in my sample code. If we get here, `startIsTextBased` is false. If `endIsTextBased` is also false, then the entire region was specified by a byte offset and a byte length, and we've already copied those properties. So the `else` should be `else if (endIsTextBased)`.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+            {
+                // If we get here, we were unable to determine region character offset, so we have to warn the caller
+                // TODO: add a warning to the list
+            }
+
+            return stream;
+        }
+
+        public TextReader GetFileTextReader(Uri uri, out Encoding encoding)
+        {
```
lgolding: You can DRY this out:
```
public TextReader GetFileTextReader(Uri uri, out Encoding encoding)
{
    TextReader reader = null;
    encoding = null;

    Stream contentStream = GetContentStream(uri);
    if (contentStream != null)
    {
        reader = new StreamReader(contentStream);
        encoding = reader.CurrentEncoding;
    }

    return reader;
}
```

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+            }
+
+            return result;
+        }
+
+        public int GetRegionCharOffset(RegionVersionOne v1Region, Uri uri)
+        {
+            int result = 0;
+
+            TextReader reader = null;
```
lgolding: I suggest you change the type to `StreamReader`, for reasons I will explain...

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+            Stream stream = null;
+
+            try
+            {
+                // Get a reader so the encoding might be detected by the StreamReader ctor
+                Encoding encoding;
+                reader = GetFileTextReader(uri, out encoding);
```
lgolding: ... and I suggest you rename this to `GetFileStreamReader`, again for reasons I will explain...

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+
+                if (reader != null)
+                {
+                    // Close the reader
+                    reader.Close();
+
+                    // Get a stream so we can do byte operations
+                    stream = GetContentStream(uri);
```
lgolding: Now... if `reader` is a `StreamReader`, you just have to do `stream = reader.BaseStream;` which is more efficient.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+
+                    if (stream != null)
```
lgolding: ... and then you won't have to test for null because `reader` is not null.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,216 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+                }
+            }
+            finally
+            {
+                if (reader != null)
+                {
+                    reader.Dispose();
+                }
+
+                if (stream != null)
```
lgolding: ... and you won't have to do this dispose.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Whoops. Fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Ah I see.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Done (now a using)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Still figuring out how to handle this...

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Fixed

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

EasyRhin0:
Whoops! Fixed x 2

EasyRhin0:
But if I read characters, I have to watch for CRLF, no?

lgolding:
Re: CRLF: You might be right. I'll look at that again in the morning.

EasyRhin0:
I'm seeing regular comments here, rather than code-context comments?

lgolding:
Re: regular comments. Sorry, that's because I was commenting on the commit, rather than the PR, because I wanted to be able to see the changes. I didn't realize that as a result, you wouldn't see the comments in context. I'll stop doing that.

lgolding:
Overall comment: There are _many_ combinations of properties that might need to be translated, in both directions. We don't have nearly the test coverage we need, and it's hard to get it by way of our normal tests that translate an entire log file from one version to another.

Ideally, we need a set of tests that look like this:
```
RegionVersionOne regionV1 = new RegionV1 { Offset = 2, ... }
Region region = TranslateRegion(regionV1) // or whatever the method is called.
region.Should().Be(ExpectedRegion);
```

Now I know this is complicated because the translation methods expect a file dictionary. You'd have to synthesize a run object and initialize the file dictionary.

However we decide to do it, this is the most complicated aspect of the translation, and it _has_ to be well covered by tests.

lgolding:
I will continue the review tomorrow morning.

EasyRhin0:
Agree on the test coverage. I started keeping notes on uncovered situations/combos/etc.
Have a good night!

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        region.StartLine = v2Region.StartLine;
+                        region.StartColumn = v2Region.StartColumn > 0
+                            ? v2Region.StartColumn
+                            : 1;
+                    }
+                    else
+                    {
+                        // The start of the region is described by character offset
+
+                        if (v2Region.CharOffset == 0)
```
lgolding: Nit: This `if` isn't needed because `ConvertCharOffsetToByteOffset` should work for a 0 offset.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                    {
+                        region.EndLine = v2Region.EndLine;
+
+                        if (v2Region.EndColumn > 0)
+                        {
+                            region.EndColumn = v2Region.EndColumn;
+                        }
+                        else
+                        {
+                            // Try to get the end column using the file encoding and contents
```
lgolding: Suggested comment:
> In v2, if endColumn is missing, then the region extends to the end of the line (exclusive of any newline sequence). Use the file contents and encoding to determine how many columns are in the end line.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                if (reader != null)
+                {
+                    if (v2Region.StartLine > 0) // Use line and column 
+                    {
+                        string sourceLine = null;
+
+                        // Read down to startLine (null return means EOF)
+                        for (int i = 1; i <= v2Region.StartLine && sourceLine != null; sourceLine = reader.ReadLine(), i++) { }
+
+                        if (sourceLine != null && sourceLine.Length <= v2Region.StartColumn)
```
lgolding: If `sourceLine.Length == v2Region.StartColumn` (for example, the line is 5 characters long and `StartColumn` is 5), then you don't need to back up.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        {
+                            // Since we read past startColumn, we need to back up using the base stream
+                            Stream stream = reader.BaseStream;
+                            stream.Position -= v2Region.StartColumn - 1;
```
lgolding: Bug: You need to back up by the full `StartColumn` characters. BUT....

Bug: `stream.Position` is measured in bytes, not characters.

EasyRhin0: Ugh

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+
+                            // Read the next charLength characters
+                            char[] buffer = new char[v2Region.CharLength];
+                            reader.Read(buffer, 0, buffer.Length);
+
+                            byteLength = encoding.GetByteCount(buffer);
```
lgolding: Bug: If the condition on the enclosing `if` statement is false, `byteLength` is never assigned.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Bug: _If_ you end up using some variant of this method, be aware that `StartColumn` might be 0, in which case you'll need to set it to 1 (using a temporary variable, since I don't think you want to modify `v2Region`):
```
int startColumn = v2Region.StartColum > 0 ? v2Region.StartColumn : 1;
```

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,284 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                if (reader != null)
+                {
+                    string sourceLine = null;
+
+                    // Read down to endLine (null return means EOF)
+                    for (int i = 1; i <= v2Region.EndLine && sourceLine != null; sourceLine = reader.ReadLine(), i++) { }
+
+                    if (sourceLine != null && sourceLine.Length < v2Region.EndColumn)
+                    {
+                        endColumn = sourceLine.Length;
```
lgolding: Bug: Must be `sourceLine.Length + 1`, because regions are "half-open": That is, they include the start column but not the end column. So if the line is 5 characters long, and you want to specify the entire line, you must set endColumn to 6.

EasyRhin0: Right, fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: I believe I have the logic wrong here. Should back up sourceLine.Length - (v2Region.StartColumn - 1) characters

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,285 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                        // Read down to startLine (null return means EOF)
+                        for (int i = 1; i <= v2Region.StartLine && sourceLine != null; sourceLine = reader.ReadLine(), i++) { }
+
+                        if (sourceLine != null)
+                        {
+                            if (sourceLine.Length > v2Region.StartColumn)
+                            {
+                                // Since we read past startColumn, we need to back up using the base stream
+                                Stream stream = reader.BaseStream;
+                                stream.Position -= encoding.GetByteCount(sourceLine.Substring(v2Region.StartColumn - 1));
```
lgolding: Bug: `StartColumn` might be 0, in which case it defaults to 1.
```
int startColumn = v2Region.StartColumn > 0
    ? v2Region.StartColumn
    : 1;
stream.Position -= encoding.GetByteCount(sourceLine.Substring(startColumn - 1));
```

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,285 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                FileData fileData;
+                if (_currentV2Run.Files.TryGetValue(uri.OriginalString, out fileData))
+                {
+                    // We need the encoding because the content might have been transcoded to UTF-8
+                    string encodingName = fileData.Encoding ?? _currentV2Run.DefaultFileEncoding;
+                    encoding = GetFileEncoding(encodingName);
+
+                    if (encoding != null)
+                    {
+                        if (fileData.Contents != null && fileData.Contents.Binary != null)
```
lgolding: Could be written `if (fileData.Contents?.Binary != null)`.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,285 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+        {
+            StreamReader reader = null;
+            encoding = null;
+
+            Stream contentStream = GetContentStream(uri, out encoding);
+            if (contentStream != null)
+            {
+                // If we found embedded content, encoding was specified on the file object
+                // If the content is in an extenal file, we let the StreamReader detect the encoding
+                reader = encoding != null ?
```
lgolding: Bug: If you don't have an encoding, there's nothing you can do. As I mentioned in the notes on the pseudo-code I sent: you never guess the encoding from the file stream (it just sniffed the file and made a guess). You can only use the encoding from the SARIF file.
```
private StreamReader GetFileStreamReader(Uri uri, out Encoding encoding)
{
    Stream contentStream = GetContentStream(uri, out encoding);
    return (contentStream != null && encoding != null)
        ? new StreamReader(contentStream, encoding, detectEncodingFromByteOrderMark: false);
        : null;
}
```

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -377,38 +379,285 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
...
+                    if (encoding != null)
+                    {
+                        if (fileData.Contents != null && fileData.Contents.Binary != null)
+                        {
+                            // Embedded binary file content
+
+                            byte[] content = Convert.FromBase64String(fileData.Contents.Binary);
+                            stream = new MemoryStream(content);
+                        }
+                        else if (fileData.Contents.Text != null)
```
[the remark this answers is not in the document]

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: In fact, you have to calculate that defaulted start column right after the test for `sourceLine != null`, because you use it in the test `sourceLine.length > v2Region.StartColumn`.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Forgot about this code. Fixed.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Two issues here:
1. As I said in my follow-on comment above, you must calculate the defaulted `startColumn` before Line 510, because the test on that line uses `v2Region.StartColumn`.
2. Although you calculated `startColumn`, you didn't actually use in on Line 517.

In summary:
```
if (sourceLine != null)
{
    int startColumn = v2Region.StartColumn > 0
        ? v2Region.StartColumn
        : 1;

    if (sourceLine.Length > startColumn)
    {
        // Since we read past startColumn, we need to back up using the base stream.
        Stream stream = reader.BaseStream;

        stream.Position -= encoding.GetByteCount(sourceLine.Substring(startColumn - 1));
        ...
```

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -582,14 +585,14 @@ private Stream GetContentStream(Uri uri, out Encoding encoding)
 
                     if (encoding != null)
                     {
-                        if (fileData.Contents != null && fileData.Contents.Binary != null)
+                        if (fileData.Contents?.Binary != null)
```
lgolding: LGTM.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -582,14 +585,14 @@ private Stream GetContentStream(Uri uri, out Encoding encoding)
...
                         {
                             // Embedded binary file content
 
                             byte[] content = Convert.FromBase64String(fileData.Contents.Binary);
                             stream = new MemoryStream(content);
                         }
-                        else if (fileData.Contents.Text != null)
+                        else if (fileData.Contents?.Text != null)
```
lgolding: LGTM.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -645,14 +648,9 @@ private StreamReader GetFileStreamReader(Uri uri, out Encoding encoding)
             encoding = null;
```
lgolding: Nit: You don't have to initialize `encoding`, because the call to `GetContentStream` will assign it. The compiler checks to make sure that `out` parameters are assigned on all code paths.

EasyRhin0: OK

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

lgolding:
Most recent changes to v2-to-v1 conversion LGTM. That takes care of Steps 1 and 2 in our 6-Step Program (1. CM: Fix v2-to-v1 conversion; 2. LJG: Review v2-to-v1 conversion.) On to Step 3 (CM: Transfer any lessons learned to v1-to-v2 conversion)!

lgolding:
Oopsie.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -630,25 +634,46 @@ internal PhysicalLocation CreatePhysicalLocation(PhysicalLocationVersionOne v1Ph
...
-                    Region = CreateRegion(v1PhysicalLocation.Region)
+                    Region = CreateRegion(v1PhysicalLocation)
                 };
             }
 
             return physicalLocation;
         }
 
-        internal Region CreateRegion(RegionVersionOne v1Region)
+        internal Region CreateRegion(PhysicalLocationVersionOne v1PhysicalLocation)
```
lgolding: This is fine. You can reduce coupling by passing `Region region, string uri` (that is, just the bits of the `v1PhysicalLocation` that you need.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -663,7 +688,7 @@ internal Region CreateRegion(PhysicalLocationVersionOne v1AnnotationLocation, Ph
             // So only copy the v1 annotations that refer to the same file as the location.
             if (v1PhysicalLocation != null && v1AnnotationLocation.Uri == v1PhysicalLocation.Uri)
             {
-                region = CreateRegion(v1PhysicalLocation.Region);
+                region = CreateRegion(v1PhysicalLocation);
```
lgolding: As above, your other option is to pass `v1PhysicalLocation.Uri` as a second argument.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -690,6 +715,151 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
...
+            int charLength = 0;
+            
+            using (StreamReader reader = GetFileStreamReader(uri))
+            {
+                if (reader != null)
+                {
+                    // Read each line preceding the startLine and copy them into a StringBuilder
+                    StringBuilder sb = new StringBuilder();
+                    string sourceLine = null;
+                    for (int i = 1; i < v1Region.StartLine; sourceLine = reader.ReadLine(), i++)
```
lgolding: Bug: As in the v2-to-v1 conversion, the condition needs a `null` check, and `sourceLine` needs to be initialized to `string.Empty`.

lgolding: Ok, you check for `null` inside the loop, but then you don't break out of the loop.

EasyRhin0: Should be consistent. Fixed.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
EasyRhin0: Agreed, changed.

PR #928, opened 2018-06-21
Transform v2 -> v1 location

EasyRhin0:


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -923,6 +923,289 @@ public void SarifTransformerTests_ToVersionOne_NotificationExceptionWithStack()
...
+            {
+              ""physicalLocation"": {
+                ""fileLocation"": {
+                  ""uri"": ""http://localhost:34420/HtmlFixes.html""
+                },
+                ""region"": {
+                  ""startLine"": 20,
+                  ""startColumn"": 69,
+                  ""endColumn"": 74,
+                  ""offset"": 720,
```
lgolding: Of course this will change when the region PR is merged.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -923,6 +923,289 @@ public void SarifTransformerTests_ToVersionOne_NotificationExceptionWithStack()
...
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner""
+      },
+      ""logicalLocations"": {
+        ""html.body.header"": {
```
lgolding: I don't think this is a good example for logical location. A `body` might have multiple `header` elements. To locate the one you want, you'd need an XPath expression like `/html/body/header[2]`. I suggest that you make up an example using regular code, for instance, pick your favorite [C# compiler error](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages).

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -923,6 +923,289 @@ public void SarifTransformerTests_ToVersionOne_NotificationExceptionWithStack()
...
+          ""kind"": ""member"",
+          ""parentKey"": ""html.body""
+        },
+        ""html.body"": {
+          ""name"": ""body"",
+          ""kind"": ""namespace"",
+          ""parentKey"": ""html""
+        },
+        ""html"": {
+          ""name"": ""html"",
```
lgolding: `name` is not needed since it matches the key. It would be a good test to omit it and see if you still emit the correct SARIF v1. (Please take a look at the [SARIF v1 spec](https://rawgit.com/sarif-standard/sarif-spec/master/Static%20Analysis%20Results%20Interchange%20Format%20(SARIF).html) and see if you can omit t there too,).

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -407,6 +502,11 @@ internal ResultVersionOne CreateResult(Result v2Result)
                     }
                 }
 
+                if (v2Result.AnalysisTarget != null)
+                {
+                    result.Locations[0].AnalysisTarget = CreatePhysicalLocation(v2Result.AnalysisTarget);
```
lgolding: Set this property on every v1 `location` object.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -291,6 +352,37 @@ internal PhysicalLocationVersionOne CreatePhysicalLocation(PhysicalLocation v2Ph
...
+                    Uri = v2FileLocation.Uri,
+                    UriBaseId = v2FileLocation.UriBaseId
+                };
+            }
+
+            return physicalLocation;
+        }
+
+        internal PhysicalLocationVersionOne CreatePhysicalLocation(Region v2Region)
+        {
```
lgolding: This can't be right. A v1 `physicalLocation` requires a [`uri` property](https://rawgit.com/sarif-standard/sarif-spec/master/Static%20Analysis%20Results%20Interchange%20Format%20(SARIF).html#physicallocation-uri). You'll see that there were a couple of exceptions to this in the v1 spec, but neither of those exceptions can be represented in v2, so the result is, when converting from v2 to v1, you always need a `uri`.

I haven't yet see the call to this method, so I don't yet know how to avoid this... stay tuned.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -234,6 +272,29 @@ internal InvocationVersionOne CreateInvocation(Invocation v2Invocation)
...
+            {
+                location = new LocationVersionOne
+                {
+                    FullyQualifiedLogicalName = v2Location.FullyQualifiedLogicalName,
+                    Properties = v2Location.Properties,
+                    ResultFile = CreatePhysicalLocation(v2Location.PhysicalLocation)
+                };
+
+                if (!string.IsNullOrWhiteSpace(v2Location.FullyQualifiedLogicalName) &&
+                    _currentV2Run.LogicalLocations?.ContainsKey(v2Location.FullyQualifiedLogicalName) == true)
```
lgolding: If your v2 location has an FQLN, and the logical locations dictionary does not yet have an entry for that FQLN, then you must create one. Otherwise you will lose the decorated name.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: **UPDATE* I was wrong. The place you use this is indeed one of the exceptions in the v1 spec. This is fine.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -35,6 +35,44 @@ public override SarifLog VisitSarifLog(SarifLog v2SarifLog)
...
+        internal AnnotatedCodeLocationVersionOne CreateAnnotatedCodeLocation(Location v2Location)
+        {
+            AnnotatedCodeLocationVersionOne annotatedCodeLocation = null;
+
+            if (v2Location != null)
+            {
+                annotatedCodeLocation = new AnnotatedCodeLocationVersionOne
+                {
+                    Annotations = v2Location.Annotations?.Select(CreateAnnotation).ToList(),
+                    LogicalLocationKey = v2Location.FullyQualifiedLogicalName,
```
lgolding: You also have to fill in v1 FQLN. `LogicalLocationKey` is only allowed in FQLN is present.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Yep

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1329,6 +1612,20 @@ public void SarifTransformerTests_ToVersionOne_TwoResultsWithFixes()
               ""DIV""
             ]
           },
+          ""locations"": [
```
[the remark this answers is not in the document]
EasyRhin0: This is the expected output which now includes locations.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Not following. I'm getting the decorated name *from* the logical location.

lgolding: Oh, I see. _If_ there is a v2 logical location object that contains the decorated name, then copy it to the v1 location object. Got it.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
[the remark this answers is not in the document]
lgolding: Oh right.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Yeah, reversed what we are doing v1 -> v2

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: I'll just use the collections::list::add set used elsewhere

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Fixed and confirmed

src/Sarif/Sarif.csproj
```diff
@@ -40,25 +40,26 @@
     <None Include="CodeGenHints.json" />
     <None Include="CopyrightNotice.txt" />
   </ItemGroup>
-  <ItemGroup>
```
lgolding: VS-induced change :( Who knows what it's thinking...

EasyRhin0: Weird!

PR #931, opened 2018-06-27
Get Multitool Validate verb working.

lgolding:
I also took the opportunity to clean up the project files and build props files.


BuildAndTest.cmd
```diff
@@ -1,116 +1,125 @@
...
+    set NoClean=true&& shift && goto :NextArg
+)
+if "%1" == "/nobuild" (
+    REM If we're not going to build, we'd better not clean.
+    set NoBuild=true&& set NoClean=true&& shift && goto :NextArg
+)
+if "%1" == "/notest" (
+    set NoTest=true&& shift && goto :NextArg
+)
+if "%1" == "/nopublish" (
```
michaelcfanning: does this mean the script by default tries to publish? that seems wrong. this should not be the default. I'll look elsewhere to confirm

michaelcfanning:
'fcib' dir should be named 'refs' instead

src/Sarif.Driver.UnitTests/Sdk/SkimmerBaseTestResources.Designer.cs
```diff
@@ -8,7 +8,7 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace Microsoft.CodeAnalysis.Sarif.Driver.Sdk {
+namespace Microsoft.CodeAnalysis.Sarif.Driver.UnitTests.Sdk {
```
michaelcfanning: >Microsoft.CodeAnalysis.Sarif.Driver.UnitTests.Sdk [](start = 10, length = 49)

why have you renamed these namespaces? generally, the unit test namespaces match the assembly under test.

michaelcfanning:
we should get this code in asap. nice work. we can clean up further later. for example, it'd be nice if you didn't have to point the validate verb to a JSON schema (but we had it built in).

michaelcfanning:
my sample validate run is emitting rule ids like JS1005. shouldn't these be SARIFXXX?

BuildAndTest.cmd
lgolding: Yes, but "publish" doesn't mean what you think :-) The `dotnet publish` command just creates a directory that includes all of the program's runtime dependencies, so you can xcopy deploy it. The "publish" portion of BuildAndTest just runs that program to create the directory, for example `bld\bin\Sarif.Multitool\AnyCPU_Release\Publish\net461\`.

src/Sarif.Driver.UnitTests/Sdk/SkimmerBaseTestResources.Designer.cs
lgolding: I don't recall renaming this manually. I'll investigate.

lgolding:
Re: your comment about rule ids like `JS1005`: To understand this you need to know that the JSchema repo includes a validation library `Json.Schema.Validation` with generic JSON schema validation rules, for example: "The schema says that property "foo" is an array with no more than 4 elements, but this one has 5 elements." The rule id for these generic rules start with "`JS`".

 The JSchema repo also includes a command line tool `JsonSchemaValidator` that just runs those generic JSON schema validation rules.

Now, in the sarif-sdk repo, `multitool Analyze` references the `Json.Schema.Validation` library, and it runs those rules _in addition to_ the SARIF-specific rules.

lgolding:
Re: your comment about embedded the schema in the Multitool: Yes, that's a good idea.

(For some reason I can't reply inline to the comments you're leaving. No matter.)

lgolding:
Re: fcib => refs: Fixed.

PR #936, opened 2018-07-06
Factor commands to distinguish between textual and binary file persistence

michaelcfanning:
Prior to this change the --persist-file-contents command for the analyze and multitool rewrite verb would persist binary as well as textual data. This led to file bloat. Embedding binary data is a corner case (and an insecure practice) while persisting textual data is much more typical (as this enables end-to-end debugging of sarif results soley by referring to information captured in the log file).

Now both analyze and rewrite honor commands that distinguish the two scenarios:

--persist-text-contents (shortcut of 't')
--persist-binary-contents (no shortcut, since we want this insecure setting to be highly readable in all usage)

Within the SDK, we have a well-known list of textual mime-types. All unrecognized mime types are regarded as binary. In the future, it is likely we may need to provide a mechanism to specify additional extensions that should be regarded as textual (as users cannot easily updated the SDK itself).

In the course of developing this change, i reran the test script that embeds file contents into our test log files. This resulted in a fair amount of new code getting embedded into those files (due to me apparently running this code for the first time in an environment where the enlistment drive mapping matched the original output production).

This change introduces a command-line breaking change. --persist-file-contents is no longer supported. We could retain this command and remap it to --persist-text-contents. I've chosen to make a breaking change instead, due to the fact that sarif v2 will entail significant breaks for future consumers. What's another break, more or less? :)


michaelcfanning:
>                 fileContent.Text = inputFileEncoding.GetString(fileContents);

hey! Look what we do here. we convert text data to UTF and we lose the information about original encoding. minimally, we should be persisting the encoding property here.

---
Refers to: src/Sarif/Core/FileData.cs:103 in 1df9d43. [](commit_id = 1df9d43c49aeef1eb4b5eef3f823928d8ecc92b3, deletion_comment = False)

src/Sarif/VersionOne/Core/FileDataVersionOne.cs
```diff
@@ -49,7 +49,7 @@ public partial class FileDataVersionOne : ISarifNodeVersionOne
                 string filePath = uri.LocalPath;
                 bool encodeAsUtf8 = (fileData.MimeType != SarifWriters.MimeType.Binary);
 
-                if (loggingOptions.Includes(Writers.LoggingOptions.PersistFileContents))
+                if (loggingOptions.Includes(Writers.LoggingOptions.PersistTextFileContents))
```
michaelcfanning: >PersistTextFileContents [](start = 67, length = 23)

we should zap these version one helpers.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -94,8 +94,18 @@ public abstract class AnalyzeOptionsBase
         public bool RichReturnCode { get; set; }
 
         [Option(
-            "persist-file-contents",
-            HelpText = "Persist a base64-encoded representation of all referenced files to the log.")]
-        public bool PersistFileContents { get; set; }
+            't', // persist 't'ext
+            "persist-text-contents",
+            HelpText = "Persist a base64-encoded representation of all referenced textual files to the log.")]
```
lgolding: It's not Base64-encoded. It's UTF-8 transcoded.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -94,8 +94,18 @@ public abstract class AnalyzeOptionsBase
...
+        public bool PersistTextFileContents { get; set; }
+
+        [Option(
+            // Lack of shortcut for this one is by design, as the need
+            // to persist this data is unusual and could lead to insecurities.
```
lgolding: Is it insecure because it's easier for a bad actor to modify the binary data without anyone noticing?

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
```diff
@@ -9,7 +9,24 @@
       },
       "files": {
         "file:///d:/src/sarif-sdk/src/sarif.functionaltests/convertertestdata/prefast/src/advancedaliasing/aa_indexalias.cpp": {
-          "mimeType": "text/x-cpp"
+          "mimeType": "text/x-cpp",
+          "contents": {
+            "binary": "Ly8gVHJhY2sgImludGVyZXN0aW5nIiB2YXJpYWJsZXMgYW5kIGRldGVjdCB3aGVuIHRoZXkgYXJlIGFsaWFzZWQuDQoNCnZvaWQgQzYzODVfSW5kZXhBbGlhcyhpbnQgaW5wdXRJbmRleCkNCnsNCglpbnQgKiBjb250ZW50U3RvcmVzID0gbmV3IGludCBbMl07DQoNCgljb250ZW50U3RvcmVzWzBdID0gMDsNCg0KCWNvbnRlbnRTdG9yZXNbMV0gPSAxOw0KDQoJd2hpbGUgKGlucHV0SW5kZXggPiAxKQ0KCXsNCgkJaW50IGluZGV4ID0gaW5wdXRJbmRleDsNCg0KCQljb250ZW50U3RvcmVzWzBdID0gY29udGVudFN0b3Jlc1tpbmRleF07DQoJfQ0KDQoJZGVsZXRlIFtdIGNvbnRlbnRTdG9yZXM7DQp9DQo="
+          },
```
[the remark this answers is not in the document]
lgolding: It's because `text/c-cpp` is not in the textual MIME types list. I don't think there's an official MIME type -- I don't see it in the [IANA registry](https://www.iana.org/assignments/media-types/media-types.xhtml). FWIW, [fileext.com](http://filext.com/file-extension.php?extension=C%2B%2B) says:

> Detailed information for file extension C++:
Primary association: C++ Source Code
Mime type: text/plain, text/x-c++src, application/x-cplusplus

src/Sarif.Multitool/MultitoolOptionsBase.cs
```diff
@@ -26,10 +26,19 @@ public class MultitoolOptionsBase
             HelpText = "Output MD5, SHA1, and SHA-256 hash of analysis targets when emitting SARIF reports.")]
         public bool ComputeFileHashes { get; set; }
 
+        [Option(
+            't', // persist 't'ext
+            "persist-text-contents",
+            HelpText = "Persist a base64-encoded representation of all referenced textual files to the log.")]
```
lgolding: Again, UTF-8 transcoded.

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -57,24 +57,55 @@ public void FileData_ComputeHashes()
...
         }
 
-        [Fact]
-        public void FileData_PersistFileContentsBinary()
+        [Theory]
+        // Unknown files are regarded as binary
+        [InlineData(".unknown", LoggingOptions.PersistBinaryContents, true)]
+        [InlineData(".unknown", LoggingOptions.PersistTextFileContents, false)]
+        [InlineData(".exe", LoggingOptions.PersistBinaryContents | LoggingOptions.PersistTextFileContents, true)]
+        [InlineData(".cs",  LoggingOptions.PersistBinaryContents | LoggingOptions.PersistTextFileContents, true)]
```
lgolding: I suggest that your inline data should be
`fileExtension, loggingOptions, textShouldBePersisted, binaryShouldBePersisted`

src/Sarif.UnitTests/Core/FileDataTests.cs
```diff
@@ -57,24 +57,55 @@ public void FileData_ComputeHashes()
...
+
+                if (shouldBePersisted)
+                {
+                    fileData.Contents.Binary.Should().Be(encodedFileContents);
+                    fileData.Contents.Text.Should().BeNull();
+                }
+                else
+                {
+                    fileData.Contents.Should().BeNull();
+                }
```
lgolding: Then this logic should be
```
if (textShouldBePersisted)
{
    fileData.Contents.Text.Should().Be(fileContents);
    fileData.Contents.Binary.Should().BeNull();
}

if (binaryShouldBePersisted)
{
    string encodedFileContents = Convert.ToBase64String(File.ReadAllBytes(filePath));
    fileData.Contents.Binary.Should().Be(encodedFileContents);
    fileData.Contents.Text.Should().BeNull();
}

if (!textShouldBePersisted && !binaryShouldBePersisted)
{
    fileData.Contents.Should().BeNull();
}
```

src/Sarif/VersionOne/Core/FileDataVersionOne.cs
lgolding: Yes. You'll recall I argued against including them at all, on the principle that the only valid use of the v1 OM was to convert it to v2 -- and so we shouldn't offer people v1 programming aids by way of these helpers.

src/Sarif/Writers/MimeType.cs
```diff
@@ -24,16 +25,18 @@ public static string DetermineFromFileExtension(string path)
                 return MimeType.Directory;
             }
 
-            foreach (ImmutableArray<string> tableEntry in s_extensionTable)
+            foreach (ImmutableArray<ImmutableArray<string>> extensionsTable in GetExtensionsTables())
```
lgolding: I know this is existing code, but... why are we crawling an array? Shouldn't we build a hash table?

lgolding: (We already have one for the textual MIME types.)

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
michaelcfanning: no. generally, it's just a 'Bad Idea' to carry binary data around (due to the possibility someone will execute it).

---
In reply to: [200718546](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r200718546) [](ancestors = 200718546)

src/Sarif/Writers/MimeType.cs
michaelcfanning: is there a note here?

---
In reply to: [200730359](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r200730359) [](ancestors = 200730359)

src/Sarif.Multitool/MultitoolOptionsBase.cs
michaelcfanning: No sir. Not at this time.

---
In reply to: [200722202](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r200722202) [](ancestors = 200722202)

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
lgolding: @michaelcfanning It _is_ true. Regardless of the encoding of the original file, the property `fileContent.text` is UTF-8. See §3.2.2, fileContent.text property:
> Since SARIF log files are encoded in UTF-8 ([RFC3629]; see §3.1), this means that if the external file is a text file in any encoding other than UTF-8, the SARIF producer SHALL transcode the text to UTF-8 before assigning it to the text property.

src/Sarif.Multitool/MultitoolOptionsBase.cs
lgolding: Again, yes, per the spec.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
michaelcfanning: I don't think you understand what I'm saying. the --persist-text-contents argument does NOT populate the Contents.Text property (which certainly is UTF-8, by spec). The reason is that we don't have a reliable mechanism to determine the encoding of a file on disk (and what this rewrite does is inject file contents from disk into the SARIF). Because we don't know the original encoding of the file, we simply inject it into Contents.Binary. We know this binary representation refers to a textual file only because of its extension. Make sense?

---
In reply to: [201072947](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r201072947) [](ancestors = 201072947)

src/Sarif.Multitool/MultitoolOptionsBase.cs
michaelcfanning: No, sire. The behavior of this command  is as follows: for any URI in the files table, if the extension of the file indicates it is a text file, base64 encode that file's contents into the file object Contents.Binary property. And so the help explanation here is accurate. In order to persist the file's contents to Contents.Text, we'd need to identify the encoding of the file on disk. There's no reliable mechanism to do this for all cases. Currently, our API doesn't allow us to do something like say 'assume any file you load from disk is UTF8'

---
In reply to: [201073313](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r201073313) [](ancestors = 201073313)

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
lgolding: Oh my gosh. Then I completely misread the code and the distinction between `--persist-text-contents` and `--persist-binary-contents`. I need to look at the product code again. Back with you shortly.

lgolding: I see now. The comments are correct. `--persist-text-contents` means "Populate `file.content`, but only for text files" and `--persist-binary-content` means "Populate `file.content`, but only for binary files". But you always, only, populate `file.content.binary`.

src/Sarif.Multitool/MultitoolOptionsBase.cs
lgolding: Yup. As I said in the other comment I just sent (we crossed wires), I completely misunderstood the purpose of these options. And that means I completely misunderstood the product code as well, since I approved the PR. Sorry.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
michaelcfanning: That's it!

---
In reply to: [201077447](https://github.com/Microsoft/sarif-sdk/pull/936#discussion_r201077447) [](ancestors = 201077447)

PR #942, opened 2018-07-12
V2 insert code snippets

michaelcfanning:
Provide mechanism to inject single line of source text in region. See SARIF spec issue below for why we might not want to do this over time.

https://github.com/oasis-tcs/sarif-spec/issues/197


michaelcfanning:
@lgolding @EasyRhin0

src/Sarif.Converters.UnitTests/CppCheckErrorTests.cs
```diff
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
```
lgolding: Yay cleanups.

src/Sarif.FunctionalTests/UpdateBaselines.ps1
```diff
@@ -2,7 +2,7 @@
     [string]$ToolName
 )
 
-$utility = "$PSScriptRoot\..\..\bld\bin\AnyCPU_Release\net452\Sarif.Multitool.exe"
+$utility = "$PSScriptRoot\..\..\\bld\bin\Sarif.Multitool\AnyCPU_Release\net461\Sarif.Multitool.exe"
```
lgolding: Remove extra backslash "`..\\bld`".

Thanks for fixing; this wasn't even on my punch list.

src/Sarif.FunctionalTests/UpdateBaselines.ps1
```diff
@@ -49,12 +49,12 @@ function Build-Baselines($toolName)
...
         Remove-Item $outputTemp -ErrorAction SilentlyContinue
         & Write-Host "$utility convert ""$input"" --tool $toolName --output ""$outputTemp"" --pretty-print --insert TextFiles"
-        &$utility convert "$input" --tool $toolName --output "$outputTemp" --pretty-print --insert TextFiles
+        &$utility convert "$input" --tool $toolName --output "$outputTemp" --pretty-print
 
         # Next, perform some rewriting. The PREfast converter in particular cannot embed file contents as the source      
         # SARIF emitted by the compiler does not contain the optional 'files' member of the 'run' object.
-        Write-Host "$utility rewrite ""$outputTemp"" --output ""$outputTemp"" --pretty-print --insert ""TextFiles;Hashes"" --force"
-        &$utility rewrite ""$outputTemp"" --output ""$outputTemp"" --pretty-print --insert "TextFiles;Hashes" --force
+        Write-Host "$utility rewrite ""$outputTemp"" --output ""$outputTemp"" --pretty-print --insert ""TextFiles;Hashes;CodeSnipets"" --force"
```
lgolding: `CodeSnipets` => `CodeSnippets` (and this is why I've never liked these echoes, here and on Line 51. I'd prefer to just remove them).

lgolding: But anyway, lovely proof of the usefulness of your "`--insert`" design.

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using FluentAssertions;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class ReformattingVisitorTests
+    {
+        //                                  0          10          20 
+        //                                  01234567 8 901234 5 678901 2 345
+        private static string s_testText = "line1\r\nline2\r\nline3\r\nline4";
```
lgolding: `private const string TestText = `

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private static Tuple<Region, string>[] s_testCases = new Tuple<Region, string>[]
```
lgolding: `private static readonly`

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            new Tuple<Region, string>(new Region() { StartLine = 3, EndLine = 4 }, "line3"),
+            new Tuple<Region, string>(new Region() { StartLine = 4, EndLine = 4 }, "line4"),
+            // Char offsets. Should return full line, no matter what subset of line is referenced
+            new Tuple<Region, string>(new Region() { CharOffset = 0, CharLength = 2 }, "line1"),
+            new Tuple<Region, string>(new Region() { CharOffset = 9, CharLength = 1 }, "line2"),
+            new Tuple<Region, string>(new Region() { CharOffset = 16, CharLength = 2 }, "line3"),
+            new Tuple<Region, string>(new Region() { CharOffset = 22, CharLength = 4 }, "line4"),
+            // Binary regions. No return value expected.
+            new Tuple<Region, string>(new Region() { ByteOffset = 20, ByteLength = 5 }, null),
+            // No explicit values is implicitly a binary region with insertion point at beginning of file
```
lgolding: I guess that's right, although the spec doesn't say it explicitly. The spec says that every valid text region includes either `startLine` or `charOffset`, so this isn't a text region. And in a binary region, `length` and `offset` are both allowed to default to 0. Huh.

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            new Tuple<Region, string>(new Region() { }, null),
+        };
+
+        [Fact]        
+        public void ReformattingVisitor_InsertCodeSnippets()
```
lgolding: Ultra-nit: The tests are named so that you can read them in English: "The ReformattingVisitor inserts code snippets.", thus, "`ReformattingVisitor_InsertsCodeSnippets`.

src/Sarif/Core/Region.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public partial class Region
+    {
+        public bool IsBinaryRegion
+        {
+            get
+            {
+                // Is this right? What about an insertion point right after a BOM in a text file??
+                // Do we need to just bite the bullet and make these Nullable type so that we have a
+                // clear indicator of whether the region is binary vs. textual? I tend to think so.
```
lgolding: I'm afraid so, but this is the best you can do right now.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+
             return node;
         }
 
-        internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
+        public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
         {
-            Uri uri;
+            // If we have a binary region, there is no work to do
+            if (node.Region == null ||
```
lgolding: You wrote a helper method `Region.IsBinaryRegion` but didn't use it here (and I like this logic better for implementing it).

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+            {
+                goto Exit;
+            }
 
-            if (!Uri.TryCreate(key, UriKind.RelativeOrAbsolute, out uri))
+            // If the user hasn't specified code snippet insertion, there is no work to do
+            bool workToDo = false;
+            workToDo |= node.Region != null && _dataToInsert.Includes(OptionallyEmittedData.CodeSnippets);
+
+            if (!workToDo)
```
lgolding: Or just:
```
if (node.Region == null ||  !_dataToInsert.Include(...))
```

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+            {
+                goto Exit;
+            }
+
+            // If we have an absolute URL specified as a file that 
+            // exist on disk, there is work to do
+            if (node.FileLocation.Uri.IsAbsoluteUri &&
+                node.FileLocation.Uri.LocalPath != null &&
+                File.Exists(node.FileLocation.Uri.LocalPath))
```
lgolding: The SARIF SDK has an `IFileSystem` that you can inject to make this more easily unit testable.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+                GetCompleteFirstLineAssociatedWithRegion(region, fileText);
+            }
+
+            Exit:
+            return base.VisitPhysicalLocation(node);
+        }
+
+        internal static void GetCompleteFirstLineAssociatedWithRegion(Region region, string fileText)
+        {
+            if (region.IsBinaryRegion)
```
lgolding: Shouldn't need to test this because you tested it on Line 67.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+        }
+        
+
+        private string GetFileText(Uri uri)
+        {
+            _uriToFileTextMap = _uriToFileTextMap ?? new Dictionary<Uri, string>();
+            
+            if (!_uriToFileTextMap.TryGetValue(uri, out string fileText))
+            {
+                fileText = _uriToFileTextMap[uri] = File.ReadAllText(uri.LocalPath);
```
lgolding: I'm worried about this. You could end up sucking most of your code base into memory; if your code base is Windows, you'll regret it.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+            else
+            {
+                // Not at end of file. Get the offset from the
+                // line after this one, then compute length from there.
+                lineInfo = lineIndex.GetLineInfoForLine(startLine + 1);
+                lineLength = lineInfo.StartOffset - lineOffset;
+            }
+
+            // Grab the line text, except for any trailing line break characters
+            region.Snippet.Text = fileText.Substring(lineOffset, lineLength).Trim();
```
lgolding: You're not just trimming the newline sequence; you're trimming trailing white space.

From an experience point of view, maybe that's ok (there's no point squiggling trailing spaces). But if that's what you intend, the comment should say so.

And in fact, you might have a style rule that says "No Trailing Whitespace", and then you'd want to squiggle it. So on further reflection, I think this is a bug.

src/Sarif/Visitors/ReformattingVisitor.cs
lgolding: Again, you can use `IFileSystem` here.

src/Sarif/Visitors/ReformattingVisitor.cs
```diff
@@ -43,23 +45,129 @@ public override Run VisitRun(Run node)
...
+            {
+                // End of file. Compute length using file length
+                lineLength = fileText.Length - lineOffset;
+            }
+            else
+            {
+                // Not at end of file. Get the offset from the
+                // line after this one, then compute length from there.
+                lineInfo = lineIndex.GetLineInfoForLine(startLine + 1);
+                lineLength = lineInfo.StartOffset - lineOffset;
```
lgolding: Won't this calculation include the newline sequence?

src/Sarif/Visitors/ReformattingVisitor.cs
lgolding: (I see you mentioned this issue in your initial comments. In any case, again, if we decide it's right to Trim, then then comment should reflect that.)

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
michaelcfanning: We should consider requiring an explicit indicator of the kind of a region. This means that one of startLine, charOffset or byteOffset must be explicit. Everything else can be used as a default. This would restrict the # of properties that need to be nullable types.

---
In reply to: [202097770](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202097770) [](ancestors = 202097770)

lgolding: You may recall, I once proposed that a region have text region and binary region sub-objects:
```
region: {
  text: {
    "startLine": 42
  },
  "binary": {
  }
}
```

michaelcfanning: you did. this doesn't resolve all the strangeness, however, due to Jim's suggestion that we implicitly allow a region to start with a specific column and include all the remaining line text. to do this, we need a nullable charLength

[https://github.com/oasis-tcs/sarif-spec/issues/201](https://github.com/oasis-tcs/sarif-spec/issues/201) 

---
In reply to: [202409592](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202409592) [](ancestors = 202409592)

lgolding: You're right. The same is true of your suggestion to have a region type indicator:
```
# charLength and endColumn missing; region extends to end of line.
"region": {
  "type": "text",
  "startLine": 5
},

# Insertion point.
"region": {
  "type": "text",
  "startLine": 5,
  "charLength": 0
}
```
... and again, the SDK can't tell the difference unless `Region.CharLength` is nullable.

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
michaelcfanning: what's the point of making the array reference read only? when i can arbitrarily swap contents of the array. swtiched to readonlycollection, though I'm not sure i understand what the additional work is buying us. :)

---
In reply to: [202095850](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202095850) [](ancestors = 202095850)

lgolding: For me it's a matter of habit. I _always_ mark data I don't want changed. I'm a creature of habit; if I always do it, I won't forget to do it when it matters. YMMV.

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: Right. This is following the workToDo pattern established elsewhere in this file. The thought is that this code will be subsequently updated to include multiline code snippets, so the logical will be built out further.

---
In reply to: [202101477](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202101477) [](ancestors = 202101477)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: This logical doesn't strictly identify a binary region. It allows the entirely empty region through, for example, which, strictly speaking, is a binary region. I'm going to hold tight on further changes here until we revisit this topic in context of the SARIF TC.

---
In reply to: [202100932](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202100932) [](ancestors = 202100932)

src/Sarif.FunctionalTests/UpdateBaselines.ps1
michaelcfanning: zapped them

---
In reply to: [202090806](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202090806) [](ancestors = 202090806)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: good eye, yes, but i want this test in the helper, so that it is complete. in product usage, i test for binary regions early as an optimization.

---
In reply to: [202102249](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202102249) [](ancestors = 202102249)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: Performance (speed) is already surfacing as an issue in the multitool. This could certainly represent a working set issue. It all depends on the scope of files referenced. We certainly don't want to reload files over and over again. Anyway, yes, we need some additional tuning, certainly before this code gets integrated into an actual service/pipeline.

---
In reply to: [202103169](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202103169) [](ancestors = 202103169)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: yes. is there a concern? that's why we have to trim it later. 

---
In reply to: [202106386](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202106386) [](ancestors = 202106386)

src/Sarif.UnitTests/Visitors/ReformattingVisitorTests.cs
michaelcfanning: Persuasive and I did it. Also useful to build the internal muscle of recalling the pattern to author.

---
In reply to: [202413355](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202413355) [](ancestors = 202413355)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: Resolved. Will expand unit tests in next check-in as well.

---
In reply to: [202102007](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202102007) [](ancestors = 202102007)

src/Sarif/Visitors/ReformattingVisitor.cs
michaelcfanning: Changed it just to trim the line break chars

---
In reply to: [202104581](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202104581) [](ancestors = 202104581)

src/Sarif/Visitors/ReformattingVisitor.cs
lgolding: Not a concern if it gets trimmed. I didn't understand well enough how the code fit together to realize that it would be trimmed later. A comment to that effect wouldn't come amiss.

michaelcfanning: I will make this change. I would appreciate your approval soon as this change is getting large and blocking other work.

---
In reply to: [202437257](https://github.com/Microsoft/sarif-sdk/pull/942#discussion_r202437257) [](ancestors = 202437257)

PR #945, opened 2018-07-13
Float file regions cache notion + basic unit test harness

michaelcfanning:
@lgolding @EasyRhin0 

This is setup work for hydrating region text snippets. In order to do so, we need to compute the actual length of a region (which isn't always explicit). I propose to do this with a 'file regions cache' object that has the capability of completing expanding a region's properties and to then compute the textual snippet associated with that precise region.

This PR just shows the basic mechanics of unit-testing with no real functionality.

Michael

### DIFF
--- a/src/Sarif.UnitTests/FileRegionsCacheTests.cs
+++ b/src/Sarif.UnitTests/FileRegionsCacheTests.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.ObjectModel;
+using FluentAssertions;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.UnitTests
+{
+    public class FileRegionsCacheTests
+    {
+        //                                 0          10          20         30
+        //                                 01234 5 6789012 3 45678901 2 345679012
+        private const string s_testText = "line1\r\n line2\r\n  line3\r\n   line4";
+
+        private static ReadOnlyCollection<Tuple<string, Region, Region>> s_testCases =
+            new ReadOnlyCollection<Tuple<string, Region, Region>>(new Tuple<string, Region, Region>[]
+            {
+                // Regions specified only by start line
+                new Tuple<string, Region, Region>(
+                    "line1",
+                    new Region() { StartLine = 1 }, 
+                    new Region() { StartLine = 1, EndLine = 1, StartColumn = 1, EndColumn = 5, CharOffset = 0, CharLength = 5 })
+            });
+
+
+        [Fact]
+        public void FileRegionsCache_PopulatesRegionsFromAbsoluteFileUri()
+        {            
+            var run = new Run();
+            var fileRegionsCache = new FileRegionsCache(run);
+
+            Uri uri = new Uri(@"c:\temp\myFile.cpp");
+            var mockFileSystem = MockFactory.MakeMockFileSystem(uri.LocalPath, s_testText);
+
+            fileRegionsCache._fileSystem = mockFileSystem;
+
+            var physicalLocation = new PhysicalLocation()
+            {
+                FileLocation = new FileLocation()
+                {
+                    Uri = uri
+                }
+            };
+
+            ExecuteTests(fileRegionsCache, physicalLocation);
+        }
+
+        private static void ExecuteTests(FileRegionsCache fileRegionsCache, PhysicalLocation physicalLocation)
+        {
+            foreach (Tuple<string, Region, Region> testCase in s_testCases)
+            {
+                string snippet = testCase.Item1;
+                Region inputRegion = testCase.Item2;
+                Region expectedRegion = testCase.Item3;
+
+                physicalLocation.Region = inputRegion;
+
+                Region actualRegion = fileRegionsCache.PopulatePrimaryRegionProperties(physicalLocation, populateSnippet: false);
+
+                actualRegion.ValueEquals(expectedRegion).Should().BeTrue();
+                actualRegion.Snippet.Should().BeNull();
+
+                actualRegion = fileRegionsCache.PopulatePrimaryRegionProperties(physicalLocation, populateSnippet: true);
+
+                expectedRegion.Snippet = new FileContent() { Text = snippet };
+                actualRegion.ValueEquals(expectedRegion).Should().BeTrue();
+                actualRegion.Snippet.Text.Should().Be(snippet);
+            }
+        }
+    }
+}
--- a/src/Sarif.UnitTests/MockFactory.cs
+++ b/src/Sarif.UnitTests/MockFactory.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using Moq;
+
+using Match = System.Text.RegularExpressions.Match; // Avoid ambiguity with Moq.Match;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal static class MockFactory
+    {
+        public static IFileSystem MakeMockFileSystem(string fileName, string fileText = null, string[] fileLines = null)
+        {
+            var mock = new Mock<IFileSystem>(MockBehavior.Strict);
+            mock.Setup(fs => fs.FileExists(It.IsAny<string>())).Returns((string s) => s.Equals(fileName, StringComparison.OrdinalIgnoreCase));
+            mock.Setup(fs => fs.GetFullPath(It.IsAny<string>())).Returns((string path) => path);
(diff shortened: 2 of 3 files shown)


src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        private const string s_testText = "line1\r\n line2\r\n  line3\r\n   line4";
+
+        private static ReadOnlyCollection<Tuple<string, Region, Region>> s_testCases =
+            new ReadOnlyCollection<Tuple<string, Region, Region>>(new Tuple<string, Region, Region>[]
+            {
+                // Regions specified only by start line
+                new Tuple<string, Region, Region>(
+                    "line1",
+                    new Region() { StartLine = 1 }, 
+                    new Region() { StartLine = 1, EndLine = 1, StartColumn = 1, EndColumn = 5, CharOffset = 0, CharLength = 5 })
```
lgolding: Shouldn't it be `EndColumn = 6`? Because the line does include column 5 (the "1" at the end of "line1"), and the length of the region is 5, and we're supposed to have length == endColumn - startColumn.

lgolding: I suggest you define a type rather than using `Tuple`. I had to say to myself, "OK, that's the original region, and that's the fully populated region"... I'd have liked to see
```
new CachedRegion
{
    Snippet = ...,
    OriginalRegion = ...,
    FullyPopulatedRegion = ...
}
```

src/Sarif/FileRegionsCache.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// This class is a file cache that can be used to populate
+    /// regions with comprehensive data, to retrieve file text
+    /// associated with a SARIF log, and to construct text
+    /// snippets associated with region instances.
+    /// </summary>
+    public class FileRegionsCache
+    {
+        internal IFileSystem _fileSystem;
+
+        private Run _run;
```
lgolding: `readonly`

src/Sarif/FileRegionsCache.cs
```diff
@@ -0,0 +1,70 @@
...
+            Region result = new Region() { StartLine = 1, EndLine = 1, StartColumn = 1, EndColumn = 5, CharOffset = 0, CharLength = 5 };
+
+            if (populateSnippet) { result.Snippet = new FileContent() { Text = "line1" }; }
+
+            return result;
+        }
+
+        private string GetFileText(FileLocation fileLocation)
+        {
+            return _fileSystem.ReadAllText(fileLocation.Uri.LocalPath);
```
lgolding: I assume at some point we'll have to worry about URI base ids, but today is not that day.

src/Sarif.UnitTests/MockFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using Moq;
+
+using Match = System.Text.RegularExpressions.Match; // Avoid ambiguity with Moq.Match;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal static class MockFactory
+    {
+        public static IFileSystem MakeMockFileSystem(string fileName, string fileText = null, string[] fileLines = null)
```
lgolding: Any reason not to just pass `fileText`? Your mock `Setup` for `ReadAllLines` could split it, and you're reduce your storage requirements by half.

src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            });
+
+
+        [Fact]
```
lgolding: Rather than looping on the test cases yourself, you could use xUnit's [`Theory` and `MemberData` attributes](https://andrewlock.net/creating-parameterised-tests-in-xunit-with-inlinedata-classdata-and-memberdata/).

src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        private static void ExecuteTests(FileRegionsCache fileRegionsCache, PhysicalLocation physicalLocation)
+        {
+            foreach (Tuple<string, Region, Region> testCase in s_testCases)
+            {
+                string snippet = testCase.Item1;
+                Region inputRegion = testCase.Item2;
+                Region expectedRegion = testCase.Item3;
+
```
lgolding: Wouldn't have to do this if you had a real type.

src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                physicalLocation.Region = inputRegion;
+
+                Region actualRegion = fileRegionsCache.PopulatePrimaryRegionProperties(physicalLocation, populateSnippet: false);
+
+                actualRegion.ValueEquals(expectedRegion).Should().BeTrue();
```
lgolding: More idiomatic is
```
actualRegion.Should().Be(expectedRegion);
```

lgolding: Oh! `ValueEquals`! I didn't even know about that one. Nice.

src/Sarif/FileRegionsCache.cs
```diff
@@ -0,0 +1,70 @@
...
+                return inputRegion;
+            }
+
+            string fileText = GetFileText(physicalLocation.FileLocation);
+            return PopulatePrimaryRegionProperties(fileText, inputRegion, populateSnippet);
+        }
+
+        private Region PopulatePrimaryRegionProperties(string fileText, Region inputRegion, bool populateSnippet)
+        {
+            if (fileText == null) { return inputRegion; }
```
lgolding: Got it, the logic goes here.

src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -0,0 +1,72 @@
...
+using FluentAssertions;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.UnitTests
+{
+    public class FileRegionsCacheTests
+    {
+        //                                 0          10          20         30
+        //                                 01234 5 6789012 3 45678901 2 345679012
+        private const string s_testText = "line1\r\n line2\r\n  line3\r\n   line4";
```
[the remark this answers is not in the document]
michaelcfanning: These are standards for public items only. For private data, it is a useful standard to use the s_ prefix, as a clear indicator the data is static (and could therefore entail special thread access considerations)

---
In reply to: [202746907](https://github.com/Microsoft/sarif-sdk/pull/945#discussion_r202746907) [](ancestors = 202746907)

src/Sarif.UnitTests/FileRegionsCacheTests.cs
michaelcfanning: um, yes, so why is the test passing? :) will go check

---
In reply to: [202748121](https://github.com/Microsoft/sarif-sdk/pull/945#discussion_r202748121) [](ancestors = 202748121)

lgolding:
Because the product code has no logic, it’s just a dummy 😊

src/Sarif.UnitTests/FileRegionsCacheTests.cs
michaelcfanning: you can't do this defining a strong typed test data class. i have taken this suggestion, but the result is not to use ClassData/MemberData/TheoryData. maybe there's a new way to mix these paradigms I haven't found? 

---
In reply to: [202753131](https://github.com/Microsoft/sarif-sdk/pull/945#discussion_r202753131) [](ancestors = 202753131)

lgolding:
Oh, wow. I'm going to stop replying to comments in email. I didn't realize that it looked so ugly and broke the discussion threading. Sorry.

src/Sarif.UnitTests/FileRegionsCacheTests.cs
lgolding: (Replied in email. You are right.)

src/Sarif.UnitTests/MockFactory.cs
michaelcfanning: ok

---
In reply to: [202750290](https://github.com/Microsoft/sarif-sdk/pull/945#discussion_r202750290) [](ancestors = 202750290)

src/Sarif.UnitTests/FileRegionsCacheTests.cs
michaelcfanning: Yes, this is how the auto-generated code handles it. You have to explicitly utilize this helper to get value type comparison semantics. raises the point that our core types have reference equality semantics when put into hash sets and tables. You can fix this by initializing the set with an autogenerated ValueComparer associated with the type. But I'll bet it will be a common pitfall not to do so. :(

---
In reply to: [202754179](https://github.com/Microsoft/sarif-sdk/pull/945#discussion_r202754179) [](ancestors = 202754179)

PR #947, opened 2018-07-17
v2 -> v1 codeflow

EasyRhin0:
Not sure if collapsing the threadFlows in CreateCodeFlow is the right behavior.


src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1963,5 +1963,286 @@ public void SarifTransformerTests_ToVersionOne_Regions()
...
+        {
+            const string V2LogText =
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner"",
+        ""semanticVersion"": ""2.1.0""
```
lgolding: This is optional. Let's make our tests a little shorter by omitting it.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
```diff
@@ -1963,5 +1963,286 @@ public void SarifTransformerTests_ToVersionOne_Regions()
...
+      },
+      ""logicalLocations"": {
```
lgolding: You don't need `logicalLocations` to test code flows.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Done

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
```diff
@@ -1658,7 +1672,7 @@ public void SarifTransformerTests_ToCurrent_CodeFlows()
               ""message"": ""Path from declaration to usage"",
               ""locations"": [
                 {
-                  ""kind"": ""declaration"",
+                  ""kind"": ""call"",
```
lgolding: Why these changes? The `snippet` shows that this line is not in fact a function call.

EasyRhin0: I wanted the v1->v2 log to match the v2->v1 log, as far as the kind vs. nestingLevel is concerned

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Removed

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: But these are just wrong (the `kind` is inconsistent with the `snippet`).

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs
EasyRhin0: Actually tool is required.

lgolding: Correct, but `tool.semanticVersion` is not.

lgolding: I was not sufficiently precise in my original comment.

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
EasyRhin0: Yeah I get that, I'll adjust.

src/Sarif.UnitTests/SarifCurrentToVersionOneTests.cs

src/Sarif.UnitTests/SarifVersionOneToCurrentTests.cs
lgolding: In fact, if you have a v2 `threadFlowLocation` whose `indentLevel` has increased by 1, then the _previous_ `threadFlowLocation` must have been the call site.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -76,6 +122,32 @@ internal AnnotationVersionOne CreateAnnotation(Region v2Region)
...
+                    Properties = v2CodeFlow.Properties
+                };
+
+                if (v2CodeFlow.ThreadFlows?.Count > 0)
+                {
+                    codeFlow.Locations = new List<AnnotatedCodeLocationVersionOne>();
+
+                    foreach (ThreadFlow tf in v2CodeFlow.ThreadFlows)
+                    {
+                        (codeFlow.Locations as List<AnnotatedCodeLocationVersionOne>).AddRange(CreateAnnotatedCodeLocation(tf));
```
lgolding: Unfortunately it's not this easy; in fact, it's going to be quite difficult. Please read the following carefully.

In v1, a `codeFlow` was described by a single list of `annotatedCodeLocation`s, in chronological order. To represent the fact that there might be multiple threads of execution, `annotatedCodeLocation` had a `threadId` property:

> ... whose value is an integer which identifies the thread that was executing when the execution of a code flow reached the location specified by this object. 

For example:
```
{                                           # A codeFlow object
  "locations": [
    {
      "step": 1,
      "threadId": 42
    },
    {
      "step": 2,
      "threadId": 54
    },
    {
      "step": 3,
      "threadId": 42
    }
  ]
}
```
Step 1 and Step 3 occur on thread 42, Step 2 occurs on thread 54, and Step 2 occurs after Step 1 and before Step 3.

In v2, we decided to represent each thread separately, so a `codeFlow` contains an array of `threadFlow`s. To represent the inter-thread ordering information of the original design, we added a property `threadFlowLocation.executionOrder`.

For example:
```
{                                           # A codeFlow object
  "threadFlows": [
    {                                       # A threadFlow object
      "id": "42",
      "locations": [
        {
          "step": 1
          "executionOrder": 1
        },
        {
          "step": 2,
          "executionOrder": 3
        }
      ]
    },
    {                                       # Another threadFlow object
      "id": 54,
      "locations": [
        {
          "step": 1
          "executionOrder": 2
        }
      ]
    }
  ]
}
```

So when you translate from v1 to v2, you have do something like this:
```
executionOrder = 0
for each location in the v1 codeFlow
  if we have not seen the location's thread id before (and this includes the case where it's absent aka 0)
    create a new thread flow
    threadFlow.id = stringified location.threadId (culture-invariant)
 add the location to the appropriate threadFlow, setting its executionOrder to the next available value
```
I'll leave it to you to figure out the v2 to v1 logic. Be aware that `executionOrder` is _optional_.

This needs to be thoroughly unit tested. If you're going to stick to the current unit test strategy of converting an entire file, make sure that you use the absolute minimum file -- no extra properties. Just the ones you need for each test case.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -57,6 +57,52 @@ internal AnnotatedCodeLocationVersionOne CreateAnnotatedCodeLocation(Location v2
...
+                annotatedCodeLocation.Module = v2ThreadFlowLocation.Module;
+                annotatedCodeLocation.Properties = v2ThreadFlowLocation.Properties;
+                annotatedCodeLocation.State = v2ThreadFlowLocation.State;
+                annotatedCodeLocation.Step = v2ThreadFlowLocation.Step;
+            }
+
+            return annotatedCodeLocation;
+        }
+
+        internal IList<AnnotatedCodeLocationVersionOne> CreateAnnotatedCodeLocation(ThreadFlow v2ThreadFlow)
```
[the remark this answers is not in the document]
EasyRhin0: Yes, fixed

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Larry, for v1 -> v2, how should step be handled? Should I always set a value?

lgolding: If you need to set it at all (it is optional), you should set it on all `threadFlowLocation`s. Each `threadFlow` has its own sequence (1, 2, 3...).

@michaelcfanning The spec is not clear on this point, but it should be: if anybody has a step, everybody should have one, because the purpose is for users to be able to look at it and say "look what happened at step 4!".

I filed oasis-tcs/sarif-spec#203 on that.

lgolding: If it's not clear to you whether you need to set it in a given `codeFlow`, just go ahead and set it always.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -57,6 +57,52 @@ internal AnnotatedCodeLocationVersionOne CreateAnnotatedCodeLocation(Location v2
...
+
+            if (v2ThreadFlow?.Locations?.Count > 0)
+            {
+                annotatedCodeLocationList = v2ThreadFlow.Locations.Select(CreateAnnotatedCodeLocation).ToList();
+
+                if (annotatedCodeLocationList.Count == v2ThreadFlow.Locations.Count)
+                {
+                    for (int i = 0; i < annotatedCodeLocationList.Count - 1; i++)
+                    {
+                        if (v2ThreadFlow.Locations[i].NestingLevel > v2ThreadFlow.Locations[i + 1].NestingLevel)
```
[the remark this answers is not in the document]
lgolding: A native v2 file never had the `kind`-dependent properties, so there's no danger of data loss when converting to v1. And a v2 file that was converted from v1 has the entire v1 run in its run property bag, so again there's no danger of data loss when converting back to v1.

EasyRhin0:
I've fixed the v1 -> v2 codeFlow transformer, and I've moved the test to the file-on-disk model. Does this kind of test need to be moved out of Sarif.UnitTets since it interacts with the file system?

lgolding:
Strictly speaking yes, but don't do it now -- especially as @michaelcfanning and I discussed yesterday a way to run the tests that didn't require reading the files from disk. The idea was:

1. build those disk files into embedded resources and access them from there
2. Write the output to a stream rather than to disk
3. Compare streams rather than files
4. _If_ a test fails, _then_ write the actual output to disk.

That way,

1. You can still _author_ the test files on disk, but
2. The tests run quickly in the overwhelming majority of cases where they pass, because the test isn't reading them from the disk at runtime, and
3. When they file, you still end up with an "actual" disk file you can examine.

If we did that, it would be fine to leave them as "unit tests".

Again, don't do that now -- let's do it in a separate PR.

lgolding:
(And to be clear, even _if_ the right answer was to move them to a separate "functional tests" assembly, I wouldn't suggest doing that as part of this PR.)

EasyRhin0:
Yeah that's how I did it, except I'm writing the files always. I'll change it to only write for failures.
I am having one problem: I used a RESX and found that the file resources don't update unless you rebuild. So this breaks the "change expected sarif log and re-run the test" scenario. I'm going to switch to embedded resources directly in the assembly.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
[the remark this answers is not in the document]

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -57,6 +57,52 @@ internal AnnotatedCodeLocationVersionOne CreateAnnotatedCodeLocation(Location v2
...
+
+        internal IList<AnnotatedCodeLocationVersionOne> CreateAnnotatedCodeLocation(ThreadFlow v2ThreadFlow)
+        {
+            List<AnnotatedCodeLocationVersionOne> annotatedCodeLocationList = null;
+
+            if (v2ThreadFlow?.Locations?.Count > 0)
+            {
+                annotatedCodeLocationList = v2ThreadFlow.Locations.Select(CreateAnnotatedCodeLocation).ToList();
+
+                if (annotatedCodeLocationList.Count == v2ThreadFlow.Locations.Count)
```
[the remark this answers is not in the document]
EasyRhin0: I'll put a todo here and surface a warning/error when I do that work item

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
EasyRhin0: Given that executionOrder isn't required, how am I to handle sequence if it's missing? Just go with the list item order? And then there's the situation where some TFLs have a value but some don't.

lgolding: Yes, go with list order.

Example: Given this v2 code flow:
```
thread 1
    step 1
    step 2
    step 3: executionOrder = 1
    step 4:

thread 2
    step 1
    step 2 executionOrder = 2
```

This means: "thread 2 step 2 occurs after thread 1 step 3" (and that's _all_ it means!).

Given that, there are many valid conversions to a v1 thread flow. One possible translation is:
```
step 1: threadId = 1
step 2: threadId = 1
step 3: threadId = 1
step 4: threadId = 2 <== this was originally step 1 in thread 2!
step 5: threadId = 2 <== this was originally step 2 in thread 2
step 6: threadId = 1 <== this was originally step 4 in thread 1
```

lgolding: @michaelcfanning You see how tricky this is. It would be reasonable to just abandon v2 => v1 down-versioning at this point. The fact that it's "almost finished" isn't a good enough reason to finish it. That's the "sunk cost fallacy". The time you'd spend on that would be better spent _removing_ v2 => v1 conversion and its tests. IMNSHO.

EasyRhin0: An important consideration is the iffy quality of the v1 log that we're able to produce. The v2 schema is a lot more sophisticated, there are few required properties, and there are many significant differences, so we end up either guessing or dropping data at a number of points during transformation.
And my current codeFlow transformer architecture won't work for the above so I'll spend extra time refactoring.

EasyRhin0: @michaelcfanning Thoughts here Michael?

PR #954, opened 2018-07-23
Clean up handling of schema file in tests

lgolding:
Sarif.FunctionalTests incorrectly copied the schema to the root of the bld\bin directory rather than to the Sarif.FunctionalTests subdirectory of the bld\bin directory. It turns out that the functional tests don't need the schema at all, so I removed the Copy task from the project file.

Sarif.ValidationTests copied the schema in such a way that not only was there a copy in each framework directory (for example. bld\bin\Sarif.ValidationTests\Any_CPU_Release\net461), but there was an extra copy in the "flavor" directory (bld\bin\Sarif.ValidationTests\Any_CPU_Release). I fixed that as well.

This was one of the tasks on my punch list as I converted the repo to the new project system. I noticed it because I wondered why there were copies of the schema directly under bld\bin\AnyCPU_Release.

### DIFF
--- a/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
+++ b/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
@@ -30,6 +30,12 @@
     <Reference Include="..\..\refs\$(TargetFramework)\*.dll" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="$(SolutionDir)\Sarif\Schemata\Sarif.schema.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
   <Target Name="CopyTestData" AfterTargets="Build">
     <ItemGroup>
       <v1ConverterTestData Include="$(SolutionDir)\Sarif.FunctionalTests\v1\ConverterTestData\**\*.sarif" />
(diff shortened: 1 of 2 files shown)


src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -45,6 +51,5 @@
     <Copy SourceFiles="@(v2ConverterTestData)" DestinationFolder="$(OutputPath)v2\ConverterTestData\%(RecursiveDir)" />
     <Copy SourceFiles="@(v2DirectProducerTestData)" DestinationFolder="$(OutputPath)v2\DirectProducerTestData\%(RecursiveDir)" />
     <Copy SourceFiles="@(v2SpecExamples)" DestinationFolder="$(OutputPath)v2\SpecExamples\%(RecursiveDir)" />
-    <Copy SourceFiles="$(SolutionDir)\Sarif\Schemata\Sarif.schema.json" DestinationFolder="$(OutputPath)" />
```
lgolding: For some reason, performing a copy like this makes copies both in the actual output directory (e.g., bld\bin\AnyCPU_Release\Sarif.ValidationTests\net461) and unnecessarily in the parent directory (e.g., bld\bin\AnyCPU_Release\Sarif.ValidationTests). Using the <None> element with CopyToOutputDirectory metadata item does what I want.

Come to think of it, I should make this change for the other copied files as well. I see that there are extra unnecessary v1 and v2 subdirectories directly under bld\bin\AnyCPU_Release\Sarif.ValidationTests.

PR #956, opened 2018-07-23
Fix copying of validation test files

lgolding:
The files were being copied in such a way that they ended up not only in the framework binaries directory (e.g., `bld\bin\Sarif.ValidationTests\AnyCPU_Release\net461`, but also in the parent directory (e.g., `bld\bin\Sarif.ValidationTests\AnyCPU_Release`). I don't know why.

This change not only avoids the extra copy, but is much simpler, and takes advantage of the project system rather than using custom code.

### DIFF
--- a/scripts/BuildAndTest.ps1
+++ b/scripts/BuildAndTest.ps1
@@ -139,6 +139,8 @@ function New-NuGetPackageFromNuspecFile($project, $version, $suffix = "") {
     if ($LASTEXITCODE -ne 0) {
         Exit-WithFailureMessage $ScriptName "$project NuGet package creation failed."
     }
+
+    Write-Information "  Successfully created package '$BinRoot\NuGet\$Configuration\$Project.$version.nupkg'."
 }
 
 function New-NuGetPackages {
--- a/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
+++ b/src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
@@ -31,25 +31,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="$(SolutionDir)\Sarif\Schemata\Sarif.schema.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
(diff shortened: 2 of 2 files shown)


src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -31,25 +31,13 @@
...
 
   <ItemGroup>
-    <None Include="$(SolutionDir)\Sarif\Schemata\Sarif.schema.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
+    <None Include="$(SolutionDir)\Sarif\Schemata\Sarif.schema.json"
+          CopyToOutputDirectory="PreserveNewest" />
+    <None Include="$(SolutionDir)\Sarif.FunctionalTests\v1\**\*.sarif"
+          CopyToOutputDirectory="PreserveNewest"
+          LinkBase="v1" />
```
lgolding: `LinkBase` is a new feature added in .NET SDK 2.0 -- and a good thing, too, because without it, you end up with
```
OutputPath\
  ConverterTestFiles\
  ...
```
rather than what you want, which is
```
OutputPath\
  v1\
    ConverterTestFiles\
    ...
```
See Microsoft/msbuild#2949.

scripts/BuildAndTest.ps1
```diff
@@ -139,6 +139,8 @@ function New-NuGetPackageFromNuspecFile($project, $version, $suffix = "") {
     if ($LASTEXITCODE -ne 0) {
         Exit-WithFailureMessage $ScriptName "$project NuGet package creation failed."
     }
+
+    Write-Information "  Successfully created package '$BinRoot\NuGet\$Configuration\$Project.$version.nupkg'."
```
lgolding: `dotnet pack` produces this message for NuGet packages build from a project file. But we weren't producing any message for NuGet packages build from a .nuspec file. This message mimics the format (including the indentation) of the message from `dotnet pack`.

PR #968, opened 2018-07-25
Restore validation rule SARIF1013: End column must not be less than start column

lgolding:


scripts/ScriptUtilities.psm1
```diff
@@ -36,9 +36,12 @@ function Exit-WithFailureMessage($scriptName, $message) {
 $RepoRoot = $(Resolve-Path $PSScriptRoot\..).Path
 $SourceRoot = "$RepoRoot\src"
 $NuGetPackageRoot = "$SourceRoot\packages"
+$JsonSchemaPath = "$SourceRoot\Sarif\Schemata\Sarif.schema.json"
```
lgolding: We export a couple more variables, which, although so far they're only used in Sarif.Multitool.FunctionalTests\Update-Expected.ps1, seem generally useful.

src/Sarif/ConsoleLogger.cs
```diff
@@ -291,7 +291,7 @@ private string FormatNotificationMessage(Notification notification)
                 sb.Append(notification.RuleId + " : ");
             }
 
-            sb.Append(notification.Message);
+            sb.Append(notification.Message.Text);
```
lgolding: Existing bug that went unnoticed when we introduced the `Message` object. I noticed this when I ran the Multitool and it failed with an `ExecutionEngine` error, but the output didn't tell me what had gone wrong (it was a file not found error). The output just said "Microsoft.CodeAnalysis.Sarif.Message".

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -428,6 +427,20 @@ private void Visit(StackFrame frame, string framePointer)
...
+
+                for (int i = 0; i < threadFlowLocations.Length; ++i)
+                {
+                    ThreadFlowLocation threadFlowLocation = threadFlowLocations[i];
+                    string threadFlowLocationPointer = threadFlowLocationsPointer.AtIndex(i);
+
+                    Visit(threadFlowLocation, threadFlowLocationPointer);
+                }
+            }
         }
```
lgolding: An existing bug that went unnoticed because I hadn't yet restored a rule that expected to find an issue in a code flow location. I had the code to visit `codeFlow` and loop over its `threadFlow`s, and I had the code to visit a `threadFlowLocation` and loop over its `location`s -- but I was missing the part in the middle where you loop over the `threadFlowLocation`s in a `threadFlow`.

src/Sarif.Multitool.FunctionalTests/Sarif.Multitool.FunctionalTests.csproj
```diff
@@ -64,4 +63,6 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
   </ItemGroup>
+
+  <ProjectExtensions><VisualStudio><UserProperties TestData_4EndColumnMustNotBeLessThanStartColumn_4EndColumnEqualsStartColumnNoEndLine_1sarif__JSONSchema="http://json.schemastore.org/sarif-2.0.0" TestData_4EndColumnMustNotBeLessThanStartColumn_4EndColumnLessThanStartColumnInRelatedLocation_1sarif__JSONSchema="http://json.schemastore.org/sarif-2.0.0" TestData_4EndColumnMustNotBeLessThanStartColumn_4EndColumnGreaterThanStartColumn_1sarif__JSONSchema="http://json.schemastore.org/sarif-2.0.0" TestData_4EndColumnMustNotBeLessThanStartColumn_4EndColumnEqualsStartColumn_1sarif__JSONSchema="http://json.schemastore.org/sarif-2.0.0" /></VisualStudio></ProjectExtensions>
```
lgolding: ??? I think VS added this when I edited the SARIF files and typed "sarif-2.0.0" into the Schema field at the top of the editor. I'm going to remove this line.

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,34 @@
 ﻿<#
 .SYNOPSIS
-Update the expected output files for the SarifCli functional tests.
+Update the expected output files for the Sarif.Multitool functional tests.
 
 .DESCRIPTION
-This script runs "SarifCli validate" on each input log file in the SarifCli
+This script runs "Sarif.Multitool validate" on each input log file in the SarifCli
 functional tests TestData directory. For each file foo.sarif, it produces
```
lgolding: Nobody had used this script since the days of SarifCli. I had forgotten it existed.

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,34 @@
...
-    Configuration
-        The build configuration. Default: Debug
+.PARAMETER Configuration
+    The configuration of the build that produced Sarif.Multitool.exe.
+    Default: Debug
+
+.PARAMETER RuleName
+    The name of the rule whose files are to be regenerated.
+    Default: $null, meaning "all rules"
 #>
```
lgolding: Enhancement. Before, this script would update all the Expected files for all the validation rules. But I'm working on one restoring one rule at a time, so I don't want to modify the files for any other rules.

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,34 @@
...
-param([string] $Configuration = "Debug")
+param(
+    [string] $Configuration = "Debug",
+    [string] $RuleName = $null
+)
 
 $ErrorActionPreference = "Stop"
 
-$SarifExtension = ".sarif"
+Import-Module -Force ..\..\scripts\ScriptUtilities.psm1
```
lgolding: `$SarifExtension` and `$JsonSchemaPath` are now exported from `ScriptUtilities.psm1`.

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,34 @@
...
+
 $ExpectedSuffix = "_Expected$SarifExtension"
 
-$sdkPath = Join-Path -Resolve $PSScriptRoot "..\.."
-$sarifCliExe = Join-Path $sdkPath "bld\bin\SarifCli\AnyCPU_$Configuration\SarifCli.exe"
-$jsonSchemaPath = Join-Path $sdkPath "src\Sarif\Schemata\Sarif.schema.json"
+$multiToolExePath = "$BinRoot\Sarif.Multitool\AnyCPU_$Configuration\net461\Sarif.Multitool.exe"
```
lgolding: Take advantage of `$BinRoot` from `ScriptUtilities.psm1`.

src/Sarif.Multitool.FunctionalTests/Rules/EndColumnMustNotBeLessThanStartColumnTests.cs
```diff
@@ -26,18 +25,6 @@ public void EndColumnMustNotBeLessThanStartColumn_EndColumnGreaterThanStartColum
...
-        public void EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInAnalysisTarget()
-        {
-            Verify(new EndColumnMustNotBeLessThanStartColumn(), "EndColumnLessThanStartColumnInAnalysisTarget.sarif");
-        }
-
-        [Fact(DisplayName = nameof(EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInAnalysisTargetNoEndLine))]
-        public void EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInAnalysisTargetNoEndLine()
-        {
-            Verify(new EndColumnMustNotBeLessThanStartColumn(), "EndColumnLessThanStartColumnInAnalysisTargetNoEndLine.sarif");
-        }
```
lgolding: These two tests are obsolete because `result.AnalysisTarget` is no longer a `location` object with a `region` -- it's just a `fileLocation`.

src/Sarif.Multitool.FunctionalTests/Rules/EndColumnMustNotBeLessThanStartColumnTests.cs
```diff
@@ -50,10 +37,10 @@ public void EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnIn
             Verify(new EndColumnMustNotBeLessThanStartColumn(), "EndColumnLessThanStartColumnInRelatedLocation.sarif");
         }
 
-        [Fact(DisplayName = nameof(EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInResultFile))]
-        public void EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInResultFile()
+        [Fact(DisplayName = nameof(EndColumnMustNotBeLessThanStartColumn_EndColumnLessThanStartColumnInResultLocation))]
```
lgolding: Changed the name of this test because there's no longer such a thing as a "result file" -- there are just `location` objects on a `result`.

lgolding:

lgolding:

src/Sarif/ConsoleLogger.cs
michaelcfanning: Nuts. You see the subtlety here? sb.Append silently fell back from calling sb.Append(params string[]) to sb.AppendLine(params object[]) when the type changed. :) AppendLine calls Message.ToString() which by default just emits its fully qualified type name. We'd need an analyze that bans all calls to String.Concat(object)/StringBuilder.Append(object) to find this thing. 

Or author a doggone unit test! (which would have caught this specific case) :)

---
In reply to: [205180107](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205180107) [](ancestors = 205180107)

lgolding: Oh yes, the failure mode was quite clear.

---
In reply to: [205192274](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205192274) [](ancestors = 205192274,205180107)

lgolding:

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -428,6 +427,20 @@ private void Visit(StackFrame frame, string framePointer)
         private void Visit(ThreadFlow threadFlow, string threadFlowPointer)
         {
             Analyze(threadFlow, threadFlowPointer);
+
+            if (threadFlow.Locations != null)
+            {
+                ThreadFlowLocation[] threadFlowLocations = threadFlow.Locations.ToArray();
+                string threadFlowLocationsPointer = threadFlowPointer.AtProperty(SarifPropertyName.Locations);
```
michaelcfanning: >threadFlowLocationsPointer [](start = 23, length = 26)

What the heck is this? a string-based pointer to something? [i realize this question is orthogonal to your change :)]

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,38 @@
 ﻿<#
 .SYNOPSIS
-Update the expected output files for the SarifCli functional tests.
+Update the expected output files for the Sarif.Multitool functional tests.
 
 .DESCRIPTION
-This script runs "SarifCli validate" on each input log file in the SarifCli
+This script runs "Sarif.Multitool validate" on each input log file in the SarifCli
 functional tests TestData directory. For each file foo.sarif, it produces
```
michaelcfanning: >foo [](start = 51, length = 3)

Can you change this use of the term 'foo' to something else? 'example.sarif' or *.sarif

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,38 @@
...
 an output file foo_Expected.sarif, overwriting any existing file of that
```
michaelcfanning: >foo_Expected [](start = 15, length = 12)

fix it here, too. this term does not meet geopol standards

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,38 @@
...
 name in the TestData directory.
 
-.PARAMETERS
-    Configuration
-        The build configuration. Default: Debug
+.PARAMETER Configuration
+    The configuration of the build that produced Sarif.Multitool.exe.
+    Default: Debug
```
michaelcfanning: >Debug [](start = 13, length = 5)

is this the right default? we typically would want to run this script against a release build wouldn't we?

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
lgolding: Fixed. Changed to "example" as you suggest.

---
In reply to: [205215091](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205215091) [](ancestors = 205215091)

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
lgolding: Fixed here too.

---
In reply to: [205215240](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205215240) [](ancestors = 205215240)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif
```diff
@@ -21,65 +20,76 @@
...
+              "physicalLocation": {
+                "fileLocation": {
+                  "uri": "file:///C:/Code/sarif-sdk/src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation.sarif"
+                }
+              },
+              "message": {
+                "text": "Analyzing 'EndColumnLessThanStartColumnInRelatedLocation.sarif'..."
+              },
+              "level": "note",
+              "time": "2018-07-25T16:16:39.813Z"
```
michaelcfanning: >2018-07-25T16:16:39.813Z [](start = 23, length = 24)

will this work? i.e., if you emit time data in your baseline will it match a generated actual?

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInResultLocation_Expected.sarif
```diff
@@ -0,0 +1,95 @@
+{
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
+  "version": "2.0.0",
+  "runs": [
+    {
+      "tool": {
+        "name": "Sarif.Multitool",
```
michaelcfanning: >Sarif.Multitool [](start = 17, length = 15)

hm. no indicator here that we're invoking your validate verb. this is in line with a comment someone raised in the SARIF TC (might have been me), SARIF doesn't clearly describe plug-ins/optionally invoked functionality in the format. i.e., assumes tool -> single configuration/plug-in set.

this observation probably warrants a TC tracking issue, assuming we don't already have one

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
lgolding: I can certainly change it. It doesn't much matter because the Debug and Release builds will of course produce the same SARIF file. My reasoning (I don't remember if I wrote this script, but I mean, my reasoning as I was re-reading and modifying it), was that during development you're more likely to have a Debug build in hand, and it's usually only your run of BuildAndTest.cmd that produces a Release build. I'd like to leave as is, but feel free to push back.

---
In reply to: [205215433](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205215433) [](ancestors = 205215433)

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -1,28 +1,38 @@
...
+.PARAMETER RuleName
+    The name of the rule whose files are to be regenerated.
+    Default: $null, meaning "all rules"
 #>
 
-param([string] $Configuration = "Debug")
+param(
+    [string]
+    [ValidateSet("Debug", "Release")]
+    $Configuration = "Debug",
```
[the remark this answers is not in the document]
lgolding: As above, by design. Remember, it's the developer, not the end user, who runs this script. Feel free to push back.

---
In reply to: [205215746](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205215746) [](ancestors = 205215746)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif
lgolding: The unit tests do a "selective compare".

---
In reply to: [205217627](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205217627) [](ancestors = 205217627)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif
```diff
@@ -21,65 +20,76 @@
...
       },
       "files": {
-        "file:///G:/Code/sarif-sdk/src/Microsoft.CodeAnalysis.Sarif.Multitool/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation.sarif": {
-          "mimeType": "application/octet-stream"
+        "file:///C:/Code/sarif-sdk/src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation.sarif": {
+          "mimeType": "application/sarif-json"
         }
       },
-      "invocation": {
-        "commandLine": "\"G:\\Code\\sarif-sdk\\bld\\bin\\SarifCli\\AnyCPU_Debug\\SarifCli.exe\"  validate G:\\Code\\sarif-sdk\\src\\Microsoft.CodeAnalysis.Sarif.Multitool\\TestData\\EndColumnMustNotBeLessThanStartColumn\\EndColumnLessThanStartColumnInRelatedLocation.sarif -o G:\\Code\\sarif-sdk\\src\\Microsoft.CodeAnalysis.Sarif.Multitool\\TestData\\EndColumnMustNotBeLessThanStartColumn\\EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif -j G:\\Code\\sarif-sdk\\src\\Sarif\\Schemata\\Sarif.schema.json ",
```
michaelcfanning: >commandLine [](start = 9, length = 11)

why did this invocation data go away? are you stripping it elsewhere (as some of our scripts do)? is the multitool not emitting it by default (reasonably).

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif
michaelcfanning: got it. the problem here though is that the presence of this data will result in unwanted changes every time we rerun your refresh scripts.

---
In reply to: [205217627](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205217627) [](ancestors = 205217627)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInRelatedLocation_Expected.sarif
lgolding: It moved up to become `invocations[]`.

---
In reply to: [205218670](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205218670) [](ancestors = 205218670)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnNotSpecified_Expected.sarif
```diff
@@ -1,46 +1,43 @@
...
         {
-          "id": "MSG001.AnalyzingTarget",
-          "physicalLocation": {
-            "uri": "file:///G:/Code/sarif-sdk/src/Microsoft.CodeAnalysis.Sarif.Multitool/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnNotSpecified.sarif"
-          },
-          "message": "Analyzing 'EndColumnNotSpecified.sarif'...",
-          "level": "note",
-          "time": "2016-08-29T22:59:10.501Z"
+          "startTime": "2018-07-25T16:16:41.527Z",
+          "endTime": "2018-07-25T16:16:42.173Z",
```
michaelcfanning: >07-25T16:16:42.173Z [](start = 27, length = 19)

same note. your matching is resilient to this non-deterministic data, but git is not: every time we run your script, we will generate a new timestamp. other tools use regex to strip this data out. might be some code in the binskim tree if it isn't in SARIF. search the powershell scripts for 'endTime'

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInResultLocation_Expected.sarif
lgolding: You're right. We do have the `commandLine` property of the `invocation` object (we're not populating it, but we could). 

---
In reply to: [205218073](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205218073) [](ancestors = 205218073)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnNotSpecified_Expected.sarif
lgolding: Hmm, that's true -- it does complicate the history. Do we have a "--deterministic" option on the Driver?

---
In reply to: [205219160](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205219160) [](ancestors = 205219160)

michaelcfanning: found it. search for endTime. no one is proud of this code. if you improve on it or add other non-deterministic things, let me know so we can update binskim. also note: i will shortly add an option to strip all non-deterministic data, so this may all be a wash in a couple of weeks anyway. so you might want to just grab this stuff in the meantime or frankly ignore the note. your instincts will be sound on how to proceed i am sure.

[https://github.com/Microsoft/binskim/blob/master/src/Test.FunctionalTests.BinSkim.Driver/UpdateBaselines.ps1](https://github.com/Microsoft/binskim/blob/master/src/Test.FunctionalTests.BinSkim.Driver/UpdateBaselines.ps1)

---
In reply to: [205219160](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205219160) [](ancestors = 205219160)

lgolding: If you're in the process of adding an option (I assume you mean a new Multitool verb, such as "Sarif.Multitool make-deterministic", then I will leave this as is.

---
In reply to: [205220055](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205220055) [](ancestors = 205220055,205219160)

src/Sarif.Multitool.FunctionalTests/Update-Expected.ps1
```diff
@@ -36,15 +46,20 @@ function Get-InputFiles($ruleDirectoryPath) {
 
 function Update-ExpectedFile($inputPath) {
     $outputPath = $inputPath.Replace($SarifExtension, $ExpectedSuffix)
-    & $sarifCliExe validate $inputPath -o $outputPath -j $jsonSchemaPath
+    & $multiToolExePath validate $inputPath --output $outputPath --json-schema $JsonSchemaPath --pretty-print --force --verbose
```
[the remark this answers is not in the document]
lgolding: I think that's a good idea, and if we add those verbs to Multitool, I'll invoke them here.

---
In reply to: [205216293](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205216293) [](ancestors = 205216293)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnNotSpecified_Expected.sarif
michaelcfanning: Sagacious.

---
In reply to: [205221602](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205221602) [](ancestors = 205221602,205220055,205219160)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
lgolding: The extension method `AtProperty` appends a property name, for instance `/runs[0]` -> `/runs[0]/locations`, and the method `AtIndex` appends an array index, for instance,  `/runs[0]/locations` -> `/runs[0]/locations[1]`

---
In reply to: [205216379](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205216379) [](ancestors = 205216379,205214627)

src/Sarif.Multitool.FunctionalTests/TestData/EndColumnMustNotBeLessThanStartColumn/EndColumnLessThanStartColumnInResultLocation_Expected.sarif
lgolding: On further reflection, I don't think the lack of "plug-ins" support is the issue here. That discussion was about the scenario where you have v1 of tool MyTool.exe, and it accepts plugins like rule sets, and you give it v2 of MySecurityRules.dll and v3 of MyAccessibilityRules.dll. Maybe you deploy them to the tool installation directory, and they're not even mentioned on the command line). Then SARIF has no place for you to mention the plug-ins or what versions they are.

Sure, a plug-in could add a whole new verb, but that's not the case here: `validate` is built-in.

Here, I think the existing `invocation.commandLine` would meet the case, but I don't think it's worth holding this merge for it.

---
In reply to: [205219537](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205219537) [](ancestors = 205219537,205218073)

lgolding: I filed [https://github.com/oasis-tcs/sarif-spec/issues/207](https://github.com/oasis-tcs/sarif-spec/issues/207), "Support tool plug-ins."

---
In reply to: [205224952](https://github.com/Microsoft/sarif-sdk/pull/968#discussion_r205224952) [](ancestors = 205224952,205219537,205218073)

PR #971, opened 2018-07-25
Restore validation rule SARIF1012: End line must not be less than start line

lgolding:


lgolding:

lgolding:

src/Sarif.Multitool.FunctionalTests/Rules/EndLineMustNotBeLessThanStartLineTests.cs
```diff
@@ -38,16 +38,10 @@ public void EndLineMustNotBeLessThanStartLine_EndLineNotSpecified()
...
+        public void EndLineMustNotBeLessThanStartLine_EndLineLessThanStartLineInResultLocation()
         {
-            Verify(new EndLineMustNotBeLessThanStartLine(), "EndLineLessThanStartLineInAnalysisTarget.sarif");
-        }
-
-        [Fact(DisplayName = nameof(EndLineMustNotBeLessThanStartLine_EndLineLessThanStartLineInResultFile))]
-        public void EndLineMustNotBeLessThanStartLine_EndLineLessThanStartLineInResultFile()
-        {
-            Verify(new EndLineMustNotBeLessThanStartLine(), "EndLineLessThanStartLineInResultFile.sarif");
+            Verify(new EndLineMustNotBeLessThanStartLine(), "EndLineLessThanStartLineInResultLocation.sarif");
```
lgolding: As in the previous PR: "`analysisTarget`" is now a `fileLocation`, so it has no `region`, so this rule does not apply to it, so remove that test. And "result file" no longer appears in the spec; the test applies to locations within a result, hence "ResultLocation".

src/Sarif.Multitool.FunctionalTests/TestData/EndLineMustNotBeLessThanStartLine/EndLineEqualsStartLine_Expected.sarif
```diff
@@ -1,46 +1,43 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-1.0.0",
```
lgolding: All the _Expected files were regenerated by Update-Expected.ps1. The console output from that tool shows me that the results are correct (I also spot-checked the resulting files).

PR #974, opened 2018-07-27
Comprehensive region properties

michaelcfanning:
@lgolding @EasyRhin0 I will annotate this PR tomorrow am.


src/ReleaseHistory.md
```diff
@@ -188,4 +188,5 @@
 ## **v2.0.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0))
 * Convert object model to conform to SARIF v2 CSD.1 draft specification
 * Distinguish textual vs. binary file persistence in rewrite option (and allow for both in multitool rewrite verb)
-*   NOTE: the change above introduces a command-line breaking change. --persist-file-contents is now renamed to --persist-text-contents
+*   NOTE: the change above introduces a command-line breaking change. --persist-file-contents is now renamed to --insert
```
lgolding: >insert [](start = 114, length = 6)

Good catch.

src/Sarif/ExtensionMethods.cs
```diff
@@ -15,6 +15,37 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        /// <summary>
+        /// Attempt to reconstruct a URI, if appropriate, using Run instance 
+        /// originalUriBaseId and uriBaseId properties. If this method cannot
+        /// successfully reconstitute an absolute URI, it will return false
+        /// and populate 'resolvedUri' with the input value.
+        /// </summary>
+        /// <param name="run"></param>
```
lgolding: >>< [](start = 29, length = 2)

Either provide text or remove empty doc comments.

src/ReleaseHistory.md
```diff
@@ -188,4 +188,5 @@
...
+* Add Regions as possible qualifier to --insert command
```
lgolding: >command [](start = 48, length = 7)

I think this word should be "option". Besides that being the more common term, our code already uses "command" to refer to the implementation of what the CommandLine NuGet package refers to as a "verb". For example, we implement the `rewrite` verb with the class `RewriteCommand`.

src/Sarif/ExtensionMethods.cs
```diff
@@ -15,6 +15,37 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+        /// <param name="uri"></param>
+        /// <param name="resolvedUri"></param>
+        /// <returns></returns>
+        public static bool TryResolveUri(this Run run, FileLocation fileLocation, out Uri resolvedUri)
+        {
+            resolvedUri = fileLocation.Uri;
+
+            // We can't restore any absolute URIs unless someone has
+            // deconstructed them using uriBaseId + originalUriBaseIds
+            if (run.OriginalUriBaseIds == null) { return false; }
```
lgolding: >false [](start = 57, length = 5)

I'm pondering this. I think it's more common, when returning `false` from a `TryXxx` method, to set the `out` parameter to `null`. I can't deny the usefulness of populating it with the input URI, but I don't know if it violates "least surprise". Your call.

lgolding: >OriginalUriBaseIds [](start = 20, length = 18)

But... I think this is a bug. If `fileLocation.Uri.IsAbsoluteUri`, then you don't need any help from URI base ids, original or otherwise. So:
```
resolvedUri = fileLocation.Uri;

if (fileLocation.Uri.IsAbsoluteUri) { return true; }

if (run.OriginalUriBaseIds == null || string.IsNullOrEmpty(fileLocation.UriBaseId)) { return false; }

...
```

src/Sarif/ExtensionMethods.cs
```diff
@@ -15,6 +15,37 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+
+            if (!string.IsNullOrEmpty(fileLocation.UriBaseId) &&
+                !fileLocation.Uri.IsAbsoluteUri)
+            {
+                if (run.OriginalUriBaseIds.TryGetValue(fileLocation.UriBaseId, out Uri originalUriBaseId))
```
lgolding: >Uri [](start = 83, length = 3)

I didn't know you could declare an `out` parameter inline like this. Nice. Is it new in C# 7?

src/Sarif/ExtensionMethods.cs
```diff
@@ -15,6 +15,37 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+        /// Attempt to reconstruct a URI, if appropriate, using Run instance 
+        /// originalUriBaseId and uriBaseId properties. If this method cannot
+        /// successfully reconstitute an absolute URI, it will return false
+        /// and populate 'resolvedUri' with the input value.
+        /// </summary>
+        /// <param name="run"></param>
+        /// <param name="uri"></param>
+        /// <param name="resolvedUri"></param>
+        /// <returns></returns>
+        public static bool TryResolveUri(this Run run, FileLocation fileLocation, out Uri resolvedUri)
```
lgolding: >Run [](start = 46, length = 3)

You can reduce coupling by just passing the the `originalUriBaseIds` dictionary, rather than the entire `Run` object. That's all you need from the `Run`.

lgolding: I understand that then, what you have is no longer an extension method on `Run`. But -- hah! -- then it could be an extension method on `fileLocation`, which I actually think is more natural:
```
public static bool TryResolveUri(this FileLocation fileLocation, IDictionary<string, string> originalUriBaseIds, out Uri resolvedUri)
```

---
In reply to: [205979411](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979411) [](ancestors = 205979411)

src/Sarif/OptionallyEmittedData.cs
```diff
@@ -53,6 +57,12 @@ public enum OptionallyEmittedData
...
         // that produce static analysis results. It is useful in some contexts, therefore,
         // to strip this information. 
-        EnvironmentVariables = 0x80
+        EnvironmentVariables = 0x80,
+
+        // A special enum value that indicates that insertion should overwrite any existing
+        // information in the SARIF log file. In the absence of this setting, any existing
+        // data that would otherwise have been overwritten by the insert operation will
+        // be preserved.
+        OverwriteExistingData = 0x100
```
lgolding: >0x100 [](start = 32, length = 5)

Ultra-nit: Since this is a special value, perhaps `0x80000000`?

src/Sarif/Sarif.csproj
```diff
@@ -38,8 +38,6 @@
     <Content Include="Schemata\Sarif.schema.json" />
 
     <!-- Link=false: Don't display this item in Solution Explorer. -->
-    <Content Include="..\ReleaseHistory.md"
```
lgolding: 1. If you remove it from the project file, it won't automatically be added to the NuGet package. That's why it's here. I suppose the comment on the line above should have mentioned that.
2. If you do decide to remove it, remove that comment.

src/Sarif/Core/FileLocation.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public partial class FileLocation
```
lgolding: >FileLocation [](start = 25, length = 12)

Oh, right! All of our object model classes are `partial`. So, if you accept my proposed redesign in `TryGetUri`, you could put that method here, rather than making it an extension method.

src/Sarif/Visitors/AddFileReferencesVisitor.cs
```diff
@@ -19,11 +19,29 @@ public override Run VisitRun(Run node)
...
+
+            string uriText = Uri.EscapeUriString(fileLocation.Uri.ToString());
 
-            string uriText = Uri.EscapeUriString(node.FileLocation.Uri.ToString());
+            if (!string.IsNullOrEmpty(fileLocation.UriBaseId))
+            {
+                // See EXAMPLE 3 of 3.11.13.2 'Property Names' of
+                // SARIF v2 'files' property specification 
+                uriText = "#" + fileLocation.UriBaseId + "#" + uriText;
+            }
```
lgolding: >} [](start = 12, length = 1)

Very nice!

src/Sarif.FunctionalTests/Sarif.FunctionalTests.csproj
```diff
@@ -25,6 +25,7 @@
     <ProjectReference Include="..\Sarif.Converters\Sarif.Converters.csproj" />
     <ProjectReference Include="..\Sarif.Driver\Sarif.Driver.csproj" />
     <ProjectReference Include="..\Sarif\Sarif.csproj" />
+    <ProjectReference Include="..\Sarif.UnitTests\Sarif.UnitTests.csproj" />
```
lgolding: >UnitTests [](start = 56, length = 9)

If `Sarif.UnitTests` has code that `Sarif.FunctionalTests` needs, then that code belongs in `Sarif.TestUtilities`, and you should reference that project instead.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+    public class InsertOptionalDataVisitor : SarifRewritingVisitor
+    {
+        internal IFileSystem s_fileSystem = new FileSystem();
+
+        private Run _run;
+        private bool _overwriteExistingData;
+        private FileRegionsCache _fileRegionsCache;
+        private OptionallyEmittedData _dataToInsert;
+        
+        public InsertOptionalDataVisitor(OptionallyEmittedData dataToInsert, bool overwriteExistingData = true)
```
lgolding: >bool overwriteExistingData = true [](start = 77, length = 33)

You don't need the argument (and I see you don't use it) because `OptionallyEmittedData` has a flag for it.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            _dataToInsert = dataToInsert;
+            _overwriteExistingData = dataToInsert.Includes(OptionallyEmittedData.OverwriteExistingData);
```
lgolding: >_overwriteExistingData [](start = 12, length = 22)

Likewise, you don't need a separate instance variable since `_dataToInsert` has the flag you need.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class InsertOptionalDataVisitor : SarifRewritingVisitor
+    {
+        internal IFileSystem s_fileSystem = new FileSystem();
+
+        private Run _run;
+        private bool _overwriteExistingData;
+        private FileRegionsCache _fileRegionsCache;
+        private OptionallyEmittedData _dataToInsert;
```
lgolding: >  [](start = 15, length = 1)

`readonly`

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+            if (node != null)
+            {
+                bool scrapeFileReferences = _dataToInsert.Includes(OptionallyEmittedData.Hashes)      ||
+                                            _dataToInsert.Includes(OptionallyEmittedData.TextFiles)   ||
+                                            _dataToInsert.Includes(OptionallyEmittedData.BinaryFiles);
+
+                if (scrapeFileReferences)
+                {
+                    var visitor = new AddFileReferencesVisitor();
+                    visitor.VisitRun(node);
```
lgolding: >visitor.VisitRun(node) [](start = 20, length = 22)

Oh my! That's very clever (and at the same time quite clear, unlike most "clever" code).

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+                }
+
+                if (node.Files != null)
+                {
+                    var keys = node.Files.Keys.ToArray();
```
lgolding: >.ToArray() [](start = 46, length = 10)

You don't need the `ToArray` since all you do with `keys` is enumerate over it.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+                        var value = node.Files[key];
+                        if (value != null)
+                        {
+                            node.Files[key] = VisitDictionaryValueNullChecked(key, value);
+                        }
+                    }
+                }
+            }
+
+            if (node.Results != null)
```
lgolding: >if (node.Results != null) [](start = 12, length = 25)

This whole block needs to be inside the top-level `if (node != null)` block.

src/Sarif.UnitTests/Sarif.UnitTests.csproj
```diff
@@ -30,4 +30,31 @@
     <ProjectReference Include="..\Sarif.Converters\Sarif.Converters.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="TestData\InsertOptionalDataVisitor\CoreTests.sarif">
```
lgolding: >sarif [](start = 63, length = 5)

I suggest we use embedded resources here.

lgolding: >None [](start = 5, length = 4)

There's no need to mention the files individually, because you are using `Update`, which [allows]() you to "Modify metadata of a file that was included using a glob". Just:
```
<None Update="TestData\**\*.sarif" CopyToOutputDirectory="true" />
```

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
+  "version": "2.0.0",
```
lgolding: Can we --pretty-print these files? They're a bit hard to read with so much nesting and multiple properties on a single line.

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
```diff
@@ -0,0 +1,22 @@
...
+  "runs": [
+    {
+      "tool": { "name": "Sarif.UnitTests",
+        "fullName": "Sarif.UnitTests.1.3.3.0-beta", "version": "1.3.3.0", "semanticVersion": "1.3.3", "sarifLoggerVersion": "1.5.22.0", "language": "en-US"},
```
lgolding: >1.5.22.0 [](start = 125, length = 8)

Is this _still_ the logger version?

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
```diff
@@ -0,0 +1,22 @@
...
+      "originalUriBaseIds": { "SRCROOT": "file:///e:/src/sarif-sdk/" },
+      "results": [        
+          {"ruleId": "TEST2001", "message": { "text": "This result is at the beginning of the file and selects 'Region_At_Start_Of_File'." },
+          "locations": [{"physicalLocation": { "fileLocation": { "uri": "src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/InsertOptionalDataVisitor.txt", "uriBaseId": "SRCROOT" },
```
lgolding: >src/Sarif.UnitTests/TestData [](start = 73, length = 28)

You could make these files shorter by including this in the URI base id, and renaming it from `SRCROOT` to `TEST_DATA_ROOT`.

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
+  "version": "2.0.0",
+  "runs": [
+    {
+      "tool": { "name": "Sarif.UnitTests",
+        "fullName": "Sarif.UnitTests.1.3.3.0-beta", "version": "1.3.3.0", "semanticVersion": "1.3.3", "sarifLoggerVersion": "1.5.22.0", "language": "en-US"},
+      "originalUriBaseIds": { "SRCROOT": "file:///e:/src/sarif-sdk/" },
```
lgolding: >e [](start = 50, length = 1)

Do you have an E: drive? Is that why the tests pass for you but not on AppVeyor?

src/Sarif.UnitTests/FileDiffingTests.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.Globalization;
+using System.IO;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.UnitTests
+{
+    public class FileDiffingTests
```
lgolding: >FileDiffingTests [](start = 17, length = 16)

Move this file to `Sarif.TestUtilities`, and now, since it's a base class, I suggest `public abstract class FileDiffingTestBase`.

src/Sarif.UnitTests/FileDiffingTests.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        protected string GenerateDiffCommand(string expected, string actual)
+        {
+            expected = Path.GetFullPath(expected).Replace(@"\", @"\\");
+            actual = Path.GetFullPath(actual).Replace(@"\", @"\\");
+
+            string beyondCompare = TryFindBeyondCompare();
+            if (beyondCompare != null)
```
lgolding: >beyondCompare [](start = 16, length = 13)

I know this is existing code, but... NO! NO! NO! This code means that a unit test could potentially give different results on machines where Beyond Compare is installed than on machines where it's not installed.

src/Sarif.UnitTests/FileDiffingTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                return String.Format(CultureInfo.InvariantCulture, "\"{0}\" \"{1}\" \"{2}\" /title1=Expected /title2=Actual", beyondCompare, expected, actual);
+            }
+
+            return String.Format(CultureInfo.InvariantCulture, "windiff \"{0}\" \"{1}\"", expected, actual);
```
lgolding: >windiff [](start = 64, length = 7)

It's worse than that. These tests won't run at all on a non-Windows machine.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -26,7 +26,10 @@ public static int Run(MergeOptions mergeOptions)
                 // Reformat the SARIF log if we need to.
```
lgolding: >Reformat the SARIF log if we need to. [](start = 19, length = 37)

This comment is obsolete.

src/Sarif.UnitTests/FileRegionsCacheTests.cs
```diff
@@ -319,20 +303,11 @@ public void FileRegionsCache_PopulatesNullRegion()
...
-                FileLocation = new FileLocation()
-                {
-                    Uri = uri
-                },
-                Region = null
-            };
-
-            fileRegionsCache.PopulateTextRegionProperties(physicalLocation, false);
-            physicalLocation.Region.Should().BeNull();
+            Region result = fileRegionsCache.PopulateTextRegionProperties(inputRegion: null, uri: uri, populateSnippet: false);
```
lgolding: >result [](start = 19, length = 6)

I suggest "`region`" since "`result`" means something in SARIF.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        private ITestOutputHelper _outputHelper;
+
+        private static string GetTestDirectory(string subdirectory = "")
+        {
+            return Path.GetFullPath(Path.Combine(@".\TestData", subdirectory));
+        }
+
+        private static string GetProductTestDataDirectory(string subdirectory = "")
+        {
+            return Path.GetFullPath(Path.Combine(@"..\..\..\..\..\\src\Sarif.UnitTests\TestData", subdirectory));
```
lgolding: >..\..\..\..\..\\ [](start = 51, length = 16)

This is another reason you might fail on AppVeyor. You're assuming that the directory from which the tests are run is next door to the `src` directory.

Do you really need to pull anything from the `src` directory? Isn't the project file already copying the files you need to the output directory? (If it isn't, it could.)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using Microsoft.CodeAnalysis.Sarif.UnitTests;
+using Newtonsoft.Json;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class InsertOptionalDataVisitorTests : FileDiffingTests
+    {
+        // To rebaseline all test files set this value to true and rerun the testa
```
lgolding: >To rebaseline all test files set this value to true and rerun the testa [](start = 11, length = 71)

Neither the `Update-Expected.ps1` script in `Sarif.Multitool.FunctionalTests` nor `UpdateBaselines.ps` in `Sarif.FunctionalTests` require a special build to update the expected files. Can't this one be managed in the same way?

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            return Path.GetFullPath(Path.Combine(@".\TestData", subdirectory));
+        }
+
+        private static string GetProductTestDataDirectory(string subdirectory = "")
+        {
+            return Path.GetFullPath(Path.Combine(@"..\..\..\..\..\\src\Sarif.UnitTests\TestData", subdirectory));
+        }
+
+        public InsertOptionalDataVisitorTests(ITestOutputHelper outputHelper)
```
lgolding: >ITestOutputHelper [](start = 46, length = 17)

In email you asked about adding diagnostic output to unit tests. That's what [`ITestOutputHelper`](https://xunit.github.io/docs/capturing-output) is for.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            _outputHelper = outputHelper;
+        }
+
+        [Theory]
+        [InlineData(OptionallyEmittedData.Hashes)]
+        [InlineData(OptionallyEmittedData.TextFiles)]
+        [InlineData(OptionallyEmittedData.RegionSnippets)]
+        [InlineData(OptionallyEmittedData.ComprehensiveRegionProperties)]
+        [InlineData(OptionallyEmittedData.ComprehensiveRegionProperties | OptionallyEmittedData.RegionSnippets | OptionallyEmittedData.TextFiles | OptionallyEmittedData.Hashes)]
+        public void InsertOptionaDataVisitorTests_InsertsOptionalDataForCommonConditions(OptionallyEmittedData optionallyEmittedData)
```
lgolding: >InsertOptionaDataVisitorTests [](start = 20, length = 29)

Nit: To make the test name read in English ("The InsertOptionalDataVistor inserts optional data for common conditions"), the test name would be `InsertOptionaDataVisitor_InsertsOptionalDataForCommonConditions`.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            expectedFileName = Path.Combine(testDirectory, expectedFileName);
+            actualFileName = Path.Combine(testDirectory, actualFileName);
+            inputFileName = Path.Combine(testDirectory, inputFileName);
+
+            File.Exists(inputFileName).Should().BeTrue();
+
+            SarifLog actualLog;
+
+            JsonSerializerSettings settings = new JsonSerializerSettings()
```
lgolding: >JsonSerializerSettings [](start = 50, length = 22)

Future: This particular `JsonSerializerSettings` object is initialized all over the place (and is even given a name (not appropriate for the current file) `SarifTransformerUtilities.JsonSettingsV2`). We should DRY it out in a separate PR.

src/Sarif.UnitTests/FileDiffingTests.cs
lgolding: Ok, ok, I see now that the result is only used in a message, telling the user how to see the differences between actual and expected files.

---
In reply to: [205981120](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205981120) [](ancestors = 205981120)

src/Sarif.UnitTests/FileDiffingTests.cs
lgolding: And ok, ok, it's not that the tests won't run on non-Windows, it's that the message will be wrong.

---
In reply to: [205981167](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205981167) [](ancestors = 205981167)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            if (insertRegionSnippets || populateRegionProperties)
+            {
+                Region result;
+
+                _fileRegionsCache = _fileRegionsCache ?? new FileRegionsCache(_run);
+                
+                // If we can resolve a file location to a newly constructed
+                // absolute URI, we will prefer that
+                if (!_run.TryResolveUri(node.FileLocation, out Uri resolvedUri))
```
lgolding: >_run.TryResolveUri [](start = 21, length = 18)

Remember my comment wondering whether it was right for you to populate the `out` parameter if the `Try` routine returns false? And I wondered if it violated "least surprise" for you to do so?

Well apparently it did, because here you surprised even yourself and re-filled `resolvedUri` with the exact value that your `Try` routine did :-)

So I think you _should_ change that initialization in `TryResolveUri`.

src/Sarif/ExtensionMethods.cs
lgolding: Update: See my comment in InsertOptionalDataVisitor.cs. This _does_ violate least surprise, and I think you should change the initialization to `null`.

---
In reply to: [205979160](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979160) [](ancestors = 205979160)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                    // We also rewrite the checked in test baselines
+                    File.WriteAllText(expectedFileName, actualSarif);
+                }
+                else
+                {
+                    File.WriteAllText(actualFileName, actualSarif);
+
+                    string errorMessage = "Expanding optional data for input '{0}' produced unexpected results for the following options: '{1}'.";
+                    sb.AppendLine(string.Format(CultureInfo.CurrentCulture, errorMessage, inputFileName, optionallyEmittedData));
```
lgolding: >sb.AppendLine(string.Format(CultureInfo.CurrentCulture [](start = 20, length = 54)

`StringBuilder.AppendFormat(IFormatProvider, ...)` is the overload you need.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
lgolding: Ok, I understand why. Both of those scripts update _all_ the expected files, and since they don't take care to make the files deterministic, they all change. We either need to
1) Make the expected files deterministic, or
2) Have an elaborate script that identifies the failing tests (say, by reading the .trx), identifies the files that need to be updated (say, by a naming convention, based on the name of the test), and updates only the necessary files.

---
In reply to: [205982194](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205982194) [](ancestors = 205982194)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: indeed. also removed var use in the enumeration in preference of explicit reference to FileData type.

---
In reply to: [205980346](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205980346) [](ancestors = 205980346)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: nice catch

---
In reply to: [205980403](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205980403) [](ancestors = 205980403)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: Did that. Also changed helper name to 'TryReconstructAbsoluteUri'. returns true if we can reconstruct an absolute uri or if the input URI was already absolute. 

---
In reply to: [205985485](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205985485) [](ancestors = 205985485)

src/Sarif/ExtensionMethods.cs
michaelcfanning: Yes. Absolutely right.

---
In reply to: [205979511](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979511) [](ancestors = 205979511,205979411)

src/Sarif/ExtensionMethods.cs
michaelcfanning: yes, almost! you also need to reset resolvedUri to null in cases where you are returning false.

---
In reply to: [205979291](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979291) [](ancestors = 205979291)

src/Sarif/OptionallyEmittedData.cs
```diff
@@ -33,10 +33,14 @@ public enum OptionallyEmittedData
         // addition to the stard/end property pairs, the CharOffset and CharLength properties
         // can be used to specify a text region. This enum value either comprehensively 
         // populates all possible region properties or reduces all regions to a minimal form.
-        Regions = 0x08,
+        ComprehensiveRegionProperties = 0x08,
```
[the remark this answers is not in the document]
michaelcfanning: Also considered 'ExpandedRegionProperties'

---
In reply to: [205979843](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979843) [](ancestors = 205979843)

src/Sarif/Sarif.csproj
michaelcfanning: Good catch, not sure how this got stripped. Restored it and added a comment about the line adding the referenced file to the NuGet, to try to ensure it remains in future. I suspect some random act in the IDE scrubbed. 

---
In reply to: [205979938](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979938) [](ancestors = 205979938)

src/Sarif/Core/FileLocation.cs
michaelcfanning: right! This helper makes sense right on the file location class...

---
In reply to: [205979970](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205979970) [](ancestors = 205979970)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: well, at the expense of having to figure out how to handle the newline addition, as embedded chars or use of Environment.NewLine (which increases line length)

---
In reply to: [205988006](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205988006) [](ancestors = 205988006)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: Sure, just a typo 'OptionaData' is not a real thing :)

---
In reply to: [205982390](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205982390) [](ancestors = 205982390)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: indeed and i found it as you can see

---
In reply to: [205982298](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205982298) [](ancestors = 205982298)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: This helper is used to write data back to the source directory, to refresh baselines. Should only be used in an ad hoc rebaseline scenario on the dev box (not in appveyor). Closing by 'non-obvious design'. Will decorate the helper with a comment.

---
In reply to: [205981836](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205981836) [](ancestors = 205981836)

src/Sarif.UnitTests/FileDiffingTests.cs
michaelcfanning: Right, part of this test methodology is to assist the developer into jumping into a diffing experience in as efficient a way as possible. Currently, we don't emit these diff commands on the console when testing (a gap). also, you have to click through to the 'Output' pane in the IDE, another gap. Ideally, you get a command you can scrape easily for bulk comparison of all diffs (with this sort of data-driven testing approach).

---
In reply to: [205982688](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205982688) [](ancestors = 205982688,205981167)

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
michaelcfanning: Nerts. of course.

---
In reply to: [205980932](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205980932) [](ancestors = 205980932)

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests.sarif
michaelcfanning: no. but i don't want to set a standard to update this information every time the logger version changes. 

---
In reply to: [205980807](https://github.com/Microsoft/sarif-sdk/pull/974#discussion_r205980807) [](ancestors = 205980807)

PR #975, opened 2018-07-27
Remove v2 DirectProducer test data files.

lgolding:
We will repopulate with genuine v2 direct producers as they become available. BinSkim will probably be first.


src/Sarif.FunctionalTests/v2/DirectProducerTestData/BinSkim/ProduceSarifTestCases.cmd
```diff
@@ -1,7 +0,0 @@
-set TOOL=d:\src\binskim\bld\bin\BinSkim.Driver\x64_Release\BinSkim.exe
```
lgolding: Hmmm. Perhaps I shouldn't remove this?

PR #976, opened 2018-07-28
Clean up and fix intermittent failure in generation of VersionConstants.cs

lgolding:
This change fixes an intermittent failure in the generation of `VersionConstants.cs`, and cleans up other aspects of the generation and usage of that file. There are several changes:

- The call to `Set-Content` in `New-VersionConstantsFile.ps` intermittently failed with the message "Stream was not readable." Others have encountered this; see this [StackOverflow post](https://stackoverflow.com/questions/48953767/set-content-sporadically-fails-with-stream-was-not-readable).

    The solution proposed in StackOverflow was to replace `Set-Content` with a call to `[System.IO.File]::WriteAllText`.

- We now generate `VersionConstants.cs` only once, in `Sarif.csproj`. All other projects refer to the resulting public class `Microsoft.CodeAnalysis.Sarif.VersionConstants`.

- We removed two spurious copies of `VersionConstants.cs` that somehow got checked in, one in `Sarif` and one in `Sarif.Driver`. But we are now _generating_ `VersionConstants.cs` in the `Sarif` directory, and checking it in, the same way we handle the `Autogenerated` object model files.

- I moved the "Clean" operation from `BuildAndTest.ps1` to `BeforeBuild.ps1`, and made it the very first thing that happens.

- We were inconsistent in our casing of `Prerelease` _vs._ `PreRelease`. According to the [APA Style Blog](http://blog.apastyle.org/apastyle/2012/03/title-case-and-sentence-case-capitalization-in-apa-style.html), it would be spelled "Pre-release" in a title, so I settled on "`Prerelease`". Feel free to change it back the other way, but whichever you choose, be consistent.


scripts/New-VersionConstantsFile.ps1
```diff
@@ -24,17 +25,17 @@ namespace $namespace
...
         public const string FileVersion = AssemblyVersion + ".0";
-        public const string Version = AssemblyVersion + PreRelease;
+        public const string Version = AssemblyVersion + Prerelease;
     }
 }
 "@
 
-if (-not (Test-Path $outputDirectory)) {
-    New-Item -ItemType Directory -Path $outputDirectory | Out-Null
-}
```
lgolding: We're generating it in the source directory, so no need to check and create the directory first.

lgolding:

lgolding:

scripts/BeforeBuild.ps1
```diff
@@ -7,10 +7,19 @@
     being performed inline in BuildAndTest.cmd, because AppVeyor cannot run BuildAndTest.
     AppVeyor only allows you to specify the project to build, and a script to run before
     the build step. So that is how we have factored the build scripts.
+.PARAMETER NoClean
+    Do not remove the outputs from the previous build.
```
lgolding: Add missing doc comments.

scripts/New-VersionConstantsFile.ps1
```diff
@@ -18,23 +19,21 @@ $versionConstantsFileContents =
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Reflection;
```
lgolding: This hasn't been needed in ages.

scripts/New-VersionConstantsFile.ps1
```diff
@@ -18,23 +19,21 @@ $versionConstantsFileContents =
...
-        public const string Version = AssemblyVersion + PreRelease;
+        public const string Version = AssemblyVersion + Prerelease;
     }
 }
 "@
 
-if (-not (Test-Path $outputDirectory)) {
-    New-Item -ItemType Directory -Path $outputDirectory | Out-Null
-}
+$outputFile = "$SourceRoot\Sarif\VersionConstants.cs"
```
lgolding: We generate the file in the source directory now, so there's no need to check and create the directory first.

src/build.common.props
```diff
@@ -45,11 +45,4 @@ whether they use the VS 2017 project system or the old project system.
   <ItemGroup>
     <Compile Include="$(MsBuildThisFileDirectory)CommonAssemblyInfo.cs" />
   </ItemGroup>
-
-  <Target BeforeTargets="PrepareForBuild" Name="CreateVersionConstantsFile">
-    <Exec Command="powershell -ExecutionPolicy RemoteSigned $(MsBuildThisFileDirectory)..\scripts\New-VersionConstantsFile.ps1 $(IntermediateOutputPath) $(AssemblyName)" />
-    <ItemGroup>
-      <Compile Include="$(IntermediateOutputPath)VersionConstants.cs" />
-    </ItemGroup>
-  </Target>
```
lgolding: We do this only in Sarif.csproj now. And, since we generate the file in the source directory, you'll note that in Sarif.csproj, we don't need to mention the `<Compile>` item for VersionConstants.cs. (In fact if you do mention it, the project system complains that it's a duplicate item, because it already knows about it.)

src/Sarif.Driver/VersionConstants.cs
```diff
@@ -1,12 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT
```
lgolding: This file should never have been checked in.

src/Sarif/VersionConstants.cs
```diff
@@ -1,12 +1,13 @@
-// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT        
-// license. See LICENSE file in the project root for full license information.
```
lgolding: This file _was_ mistakenly checked in, but now we're _generating_ it here, so it's all good.

PR #979, opened 2018-08-01
Add multiline context regions to output

michaelcfanning:
@Evmaus-MS @EasyRhin0 @lgolding

### DIFF
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -99,6 +99,32 @@ namespace Microsoft.CodeAnalysis.Sarif
             return region;
         }
 
+        internal Region ConstructMultilineContextSnippet(Region inputRegion, Uri uri)
+        {
+            if (inputRegion == null || inputRegion.IsBinaryRegion)
+            {
+                // Context snippets are relevant only for textual regions.
+                return null;
+            }
+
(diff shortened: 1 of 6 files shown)


src/Sarif/FileRegionsCache.cs
```diff
@@ -99,6 +99,32 @@ private Region PopulateTextRegionProperties(NewLineIndex lineIndex, Region input
...
+            if (newLineIndex == null)
+            {
+                return null;
+            }
+
+            int maxLineNumber = newLineIndex.MaximumLineNumber;
+
+            // Currently, we just grab a single line before and after the region start
+            // and end lines, respectively. In the future, we could make this configurable.
+            var region = new Region()
```
Evmaus-MS: Do we want to allow invalid input regions to get more invalid here?

(E.x. if start line is < 1 or if end line is > max line number, we seem to make the problem worse.  Yes, the file is already invalid per the spec, but tools have bugs...)

Rest of it LGTM

michaelcfanning: this is a very good point but note that this is an internal helper. we have a single caller that comprehensively populates the region for us. i filed an issue in #980 to do more testing of invalid data

---
In reply to: [207065625](https://github.com/Microsoft/sarif-sdk/pull/979#discussion_r207065625) [](ancestors = 207065625)

Issue #983, opened 2018-08-03
Weakly typed objects


michaelcfanning:
We need updates to codegen hints to resolve these weakly typed objects

EdgeTraversal.cs:        public object FinalState { get; set; }
GraphTraversal.cs:        public object InitialState { get; set; }
Resources.cs:        public object MessageStrings { get; set; }
RuleConfiguration.cs:        public object Parameters { get; set; }
ThreadFlowLocation.cs:        public object State { get; set; }

michaelcfanning:
Please uncomment code in this file on resolving this issue. Also need to add some relevant tests to CoreTests.sarif (introducing some notifications that refer to strings in run.resources will be sufficient)

E:\src\sarif-sdk\src\Sarif\Visitors\InsertOptionalDataVisitor.cs

michaelcfanning:
fixed

PR #984, opened 2018-08-03
Add option to 'flatten' messages, saving consumers the complexity of …

michaelcfanning:
…following SARIF resource string look-up semantics. @Evmaus-MS @cfaucon @lgolding @EasyRhin0


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -143,5 +145,36 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
...
+        public override Result VisitResult(Result node)
+        {
+            _ruleId = node.RuleId;
+            node = base.VisitResult(node);
+            _ruleId = null;
+
+            return node;
+        }
+
+        public override Message VisitMessage(Message node)
```
michaelcfanning: >VisitMessage [](start = 32, length = 12)

SARIF string look-up semantics are very complex due to the format supporting localization. We can refer to an external resources file. We have a global strings table. We have per-rule strings tables. And so, this message string flattener will be very useful for pipeline developers. We need to extract this logic into a separate helper, I think, so that viewers can also easily get access to it. I'll do this when I tackle #983 below.

src/Sarif.UnitTests/Sarif.UnitTests.csproj
```diff
@@ -48,7 +48,10 @@
     <None Update="TestData\InsertOptionalDataVisitor\CoreTests_ContextCodeSnippets.sarif">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="TestData\InsertOptionalDataVisitor\CoreTests_Hashes+TextFiles+ComprehensiveRegionProperties+RegionSnippets+ContextCodeSnippets.sarif">
+    <None Update="TestData\InsertOptionalDataVisitor\CoreTests_FlattenedMessages.sarif">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
michaelcfanning: >CopyToOutputDirectory [](start = 7, length = 21)

Larry, would love to deploy your recommendation to provide a global copy reference here. The code you provided worked on the command-line but failed for me within the IDE however. So I continue the inferior pattern here. I don't know whether this is a bug in the IDE or if it relates to some other configuration gap...

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -59,13 +60,16 @@ private void RunTest(string testDirectory, string inputFileName, OptionallyEmitt
             string optionsNameSuffix = "_" + NormalizeOptionallyEmittedDataToString(optionallyEmittedData);
 
             string expectedFileName = inputFileName + optionsNameSuffix + ".sarif";
-            string actualFileName = inputFileName + optionsNameSuffix + ".actual.sarif";
```
michaelcfanning: >actual [](start = 74, length = 6)

We now persist all generated files to an 'Actual' directory. This allows me to now produce a single diff command-line that compares output from the entire test suite (by comparing the expected vs. actual dirs). this is helpful when introducing changes that will flow to all test files.

Evmaus-MS:
>     <None Update="TestData\InsertOptionalDataVisitor\CoreTests.sarif">

Can we use embedded resources, a la the TestLogs above?  Not blocking, but it'd be good to be consistent.

---
Refers to: src/Sarif.UnitTests/Sarif.UnitTests.csproj:42 in db85c1e. [](commit_id = db85c1e51a29ce0d64798baedb35277fd88263af, deletion_comment = False)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -140,6 +146,13 @@ private void RunTest(string testDirectory, string inputFileName, OptionallyEmitt
             ValidateResults(sb.ToString());
         }
 
+        private string GenerateDirectoryDiffCommand(string v1, string v2)
+        {
+            // "C:\Program Files\Beyond Compare 4\BComp.exe" "E:\\src\\sarif-sdk\\bld\\bin\\Sarif.UnitTests\\AnyCPU_Debug\\netcoreapp2.0\\TestData\\InsertOptionalDataVisitor\" "E:\\src\\sarif-sdk\\bld\\bin\\Sarif.UnitTests\\AnyCPU_Debug\\netcoreapp2.0\\TestData\\InsertOptionalDataVisitor\Actual\"
+
+            throw new NotImplementedException();
```
Evmaus-MS: Can we delete this?  Or Implement it?

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -143,5 +145,36 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
...
+                if ((bool)_run.Resources?.Rules.TryGetValue(_ruleId, out rule))
+                {
+                    node.Text = rule.Format(node.MessageId, node.Arguments.ToArray());
+                }
+                // Can't enable this code yet, due to MessageStrings being weakly typed
+                //
+                // https://github.com/Microsoft/sarif-sdk/issues/983
+                //else if ((bool)_run.Resources?.MessageStrings?.TryGetValue(node.MessageId, out string formatString))
+                //{
+                //    node.Text = string.Format(CultureInfo.CurrentCulture, formatString, node.Arguments);
```
lgolding: >Arguments [](start = 95, length = 9)

`.ToArray()`

src/Sarif.UnitTests/Sarif.UnitTests.csproj
lgolding: Ok, thanks for trying.

---
In reply to: [207575996](https://github.com/Microsoft/sarif-sdk/pull/984#discussion_r207575996) [](ancestors = 207575996)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
lgolding: Got it. The file names of the "actual" files can now be the same as the file names of the "expected" files, because they are in different directories.

---
In reply to: [207576383](https://github.com/Microsoft/sarif-sdk/pull/984#discussion_r207576383) [](ancestors = 207576383)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -143,5 +145,36 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
...
+        }
+
+        public override Message VisitMessage(Message node)
+        {
+            if ((node.Text == null || _dataToInsert.Includes(OptionallyEmittedData.OverwriteExistingData)) &&
+                _dataToInsert.Includes(OptionallyEmittedData.FlattenedMessages))
+            {
+                Rule rule = null;
+                //string formatString = null;
+                if ((bool)_run.Resources?.Rules.TryGetValue(_ruleId, out rule))
```
lgolding: >(bool) [](start = 20, length = 6)

This fails at runtime if the `Nullable<bool>` that results from this expression actually is `null`.

This alternative always works:
```
if (_run.Resources?.Rules.TryGetValue(_ruleId, out rule) == true)
```

PR #989, opened 2018-08-04
Weak types

michaelcfanning:
I encountered several weakly typed things in v2 and one in v1 and have updated them.

@Evmaus-MS @lgolding @EasyRhin0

### DIFF
--- a/src/Sarif/CodeGenHints.json
+++ b/src/Sarif/CodeGenHints.json
@@ -332,6 +332,18 @@
       "kind": "DictionaryHint"
     }
   ],
+  "RuleConfiguration.Parameters": [
+    {
+      "kind": "DictionaryHint",
+      "arguments": {
+        "valueTypeName": "SerializedPropertyInfo",
+        "namespaceName": "Microsoft.CodeAnalysis.Sarif.Readers",
+        "comparisonKind": "ObjectEquals",
+        "hashKind": "ScalarReferenceType",
+        "initializationKind": "SimpleAssign"
+      }
+    }
+  ],
   "run": [
     {
       "kind": "BaseTypeHint",
--- a/src/Sarif/CodeGenHints.json
+++ b/src/Sarif/CodeGenHints.json
@@ -1,23 +1,4 @@
 {
-  "threadFlowLocation": [
-    {
-      "kind": "BaseTypeHint",
-      "arguments": {
-        "baseTypeNames": [
-          "PropertyBagHolder"
-        ]
-      }
-    }
-  ],
-  "ThreadFlowLocation.Importance": [
-    {
-      "kind": "EnumHint",
-      "arguments": {
-        "typeName": "ThreadFlowLocationImportance",
-        "description": "Values specifying the importance of an \"threadFlowLocation\" within the \"codeFlow\" in which it occurs"
-      }
-    }
-  ],
   "*.Properties": [
     {
       "kind": "DictionaryHint",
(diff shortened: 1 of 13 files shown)


src/Sarif/CodeGenHints.json
```diff
@@ -434,6 +442,30 @@
...
+      "kind": "DictionaryHint"
+    }
+  ],
+  "ThreadFlowLocation.Importance": [
+    {
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "ThreadFlowLocationImportance",
+        "description": "Values specifying the importance of an \"threadFlowLocation\" within the \"codeFlow\" in which it occurs"
+      }
```
lgolding: >} [](start = 6, length = 1)

Bug: You need to supply `arguments.memberNames`.

lgolding: >} [](start = 6, length = 1)

Bug: As a result of this declaration, I would expect the generated code to include a new file Autogenerated\ThreadFlowLocationImportance.cs. (See Autogenerated\Baseline.cs for an example of what you'd expect.)

I don't know if it's missing because you neglected to omit it, or whether the code generator decided not to bother emitting it because you left out `arguments.memberNames`. Either way, it needs to be there.

lgolding: And I don't know why AppVeyor doesn't catch this. The AppVeyor build should break because `ThreadFlowLocation` declares a member `Importance` of type `ThreadFlowLocationImportance`, but that type isn't declared anywhere.

---
In reply to: [208047570](https://github.com/Microsoft/sarif-sdk/pull/989#discussion_r208047570) [](ancestors = 208047570)

michaelcfanning: Note that I only moved this code and did not author it. (for ThreadFlowLocation.Importance). The enum is created in my clean build, with correct enum values. Are you sure that your autogen doesn't provide correct behavior by default? perhaps arguments.memberNames is only required if we insist on creating an additional member name that's not explicit in the schema. Or if some renaming heuristics are broken?

---
In reply to: [208046177](https://github.com/Microsoft/sarif-sdk/pull/989#discussion_r208046177) [](ancestors = 208046177)

michaelcfanning: CodeGen is working here. I looked at jschema code, the enum member name is derived on line 83 of the following file:

[https://github.com/Microsoft/jschema/blob/master/src/Json.Schema.ToDotNet/EnumGenerator.cs](https://github.com/Microsoft/jschema/blob/master/src/Json.Schema.ToDotNet/EnumGenerator.cs)

---
In reply to: [208048947](https://github.com/Microsoft/sarif-sdk/pull/989#discussion_r208048947) [](ancestors = 208048947,208047570)

src/Sarif/VersionOne/Autogenerated/AnnotatedCodeLocationVersionOne.cs
```diff
@@ -111,7 +111,7 @@ public SarifNodeKindVersionOne SarifNodeKindVersionOne
         /// A dictionary, each of whose keys specifies a variable or expression, the associated value of which represents the variable or expression value. For an annotation of kind 'continuation', for example, this dictionary might hold the current assumed values of a set of global variables.
         /// </summary>
         [DataMember(Name = "state", IsRequired = false, EmitDefaultValue = false)]
-        public object State { get; set; }
+        public IDictionary<string, string> State { get; set; }
```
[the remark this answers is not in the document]
michaelcfanning: Indeed. 

---
In reply to: [208046870](https://github.com/Microsoft/sarif-sdk/pull/989#discussion_r208046870) [](ancestors = 208046870)

michaelcfanning:
@lgolding merging, as the only issue you pointed out was in code I hadn't changed (and which is working as required, per stepping through via jschema todotnet exe.

PR #992, opened 2018-08-07
Pr notes + flatten notifications

michaelcfanning:
@EasyRhin0 @cfaucon @lgolding @Evmaus-MS

### DIFF
--- a/src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests_ComprehensiveRegionProperties.sarif
+++ b/src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests_ComprehensiveRegionProperties.sarif
@@ -11,6 +11,30 @@
         "sarifLoggerVersion": "1.5.22.0",
         "language": "en-US"
       },
+      "invocations": [
+        {
+          "toolNotifications": [
+            {
+              "message": {
+                "messageId": "NotificationMessage",
+                "arguments": [
+                  "tool notification"
+                ]
+              }
+            }
+          ],
+          "configurationNotifications": [
+            {
+              "message": {
+                "messageId": "NotificationMessage",
+                "arguments": [
+                  "configuration notification"
+                ]
+              }
+            }
+          ]
+        }
+      ],
       "originalUriBaseIds": {
         "TESTROOT": "file:///REPLACED_AT_TEST_RUNTIME/src/Sarif.UnitTests/TestData/"
       },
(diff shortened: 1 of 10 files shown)


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -55,19 +55,7 @@ public override Run VisitRun(Run node)
                 }
             }
 
-            if (node.Results != null)
```
michaelcfanning: >Results [](start = 21, length = 7)

results objects are mutated in place, so we don't need this. also by not calling the base visit, we weren't visiting other constructs (such as those that contain messages but aren;t result instances, like notifications)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -161,18 +149,20 @@ public override Message VisitMessage(Message node)
...
             {
                 Rule rule = null;
-                //string formatString = null;
-                if ((bool)_run.Resources?.Rules.TryGetValue(_ruleId, out rule))
+                string formatString = null;
+                if (_ruleId != null &&
+                    _run.Resources?.Rules.TryGetValue(_ruleId, out rule) == true)
+                {
+                    node.Text = node.Arguments?.Count > 0 
+                        ? rule.Format(node.MessageId, node.Arguments.ToArray())
```
lgolding: >Format [](start = 31, length = 6)

It occurs to me that, since we own the method `Rule.Format`, we could have it accept the arguments as `IEnumerable` rather than as an array. Then you wouldn't have to do the `ToArray` in each caller, you could do it once in `Rule.Format`

The downside is that all the .NET `Format` methods we know and love take a `params object[]`, so callers probably wouldn't think to use the labor-saving `IEnumerable` feature and would continue to call `ToArray`.

Up to you.

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests_ComprehensiveRegionProperties.sarif
```diff
@@ -11,6 +11,30 @@
         "sarifLoggerVersion": "1.5.22.0",
         "language": "en-US"
       },
+      "invocations": [
```
lgolding: >invocations [](start = 7, length = 11)

For the sake of brevity, I suggest we not include optional elements that aren't related to the test...

Oh... is your idea to test that (for example), when _only_ the "populate comprehensive region properties" flag is set, that it does _not_ also flatten the notifications? That makes sense. Ok. Stream of consciousness comment withdrawn.

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/CoreTests_FlattenedMessages.sarif
```diff
@@ -11,6 +11,32 @@
         "sarifLoggerVersion": "1.5.22.0",
         "language": "en-US"
       },
+      "invocations": [
+        {
+          "toolNotifications": [
+            {
+              "message": {
+                "text": "A notification was raised. Notification kind was: 'tool notification'.",
```
lgolding: >text [](start = 17, length = 4)

Yup. And here, when you _do_ set that flag, it _does_ flatten the notifications.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: we should fix it.

---
In reply to: [208325436](https://github.com/Microsoft/sarif-sdk/pull/992#discussion_r208325436) [](ancestors = 208325436)

lgolding: Change looks good.

---
In reply to: [208381881](https://github.com/Microsoft/sarif-sdk/pull/992#discussion_r208381881) [](ancestors = 208381881,208325436)

PR #994, opened 2018-08-07
Changing TryReconstructAbsoluteUri to respect if the uri is already a…

cfaucon:
…bsolute instead of returning a null uri.  Fixing TSLintConverter length calculation to not add 1.

### DIFF
--- a/src/Sarif.Converters/TSLintConverter.cs
+++ b/src/Sarif.Converters/TSLintConverter.cs
@@ -94,7 +94,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 CharOffset = entry.StartPosition.Position
             };
 
-            int length = entry.EndPosition.Position - entry.StartPosition.Position + 1;
+            int length = entry.EndPosition.Position - entry.StartPosition.Position;
             region.CharLength = length > 0 ? length : 0;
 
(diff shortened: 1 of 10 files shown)


src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -94,7 +94,7 @@ internal Result CreateResult(TSLintLogEntry entry)
                 CharOffset = entry.StartPosition.Position
             };
 
-            int length = entry.EndPosition.Position - entry.StartPosition.Position + 1;
```
michaelcfanning: > + 1 [](start = 82, length = 4)

Arg!

src/Sarif/Core/FileLocation.cs
```diff
@@ -20,7 +20,7 @@ public partial class FileLocation
         /// <returns></returns>
         public bool TryReconstructAbsoluteUri(IDictionary<string, Uri> originalUriBaseIds, out Uri resolvedUri)
         {
-            resolvedUri = null;
+            resolvedUri = this.Uri.IsAbsoluteUri ? this.Uri : null;
```
michaelcfanning: >IsAbsoluteUri [](start = 35, length = 13)

we missed a unit test here. please drop one in quickly? just a single test for this single change. will create a placeholder for us to add to.

PR #995, opened 2018-08-07
Update all packages to meet nuget ship requirements

michaelcfanning:
@Evmaus-MS


src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -4,13 +4,15 @@
     <id>Sarif.Multitool</id>
     <version>$version$</version>
     <title>Microsoft SARIF Multitool (includes SARIF SDK)</title>
```
Evmaus-MS: >(includes SARIF SDK) [](start = 37, length = 20)

Is this strictly true anymore?  It distributes the DLL, but we probably don't want to mark this as "including" the SARIF SDK (it doesn't package it as a library).

Issue #934, opened 2018-07-06
Reformatting visitor does not consult node.Encoding when embedded file contents


michaelcfanning:
See VisitDictionaryValueNullChecked(string key, FileData node)

PR #998, opened 2018-08-08
Fix UpdateBaselines #997 + issue #934

EasyRhinoMSFT:
Also persist file length property

### DIFF
--- a/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
+++ b/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
@@ -129,11 +129,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 
                 Encoding encoding = null;
 
-                try
+                if (!string.IsNullOrWhiteSpace(node.Encoding))
                 {
-                    encoding = Encoding.GetEncoding(node.Encoding);
+                    try
+                    {
+                        encoding = Encoding.GetEncoding(node.Encoding);
+                    }
+                    catch (ArgumentException) { }
                 }
-                catch (ArgumentException) { }
 
                 int length = node.Length;
                 node = FileData.Create(uri, _dataToInsert, node.MimeType, encoding: encoding);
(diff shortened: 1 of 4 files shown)


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -126,9 +127,17 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
             {
                 fileLocation.TryReconstructAbsoluteUri(_run.OriginalUriBaseIds, out Uri uri);
 
-                // TODO: we should convert node.Encoding to a .NET equivalent and pass it here
-                // https://github.com/Microsoft/sarif-sdk/issues/934
-                node = FileData.Create(uri, _dataToInsert, node.MimeType, encoding: null);
+                Encoding encoding = null;
+
+                try
```
michaelcfanning: >try [](start = 16, length = 3)

test for non-null node.Encoding first, and don't enter this code path for that case. will prevent us raising/catching exceptions unnecessarily.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -126,9 +127,17 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
...
+                {
+                    encoding = Encoding.GetEncoding(node.Encoding);
+                }
+                catch (ArgumentException) { }
+
+                int length = node.Length;
+                node = FileData.Create(uri, _dataToInsert, node.MimeType, encoding: encoding);
+                node.Length = length;
```
michaelcfanning: >length [](start = 30, length = 6)

add simple unit tests for the nice changes you have made. they don't need to be complex or super comprehensive, just covering your code delta.

michaelcfanning: this change is approved, check in when you fix the suggestions above. 

---
In reply to: [208705942](https://github.com/Microsoft/sarif-sdk/pull/998#discussion_r208705942) [](ancestors = 208705942)

PR #1001, opened 2018-08-09
Add option to RebaseUri verb to allow rebase of all relative uris to provided path

cfaucon:
Add option to RebaseUri verb to allow rebase of all relative uris to provided path

### DIFF
--- a/src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
+++ b/src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
@@ -75,9 +75,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 
             Run newRun = rebaseUriVisitor.VisitRun(oldRun);
 
-            newRun.Properties.Should().ContainKey(RebaseUriVisitor.BaseUriDictionaryName);
-
-            Dictionary<string, Uri> baseUriDictionary = RebaseUriVisitor.DeserializePropertyDictionary(newRun.Properties[RebaseUriVisitor.BaseUriDictionaryName]);
+            IDictionary<string, Uri> baseUriDictionary = newRun.OriginalUriBaseIds;
 
             baseUriDictionary.Should().ContainKey("SRCROOT");
             baseUriDictionary.Should().ContainValue(new Uri(@"C:\src\root"));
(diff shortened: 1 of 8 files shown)


src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -64,38 +85,75 @@ public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
...
             Run newRun = base.VisitRun(node);
 
-            if (newRun.Files != null)
-            {
-                FixFiles(newRun);
-            }
-            
+            newRun.Files = _files;
+
             if (node.Properties == null)
```
Evmaus-MS: If OriginalUriBaseIds is now a first class object, we don't need to touch the propertybag here.

Evmaus-MS:
>         internal void FixFiles(Run run)

I think this is dead code now?  We should clean up up if it isn't used.

---
Refers to: src/Sarif/Visitors/RebaseUriVisitor.cs:166 in 48b1733. [](commit_id = 48b1733ab0eabfb71b1fb5b9e35e259df9733a86, deletion_comment = False)

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -95,50 +93,21 @@ public void RebaseUriVisitor_VisitRun_UpdatesBaseUriDictionaryWhenPresent()
...
 
-            newRun.Properties.Should().ContainKey(RebaseUriVisitor.BaseUriDictionaryName);
-
-            Dictionary<string, Uri> baseUriDictionary = RebaseUriVisitor.DeserializePropertyDictionary(newRun.Properties[RebaseUriVisitor.BaseUriDictionaryName]);
+            IDictionary<string, Uri> baseUriDictionary = newRun.OriginalUriBaseIds;
 
             baseUriDictionary.Should().ContainKey(srcRoot);
             baseUriDictionary[srcRoot].Should().BeEquivalentTo(srcRootUri);
             baseUriDictionary.Should().ContainKey(bldRoot);
             baseUriDictionary[bldRoot].Should().BeEquivalentTo(bldRootUri);
```
Evmaus-MS: Nit--this pattern was just because we had to retrieve the dictionary to get it as a dictionary and that was overly verbose.  Can we just do the checks on the property since it's now part of the run?

Evmaus-MS:
A few nits around dead code and small refactorings, but in general this looks good.

PR #1018, opened 2018-08-15
Restore three more validation rules and remove one

lgolding:
This change restores the following validation rules:

- `SARIF1003`: URI must be valid.
- `SARIF1009`: Step values must form one-based sequence.
- `SARIF1014`: `UriBaseId` requires relative URI.

It also removes the following rule:

- `SARIF1010`: Step must appear only on code flow locations.
    This rule was necessary because `annotatedCodeLocations` appeared in many places, but some of their properties, including the `step` property, were only applicable to code flows. Now, with the [`location` object redesign ](oasis-tcs/sarif-spec#130), the `annotatedCodeLocation` object is gone, all but one of the places where it was previously used now use the improved general-purpose `location` object, and `step` only appears on the single-purpose `threadFlowLocation` object.

This description might look very similar to a previous PR that also restored three rules and remove one. Don't worry, this is a different set of rules.

After this, there is only one rule left to restore. I could have done it in this PR but it's already too big and it's the end of the day.


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif/Errors.cs
```diff
@@ -74,7 +74,7 @@ public static void LogExceptionLoadingPdb(IAnalysisContext context, Exception ex
                     exception,
                     false,
                     context.TargetUri.GetFileName(),
-                    context.Rule.Name));
+                    context.Rule.Name.Text));
```
lgolding: >Text [](start = 38, length = 4)

Existing bug, introduced when we introduced the message object. I noticed it when an unrelated bug in one of the rules I was restoring caused one of the messages in this file to be generated.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -40,12 +36,21 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>MultitoolResources.resx</DependentUpon>
     </Compile>
+    <Compile Update="Rules\RuleResources.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>RuleResources.resx</DependentUpon>
+    </Compile>
```
lgolding: >Compile [](start = 6, length = 7)

When I created the new-style project files, I missed this. As a result, the embedded resource wasn't being updated when I edited the `.resx` file.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -23,10 +23,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Remove="Rules\StepMustAppearOnlyInCodeFlowLocations.cs" />
-    <Compile Remove="Rules\StepValuesMustFormOneBasedSequence.cs" />
-    <Compile Remove="Rules\UriBaseIdRequiresRelativeUri.cs" />
-    <Compile Remove="Rules\UrisMustBeValid.cs" />
     <Compile Remove="Rules\UseAbsolutePathsForNestedFileUriFragments.cs" />
```
lgolding: >UseAbsolutePathsForNestedFileUriFragments [](start = 27, length = 41)

This is the one rule that remains to restore.

src/Sarif.Multitool/SarifPropertyName.cs
```diff
@@ -6,17 +6,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
     public static class SarifPropertyName
     {
         public const string AnalysisTarget = "analysisTarget";
+        public const string AnalysisToolLogFiles = "analysisToolLogFiles";
```
lgolding: >AnalysisToolLogFiles [](start = 28, length = 20)

A bunch of new properties I have to visit to fully implement the restored rules in SARIF v2.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -57,18 +57,38 @@ protected void LogResult(string jPointer, string formatId, params string[] args)
                 RuleUtilities.BuildResult(DefaultLevel, Context, region, formatId, argsWithPointer));
         }
 
+        protected virtual void Analyze(Attachment attachment, string attachmentPointer)
```
lgolding: >attachment [](start = 50, length = 10)

A bunch of new objects I have to visit to implement the restored rules in SARIF v2. If you haven't seen this code before, here's how it works:

There are a bunch of private `Visit` methods, below, that are responsible for walking down the object tree. Each `Visit` method first calls one of these virtual `Analyze` methods to validate the object at this level, and then drills down into the object model by calling the `Visit` method on its member objects. Individual `Rule` classes override the appropriate `Analyze` method(s) to implement their validation logic.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -113,31 +144,47 @@ protected virtual void Analyze(ThreadFlowLocation threadFlowLocation, string thr
...
+            Analyze(log, logPointer);
+
             if (log.Runs != null)
             {
                 Run[] runs = log.Runs.ToArray();
                 string runsPointer = logPointer.AtProperty(SarifPropertyName.Runs);
 
                 for (int i = 0; i < runs.Length; ++i)
                 {
-                    Run run = runs[i];
```
lgolding: >run [](start = 24, length = 3)

Here and throughout, I decided to tighten up the code by removing these temporaries and evaluating the expressions at the point of call to the next `Visit` method. I think it's just as clear.

src/Sarif.Multitool/Rules/UriBaseIdRequiresRelativeUri.cs
```diff
@@ -29,47 +33,14 @@ protected override IEnumerable<string> FormatIds
             }
         }
 
-        protected override void Analyze(
```
lgolding: >Analyze [](start = 32, length = 7)

There used to be several objects with pairs of properties `uri`, `uriBaseId`. Now, only the `fileLocation` object has this pair, so we only have to check for this condition in one place. We do have to make sure to visit all the `fileLocation` objects in the model.

src/Sarif.Multitool.FunctionalTests/TestData/StepValuesMustFormOneBasedSequence/InvalidStepValues.sarif
```diff
@@ -1,6 +1,5 @@
-{
-  "$schema": "http://json.schemastore.org/sarif-1.0.0",
-  "version": "1.0.0",
+{  "$schema": "http://json.schemastore.org/sarif-2.0.0",
```
lgolding: >$schema [](start = 4, length = 7)

One hardly knows what to say about diffs this large. You're probably better off looking at "right side only" and treating all the test files as new.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
lgolding: Really we ought to be able to code gen this visitor.

---
In reply to: [210441226](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210441226) [](ancestors = 210441226)

src/Sarif/Errors.cs
michaelcfanning: this is a tough one. if we leave CreateNotification as weakly typed, everyone will continue to fall into this trap. The introduction of the Message object makes it super easy to recompile with an error. If we make our helpers more strongly typed, the call sites are more verbose, everyone needs to explicitly call ToString(). Your thoughts on how to address?




---
In reply to: [210439893](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210439893) [](ancestors = 210439893)

michaelcfanning: Check it out, this bug shipped in current BinSkim. Bad news.

              "message": {
                "text": "'MixedMode_x86_VS2013_MissingPdb.dll' was not evaluated for check 'Microsoft.CodeAnalysis.Sarif.Message' because its PDB could not be loaded. (E_PDB_NOT_FOUND (File not found))"
              },

---
In reply to: [210450798](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210450798) [](ancestors = 210450798,210439893)

src/Sarif.Multitool/Rules/StepValuesMustFormOneBasedSequence.cs
```diff
@@ -33,46 +38,46 @@ protected override IEnumerable<string> FormatIds
...
+            ThreadFlowLocation[] locations,
+            JArray threadFlowLocationsArray,
+            string threadFlowLocationsPointer)
         {
-            JObject[] annotatedCodeLocationObjects = annotatedCodeLocationArray.Children<JObject>().ToArray();
+            JObject[] threadFlowLocationObjects = threadFlowLocationsArray.Children<JObject>().ToArray();
 
             for (int i = 0; i < locations.Length; ++i)
             {
                 // Only report "invalid step value" for locations that actually specify
```
michaelcfanning: >Only report "invalid step value" for locations that actually specify [](start = 19, length = 68)

gee,all this makes me realize we have a step priority that you must populate but which can always be inferred simply by tracking the index value of the object array.

if a producer doesn't provide it, any consumer that wanted it could manufacture it. any consumer could correct the bad output from the tool. makes me think we should just cut this property and put a note in the spec: it may be helpful to compute a 'step' value for display. this can be inferred from the array index, etc..

src/Sarif.Multitool/Rules/UrisMustBeValid.cs
```diff
@@ -29,69 +35,80 @@ protected override IEnumerable<string> FormatIds
...
-            string parentPointer,
-            string childPropertyName = SarifPropertyName.Uri)
+        private void AnalyzeUri(string uri, string pointer)
         {
             if (uri != null)
             {
                 try
                 {
-                    Uri fileUri = new Uri(uri.OriginalString);
+                    Uri fileUri = new Uri(uri);
```
michaelcfanning: >Uri fileUri [](start = 20, length = 11)

Why aren't you using Uri.TryCreate here? Suggest doing so to prevent preventable exceptions.

src/Sarif.Multitool.FunctionalTests/Rules/StepValuesMustFormOneBasedSequenceTests.cs
```diff
@@ -31,5 +31,11 @@ public void StepValuesMustFormOneBasedSequence_InvalidStepValues()
         {
             Verify(new StepValuesMustFormOneBasedSequence(), "InvalidStepValues.sarif");
         }
+
+        [Fact(DisplayName = nameof(StepValuesMustFormOneBasedSequence_MultipleThreadFlows))]
```
michaelcfanning: >Fact [](start = 9, length = 4)

thank you for not using a theory for this! facts are faster to process

src/Sarif/Errors.cs
lgolding: Ugh.

---
In reply to: [210451144](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210451144) [](ancestors = 210451144,210450798,210439893)

src/Sarif.Multitool/Rules/UrisMustBeValid.cs
lgolding: I didn't think of it. Much better. Will change.

---
In reply to: [210451903](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210451903) [](ancestors = 210451903)

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRelatedLocation.sarif
```diff
@@ -10,25 +10,13 @@
         {
           "ruleId": "TST0001",
           "level": "error",
-          "formattedRuleMessage": {
-            "formatId": "Default"
-          },
-          "locations": [
```
michaelcfanning: >locations [](start = 11, length = 9)

why did this go?

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInInvocationAttachment.sarif
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
+  "version": "2.0.0",
+  "runs": [
+    {
+      "tool": {
+        "name": "CodeScanner"
```
michaelcfanning: it would be great if a test file had some marker to explicitly denote all expected failures. the property bag could be utilized. what's your general approach here for making this explicit? where are you putting positive test cases? sorry i am so late to this code, we could get on the phone for a chat about it if you'd like to fill me in that way.

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRunGraphNode_Expected.sarif
```diff
@@ -0,0 +1,93 @@
...
+          "ProductName": "Microsoft SARIF SDK"
+        }
+      },
+      "results": [
+        {
+          "ruleId": "SARIF1014",
+          "level": "error",
+          "message": {
+            "arguments": [
+              "/runs/0/graphs/0/nodes/0/location/physicalLocation/fileLocation/uri",
```
michaelcfanning: >runs/0/graphs/0/nodes/0/location/physicalLocation/fileLocation/uri" [](start = 16, length = 67)

this is a logical location presumably! right? :) so why aren't you specifying the fullyQualifiedName or the logical locations table. heh. would be useful test data for some of our UX explorations.

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRelatedLocation.sarif
lgolding: Each test file has only and exactly the elements it needs to demonstrate the proposition stated by the test name. I don't need `result.locations` to demonstrate that I can detect an unwanted absolute URI in `relatedLocations`. I'm trying to keep the test content as small and focused as possible.

---
In reply to: [210452611](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452611) [](ancestors = 210452611)

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/UrisAreRelative.sarif
```diff
@@ -1,52 +1,204 @@
...
-                  "physicalLocation": {
-                    "uriBaseId": "%SRCROOT%",
-                    "uri": "/c:/src/file1.c"
-                  }
+                  "locations": [
+                    {
+                      "location": {
+                        "physicalLocation": {
+                          "fileLocation": {
+                            "uriBaseId": "%SRCROOT%",
```
michaelcfanning: >SRCROOT [](start = 43, length = 7)

UBER NIT: SRCROOT doesn't seem like the right variable name for a relative path that starts with a drive letter. :) No change proposed, just commenting to amuse myself.

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInInvocationAttachment.sarif
lgolding: For each rule I have a separate negative test case for each possible error, and then a single positive case showing all the correct content. So for example `TestData\UriBaseIdRequiresRelativeUri\UrisAreRelative.sarif` is the positive test case. Likewise, `TestData\UrisMustBeValid\ValidUris.sarif` is the positive test case for that rule.

I'd be willing to discuss adding a documenting `message` property somewhere in each test file, but I think if I keep them small enough (as I mentioned in response to another comment) they will be understandable.

---
In reply to: [210452808](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452808) [](ancestors = 210452808)

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidFileLocationUri_Expected.sarif
```diff
@@ -0,0 +1,93 @@
...
+      },
+      "invocations": [
+        {
+          "startTime": "2018-08-15T23:04:47.640Z",
+          "endTime": "2018-08-15T23:04:48.453Z",
+          "toolNotifications": [
+            {
+              "id": "MSG001.AnalyzingTarget",
+              "physicalLocation": {
+                "fileLocation": {
```
michaelcfanning: >fileLocation [](start = 17, length = 12)

is this file manually or auto-generated?

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/UrisAreRelative.sarif
lgolding: Oh, good point. I think I'll fix it; it will just make the files that much shorter.

---
In reply to: [210453175](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210453175) [](ancestors = 210453175)

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRunGraphNode_Expected.sarif
lgolding: Not sure what you're asking. This `_Expected` file shows that the validator detected a problem in the `uri` property of the `fileLocation` property of the `physicalLocation` property of the `location` property of the 0th element of a graph's `nodes` array. There's no logical location involved. `nodes[0].location` could certainly include a logical location, but that's not relevant to the validation rule I'm testing.

---
In reply to: [210452957](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452957) [](ancestors = 210452957)

src/Sarif.Multitool.FunctionalTests/Rules/StepValuesMustFormOneBasedSequenceTests.cs
lgolding: Yes, you've sold me on that. I had no idea it was so slow compared even to file i/o.

---
In reply to: [210452008](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452008) [](ancestors = 210452008)

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidFileLocationUri_Expected.sarif
lgolding: It is generated by the script `Update-Expected.ps1` in the Sarif.Multitool.FunctionalTests directory. I run that script by hand to update the "expected" files for each rule I implement. I check the output to ensure that the "expected" content has exactly the results I expect.

---
In reply to: [210453490](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210453490) [](ancestors = 210453490)

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidRuleHelpUri_Expected.sarif
```diff
@@ -0,0 +1,93 @@
...
+              "text": "UrisMustBeValid"
+            },
+            "shortDescription": {
+              "text": "Specify a valid URI reference for every URI-valued property."
+            },
+            "fullDescription": {
+              "text": "Specify a valid URI reference for every URI-valued property."
+            },
+            "messageStrings": {
+              "Default": "{0}: The string \"{1}\" is not a valid URI reference."
```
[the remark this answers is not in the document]
lgolding: IMO the fact that the strings are persisted in JSON, and so require escaping, is neither here nor there. I'm concerned solely with how it looks to the user. I feel pretty strongly (can you tell?) that in US English top-level quotes are double quotes. I know we differ on this.

---
In reply to: [210453364](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210453364) [](ancestors = 210453364)

lgolding:
I see you approved, thanks. I'll address a few of the things you commented on in the morning, then merge it.

src/Sarif.Multitool/Rules/StepValuesMustFormOneBasedSequence.cs
lgolding: Nice idea. It does assume a "smart" consumer, and not two people poring over a raw log file together.

src/Sarif.Multitool/Rules/UrisMustBeValid.cs
lgolding: It doesn't work. `Uri.TryCreate` happily returns `true` and produces a `Uri` object whose `OriginalString` is `fi%le://whatever`.

---
In reply to: [210452152](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452152) [](ancestors = 210452152,210451903)

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRunGraphNode_Expected.sarif
lgolding: Oh, _now_ I see what you're asking! You're asking why my validator tool doesn't produce a logical location corresponding to the JPointer! That's a really nice idea.

---
In reply to: [210453784](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210453784) [](ancestors = 210453784,210452957)

src/Sarif.Multitool/Rules/UrisMustBeValid.cs
michaelcfanning: Broken API.

Try using Uri.IsWellFormedUriString FTW

---
In reply to: [210650717](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210650717) [](ancestors = 210650717,210452152,210451903)

lgolding: Works.

---
In reply to: [210653881](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210653881) [](ancestors = 210653881,210650717,210452152,210451903)

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/AbsoluteUriInRelatedLocation.sarif
michaelcfanning: Right but this approach leads to file/test proliferation. Alternate philosophy drives maximal efficiency from least content. I prefer this for data-driven approaches. Use the property bag to store important clarifying information for individual results. Just my opinion. If you want small, focused testing, your string literals approach makes more sense. 

obviously your mileage varies philosophically. 

---
In reply to: [210452997](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210452997) [](ancestors = 210452997,210452611)

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidRuleHelpUri_Expected.sarif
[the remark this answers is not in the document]
michaelcfanning: I just like pulling this one out of the stable every now and then for a quick ride.

---
In reply to: [210454237](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210454237) [](ancestors = 210454237,210453364)

src/Sarif.Multitool.FunctionalTests/Rules/StepValuesMustFormOneBasedSequenceTests.cs
michaelcfanning: I wonder whether it was the enumeration over the test case data specifically that caused the problem. might poke around on it at a future time.

---
In reply to: [210453847](https://github.com/Microsoft/sarif-sdk/pull/1018#discussion_r210453847) [](ancestors = 210453847,210452008)

PR #1022, opened 2018-08-16
Rermove errant commas from calls to Get-ProjectBinDirectory

EasyRhinoMSFT:


scripts/BuildAndTest.ps1
```diff
@@ -97,7 +97,7 @@ function New-SigningDirectory {
     }
 
     foreach ($project in $Projects.Products) {
-        $projectBinDirectory = (Get-ProjectBinDirectory $project, $configuration)
+        $projectBinDirectory = (Get-ProjectBinDirectory $project $configuration)
```
lgolding: >$project $configuration [](start = 56, length = 23)

In case you're not familiar with the PowerShell argument-passing syntax: The original call passed a single argument to `Get-ProjectBinDirectory`; that argument was a list containing two items `($project, $configuration)`. The corrected version passes two arguments, `$project` and `$configuration`.

lgolding: Also: You don't need the parentheses enclosing the call to `Get-ProjectBinDirectory`.

---
In reply to: [210758035](https://github.com/Microsoft/sarif-sdk/pull/1022#discussion_r210758035) [](ancestors = 210758035)

PR #1023, opened 2018-08-16
Restore last rule: UseAbsolutePathsForNestedFileUriFragments

lgolding:
As of this change, the v2 Multitool `validate` command is at parity with the old v1 validator tool.

Just as was the case in the previous PR, where we restored the rule `UrisMustBeValid`, we require little test coverage for this rule. We need only verify that it is checked for URIs that appear as the value of the `fileLocation.uri` property, and for URIs that appear as property names in the `run.files` dictionary.

The exhaustive test coverage for the rule `UriBaseIdRequiresRelativeUri` ensures that we actually visit every `fileLocation` object in the schema. So for any other rule that validates any aspect of a `fileLocation` object, we only require one test case -- any SARIF file that contains a `fileLocation` object anywhere will do.


src/Sarif.Multitool.FunctionalTests/Sarif.Multitool.FunctionalTests.csproj
```diff
@@ -40,4 +36,6 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
   </ItemGroup>
+
+  <ProjectExtensions><VisualStudio><UserProperties TestData_4UseAbsolutePathsForNestedFileUriFragments_4NestedFileUriFragmentsAreAbsolute_1sarif__JSONSchema="http://json.schemastore.org/sarif-2.0.0" /></VisualStudio></ProjectExtensions>
 </Project>
```
EasyRhin0: Use indented formatting

PR #1024, opened 2018-08-17
Restore the SDK sample to working order.

lgolding:


lgolding:

lgolding:

lgolding:

lgolding:

src/Samples/Sarif.Sdk.Sample/Program.cs
```diff
@@ -41,6 +41,8 @@ static int LoadSarifLogFile(LoadOptions options)
             string logText = File.ReadAllText(options.InputFilePath);
             SarifLog log = JsonConvert.DeserializeObject<SarifLog>(logText, settings);
 
+            Console.WriteLine($"The log file \"{options.InputFilePath}\" contains {log.Runs[0]?.Results.Count} results.");
```
lgolding: Just so this verb produces some output.

src/Samples/Sarif.Sdk.Sample/Program.cs
```diff
@@ -52,9 +54,9 @@ static int LoadSarifLogFile(LoadOptions options)
         static int CreateSarifLogFile(CreateOptions options)
         {
             // We'll use this source file for several defect results -- the
-			// SampleSourceFiles folder should be at the same level as the project folder
+            // SampleSourceFiles folder should be at the same level as the project folder
```
lgolding: Tabs to spaces.

src/Samples/Sarif.Sdk.Sample/Program.cs
```diff
@@ -226,7 +239,7 @@ static int CreateSarifLogFile(CreateOptions options)
                         var result = new Result()
                         {
                             RuleId = rule.Id,
-                            AnalysisTarget = new FileLocation { Uri = new Uri(@"file://d:/src/module/foo.dll") }, // This is the file that was analyzed
```
lgolding: Not MS compliant

src/Samples/Sarif.Sdk.Sample/Sarif.Sdk.Sample.csproj
```diff
@@ -38,7 +38,13 @@
     <Reference Include="Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
       <HintPath>..\packages\Newtonsoft.Json.11.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
     </Reference>
+    <Reference Include="Sarif, Version=2.0.0.0, Culture=neutral, PublicKeyToken=21a5e83f6f5bb844, processorArchitecture=MSIL">
+      <HintPath>..\packages\Sarif.Sdk.2.0.0-csd.1\lib\net461\Sarif.dll</HintPath>
+    </Reference>
```
lgolding: Consume SDK as NuGet package.

src/Samples/Sarif.Sdk.Sample/packages.config
```diff
@@ -2,7 +2,9 @@
 <packages>
   <package id="CommandLineParser" version="2.2.1" targetFramework="net461" />
   <package id="Newtonsoft.Json" version="11.0.2" targetFramework="net461" />
```
EasyRhin0: >Newtonsoft.Json" version="11.0.2"  [](start = 15, length = 34)

Larry please move this down to 10.0.3 to match the SDK.

src/Samples/Sarif.Sdk.Sample/packages.config
```diff
@@ -2,7 +2,9 @@
...
+  <package id="Sarif.Sdk" version="2.0.0-csd.1" targetFramework="net461" />
   <package id="System.Collections" version="4.0.11" targetFramework="net461" />
+  <package id="System.Collections.Immutable" version="1.5.0" targetFramework="net461" />
```
lgolding: Dependency of Sarif.Sdk.

src/Samples/Sarif.Sdk.Sample/packages.config
lgolding: @EasyRhin0  Thanks, fixed.

src/Samples/Sarif.Sdk.Sample/Program.cs
michaelcfanning: Hooray! My pet peeve addressed.

---
In reply to: [210981653](https://github.com/Microsoft/sarif-sdk/pull/1024#discussion_r210981653) [](ancestors = 210981653)

michaelcfanning:
Please add a build step to our scripts that validates this thing is working. :) Then we won't get into a regressed condition again. After that LGTM

lgolding:
@michaelcfanning Roger wilco on the build step.

lgolding:
@michaelcfanning See changes to BuildAndTest.ps1 to build the sample to avoid regression.

@EasyRhinoMSFT FYI

scripts/BuildAndTest.ps1
```diff
@@ -14,6 +14,8 @@
     Do not rebuild the SARIF object model from the schema.
 .PARAMETER NoBuild
     Do not build.
+.PARAMETER NoBuildSample
```
michaelcfanning: >NoBuildSample [](start = 11, length = 13)

Why do we need this knob? Is there really a situation where disabling the sample build is of material interest? Who would discover this knob in any case? :)

If you want something like this, you should consider a setting for 'Comprehensive' or 'FullGate' and then have things like this disabled by default for happy path developer use. Those settings make sense 'fast and most relevant pre-check-in' and 'slow and comprehensive, verifies that the full VSTS build can do its thing'.

lgolding: I have a switch like this for every step in BuildAndTest. I agree it's not best. I'll do something like you suggest in a separate PR.

Re: discoverability, it's the same as any PowerShell comment that uses the standard doc comment header:
```
Get-Help .\BuildAndTest.ps1 -Detailed
```

michaelcfanning: Roger.

---
In reply to: [211004442](https://github.com/Microsoft/sarif-sdk/pull/1024#discussion_r211004442) [](ancestors = 211004442)

lgolding: @michaelcfanning Come to think of it, I use these knobs all the time when testing changes to the script. But yes, there's probably a better way.

lgolding:
@michaelcfanning Closing the loop: per your comment "Please add a build step to our scripts that validates this thing is working...", we now build the sample in `BuildAndTest.ps1`, both on the desktop and in AppVeyor.

Do you think it's worth adding tests of the sample's functionality? All it does is
1. `load` verb: Read in a SARIF log file and print out how many results it contains.
2. `create` verb" Serialize a SARIF log file from a hard-coded object model.

IMO it's not worth it. If it builds at all, it's hard to see how it can fail to do what it says.

PR #1025, opened 2018-08-17
Make CreateNotifications strongly typed to prevent silent use-facing …

michaelcfanning:
…string construction woes. @Evmaus-MS @lgolding @cfaucon 

Turns out Larry fixed all bad arguments passed to CreateNotification helper, conversion to string[] recompiles cleanly. This change will break an API consumers who depend on the implicit object.ToString() conversion that previously took place in the underlying .NET formatting API.

We need to ship this fix immediately as BinSkim has a reporting bug based on the problem.


src/ReleaseHistory.md
```diff
@@ -185,8 +185,13 @@
 * Multitool improvements--added "rebaseUri" and "absoluteUri" tasks, which will either make the URIs in a SARIF log relative to some base URI, or take base URIs stored in SARIF and make the URIs absolute again.
 * Added a "processing pipeline" model to the SARIF SDK in order to allow easy chaining of operations on SARIF logs (like making all URIs relative/absolute).
 
-## **v2.0.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0))
+## **v2.0.0-csd.1** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1))
```
michaelcfanning: >## **v2.0.0-csd.1 [](start = 0, length = 17)

we shipped, so we should have finalized this information. unfortunately, the readme in the nuget package has broken links

PR #1029, opened 2018-08-17
Fix build breaks related to SDK sample project.

lgolding:
Just comment out the restore and build of the sample solution.


lgolding:

PR #1030, opened 2018-08-18
csd1.0.1 RTM

michaelcfanning:


lgolding:
Without suggesting that I looked at all 379 files, a check of the history difference between `develop` and `master` shows that this is the right set of commits, so... Ship it!

PR #1031, opened 2018-08-19
Restore AppVeyor build of sample app

lgolding:
We expect it to fail on "nuget.exe restore", as it did before, but this is the first step.

**PLEASE DO NOT REVIEW THIS YET**

Also:
- If we're using BuildAndTest.ps1 as our build script, we don't need to explicitly run BeforeBuild.ps1 (BuildAndTest does that).
- Add a bunch of script and build-related files to a solution-level folder in the .sln file, so we can more easily access them in VS.
- Remove unused `native.props`


lgolding:
We'll try again in a fresh PR.

PR #1035, opened 2018-08-20
SDK sample: Add result message strings and correct the regions.

lgolding:
Now that I can build and install the VSIX viewer, I used it to validate the SARIF file produced by the SDK sample app's `create` verb. As a result:

1. I added result message strings, two of which are parameterized with arguments.
2. I corrected the regions for each result.

Also, I noticed and filed two bugs on the viewer's handling of fixes:

Microsoft/sarif-visualstudio-extension#12: Fixes don't recognize text-based regions.
Microsoft/sarif-visualstudio-extension#13: Preview and Apply action links are disabled in Fixes.

### DIFF
--- a/src/Samples/Sarif.Sdk.Sample/Program.cs
+++ b/src/Samples/Sarif.Sdk.Sample/Program.cs
@@ -67,38 +67,42 @@ namespace Sarif.Sdk.Sample
                 {
                     Id ="CA1819",
                     Name = new Message { Text = "Properties should not return arrays" },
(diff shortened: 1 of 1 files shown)


EasyRhinoMSFT:
We should also think about an example of embedded linking. Could be done separately.

lgolding:
Sure thing. The task on my plate was to restore the sample to parity with its previous functionality, and fix obvious bugs (the missing messages and incorrect regions). Now that this is done, we have clear sailing to add anything else we link.

---
In reply to: [414471184](https://github.com/Microsoft/sarif-sdk/pull/1035#issuecomment-414471184) [](ancestors = 414471184)

PR #1036, opened 2018-08-20
Build the SDK sample, but only on the desktop

lgolding:
On AppVeyor, the invocation of `NuGet restore` results in a null reference exception for reasons we don't yet understand. So in AppVeyor, disable the building of the sample.

*grin* to MikeFan, who asked me of what use the "-NoBuildSample" switch might be ;-)

Also:
- Add a bunch of build-related files to the solution so we can edit them more easily in VS.
- Remove now-unused native.props, and merge build.common.props into build.props, because we no longer have any "old-style" projects in our solution (they moved into the sarif-visualstudio-extension repo).


appveyor.yml
```diff
@@ -9,10 +9,6 @@ branches:
     - develop
     - sarif-v1
 
-before_build:
-  - '"%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Community\Common7\Tools\VsDevCmd.bat"'
```
lgolding: >VsDevCmd [](start = 79, length = 8)

No need for this; AppVeyor automatically sets up a VS command prompt window for you (as it turns out).

appveyor.yml
```diff
@@ -9,10 +9,6 @@ branches:
...
-  - ps: .\scripts\BeforeBuild.ps1 -NuGetVerbosity detailed
```
lgolding: >BeforeBuild [](start = 18, length = 11)

No Need for this; `BuildAndTest.ps` runs `BeforeBuild.ps1`.

appveyor.yml
```diff
@@ -22,7 +18,7 @@ configuration:
   - Release
 
 build_script:
-  - ps: .\scripts\BuildAndTest.ps1 -NoTest -NoPackage -NoPublish -NoSigningDirectory -Configuration Release
+  - ps: .\scripts\BuildAndTest.ps1 -NoBuildSample -NoTest -NoPackage -NoPublish -NoSigningDirectory -Configuration Release
```
lgolding: >NoBuildSample [](start = 36, length = 13)

We can't yet build the sample on AppVeyor because the invocation of `NuGet.exe restore` causes a null reference exception that I haven't yet been able to debug. But we _can_ (and do) build it on the desktop.

src/build.props
```diff
@@ -37,13 +28,49 @@
   </PropertyGroup>
 
   <PropertyGroup Label="Build">
+    <OutputSubDir>$(Platform)_$(Configuration)</OutputSubDir>
```
lgolding: >OutputSubDir [](start = 5, length = 12)

All the added lines come from the now-removed `build.common.props`. I interleaved its contents in the appropriate places.

src/build.props
```diff
@@ -1,20 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<!--
-  Projects that use the new VS 2017 project system import this file. Projects that
-  use the old (pre-VS 2017) project system import "build.old-project.props" instead.
-  This file gets the simple name "build.props" rather than being called
-  "build.new-project.props" because almost every project in this repo uses the
-  VS 2017 project system.
```
lgolding: Nope, everyone uses the new project system now.

src/Sarif.Sdk.sln
```diff
@@ -24,6 +24,48 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sarif.FunctionalTests", "Sa
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sarif.Multitool.FunctionalTests", "Sarif.Multitool.FunctionalTests\Sarif.Multitool.FunctionalTests.csproj", "{859920F3-DB09-4C1D-9EAD-DFB1D7A03A46}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "root", "root", "{BBDE0751-B8A3-4EA6-A3DF-3CE7BA41EA0F}"
+	ProjectSection(SolutionItems) = preProject
+		..\.gitattributes = ..\.gitattributes
```
lgolding: >gitattributes [](start = 6, length = 13)

Add solution items for all loose files. All of them are grouped under a single top-level solution folder `root` and its subfolders `src` and `scripts`, so as not to pollute Solution Explorer's view.

lgolding:

lgolding:

lgolding:

lgolding:

PR #1038, opened 2018-08-20
Fix #1037 Sample log creation generates incorrect source file path

### LINKED ISSUE #1037: Sample log creation generates incorrect source file path
Dot-dots one too many times

EasyRhinoMSFT:


src/Samples/Sarif.Sdk.Sample/Program.cs
```diff
@@ -57,7 +57,7 @@ static int CreateSarifLogFile(CreateOptions options)
             // SampleSourceFiles folder should be at the same level as the project folder
             // Because this file can actually be accessed by this app, its
             // content will be embedded in the log file.
-            var fileLocation = new FileLocation { Uri = new Uri($"file://{AppDomain.CurrentDomain.BaseDirectory}/../../../../SampleSourceFiles/AnalysisSample.cs") };
+            var fileLocation = new FileLocation { Uri = new Uri($"file://{AppDomain.CurrentDomain.BaseDirectory}/../../../SampleSourceFiles/AnalysisSample.cs") };
```
lgolding: >SampleSourceFiles [](start = 122, length = 17)

Is the idea to prevent it from prompting if you run the sample directly from the bin directory into which it was built?

EasyRhin0: That seems like the least-confusing assumption to me

---
In reply to: [211441067](https://github.com/Microsoft/sarif-sdk/pull/1038#discussion_r211441067) [](ancestors = 211441067)

lgolding: (For some reason I don't see your reply here, but... ok, that's fine. But add a comment explaining why you chose this path. This is sample code, so people will examine it, and try to understand it, and model their code after it... so we want to be really clear about expedient choices like this.

---
In reply to: [211441067](https://github.com/Microsoft/sarif-sdk/pull/1038#discussion_r211441067) [](ancestors = 211441067)

PR #1040, opened 2018-08-21
Add new validation rule UriMustBeAbsolute

lgolding:
Some URIs in SARIF are required to be absolute. Verify that the ones that must be, are:

- The values in the `run.originalUriBaseIds` dictionary.
- `sarifLog.$schema`
- `tool.downloadUri`
- `invocation.executableLocation.fileLocation.uri` (but see oasis-tcs/sarif-spec#221)
- `versionControlDetails.uri`
- The elements of the `result.workItemUris` array.
- `rule.HelpUri`

@michaelcfanning Note that I've taken your advice on structuring the tests. There is one test file where all the relevant properties are invalid, and one where they are all valid.

Also:
- Shorten some properties to expression bodies.

### DIFF
--- a/src/Sarif.Multitool.FunctionalTests/Rules/UriMustBeAbsoluteTests.cs
+++ b/src/Sarif.Multitool.FunctionalTests/Rules/UriMustBeAbsoluteTests.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class UriMustBeAbsoluteTests : SkimmerTestsBase
+    {
+        [Fact(DisplayName = nameof(UriMustBeAbsolute_ReportsInvalidSarif))]
+        public void UriMustBeAbsolute_ReportsInvalidSarif()
+        {
+            Verify(new UriMustBeAbsolute(), "Invalid.sarif");
+        }
+
+        [Fact(DisplayName = nameof(UriMustBeAbsolute_AcceptsValidSarif))]
+        public void UriMustBeAbsolute_AcceptsValidSarif()
+        {
+            Verify(new UriMustBeAbsolute(), "Valid.sarif");
+        }
+    }
+}
--- a/src/Sarif.Multitool/Rules/DoNotUseFriendlyNameAsRuleId.cs
+++ b/src/Sarif.Multitool/Rules/DoNotUseFriendlyNameAsRuleId.cs
@@ -22,16 +22,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
         /// </summary>
         public override string Id => RuleId.DoNotUseFriendlyNameAsRuleId;
 
-        protected override IEnumerable<string> MessageResourceNames
+        protected override IEnumerable<string> MessageResourceNames => new string[]
         {
-            get
-            {
-                return new string[]
-                {
-                    nameof(RuleResources.SARIF1001_Default)
-                };
-            }
-        }
+            nameof(RuleResources.SARIF1001_Default)
+        };
 
         protected override void Analyze(Rule rule, string rulePointer)
         {
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -181,9 +181,9 @@
     <value>If a fileLocation object contains a "uriBaseId" property, the value of the "uri" property must be a relative URI reference.</value>
   </data>
   <data name="SARIF1015_Default" xml:space="preserve">
-    <value>{0}: The SARIF specification requires the value of this property to be an absolute URI, but "{1}" is a relative URI reference.</value>
+    <value>{0}: The value of this property is required to be an absolute URI, but "{1}" is a relative URI reference.</value>
   </data>
   <data name="SARIF1015_UriMustBeAbsolute" xml:space="preserve">
-    <value>Certain URIs in the SARIF specification are required to be absolute.</value>
+    <value>Certain URIs are required to be absolute.</value>
   </data>
 </root>
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -180,4 +180,10 @@
   <data name="SARIF1014_UriBaseIdRequiresRelativeUri" xml:space="preserve">
     <value>If a fileLocation object contains a "uriBaseId" property, the value of the "uri" property must be a relative URI reference.</value>
   </data>
+  <data name="SARIF1015_Default" xml:space="preserve">
+    <value>{0}: The SARIF specification requires the value of this property to be an absolute URI, but "{1}" is a relative URI reference.</value>
+  </data>
+  <data name="SARIF1015_UriMustBeAbsolute" xml:space="preserve">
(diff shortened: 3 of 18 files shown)


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif.Multitool/Rules/DoNotUseFriendlyNameAsRuleId.cs
```diff
@@ -22,16 +22,10 @@ public class DoNotUseFriendlyNameAsRuleId : SarifValidationSkimmerBase
         /// </summary>
         public override string Id => RuleId.DoNotUseFriendlyNameAsRuleId;
 
-        protected override IEnumerable<string> MessageResourceNames
```
lgolding: >MessageResourceNames [](start = 47, length = 20)

Change all these to expression bodies; saves 6 lines per file.

src/Sarif.Multitool/Rules/UriMustBeAbsolute.cs
```diff
@@ -0,0 +1,110 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Json.Pointer;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class UriMustBeAbsolute : SarifValidationSkimmerBase
```
lgolding: >UriMustBeAbsolute [](start = 17, length = 17)

Note that I named this `UriMustBeAbsolute` in contrast to an existing rule `UrisMustBeValid`. My reasoning is that _all_ URIs must be valid, so we say to the user "(All)UrisMustBeValid". Whereas only certain URIs are required to be absolute, so we say to the user "(This)UriMustBeAbsolute".

Thoughts?

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -180,4 +180,10 @@
   <data name="SARIF1014_UriBaseIdRequiresRelativeUri" xml:space="preserve">
     <value>If a fileLocation object contains a "uriBaseId" property, the value of the "uri" property must be a relative URI reference.</value>
   </data>
+  <data name="SARIF1015_Default" xml:space="preserve">
```
michaelcfanning: >SARIF1015_Default [](start = 14, length = 17)

It would be nice if we had a convention where every message was followed by a reference to the relevant section of the specification. This would be difficult to maintain but would be extremely helpful. We could consider making a pass towards the end when the spec ships. Alternately, we could contribute a bookmarking scheme that allows us to publish a stable identifier to a web-hosted spec.

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -180,4 +180,10 @@
...
+    <value>{0}: The SARIF specification requires the value of this property to be an absolute URI, but "{1}" is a relative URI reference.</value>
```
michaelcfanning: >SARIF specification requires the value [](start = 20, length = 38)

the SARIF spec requires all of this validation, so it is not helpful to note this. instead, prefer the spec language itself, such as The value of this property must be an absolute URI, or should or whatever.

src/Sarif.Multitool/Rules/RuleResources.resx
lgolding: That's a really nice idea. I agree it's too early, because section numbers keep changing, but yes, if we had a stable URI that would help.

---
In reply to: [211694972](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211694972) [](ancestors = 211694972)

src/Sarif.Multitool/Rules/RuleResources.resx
lgolding: How's this: "The value of this property is required to be an absolute URI, but "{1}" is a relative URI reference."

---
In reply to: [211695585](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211695585) [](ancestors = 211695585)

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -180,4 +180,10 @@
...
+  </data>
+  <data name="SARIF1015_UriMustBeAbsolute" xml:space="preserve">
+    <value>Certain URIs in the SARIF specification are required to be absolute.</value>
```
lgolding: >Certain [](start = 11, length = 7)

Likewise here, just remove the words "in the SARIF specification".

src/Sarif.Multitool/Rules/UriMustBeAbsolute.cs
EasyRhin0: Accurately describes what it is, sounds good to me.

---
In reply to: [211458124](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211458124) [](ancestors = 211458124)

src/Sarif.Multitool/Rules/RuleResources.resx
michaelcfanning: yes

---
In reply to: [211696777](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211696777) [](ancestors = 211696777)

src/Sarif.Multitool/Rules/UriMustBeAbsolute.cs
michaelcfanning: This is the rule that enforces uris that must be absolute, so agree with Chris.

---
In reply to: [211698838](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211698838) [](ancestors = 211698838,211458124)

src/Sarif.Multitool.FunctionalTests/Rules/UriMustBeAbsoluteTests.cs
```diff
@@ -0,0 +1,22 @@
...
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class UriMustBeAbsoluteTests : SkimmerTestsBase
+    {
+        [Fact(DisplayName = nameof(UriMustBeAbsolute_ReportsInvalidSarif))]
+        public void UriMustBeAbsolute_ReportsInvalidSarif()
+        {
+            Verify(new UriMustBeAbsolute(), "Invalid.sarif");
```
michaelcfanning: >UriMustBeAbsolute [](start = 23, length = 17)

Files should be named UriMustBeAbsolute_Valid.sarif and UriMustBeAbsolute_Invalid.sarif, as we discussed. Put them all in one directory. Now you can write a test that works just against your rule name and it will locate the file and do the right thing. All TBD of course, don't block.

src/Sarif.Multitool/Rules/UriMustBeAbsolute.cs
lgolding: Not to look a gift horse in the mouth, but I was asking your opinion on the (intentional) use of the singular "Uri" in this rule name, while I used the plural "Uris" in the other rule name.

---
In reply to: [211705887](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211705887) [](ancestors = 211705887,211698838,211458124)

michaelcfanning: I like the distinction. Plural is a rule that applies universally. Singular does not.

---
In reply to: [211707455](https://github.com/Microsoft/sarif-sdk/pull/1040#discussion_r211707455) [](ancestors = 211707455,211705887,211698838,211458124)

PR #1043, opened 2018-08-21
Fix build break in VS due to collision on VersionConstants.cs

lgolding:
In VS, it seems, the two platform versions of the Sarif project (net461 and netstandard20) are built simultaneously. Sometimes that results in a collision in the attempt to create `VersionConstants.cs`.

Instead, create it in `BeforeBuild.ps1`. This makes life a little more complicated, because now we have to read a property value from `build.props` instead of letting MSBuild do it for us. Note that `Get-VersionConstants` was already doing that, so we already had code that demonstrated the pattern.

We perform a little refactoring to DRY out a couple of Powershell variables that are now used in multiple places.

If we only built for one platform -- netstandard2.0, as @Evmaus-MS has suggested -- we wouldn't have hit this problem. Still, it's not a bad idea to move this into `BeforeBuild.ps1` in case we ever do in future build multiple platforms for flavors simultaneously (such as a Debug and a Release build).

### DIFF
--- a/scripts/BeforeBuild.ps1
+++ b/scripts/BeforeBuild.ps1
@@ -55,11 +55,34 @@ function Remove-BuildOutput {
     }
 }
 
+# Install-VersionConstantsFile
+#
+# Create a source file containing a class that specifies the NuGet package
+# version number, and place it in the Sarif project directory.
+#
+function Install-VersionConstantsFile {
+    # The name of the project in which to install VersionConstants.cs.
+    $targetProjectName = "Sarif"
+
+    # The element in build.props from which to extract the root of the namespace
+    # used in VersionConstants.cs.
+    $xPath = "/msbuild:Project/msbuild:PropertyGroup[@Label='Build']/msbuild:RootNamespaceBase"
+    $xml = Select-Xml -Path $BuildPropsPath -Namespace $MSBuildXmlNamespaces -XPath $xPath
+    $rootNamespaceBase = $xml.Node.InnerText
+
+    $targetProjectDirectory = "$SourceRoot\$targetProjectName"
+    $rootNamespace = "$rootNamespaceBase.$targetProjectName"
+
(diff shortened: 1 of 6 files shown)


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

lgolding:
> <?xml version="1.0" encoding="utf-8"?>

Oh wow. This duplicate NuGet file, which I removed in the previous PR, is auto-generated. Maybe by the call to "NuGet.exe restore"? That doesn't make sense. The command line specifies the path to another NuGet.config. Oh well. I'm going to delete it again. Ideas welcome.

---
Refers to: src/NuGet.Config:1 in 62b0766. [](commit_id = 62b0766cebed488ada2ae157132e9feb401042eb, deletion_comment = False)

EasyRhin0:
> <?xml version="1.0" encoding="utf-8"?>

gitignore it for now

---
In reply to: [414848302](https://github.com/Microsoft/sarif-sdk/pull/1043#issuecomment-414848302) [](ancestors = 414848302)

---
Refers to: src/NuGet.Config:1 in 62b0766. [](commit_id = 62b0766cebed488ada2ae157132e9feb401042eb, deletion_comment = False)

lgolding:
> <?xml version="1.0" encoding="utf-8"?>

That's a good idea and I will do that. Unfortunately, it doesn't solve the whole problem, which is that I _saw_ two different NuGet.configs and wondered which was the real one... Oh! The problem is that I chose to use the one in RepoRoot rather than the one in src! I'll fix that. Stand by.

---
In reply to: [414849458](https://github.com/Microsoft/sarif-sdk/pull/1043#issuecomment-414849458) [](ancestors = 414849458,414848302)

---
Refers to: src/NuGet.Config:1 in 62b0766. [](commit_id = 62b0766cebed488ada2ae157132e9feb401042eb, deletion_comment = False)

PR #1044, opened 2018-08-22
FxCop converter improvements

michaelcfanning:


src/Sarif/FileRegionsCache.cs
```diff
@@ -234,7 +234,7 @@ private void PopulateEndColumn(NewLineIndex lineIndex, Region region, string fil
             }
         }
 
-        private static void PopulateCharOffset(NewLineIndex lineIndex, Region region, string fileText)
```
michaelcfanning: >string fileText [](start = 86, length = 15)

FxCop notices unused parameters. I scrubbed some of these issues in order to generate 'absent' messages in the project file.

michaelcfanning:
This change significantly improves the FxCop converter, including adding the ability to convert an FxCop project file that contains a current analysis report. As part of this, FxCop itself discovered various correctness bugs, some of which I fixed in order to make FxCop messages disappear (invoking the 'absent' state). Note that the updated FxCop content ('enriched') is parked in its own directory. I haven't yet added the capability to regenerate this content alongside the SDK directory, this will happen in the next check-in.

src/Sarif/FileRegionsCache.cs
michaelcfanning: @lgolding @Evmaus-MS @EasyRhin0

scripts/BuildAndTest.ps1
```diff
@@ -200,7 +200,7 @@ if (-not $NoBuildSample) {
 }
 
 if (-not $NoTest) {
-    & $PSScriptRoot\Run-Tests.ps1
+    & $PSScriptRoot\Run-Tests.ps1 $Configuration
```
rtaket: I suspect you either need to specify the -Configuration switch or add the Position attribute to the param in Run-Tests.ps1.

michaelcfanning: This definitely invokes the debug tests. It's all academic, however, as the debug builds don't pass tests. Likely need to scrub for asserts that are firing.

michaelcfanning:
@lgolding @Evmaus-MS @EasyRhin0 glad to take your retroactive comments on this PR.

lgolding:
@mikefan@microsoft.com looking now.

---
In reply to: [415455962](https://github.com/Microsoft/sarif-sdk/pull/1044#issuecomment-415455962) [](ancestors = 415455962)

scripts/BuildAndTest.ps1
lgolding: Yes and no.

1. By default `Run-Tests.ps1` runs the `Release` tests, not the `Debug` tests. See the declaration of the `$Configuration` parameter in `Run-Tests.ps1` on line 21.
2. It is definitely a bug that this invocation of `Run-Tests.ps1` did not pass the `$Configuration` value from the calling script. Thanks for fixing.
3. Even though `$Configuration` is a positional parameter, and the first one (so that the change you made works properly), I always prefer to spell out the parameter name, thus:
```
& $PSScriptRoot\Run-Tests.ps1-Configuration $Configuration
```
This is good documentation (although admittedly of limited value in this particular case), and it makes the caller resilient against changes to the parameter declaration order in the callee.

---
In reply to: [212344747](https://github.com/Microsoft/sarif-sdk/pull/1044#discussion_r212344747) [](ancestors = 212344747)

src/Sarif/PropertiesDictionary.cs
```diff
@@ -158,9 +150,14 @@ private static bool TryConvertFromString<T>(string source, out T destination)
 
         public void SaveToJson(string filePath, bool prettyPrint = true)
         {
+
+            Newtonsoft.Json.Formatting formatting = prettyPrint
```
lgolding: >Formatting [](start = 28, length = 10)

Consider:
```
using Formatting = Newtonsoft.Json.Formatting;
```

lgolding:
>         private void AddFile(PhysicalLocation physicalLocation)

Consider an opportunistic refactor. This method really does two things:
1. Create a `FileData` object from a `PhysicalLocation`.
2. Add the resulting `FileData` object to the dictionary.

Consider extracting `private void CreateFileDataFrom(PhysicalLocation physicalLocation)`. That method is more easily unit testable.

In general, consider separating creation responsibility from usage responsibility (an example of Single Responsibility Principle).

---
Refers to: src/Sarif.Converters/FileInfoFactory.cs:87 in 2608e16. [](commit_id = 2608e16326548fffea760b4b37cfc0809147b0f1, deletion_comment = False)

src/Sarif.Converters/FileInfoFactory.cs
```diff
@@ -96,6 +93,13 @@ private void AddFile(PhysicalLocation physicalLocation)
 
             Uri uri = physicalLocation.FileLocation.Uri;
             string key = UriHelper.MakeValidUri(uri.OriginalString);
+
+            string uriBaseId = physicalLocation.FileLocation.UriBaseId;
+            if (!string.IsNullOrEmpty(uriBaseId))
+            {
+                key = "#" + uriBaseId + "#" + key;
```
lgolding: >"#" + uriBaseId + "#" + key; [](start = 22, length = 28)

This is ok if the SDK is consistent about it, and I suppose it's the safest course.

The spec doesn't require the `#<uriBaseId>` prefix _unless_ the dictionary needs to contain entries for two files with the same `uri` but different `uriBaseIds`. You can't actually determine that without looking at every `fileLocation` object in the log file.

So yes, I think this is the right thing to do. Just thinking out loud.

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
+Notes on producing test examples.
+
+This output is an FxCop run against a debug build of SARIF-SDK
```
lgolding: >SARIF-SDK [](start = 53, length = 9)

The actual repo name is `sarif-sdk`.

Consider "... a debug build of the sarif-sdk repo."

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
+Notes on producing test examples.
+
+This output is an FxCop run against a debug build of SARIF-SDK 
+
+0. Enlist in the SARIF-SDK. You will update the test files in this version.
```
lgolding: >version [](start = 67, length = 7)

Consider: "version" => "enlistment".

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+1. Git clone into a second version of the tree, e.g.:
+   git clone https://github.com/microsoft/sarif-sdk sarif-sdk-fxcop-test
+2. cd sarif-sdk-fxcop-test. git checkout git checkout fc172277895ed80cb3c03d0ce66738cff89aad75
```
lgolding: >git checkout [](start = 28, length = 12)

These words are duplicated, and I suspect you wanted a line break and a new numbered list item for this command.

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
+Notes on producing test examples.
+
+This output is an FxCop run against a debug build of SARIF-SDK 
+
+0. Enlist in the SARIF-SDK. You will update the test files in this version.
+1. Git clone into a second version of the tree, e.g.:
```
lgolding: >version of the t [](start = 27, length = 16)

Again, consider: "version of the tree" => "copy of the repo".

Nit, but "version" means something specific in a VCS.

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+   git clone https://github.com/microsoft/sarif-sdk sarif-sdk-fxcop-test
+2. cd sarif-sdk-fxcop-test. git checkout git checkout fc172277895ed80cb3c03d0ce66738cff89aad75
+3. BuildAndTest.cmd --Configuration Debug -NoTest (debug to emit in-source suppressions)
```
lgolding: >  [](start = 49, length = 1)

You can save them time with `-NoBuildSample -NoPackage`.

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
+Notes on producing test examples.
+
+This output is an FxCop run against a debug build of SARIF-SDK 
+
+0. Enlist in the SARIF-SDK. You will update the test files in this version.
+1. Git clone into a second version of the tree, e.g.:
+   git clone https://github.com/microsoft/sarif-sdk sarif-sdk-fxcop-test
```
lgolding: >git [](start = 3, length = 3)

Nit: Since this is a Markdown file, consider enclosing command lines in backticks.

src/Sarif.FunctionalTests/v2/ConverterTestData/FxCop/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+2. cd sarif-sdk-fxcop-test. git checkout git checkout fc172277895ed80cb3c03d0ce66738cff89aad75
+3. BuildAndTest.cmd --Configuration Debug -NoTest (debug to emit in-source suppressions)
+4. Open SarifSdk.FxCop in the non-test enlistment.
```
lgolding: >FxCop [](start = 17, length = 5)

Does this assume they have a command-line version of FxCop installed? If so, mention it at the top. (Or does VS install the necessary shell association for files of type .FxCop? Is there a dependency on a particular version of VS here?)

lgolding:
> function Build-Baselines($toolName)

Existing nit, but IMO worth fixing: Because PowerShell variable names are case-insensitive, this parameter name hides the script parameter name `$ToolName` declared on line 2. This is confusing enough, so to avoid additional confusion, please consistently use lower-case `$toolName` throughout this function. I noticed this because of the inconsistency between lines 45 and 48.

---
Refers to: src/Sarif.FunctionalTests/UpdateBaselines.ps1:16 in 2608e16. [](commit_id = 2608e16326548fffea760b4b37cfc0809147b0f1, deletion_comment = False)

src/Sarif.FunctionalTests/UpdateBaselines.ps1
```diff
@@ -54,17 +61,16 @@ function Build-Baselines($toolName)
         # Actually run the converter
         Remove-Item $outputTemp -ErrorAction SilentlyContinue
         &$utility convert "$input" --tool $toolName --output "$outputTemp" --pretty-print
-
         # Next, perform some rewriting. The PREfast converter in particular cannot embed file contents as the source      
         # SARIF emitted by the compiler does not contain the optional 'files' member of the 'run' object.
-        &$utility rewrite ""$outputTemp"" --output ""$outputTemp"" --pretty-print --insert "TextFiles;Hashes;RegionSnippets;ContextCodeSnippets" --force
+        &$utility rewrite ""$outputTemp"" --output ""$outputTemp"" --pretty-print --insert """TextFiles;Hashes;RegionSnippets;ContextRegionSnippets""" --force
```
lgolding: >""" [](start = 91, length = 3)

If the reason for this change is that Powershell interpreted the `;` as a statement terminator, you could instead escape the `;` with a backtick: ``--insert TextFiles`;Hashes`;RegionSnippets`;ContextRegionSnippets``

src/Sarif.FunctionalTests/UpdateBaselines.ps1
```diff
@@ -42,6 +42,13 @@ function Build-Baselines($toolName)
       $sourceExtension = "fpr"
     }
 
+    if ($ToolName -eq "FxCopProject")
+    {
+      $sourceExtension = "fxcop"
+      $toolName = "FxCop"
```
lgolding: >FxCop [](start = 19, length = 5)

Rather than special-casing `FxCopyProject` as you do here and on line 73, would you consider placing the `FxCopProject` files in a separate same-named subdirectory of `ConverterTestData`?

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -88,12 +91,40 @@ private void BatchRunConverter(string tool, string inputFilter = "*.xml")
...
+            {
+                Formatting = Formatting.Indented,
+                ContractResolver = SarifContractResolver.Instance
+            };
+
+            string logText = File.ReadAllText(actualFilePath);
+
+            SarifLog actualLog = JsonConvert.DeserializeObject<SarifLog>(logText, settings);
+
+            OptionallyEmittedData dataToInsert =
```
lgolding: >OptionallyEmittedData [](start = 12, length = 21)

Consider a comment: "If you change the set of data to insert, make the corresponding change to the invocation of Sarif.Multitool in UpdateBaselines.ps1"

lgolding: Note to self: _are_ those two things related?

---
In reply to: [212415835](https://github.com/Microsoft/sarif-sdk/pull/1044#discussion_r212415835) [](ancestors = 212415835)

src/Sarif.FunctionalTests/SarifConverterTests.cs
```diff
@@ -116,7 +147,7 @@ private void RunConverter(StringBuilder sb, string toolFormat, string inputFileN
             {
                 sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "The converter {0} threw an exception for input \"{1}\".", toolFormat, inputFileName));
                 sb.AppendLine(ex.ToString());
-                return;
+                return generatedFileName;
```
lgolding: >generatedFileName [](start = 23, length = 17)

How about returning `null` on error, and then caller can test for `null` instead of having to test `File.Exists`? I thought that was odd when I saw the test on line 96; I wondered to myself "Why would it not exist?".

src/Sarif.Converters.UnitTests/FxCopConverterTests.cs
```diff
@@ -360,14 +360,15 @@ public void FxCopConverter_CreateResult_FakeContext_Member()
...
+            context.RefineIssue(null, "test", "uncertain", "error", @"source", "myfile.cs", 13);
+            context.RefineItem("hello!");
 
             string expectedLogicalLocation = "mynamespace.mytype.mymember(string)";
 
             var expectedResult = new Result
             {
                 RuleId = "CA0000",
-                Message = new Message { Text = "hello!" },
+                Message = new Message { Arguments = new List<string>(new string[] { "hello!" }) },
```
lgolding: >{ [](start = 38, length = 1)

If we're formatting the arguments into a string, don't we need `MessageId`?

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     ///</remarks>
     internal sealed class FxCopConverter : ToolFileConverterBase
     {
+        private const string ProjectDirectoryVariable = "$(ProjectDir)";
```
lgolding: >private [](start = 8, length = 7)

Ultra-nit: blank line below.

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -108,7 +137,7 @@ internal Result CreateResult(FxCopLogReader.Context context)
             }
 
             result.RuleId = context.CheckId;
-            result.Message = !string.IsNullOrWhiteSpace(context.Message) ? new Message { Text = context.Message } : null;
+            result.Message = new Message { Arguments = context.Items, MessageId = context.ResolutionName, Text = context.Message };
```
lgolding: >MessageId = context.ResolutionName, Text = context.Message [](start = 70, length = 58)

The message text can be derived from the message format string + arguments. Do you want to include both? This was the same question I asked when I reviewed the SARIF you gen'd up for <redacted>.

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -119,10 +148,7 @@ internal Result CreateResult(FxCopLogReader.Context context)
                 !string.IsNullOrWhiteSpace(targetFile) &&
                 !sourceFile.Equals(targetFile))
             {
-                result.AnalysisTarget = new FileLocation()
-                {
-                    Uri = new Uri(targetFile, UriKind.RelativeOrAbsolute)
-                };
+                result.AnalysisTarget = BuildFileLocationFromFxCopReference(targetFile);
```
lgolding: >BuildFileLocationFromFxCopReference [](start = 40, length = 35)

We could use a naming convention for methods like this. In FileInfoFactory.cs, where you had a method that both created a `FileInfo` object and added it to the `Files` dictionary, I suggested you factor out a creation method `private FileData CreateFrom(PhysicalLocation)`. I'm not wedded to that convention at all, it's just that this is the first time I'm seen a creation method in this code base named `BuildXxx`.

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -422,14 +477,26 @@ public void RefineMessage(string checkId, string typename, string messageId, str
             public void RefineIssue(string message, string result, string certainty, string level, string path, string file, int? line)
```
lgolding: >result [](start = 59, length = 6)

Can we rename local variable and parameter names `result` to `resolutionName` in this class, wherever it occurs? It took me reading down to here to understand that `result` wasn't a SARIF `Result` object.

lgolding:
>             public void RefineReport(string report)

What does "Refine" mean in these method names?

---
Refers to: src/Sarif.Converters/FxCopConverter.cs:414 in 2608e16. [](commit_id = 2608e16326548fffea760b4b37cfc0809147b0f1, deletion_comment = False)

src/Sarif.Converters/FxCopConverter.cs
lgolding: >RefineIssue [](start = 24, length = 11)

Also, does "Issue" here refer to a SARIF `Result` object? If you renamed the parameter `result` to `resolutionName`, would it be ok to rename this method to `RefineResult`?

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -647,7 +735,25 @@ public void Read(Context context, Stream input)
 
             using (var sparseReader = SparseReader.CreateFromStream(_dispatchTable, input, schemaSet))
             {
-                if (sparseReader.LocalName.Equals(SchemaStrings.ElementFxCopReport))
+                // FxCop distinctions between project and report files.
```
lgolding: >distinctions [](start = 25, length = 12)

Did you mean to say "distinguishes"? Or else to end this sentence with a colon?

src/Sarif.Converters/FxCopConverter.cs
lgolding: I see. The code below says that a project file produces a fully formed message, whereas a report file produces a message format string + arguments. So you'll never have both. That is totally worth a comment here.

---
In reply to: [212451123](https://github.com/Microsoft/sarif-sdk/pull/1044#discussion_r212451123) [](ancestors = 212451123)

lgolding:
@michaelcfanning Done! Many comments but mostly trivia, a couple of suggestions for design refinements, and one possible bug in FxCopConverterTests.cs.

---
In reply to: [415504250](https://github.com/Microsoft/sarif-sdk/pull/1044#issuecomment-415504250) [](ancestors = 415504250,415455962)

PR #1048, opened 2018-08-24
Fix #1047: correct message string creation for rule & resource strings

### LINKED ISSUE #1047: Result.GetMessageText doesn't correctly construct strings from result.arguments + rule.messageStrings

EasyRhinoMSFT:


src/Sarif/ExtensionMethods.cs
```diff
@@ -182,56 +182,60 @@ public static string GetMessageText(this Result result, IRule rule, bool concise
...
+                if (result.Message?.Arguments != null)
+                {
+                    arguments = new string[result.Message.Arguments.Count];
+                    result.Message.Arguments.CopyTo(arguments, 0);
+                }
+                else
                 {
-                    string messageString;
+                    arguments = new string[0];
+                }
```
lgolding: You may as well defer constructing the `arguments` array until you know whether you have a message format string. I'd move this into the `if` block starting on Line 199.

EasyRhin0: Left over from the refactor...

---
In reply to: [212701501](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212701501) [](ancestors = 212701501)

lgolding: Actually you also don't need to fetch the messageId if you don't even have a rule. Most efficient (and, IMO, clearest) is:
```
Fetch message text
IF message text is null or empty
    IF there is a rule
        Fetch message id
        IF there is a message id AND there is a rule AND the rule has a message with that id
      
    ELSE
        Set text to empty string.
```

---
In reply to: [212702127](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212702127) [](ancestors = 212702127,212701501)

src/Sarif/ExtensionMethods.cs
```diff
@@ -182,56 +182,60 @@ public static string GetMessageText(this Result result, IRule rule, bool concise
...
 
-                    if (rule.MessageStrings?.ContainsKey(messageId) == true)
-                    {
-                        messageString = rule.MessageStrings[messageId];
+            return text;
+        }
+
+        internal static string GetFormattedMessage(string formatString, string[] arguments)
+        {
+            string result = null;
```
lgolding: >result [](start = 19, length = 6)

Consider `result` => `formattedMessage`

src/Sarif/ExtensionMethods.cs
EasyRhin0: Looks good

---
In reply to: [212703463](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212703463) [](ancestors = 212703463,212702127,212701501)

src/Sarif.Driver.UnitTests/Sdk/FormatForVisualStudioTests.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     public class FormatForVisualStudioTests
     {
         private const string TestRuleId = "TST0001";
-        private const string TestMessageStringId = "testMessageStringId";
```
lgolding: >TestMessageStringId [](start = 29, length = 19)

No reason for this rename now.

src/Sarif.UnitTests/SarifExtensionsTests.cs
```diff
@@ -47,5 +47,71 @@ public void SarifExtensions_IsSemanticVersioningCompatible()
...
+                }
+            };
+
+            string expected = "The quick brown fox jumps over the lazy dog.";
+            string actual = result.GetMessageText(null);
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public void SarifExtensions_Result_GetMessageText()
```
lgolding: >SarifExtensions_Result_GetMessageText [](start = 20, length = 37)

`_WithArguments`

src/Sarif/ExtensionMethods.cs
EasyRhin0: Bad habit

---
In reply to: [212704164](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212704164) [](ancestors = 212704164)

lgolding:
Basically looks good. I think the code reordering I proposed in pseudocode is an important change. (I know this was existing code.)

Rationale: I really don't like it when I spend brain cycles to understand a piece of code, then dive into a conditional and find out that I didn't need to have read that code at all.

src/Sarif/ExtensionMethods.cs
EasyRhin0: I left the text = "" line where it is so we never return null

---
In reply to: [212704250](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212704250) [](ancestors = 212704250,212703463,212702127,212701501)

lgolding: Sorry, what is the "this" that returns null?

---
In reply to: [212704250](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212704250) [](ancestors = 212704250,212703463,212702127,212701501)

lgolding: Oh, did I miss a case. Ok.

---
In reply to: [212705797](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212705797) [](ancestors = 212705797,212704250,212703463,212702127,212701501)

EasyRhin0: Sorry, if this fails: IF there is a message id AND the rule has a message with that id
text will be null

---
In reply to: [212705834](https://github.com/Microsoft/sarif-sdk/pull/1048#discussion_r212705834) [](ancestors = 212705834,212704250,212703463,212702127,212701501)

src/Sarif/ExtensionMethods.cs
```diff
@@ -183,45 +183,27 @@ public static string GetMessageText(this Result result, IRule rule, bool concise
...
-                    }
-                    else
-                    {
-                        arguments = new string[0];
-                    }
+                    string messageId = result.Message?.MessageId;
 
-                    if (rule.MessageStrings?.ContainsKey(messageId) == true)
+                    if (!string.IsNullOrWhiteSpace(messageId)
+                        && rule.MessageStrings?.ContainsKey(messageId) == true)
```
lgolding: >ContainsKey [](start = 48, length = 11)

One last tweak. You can save the double dictionary lookup by using `rule.MessageStrings?.TryGetValue(messageId,out formatString) == true`

lgolding:
Approved with one last suggestion.

lgolding:
Ship it!

Issue #1049, opened 2018-08-26
GetFirstSentence can't handle unmatched apostrophes


EasyRhinoMSFT:
Such as contractions e.g. can't don't won't
This is pretty common and can lead to large amounts of text being displayed in the Viewer's error list items.

PR #1050, opened 2018-08-27
Fix issue #1049 GetFirstSentence

EasyRhinoMSFT:
Using a regex improves our accuracy quite a bit, and we can make adjustments much more easily.


src/Sarif.Driver.UnitTests/Sdk/ExtensionMethodsTests.cs
```diff
@@ -10,22 +10,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     public class ExtensionMethodsTests
     {
         [Theory]
-        [InlineData("first (foo.dll) sentence. more text", "first (foo.dll) sentence.")]
-        [InlineData("first 'foo.dll' sentence. more text", "first 'foo.dll' sentence.")]
-        [InlineData("first (') sentence. more text", "first (') sentence.")]
-        [InlineData("first '(' sentence. more text", "first '(' sentence.")]
+        [InlineData("first (foo.dll) sentence. uncapitalized text", "first (foo.dll) sentence. uncapitalized text.")]
```
lgolding: >foo [](start = 28, length = 3)

Change `foo` to `example` throughout.

src/Sarif/ExtensionMethods.cs
```diff
@@ -253,86 +254,26 @@ public static string GetFirstSentence(string text)
...
-                    // If we encounter a line-break, we return all leading text.
-                    case '\n':
-                    case '\r':
-                    {
-                        if (withinQuotes || withinParentheses) { continue; }
-                        return text.Substring(0, length).TrimEnd('\r', '\n', ' ', '.') + ".";
-                    }
+                if (index > -1)
+                {
+                    text = text.Substring(0, index);
```
lgolding: >text [](start = 20, length = 4)

Once you've found the first line break character, you won't find any more, so `break` here.

EasyRhin0: Good eye

---
In reply to: [213073698](https://github.com/Microsoft/sarif-sdk/pull/1050#discussion_r213073698) [](ancestors = 213073698)

src/Sarif/ExtensionMethods.cs
```diff
@@ -253,86 +254,26 @@ public static string GetFirstSentence(string text)
...
 
-                    // If we encounter a space following a period, return 
-                    // all text terminating in the period (inclusive).
-                    case ' ':
-                    {
-                        if (!lastEncounteredWasDot) continue;
-                        if (withinQuotes || withinParentheses) { continue; }
-                        return text.Substring(0, length).TrimEnd('\r', '\n', ' ', '.') + ".";
-                    }
+            Match match = Regex.Match(text, @"^.*?[.?!][\)""']*(?=\s+\p{P}*[\p{Lu}\p{N}]|\s*$)", RegexOptions.Compiled);
```
lgolding: >@"^.*?[.?!][\)""']*(?=\s+\p{P}*[\p{Lu}\p{N}]|\s*$ [](start = 44, length = 49)

Oh good heavens :-)

I suggest you also use [`RegexOptions.IgnorePatternWhitespace`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netframework-4.7.2#System_Text_RegularExpressions_RegexOptions_IgnorePatternWhitespace), which allows you to document your pattern inline, for example:
```
@"^    # Beginning of string
.*?      # Any characters <-- and by the way, you don't need the ? because .* already means "0 or more"
[.?!]   # Sentence-ending punctuation
[\)""']*   # Zero or more quotation marks <-- and by the way (1) Why is the paren character in this character class, and (2) why is it escaped? You don't have to escape anything in a character class except you have to be careful with ^ and ]
(?=...    # Good heavens, I don't know what happens from here on out :-) :)
```

lgolding:
Code looks much better, but that regex really needs a comment. Also, I am concerned that the part of your regex that means "sentence-ending punctuation" (`[.!?]`) is limited to English. A Spanish upside-down exclam or question mark won't match. Is there a Unicode character class for this?

src/Sarif/ExtensionMethods.cs
lgolding: Ok, I understand why your character class `[\)""']` includes the closing paren: because you want to match `(This was a test.)` as a first sentence.

But no, it [doesn't need to be escaped](https://www.regular-expressions.info/charclass.html):

> The usual metacharacters are normal characters inside a character class, and do not need to be escaped by a backslash. To search for a star or plus, use [+*]. Your regex will work fine if you escape the regular metacharacters inside a character class, but doing so significantly reduces readability.

---
In reply to: [213076317](https://github.com/Microsoft/sarif-sdk/pull/1050#discussion_r213076317) [](ancestors = 213076317)

EasyRhin0: Correct re: paren et. al.
Removed the escape, remnant from earlier generation of the pattern.

---
In reply to: [213100457](https://github.com/Microsoft/sarif-sdk/pull/1050#discussion_r213100457) [](ancestors = 213100457,213076317)

EasyRhin0:
Sadly, no. The inverted symbols in Spanish are at the beginning of the sentence though.

---
In reply to: [416330599](https://github.com/Microsoft/sarif-sdk/pull/1050#issuecomment-416330599) [](ancestors = 416330599)

lgolding:
Oh right :) Ok, bad example. Please file an issue so we can decide whether to address is.

---
In reply to: [416392588](https://github.com/Microsoft/sarif-sdk/pull/1050#issuecomment-416392588) [](ancestors = 416392588,416330599)

src/Sarif/ExtensionMethods.cs
lgolding: Nice documentation.

---
In reply to: [213135448](https://github.com/Microsoft/sarif-sdk/pull/1050#discussion_r213135448) [](ancestors = 213135448,213100457,213076317)

PR #1052, opened 2018-08-28
Pick up JSchema references from NuGet. Eliminate result.ruleMessageId

michaelcfanning:
Eliminate result.ruleMessageId. Pick up some Microsoft.json references from nuget rather than foreign checked in binaries.

@lgolding @rtaket @Evmaus-MS @EasyRhin0


michaelcfanning:
We have a string of things that need to be sequenced. I can't update the todotnet.exe code, for example, as it isn't getting packaged properly. We need to eliminate ruleMessageId asap as an important breaking change. BinSkim requires an SDK update with this change (and another which breaks some standard driver output). So, this change removes ruleMessageId but leaves the checked in utility code (which will be removed after the packaging issue is resolved).

src/Sarif.FunctionalTests/v2/SpecExamples/Comprehensive.sarif
```diff
@@ -170,8 +170,8 @@
             "uri": "file:///home/buildAgent/src/collections/list.cpp"
           },
           "ruleId": "C2001",
-          "ruleMessageId": "default",
           "message": {
+            "messageId": "default",
```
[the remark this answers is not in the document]
lgolding: Yes. Having updated the input files, there is then a script Update-Expected.ps1 to update the "expected" files. We've discussed going to a one-file-for-positives, one-file-for-negatives scheme -- and I did that in the rule I added most recently (`UriMustBeAbsolute)`, but I haven't yet updated the other tests for the other rules.

---
In reply to: [213381910](https://github.com/Microsoft/sarif-sdk/pull/1052#discussion_r213381910) [](ancestors = 213381910)

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -14,6 +14,8 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.2.1" />
+    <PackageReference Include="Microsoft.Json.Pointer" Version="0.57.0" />
```
Evmaus-MS: Can we get rid of the "copy all refs" below if this is here?

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -14,6 +14,8 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.4.0" />
+    <PackageReference Include="Microsoft.Json.Schema" Version="0.57.0" />
```
Evmaus-MS: Same comment about mass dll copy below.

PR #1055, opened 2018-08-28
CSD 1.0.1

michaelcfanning:
@lgolding @cfaucon @Evmaus-MS @rtaket


lgolding:
Rubber stamp. I only spot-checked a few files.

PR #1057, opened 2018-08-29
Remove checked in .NET object model generation utility in preference …

michaelcfanning:
…of NuGet-delivered content. Update Jschema generally to 0.58.0.

@lgolding @rtaket @EasyRhinoMSFT


src/Sarif/ToDotNet/packages.config
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.Json.Schema.ToDotNet" version="0.49.0" targetFramework="net451" />
+  <package id="Microsoft.Json.Schema.ToDotNet" version="0.58.0" targetFramework="net461" />
```
Evmaus-MS: Can we just add this as a private package dependency to the Sarif.csproj?

Evmaus-MS:
LGTM

src/Sarif/ToDotNet/packages.config
lgolding: I don't remember why I did it this way. It might have been because I didn't know what a "private package dependency" was :-). If you can make this work without a separate packages.config that's fine.

Evmaus-MS: https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

It should be possible to just do something like this:
<PackageReference Include="Contoso.Utility.UsefulStuff" Version="3.6.0">
        <PrivateAssets>all</PrivateAssets>
    </PackageReference>

That will mark everything in the package as "private" and not flow them to build output/etc.

michaelcfanning:
I will make the change you suggest next, @Evmaus-MS

PR #1058, opened 2018-08-29
Typescript Schema Regen & File Move

Evmaus-MS:
Updates the typescript schema and expands slightly on the script for building the package.

Does not add automatic package generation or build; that should be done later.


src/TypeScript/BuildTypeScript.ps1
```diff
@@ -0,0 +1,7 @@
+
+# TODO:  Expand on this.
+cd Sarif.Sdk
```
michaelcfanning: you will find pushd/popd is preferable for this sort of thing, to avoid need to update relative paths in a complex script when directory nesting changes. just an historical maintainability suggestion for batch files. :)

src/TypeScript/BuildTypeScript.ps1
```diff
@@ -0,0 +1,7 @@
...
+npm install
```
michaelcfanning: obviously, no error checking here. we have lots of examples that provide basic structure for erroring out. and so, i'd go grab that pattern rather than checking in nothing at all (since the incidental costs of stealing existing code is low)

src/TypeScript/Sarif.SDK/package-lock.json
```diff
@@ -440,6 +470,7 @@
       "version": "0.4.12",
       "resolved": "https://registry.npmjs.org/memoizee/-/memoizee-0.4.12.tgz",
       "integrity": "sha512-sprBu6nwxBWBvBOh5v2jcsGqiGLlL2xr2dLub3vR8dnE8YB17omwtm/0NSHl8jjNbcsJd5GMWJAnTSVe/O0Wfg==",
+      "dev": true,
```
michaelcfanning: did npm inject all these>?

src/TypeScript/Sarif.SDK/lib/AutoGenerated/sarif-types.ts
```diff
@@ -309,6 +312,10 @@ export interface Run {
   logicalLocations?: {
     [k: string]: LogicalLocation;
   };
+  /**
+   * An array of one or more unique 'graph' objects.
+   */
+  graphs?: Graph[];
```
michaelcfanning: you'll want to regenerate after someone reviews my open change to update SARIF schema to absolute current.

src/TypeScript/Sarif.SDK/package-lock.json
Evmaus-MS: Yep, dependencies of various packages like TypeScript and the ts2json.

---
In reply to: [214737556](https://github.com/Microsoft/sarif-sdk/pull/1058#discussion_r214737556) [](ancestors = 214737556)

PR #1060, opened 2018-08-30
Result Matching Bugfix

Evmaus-MS:
Small bugfix to the result matching algorithm around handling of null/empty inputs after converting everything to IEnumerables.

Adds unit tests to prevent regression.


michaelcfanning:
>     <PreviousVersionPrefix>2.0.0-csd.1</PreviousVersionPrefix>

need to update this as well

---
Refers to: src/build.props:25 in ec07d40. [](commit_id = ec07d40e49bd258aea28051b56bb083bc8fba912, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -196,3 +196,6 @@
 * Fix weakly typed CreateNotification calls and make API more strongly typed (BREAKING CHANGE)
 * Rename OptionallyEmittedData.ContextCodeSnippets to ContextRegionSnippets (BREAKING CHANGE)
 * Eliminate result.ruleMessageId (in favor of result.message.messageId) (BREAKING CHANGE)
+
+## **v2.0.0-csd.1.0.2** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1.0.1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1.0.1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1.0.1)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1.0.1))
+* Bugfix to result matching algorithm.
```
michaelcfanning: >Bugfix [](start = 2, length = 6)

What's the bug symptom? Add something like this:

Resolves unhandled NullReferenceException raised in MyClass:MyMethod() on matching an empty result set.

or whatever. The idea is that if someone sees a funny break, they can look in the readme and see terms/etc. that identify the fix as the one they need.

PR #1062, opened 2018-08-31
Semmle test

michaelcfanning:
@lgolding @jeffersonking @rtaket @Evmaus-MS


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -122,21 +135,22 @@ internal FileData VisitDictionaryValueNullChecked(string key, FileData node)
...
 
             if (workToDo)
             {
                 if (fileLocation.TryReconstructAbsoluteUri(_run.OriginalUriBaseIds, out Uri uri))
                 {
-
                     Encoding encoding = null;
 
-                    if (!string.IsNullOrWhiteSpace(node.Encoding))
+                    string encodingText = node.Encoding ?? _run.DefaultFileEncoding;
```
michaelcfanning: >_run.DefaultFileEncoding [](start = 59, length = 24)

Oops. Didn't honor DefaultFileEncoding previously. Added test.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -39,24 +50,9 @@ public override Run VisitRun(Run node)
...
-                foreach (string key in keys)
-                {
-                    var value = node.Files[key];
-                    if (value != null)
-                    {
-                        node.Files[key] = VisitDictionaryValueNullChecked(key, value);
-                    }
-                }
-            }
+            Run visited = base.VisitRun(node);
```
lgolding: >visited [](start = 16, length = 7)

Ultra-nit (I'm not even asking you to change this, since you're clearly expressing a personal preference here): The method declaration already tells us it returns a `Run`, so IMO we don't need the extra documentation we get by assigning the result of this call to a temp variable.

Again, that's just me: I leave the line `return base.VisitRun(node)` alone.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -39,24 +50,9 @@ public override Run VisitRun(Run node)
                 visitor.VisitRun(node);
             }
 
-            if (node.Files != null)
```
lgolding: >null [](start = 30, length = 4)

Why do we no longer visit the `run.files` dictionary?

lgolding: Oh! because `base.VisitRun` does it. You're not doing anything derived-class-specific here; you're just doing the deep dive that the base class already does. Good catch!

---
In reply to: [214519859](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214519859) [](ancestors = 214519859)

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -32,14 +32,20 @@ public static int Run(MergeOptions mergeOptions)
...
                 
                 var formatting = mergeOptions.PrettyPrint
                     ? Formatting.Indented
                     : Formatting.None;
 
-                Directory.CreateDirectory(mergeOptions.OutputFolderPath);
+                if (!Directory.Exists(outputDirectory))
+                {
+                    Directory.CreateDirectory(mergeOptions.OutputFolderPath);
+                }
```
lgolding: >} [](start = 16, length = 1)

Oh wow. I think I understand that you're doing this as a micro-optimization, but the code is a bit unclear.

It works because if the calculated `outputDirectory` doesn't exist, it must have been specified in the `mergeOptions` -- because otherwise, `outputDirectory` would be the current working directory, and that _must_ exist. Therefore it is safe to say, as you do here: "if the output directory doesn't exist, get it from the options." If you'd just written "if the output directory doesn't exist, create it", I wouldn't have given it a second thought.

Finally, since `CreateDirectory` is idempotent, you might just as well call it. I don't see the savings as worth the extra line of code. I've just spent more time thinking about this then we'll save in CPU time until the heat death of the universe. Was it worth it? ;-)

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -32,14 +32,20 @@ public static int Run(MergeOptions mergeOptions)
                     combinedLog = new InsertOptionalDataVisitor(dataToInsert).VisitSarifLog(combinedLog);
                 }
 
+                string outputDirectory = mergeOptions.OutputFolderPath ?? Environment.CurrentDirectory;
```
lgolding: >outputDirectory [](start = 23, length = 15)

I see that this is a bug fix. Before it `mergeOptions.OutputFolderPath` was missing, you'd crash with a nullref on the next line.

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -39,6 +41,23 @@ public static int Run(RewriteOptions rewriteOptions)
             return 0;
         }
 
+        private static IDictionary<string, Uri> ConstructUriBaseIds(IEnumerable<string> uriBaseIds)
+        {
+            if (uriBaseIds == null) { return null; }
+
+            IDictionary<string, Uri> result = new Dictionary<string, Uri>();
```
lgolding: >result [](start = 37, length = 6)

Nit: I always suggest to Chris, when I see this, that he find another variable name, since "`result`" both is very generic and has a special meaning is SARIF. If it's me, I call it `uriBaseIdsDictionary`, but YMMV.

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -39,6 +41,23 @@ public static int Run(RewriteOptions rewriteOptions)
             return 0;
         }
 
+        private static IDictionary<string, Uri> ConstructUriBaseIds(IEnumerable<string> uriBaseIds)
```
lgolding: >ConstructUriBaseIds [](start = 48, length = 19)

Since you defined the multi-valued `UriBaseIds` command line option in the driver, I suggest you put the code to parse it there as well, rather than restricting it to a single verb in a single tool.

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -39,6 +41,23 @@ public static int Run(RewriteOptions rewriteOptions)
...
+
+            IDictionary<string, Uri> result = new Dictionary<string, Uri>();
+
+            foreach (string uriBaseId in uriBaseIds)
+            {
+                string[] tokens = uriBaseId.Split('=');
+                string key = tokens[0];
+                Uri value = new Uri(tokens[1], UriKind.Absolute);
+                result[key] = value;
+            }
```
lgolding: >} [](start = 12, length = 1)

Since the command line parser doesn't enforce the expected `x=y` format, this will crash if there is no `=` (because `tokens[0]` won't exist), or if there's nothing on the right (because `new Uri` will throw, or if there's nothing on the left (because a dictionary can't have a null key).

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -19,8 +20,9 @@ public static int Run(RewriteOptions rewriteOptions)
                 SarifLog actualLog = MultitoolFileHelpers.ReadSarifFile<SarifLog>(rewriteOptions.InputFilePath);
 
                 OptionallyEmittedData dataToInsert = rewriteOptions.DataToInsert.ToFlags();
+                IDictionary<string, Uri> originalUriBaseIds = ConstructUriBaseIds(rewriteOptions.UriBaseIds);
```
lgolding: >ConstructUriBaseIds [](start = 62, length = 19)

Now that I've seen what this code does, perhaps `ParseUriBaseIdsOption`?

src/Sarif.Multitool/RewriteCommand.cs
lgolding: In fact it could even be an extension method on `CommonOptionsBase`, so you could say
```
IDictionary<string, Uri> originalUriBaseIds = rewriteOptions.ParseUriBaseIdsOption();
```
It could do the null check if the argument is absent, just as you do here.


---
In reply to: [214520378](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214520378) [](ancestors = 214520378)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -162,38 +163,53 @@ private void ValidateResults(string output)
         }
 
         [Fact]
-        public void InsertOptionalDataVisitorTests_VisitDictionaryValueNullChecked_ValidEncoding()
+        public void InsertOptionalDataVisitorTests_ResolveOriginalUriBaseIds()
```
lgolding: >InsertOptionalDataVisitorTests_ResolveOriginalUriBaseIds [](start = 20, length = 56)

I advocate a `Noun_VerbPhrase` naming convention so the test name reads naturally in English. In this case, consider: `InsertOptionalDataVisitor_ResolvesOriginalUriBaseIds`.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -162,38 +163,53 @@ private void ValidateResults(string output)
...
-            string uriString = "file:///C:/src/foo.cs";
+            run.OriginalUriBaseIds.Should().BeNull();
+            run.Files.Keys.Count.Should().Be(1);
+            run.Files[fileKey].Contents.Should().BeNull();
 
-            FileData fileData = FileData.Create(new Uri(uriString), mimeType: "text/x-csharp");
-            fileData.Encoding = "invalid";
-            fileData.Length = 54321;
+            visitor = new InsertOptionalDataVisitor(OptionallyEmittedData.TextFiles, originalUriBaseIds);
+            visitor.VisitRun(run);
```
lgolding: >VisitRun [](start = 20, length = 8)

This is really two tests

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -162,38 +163,53 @@ private void ValidateResults(string output)
...
 
-            FileData outputFileData = visitor.VisitDictionaryValueNullChecked(uriString, fileData);
-            outputFileData.MimeType.Should().Be(fileData.MimeType);
-            outputFileData.Encoding.Should().BeNull();
-            outputFileData.Length.Should().Be(fileData.Length);
+            run.OriginalUriBaseIds.Should().Equal(originalUriBaseIds);
+            run.Files[fileKey].Contents.Text.Should().Be(File.ReadAllText(Path.Combine(testDirectory, inputFileName)));
```
lgolding: >File [](start = 57, length = 4)

In a separate PR I advocate injecting an `IFileSystem` into the `InsertOptionalDataVisitor`'s ctor so this can all be mocked and no file system access is needed.

src/Sarif.Driver/Sdk/CommonOptionsBase.cs
```diff
@@ -19,15 +20,23 @@ public class MultitoolOptionsBase
             "force",
             Default = false,
             HelpText = "Force overwrite of output file if it exists.")]
-        public bool Force { get; internal set; }
+        public bool Force { get; set; }
```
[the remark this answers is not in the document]
lgolding: Ok, I never found out. Why does this need a public setter?

---
In reply to: [214519998](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214519998) [](ancestors = 214519998)

lgolding:
A few things I think could be improved, but nothing that I'd block the merge for.

src/Sarif.Driver/Sdk/CommonOptionsBase.cs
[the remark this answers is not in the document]
michaelcfanning: You mean internal.  These are indicators that unit tests are creating/initializing these object (which are otherwise constructed only by the commandlineparser library). We could skip these hijinks by the expedient of a ctor to init...)

---
In reply to: [214521824](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214521824) [](ancestors = 214521824,214519998)

src/Sarif.Multitool/RewriteCommand.cs
michaelcfanning: yes. 'result' is a particularly poor name in a SARIF context.

---
In reply to: [214520277](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214520277) [](ancestors = 214520277)

src/Sarif.Multitool/RewriteCommand.cs
michaelcfanning: This is a common practice of mine, i don't harden helpers of this kind for specific negative cases. The crash is the current alert to the user. What I usually do is try to put a throttling catch somewhere during the command configuration phase that captures the exception (and reports it), wrapping it as the inner exception of an exception type dedicated to this general condition. This is the way the driver framework operates. Right now, we have lots of inscrutable reports due to missing/improperly constructed arguments.

---
In reply to: [214520418](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214520418) [](ancestors = 214520418)

src/Sarif.Multitool/RewriteCommand.cs
michaelcfanning: I changed it to ConstructUriBaseIdsDictionary, since this clearly expresses the outcome of the parsing.

---
In reply to: [214520447](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214520447) [](ancestors = 214520447)

src/Sarif.Multitool/MergeCommand.cs
lgolding: @michaelcfanning Won't we crash now if `OutputFolderPath` is not specified? Shouldn't this be `outputDirectory`?

---
In reply to: [214520177](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214520177) [](ancestors = 214520177)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
lgolding: @michaelcfanning I wasn't clear. I was quibbling about the subject of the sentence being "the tests". It's not the tests that resolve the uriBaseIds, it's the visitor. Thus my naming suggestion `InsertOptionalDataVisitor_ResolvesOriginalUriBaseIds`.

---
In reply to: [214551684](https://github.com/Microsoft/sarif-sdk/pull/1062#discussion_r214551684) [](ancestors = 214551684,214520596)

PR #1063, opened 2018-08-31
Release CSD1.0.2 with Result Matching Bugfixes

Evmaus-MS:
Also proves out the CI/CD pipeline that's been set up for the SARIF SDK.


michaelcfanning:
What's going on here? Are these a bunch of changes we failed to squash on merge to master? Why are they recapitulated? :)

Git commit hygiene is hard.

michaelcfanning:
Why are we seeing all these commits? Did we fail to squash on a previous merge to master?

Evmaus-MS:
So I think I've identified the issue.  We should either merge the squashed commit back to develop immediately (should be no changes, but it's technically a "new commit"), or we should use a different commit model that preserves the history.

PR #1067, opened 2018-09-02
Reconcile SDK schema with current spec schema

michaelcfanning:
We maintain three copies of the schema, the definitive version in the OASIS repo, the one here, and the one published to schemastore.org. This change attempts to sync the SDK version with the spec. On acceptance, I will update schemastore.

Changes include:
* Incidental updates of strings, whitespace, etc.
* attachment no longer can publish arbitrary additional properties
* add the new 'externalFiles' object (@lgolding your attention here particularly necessary)
* some incidental reordering
* preserves SDK specific removal of 'ruleMessageId'
* preserves SDK schema change that makes 'results' not required (there is an open issue on this in the OASIS repo)
* preserves SDK schema requirement that defaultLevel's default is 'warning'
* Renames SDK schema to conform to OASIS repo naming convention.


michaelcfanning:
@Evmaus-MS @rtaket @lgolding 
Larry, this is an important change, your careful review of the schema changes would be appreciated. We will have another chance to examine deltas on pushing this back to the OASIS repo (as the SDK version is preferable in some cases).

michaelcfanning:
It'd be nice to get this reviewed before #1058, so that Everett can refresh his code gen. @lgolding, any chance we can pull you back from your righteous pursuit of SARIF validation? :)

lgolding:
@michaelcfanning Looking now.

src/ReleaseHistory.md
```diff
@@ -193,10 +193,12 @@
...
+* BREAKING CHANGE: Eliminate result.ruleMessageId (in favor of result.message.messageId)
 
-## **v2.0.0-csd.1.0.2** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1.0.1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1.0.1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1.0.1)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1.0.1))
+## **v2.0.0-csd.1.0.2** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1.0.2) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1.0.2) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1.0.2)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1.0.2))
 * Bugfix to result matching algorithm where empty or null previous log sets caused a NullReferenceException.
 * Bugfix to result matching algorithm where we were incorrectly detecting duplicate data across files, and changed a "NotImplementedException" to the correct "InvalidOperationException".
+* BREAKING CHANGE: invocation.workDirectory is now a FileLocation object (and not a URI expressed as a string)
+* Add run.externalFiles object to schema
```
lgolding: >externalFiles [](start = 10, length = 13)

We've already published NuGet packages 1.0.2, and we don't yet have `run.externalFiles`. That will have to be in a future version.

src/Sarif/Autogenerated/FileChange.cs
```diff
@@ -34,11 +34,11 @@ public SarifNodeKind SarifNodeKind
         /// <summary>
         /// The location of the file to change.
         /// </summary>
-        [DataMember(Name = "fileLocation", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "fileLocation", IsRequired = true)]
         public FileLocation FileLocation { get; set; }
 
         /// <summary>
-        /// An array of replacement objects, each of which represents the replacement of a single range of bytes in a single file specified by 'fileLocation'.
+        /// An array of replacement objects, each of which represents the replacement of a single range of bytes in a single file specified by 'uri'.
```
lgolding: >range of bytes [](start = 98, length = 14)

"range of bytes" => "region". Let's keep track of the changes I mention here and push them back into OASIS.

lgolding: >uri [](start = 144, length = 3)

The OASIS schema is wrong and the SDK's version is right. The property is named `fileLocation`, not `uri`.

src/Sarif/Autogenerated/Fix.cs
```diff
@@ -32,7 +32,7 @@ public SarifNodeKind SarifNodeKind
         }
 
         /// <summary>
-        /// A message that describes the proposed fix, enabling viewers to present the proposed change to an end user.
+        /// A plain text message that describes the proposed fix, enabling viewers to present the proposed change to an end user.
```
lgolding: >plain text [](start = 14, length = 10)

The OASIS schema is wrong and the SDK's version is right. This property is a `message` object, not a plain text message.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -92,6 +97,19 @@ private static string Redact(string text, IEnumerable<string> tokensToRedact)
...
 
+        private static Uri Redact(Uri uri, IEnumerable<string> tokensToRedact)
+        {
+            if (uri == null) { return uri; }
+
+            string uriText = uri.OriginalString;
+
+            foreach (string tokenToRedact in tokensToRedact)
+            {
+                uriText = uriText.Replace(tokenToRedact, SarifConstants.RemovedMarker);
```
lgolding: >RemovedMarker [](start = 72, length = 13)

Can we call it `RedactedMarked` while we're at it?

src/ReleaseHistory.md
michaelcfanning: Got it.

---
In reply to: [214740439](https://github.com/Microsoft/sarif-sdk/pull/1067#discussion_r214740439) [](ancestors = 214740439)

src/Sarif/Autogenerated/FileChange.cs
michaelcfanning: fixed. note that i will push the updated SDK schema back to OASIS in a separate change. the goal is a shared file.

---
In reply to: [214740638](https://github.com/Microsoft/sarif-sdk/pull/1067#discussion_r214740638) [](ancestors = 214740638)

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: Good bug. the value of this constant was the old [REMOVED] value which you and Jim updated to [REDACTED] in v2. fixed that as well.

---
In reply to: [214741183](https://github.com/Microsoft/sarif-sdk/pull/1067#discussion_r214741183) [](ancestors = 214741183)

src/Sarif/Autogenerated/Result.cs
```diff
@@ -50,6 +50,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "message", IsRequired = false, EmitDefaultValue = false)]
         public Message Message { get; set; }
 
+        /// <summary>
+        /// A string that identifies the message within the metadata for the rule used in this result.
+        /// </summary>
+        [DataMember(Name = "ruleMessageId", IsRequired = false, EmitDefaultValue = false)]
+        public string RuleMessageId { get; set; }
```
lgolding: >RuleMessageId [](start = 22, length = 13)

@michaelcfanning ?? It looks like you _added_ `RuleMessageId`, but we just got finished removing it.

lgolding: Oh, I'm not looking at the latest commit.

---
In reply to: [215004356](https://github.com/Microsoft/sarif-sdk/pull/1067#discussion_r215004356) [](ancestors = 215004356)

PR #1071, opened 2018-09-06
Add missing Autogenerated files.

lgolding:
@michaelcfanning made a schema change that added an `externalFiles` object to the schema, but neither `ExternalFiles.cs` nor `ExternalFilesEqualityComparer.cs` was committed because `.gitignore` mentions `/src/Sarif/Autogenerated`.

I started a thread with Michael to settle whether that line should be in `.gitignore`. For now, I’m force-committing the missing files.


lgolding:

lgolding:

PR #1073, opened 2018-09-07
Fix #1045: PropertyBagConverter uses JavaScriptStringEncode

### LINKED ISSUE #1045: PropertyBagConverter uses JavaScriptStringEncode
@boAndron reported this bug:

The `PropertyBagConverter` class uses `System.Web.HttpUtility.JavaScriptStringEncode(serializedValue)` to encode the value of property values if they are strings (line 48). This is inconsistent with the serialization of any other string in the model, because `JavaScriptStringEncode` will replace some control characters with their Unicode value (such as apostrophes to `\u0027`). Also, the encoding is done just the deserialization step (not the serialization step), leading to more inconsistency.

lgolding:
This bug caused string-valued properties to be deserialized incorrectly, so when they were re-serialized, the resulting string did not match the original string. The fix is to use `JsonConvert.ToString` rather than `JavaScriptStringEncode`.

Also:
- Start marking tests with the bug the were introduced to regress, using the XUnit `Trait` attribute.


lgolding:

lgolding:
@boAndron is added to the review. #Closed

lgolding:

lgolding:

src/Sarif.FunctionalTests/PropertyBagConverterTests.cs
```diff
@@ -0,0 +1,52 @@
...
+  ""version"": """ + SarifUtilities.SarifFormatVersion + @""",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""CodeScanner""
+      },
+      ""properties"": {
+        ""int"": 42,
+        ""string"": ""'\""\\'""
+      }
```
lgolding: @boAndron: This is a minimal repro based on the sample you provided. It fails without the fix, exactly as your sample did, and succeeds with the fix.

src/Sarif.FunctionalTests/SarifLogEqualityComparerTests.cs
```diff
@@ -4,10 +4,11 @@
 using System.IO;
 using FluentAssertions;
 using Microsoft.CodeAnalysis.Sarif.Readers;
+using Microsoft.CodeAnalysis.Sarif.TestUtilities;
 using Newtonsoft.Json;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Sarif.UnitTests.Autogenerated
+namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests
```
lgolding: >FunctionalTests [](start = 39, length = 15)

Namespace was wrong.

src/Sarif.FunctionalTests/PropertyBagConverterTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+  ]
+}";
```
lgolding: @michaelcfanning: Yeah, yeah, test framework, yada yada. We'll get there.

src/Sarif.FunctionalTests/PropertyBagConverterTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            var serializerSettings = new JsonSerializerSettings
+            {
+                ContractResolver = SarifContractResolver.Instance,
+                Formatting = Formatting.Indented,
```
EasyRhin0: >, [](start = 48, length = 1)

Nit

PR #1077, opened 2018-09-08
Provide v1 downgrade. Fix log file emit bug. Update schema to current.

michaelcfanning:
@Evmaus-MS @EasyRhinoMSFT @rtaket @lgolding

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -202,7 +202,7 @@
 * Bugfix to result matching algorithm where we were incorrectly detecting duplicate data across files, and changed a "NotImplementedException" to the correct "InvalidOperationException".
 
 ## **v2.0.0-csd.1.0.3** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1.0.3) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1.0.3) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1.0.3)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1.0.3))
-* Bugfix to result matching algorithm where empty or null previous log sets caused a NullReferenceException.
-* Bugfix to result matching algorithm where we were incorrectly detecting duplicate data across files, and changed a "NotImplementedException" to the correct "InvalidOperationException".
 * BREAKING CHANGE: invocation.workDirectory is now a FileLocation object (and not a URI expressed as a string)
-* Add run.externalFiles object to schema
+* Add run.externalFiles object to schema. Sync generally to OASIS TC schema.
+* Add --sarif-version command to driver (to transform SARIF output to v1 format)
+* Fix bug that persisted tokens to redact as file references.
--- a/src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
+++ b/src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Reflection;
 using FluentAssertions;
 using Microsoft.CodeAnalysis.Sarif.Readers;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
 using Newtonsoft.Json;
 using Xunit;
 
@@ -505,6 +506,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                     ConfigurationFilePath = configFileName ?? TestAnalyzeCommand.DefaultPolicyName,
                     Recurse = true,
                     OutputFilePath = path,
+                    SarifVersion = SarifVersion.TwoZeroZero
                 };
 
                 var command = new TestAnalyzeCommand();
@@ -557,6 +559,53 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
             toolNotificationCount.Should().Be(1);
             configurationNotificationCount.Should().Be(0);
+        }
+
+
+        [Fact]
+        public void AnalyzeCommand_PersistsSarifOneZeroZero()
+        {
+            string fileName = GetThisTestAssemblyFilePath();
+            string path = Path.GetTempFileName();
+
+            try
+            {
+                var options = new TestAnalyzeOptions
+                {
+                    TargetFileSpecifiers = new string[] { fileName },
+                    Verbose = true,
+                    Statistics = true,
+                    Quiet = true,
+                    ComputeFileHashes = true,
+                    ConfigurationFilePath = TestAnalyzeCommand.DefaultPolicyName,
+                    Recurse = true,
+                    OutputFilePath = path,
+                    PrettyPrint = true,
+                    SarifVersion = SarifVersion.OneZeroZero
+                };
+
+                var command = new TestAnalyzeCommand();
+                command.DefaultPlugInAssemblies = new Assembly[] { this.GetType().Assembly };
+                int result = command.Run(options);
+
+                result.Should().Be(0);
+
+                command.RuntimeErrors.Should().Be(RuntimeConditions.None);
+
+                JsonSerializerSettings settings = new JsonSerializerSettings()
+                {
+                    ContractResolver = SarifContractResolverVersionOne.Instance
+                };
+
+                SarifLogVersionOne log = JsonConvert.DeserializeObject<SarifLogVersionOne>(File.ReadAllText(path), settings);
+                Assert.NotNull(log);
+                Assert.Equal<int>(1, log.Runs.Count);
+
+            }
+            finally
+            {
+                File.Delete(path);
+            }
         }
 
         [Fact]
--- a/src/Sarif/Writers/SarifLogger.cs
+++ b/src/Sarif/Writers/SarifLogger.cs
@@ -13,7 +13,7 @@ using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Sarif.Writers
 {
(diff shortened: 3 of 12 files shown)


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -121,18 +121,19 @@ private static Uri Redact(Uri uri, IEnumerable<string> tokensToRedact)
...
-                  dataToInsert,
-                  tool,
-                  run,
-                  invocationTokensToRedact,
-                  invocationPropertiesToLog)
+                  loggingOptions: loggingOptions,
+                  dataToInsert: dataToInsert,
+                  tool: tool,
+                  run: run,
+                  analysisTargets: analysisTargets,
```
michaelcfanning: >analysisTargets [](start = 18, length = 15)

failure to explicitly provide analysisTarget argument results in invocationsToRedact argument being passed as targets list. gap in unit tests. we need this fix asap to fix issues in checked in binskim test files. PENDING

michaelcfanning:
This change adds a new --sarif-version argument to standard commands. This arg maps to the SarifVersion enum. When set to OneZeroZero, the driver framework invokes a SarifLogger wrapper that on calls to Dispose reads the persisted v2 file, transforms to v1 then rewrites the v2 file with v1 contents. Not efficient but a surprisingly small amount of code.

lgolding:
> # Sarif Driver and SDK Release History

Since this file is included only in Sarif.Sdk.nupkg, I think it should only have Sarif.Sdk-related changes. As matters stand, Sarif.Driver.nupkg doesn't have a history file at all.

---
Refers to: src/ReleaseHistory.md:1 in 7df67c3. [](commit_id = 7df67c36449a3e89427b086beb05af3f6235fd8b, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -202,7 +202,7 @@
 * Bugfix to result matching algorithm where we were incorrectly detecting duplicate data across files, and changed a "NotImplementedException" to the correct "InvalidOperationException".
 
 ## **v2.0.0-csd.1.0.3** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.0.0-csd.1.0.3) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.0.0-csd.1.0.3) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.0.0-csd.1.0.3)) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.0.0-csd.1.0.3))
-* Bugfix to result matching algorithm where empty or null previous log sets caused a NullReferenceException.
-* Bugfix to result matching algorithm where we were incorrectly detecting duplicate data across files, and changed a "NotImplementedException" to the correct "InvalidOperationException".
 * BREAKING CHANGE: invocation.workDirectory is now a FileLocation object (and not a URI expressed as a string)
-* Add run.externalFiles object to schema
+* Add run.externalFiles object to schema. Sync generally to OASIS TC schema.
+* Add --sarif-version command to driver (to transform SARIF output to v1 format)
```
lgolding: > Add --sarif-version command to driver (to transform SARIF output to v1 format) [](start = 1, length = 79)

This is the line that made me realize that we're mixing the history into one file even though there are two packages.

src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -559,6 +561,53 @@ public void AnalyzeCommand_DefaultEndToEndAnalysis()
             configurationNotificationCount.Should().Be(0);
         }
 
+
+        [Fact]
+        public void AnalyzeCommand_PersistsSarifOneZeroZero()
```
lgolding: >AnalyzeCommand_PersistsSarifOneZeroZero [](start = 20, length = 39)

Good name.

src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -559,6 +561,53 @@ public void AnalyzeCommand_DefaultEndToEndAnalysis()
...
+                    ConfigurationFilePath = TestAnalyzeCommand.DefaultPolicyName,
+                    Recurse = true,
+                    OutputFilePath = path,
+                    PrettyPrint = true,
+                    SarifVersion = SarifVersion.OneZeroZero
+                };
+
+                var command = new TestAnalyzeCommand();
+                command.DefaultPlugInAssemblies = new Assembly[] { this.GetType().Assembly };
+                int result = command.Run(options);
```
lgolding: >result [](start = 20, length = 6)

Consider: `returnValue`

src/Sarif.Driver.UnitTests/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -559,6 +561,53 @@ public void AnalyzeCommand_DefaultEndToEndAnalysis()
...
+                command.RuntimeErrors.Should().Be(RuntimeConditions.None);
+
+                JsonSerializerSettings settings = new JsonSerializerSettings()
+                {
+                    ContractResolver = SarifContractResolverVersionOne.Instance
+                };
+
+                SarifLogVersionOne log = JsonConvert.DeserializeObject<SarifLogVersionOne>(File.ReadAllText(path), settings);
+                Assert.NotNull(log);
+                Assert.Equal<int>(1, log.Runs.Count);
```
lgolding: >Assert [](start = 16, length = 6)

Use FluentAssertions:
```
log.Should().NotBeNull();
log.Runs.Count.Should().Be(1);
```

src/Sarif/Writers/SarifOneZeroZeroLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+            Run run = null,
+            IEnumerable<string> analysisTargets = null,
+            IEnumerable<string> invocationTokensToRedact = null,
+            IEnumerable<string> invocationPropertiesToLog = null,
+            string defaultFileEncoding = null)
+            : base(new StreamWriter(new FileStream(outputFilePath, FileMode.Create, FileAccess.Write, FileShare.None)),
+                  loggingOptions,
+                  dataToInsert,
+                  tool,
+                  run,
```
lgolding: >run [](start = 18, length = 3)

BUG: `analysisTargets` not passed down to base class.

src/Sarif/Writers/SarifOneZeroZeroLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+                  invocationPropertiesToLog)
+        {
+            _outputFilePath = outputFilePath;
+        }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+
+            JsonSerializerSettings v2Settings = new JsonSerializerSettings()
```
lgolding: Shouldn't `base.Dispose` happen last? Conceivably the derived class `Dispose` method might want to use a disposable resource held by the base class, and now it's gone.

src/Sarif/Writers/SarifOneZeroZeroLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif.Writers
+{
+    /// <summary>
+    /// A SARIF logger that works by generating a SARIF v2 log file which is transformed into 
+    /// SARIF v1 when the instance is disposed. The file location used to produced the preliminary
+    /// v2 content is overwritten in place to produce the transformed v1 file.
```
lgolding: >file [](start = 73, length = 4)

That is awfully clever :-) And clever code needs a good comment, and this is one.

src/Sarif/Writers/SarifOneZeroZeroLogger.cs
Evmaus-MS: base.Dispose() finishes writing the original SARIF log, which I think is then needed here--this seems correct to me, at least as far as the base class is implemented.

---
In reply to: [216135766](https://github.com/Microsoft/sarif-sdk/pull/1077#discussion_r216135766) [](ancestors = 216135766)

src/Sarif/Writers/SarifOneZeroZeroLogger.cs
michaelcfanning: Oof. You are too polite to say it, so i will :MILE WIDE UNIT TEST HOLE. I will address shortly (not this PR).

---
In reply to: [216135732](https://github.com/Microsoft/sarif-sdk/pull/1077#discussion_r216135732) [](ancestors = 216135732)

lgolding:
> # Sarif Driver and SDK Release History

How can it be "by design" that one package has a release history that mentions changes to another package, and the other package doesn't have a release history at all? I could buy "pending" with an issue filed.

---
In reply to: [419657275](https://github.com/Microsoft/sarif-sdk/pull/1077#issuecomment-419657275) [](ancestors = 419657275)

---
Refers to: src/ReleaseHistory.md:1 in 7df67c3. [](commit_id = 7df67c36449a3e89427b086beb05af3f6235fd8b, deletion_comment = False)

michaelcfanning:
> # Sarif Driver and SDK Release History

Yes, my apologies, mismanaged this. Turns out that SDK and Multitool both reference the release markdown. Each package should have its own. Will file a bug.

---
In reply to: [420353600](https://github.com/Microsoft/sarif-sdk/pull/1077#issuecomment-420353600) [](ancestors = 420353600,419657275)

---
Refers to: src/ReleaseHistory.md:1 in 7df67c3. [](commit_id = 7df67c36449a3e89427b086beb05af3f6235fd8b, deletion_comment = False)

PR #1081, opened 2018-09-10
Introduce new validation verification method

lgolding:
We introduce a new method of verifying the results of running a validation rule on a SARIF test file. The test file itself will now contain a custom property that summarizes the expected results. In particular, it will contain the set of JSON pointers where the results should be found.

With this design, there is no need for a separate "_Expected.sarif" file, and no need for the `SelectiveCompare` method.

In this PR, we demonstrate this change on a single, "valid" file. That means that there is no code yet to actually compare the lists of JSON pointers. That will come next.

We also do not yet have a script to auto-populate this custom property.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif.Multitool.FunctionalTests/TestData/UriMustBeAbsolute/Valid.sarif
```diff
@@ -38,7 +38,12 @@
         {
           "uri": "https://example.com/my-project"
         }
-      ]
+      ],
+      "properties": {
+        "expectedResults": {
+          "resultLocationPointers": []
```
michaelcfanning: >resultLocationPointers [](start = 11, length = 22)

shouldn't this expected results object contain a rule id of some kind?

lgolding: I considered it. I didn't do it because I reasoned "all the results will have the same ruleId". But now you mention it, it's possible that a bug could cause the wrong rule to fire on the expected location. Ok, I'll do it in next PR.

---
In reply to: [216740798](https://github.com/Microsoft/sarif-sdk/pull/1081#discussion_r216740798) [](ancestors = 216740798)

lgolding: No, I don't need to do that. All I have to do is have the validate method ensure that each result has the ruleId for the rule under test. I'll show you in next PR.

---
In reply to: [216741872](https://github.com/Microsoft/sarif-sdk/pull/1081#discussion_r216741872) [](ancestors = 216741872,216740798)

michaelcfanning:
Sorry, I have another comment. What you should actually consider is placing something like 'expectedRuleIds' in a property bag on the target thing where possible. 

why have indirect references, after all, in a file where the direct targets exist?

there will be scoping problems to my suggestion: not everything has a property bag. but many containers do, of course. so my idea depends on the code being able to talk up to an appropriate parent.

food for though. alternately, you could come back and assert that the JSON references won't be burdensome to update (and we'll want proof points in practice).

ALSO: please note: instead of string references to the JSON, you could populate the SARIF with a logical locations table. the logical location itself could contain the expectedRulesId. this creates more burden in the test file authoring but is a much more SARIF sort of approach. maybe the additional testing of logical locations is useful. note that this reduces fragility costs in the string references, as any changes to test file structure entails updating the unique relevant logical locations only, instead of every string reference.

lgolding:
I think it will be easy to update references. I'll demonstrate in my next PR.

src/Sarif.Multitool.FunctionalTests/TestData/UriMustBeAbsolute/Valid.sarif
lgolding: @michaelcfanning Turns out the "update" functionality isn't as easy as I thought. I'm going to let this percolate for a little while.

PR #1082, opened 2018-09-11
Add a EnumConverter for VersionOne

rtaket:
Some of the v1 SARIF files in the Results Store can't be converted because the RuleConfiguration.Configuration JSON values are enum integer values instead of enum name strings.

So I added a converter for v1 that will convert the integers into their enum type.


src/Sarif/VersionOne/Readers/EnumConverterVersionOne.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (reader == null)
+            {
+                throw new ArgumentNullException(nameof(reader));
+            }
+
+            string value;
+
+            if (reader.Value is long)
+            {
+                value = Enum.GetName(objectType, reader.Value);
```
rtaket: This is the line I added from the original EnumConverter definition.

michaelcfanning:
Let's not do this. This is a fix for version but doesn't fix other issues if someone emitted SARIF v1 without the resolver. DateTimes are also broken, I believe, for example. A better solution: we should decorate the OM so that you don't have to use the contract resolver to emit valid SARIF.

PR #1089, opened 2018-09-14
Miscellaneous schema and hint fixes related to dictionaries

lgolding:
While diagnosing a bug related to code generation for dictionaries, I inventoried all uses of `DictionaryHint` and discovered a few unrelated bugs:
- In the schema, a couple of dictionary-valued properties failed to declare the type of the dictionary values. This didn't cause a bug because the code generator defaults the dictionary value type to `string`. But it would have incorrectly allowed SARIF files in which the dictionary contained other value types to validate against the schema.
- In the code gen hints file, there were a couple of also harmless mistakes:
    - one obsolete property was still mentioned
    - one non-dictionary-valued property had a dictionary hint; this was harmless because the code generator correctly ignores dictionary hints on properties whose JSON type is not `object`.

Also:
- Improved the description of the `result.occurrenceCount` property, per @michaelcfanning.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif/CodeGenHints.json
```diff
@@ -146,11 +146,6 @@
       "kind": "DictionaryHint"
     }
   ],
-  "Invocation.ResponseFiles": [
-    {
-      "kind": "DictionaryHint"
-    }
```
lgolding: >} [](start = 4, length = 1)

This property is not a dictionary. `DictionaryHint` was (correctly) ignored.

src/Sarif/CodeGenHints.json
```diff
@@ -405,11 +400,6 @@
       "kind": "DictionaryHint"
     }
   ],
-  "Run.Rules": [
-    {
-      "kind": "DictionaryHint"
-    }
```
lgolding: >} [](start = 4, length = 1)

This property no longer exists (it lives on `Resources` object).

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1872,7 +1873,10 @@
 
         "state": {
           "description": "A dictionary, each of whose keys specifies a variable or expression, the associated value of which represents the variable or expression value. For an annotation of kind 'continuation', for example, this dictionary might hold the current assumed values of a set of global variables.",
-          "type": "object"
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
```
lgolding: >} [](start = 10, length = 1)

This bug would have allowed arbitrary property types to be present.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -761,8 +761,9 @@
         "environmentVariables": {
           "description": "The environment variables associated with the analysis tool process, expressed as key/value pairs.",
           "type": "object",
-          "additionalProperties": true,
-          "default": {}
+          "additionalProperties": {
+            "type":  "string"
+          }
```
lgolding: >} [](start = 10, length = 1)

This bug would have allowed arbitrary property types to be present.

src/Sarif/ToDotNet/ToDotNet.targets
```diff
@@ -46,12 +46,12 @@
          this issue again, we can easily enable the sarif-sdk to support the
          required object model without waiting for Microsoft.JSchema.ToDotNet
          to catch up.  
-      -->
```
lgolding: >--> [](start = 6, length = 3)

Removed trailing white space from a few lines.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1268,7 +1269,7 @@
         },
 
         "occurrenceCount": {
-          "description": "A positive integer specifying the number of times a result with this result object's correlationGuid has been observed",
+          "description": "A positive integer specifying the number of times this logically unique result was observed in this run.",
```
lgolding: >description [](start = 11, length = 11)

@michaelcfanning's improved description.

src/Sarif/CodeGenHints.json
michaelcfanning: From the property description, it sounds like this is in fact a dictionary. What am I missing?

The property names in this object SHALL contain the names of all the environment variables in the tool's execution environment. The value of each property SHALL be a string containing the value of the specified environment variable. If the value of the environment variable is an empty string, the corresponding property value SHALL be an empty string.

---
In reply to: [217861968](https://github.com/Microsoft/sarif-sdk/pull/1089#discussion_r217861968) [](ancestors = 217861968)

lgolding: As we discussed: `invocation.responseFiles` is (appropriately) a list of `fileLocation`. Since it is (correctly) not declared as an `object` in the schema, this `DictionaryHint` is ignored.

I think you are reviewing the wrong PR. You already approved this one and it is merged. Please take a look at #1091. That's the one currently in review.

PR #1091, opened 2018-09-15
Fix #1090: SarifRewritingVisitor does not visit some dictionaries.

### LINKED ISSUE #1090: SarifRewritingVisitor does not visit some dictionaries
By default, when the `JsonSchemaToDotNet` code generator encounters a property whose JSON type is `object`, it generates a .NET property of type `System.Object`. However, if you specify a `DictionaryHint` for that property, the code generator instead generates a .NET property of type `IDictionary<TKey, TValue>`. If you don't supply any arguments to the `DictionaryHint`, the code generator sets `TKey` to `string`, and it infers `TValue` from the schema. However, you can supply arguments to the dictionary hint to specify the key type, the value type, and how the value type should be treated when its cloning and visiting code is generated.

In the file `CodeGenHints.json`, the `DictionaryHint`s for `run.files` and `run.logicalLocation` do not have any arguments, and their cloning code and visiting code is correctly generated. But -- for no particular reason; probably because I'd forgotten how the code generator infers the dictionary type by default -- the `DictionaryHint`s for `resources.rules`, `run.graphs`, and `result.graphs` _do_ have arguments.

The arguments I supplied should match the defaults inferred by the code generator. But _either_ because I got those arguments wrong, _or_ because there is a bug in the code generator (this possibility is tracked in Microsoft/jschema#37), the code generator did not generator any code in the `SarifRewritingVisitor` to visit those dictionaries.

Removing the arguments from those `DictionaryHint`s allowed the code generator to correctly infer how to generate the dictionary, its cloning code, and its visiting code.

lgolding:
By default, when the `JsonSchemaToDotNet` code generator encounters a property whose JSON type is `object`, it generates a .NET property of type `System.Object`. However, if you specify a `DictionaryHint` for that property, the code generator instead generates a .NET property of type `IDictionary<TKey, TValue>`. If you don't supply any arguments to the `DictionaryHint`, the code generator sets `TKey` to `string`, and it infers `TValue` from the schema. However, you can supply arguments to the dictionary hint to specify the key type, the value type, and how the value type should be treated when its cloning and visiting code is generated.

In the file `CodeGenHints.json`, the `DictionaryHint`s for `run.files` and `run.logicalLocation` do not have any arguments, and their cloning code and visiting code is correctly generated. But -- for no particular reason; probably because I'd forgotten that the code generator can infer without any help -- the `DictionaryHint`s for `resources.rules`, `run.graphs`, and `result.graphs` _do_ have arguments.

The arguments I supplied should match the defaults inferred by the code generator. But _either_ because I got those arguments wrong, _or_ because there is a bug in the code generator (this possibility is tracked in Microsoft/jschema#37), the code generator did not generator any code in the `SarifRewritingVisitor` to visit those dictionaries.

Removing the arguments from those `DictionaryHint`s allowed the code generator to correctly infer how to generate the dictionary, its cloning code, and its visiting code.

### DIFF
--- a/src/Sarif/Autogenerated/SarifRewritingVisitor.cs
+++ b/src/Sarif/Autogenerated/SarifRewritingVisitor.cs
@@ -544,6 +544,18 @@ namespace Microsoft.CodeAnalysis.Sarif
         {
             if (node != null)
             {
+                if (node.Rules != null)
+                {
+                    var keys = node.Rules.Keys.ToArray();
+                    foreach (var key in keys)
+                    {
+                        var value = node.Rules[key];
+                        if (value != null)
+                        {
+                            node.Rules[key] = VisitNullChecked(value);
+                        }
+                    }
+                }
             }
 
             return node;
@@ -576,6 +588,19 @@ namespace Microsoft.CodeAnalysis.Sarif
                     for (int index_0 = 0; index_0 < node.CodeFlows.Count; ++index_0)
                     {
                         node.CodeFlows[index_0] = VisitNullChecked(node.CodeFlows[index_0]);
+                    }
+                }
+
+                if (node.Graphs != null)
+                {
+                    var keys = node.Graphs.Keys.ToArray();
+                    foreach (var key in keys)
+                    {
+                        var value = node.Graphs[key];
+                        if (value != null)
+                        {
+                            node.Graphs[key] = VisitNullChecked(value);
+                        }
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


src/Sarif/Autogenerated/SarifRewritingVisitor.cs
```diff
@@ -544,6 +544,18 @@ public virtual Resources VisitResources(Resources node)
...
+                    var keys = node.Rules.Keys.ToArray();
+                    foreach (var key in keys)
+                    {
+                        var value = node.Rules[key];
+                        if (value != null)
+                        {
+                            node.Rules[key] = VisitNullChecked(value);
+                        }
+                    }
+                }
```
lgolding: >} [](start = 16, length = 1)

Here, after all the shouting, is the correctly generated dictionary visitation code for `resources.rules`. `run.graphs` and `result.graphs` are visited below.

PR #1098, opened 2018-09-19
Remove duplicate FileData entries after rebasing

rtaket:


lgolding:
This needs unit tests.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -61,23 +63,27 @@ public RebaseUriVisitor(string baseName, bool rebaseRelativeUris, Uri baseUri)
             _rebaseRelativeUris = rebaseRelativeUris;
         }
 
-        public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
```
lgolding: >VisitPhysicalLocation [](start = 41, length = 21)

I take it that the reason for this change is that you want to visit every `FileLocation`, and not every `FileLocation` occurs within a `PhysicalLocation`.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -109,15 +144,15 @@ public override Run VisitRun(Run node)
         /// If we are changing the URIs in Results to be relative, we need to also change the URI keys in the files dictionary
         /// to be relative.
         /// </summary>
-        /// <param name="node">Result location being changed to relative.</param>
-        internal void RebaseFilesDictionary(PhysicalLocation node)
+        /// <param name="node">File location being changed to relative.</param>
+        internal void RebaseFilesDictionary(FileData node)
```
lgolding: >RebaseFilesDictionary [](start = 22, length = 21)

It took me a while to figure out what this did because of the misleading name. It actually rebases only a single dictionary entry, not the whole dictionary. Please rename to `RebaseFilesDictionaryEntry`.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -87,8 +93,37 @@ public override Run VisitRun(Run node)
...
+            // Save the file key in the property bag so we can remap it during rebasing.
+            foreach (string key in _files.Keys)
+            {
+                _files[key].SetProperty(OriginalFileKey, key);
+            }
+
             Run newRun = base.VisitRun(node);
 
+            // Remove the remapped files (they are duplicated during remapping).
+            foreach (string key in _files.Keys.ToArray())
```
lgolding: >.ToArray() [](start = 46, length = 10)

`.ToArray()` isn't necessary since `IDictionary.Keys` is `IEnumerable`.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -87,8 +93,37 @@ public override Run VisitRun(Run node)
...
+                {
+                    if (_files.ContainsKey(originalFileKey))
+                    {
+                        _files.Remove(originalFileKey);
+                    }
+                    else
+                    {
+                        _files[key].RemoveProperty(OriginalFileKey);
+
+                        if (_files[key].Properties.Count == 0)
```
lgolding: >_files[key].Properties.Count == 0 [](start = 28, length = 33)

Or, if you're into LINQ idioms, `if (!_files[key].Properties.Any())`

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -87,8 +93,37 @@ public override Run VisitRun(Run node)
...
+
             Run newRun = base.VisitRun(node);
 
+            // Remove the remapped files (they are duplicated during remapping).
+            foreach (string key in _files.Keys.ToArray())
+            {
+                string originalFileKey = _files[key].GetProperty<string>(OriginalFileKey);
+
+                if (originalFileKey != null)
+                {
```
lgolding: >{ [](start = 16, length = 1)

I see what's going on here. Both the original entry and the duplicate have the `OriginalFileKey` property. Get rid of the original entry altogether, and as for the duplicate (the new one), get rid of the property.

So two things:
1. This is worth a comment.
2. BUG: If you apply this transformation to a file that has _already_ been remapped (that is, its `run.files` dictionary keys are already of the form `#<base>#<relativeUri>`). In that case, you will _not_ actually duplicate the dictionary entry; when you get here you'll find that there is exactly one entry, and it is the original entry, and you're going to delete it.

lgolding: (You'll definitely need a test case for <span>#</span>2.)

---
In reply to: [218974200](https://github.com/Microsoft/sarif-sdk/pull/1098#discussion_r218974200) [](ancestors = 218974200)

lgolding: You could avoid this, I think. Why are you leaving the duplicate entries to be cleaned up later? At the point you create the duplicate (L165), you already know that you don't need the old node -- and you don't need the `OriginalFileKey` property on the new node any more either. Get rid of them both, right then, and you won't need the post-processing in `VisitRun` (the whole code block starting on L105). Then you're in no danger of clobbering nodes that weren't duplicated.

---
In reply to: [218974651](https://github.com/Microsoft/sarif-sdk/pull/1098#discussion_r218974651) [](ancestors = 218974651,218974200)

lgolding: Oh crud. You can't do that while you're in the process of enumerating the keys. Here's what you can do.

1. As soon as you create the new node:
    1. Remove the `OriginalFileKey` property from the new node.
    2. Mark the old node with a property `ToBeRemoved`.
2. In the `VisitRun` post-processing:
    - Remove any node with the `ToBeRemoved` property.

---
In reply to: [218978128](https://github.com/Microsoft/sarif-sdk/pull/1098#discussion_r218978128) [](ancestors = 218978128,218974651,218974200)

lgolding: Hah hah. You don't need the `ToBeRemoved` property. Instead:

1. As soon as you create the new node, remove its `OriginalFileKey` property.
2. In the `VisitRun` post-processing, remove any node that still has the `OriginalFileKey` property. (Those are the old nodes.)

---
In reply to: [218981486](https://github.com/Microsoft/sarif-sdk/pull/1098#discussion_r218981486) [](ancestors = 218981486,218978128,218974651,218974200)

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -17,6 +18,7 @@ public class RebaseUriVisitor : SarifRewritingVisitor
     {
         internal const string BaseUriDictionaryName = "originalUriBaseIds";
         internal const string IncorrectlyFormattedDictionarySuffix = ".Old";
+        internal const string OriginalFileKey = "OriginalFileKey";
```
michaelcfanning: >OriginalFileKey [](start = 49, length = 15)

you should initialize this to an instantiated GUID and then you will never risk a collision with an existing property

src/Sarif/Visitors/RebaseUriVisitor.cs
michaelcfanning: +1

---
In reply to: [218962150](https://github.com/Microsoft/sarif-sdk/pull/1098#discussion_r218962150) [](ancestors = 218962150)

src/Sarif/NotYetGenerated/SarifRewritingVisitor.cs
```diff
@@ -125,14 +125,56 @@ public virtual object VisitActual(ISarifNode node)
...
+            }
+
+            return (T)VisitDictionaryEntry(node, ref key);
+        }
+
+        private T VisitDictionaryEntry(ISarifNode node, ref string key) where T : class, ISarifNode
+        {
+            if (node == null)
+            {
+                throw new ArgumentNullException(nameof(key));
```
rtaket: >key [](start = 55, length = 3)

Node

src/Sarif/NotYetGenerated/SarifRewritingVisitor.cs
```diff
@@ -684,7 +726,14 @@ public virtual Run VisitRun(Run node)
                         var value = node.Files[key];
                         if (value != null)
                         {
-                            node.Files[key] = VisitNullChecked(value);
+                            string newKey = key;
+                            node.Files.Remove(key);
+                            value = VisitNullChecked(value, ref newKey);
+
+                            if (newKey != null)
```
rtaket: >newKey != null [](start = 32, length = 14)

if newKey is null, shouldn't we throw? or at least replace the original key/value?

src/Sarif/NotYetGenerated/SarifRewritingVisitor.cs
```diff
@@ -684,7 +726,14 @@ public virtual Run VisitRun(Run node)
...
+                            {
+                                node.Files[newKey] = VisitNullChecked(value);
```
rtaket: >Files [](start = 37, length = 5)

Do you want a check for a duplicate key? Then maybe raise a warning to the user?

rtaket: >VisitNullChecked [](start = 53, length = 16)

Doesn't this cause us to visit the FileData/FileLocation twice. Once from visiting the dictionaryentry and again here?

rtaket:

src/Sarif/NotYetGenerated/SarifRewritingVisitor.cs
```diff
@@ -125,14 +125,56 @@ public virtual object VisitActual(ISarifNode node)
...
+            }
+
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+
+            switch (node.SarifNodeKind)
+            {
+                case SarifNodeKind.FileData:
```
lgolding: >FileData [](start = 35, length = 8)

I guess the generator could identify the set of types that can occur as dictionary entries, and generate a `case` for each, but I'm going to consider whether there's a more generic way of doing this.

michaelcfanning:
@lgolding this is all working now.

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -109,44 +101,39 @@ public override Run VisitRun(Run node)
...
-                {
-                    _files[uriText] = new FileData()
-                    {
-                        MimeType = mimeType
-                    };
+                    key = "#" + _baseName + "#" + key.Substring(_baseUri.OriginalString.Length);
                 }
             }
+
+            if (node.ParentKey != null &&
```
michaelcfanning: We neglected the parent key case originally, now covered in unit test

michaelcfanning:
@ raket

---
In reply to: [423781549](https://github.com/Microsoft/sarif-sdk/pull/1098#issuecomment-423781549) [](ancestors = 423781549)

michaelcfanning:
@rtaket approved my changes to his changes

PR #1099, opened 2018-09-19
[Design Only] Deferred Reader

vScottLouvau:
I'd like feedback on the design, usage, and design compromises of the reader.

See the executable (won't be submitted) for sample usage. Essentially deferred reading is transparent to the consumer other than changing the ContractResolver and JsonTextReader types.

It uses much less RAM for Results, and somewhat less for files depending on how big the FileData objects are relative to the URIs. 

You can only enumerate the results in order, and you can currently randomly access any items in the Files Dictionary.

The design is generic; any JSON array can be deferred via DeferredList<T> and any JSON object (string to object map) can be deferred via DeferredDictionary<T>.

Key questions:
  - Does this user experience look reasonable?
  - Should this be submitted to Sarif.SDK? If not, where?
  - Do the usage compromises seem ok? (enumerate lists in order only)
  - How much do we want to constrain RAM vs. maintain performance? (will bring up in standup)

Thanks!

### DIFF
--- a/src/Sarif.Sdk.sln
+++ b/src/Sarif.Sdk.sln
@@ -1,4 +1,5 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.27130.2010
 MinimumVisualStudioVersion = 15.0.26124.0
@@ -66,7 +67,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{90E0CA8D-0D6
 		ReleaseHistory.md = ReleaseHistory.md
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sarif.Multitool.UnitTests", "Sarif.Multitool.UnitTests\Sarif.Multitool.UnitTests.csproj", "{706D3D4B-B669-496A-B4C2-7B9E91C4E684}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sarif.Multitool.UnitTests", "Sarif.Multitool.UnitTests\Sarif.Multitool.UnitTests.csproj", "{706D3D4B-B669-496A-B4C2-7B9E91C4E684}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SarifDeferredSample", "SarifDeferredSample\SarifDeferredSample.csproj", "{EBE491D1-790D-43CA-899B-FCD2FDD21823}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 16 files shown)


src/SarifDeferredSample/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                } 
+
+                return $"Enumerated {fileCount:n0} Files, URI total {uriLengthTotal:n0}b.";
+            });
+
+            GC.KeepAlive(log);
+        }
+
+        private static void Measure(Func<string> action)
```
lgolding: >Measure [](start = 28, length = 7)

Love it. I'm stealing this :-)

src/SarifDeferredSample/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+            Console.WriteLine($"Loading {filePath}{(deferred ? " (deferred)" : "")}...");
+
+            Measure(() =>
+            {
+                JsonSerializer serializer = new JsonSerializer();
+                serializer.ContractResolver = (deferred ? SarifDeferredContractResolver.Instance : SarifContractResolver.Instance);
+
+                using (JsonTextReader reader = (deferred ? new JsonPositionedTextReader(filePath) : new JsonTextReader(new StreamReader(filePath))))
+                {
+                    log = serializer.Deserialize<SarifLog>(reader);
```
lgolding: >Deserialize [](start = 37, length = 11)

No matter the outcome of this investigation, we should change the usage pattern recommended in [https://github.com/Microsoft/sarif-sdk/blob/master/docs/how-to.md](https://github.com/Microsoft/sarif-sdk/blob/master/docs/how-to.md). It recommends reading the entire file contents into a string, then deserializing with `JsonConvert.DeserializeObject<T>`. Instead, it should use this pattern of creating a reader, creating a serializer, and calling `JsonSerializer.Deserialize<T>`. The current pattern is simpler, but we shouldn't recommend a pattern that uses that much memory.

src/SarifDeferredSample/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+                    {
+                        FileData file = run.Files[fileKey];
+                        uriLengthTotal += file?.FileLocation?.Uri?.OriginalString?.Length ?? 0;
+                    }
+                } 
+
+                return $"Enumerated {fileCount:n0} Files, URI total {uriLengthTotal:n0}b.";
+            });
+
+            GC.KeepAlive(log);
```
lgolding: >log [](start = 25, length = 3)

Interesting. Should our `JsonSchemaToDotNet` code generator have an option to make the root class (`SarifLog`, in our case) `IDisposable`, so this is not necessary? (This is all new to me, by the way -- I had to read [The Old New Thing -- When do I need to use GC.KeepAlive](https://blogs.msdn.microsoft.com/oldnewthing/20100813-00/?p=13153) to understand this.

The downside would be that you'd _have_ to remember to dispose the log -- that is, you'd have to remember to create it in a `using` statement -- to prevent it from leaking.

src/SarifDeferredSample/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            string message = action();
+
+            w.Stop();
+            long ramAfter = GC.GetTotalMemory(true);
+
+            Console.WriteLine($"{message} in {w.ElapsedMilliseconds:n0}ms using {(ramAfter - ramBefore) / (BytesPerMB):n3}MB RAM.");
+        }
+
+        private static string SeekAndRead(string filePath, long position)
```
lgolding: >SeekAndRead [](start = 30, length = 11)

Not used.

lgolding:
Scott, the design looks fine to me. I will defer (:-)) detailed review of the implementation classes `DeferredDictionary`, `DeferredList`, etc. until your next push.

src/Sarif.Sdk.sln
```diff
@@ -1,4 +1,5 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
vScottLouvau: Removed. Weird.

src/SarifDeferredSample/Program.cs
vScottLouvau: =) Yep. C# null propagation makes the usual safety checks much more pleasant to read.

src/SarifDeferredSample/Program.cs
vScottLouvau: Yes. Using a JsonTextReader instead of File.ReadAllText halves the peak memory use when reading any Sarif.

src/SarifDeferredSample/Program.cs
vScottLouvau: GC.KeepAlive is only used here to make the memory use measurements accurate. Otherwise, release builds will sometimes optimize away things it knows the downstream code isn't using.

src/SarifDeferredSample/Program.cs
vScottLouvau: Yep - this was for debugging weird behaviors. The whole executable won't be in the final submit.

src/Sarif/Readers/DeferredListConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            JsonPositionedTextReader r = reader as JsonPositionedTextReader;
+            if (r == null) throw new InvalidOperationException($"DeferredListConverter requires a JsonPositionedTextReader be used for deserialization.");
+
+            return new DeferredList<T>(serializer, r);
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            // Default Serialization is fine
```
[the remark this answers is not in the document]
vScottLouvau: Yes. It's an abstract method on the base class, so I have to.

src/Sarif/Readers/DeferredDictionaryConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    /// <summary>
+    ///  DeferredDictionaryConverter is a JsonConverter which allows reading IDictionary&lt;string, T&gt;
+    ///  items only at enumeration time, saving the memory cost of keeping every object around.
+    ///  
+    ///  The set of keys and the file position in the JSON stream of the values is pre-loaded and kept around
+    ///  for fast retrieval.
```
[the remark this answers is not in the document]
vScottLouvau: Added.

lgolding:
To answer directly the question you posed in the PR description:

1. Yes, the experience (using a "deferred contract resolver" and a "positioned text reader") is reasonable. The SDK should probably provide a helper, even so.
2. Yes, the implementation classes should be submitted to the SDK. I agree with where you put them in the source tree. Also, the SDK has a Samples directory (which at the moment is structured on the assumption that there is only one sample ;-)), and you could rework the sample app to go there.
3. Yes, I think the enumeration compromise is acceptable, but you'll want to check with Michael, since he's seen more real-world usage than I have.
4. "How much do we want to constrain RAM vs. maintain performance?" I think you're asking whether it's more important to use less memory or to be fast. From the point of view of the engineer at MicroFocus who first raised the issue, perf was more important.

PR #1100, opened 2018-09-20
DeferredReader

vScottLouvau:
The real code review. I will not submit the test executable - I'll remove it and may submit it to the Samples directory later.

DeferredList and DeferredDictionary are now more complicated. They both support loading nothing up front. Both collections can be enumerated without any extra loading or allocation; the items are read by a single reader and returned as you enumerate.

Random access is supported for both types. If you access an item by index (or key), a map is made from the item key or index to the file position of the item. Reading everything this way is slower, as the collections must seek in the file for each item read. 

The list pre-builds the map by default, but the dictionary does not. This can be overridden with a constructor argument. The list pre-build adds less than 10% to normal deserialization time, and every million results only uses 8MB in the map. The Dictionary map takes more space and significantly slows reading, so it's left until first use.

When externalFiles is present, we can initialize DeferredLists and DeferredDictionaries which will read the external files instead of a position in the base file and we can not pre-load anything. This means the user only has to wait for the base file parse unless the collections are actually used. Also, if the collections are enumerated, they'll only pay the parsing cost for the part of the collection read.


src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            }
+            else
+            {
+                reader.Skip();
+            }
+        }
+
+        private void EnsurePositionsBuilt()
+        {
+            if (_itemPositions == null) BuildPositions();
```
michaelcfanning: >BuildPositions [](start = 40, length = 14)

here's a nice nit for you, prefer adding braces even for single line if statements: 

if (_itemPostions == null) { BuildPositions(); }

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+                stream.Seek(_start, SeekOrigin.Begin);
+                reader.Read();
+
+                BuildPositions(reader, _start);
+            }
+        }
+
+        private void BuildPositions(JsonPositionedTextReader reader, long currentOffset)
+        {
+            Dictionary<int, long> result = new Dictionary<int, long>();
```
michaelcfanning: >Dictionary<int, long [](start = 12, length = 20)

nice spot for 'var' usage. YMMV

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            while (true)
+            {
+                // Find the position just before the PropertyName (we need to read it back later to confirm the string matches)
+                long keyPosition = currentOffset + reader.TokenPosition + 1;
```
michaelcfanning: > + 1 [](start = 71, length = 4)

is there a guarantee that this '+1' won't result in an exception for any corner case, due to reading beyond the literal stream?

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+                reader.Read();
+                if (reader.TokenType == JsonToken.EndObject) break;
+
+                if (reader.TokenType != JsonToken.PropertyName) throw new InvalidDataException($"@({reader.LineNumber}, {reader.LinePosition}): Expected property name, found {reader.TokenType} \"{reader.Value}\".");
```
michaelcfanning: >if [](start = 16, length = 2)

this is a rather long line. more NITS for you to consider. recall my NIT around always including braces even for single lines (as in the line previous). i will stop remarking on this now.

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+            _itemPositions = result;
+        }
+
+        public T this[string key]
+        {
+            get
+            {
+                T value;
+                if (!this.TryGetValue(key, out value)) throw new KeyNotFoundException(key);
```
michaelcfanning: >if (!this.TryGetValue(key, out value)) throw new KeyNotFoundException(key); [](start = 16, length = 75)

you can skip the var declaration and put this in a single line now by declaring the var inside the TryGetValue call, e.g.,

if (!this.TryGetValue(key, out T value)) { throw new KeyNotFoundException(key); }

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+        public bool Remove(KeyValuePair<string, T> item)
+        {
+            throw new NotSupportedException();
+        }
+        #endregion
+
+        public bool Contains(KeyValuePair<string, T> item)
+        {
+            T value;
+            if (!this.TryGetValue(item.Key, out value)) return false;
```
michaelcfanning: >this.TryGetVal [](start = 17, length = 14)

same not on inlining var declaration, will now cease repeating this.

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            public override int Read(char[] buffer, int index, int count)
+            {
+                int countRead = 0;
+
+                // We are seeking to a position just before a new PropertyName in an object.
+                if(!_readCalled)
+                {
+                    _readCalled = true;
+
+                    // We'll add a '{' to the JsonTextReader understands we're in an object
```
michaelcfanning: >to the [](start = 39, length = 6)

typo, s/be 'so the' not 'to the'

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            {
+                _jsonSerializer = jsonSerializer;
+                _streamProvider = streamProvider;
+                _start = start;
+
+                Reset();
+            }
+
+            public KeyValuePair<string, U> Current { get; private set; }
+            object IEnumerator.Current => Current;
```
michaelcfanning: >object [](start = 12, length = 6)

add 'private' modifier here

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+        }
+
+        /// <summary>
+        ///  JsonDeferredDictionaryEnumerator provides enumeration over a Dictionary in a
+        ///  Json object starting at a given position.
+        ///  
+        ///  Items are not pre-loaded and are not kept after enumeration, allowing use with
+        ///  collections too large for memory.
+        /// </summary>
+        /// <typeparam name="U">Dictionary Value item type</typeparam>
```
michaelcfanning: >U [](start = 29, length = 1)

just curious, why 'U' as a type parameter name?

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            {
+                throw new NotSupportedException();
+            }
+            #endregion
+
+            public bool Contains(U item)
+            {
+                EqualityComparer<U> comparer = EqualityComparer<U>.Default;
+
+                foreach(U value in this)
```
michaelcfanning: >foreach [](start = 16, length = 7)

here and elsewhere need space before parens, see foreach and if statements in this class

src/Sarif/Readers/DeferredDictionaryConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+    ///  
+    ///  The set of keys and the file position in the JSON stream of the values is pre-loaded and kept around
+    ///  for fast retrieval.
+    /// </summary>
+    /// <typeparam name="T">Type of items in the Dictionary</typeparam>
+    public class DeferredDictionaryConverter<T> : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(IDictionary<string, T>);
```
michaelcfanning: >objectType [](start = 19, length = 10)

Not sure how important this is but your converter on surface should work for both IDictionary<string, T> and the concrete Dictionary<string, T> types in both read and write cases. you could adjust your CanConvert implementation to reflect this, either by using IsAssignableTo or by adding a second typeof comparison.

src/Sarif/Readers/DeferredDictionaryConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            JsonPositionedTextReader r = reader as JsonPositionedTextReader;
+            if (r == null) throw new InvalidOperationException($"DeferredDictionaryConverter requires a JsonPositionedTextReader be used for deserialization.");
```
michaelcfanning: >DeferredDictionaryConverter [](start = 65, length = 27)

You can use nameof in your interpolated strings if you care to do so.

if (r == null) { throw new InvalidOperationException($"{nameof(DeferredDictionaryConverter<T>)} requires a {nameof(JsonPositionedTextReader)} be used for deserialization."); }

src/Sarif/Readers/DeferredList.cs
```diff
@@ -0,0 +1,272 @@
...
+
+            _itemPositions = positions.ToArray();
+            _count = positions.Count;
+        }
+
+        public int Count
+        {
+            get
+            {
+                if(_count < 0) EnsurePositionsBuilt();
```
michaelcfanning: >if [](start = 16, length = 2)

need space after 'if'

src/Sarif/Readers/DeferredDictionaryConverter.cs
michaelcfanning: all comments for this file apply to DeferredListConverter as well.

---
In reply to: [220976161](https://github.com/Microsoft/sarif-sdk/pull/1100#discussion_r220976161) [](ancestors = 220976161)

src/Sarif/Readers/JsonPositionedTextReader.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    ///  JsonPositionedTextReader is a JsonTextReader which exposes TokenPosition(),
+    ///  the byte offset in the stream where the current token begins. It is used to
+    ///  enable deferred collections, which look like other collections but are deserialized
+    ///  only when enumerated, saving memory.
+    /// </summary>
+    public class JsonPositionedTextReader : JsonTextReader
+    {
+        private LineMappingStreamReader StreamReader { get; set; }
+        private int LastReadLineNumber { get; set; }
```
michaelcfanning: >private int LastReadLineNumber [](start = 8, length = 30)

since these are private, why not just use instance fields? just curious

src/Sarif/Readers/JsonPositionedTextReader.cs
```diff
@@ -0,0 +1,44 @@
...
+        private int LastReadLinePosition { get; set; }
+
+        public Func<Stream> StreamProvider { get; private set; }
```
michaelcfanning: >StreamProvider [](start = 28, length = 14)

i'm not sure I understand the utility of this. in your usage, this delegate simply provides a reference to a stream, so it could just as easily have been passed as a ctor arg. if you did, in fact, use the delegate as some kind of factory, one confusion is how that stream is tracked for dispose, this type retains no reference to it.

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+                using (JsonTextReader reader = new JsonTextReader(new JsonObjectMemberStreamReader(_stream)))
+                {
+                    reader.CloseInput = false;
+
+                    // Read past the mock object start from JsonObjectMemberStreamReader [so JsonTextReader expects to see a PropertyName]
+                    reader.Read();
+
+                    // Get the string key for the item with this hash
+                    reader.Read();
+                    string foundKey = (string)reader.Value;
```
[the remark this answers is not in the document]
michaelcfanning: ok, looking at the broader context, I see I missed the obvious point: this dictionary is associated with a single JSON object, the property names of which will not be duplicated.

---
In reply to: [220969257](https://github.com/Microsoft/sarif-sdk/pull/1100#discussion_r220969257) [](ancestors = 220969257)

src/Sarif/Readers/LineMappingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    /// <summary>
+    ///  LineMappingStreamReader is a StreamReader which additionally tracks the byte offsets of each line start
+    ///  in the part of the file currently read. This allows it to map a line and char back to a byte offset
+    ///  to allow consumers to seek back to the position later.
+    /// </summary>
+    internal class LineMappingStreamReader : StreamReader
```
michaelcfanning: >LineMappingStreamReader [](start = 19, length = 23)

Would you mind taking a look at NewlineIndex.cs to see whether we have convergence possibilities?

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: I fixed this class. I'm not sure if I'll fix all of them on this visit.

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: Good question - no, because we call reader.Read() right afterward, which means there must be something afterward. If we don't get another object, the code afterward breaks or throws, so the invalid position won't be added to the collection.

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: Huh - cool. I'll start using that pattern.

src/Sarif/Readers/DeferredDictionary.cs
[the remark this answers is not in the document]
vScottLouvau: Yep - there shouldn't be duplicates, and if there are, we would only return the first one. This is similar to the Dictionary<string, FileData> would otherwise do.

src/Sarif/Readers/DeferredDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+        ///  value somewhere in an outer object.
+        ///  
+        ///  It alters the first read so that there's a fake outer object (a starting '{') and
+        ///  whitespace and any comma before the property name are hidden.
+        /// </summary>
+        private class JsonObjectMemberStreamReader : StreamReader
+        {
+            private bool _readCalled;
+
+            public JsonObjectMemberStreamReader(Stream stream) : base(stream, Encoding.Default, true, 1024, true)
```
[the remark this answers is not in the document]
vScottLouvau: Sure - fixed.

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: Hmm. It doesn't compile if I add it. I guess the two 'Currents' conflict between IEnumerable<T> and IEnumerable, so the untyped one has to be implemented explicitly and that's what's happened. This is the code VS adds when you run the 'Implement Interface' code fixer.

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: Did a format document.

src/Sarif/Readers/DeferredDictionaryConverter.cs
vScottLouvau: Good idea - fixed both.

src/Sarif/Readers/JsonPositionedTextReader.cs
vScottLouvau: Made consistent with other classes and removed the two line numbers, which are unused and were leftover.

src/Sarif/Readers/DeferredList.cs
vScottLouvau: Fixed via Format Document.

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: I made this a private class (since it's used for the containing type) and you have to use different type parameter names than the enclosing class or it's unclear to the compiler whether it's a generic class or specific to the particular <T> of the compiled outer class.

(Short answer: You get a compiler warning if you don't.)

src/Sarif/Readers/DeferredDictionary.cs
vScottLouvau: Fixed. =)

src/Sarif/Readers/JsonPositionedTextReader.cs
vScottLouvau: Added a comment on the member and constructor to explain. It's subtle.

PR #1103, opened 2018-09-27
Eliminate need to specify the SARIF contract resolver. All serializat…

michaelcfanning:
…ion and deserialization now works by default.

This change eliminates the need to specify a contract resolver in a json settings object that is passed to API such as JsonConvert to serialize/deserialize SARIF. We require special converters to perform operations such as converting [Flags] enum values into string arrays in JSON and vice verse. This change was prompted by the observation that several customers fell into the trap of producing SARIF that could not subsequently be deserialized.

The core change is accomplished by using the Jschema AttributeHint to mark relevant types with the JsonConverterAttribute (which passes the type of the converter as an argument). This converter specification cannot be overridden in cases where a containing class is being serialized or deserialized. The attribute itself is not recognized in cases where a user may be serializing/deserializing a specific member directly.

This change was complicated by inconsistencies produced by the mixture of IRule and Rule in our OM. I made the choice to drive IRule into the OM. Moving forward, I believe we should back away from this and eliminate IRule in favor of an overridable Rule class. Similarly, we should look at eliminating ISarifNode in favor of a common abstract base class for all generated SARIF nodes.

I verified this change by completing eliminating the public SarifContractResolver.Instance member that's commonly used for serializing SARIF. Once I had eliminated all uses, I restored this member and marked it as obsolete (but not an error to use). Legacy code should therefore still compile but emit a warning that this resolver is generally no longer required.


src/Sarif/CodeGenHints.json
```diff
@@ -18,6 +18,14 @@
           "override"
         ]
       }
+    },
+    {
+      "kind": "AttributeHint",
```
michaelcfanning: >AttributeHint [](start = 15, length = 13)

@lgolding FYI, this hint doesn't appear to be operational for a wildcard spec like this one.

src/Sarif/CodeGenHints.json
```diff
@@ -81,6 +89,16 @@
       }
     }
   ],
+  "FileData.LastModifiedTime": [
+    {
+      "kind": "AttributeHint",
+      "arguments": {
+        "namespaceName": "Newtonsoft.Json",
```
michaelcfanning: >Newtonsoft.Json [](start = 26, length = 15)

@lgolding, another jschema bug, the namespaceName element doesn't appear to be honored by the code emit for this kind of attribute.

src/Sarif/ExtensionMethods.cs
```diff
@@ -16,6 +16,11 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        public static string Format(this IRule rule, string messageId, IEnumerable<string> arguments)
```
michaelcfanning: >Format [](start = 29, length = 6)

This extension method created as a result of pushing IRule into the OM. Method previously existed on Rule, which is defined as a partial class. Can't do this (partial class) for an interface, obviously

src/Sarif/Autogenerated/Resources.cs
```diff
@@ -41,7 +43,8 @@ public SarifNodeKind SarifNodeKind
         /// A dictionary, each of whose keys is a string and each of whose values is a 'rule' object, that describe all rules associated with an analysis tool or a specific run of an analysis tool.
         /// </summary>
         [DataMember(Name = "rules", IsRequired = false, EmitDefaultValue = false)]
-        public IDictionary<string, Rule> Rules { get; set; }
+        [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.RuleDictionaryConverter))]
+        public IDictionary<string, IRule> Rules { get; set; }
```
michaelcfanning: >IRule [](start = 35, length = 5)

We now emit IRule into the OM. We should either prefer this or eliminate IRule altogether. Probably should prefer the latter. this would be a large breaking change, so should just do it if we are. I think it may be preferable, otherwise we need to figure out how to code gen things like IRuleEquatable.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -99,9 +99,15 @@ public T GetProperty<T>(string propertyName)
                         propertyName));
             }
 
-            return JsonConvert.DeserializeObject<T>(Properties[propertyName].SerializedValue);
+            return JsonConvert.DeserializeObject<T>(Properties[propertyName].SerializedValue, _settingsWithComprehensiveV2ContractResolver);
         }
 
+        private static readonly JsonSerializerSettings _settingsWithComprehensiveV2ContractResolver = new JsonSerializerSettings
```
michaelcfanning: >_settingsWithComprehensiveV2ContractResolver [](start = 55, length = 44)

this is how we solve the problem of persisting SARIF types that require special serialization/deserialization to and from the property bag.

src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
```diff
@@ -594,7 +594,14 @@ public virtual Resources VisitResources(Resources node)
                         var value = node.Rules[key];
                         if (value != null)
                         {
-                            node.Rules[key] = VisitNullChecked(value);
+                            ISarifNode sarifNode = value as ISarifNode;
```
michaelcfanning: >ISarifNode [](start = 28, length = 10)

This one-off code required as IRule itself doesn't require ISarifNode to implement. All this means 1) we need special-casing to emit this method. 2) a class that implements IRule needs to also implement ISarifNode in order for the visitor to walk it successfully. Neither is very desirable.

src/Sarif.UnitTests/Baseline/ResultMatching/ResultMatchingBaselinerTests.cs
```diff
@@ -58,11 +58,18 @@ public void ResultMatchingBaseliner_BaselinesTwoSimpleSarifLogs()
...
 
-                calculatedNextBaseline.Runs[0].Results.Where(r => r.BaselineState == BaselineState.Existing).Should().HaveCount(currentLog.Runs[0].Results.Count - 1);
 
-                calculatedNextBaseline.Runs[0].Results.Where(r => r.BaselineState == BaselineState.Existing).First().TryGetProperty(SarifLogResultMatcher.ResultMatchingResultPropertyName, out Dictionary<string, string> CurrentResultProperties).Should().BeTrue();
-                CurrentResultProperties.Should().ContainKey("Run");
-                CurrentResultProperties["Run"].Should().BeEquivalentTo(currentLog.Runs[0].InstanceGuid);
+                int existingCount = currentLog.Runs[0].Results.Count - 1;
+                calculatedNextBaseline.Runs[0].Results.Where(r => r.BaselineState == BaselineState.Existing).Count().Should().Be(existingCount);
+
+                if (existingCount > 0)
```
michaelcfanning: >existingCount [](start = 20, length = 13)

This is a fix for a bug in a test that produces random content (that happened to fail intermittently in one of my validation runs)

src/Sarif/Autogenerated/Resources.cs
lgolding: @michaelcfanning I totally agree. Please remind me what the original motivation was for `IRule` -- I know you just told me recently.

---
In reply to: [221101257](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221101257) [](ancestors = 221101257)

src/Sarif.UnitTests/Baseline/ResultMatching/ResultMatchingBaselinerTests.cs
```diff
@@ -24,7 +24,7 @@ public ResultMatchingBaselinerTests(ITestOutputHelper outputHelper)
 
         [Fact]
         public void ResultMatchingBaseliner_BaselinesTwoSimpleSarifLogs()
-        {
+        {
```
lgolding: >            [](start = 9, length = 11)

Superfluous white space added.

src/Sarif/CodeGenHints.json
```diff
@@ -18,10 +18,18 @@
...
+        "typeName": "JsonConverter",
+        "arguments": [ "typeof(Microsoft.CodeAnalysis.Sarif.Readers.PropertyBagConverter)" ]
+      }
     }
   ],
   "codeFlow": [
-    {
+    {
```
lgolding: >  [](start = 5, length = 1)

Nit: trailing white space.

src/Sarif/CodeGenHints.json
lgolding: That's disturbing. I filed Microsoft/JSchema#51, "AttributeHint doesn't work on a wildcard."

---
In reply to: [221087141](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221087141) [](ancestors = 221087141)

src/Sarif/CodeGenHints.json
```diff
@@ -281,6 +387,24 @@
...
+        "arguments": [ "typeof(FlagsEnumConverter)" ]
+      }
+    }
+  ],
+  "Result.WorkItemUris": [
+    {
+      "kind": "AttributeHint",
+      "arguments": {
+        "namespaceName": "Newtonsoft.Json",
+        "typeName": "JsonConverter",
```
lgolding: >JsonConverter [](start = 21, length = 13)

Does the converter work on the array elements when you apply the attribute to an array? Quick search didn't turn up the answer. Is this tested?

src/Sarif/CodeGenHints.json
```diff
@@ -379,6 +521,14 @@
   "Run.OriginalUriBaseIds": [
     {
       "kind": "DictionaryHint"
+    },
+    {
+      "kind": "AttributeHint",
+      "arguments": {
+        "namespaceName": "Newtonsoft.Json",
+        "typeName": "JsonConverter",
```
lgolding: >JsonConverter [](start = 21, length = 13)

Same question: Does the converter work on the object's property values when you apply the attribute to an object?

src/Sarif/CodeGenHints.json
```diff
@@ -445,17 +612,35 @@
       }
     }
   ],
+  "ThreadFlowLocation.Importance": [
+    {
+      "kind": "EnumHint",
```
lgolding: >EnumHint [](start = 15, length = 8)

So this was just missing -- reveals test gap.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -102,6 +102,12 @@ public T GetProperty<T>(string propertyName)
             return JsonConvert.DeserializeObject<T>(Properties[propertyName].SerializedValue);
         }
 
+        private static readonly JsonSerializerSettings _settingsWithComprehensiveV2ContractResolver = new JsonSerializerSettings
```
lgolding: >_settingsWithComprehensiveV2ContractResolver [](start = 55, length = 44)

Nit: Do we use `s_` for names of statics?

src/Sarif/Core/IRuleEqualityComparer.cs
```diff
@@ -0,0 +1,256 @@
...
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Defines methods to support the comparison of objects of type Rule for equality.
+    /// </summary>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.58.0.0")]
+    internal sealed class IRuleEqualityComparer : IEqualityComparer<IRule>
```
lgolding: >IRuleEqualityComparer [](start = 26, length = 21)

I assume this is copied from the `RuleEqualityComparer` so I won't review it.

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
```diff
@@ -594,7 +594,14 @@ public virtual Resources VisitResources(Resources node)
                         var value = node.Rules[key];
                         if (value != null)
                         {
-                            node.Rules[key] = VisitNullChecked(value);
+                            ISarifNode sarifNode = value as ISarifNode;
```
lgolding: >ISarifNode [](start = 28, length = 10)

Don't put this in the `Autogenerated` code. It will be overwritten on the next build, and you've already copied the visitor to `NotYetAutogenerated` anyway.

src/Sarif.FunctionalTests/v2/SpecExamples/Comprehensive.sarif
```diff
@@ -20,6 +20,15 @@
           "copyright": "Copyright (c) 2016 by Example Corporation. All rights reserved."
         }
       },
+      "versionControlProvenance": [
+        {
+          "uri": "https://github.com/contoso/example",
```
lgolding: >github [](start = 26, length = 6)

`example.com`

lgolding:
After reviewing this, I can't figure out exactly what forced you to make the  `Rule`-to-`IRule` change.

src/Sarif/CodeGenHints.json
lgolding: Ok, I just found the code you wrote to dive into the dictionary.

---
In reply to: [221316364](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221316364) [](ancestors = 221316364)

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -241,13 +240,14 @@ public void WriteRules(IDictionary<string, IRule> rules)
             EnsureResultsArrayIsNotOpen();
             EnsureStateNotAlreadySet(Conditions.Disposed | Conditions.RulesWritten);
 
-            _jsonWriter.WritePropertyName("resources");
-            _jsonWriter.WriteStartObject(); // Begin: resources
 
-            _jsonWriter.WritePropertyName("rules");
-            _serializer.Serialize(_jsonWriter, rules, typeof(Dictionary<string, IRule>));
```
lgolding: >IRule [](start = 80, length = 5)

If you had changed this to `Rule` (because in fact `resources.Rules` is a `Dictionary<string,Rule>), would the rest of the `Rule`-to-`IRule` changes have been needed?

src/Sarif/CodeGenHints.json
michaelcfanning: Now that you know this, I have removed the non-functional hint. I was able to get things to work by providing conversion at a lower level, the SerializedPropertyInfo...

---
In reply to: [221313782](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221313782) [](ancestors = 221313782,221087141)

michaelcfanning:
1) we have an admixture of abstract and concrete types in our OM, 2) this means that any type that implements IRule has to be converted to a Rule instance when being serialized. You cannot deserialize an alternate IRule provider (you could with a converter/contract resolver, if we took this first change). in any case, i now believe, as you agree, that we should eliminate as many interfaces as possible in preference of well-constructed unsealed base types. Our object hierarchy will start with ISarifNode by providing a kind and deep clone, we will layer property bag holding on type, and then just generate concrete types. this will be a big breaking change for some consumers (like binskim).

---
In reply to: [425515152](https://github.com/Microsoft/sarif-sdk/pull/1103#issuecomment-425515152) [](ancestors = 425515152)

src/Sarif/CodeGenHints.json
michaelcfanning: Yes, this works. The UriConverter actually operates against List<uri> as well.

---
In reply to: [221316128](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221316128) [](ancestors = 221316128)

src/Sarif/CodeGenHints.json
michaelcfanning: Yep. Found this by updating the comprehensive.sarif example. We should really examine that thing more closely and continue to fill it out. very useful in many test scenarios.

---
In reply to: [221316544](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221316544) [](ancestors = 221316544)

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
michaelcfanning: We need this in the autogenerated code. Otherwise, when others sync to this change, the new file will be copied over and everyone will have this delta in their Autogenerated copy. This is just the way things work. :)

---
In reply to: [221318810](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221318810) [](ancestors = 221318810)

src/Sarif.FunctionalTests/v2/SpecExamples/Comprehensive.sarif
michaelcfanning: microsoft has created a contoso organization on github.com, do you see an issue using it? 

---
In reply to: [221319843](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221319843) [](ancestors = 221319843)

lgolding: I didn't know that.

---
In reply to: [221400882](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221400882) [](ancestors = 221400882,221319843)

src/Sarif/Autogenerated/ResourcesEqualityComparer.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.CodeDom.Compiler;
 using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Sarif;
```
lgolding: >Sarif [](start = 29, length = 5)

Odd. Was this a hand-edit you forgot to revert? I don't remember making a code gen change that would have had the effect of generating this superfluos `using` directive.

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
```diff
@@ -592,6 +592,7 @@ public virtual Resources VisitResources(Resources node)
                     foreach (var key in keys)
                     {
                         var value = node.Rules[key];
```
lgolding: >var [](start = 24, length = 3)

Extra blank line added by hand. Just revert this auto-generated file.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -157,7 +157,7 @@ private List<ExtractedResult> GetMatchingResultsFromRuns(IEnumerable<Run> sarifR
                 {
                     foreach (Result result in run.Results)
                     {
-                        Rule rule = GetRuleFromResources(result, run.Resources.Rules);
+                        IRule rule = GetRuleFromResources(result, run.Resources.Rules);
```
lgolding: >IRule [](start = 24, length = 5)

I thought we agreed to revert all `Rule` => `IRule` changes, and then implement `IRule` => `Rule` in a separate PR.

lgolding:
There are still some left-over `Rule` => `IRule` changes. Please revert, then ok to merge.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
michaelcfanning: this rule instance isn't even used, which is how it remained out of sight. @evmaus, why is this code getting called? unnecessary code?

---
In reply to: [221404351](https://github.com/Microsoft/sarif-sdk/pull/1103#discussion_r221404351) [](ancestors = 221404351)

michaelcfanning:
Nice catch. Fixed, pushed and will merge after build succeeds. Thanks for the review.

PR #1105, opened 2018-09-30
Partially DRY out declaration of schema version.

lgolding:
### Summary

**The purpose of this PR is to minimize the cost of changing the schema version number by DRYing out, as far as practical, the version number and the schema URI that depends on it.**

This would be a good change even if we hadn't decided to improve our process by updating the schema version after each TC meeting. It was always wrong to hard-code the version number in so many places.

**NOTE:** This PR does _not_ include the actual schema changes we adopted in TC <span>#</span>24, such as removing the `hash` object. This is a large PR, and I didn't want to mix the substantive changes in with the mechanics of DRYing out and updating the schema version. 

### Details

This change lays the foundation for changing the schema's pre-release version component
after the changes from each TC meeting have been applied. For example, after the changes from TC <span>#</span>24 (held on September 26th, 2018) are applied, the pre-release version component will change to `-beta.2018-09-26`.

The problem is that the string `"2.0.0"` is hard-coded in many places:

- In the `SarifVersion` enumeration (`SarifVersion.TwoZeroZero`), which in turn is used in dozens of places throughout the code.
- In the test files, in both the `$schema` and `version` properties of the top-level JSON object.
- In the inline SARIF text used in unit tests, again in both the `$schema` and `version` properties.

The solution is:

- In `CodeGenHints.json`, rename the `SarifVersion` enumeration constant from `TwoZeroZero` to `Current`.
- In `New-VersionConstantsFile.ps1`, omit the `Version` property (which was never used) and introduce the `SemanticVersion` property.
- Remove `SarifUtilities.V2_0_0` and use the new `VersionConstants.SemanticVersion` instead.
- In the inline SARIF text used in unit tests, interpolate `SarifUtilities.SarifSchemaUri` into the `$schema` property, and `VersionConstants.SemanticVersion` into the `version` property, instead of hard-coding the string `"2.0.0"` in those paces.
- Introduce a helper `Utilities.UpdateVersionNumberToCurrent` in the `Sarif.TestUtilities` project. It accepts the text of a SARIF log file and replaces `"2.0.0"` with the current version in both the `$schema` and `version` properties.
- Whenever _test code_ reads a SARIF log from a file or from a resource stream, invoke `UpdateVersionNumberToCurrent`.

In addition, I ran `Update-Baseline.ps1` to update the functional test data files. I couldn't find an easy way to update these files as they are read in, because the converter _product code_ reads the files in, and so we can't get our hands on them before the converter processes the text of those files. Unfortunately, this affects hundreds of test files each time we bump the version number. Perhaps in a future PR I'll find a way around this, but this is already a big change. The good news is that a single command updates all the files; at least we don't have to hand-edit them.

Also:

- Fixed two bugs in `InsertOptionalDataVisitor` that caused functional tests to fail. I _think_ we didn't see these before because we hadn't regenerated the "expected" files recently:
    - The `InsertOptionalDataVisitor` unnecessarily emitted an empty `run.originalUriBaseIds`.
    - The `InsertOptionalDataVisitor` omitted two `null` checks in `VisitFileData`.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:
@michaelcfanning: Regenerating the "expected" functional test files by running `UpdateBaseline.ps1` appears to have dropped alot of generated content. See for example: `src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif`. I suspect this is a result of the `SarifRewritingVisitor` work but can't be sure.

lgolding:
Yes, looking at the missing content, it is as far as I can see all under the `files` dictionary. That makes it seem likely to me that the `InsertOptionalDataVisitor` doesn't interact properly with the modified `SarifRewritingVisitor`. I do _not_ want to fix this in this already-large PR. Let's address is separately.

---
In reply to: [425684719](https://github.com/Microsoft/sarif-sdk/pull/1105#issuecomment-425684719) [](ancestors = 425684719)

lgolding:
So we have a process problem here, in deciding when to run `UpdateBaselines.ps1`:

- If you make a change and a functional test fails, it's obvious you need to run it.
- If all the functional tests pass, you might still need to run it (as was the case here).
- But you can't just run it every time as part of the build, because then the functional tests will _always_ pass and you don't learn anything from them.

Also: I think the reason the functional tests passed _despite_ the fact that the "expected" files _would_ have changed is that we don't compare the whole files; we do this "selective compare" that's supposed to weed out the non-deterministic content. And I suspect that selective compare doesn't look at all relevant properties of the `file` object.

The root problem here is that the selective compare needs to be kept up to date with schema changes, and that's easy to forget.

The real solution is to add that "determinize" command to the multitool. If you added that to the `UpdateBaselines.ps1` script, then the "expected" files would be fully deterministic, and the functional tests could compare the entire object models with `SarifLog.ValueEquals`, and they'd always get the right answer, fully in sync with the spec.

I propose this plan:
1. Merge this PR.
2. Next PR: Implement the actual TC 24 schema changes.
3. Next PR: Write the determinize command.
4. Next PR: Modify `UpdateBaselines.ps1` to run the determinizer.
5. Next PR: Modify functional tests to compare full object models.

---
In reply to: [425684888](https://github.com/Microsoft/sarif-sdk/pull/1105#issuecomment-425684888) [](ancestors = 425684888,425684719)

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -13,7 +13,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., build.props))\build.props" />
 
   <ItemGroup>
-    <PackageReference Include="FluentAssertions" Version="5.4.0" />
+    <PackageReference Include="FluentAssertions" Version="5.4.1" />
```
lgolding: >5.4.1 [](start = 58, length = 5)

Because this project now has a reference to the `TestUtilities` project, and that project uses `5.4.1`.

src/Sarif/SarifUtilities.cs
```diff
@@ -20,14 +20,12 @@ public static bool IsSemanticVersioningCompatible(this string versionText)
             return s_semVer200.IsMatch(versionText);
         }
 
-        private const string V1_0_0 = "1.0.0";
-        private const string V1_0_0_BETA_5 = "1.0.0-beta.5";
-        private const string V2_0_0 = "2.0.0";
+        public const string V1_0_0 = "1.0.0";
+        public const string V1_0_0_BETA_5 = "1.0.0-beta.5";
```
lgolding: >V1_0_0_BETA_5 [](start = 28, length = 13)

Nobody uses this. We should remove it.

scripts/New-VersionConstantsFile.ps1
```diff
@@ -25,7 +25,7 @@ namespace $namespace
         public const string Prerelease = "$versionSuffix";
         public const string AssemblyVersion = "$versionPrefix";
         public const string FileVersion = AssemblyVersion + ".0";
-        public const string Version = AssemblyVersion + Prerelease;
+        public const string SemanticVersion = AssemblyVersion + "-" + Prerelease;
```
michaelcfanning: >SemanticVersion [](start = 28, length = 15)

Is this going to work when we don't actually have any prerelease data? Looks like we'll end up sticking a hyphen at the end of our final release.

src/Sarif/SarifUtilities.cs
michaelcfanning: go right ahead

---
In reply to: [221469436](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221469436) [](ancestors = 221469436)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "title": "Static Analysis Results Format (SARIF) Version 2.0.0 JSON Schema",
-  "description": "Static Analysis Results Format (SARIF) Version 2.0.0 JSON Schema: a standard format for the output of static analysis tools.",
+  "title": "Static Analysis Results Format (SARIF) Version 2.0.0-beta.2018-09-26 JSON Schema",
```
michaelcfanning: >beta [](start = 65, length = 4)

csd2 might be a better term here:

2.0.0-csd2.2018-09-26

or perhaps add both for clearest communication?

2.0.0-beta.csd2.2018-09-26

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -113,34 +113,36 @@ public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
         public override FileData VisitFileData(FileData node)
         {
             FileLocation fileLocation = node.FileLocation;
+            if (fileLocation != null && _run.OriginalUriBaseIds != null)
```
michaelcfanning: >_run.OriginalUriBaseIds [](start = 40, length = 23)

this isn't right. a file location might be rendered as an absolute file:// URL, which is appropriate for computing hashes or injecting contents.

michaelcfanning: this is another place where we need the key that is associated with the file node, as the file location might be expressed there. how's that full change coming along?

---
In reply to: [221752670](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221752670) [](ancestors = 221752670)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0",
```
michaelcfanning: >json [](start = 21, length = 4)

you can't check in these updates, as they have improperly stripped information we expect in the test files. in order for updating to work properly, you need to enlist in the d: drive, have you done so locally? otherwise the tools can't locate the appropriate files to inject.

note that we could make this test code more resilient by converting these file locations and using uriBaseIds. the rewriters can accept arbitrary values for the uri base id values. this rewrite always injects the non-deterministic uri base id into the rewritten file as part of originalUriBaseIds, however. would be useful to make this behavior configurable, so that we could avoid injecting it into test content. another way to do so would be to write the 'remove non-deterministic content' rewrite feature which has been open for a while.

michaelcfanning:
I think the issue may be simpler: you haven't enlisted in SARIF with d:\src\sarif-sdk as a root. this is required for some PREfast content in order to inject code snippets as a post-processing step.

---
In reply to: [425685813](https://github.com/Microsoft/sarif-sdk/pull/1105#issuecomment-425685813) [](ancestors = 425685813,425684888,425684719)

src/Sarif/Schemata/sarif-schema.json
lgolding: That sounds good. Be aware that according to SemVer the, pre-release version is supposed to sort chronologically (so that, for example, `alpha.1` < `alpha.2` < `beta.1`. Whether we stay with what Jim proposed, adopt one of your two suggestions, or go with `csd2.beta.YYYY-MM-DD` (which I like better because "beta" qualifies "CSD.2" -- these are beta versions of CSD.2(, we will break that rule.

I'm going to go with `2.0.0-csd.2.beta.YYYY-MM-DD`.

---
In reply to: [221751980](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221751980) [](ancestors = 221751980)

scripts/New-VersionConstantsFile.ps1
lgolding: Nope, that's a bug. So I won't be able to get away with `SemanticVersion` being a compile-time constant, so I'll expose it as a computed property in `SarifUtilities`. Pity.

As I mentioned elsewhere in my comments, the `Version` property as defined wasn't used anywhere, so I'll just stop generating it.

---
In reply to: [221750223](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221750223) [](ancestors = 221750223)

michaelcfanning:
All you need to do is to 'net use' to d: from your c drive and run the update script again.

net use d: \\blah\c$ etc.

---
In reply to: [426067905](https://github.com/Microsoft/sarif-sdk/pull/1105#issuecomment-426067905) [](ancestors = 426067905,426061247,425685813,425684888,425684719)

lgolding:
Thanks, that sure helps. I'll still ask `UpdateBaseline.ps1` to help out.

---
In reply to: [426070608](https://github.com/Microsoft/sarif-sdk/pull/1105#issuecomment-426070608) [](ancestors = 426070608,426067905,426061247,425685813,425684888,425684719)

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -36,6 +36,8 @@ public void ValidatesAllTestFiles()
             foreach (string inputFile in TestCases)
             {
                 string instanceText = File.ReadAllText(inputFile);
+                instanceText = Utilities.UpdateVersionNumberToCurrent(instanceText);
```
michaelcfanning: >UpdateVersionNumberToCurrent [](start = 41, length = 28)

we're going to need a general utility that upgrades in-flight v2 SARIF. ideally, users will be able to seamlessly load their older SARIF v2 which will be automatically transformed to the current version. if we don't build this, we'll need to provide API snapshots for all versions and require users to handle their own fix-ups. not a good idea. let's huddle with scott louvau on an efficient approach for this, he'll have good ideas to offer.

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/AA_IndexAlias.cpp.xml.sarif
lgolding: Running `UpdateBaseline.ps1` in an enlistment rooted in `D:\src\sarif-sdk` restored the `region` and `snippet` properties.

---
In reply to: [221754585](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221754585) [](ancestors = 221754585)

lgolding: ... but it does _not_ restore the `content` and `hashes` properties in the `files` dictionary (see `Sarif.FunctionalTests\v2\ConverterTestData\FxCop\SarifSdkTests.xml.sarif. I still think this might be due to the recent dictionary visitor work. @michaelcfanning.

---
In reply to: [222757596](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r222757596) [](ancestors = 222757596,221754585)

lgolding: As I just mailed you: it's because that file was generated in an enlistment rooted at `E:/src/sarif-sdk-test`, not `D:/src/sarif-sdk`. I'm going to fix that.

---
In reply to: [222761104](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r222761104) [](ancestors = 222761104,222757596,221754585)

lgolding: Fixed.

---
In reply to: [222785738](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r222785738) [](ancestors = 222785738,222761104,222757596,221754585)

src/Sarif/SarifUtilities.cs
lgolding: Fixed.

---
In reply to: [221750460](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221750460) [](ancestors = 221750460,221469436)

scripts/New-VersionConstantsFile.ps1
lgolding: Fixed.

---
In reply to: [221759286](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221759286) [](ancestors = 221759286,221750223)

src/Sarif/Schemata/sarif-schema.json
lgolding: Fixed.

---
In reply to: [221758555](https://github.com/Microsoft/sarif-sdk/pull/1105#discussion_r221758555) [](ancestors = 221758555,221751980)

PR #1106, opened 2018-10-01
DeferredWriter

vScottLouvau:
Adding 'SarifWriter', a writer for the Sarif format which takes Runs, Files, and Results as they are generated and writes them with minimal memory use.

Each collection is written to a separate file and then merged into the main file during Dispose. This is fast as it's just a Stream copy, so overall writing time is still slightly faster than writing normally with JsonSerializer.Serialize.

When the 'ExternalFiles' design is in the SDK, we can write the externalFiles attributes and avoid merging the files at all; I'll likely add a constructor enum to allow the caller to specify whether they'd like merging.


src/Sarif/Readers/LineMappingStreamReader.cs
```diff
@@ -13,123 +12,142 @@ namespace Microsoft.CodeAnalysis.Sarif.Readers
...
+        private FilePosition CountUpTo(int index)
+        {
+            if (index < 0 || index >= _bufferLength) throw new IndexOutOfRangeException("index");
 
-            // Ensure space to hold start of each line
-            if (_lineStartByteOffsets == null || _lineStartByteOffsets.Length < charsRead) _lineStartByteOffsets = new int[charsRead];
-            if (_lineStartIndices == null || _lineStartIndices.Length < charsRead) _lineStartIndices = new int[charsRead];
-            _lineCount = 0;
+            FilePosition current = _lastReturnedPosition;
+            if (_lastReturnedPosition.BufferIndex > index) current = _bufferStartPosition;
```
EasyRhin0: >current = _bufferStartPosition; [](start = 59, length = 31)

Put this on its own line with curly braces. I find this improves readability. StyleCop 1503

EasyRhin0: OK so I confirmed that our coding standard permits the single-line format, but it needs to have curly braces.
if (blah) { method(); }

---
In reply to: [222033337](https://github.com/Microsoft/sarif-sdk/pull/1106#discussion_r222033337) [](ancestors = 222033337)

vScottLouvau:
Formatting fixed, merging.

PR #1114, opened 2018-10-18
Update schema and tests to SARIF TC#25 conclusions, 2.0.0-csd.2.beta.2018-10-10

michaelcfanning:
Changes in this PR:

CRITICAL FILES TO REVIEW:

Sarif\Schema\sarif-schema.json and

Sarif\Writers\PrereleaseCompatibilityTransformer.cs

Update schema/OM/etc. to schema as of SARIF TC#25, 10-10-2018

originalUriBaseIds is now a dictionary of file locations, not strings.
Suffix invocation.startTime, invocation.endTime, file.lastModifiedTime and notification.time with Utc (startTimeUtc, endTimeUtc, etc.).
threadflowLocation.timestamp renamed to 'executionTimeUtc'.
versionControlDetails.timestamp renamed to 'asOfTimeUtc'.
versionControlDetails.uri renamed to 'repositoryUri'.
exception.message type converted from string to message object.
all the things have property bags
Provide PrereleaseCompatibilityTransformer functionality that automatically upgrades prerelease v2 to current.

NOTE: this mechanism is used to update all existing checked in test files (a significant proof point).
Author new comprehensive prerelease SARIF v1 file to ensure code coverage of all transformations
TODO: author detailed unit tests for this transformer
Remove all inlined schema property bag definitions in preferences of reference to single 'propertyBag' definition

Remove -NoBuildSample from build.

Clean-up alphabetization of CodeGenHints.json

Convert all SARIF v1 <-> v2 tests to operate against checked in files rather than embedded string literals.

Move transformer tests (and test data files) and rename to conform to existing standards.

Delete 'StepValuesMustFormOneBasedSequence' rule. Note to self: the requirement to author a validation rule sometimes a signal of a design flaw. :)

@lgolding @rtaket

### DIFF
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -91,7 +91,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 }
             }
 
-            return true;
+            return modifiedLog;
         }
 
         private static bool UpdateSarifLogVersion(JObject sarifLog)


scripts/BeforeBuild.ps1
```diff
@@ -16,8 +16,6 @@
     Do not restore NuGet packages.
 .PARAMETER NoObjectModel
     Do not rebuild the SARIF object model from the schema.
-.PARAMETER NoBuildSample
```
michaelcfanning: >NoBuildSample [](start = 11, length = 13)

We have too many knobs in our build script and no easy ways to do common scenarios such as (please just rebuild and run tests rather than everything else). We also don't succeed in Debug configurations.

src/Sarif/CodeGenHints.json
```diff
@@ -360,11 +520,6 @@
       "kind": "DictionaryHint"
     }
   ],
-  "Result.Fingerprints": [
```
michaelcfanning: >Result.Fingerprints [](start = 3, length = 19)

FYI, i realphabetized some content in this file, as it had gotten out of date based on some renaming. This construct has moved earlier in the file, for example.

src/Sarif.TestUtilities/Utilities.cs
```diff
@@ -17,27 +17,6 @@ namespace Microsoft.CodeAnalysis.Sarif
...
-        // in the test files every time this happens, we replace the version
-        // in the test file with the current version after reading the file
-        // into memory.
-        private const string VersionPropertyPattern = @"""version""\s*:\s*""[^""]+""";
-        private static readonly Regex s_VersionRegex = new Regex(VersionPropertyPattern, RegexOptions.Compiled);
-
-        private const string SchemaPropertyPattern = @"""\$schema""\s*:\s*""[^""]+""";
-        private static readonly Regex s_SchemaRegex = new Regex(SchemaPropertyPattern, RegexOptions.Compiled);
-
-        public static string UpdateVersionNumberToCurrent(string logContents)
```
michaelcfanning: >UpdateVersionNumberToCurrent [](start = 29, length = 28)

This utility also previously replaced the run.tool version! Oops. We now use the JParser to make the sarifLog.version update.

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidVersionControlDetailsUri_Expected.sarif
```diff
@@ -35,7 +35,7 @@
                 },
                 "region": {
                   "startLine": 12,
-                  "startColumn": 50
+                  "startColumn": 60
```
michaelcfanning: >startColumn [](start = 19, length = 11)

introduction of 'repository' to property name offsets startColumn by 10 chars

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidUriInOriginalUriBaseIds_Expected.sarif
```diff
@@ -34,8 +34,8 @@
                   "uri": "file:///C:/Code/sarif-sdk/src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/InvalidUriInOriginalUriBaseIds.sarif"
                 },
                 "region": {
-                  "startLine": 10,
-                  "startColumn": 51
+                  "startLine": 11,
```
michaelcfanning: >startLine [](start = 19, length = 9)

conversion of originalUriBaseIds values from string to file location introduces a new line in the file, hence changes in offsets

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -251,8 +251,7 @@ private HashSet<string> CreateTargetsSet(TOptions analyzeOptions)
                 var fileSpecifier = new FileSpecifier(normalizedSpecifier, recurse: analyzeOptions.Recurse, fileSystem: FileSystem);
                 foreach (string file in fileSpecifier.Files) { targets.Add(file); }
             }
-            Debug.Assert(targets.Count > 0);
```
michaelcfanning: >Debug.Assert [](start = 12, length = 12)

this assert prevents debug unit tests from completing

michaelcfanning:
Changes in this PR:

CRITICAL FILES TO REVIEW: 
* Sarif\Schema\sarif-schema.json and 
* Sarif\Writers\PrereleaseCompatibilityTransformer.cs

* Update schema/OM/etc. to schema as of SARIF TC#25, 10-10-2018
  - originalUriBaseIds is now a dictionary of file locations, not strings.
  - Suffix invocation.startTime, invocation.endTime, file.lastModifiedTime and notification.time with Utc (startTimeUtc, endTimeUtc, etc.).
  - threadflowLocation.timestamp renamed to 'executionTimeUtc'.
  - versionControlDetails.timestamp renamed to 'asOfTimeUtc'.
  - versionControlDetails.uri renamed to 'repositoryUri'.
  - exception.message type converted from string to message object.
  - all the things have property bags
* Provide PrereleaseCompatibilityTransformer functionality that automatically upgrades prerelease v2 to current.
  - NOTE: this mechanism is used to update all existing checked in test files (a significant proof point).
  - Author new comprehensive prerelease SARIF v1 file to ensure code coverage of all transformations
  - TODO: author detailed unit tests for this transformer
* Remove all inlined schema property bag definitions in preferences of reference to single 'propertyBag' definition
* Remove -NoBuildSample from build.
* Clean-up alphabetization of CodeGenHints.json
* Convert all SARIF v1 <-> v2 tests to operate against checked in files rather than embedded string literals.
* Move transformer tests (and test data files) and rename to conform to existing standards.
* Delete 'StepValuesMustFormOneBasedSequence' rule. Note to self: the requirement to author a validation rule sometimes a signal of a design flaw. :)

@lgolding @rtaket

src/Sarif/CodeGenHints.json
```diff
@@ -63,6 +83,26 @@
...
+    {
+      "kind": "BaseTypeHint",
+      "arguments": {
+        "baseTypeNames": [
+          "PropertyBagHolder"
+        ]
+      }
+    }
+  ],
+  "externalFiles": [
```
lgolding: >externalFiles [](start = 3, length = 13)

This doesn't look right. The `run` object has a property `externalFiles`, which is of type "array of `externalFile`, and the `externalFile` object is the thing that has a property bag.

src/Sarif/CodeGenHints.json
```diff
@@ -512,13 +657,25 @@
   "Run.OriginalUriBaseIds": [
     {
       "kind": "DictionaryHint"
-    },
+    }
```
lgolding: >} [](start = 4, length = 1)

I assume you omitted the `AttributeHint` because with the new "chained" design for `originalUriBaseIds`, the dictionary entries are no longer necessarily URIs.

src/Sarif/ExtensionMethods.cs
```diff
@@ -16,6 +16,40 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        public static string InstanceIdInstanceComponent(this RunAutomationDetails runAutomationDetails)
```
lgolding: >InstanceIdInstanceComponent [](start = 29, length = 27)

Consider: Make this an extension method on `string` so it would be used like this:
```
string instanceComponent = runAutomationDetails.InstanceId.InstanceComponent;
```
That seems more natural to me than
```
string instanceComponent = runAutomationDetails.InstanceIdInstanceComponent;
```
and it reduces the coupling of this method (you could unit test it without having the definition of `RunAutomationDetails`).

src/Sarif/ExtensionMethods.cs
```diff
@@ -16,6 +16,40 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+            {
+                return null;
+            }
+
+            if (!instanceId.Contains("/"))
+            {
+                return String.Empty;
+            }
+
+            return instanceId.Substring(0, instanceId.LastIndexOf('/') + 1);
```
lgolding: >+ 1 [](start = 71, length = 3)

This includes the final slash. I don't agree with that. If `instanceId` is `x/y`, then the "category" (the logical id) is `x`, not `x/`. I know you love your trailing slashes on directory-valued URIs, but I don't think it's right here.

lgolding:
>       },

You omitted the declaration of the `properties` property, here and in some other classes (`message`, `rectangle`, `logicalLocation`, and perhaps others) that are newly declared as deriving from `PropertyBagHolder`. You did add it to the `sarifLog` class.

This won't affect C# code because `PropertyBagHolder` itself declares a `Properties` dictionary. But without the declaration in the schema, an `attachement` object with a `properties` property won't validate.


---
Refers to: src/Sarif/Schemata/sarif-schema.json:74 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/Sarif/CodeGenHints.json
lgolding: Sorry, I got this wrong. The spec says that the value of `run.externalFiles` is an object, and then it goes on to describe the properties of that object -- but it doesn't give that set of properties a name, unlike every other object in the spec. I'll file an editorial issue to fix that.

---
In reply to: [226708519](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226708519) [](ancestors = 226708519)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1693,6 +1660,41 @@
...
 
+
+    "runAutomationDetails": {
+      "description": "Information that describes a run's identity and role within an engineering system process.",
+      "additionalProperties": false,
+      "type": "object",
+      "properties": {
+
+        "description": {
+          "description": "A description of the role played within the engineering system by this object's containing run object.",
```
lgolding: >  [](start = 46, length = 1)

Nit: The spec says here "identity and role", which I think is useful as it captures the purpose of `instanceGuid` and the trailing component of `instanceId`.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1693,6 +1660,41 @@
...
+          "$ref": "#/definitions/message"
+        },
+
+        "instanceId": {
+          "description": "A hierarchical string that uniquely identifies this object's containing run object.",
+          "type": "string"
+        },
+
+        "instanceGuid": {
+          "description": "A string that uniquely identifies this object's containing run object in the form of a GUID.",
```
lgolding: >string that uniquely identifies [](start = 28, length = 31)

The spec might not be uniform in its phrasing. I like the formulation "A stable, unique identifier for <the thing> in the form of a GUID."

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1693,6 +1660,41 @@
...
+          "type": "string"
+        },
+
+        "correlationGuid": {
+          "description": "A stable, unique identifier for the equivalence class of runs to which this object's containing run object belongs in the form of a GUID.",
+          "type": "string"
+        },
+
+        "properties": {
+          "description": "Key/value pairs that provide additional information about the scan automation details.",
```
lgolding: >scan [](start = 88, length = 4)

scan => run

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1705,7 +1707,7 @@
         },
 
         "frames": {
-          "description": "An array of stack frames that represent a sequence of calls, rendered in reverse chronological order, that comprise the call stack.",
+          "description": "An array of stack frames that represents a sequence of calls, rendered in reverse chronological order, that comprise the call stack.",
```
lgolding: >represents [](start = 56, length = 10)

Thanks.

lgolding: Although I suppose you could argue whether the word modifies "array" or "stack frames".

---
In reply to: [226731080](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226731080) [](ancestors = 226731080)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1901,7 +1897,7 @@
         },
 
         "properties": {
-          "description": "Key/value pairs that provide additional information about the code location.",
+          "description": "Key/value pairs that provide additional information about the threadflow location.",
           "type": "object",
```
lgolding: >type": "object", [](start = 11, length = 16)

DRY this one out to `#/definitions/propertyBag`.

lgolding:
>           "type": "object",

DRY this one out to `propertyBag`.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:1970 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>           "type": "object",

DRY this one out to `propertyBag`.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:391 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>           "type": "object",

DRY this one out to `propertyBag`.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:528 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>           "type": "object",

DRY this one out to `propertyBag`.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:582 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>           "type": "object",

DRY this one out to `propertyBag`.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:752 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>           "type": "object",

I'll stop enumerating these. Find all of them by searching for `"tags":`.

---
In reply to: [431446344](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431446344) [](ancestors = 431446344)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:752 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -106,7 +108,7 @@ internal ExceptionDataVersionOne CreateExceptionData(ExceptionData v2ExceptionDa
                 {
                     InnerExceptions = v2ExceptionData.InnerExceptions?.Select(CreateExceptionData).ToList(),
                     Kind = v2ExceptionData.Kind,
-                    Message = v2ExceptionData.Message,
+                    Message = v2ExceptionData.Message.Text,
```
lgolding: >. [](start = 53, length = 1)

"." => "?."

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -226,26 +228,20 @@ internal FixVersionOne CreateFix(Fix v2Fix)
...
 
-            if (v2Hash != null)
+            foreach (string key in v2Hashes.Keys)
             {
-                AlgorithmKindVersionOne algorithm;
-                if (!Utilities.AlgorithmNameKindMap.TryGetValue(v2Hash.Algorithm, out algorithm))
-                {
-                    algorithm = AlgorithmKindVersionOne.Unknown;
-                }
+                Utilities.AlgorithmNameKindMap.TryGetValue(key, out AlgorithmKindVersionOne algorithm);
```
lgolding: >Utilities [](start = 16, length = 9)

Retain the logic that if this method returns `false`, `algorithm` should be `AlgorithmKindVersionOne.Unknown`.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -864,16 +860,19 @@ internal RunVersionOne CreateRun(Run v2Run)
                     _currentRun = run;
 
                     run.Architecture = v2Run.Architecture;
-                    run.AutomationId = v2Run.AutomationLogicalId;
                     run.BaselineId = v2Run.BaselineInstanceGuid;
                     run.Files = v2Run.Files?.ToDictionary(v => v.Key, v => CreateFileData(v.Value));
-                    run.Id = v2Run.InstanceGuid;
+
+                    run.Id = v2Run.Id?.InstanceGuid;
+                    run.AutomationId = v2Run.AggregateIds?.First().InstanceId;
```
lgolding: >First [](start = 59, length = 5)

`FirstOrDefault()?.InstanceId`, on the off chance that the array is present but empty.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -124,7 +125,7 @@ internal ExceptionData CreateExceptionData(ExceptionDataVersionOne v1ExceptionDa
                 {
                     InnerExceptions = v1ExceptionData.InnerExceptions?.Select(CreateExceptionData).ToList(),
                     Kind = v1ExceptionData.Kind,
-                    Message = v1ExceptionData.Message,
+                    Message = v1ExceptionData.Message.ToMessage(),
```
lgolding: >. [](start = 53, length = 1)

"." => "?."

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -191,6 +192,20 @@ internal FileData CreateFileData(FileDataVersionOne v1FileData)
...
 
+        private IDictionary<string, string> BuildHashesDictionary(IList<HashVersionOne> hashes)
+        {
+            if (hashes == null) { return null; }
+
+            var v2Hashes = new Dictionary<string, string>();
+
+            foreach (HashVersionOne v1Hash in hashes)
+            {
+                v2Hashes[Utilities.AlgorithmKindNameMap[v1Hash.Algorithm]] = v1Hash.Value;
```
lgolding: >Algorithm [](start = 63, length = 9)

Is there a mapping for every value of `v1Hash.Algorithm` or do we need `TryGet` logic?

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -72,9 +72,9 @@ public Result CalculateNewBaselineResult()
                 OriginalResultMatchingProperties = new Dictionary<string, object>();
             }
 
-            if (PreviousResult.OriginalRun.InstanceGuid != null)
+            if (PreviousResult.OriginalRun.Id.InstanceGuid != null)
```
lgolding: >. [](start = 45, length = 1)

"." => "?."

lgolding: Most of the changed lines below contain multi-level property accesses that require "?." at each level after the first.

---
In reply to: [226746235](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226746235) [](ancestors = 226746235)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.CodeAnalysis.Sarif.Writers;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.UnitTests.Writers
+{
+    public class PrereleaseCompatibilityTransformerTests
+    {
+        [Fact]
+        public void PrereleaseCompatibilityTransformer_UpgradePrereleaseTwoZeroZero()
```
lgolding: Ultra-nit: "upgrages" ("The transformer upgrades the old release").

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            string comprehensiveSarifPath = Path.Combine(Environment.CurrentDirectory, @"v2\ObsoleteFormats\ComprehensivePrereleaseTwoZeroZero.sarif");
+
+            string sarifText = File.ReadAllText(comprehensiveSarifPath);
+
+            sarifText = PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(sarifText);
+
+            // If this obsolete prerelease v2 file deserializes to the current
+            // version without raising an exception, this step succeeds.
```
lgolding: >succeeds [](start = 63, length = 8)

You can say that explicitly in FluentAssertions without requiring the comment, and with a more informative error message if the test fails:
```
Action action = () => {
    SarifLog sarifLog = JsonConvert.DeserializeObject<SarifLog>(sarifText);
    JsonConvert.SerializeObject(sarifLog);
};

action.Should().NotThrow();
```

lgolding:
>         public void ResultLogJsonWriter_WritesIdAndAutomationId()

Rename: "WritesAutomationDetails".

---
Refers to: src/Sarif.UnitTests/Writers/ResultLogJsonWriterTests.cs:184 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/Sarif.TestUtilities/Utilities.cs
lgolding: I hope this is the last time I ever think I can get away with using a regex to modify a JSON, XML, or HTML file.

---
In reply to: [226504993](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226504993) [](ancestors = 226504993)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+        private static readonly Regex s_SchemaRegex = new Regex(SchemaPropertyPattern, RegexOptions.Compiled);
+
+        public static string UpdateToCurrentVersion(string prereleaseSarifLog, bool forceUpdate = false, Formatting formatting = Formatting.None)
+        {
+            bool modifiedLog = false;
+            JObject sarifLog = JObject.Parse(prereleaseSarifLog);
+
+            // Some tests update the semantic version to current for non-updated content. For this situation, we 
+            // allow the test code to force a transform, despite the fact that the provided version doesn't call for it.
+            string version = forceUpdate ? "2.0.0" : (string)sarifLog["version"];
```
lgolding: >"2.0.0" [](start = 43, length = 7)

Suggestion: use `VersionConstants.AssemblyVersion` here and on Line 38.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+
+            switch (version)
+            {
+                // This case actually handles all SARIF v2 prerelease before we 
+                // started properly updating the schema.
```
lgolding: >schema [](start = 49, length = 6)

Good comment.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                case "2.0.0":
+                {
+                    modifiedLog |= ApplyCoreTransformations(sarifLog);
+                    break;
+                }
+
+                case "2.0.0-csd.2.beta.2018-10-10":
+                {
+                    modifiedLog |= ApplyCoreTransformations(sarifLog);
+                    modifiedLog |= ApplyChangesSinceTechnicalCommitteeMeeting25();
```
lgolding: >ApplyChangesSinceTechnicalCommitteeMeeting25 [](start = 35, length = 44)

Yes, nice design.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                    break;
+                }
+
+                default:
+                    break;
+            }
+
+            string text = modifiedLog ? sarifLog.ToString(formatting) : prereleaseSarifLog;
```
lgolding: >text [](start = 19, length = 4)

Or just dispense with the temporary and return the result of the ternary expression.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+
+            return text;
+        }
+        private static bool ApplyCoreTransformations(JObject sarifLog)
```
lgolding: >p [](start = 8, length = 1)

Nit: Insert blank line above.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+            if (version != SarifUtilities.SemanticVersion)
+            {
+                sarifLog["version"] = SarifUtilities.SemanticVersion;
+                modifiedLog = true;
+            }
+
+            string schema = (string)sarifLog["$schema"];
+            if (schema != SarifUtilities.SarifSchemaUri)
+            {
+                sarifLog["$schema"] = SarifUtilities.SarifSchemaUri;
```
lgolding: >SarifSchemaUri [](start = 53, length = 14)

So happy to see these in use!

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+        {
+            bool modifiedLog = UpdateSarifLogVersion(sarifLog); 
+
+            var runs = (JArray)sarifLog["runs"];
+
+            if (runs != null)
+            {
+                // Move result.ruleMessageId: https://github.com/oasis-tcs/sarif-spec/issues/216
+                // Update name of threadflowLocation.executionTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                modifiedLog |= UpdateRunResults(runs);
```
lgolding: >UpdateRunResults [](start = 31, length = 16)

Oh my heavens! I _love_ this approach!

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+            {
+                JArray results = (JArray)run["results"];
+
+                if (results == null) { continue; }
+
+                foreach (JObject result in results)
+                {
+                    // https://github.com/oasis-tcs/sarif-spec/issues/216
+                    //
+                    // result.ruleMessageId is deprecated. This property value
```
lgolding: >deprecated [](start = 47, length = 10)

It's not deprecated, it's just plain gone.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                    // e.g.:
+                    // "message": { "messageId" : "default"}
+
+                    string ruleMessageId = (string)result["ruleMessageId"];
+
+                    if (ruleMessageId != null)
+                    {
+                        result.Property("ruleMessageId").Remove();
+                        result["message"]["messageId"] = ruleMessageId;
+                        modifiedRun = true;
```
lgolding: >modifiedRun [](start = 24, length = 11)

This is my only quibble with this design: that you have to do this at every change; duplicated code, easy to forget to do it. It would be slower but easier and more reliable to just compare the root-level `JObject`s at the end.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: >UpdateRunResults [](start = 31, length = 16)

Naming nit: Since almost everything is under `Run`, you could just name the methods `UpdateResults`, `UpdateInvocation`, _etc._

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                // Update name of file.lastModifiedTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                modifiedLog |= UpdateRunFiles(runs);
+
+                // Update names of notification.startTimeUtc and endTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                modifiedLog |= UpdateRunNotifications(runs);
+
+                // Update name to versionControlDetails.repositoryUri: https://github.com/oasis-tcs/sarif-spec/issues/244
+                modifiedLog |= UpdateVersionControlProvenance(runs);
+
+                modifiedLog |= UpdateRunAutomationDetails(runs);
```
lgolding: >Update [](start = 31, length = 6)

Although it's a naming inconsistency with the other methods, a more accurate description of what this method does is `IntroduceRunAutomationDetails`.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: Random thought: We may well update `results` or other objects more than once. I suggest method names like `UpdateResultsToTc27` if that happens.

---
In reply to: [226761963](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226761963) [](ancestors = 226761963)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+            if (runs != null)
+            {
+                // Move result.ruleMessageId: https://github.com/oasis-tcs/sarif-spec/issues/216
+                // Update name of threadflowLocation.executionTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                modifiedLog |= UpdateRunResults(runs);
+
+                // invocation.workingDirectory now a file location: https://github.com/oasis-tcs/sarif-spec/issues/222
+                // Update names of invocation.startTimeUtc and endTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                // Convert exception.message to message object: https://github.com/oasis-tcs/sarif-spec/issues/240
+                modifiedLog |= UpdateRunInvocations(runs);
```
lgolding: >runs [](start = 52, length = 4)

Now that I've seen a few of these methods, and they _all_ loop over runs, I suggest hoisting the loop into this method and passing a single `run` into each called method.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                    run.Remove("description");
+                    run.Remove("instanceGuid");
+                    run.Remove("correlationGuid");
+
+                    var runId = new JObject();
+
+                    if (instanceGuid != null)
+                    {
+                        // We can only effectively populate the new instanceId in a case where
+                        // the log if previously uniquely identified a run by a guid.
```
lgolding: >if  [](start = 35, length = 3)

Extra word "if".

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                // This property is an element of what v2 now refers to as aggregateIds
+                JToken automationLogicalId = run["automationLogicalId"];
+
+                if (automationLogicalId != null)
+                {
+                    run.Remove("automationLogicalId");
+
+                    var aggregateId = new JObject();
+
+                    // For the aggregating automat id, we have to provide only the
```
lgolding: >the [](start = 79, length = 3)

"... only the logical id component"?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                    aggregateId["instanceId"] = automationLogicalId + "/";
+
+                    var aggregateIds = new JArray(aggregateId);
```
lgolding: >aggregateIds [](start = 24, length = 12)

Remove temporary?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+            }
+            return modifiedRun;
+        }
+
+        private static bool UpdateVersionControlDetails(JObject versionControlDetails)
+        {
+            bool modifiedVersionControlDetails = false;
+
+            // https://github.com/oasis-tcs/sarif-spec/issues/242
+            modifiedVersionControlDetails |= RenameProperty(versionControlDetails, previousName: "timestamp", newName: "asOfTimeUtc");
```
lgolding: >RenameProperty [](start = 45, length = 14)

Nice! This was so smooth I almost didn't notice that you'd introduced this helper; it seemed like part of the language :-)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                    if (threadFlowLocations == null) { continue; }
+
+                    foreach (JObject threadFlowLocation in threadFlowLocations)
+                    {
+                        // Update name of threadFlowLocation.executionTimeUtc: https://github.com/oasis-tcs/sarif-spec/issues/242
+                        modifiedThreadFlowLocation |= RenameProperty(threadFlowLocation, previousName: "timestamp", newName: "executionTimeUtc");
+
+                        // Delete threadFlowLocation.step property: https://github.com/oasis-tcs/sarif-spec/issues/203
+                        JToken step = threadFlowLocation["step"];
+                        if (step?.Type == JTokenType.Integer)
```
lgolding: >if (step?.Type == JTokenType.Integer) [](start = 24, length = 37)

Why the test? You wouldn't keep it around if it were (for example) a string, right?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+
+                foreach (JObject invocation in invocations)
+                {
+                    modifiedRun |= UpdateInvocationObject(invocation);
+                }
+            }
+            return modifiedRun;
+        }
+
+        internal static bool UpdateInvocationObject(JObject invocation)
```
lgolding: >Object( [](start = 45, length = 7)

I think `UpdateInvocation` suffices.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+                string workingDirectory = (string)invocation["workingDirectory"];
+
+                invocation.Property("workingDirectory").Remove();
+
+                var fileLocation = new JProperty("uri", workingDirectory);
+                var fileLocationObject = new JObject(fileLocation);
+                invocation["workingDirectory"] = fileLocationObject;
+
+                modifiedInvocation = true;
+            }
```
lgolding: Bonus points: DRY out by extracting a method `ReplaceUriStringWithFileLocation(JObject containing object, string propertyName)`. You have the same logic, expressed slightly differently in `UpdateOriginalUriBaseIds` below.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,553 @@
...
+        }
+
+
+        internal static bool UpdateNotifications(JArray notifications)
+        {
+            bool modifiedNotification = false;
+
+            foreach (JObject notification in notifications)
+            {
+                modifiedNotification |= UpdateNotificationObjects(notification);
```
lgolding: >Objects [](start = 58, length = 7)

You're passing only one object, so it would be `UpdateNotificationObject`, but `UpdateNotification` suffices.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using System.Resources;
 using Microsoft.CodeAnalysis.Sarif.Driver;
+using Microsoft.CodeAnalysis.Sarif.Writers;
```
lgolding: >Writers [](start = 35, length = 7)

Is this a remnant of a change you decided not to make? Since there are no other changes, and it built before, this can't be necessary (unless, I suppose, you changed the namespace of some other class that this class uses).

src/Sarif/ExtensionMethods.cs
lgolding: Although I suppose you could argue that the current design eliminates the potential error of forgetting to null-check `runAutomationDetails.InstanceId`. Ok, I withdraw this.

---
In reply to: [226710524](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226710524) [](ancestors = 226710524)

src/Sarif.Multitool.FunctionalTests/Rules/SkimmerTestsBase.cs
```diff
@@ -11,18 +11,24 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
 {
-    public abstract class SkimmerTestsBase<TSkimmer> : SarifMultitoolTestBase
+    public abstract class ValidationSkimmerTestsBase<TSkimmer> : SarifMultitoolTestBase
```
lgolding: >ValidationSkimmerTestsBase [](start = 26, length = 26)

Rename the file to match the class.

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -27,17 +105,17 @@ protected string GenerateDiffCommand(string expected, string actual)
             return String.Format(CultureInfo.InvariantCulture, "windiff \"{0}\" \"{1}\"", expected, actual);
         }
 
-        protected static bool AreEquivalentSarifLogs(string actualSarif, string expectedSarif)
+        protected static bool AreEquivalentSarifLogs<T>(string actualSarif, string expectedSarif, IContractResolver contractResolver = null)
```
lgolding: >T [](start = 53, length = 1)

Why are we exposing this?

src/Sarif.UnitTests/Visitors/SarifCurrentToVersionOneVisitorTests.cs
```diff
@@ -0,0 +1,177 @@
...
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
+using Microsoft.CodeAnalysis.Sarif.Visitors;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+using Newtonsoft.Json;
+using System;
+using System.IO;
+using System.Text;
+using Xunit;
```
lgolding: `System` usings first.

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -4,15 +4,93 @@
...
+
+        // Retrieving the source path of the tests is only used in developer ad hoc
+        // rebaselining scenarios. i.e., this path won't be consumed by AppVeyor.
+        public static string GetProductTestDataDirectory(string subdirectory = "")
+        {
+            return Path.GetFullPath(Path.Combine(@"..\..\..\..\..\src\Sarif.UnitTests\TestData", subdirectory));
+        }
+
+
+        private ITestOutputHelper _outputHelper;
```
lgolding: >private [](start = 8, length = 7)

`readonly`

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -4,15 +4,93 @@
...
+            if (Directory.Exists(OutputFolderPath))
+            {
+                Directory.Delete(OutputFolderPath, recursive: true);
+            }
+
+            Directory.CreateDirectory(OutputFolderPath);
+        }
+
+
+        protected virtual Assembly ThisAssembly { get { return this.GetType().Assembly; } }
```
lgolding: >ThisAssembly [](start = 35, length = 12)

Consider expression body syntax:
```
protected virtual Assembly ThisAssembly => this.GetType().Assembly;
```

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -4,15 +4,93 @@
...
+
+        protected virtual string OutputFolderPath { get { return Path.Combine(Path.GetDirectoryName(ThisAssembly.Location), "UnitTestOutput." + TypeUnderTest); } }
+
+        protected virtual string ProductTestDataDirectory {  get { return GetProductTestDataDirectory(TypeUnderTest); } }
+
+        protected virtual string TestLogResourceNameRoot { get { return "Microsoft.CodeAnalysis.Sarif.UnitTests.TestData." + TypeUnderTest; } }
+
+        protected string GetOutputFilePath(string directory, string resourceName)
+        {
+            string fileName = string.Format("{0}.sarif",
```
lgolding: FxCop would complain that you didn't specify `CultureInfo.InvariantCulture`.

lgolding:
Many more or less minor nits and quibbles, some potential failure-to-null-check bugs, but nothing that I have to revisit once you've addressed.

michaelcfanning:
>     <PreviousVersionSuffix>csd.1.0.3</PreviousVersionSuffix>

We didn't ship this version in NuGet, so that's not literally correct. AFAIK, this suffix is no longer used to delist publishes (its original intent). So I will just delete it.

---
In reply to: [431419604](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431419604) [](ancestors = 431419604)

---
Refers to: src/build.props:26 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -201,8 +201,14 @@
...
 * Add --sarif-version command to driver (to transform SARIF output to v1 format)
-* Fix bug that persisted tokens to redact as file references.
\ No newline at end of file
+* Fix bug that persisted tokens to redact as file references.
+* BREAKING: originalUriBaseIds is now a dictionary of file locations, not strings.
+* BREAKING: Suffix invocation.startTime, invocation.endTime, file.lastModifiedTime and notification.time with Utc (startTimeUtc, endTimeUtc, etc.).
+* BREAKING: threadflowLocation.timestamp renamed to 'executionTimeUtc'.
+* BREAKING: versionControlDetails.timestamp renamed to 'asOfTimeUtc'.
+* BREAKING: versionControlDetails.uri renamed to 'repositoryUri'.
+* BREAKING: exception.message type converted from string to message object.
```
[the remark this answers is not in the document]
michaelcfanning: Yes, also neglected file.hashes, run.id and run.aggregateIds and threadFlowLocation.step removal.

---
In reply to: [226705862](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226705862) [](ancestors = 226705862)

src/Sarif.TestUtilities/FileDiffingTests.cs
michaelcfanning: it is consumed external to this assembly, by sarif.unit. tests

---
In reply to: [226787959](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226787959) [](ancestors = 226787959)

src/Sarif.TestUtilities/FileDiffingTests.cs
michaelcfanning: yes, good place for this pattern.

---
In reply to: [226788839](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226788839) [](ancestors = 226788839)

src/Sarif.TestUtilities/FileDiffingTests.cs
michaelcfanning: good fix of code i moved. using string.format for what is effectively simple string concat not a good idea, from a micro-optimization standpoint.

---
In reply to: [226788994](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226788994) [](ancestors = 226788994)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
michaelcfanning: well, every line of the test is implicitly expected not to throw. I had this comment because the deserialization was previously the last line of the test. Now it isn't. I'll just delete the comment.

---
In reply to: [226748533](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226748533) [](ancestors = 226748533)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
michaelcfanning: Good catch. This is a problem that this breaking change is likely to provoke for others as well.

---
In reply to: [226746466](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226746466) [](ancestors = 226746466,226746235)

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: By making it singular, there is no argument. :) There was no argument before, it is just my sense that this form is less likely to be construed as an error. :)

---
In reply to: [226731287](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226731287) [](ancestors = 226731287,226731080)

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: Forgot to complete this scrub! Missed many examples, now resolved, thanks.

---
In reply to: [226731555](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226731555) [](ancestors = 226731555)

michaelcfanning:
>       },

OK, thanks for catching this. Missing constructs from our comprehensive sarif file, will take a TODO to fix that.

---
In reply to: [431436145](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431436145) [](ancestors = 431436145)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:74 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

michaelcfanning:
>           "type": "object",

fixed this everywhere

---
In reply to: [431445699](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431445699) [](ancestors = 431445699)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:1970 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

michaelcfanning:
>           "type": "object",

Just fyi, a note like: I will stop remarking on this, is a clue that it my responsible to scrub for it, as I have. :) Then I don't have to close/response to many identical notes. I will not stop explicitly commenting on your repeated notes. :)

---
In reply to: [431445929](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431445929) [](ancestors = 431445929)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:391 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

michaelcfanning:
>           "type": "object",

Whew. :) Usually, I stop commenting and request a scrub the second time I remark on the same pattern. 

---
In reply to: [431446562](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431446562) [](ancestors = 431446562,431446344)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:752 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: It does. Unknown is the first element of the enum, assigned a value of '0' (the value to which an enum variable is initialized).

---
In reply to: [226735147](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226735147) [](ancestors = 226735147)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: we actually should be scrubbing for this, nearly every array in the standard should be minValues == 1. will open a tracking issue.

---
In reply to: [226735459](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226735459) [](ancestors = 226735459)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: I'm sure this isn't correct, filed a bug. Not very high priority item, I think, v1 hash production not a common scenario. [https://github.com/Microsoft/sarif-sdk/issues/1116](https://github.com/Microsoft/sarif-sdk/issues/1116)

---
In reply to: [226736401](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226736401) [](ancestors = 226736401)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Won't take this suggestion, because some critical logical would change abruptly if that constant changes (as it should be able to do).

---
In reply to: [226758312](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226758312) [](ancestors = 226758312)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: I organized this so that we can strongly associate TC changes with code, though you would approve. :)

---
In reply to: [226759993](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226759993) [](ancestors = 226759993)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Agreed, this is error prone. I need detailed unit tests, will add a todo. The problem with your suggestion is that it is requires a deep clone of each object. we're trying to minimize traversal/object creation.

---
In reply to: [226760895](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226760895) [](ancestors = 226760895)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: 'For the aggregating automation id, we can only provide the logical component'

The reason is that we didn't previously define an instance id for this item

---
In reply to: [226764086](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226764086) [](ancestors = 226764086)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: There's actually very little shared code here. I did pull out the code in one place but the other didn't naturally utilize it. Oh well.

---
In reply to: [226765536](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226765536) [](ancestors = 226765536)

src/Sarif/Schemata/sarif-schema.json
lgolding: Almost! Do the same for `ruleConfiguration.parameters`.

---
In reply to: [226844215](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226844215) [](ancestors = 226844215,226731555)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: You added the "?" but you didn't change `First` to `FirstOrDefault`. `First` will throw if the array is empty. (I know you opened a tracking issue but we have the hood up on this one).

---
In reply to: [226844620](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226844620) [](ancestors = 226844620,226735459)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: You missed Line 102.

---
In reply to: [226844027](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226844027) [](ancestors = 226844027,226746466,226746235)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
lgolding: I won't press this point to the extent of reactivating this comment, but -- there are two advantages to `ShouldNotThrow`:
1. It makes it immediately obvious to the reader of the test what you are asserting.
2. If the assertion fires, you get a nice message like "`JsonConvert.SerializeObject(sarifLog) was not expected to throw an exception, but it threw `ArgumentNullException`" -- again making it clear what expectation was violated.

---
In reply to: [226843996](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226843996) [](ancestors = 226843996,226748533)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: Well, I meant the name to imply that we were introducing the _object_ `RunAutomationDetails`, not that we were introducing the details themselves -- as you say, they already existed -- but ok.

---
In reply to: [226844923](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226844923) [](ancestors = 226844923,226762406)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -0,0 +1,541 @@
...
+                    modifiedLog |= UpdateRunNotifications(run);
+
+                    // Update name to versionControlDetails.repositoryUri: https://github.com/oasis-tcs/sarif-spec/issues/244
+                    modifiedLog |= UpdateRunVersionControlProvenance(run);
+
+                    modifiedLog |= RefactorRunAutomationDetails(run);
+                }
+            }
+
+            return true;
```
lgolding: >true [](start = 19, length = 4)

Yikes! how did I miss this? You need to return `modifiedLog`.

src/Sarif/ExtensionMethods.cs
lgolding: (Reactivated to call my response to your attention.)

---
In reply to: [226872911](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226872911) [](ancestors = 226872911,226844096,226712702)

lgolding:
>           "type": "object",

:-)

---
In reply to: [431629068](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431629068) [](ancestors = 431629068,431445929)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:391 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

lgolding:
>     <PreviousVersionSuffix>csd.1.0.3</PreviousVersionSuffix>

`DelistCurrentPackages.cmd` still exists, so we should keep this. But thanks for explaining why you didn't change it.

So there is a subtlety to using these properties correctly. We should update `<VersionSuffix>` as soon as we update the schema. But we should not update `<PreviousVersionSuffix>` until just before we publish, so that `DelistCurrentPackage` sees the right value.

---
In reply to: [431627246](https://github.com/Microsoft/sarif-sdk/pull/1114#issuecomment-431627246) [](ancestors = 431627246,431419604)

---
Refers to: src/build.props:26 in 7aaa3e9. [](commit_id = 7aaa3e9713510791f0d7f7c25ed72d13d5ff46c4, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: This naming convention looks a bit cumbersome but is intended to reflect it operates against runs. Later, we UpdateNotifications for arrays of notifications (run has two of these, configuration and tool notifications). This becomes cleaner when I take your note to hoist the runs array iteration, so that each method does actually receive and operate on a single run.

---
In reply to: [226761963](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226761963) [](ancestors = 226761963)

michaelcfanning: This naming convention looks a bit cumbersome but is intended to reflect it operates against runs. Later, we UpdateNotifications for arrays of notifications (run has two of these, configuration and tool notifications). This becomes cleaner when I take your note to hoist the runs array iteration, so that each method does actually receive and operate on a single run.

---
In reply to: [226761963](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226761963) [](ancestors = 226761963)

src/Sarif/Schemata/sarif-schema.json
lgolding: Resolved in subsequent PR.

---
In reply to: [226873010](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226873010) [](ancestors = 226873010,226844215,226731555)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: Resolved in subsequent PR.

---
In reply to: [226873134](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226873134) [](ancestors = 226873134,226844027,226746466,226746235)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Resolved in subsequent PR.

---
In reply to: [226873071](https://github.com/Microsoft/sarif-sdk/pull/1114#discussion_r226873071) [](ancestors = 226873071,226844620,226735459)

PR #1117, opened 2018-10-22
Resolve tail issues from PR #1114

lgolding:
- Avoid two potential null derefs.
- DRY out the schema declaration of `ruleConfiguration.parameters`


src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -865,7 +865,7 @@ internal RunVersionOne CreateRun(Run v2Run)
                     run.Files = v2Run.Files?.ToDictionary(v => v.Key, v => CreateFileData(v.Value));
 
                     run.Id = v2Run.Id?.InstanceGuid;
-                    run.AutomationId = v2Run.AggregateIds?.First()?.InstanceId;
```
michaelcfanning: >First [](start = 59, length = 5)

What's the distinction here? FirstOrDefault returns null but First() throws for an empty collection?

lgolding: Exactly.

PR #1118, opened 2018-10-23
Merge properties into run objects during result matching. Also preserves current run properties.

michaelcfanning:
The information we preserve during results matching is a bit tricky. Our behavior with run-level properties is currently to start with the current run's property bag and then to merge each baseline property bag into current. 

This merging operation favors existing values. And so, the most current property values will be retained.

@rtaket @lgolding


src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
```diff
@@ -36,6 +36,11 @@ public void SarifLogResultMatcher_BaselinesTwoSimpleSarifLogs()
                 currentLog.Runs[0].Results[0].Tags.Add("New Unused Tag");
             }
 
+
+            string propertyName = "WeLikePi";
+            float propertyValue = 3.1415926535897932384626433832795F;
```
lgolding: >1415926535897932384626433832795F [](start = 36, length = 32)

Nit: Too many digits for a float, will be truncated; is this by design?

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -180,7 +181,7 @@ private Rule GetRuleFromResources(Result result, IDictionary<string, Rule> rules
 
         private SarifLog ConstructSarifLogFromMatchedResults(IEnumerable<MatchedResults> results, IEnumerable<Run> previous, IEnumerable<Run> currentRuns)
```
lgolding: >previous [](start = 115, length = 8)

While we're in here, can we rename this to `previousRuns` for consistency with `currentRuns`?

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -194,11 +195,18 @@ private SarifLog ConstructSarifLogFromMatchedResults(IEnumerable<MatchedResults>
...
             };
 
+            IDictionary<string, SerializedPropertyInfo> properties = null;
+
             if (previous != null && previous.Count() != 0)
             {
-                run.BaselineInstanceGuid = previous.First().Id.InstanceGuid;
+                // We currently only flow the baseline instance id forward (which becomes the 
+                // baseline guid for the merged log) and the baseline properties bag.
+                run.BaselineInstanceGuid = previous.First().Id?.InstanceGuid;
```
lgolding: >First [](start = 52, length = 5)

I don't understand the significance of the previous runs being plural (as opposed to a single previous run), so I don't know if it's right to use the first or the last.

I don't even know if it might not be correct to merge the nth properties dictionary (where nth is the oldest) forward into the n-1st, then merge the result into the n-2nd, and so forth until you come to the newest. It's not obvious to me that merging "first" directly up to "current" is the right thing to do.

src/Sarif/Readers/SerializedPropertyInfoEqualityComparer.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.CodeDom.Compiler;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+
+namespace Microsoft.CodeAnalysis.Sarif
```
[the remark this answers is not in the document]
michaelcfanning: Will put it in readers.

---
In reply to: [227571133](https://github.com/Microsoft/sarif-sdk/pull/1118#discussion_r227571133) [](ancestors = 227571133)

src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
```diff
@@ -74,6 +82,44 @@ public void SarifLogResultMatcher_BaselinesTwoSimpleSarifLogs()
...
+            string currentSharedPropertyValue = Guid.NewGuid().ToString();
+            string baselineSharedPropertyValue = Guid.NewGuid().ToString();
+
+            string uniqueToBaselinePropertyName = nameof(uniqueToBaselinePropertyName);
+            string uniqueToBaselinePropertyValue = Guid.NewGuid().ToString();
+
+            string uniqueToCurrentPropertyName = nameof(uniqueToCurrentPropertyName);
+            string uniqueToCurrentPropertyValue = Guid.NewGuid().ToString();
+
+            baselineLog.Runs[0].SetProperty(sharedPropertyName, currentSharedPropertyValue);
```
lgolding: >currentSharedPropertyValue [](start = 64, length = 26)

`baselineSharedPropertyValue`? Otherwise that variable isn't used.

But I'm confused because I thought you intended "shared" to mean "the same in the current run as it was in the baseline run".

lgolding:
>             if (currentLog.Runs[0].Results.Any())

I understand the comment, but... if  `currentLog.Runs[0]` has no results, then this code will not be executed. So either:
1. `RandomSarifLogGenerator.GenerateRandomAndLog` is _guaranteed_ to produce results, in which case the test is unnecessary, or
2. `RandomSarifLogGenerator.GenerateRandomAndLog` is _not_ guaranteed to produced results, in which case the baseline and current runs will _not_ diverge.

---
In reply to: [432425245](https://github.com/Microsoft/sarif-sdk/pull/1118#issuecomment-432425245) [](ancestors = 432425245,432424780)

---
Refers to: src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs:34 in 01ad15c. [](commit_id = 01ad15c432fbb19bbf585c5aeac77e7a30218948, deletion_comment = False)

michaelcfanning:
>             if (currentLog.Runs[0].Results.Any())

You are correct with #2. That is why all the validation logic starting at line 58 is only conditionally applied. i.e., in the event that our random file generator produces no results, there is no divergence, so none of that validation runs.

---
In reply to: [434062789](https://github.com/Microsoft/sarif-sdk/pull/1118#issuecomment-434062789) [](ancestors = 434062789,432425245,432424780)

---
Refers to: src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs:34 in 01ad15c. [](commit_id = 01ad15c432fbb19bbf585c5aeac77e7a30218948, deletion_comment = False)

lgolding:
>             if (currentLog.Runs[0].Results.Any())

But this is just bad. We have a unit test whose behavior is unpredictable. Sure, maybe that code always runs today, because the random seed is fixed, but on the day that a new version of .NET comes out that changes the random number generator, it will all end in tears.

It just cannot be right to have a test in which you _cannot predict_ whether some of its assertions will even be evaluated!

---
In reply to: [434062789](https://github.com/Microsoft/sarif-sdk/pull/1118#issuecomment-434062789) [](ancestors = 434062789,432425245,432424780)

---
Refers to: src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs:34 in 01ad15c. [](commit_id = 01ad15c432fbb19bbf585c5aeac77e7a30218948, deletion_comment = False)

src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
michaelcfanning: you're right in that baselineSharedPropertyValue is unused code that needs to be discarded. Done.

---
In reply to: [229080542](https://github.com/Microsoft/sarif-sdk/pull/1118#discussion_r229080542) [](ancestors = 229080542)

michaelcfanning:
>             if (currentLog.Runs[0].Results.Any())

That's the way this test was constructed. It produces random content by design. It emits the random seed in test output for reproducibility in case the randomness generates a previously unknown problem. In the event the random generator produces absolutely no results, the test needs to accommodate that. Sounds like you don't agree with this approach. I tend to agree it can be problematic. In my experience, this kind of test is a coin flip between producing entirely unwanted breaks in an unpredictable way vs. uncovering some new test value. 

But, this current change isn't about addressing any problems in the test pattern.

---
In reply to: [434070534](https://github.com/Microsoft/sarif-sdk/pull/1118#issuecomment-434070534) [](ancestors = 434070534,434062789,432425245,432424780)

---
Refers to: src/Sarif.UnitTests/Baseline/ResultMatching/SarifLogResultMatcherTests.cs:34 in 01ad15c. [](commit_id = 01ad15c432fbb19bbf585c5aeac77e7a30218948, deletion_comment = False)

PR #1121, opened 2018-10-29
Work to enable better serialization behaviors for empty arrays. Various schema fixes.

michaelcfanning:
This PR has three substantive changes:

1) Adds some test infrastructure that creates test content exhaustively directly from the object model.
2) Fixes two types in schema that did not include a property bag (revealed by test data above)
3) Provides better serialization behavior, whereby non-populated collections are not persisted to JSON. Without this functionality, it is easier for users to create JSON that doesn't actually validate. Ideally, default serialization would not put users in this condition.

@lgolding @rtaket


src/Sarif/CodeGenHints.json
```diff
@@ -479,16 +479,6 @@
       "kind": "DictionaryHint"
     }
   ],
-  "Result.HelpUri": [
```
lgolding: >HelpUri [](start = 10, length = 7)

Don't delete it -- this was a typo for the real property, `Rule.HelpUri`. Just change that and move it down into alphabetical order.

lgolding: Oh durr, you already did.

---
In reply to: [229077825](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229077825) [](ancestors = 229077825)

src/Sarif/CodeGenHints.json
```diff
@@ -685,7 +695,9 @@
       "arguments": {
         "dotNetPropertyName": "SchemaUri"
       }
-    },
+    }
+  ],
+  "SarifLog.SchemaUri": [
```
lgolding: >SchemaUri [](start = 12, length = 9)

Oh! Nice catch. So `PropertyNameHint` suffers from the same usage difficulty as `ClassNameHint`.

src/Sarif/ExtensionMethods.cs
```diff
@@ -16,6 +16,11 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        public static bool HasAtLeastOneNonNullValue(this IList<string> list)
+        {
+            return list != null && list.Where((s) => s != null).FirstOrDefault() != null;
```
lgolding: >list.Where((s) => s != null).FirstOrDefault() != null; [](start = 35, length = 54)

You can simplify this because `Any` can take a predicate:
```
return list != null && list.Any(s => s != null);
```

src/Sarif/Core/Run.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public partial class Run
+    {
+        public bool ShouldSerializeInvocations() { return this.Invocations != null && this.Invocations.Where((s) => s != null).FirstOrDefault() != null; }
```
lgolding: >this.Invocations.Where((s) => s != null).FirstOrDefault() != null [](start = 86, length = 65)

Again, use `Any` with predicate. Won't continue to comment on this issue.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -180,5 +180,18 @@ public void RemoveProperty(string propertyName)
...
         public TagsCollection Tags { get; }
+
+        public virtual bool ShouldSerializeProperties()
+        {
+            return PropertyBagHasAtLeastOneNonNullValue();
+        }   
+
+        public bool PropertyBagHasAtLeastOneNonNullValue()
+        {
+            bool hasAtLeastOneNonNullTag = this.Tags.Where(t => !string.IsNullOrEmpty(t)).Count() > 0;
```
lgolding: >. [](start = 52, length = 1)

Missing null check.

src/Sarif/Core/Run.cs
lgolding: You can also do this in places where the code checks for `Count() > 0`.

---
In reply to: [229091708](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229091708) [](ancestors = 229091708)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                { typeof(DateTime),(isRequired) => { return isRequired ? DateTime.UtcNow : new DateTime(); } },
+                { typeof(Uri),     (isRequired) => { return isRequired ? new Uri("https://required.uri.contoso.com") : null; } }
+            };
+        }
+
+        // These builders exhaustively populate the generated SARIF file, including non-required JSON primitives
+        public static IDictionary<Type, PrimitiveValueBuilder> PopulateAllPrimitives()
+        {
+            // The values in this output are designed to allow conformance to the schema. In our schema,.
+            // for example, some integer values must be > 0. And so, out optional integer value is
```
lgolding: >t [](start = 71, length = 1)

"out" => "our"

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+    /// </summary>
+    public class DefaultObjectPopulatingVisitor : SarifRewritingVisitor
+    {
+        public delegate object PrimitiveValueBuilder(bool isRequired);
+
+        // The default builders create a placeholder value for all required property primitives
+        // and emits default (null, 0, false, etc.) values for the remainder. As a result, a
+        // default hydration of this class should contain no non-default values for 
+        // JSON primitive-valued properties that aren't required
+        public static IDictionary<Type, PrimitiveValueBuilder> PopulateRequiredPrimitives()
```
lgolding: >PopulateRequiredPrimitives [](start = 63, length = 26)

This method and the next don't do what their names say they do. Consider: `GetRequiredPrimitiveBuilders` (which isn't great because it suggests that it's the builders that are required), or perhaps better, `GetBuildersForRequiredPrimitives`.

For the other one, perhaps `GetBuildersForAllPrimitives`.

src/Sarif/Core/PropertyBagHolder.cs
michaelcfanning: Tags is always non-null, check out the default ctor.

---
In reply to: [229092005](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229092005) [](ancestors = 229092005)

src/Sarif/ExtensionMethods.cs
michaelcfanning: like it, thanks.

---
In reply to: [229079586](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229079586) [](ancestors = 229079586)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            // The values in this output are designed to allow conformance to the schema. In our schema,.
+            // for example, some integer values must be > 0. And so, out optional integer value is 
+            // written as 1.
+            return new Dictionary<Type, PrimitiveValueBuilder>()
+            {
+                { typeof(bool),    (isRequired) => { return true; } },
+                { typeof(int),     (isRequired) => { return isRequired ? int.MaxValue : 1; } },
+                { typeof(double),  (isRequired) => { return isRequired ? double.MaxValue : 1; } },
+                { typeof(string),  (isRequired) => { return isRequired ? "[Required string value]" : "[Optional string value]"; } },
+                { typeof(DateTime),(isRequired) => { return isRequired ? DateTime.UtcNow : new DateTime(1776, 7, 4).ToUniversalTime(); } },
```
lgolding: >DateTime.UtcNow [](start = 73, length = 15)

This code will be used in unit tests, and`DateTime.UtcNow` isn't determinstic. Just use a fixed date, _e.g_ `new DateTime(2018, 10, 29).ToUniversalTime()`.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                { typeof(Uri),     (isRequired) => { return isRequired ? new Uri("https://required.uri.value.contoso.com") : new Uri("https://optional.uri.value.contoso.com"); } }
+            };
+        }
+
+        public DefaultObjectPopulatingVisitor(JsonSchema schema, IDictionary<Type, PrimitiveValueBuilder> primitiveValueBuilders = null)
```
lgolding: >primitiveValueBuilders [](start = 106, length = 22)

I see, so you can call this by passing the return value of one of the two above methods.

I don't think that's easy to understand. It unnecessarily exposes those two methods. For that matter, it unnecessarily exposes the delegate time `PrimitiveValueBuilder`. Consider instead using creation methods with revealing names. Make the ctor private, make the builder-dictionary-creating methods private, and make the delegate private. Then define:

```
public static DefaultObjectPopulatingVisitor CreateVisitorToPopulateRequiredPrimitivesOnly(JsonSchema schema);
public static DefaultObjectPopulatingVisitor CreateVisitorToPopulateAllPrimitives(JsonSchema schema);
```
... and the creation methods invoke the ctor with the appropriate set of builders.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+        {
+            _schema = schema;
+            _typeToPropertyValueConstructorMap = primitiveValueBuilders ?? PopulateRequiredPrimitives();
```
lgolding: >primitiveValueBuilders [](start = 49, length = 22)

This is another non-obvious feature of the current design. It won't be obvious to the reader that omitting the second argument results in populating only the required primitives. Using the "creation method" pattern makes this explicit.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        private JsonSchema _schema;
+        private IDictionary<Type, PrimitiveValueBuilder> _typeToPropertyValueConstructorMap;
```
lgolding: >private [](start = 8, length = 7)

Both of these fields can be `readonly`.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+
+        // There are two places in the format where object nest references to the defining
```
lgolding: "object" => "objects"?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        private JsonSchema _schema;
+        private IDictionary<Type, PrimitiveValueBuilder> _typeToPropertyValueConstructorMap;
+
+        // There are two places in the format where object nest references to the defining 
+        // type. 'exception.innerExceptions' is an array of exception objects. 
+        // 'node.children' is another collection of nodes. We need to watch for these
+        // scenarios to prevent re-entrancy and eventual stack consumption when 
+        // producing samples of these types.
```
lgolding: >. [](start = 43, length = 1)

Good comment.

src/Sarif/Core/PropertyBagHolder.cs
lgolding: ok.

---
In reply to: [229102696](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229102696) [](ancestors = 229102696,229092005)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            PopulateInstanceWithDefaultMemberValues(node);
+
+            return base.VisitActual(node);
+        }
+
+        // Retain nesting level for visiting exceptions to prevent
+        // unbounded re-entrance populating exception.innerExceptions
+        public override ExceptionData VisitExceptionData(ExceptionData node)
+        {
+            exceptionDataParentCount++;
```
lgolding: >exceptionDataParentCount [](start = 12, length = 24)

While you're counting, do you want to throw if the nesting level exceeds a maximum?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+
+        private void PopulateInstanceWithDefaultMemberValues(ISarifNode node)        {
+           
+            Type nodeType = node.GetType();
+            var binding = BindingFlags.Public | BindingFlags.Instance;
+            foreach (PropertyInfo property in nodeType.GetProperties(binding))
+            {
+                // The node kind is always properly set in the OM and
+                // isn't relevant to the SARIF schema itself
+                if (property.Name == "SarifNodeKind") { continue; }
```
lgolding: >SarifNodeKind [](start = 38, length = 13)

Huh. So the SDK introduces a _de facto_ reserved property that the spec doesn't mention. Do we want to do what we've discussed in the past, that is, reserving property names that start with "Sarif" (any case) to the implementation?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: will fix it next change

---
In reply to: [229109180](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229109180) [](ancestors = 229109180)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Will fix creation pattern in future change.

---
In reply to: [229108615](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229108615) [](ancestors = 229108615)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Nesting is impossible due to other code (that refuses to populate data members to recurse into).

---
In reply to: [229110609](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229110609) [](ancestors = 229110609)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Hm. Let's discuss offline.

---
In reply to: [229111818](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229111818) [](ancestors = 229111818)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+
+        // There are two places in the format where object nest references to the defining 
+        // type. 'exception.innerExceptions' is an array of exception objects. 
+        // 'node.children' is another collection of nodes. We need to watch for these
+        // scenarios to prevent re-entrancy and eventual stack consumption when 
+        // producing samples of these types.
+        private int graphNodeParentCount;
+        private int exceptionDataParentCount;
+
+        public override object VisitActual(ISarifNode node)
```
lgolding: Needs a big ol' comment to explain why you're overriding VisitActual and using reflection to access the properties, rather than using the usual VisitXxx overrides and the strongly typed OM.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
lgolding: This is not fixed in Iteration 3 (the latest iteration I have). Do you have more changes to push?

---
In reply to: [229106315](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229106315) [](ancestors = 229106315)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                { typeof(bool),    (isRequired) => { return true; } },
+                { typeof(int),     (isRequired) => { return isRequired ? int.MaxValue : 1; } },
+                { typeof(double),  (isRequired) => { return isRequired ? double.MaxValue : 1; } },
+                { typeof(string),  (isRequired) => { return isRequired ? "[Required string value]" : "[Optional string value]"; } },
+                { typeof(DateTime),(isRequired) => { return isRequired ? DateTime.UtcNow : new DateTime(1776, 7, 4).ToUniversalTime(); } },
+                { typeof(Uri),     (isRequired) => { return isRequired ? new Uri("https://required.uri.value.contoso.com") : new Uri("https://optional.uri.value.contoso.com"); } }
+            };
+        }
+
+        public DefaultObjectPopulatingVisitor(JsonSchema schema, IDictionary<Type, PrimitiveValueBuilder> primitiveValueBuilders = null)
```
lgolding: >schema [](start = 57, length = 6)

Note to self: It is unusual for a visitor to take the schema, since its generated code incorporates knowledge of the schema. But @michaelcfanning has told me that this visitor engages in some hijinks with the schema, so I will wait and see...

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                { typeof(string),  (isRequired) => { return isRequired ? "Required string value." : null; } },
+                { typeof(DateTime),(isRequired) => { return isRequired ? DateTime.UtcNow : new DateTime(); } },
+                { typeof(Uri),     (isRequired) => { return isRequired ? new Uri("https://required.uri.contoso.com") : null; } }
+            };
+        }
+
+        // These builders exhaustively populate the generated SARIF file, including non-required JSON primitives
+        public static IDictionary<Type, PrimitiveValueBuilder> GetBuildersForAllPrimitives()
+        {
+            // The values in this output are designed to allow conformance to the schema. In our schema,.
```
lgolding: >. [](start = 104, length = 1)

I'll bet you meant to put this period on Line 38 :-)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
lgolding: I'm wrong. This is not a member of the property bag (_e.g._, `physicalLocation.Properties["SarifNodeKind"]`). Rather, it is a first-class .NET property of the generated class (_e.g._, `physicalLocation.SarifNodeKind`). Forget I said anything.

---
In reply to: [229116171](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229116171) [](ancestors = 229116171,229111818)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            Type nodeType = node.GetType();
+            var binding = BindingFlags.Public | BindingFlags.Instance;
+            foreach (PropertyInfo property in nodeType.GetProperties(binding))
+            {
+                // The node kind is always properly set in the OM and
+                // isn't relevant to the SARIF schema itself
+                if (property.Name == "SarifNodeKind") { continue; }
+
+                // Property bags and tags are populated via special methods on 
+                // the class rather than direct access of properties.
```
lgolding: >. [](start = 68, length = 1)

Worth a comment: `PropertyNames` and `Tags` are inherited from `PropertyBagHolder`.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                propertyBagHolder.Tags.Remove(meaninglessValue);
+            }
+        }
+
+        private void PopulatePropertyWithDefaultValue(ISarifNode node, PropertyInfo property)
+        {
+            object propertyValue = null;
+            Type propertyType = property.PropertyType;
+
+            // If we can't set this property, it is not of interest
```
lgolding: >t [](start = 66, length = 1)

Micro-optimization: may as well do this first.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            // exception instance itself contains an array of exceptions
+            // (exception.innerExceptions). We don't want to hydrate 
+            // the innerExceptions of any innerExceptions, which results
+            // in re-entrancy that ends up consuming all stack space.
+            //
+            // Once we have populated a single exceptions.innerExceptions array, 
+            // we have accomplished all the testing we need in any case.
+            if (node.SarifNodeKind == SarifNodeKind.ExceptionData &&
+                property.Name == "InnerExceptions" &&
+                exceptionDataParentCount > 0) { return; }
```
lgolding: >0 [](start = 43, length = 1)

Yes, I now see what the override of `VisitExceptionData` did for you: it just set this flag.

Would it be better if it _where_ a flag, like `bool inExceptionData`?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+
+            // This special-casing is required to account for the fact that an 
+            // exception instance itself contains an array of exceptions
+            // (exception.innerExceptions). We don't want to hydrate 
+            // the innerExceptions of any innerExceptions, which results
+            // in re-entrancy that ends up consuming all stack space.
+            //
+            // Once we have populated a single exceptions.innerExceptions array, 
+            // we have accomplished all the testing we need in any case.
+            if (node.SarifNodeKind == SarifNodeKind.ExceptionData &&
```
lgolding: >if [](start = 12, length = 2)

May as well do these two tests first, as well (waiting to declare the variables `propertyValue` and `propertyType` until the point where they are used). As the code stands, I have these two variables stored on my mental stack, examining each line to see if they're needed yet.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+                property.Name == "InnerExceptions" &&
+                exceptionDataParentCount > 0) { return; }
+
+            // Sample approach applies for graph node.children
```
lgolding: >Sample [](start = 15, length = 6)

"Sample" => "Similar"?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            if (node.SarifNodeKind == SarifNodeKind.Node &&
+                property.Name == "Children" && 
+                graphNodeParentCount > 0) { return; }
+
+            bool isRequired = PropertyIsRequiredBySchema(node.GetType().Name, property.Name);
```
lgolding: >PropertyIsRequiredBySchema [](start = 30, length = 26)

If this is the only thing you need the schema for, you can get it by reflection by looking at the `IsRequired` member of the property's `DataMember` attribute.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
lgolding: Ok, now I know what it's for. It isn't actually needed; see comment below about reflecting on `DataMemberAttribute`.

---
In reply to: [229125407](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229125407) [](ancestors = 229125407)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+
+            if (_typeToPropertyValueConstructorMap.TryGetValue(propertyType, out PrimitiveValueBuilder propertyValueConstructor))
```
lgolding: >propertyValueConstructor [](start = 103, length = 24)

Nit: it isn't a constructor. How about `propertyValueBuilder`?

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            {
+                propertyValue = propertyValueConstructor(isRequired);
+            }
+            else if (HasParameterlessConstructor(property.PropertyType))
```
lgolding: >property.PropertyType [](start = 49, length = 21)

Nit: `property.PropertyType` => `propertyType`

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            // These constructed types are creatable via a parameterless constructor
+            // (which is what is ultimately invoked by Activator.CreateInstance).
+            var listType = typeof(Dictionary<,>);
+            var constructedType = listType.MakeGenericType(propertyType.GenericTypeArguments[0], propertyType.GenericTypeArguments[1]);
+            return Activator.CreateInstance(constructedType);
+        }
+
+        private static bool IsList(Type propertyType)
+        {
+            return propertyType.Name.StartsWith("IList");
```
lgolding: >StartsWith [](start = 37, length = 10)

Isn't there anything in the `Type` object that gives you "`IList`" exactly? This will match "`IListicle`". Similarly for `IDictionary`.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: At this point, it's all we need, as we don't permit another level of nesting. that might make sense for the graph object in future. but will change to bool for now, cleaner code.

---
In reply to: [229127945](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229127945) [](ancestors = 229127945)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Huh? How did autocorrect strike while I was typing this comment? ;)

---
In reply to: [229128421](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229128421) [](ancestors = 229128421)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Right. Left it this way assuming that other stuff might come in when we examine other schema-derived things, like minItems, that aren't covered by DataMember. We will see. Also, the schema is already populated and this saves some reflection calls. 

---
In reply to: [229129386](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229129386) [](ancestors = 229129386)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Well, sure, now that you mention it. :)  I

            return propertyType.IsGenericType && propertyType.GetGenericTypeDefinition() == typeof(IList<>);


---
In reply to: [229132248](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229132248) [](ancestors = 229132248)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -0,0 +1,293 @@
...
+            // If _schema.Definitions does not contain the jsonTypeName, we are operating
+            // against the root sarifLog schema, which is what's stored in _schema
+            JsonSchema propertySchema = _schema.Definitions.ContainsKey(jsonTypeName) ? _schema.Definitions[jsonTypeName] : _schema;
+
+            return propertySchema.Required != null && propertySchema.Required.Contains(jsonPropertyName);
+        }
+
+        private string GetJsonNameFor(string name)
+        {
+            return name.Substring(0, 1).ToLower() + name.Substring(1);
```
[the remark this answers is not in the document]
michaelcfanning: Geez, of all people, I should know better. :)

---
In reply to: [229132581](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229132581) [](ancestors = 229132581)

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
michaelcfanning: Yes, what you say is why the design is still open. We need to author a few more tests (or at least I do) before all the requirements for customizing the default value emit sink in. It appears most tests will start with some baseline builders and go from there. It is very likely the delegate will change, for example, it likely needs to accept a JsonSchema object, not just the isRequired bool (but YAGNI dictates I start with the more focused argument). I suspect the 'real' change to make is to stop picking up default args with the populating visitor, which will force everyone to go through/discover the default builders.

---
In reply to: [229114719](https://github.com/Microsoft/sarif-sdk/pull/1121#discussion_r229114719) [](ancestors = 229114719,229108615)

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -28,6 +30,71 @@ public SarifValidatorTests()
             _schema = SchemaReader.ReadSchema(schemaText, JsonSchemaFile);
         }
 
+        //[Fact]
+        internal void ValidateAllTheThings()
+        {
+            // First, we start with builders that only populate required properties that are backed by primitives.
+            IDictionary<Type, DefaultObjectPopulatingVisitor.PrimitiveValueBuilder> propertyValueBuilders =
+                DefaultObjectPopulatingVisitor.GetBuildersForAllPrimitives();
```
lgolding: >GetBuildersForAllPrimitives [](start = 47, length = 27)

Should this have been `ForRequiredPrimitives`? Either this or the comment seems wrong (I think the comment is wrong, given the test name).

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -28,6 +30,71 @@ public SarifValidatorTests()
             _schema = SchemaReader.ReadSchema(schemaText, JsonSchemaFile);
         }
 
+        //[Fact]
+        internal void ValidateAllTheThings()
+        {
+            // First, we start with builders that only populate required properties that are backed by primitives.
```
lgolding: >// [](start = 12, length = 2)

Comment is inconsistent with the name of the test and with the call to `GetBuildersForAllPrimitives`. Copy/paste error in comment?

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -28,6 +30,71 @@ public SarifValidatorTests()
...
+            // error JSON1015: at "runs[0].conversion.tool.fileVersion": The string "{Optional string value}" does not match the regular expression "[0-9]+(\.[0-9]+){3}".
+            // error JSON1015: at "runs[0].files.PlaceholderDictionaryKey.mimeType": The string "{Optional string value}" does not match the regular expression "[^/]+/.+".
+        }
+
+        [Fact]
+        public void ValidatesUriConversion()
+        {
+            // First, we start with builders that only populate required properties that are backed by primitives.
+            IDictionary<Type, DefaultObjectPopulatingVisitor.PrimitiveValueBuilder> propertyValueBuilders =
+                DefaultObjectPopulatingVisitor.GetBuildersForRequiredPrimitives();
```
lgolding: >GetBuildersForRequiredPrimitives [](start = 47, length = 32)

Bug? You want to validate _all_ URI-valued properties, but not all of them are required. (_e.g._, `tool.resourceLocation`, `rule.helpUri`, `result.hostedViewerUri`, ...)

PR #1122, opened 2018-10-29
Add result.hostedViewerUri to schema

michaelcfanning:
@lgolding, this looks to be the final blocking issue before shipping the 10-10 schema.

### DIFF
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -1307,8 +1307,14 @@
           }
         },
 
+        "hostedViewerUri": {
+          "description": "An absolute URI at which the result can be viewed",
+          "type": "string",
+          "format": "uri"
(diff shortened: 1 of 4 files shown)


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1307,8 +1307,14 @@
           }
         },
 
+        "hostedViewerUri": {
+          "description": "An absolute URI at which the result can be viewed",
```
lgolding: >d [](start = 74, length = 1)

Add a period.

michaelcfanning: Zapped 'em all.

---
In reply to: [229075914](https://github.com/Microsoft/sarif-sdk/pull/1122#discussion_r229075914) [](ancestors = 229075914)

PR #1125, opened 2018-10-30
Update v2

michaelcfanning:
This change allows users to pass a reference to a pre-release v2 SARIF file and have it transformed to current v2. As part of this, I dropped in a FileSystem reference to allow mocked unit testing and prepped other commands for this as well. The Convert command calls to underlying code that operates strictly against a stream and I didn't take the time to push any testing machinery there, we have not immediate additional testing in any case.

@lgolding @EasyRhinoMSFT

### DIFF
--- a/src/Sarif.Driver/FileHelpers.cs
+++ b/src/Sarif.Driver/FileHelpers.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
-using Microsoft.CodeAnalysis.Sarif.Readers;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
@@ -15,9 +13,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     /// </summary>
     static class FileHelpers
     {
-        public static T ReadSarifFile<T>(string filePath, IContractResolver contractResolver = null)
+        public static T ReadSarifFile<T>(IFileSystem fileSystem, string filePath, IContractResolver contractResolver = null)
         {
-            string logText = File.ReadAllText(filePath);
+            string logText = fileSystem.ReadAllText(filePath);
 
             var settings = new JsonSerializerSettings
             {
@@ -28,7 +26,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             return JsonConvert.DeserializeObject<T>(logText);
         }
 
-        public static void WriteSarifFile<T>(T sarifFile, string outputName, Formatting formatting = Formatting.None, IContractResolver contractResolver = null)
+        public static void WriteSarifFile<T>(IFileSystem fileSystem, T sarifFile, string outputName, Formatting formatting = Formatting.None, IContractResolver contractResolver = null)
         {
             var settings = new JsonSerializerSettings
             {
@@ -36,7 +34,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 Formatting = formatting
             };
 
-            File.WriteAllText(outputName, JsonConvert.SerializeObject(sarifFile, settings));
+            fileSystem.WriteAllText(outputName, JsonConvert.SerializeObject(sarifFile, settings));
         }
 
         public static HashSet<string> CreateTargetsSet(IEnumerable<string> targetSpecifiers, bool recurse)
--- a/src/Sarif.Driver/FileHelpers.cs
+++ b/src/Sarif.Driver/FileHelpers.cs
@@ -23,7 +23,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             };
 
 
-            return JsonConvert.DeserializeObject<T>(logText);
+            return JsonConvert.DeserializeObject<T>(logText, settings);
         }
 
         public static void WriteSarifFile<T>(IFileSystem fileSystem, T sarifFile, string outputName, Formatting formatting = Formatting.None, IContractResolver contractResolver = null)
--- a/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
+++ b/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Moq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Sarif.Multitool.UnitTests
+{
+    public class TransformCommandTests
+    {
+        // Regression test for Issue #1064, "ValidatorCommand fails when target file name contains a space"
+        [Fact]
+        public void TransformCommand_TransformsMinimalFile()
+        {
+            // A minimal valid log file.
+            string logFileContents =
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0-beta.2018-09-26"",
+  ""runs"": [
+    {
+      ""tool"": {
+        ""name"": ""TestTool""
+      },
+      ""results"": []
+    }
+  ]
+}";
+            // First, ensure that our test sample\ schema uri and SARIF version differs 
+            // from current. Otherwise we won't realize any value from this test
+            var sarifLog = JsonConvert.DeserializeObject<SarifLog>(logFileContents);
+            sarifLog.SchemaUri.Should().NotBe(SarifUtilities.SarifSchemaUri);
+            sarifLog.Version.Should().NotBe(SarifUtilities.SemanticVersion);
+
+
+            string logFilePath = @"c:\logs\mylog.sarif";
+            string transformedContents = null;
+
+            var mockFileSystem = new Mock<IFileSystem>();
+            mockFileSystem.Setup(x => x.ReadAllText(logFilePath)).Returns(logFileContents);
+            mockFileSystem.Setup(x => x.WriteAllText(logFilePath, It.IsAny<string>())).Callback<string, string>((path, contents) => { transformedContents = contents; });
+
+            var transformCommand = new TransformCommand(mockFileSystem.Object, testing: true);
+
+            var options = new TransformOptions
+            {
+                Inline = true,
+                Version = SarifVersion.Current,
+                InputFilePath = logFilePath
+            };
+
+            int returnCode = transformCommand.Run(options);            
+            returnCode.Should().Be(0);
+
+            // Finally, ensure that transformation corrected schema uri and SARIF version.
+            sarifLog = JsonConvert.DeserializeObject<SarifLog>(transformedContents);
+            sarifLog.SchemaUri.Should().Be(SarifUtilities.SarifSchemaUri);
+            sarifLog.Version.Should().Be(SarifVersion.Current);
+        }
+    }
+}
--- a/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
+++ b/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
@@ -19,7 +19,7 @@ namespace Sarif.Multitool.UnitTests
         public const string MinimalPrereleaseV2 =
     @"{
   ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
-  ""version"": ""2.0.0-beta.2018-09-26"",
+  ""version"": ""2.0.0-csd.2.beta.2018-09-26"",
   ""runs"": [
     {
       ""tool"": {
--- a/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
+++ b/src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.IO;
 using FluentAssertions;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Microsoft.CodeAnalysis.Sarif.VersionOne;
 using Moq;
 using Newtonsoft.Json;
 using Xunit;
@@ -13,13 +16,8 @@ namespace Sarif.Multitool.UnitTests
 {
     public class TransformCommandTests
     {
-        // Regression test for Issue #1064, "ValidatorCommand fails when target file name contains a space"
-        [Fact]
-        public void TransformCommand_TransformsMinimalFile()
-        {
-            // A minimal valid log file.
-            string logFileContents =
-@"{
+        public const string MinimalPrereleaseV2 =
+    @"{
   ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
   ""version"": ""2.0.0-beta.2018-09-26"",
   ""runs"": [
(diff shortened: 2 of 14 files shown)


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -33,23 +33,18 @@ public static string UpdateToCurrentVersion(string prereleaseSarifLog, bool forc
...
-                {
-                    modifiedLog |= ApplyCoreTransformations(sarifLog);
-                    break;
-                }
 
                 case "2.0.0-csd.2.beta.2018-10-10":
                 {
-                    modifiedLog |= ApplyCoreTransformations(sarifLog);
-                    modifiedLog |= ApplyChangesSinceTechnicalCommitteeMeeting25();
+                    // NOthing to do, this is current
```
lgolding: >O [](start = 24, length = 1)

"O" => "o"

src/Sarif.Driver/FileHelpers.cs
```diff
@@ -15,9 +13,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     /// </summary>
     static class FileHelpers
     {
-        public static T ReadSarifFile<T>(string filePath, IContractResolver contractResolver = null)
+        public static T ReadSarifFile<T>(IFileSystem fileSystem, string filePath, IContractResolver contractResolver = null)
```
lgolding: >fileSystem [](start = 53, length = 10)

Thank you!

In a future PR (or this one, if you're feeling ambitious), take a look at making `FileHelpers` and instance class, passing the `IFileSystem` in the ctor, and then you wouldn't have to pass it on each call.

src/Sarif.UnitTests/Readers/SampleModel/LogModel.cs
```diff
@@ -114,7 +114,7 @@ protected override JsonContract CreateContract(Type objectType)
 
     internal class LogModelSampleBuilder
     {
-        public const string SampleLogPath = "CodeCrawler.log.json";
+        public const string SampleLogPath = @"e:\repros\CodeCrawler.log.json";
```
lgolding: >repros [](start = 49, length = 6)

Is this going to be another one of those cases where the test won't work on a machine that doesn't have your drives and directory layout?

src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        // Regression test for Issue #1064, "ValidatorCommand fails when target file name contains a space"
+        [Fact]
+        public void TransformCommand_TransformsMinimalFile()
+        {
+            // A minimal valid log file.
+            string logFileContents =
+@"{
+  ""$schema"": ""http://json.schemastore.org/sarif-2.0.0"",
+  ""version"": ""2.0.0-beta.2018-09-26"",
```
lgolding: >- [](start = 22, length = 1)

"`2.0.0-csd.2.beta.2018-09-26`"?

src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Moq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Sarif.Multitool.UnitTests
+{
+    public class TransformCommandTests
+    {
+        // Regression test for Issue #1064, "ValidatorCommand fails when target file name contains a space"
```
lgolding: >space [](start = 101, length = 5)

But the target file name below, `c:\logs\mylog.sarif`, does not contain a space.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -11,9 +11,16 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
-    internal static class MergeCommand
+    internal class MergeCommand
     {
-        public static int Run(MergeOptions mergeOptions)
+        private IFileSystem _fileSystem;
```
lgolding: >private [](start = 8, length = 7)

`readonly`, here and in all the command classes. (See? I only said it once this time.)

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -44,7 +51,7 @@ public static int Run(MergeOptions mergeOptions)
 
                 Directory.CreateDirectory(mergeOptions.OutputFolderPath);
 
-                FileHelpers.WriteSarifFile(combinedLog, outputName, formatting);
+                FileHelpers.WriteSarifFile(_fileSystem, combinedLog, outputName, formatting);
```
lgolding: >WriteSarifFile [](start = 28, length = 14)

Oh, yeah, this reinforces my idea for `FileHelpers`. Having seen the usage pattern, I _do_ think you should do this:

1. Rename `FileHelpers` to `FileHelper` (both the class and the source file).
2. Make `FileHelper` an instance class, and pass `IFileSystem` to its ctor.
3. In the command class ctors, instantiate a `FileHelper` object and pass it the `IFileSystem`. Probably the command classes don't need to retain their own copy of the `IFileSystem` (certainly this particular class doesn't).

src/Sarif.Multitool/TransformOptions.cs
```diff
@@ -10,10 +10,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
     {
         [Option(
-            't',
+            'v',
             "target-version",
             HelpText = "The SARIF version to which the input file will be transformed.",
-            Default = 2)]
-        public int Version { get; internal set; }
+            Default = SarifVersion.Current)]
+        public SarifVersion Version { get; set; }
```
lgolding: >Version [](start = 28, length = 7)

There are two versions in play, the source and the target. And you helpfully named the command line option `--target-version`. So IMO the property should also be named `TargetVersion` (and then IMO the shortcut option should still be `-t`, to match).

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -2,21 +2,33 @@
...
+            _fileSystem = fileSystem ?? new FileSystem();
+            _testing = testing;
+        }
+
+        public int Run(TransformOptions transformOptions)
         {
             try
             {
-                if (transformOptions.Version < 1 || transformOptions.Version > 2)
+                if (transformOptions.Version != SarifVersion.OneZeroZero && transformOptions.Version != SarifVersion.Current)
```
lgolding: >Version [](start = 37, length = 7)

Got it, so the Transform command can either downgrade to v1, or upgrade to the latest v2. You can't transform to an intermediate version. Probably worth a comment.

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -58,5 +81,28 @@ public static int Run(TransformOptions transformOptions)
 
             return 0;
         }
+
+        private string SniffVersion(string sarifPath, int tokenCountLimit = int.MaxValue)
+        {
+            int tokenCount = 0;
+
+            // When we're unit-testing, we'll retrieve a string representation of the file contents and pass this to a stream reader instance.
+            TextReader textReader = _testing ? new StreamReader(new MemoryStream(Encoding.UTF8.GetBytes(_fileSystem.ReadAllText(sarifPath)))) : new StreamReader(sarifPath);
```
lgolding: >textReader [](start = 23, length = 10)

Long line; suggest breaking it:
```
TextReader textReader = _testing
    ? the-one-thing
    : the-other-thing;
```

lgolding: You don't need the `_testing` flag at all. The first branch of the ternary operator works whether you have a real or a fake file system:
```
TextReader textReader = new StreamReader(new MemoryStream(Encoding.UTF8.GetBytes(_fileSystem.ReadAllText(sarifPath))));
```
If you chose the conditional to avoid having to read a whole, potentially large, "real" log file into memory, that's worth a comment.

---
In reply to: [229510405](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229510405) [](ancestors = 229510405)

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -58,5 +81,28 @@ public static int Run(TransformOptions transformOptions)
...
+
+            using (JsonTextReader reader = new JsonTextReader(textReader))
+            {
+                while (reader.Read() && tokenCount < tokenCountLimit)
```
lgolding: >tokenCountLimit [](start = 53, length = 15)

I understand the sentiment, but I don't think we should limit it. Some implementations ignore the spec's advice to put `version` first. Consider the file that Aaron just sent us. It emits the properties in alphabetical order, so `version` comes last. If such an implementation produces a large file, we'll never get to `version`.

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -33,21 +45,32 @@ public static int Run(TransformOptions transformOptions)
...
+                if (transformOptions.Version == SarifVersion.Current)
                 {
-                    SarifLogVersionOne actualLog = FileHelpers.ReadSarifFile<SarifLogVersionOne>(transformOptions.InputFilePath, SarifContractResolverVersionOne.Instance);
-                    var visitor = new SarifVersionOneToCurrentVisitor();
-                    visitor.VisitSarifLogVersionOne(actualLog);
+                    string inputFilePath = transformOptions.InputFilePath;
+                    string inputVersion = SniffVersion(inputFilePath);
 
-                    FileHelpers.WriteSarifFile(visitor.SarifLog, fileName, formatting);
+                    if (inputFilePath == null || inputFilePath == "1.0.0")
```
lgolding: >inputFilePath [](start = 49, length = 13)

Bug: `inputFilePath` => `inputVersion`

lgolding: I don't understand the null check, either.

---
In reply to: [229513358](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229513358) [](ancestors = 229513358)

src/Sarif.Multitool/TransformCommand.cs
michaelcfanning: In Aaron's case, his file has fewer than  2,147,483,647 JSON tokens and so we do end up finding his version. :) this configurability isn't exposed anywhere currently, so i could just disable it entirely for now.

---
In reply to: [229512144](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229512144) [](ancestors = 229512144)

src/Sarif.Multitool/TransformOptions.cs
michaelcfanning: -v is easy to remember as the 'version'. -t is not. another idea here is to forbid a short version altogether. or we could prefer -v and --version

---
In reply to: [229509857](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229509857) [](ancestors = 229509857)

src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
michaelcfanning: comment left from copying another test class to use as basis for this one.

---
In reply to: [229507635](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229507635) [](ancestors = 229507635)

src/Sarif.Multitool/TransformCommand.cs
lgolding: Yes, I think we should. My point wasn't that Aaron's file was too large; it was that since he emitted `version` last, his implementation was _in danger of_ creating a file that you can't transform.

---
In reply to: [229513592](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229513592) [](ancestors = 229513592,229512144)

src/Sarif.Multitool/MergeCommand.cs
michaelcfanning: i make this change only because you care. i honestly don't see how applying 'readonly' assists in more maintainable code. it just comes across as busy work. can you explain the substantive issue we are protecting against? nice job only making a single comment, btw. :)

---
In reply to: [229508070](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229508070) [](ancestors = 229508070)

src/Sarif.Multitool/TransformOptions.cs
lgolding: I'd accept `-v` as the short option name but I feel pretty strongly from a code clarity standpoint that the property should be named `TargetVersion`.

---
In reply to: [229513921](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229513921) [](ancestors = 229513921,229509857)

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -33,21 +45,32 @@ public static int Run(TransformOptions transformOptions)
...
+                    {
+                        SarifLogVersionOne actualLog = FileHelpers.ReadSarifFile<SarifLogVersionOne>(_fileSystem, transformOptions.InputFilePath, SarifContractResolverVersionOne.Instance);
+                        var visitor = new SarifVersionOneToCurrentVisitor();
+                        visitor.VisitSarifLogVersionOne(actualLog);
+                        FileHelpers.WriteSarifFile(_fileSystem, visitor.SarifLog, fileName, formatting);
+                    }
+                    else
+                    {
+                        // We have a pre-release v2 file that we should upgrade to current. 
+                        string sarifText = PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(_fileSystem.ReadAllText(inputFilePath), formatting: formatting);
```
lgolding: >UpdateToCurrentVersion [](start = 78, length = 22)

Hmm. Elsewhere in this review I made a comment to the effect that perhaps you didn't want to read an entire potentially large file into memory. But we are doing it here.

src/Sarif.Multitool/MergeCommand.cs
michaelcfanning: I am not going to take your suggestion at this time. The reason is that I am not sure that file helpers should survive at all. it is less helpful now that we've eliminated the need for the contract resolver. it is redundant with some other utility code and extension methods. Also, I'd like to see how the Convert command can/should be changed to improve file access to enable testing. all of this is more work than I can invest. :) I made some improvements of testability and am drawing the line here. I will look further into improving the situation next time the hood is up.

---
In reply to: [229509097](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229509097) [](ancestors = 229509097,229508810)

src/Sarif.Multitool/TransformOptions.cs
michaelcfanning: I strongly agree with you. to be clear, you're ok with 'v' and --target-version and TargetVersion?

---
In reply to: [229514757](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229514757) [](ancestors = 229514757,229513921,229509857)

lgolding: Yes.

---
In reply to: [229515780](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229515780) [](ancestors = 229515780,229514757,229513921,229509857)

src/Sarif.Multitool/TransformCommand.cs
michaelcfanning: oh gee. you make me realize, we should update this code to transform pre-release v2 to current v2 before dropping down to v2. :) this allows ongoing transformation of pre-release v2 to v1. this is what commenting code does for you, makes you think a bit more. :)

---
In reply to: [229510132](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229510132) [](ancestors = 229510132)

lgolding: "pre-release v2 to current v2 to v1", I assume you mean. It's up to you, if you think it's a real scenario. Like, some MS internal tool is generating pre-release v2, but wants to store it in an issue store that still accepts only v1.

---
In reply to: [229516417](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229516417) [](ancestors = 229516417,229510132)

src/Sarif.Multitool/TransformCommand.cs
michaelcfanning: ouch. mile wide unit test hole, will close. thankfully, i created a nice pattern for doing so. :)

---
In reply to: [229513479](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229513479) [](ancestors = 229513479,229513358)

src/Sarif.Multitool/MergeCommand.cs
lgolding: In reply to a comment of yours that came across in email, but which I think you've since edited out: I make this a habit to document to the maintainer that the value should not be changed, and to prevent the author from accidentally changing it. It costs me nothing to add the keyword.

michaelcfanning: I did edit that out, because on reflection i also see the value. this case is similar to another standard i like to enforce, that methods that fail to access the 'this' pointer are declared static. it is helpful to be explicit and precise even in these small things. i find 'readonly' a thorn because I don't have an internal habit to think about it. i certainly can build that habit, though. resolved!

---
In reply to: [229518035](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229518035) [](ancestors = 229518035)

src/Sarif.Multitool/TransformCommand.cs
michaelcfanning: Right! This is a limit of JObject.Parse, which operates against a sting representation. I didn't want to exacerbate the situation further. The prerelease transformation is intended to wither away and it is already sufficiently burdensome to maintain this code. And so, performance is a lower priority. The goal, of course, is to finalize v2 and dispense with the rewrite altogether. In the future, maybe someone can go back and rewrite the transformation to be more performant. 

---
In reply to: [229514883](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229514883) [](ancestors = 229514883)

src/Sarif.Multitool/TransformCommand.cs
michaelcfanning: right, here's the scenario: i've integrated a pre-release v2-generating tool to my pipeline but need to drop it to v1 for some reason. i

---
In reply to: [229516755](https://github.com/Microsoft/sarif-sdk/pull/1125#discussion_r229516755) [](ancestors = 229516755,229516417,229510132)

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -2,21 +2,33 @@
...
+using Microsoft.CodeAnalysis.Sarif.Writers;
 using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
-    internal static class TransformCommand
+    internal class TransformCommand
     {
-        public static int Run(TransformOptions transformOptions)
+        private bool _testing;
```
lgolding: >bool [](start = 16, length = 4)

`readonly`

Issue #1128, opened 2018-10-31
Fortify FPR converter doesn't handle certain Vulnerabilities properly


EasyRhinoMSFT:
Vulnerabilities where the <Node> element doesn't have a label attribute AND doesn't have an <Action> element leads to an exception.
1. Handle this circumstance to fix the exception
2. Since we won't have a message.text string, don't create the message object on threadFlowLocation.location

Issue #1127, opened 2018-10-31
Fortify FPR converter crashes on Vulnerabilities with multiple traces


EasyRhinoMSFT:
In ParseLocationFromTrace, need to check _resultToSnippetIdDictionary for existing key before adding the result.

PR #1130, opened 2018-10-31
Fix Fortify FPR converter bugs #1127 and #1128

EasyRhinoMSFT:
This change includes a refactor that puts each <Trace> block into its own codeFlow, rather than a single codeFlow with a threadFlow for each <Trace>. I think this is the right way to translate.


EasyRhinoMSFT:
I don't know what's wrong with AppVeyor, everything passes in both BuildAndTest and Visual Studio.

EasyRhinoMSFT:
Yes, I'm going to run Fortify against Sarif.Sdk and the Viewer codebases. I've hit a problem that causes the scan to fail and I'm working with Prachi to figure it out. Issue #1131

PR #1133, opened 2018-11-07
Properties merge

michaelcfanning:
* Deleted an 'Includes' extension to various enums that effectively duplicates the Enum.HasFlag method.
* Adjust Result equality comparer to only compare correlation guids if that property is non-null in both results being compared
* Define an enum to govern dictionary merge behaviors (currently restricted to preserving the oldest or the newest dictionary encountered)
* Modify 'exact' result matcher to make it configurable as to whether property bags are relevant in comparison
* Preserve files dictionary in baseline
* Improve serialization so that it does not emit empty arrays/default object instances
* Update C# language to 7.3 (note: I though I needed this update to get access to a new kind of generics constraint, but in the event this wasn't required. Still, I didn't revert the change).

NOTE: result matching is complex. Some kinds of information have affinity with a baseline, for example, a correlation guid that has already been assigned. On the other hand, some information has a strong affinity for the most recently computed version, for example, the hash of a file (that may have changed since the baseline). In general, it's hard to know how to handle property bags. Are they essential information which should be examined when considering object instance identity? Or are they non-essential details which should be disregarded?

This change is a preliminary start for considering these issues. These changes are sufficient to support the Keros team in their scenarios, which is the immediate goal. A more thoughtful analysis and corresponding API work should occur afterwards (in context of the next schema updated, most likely).


src/Sarif/NotYetAutoGenerated/ResultEqualityComparer.cs
```diff
@@ -0,0 +1,596 @@
...
+                    }
+                }
+            }
+
+            if (left.InstanceGuid != right.InstanceGuid)
+            {
+                return false;
+            }
+
+            // We only insist that a correlation guid matches a non-null value. The reason is that
```
michaelcfanning: >//  [](start = 12, length = 3)

This is the relevant comment/change to review.

src/Sarif/NotYetAutoGenerated/ResultEqualityComparer.cs
```diff
@@ -0,0 +1,596 @@
...
+                    }
+
+                    if (!object.Equals(value_6.Value, value_7))
+                    {
+                        return false;
+                    }
+                }
+            }
+
+            // If we find a null correlation guid on either side, we will overwrite that null
```
michaelcfanning: >  // [](start = 10, length = 4)

please note and comment on this behavior, which is non-obvious. if we compare two results, one of which has a correlation guid and the other which does not, we will flow the non-null correlation guid to the other result instance where it is null. the reason is that we've determined these two results are identical, and therefore the correlation guid relates to both.

PR #1142, opened 2018-11-21
Fix #1138: Add validation rule: contextRegion requires region

### LINKED ISSUE #1138: Validation rule: contextRegion requires region
The spec says:

> If a `physicalLocation` object contains a region property (§3.24.4), it **MAY** also contain a property named `contextRegion` whose value is a `region` object (§3.25) which specifies a region that is a proper superset of the region specified by the `region` property. If `region` is absent, `contextRegion` **SHALL** be absent.

Check for `region` if `contextRegion` is present.

This is an error-level diagnostic.

lgolding:
Also:
- Enhance the "new-style" verification so that we no longer require the file "Invalid_Expected.sarif". Each file can now contain a property that specifies the expected locations of all the validation results.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:
@vScottLouvau is added to the review. #Closed

src/Sarif.Multitool/Rules/ContextRegionRequiresRegion.cs
```diff
@@ -0,0 +1,37 @@
...
+        public override string Id => RuleId.ContextRegionRequiresRegion;
+
+        protected override IEnumerable<string> MessageResourceNames => new string[]
+        {
+            nameof(RuleResources.SARIF1016_Default)
+        };
+
+        protected override void Analyze(PhysicalLocation physicalLocation, string physicalLocationPointer)
+        {
+            if (physicalLocation.ContextRegion != null && physicalLocation.Region == null)
```
michaelcfanning: >physicalLocation.ContextRegion [](start = 16, length = 30)

What if someone provides an entirely empty region? A region has no required items. What if someone populated the context region with line/other information, but the Region object only includes a 'Message' string? Etc.

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif
```diff
@@ -0,0 +1,61 @@
...
+      ],
+      "resources": {
+        "rules": {
+          "TST0001": {
+            "id": "TST0001",
+            "helpUri": "ht%tp://www.example.com/rules/tst0001.html"
+          }
+        }
+      },
+      "properties": {
```
michaelcfanning: >properties [](start = 7, length = 10)

Ok. I still have concerns about the fragility of this mechanism (for example, the need to change the index). The implied path changes seem trivial to update, however. And I get that you are providing data that points to to primitive values. I have had issues with this sort of approach in the past where I have attempted to constrain the reporting to a single check. Sometimes it is impossible (and perhaps not even desirable) to limit things in this way. In which chase, it is helpful to be able to comprehensively describe the rule ids that are fired for certain patterns.

src/Sarif.Multitool/Rules/RuleId.cs
```diff
@@ -16,5 +16,6 @@ public static class RuleId
         public const string EndColumnMustNotBeLessThanStartColumn = "SARIF1013";
         public const string UriBaseIdRequiresRelativeUri = "SARIF1014";
         public const string UriMustBeAbsolute = "SARIF1015";
+        public const string ContextRegionRequiresRegion = "SARIF1016";
```
michaelcfanning: >SARIF1016 [](start = 59, length = 9)

I have a funky idea for you which is to at least consider the possibility that the relevant SARIF spec section comprise the rule id. So SARIF.3.24.5 for this case (the analysis that addresses subtleties in the physicalLocation.contextRegion object. NOW: of course, the spec is malleable at this point in time. so this may simply be a lousy idea. Or, we might proceed with writing validators until the spec is finalized, then do a scrub to convert to final section numbers. Don't forget, we have the hierarchical string to allow some flexibility as well in creating sub-checks that roll up into a master rule id. SARIF.3.24/5, for example, would point to a master rule, dedicated to validating the physicalLocation object, the sub-rule of which validates the plc.contextRegion is handled properly. Something to ponder.

src/Sarif.Multitool/Rules/ContextRegionRequiresRegion.cs
lgolding: The schema doesn't declare any required properties, but a `region` object with no properties is invalid, because:

- §3.25.5: When a `region` object represents a text region specified by line/column properties, it **SHALL** contain a property named `startLine`...
- §3.25.9: When a `region` object represents a text region specified by offset/length properties, it **SHALL** contain a property named `charOffset`...
- §3.25.11: When a `region` object represents a binary region, it **SHALL** contain a property named `byteOffset`...

I filed #1143: "Add validation rule: at least one of the region "start" properties must be present."

---
In reply to: [238499344](https://github.com/Microsoft/sarif-sdk/pull/1142#discussion_r238499344) [](ancestors = 238499344)

michaelcfanning: You should satisfy that work item by updating the schema itself, rather than authoring a validator rule. An 'anyOf' rule should cover this, e.g.:

  "anyOf": [
    {"required" : ["startLine"]},
    {"required" : ["charOffset"]},
    {"required" : ["byteOffset"]}
  ]

IOW, we should favor schema-driven solutions wherever possible. Of course, this may prompt us to author improved validation within jschema itself...

---
In reply to: [238504114](https://github.com/Microsoft/sarif-sdk/pull/1142#discussion_r238504114) [](ancestors = 238504114,238499344)

src/Sarif.Multitool/Rules/RuleId.cs
lgolding: I worry about those ids breaking in the next version of the spec. Even an editorial section reordering would break them, let alone new content ("3.25 metrics object").

---
In reply to: [238501875](https://github.com/Microsoft/sarif-sdk/pull/1142#discussion_r238501875) [](ancestors = 238501875)

PR #1147, opened 2018-12-05
Mark result.message as required in the schema

michaelcfanning:
@EasyRhinoMSFT @rtaket @cfaucon @lgolding


src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError.sarif
```diff
@@ -8,6 +8,7 @@
       },
       "results": [
         {
+          "message": { "text": "Some testing occurrred." },
```
lgolding: >rrr [](start = 49, length = 3)

"rrr" => "rr"

PR #1149, opened 2018-12-06
Upgrade more validator functional tests

lgolding:
We apply the new functional test pattern to four more rules:
- `EndColumnMustNotBeLessThanStartColumn`
- `EndLineMustNotBeLessThanStartLine`
- `EndTimeMustBeAfterStartTime` (which is misnamed, and in a future PR we will rename it to `EndTimeMustNotBeBeforeStartTime`)
- `MessageShouldEndWithPeriod`

In addition, we remove the test data for a rule that no longer exists, `HashAlgorithmsMustBeUnique` (which no longer applies because `file.hashes` is now an object keyed by the hash algorithm).

Because there are so many properties of type `message`, exhaustively testing the rule `MessageShouldEndWithPeriod` revealed many holes in the visitor class `SarifValidationSkimmerBase`, which I plugged. As we have discussed, we should generate this class from the schema.

After this, there are only two more rules to convert:
- `UriBaseIdRequiresRelativeUri`
- `UseAbsolutePathsForNestedFileUriFragments`

... but this PR is already large enough.


src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -219,6 +227,33 @@ private void Visit(Attachment attachment, string attachmentPointer)
...
             }
+
+            if (attachment.Description != null)
+            {
+                Visit(attachment.Description, attachmentPointer.AtProperty(SarifPropertyName.Description));
+            }
+
+            if (attachment.Regions != null)
+            {
+                Region[] regions = attachment.Regions.ToArray();
```
michaelcfanning: >egions.ToArray() [](start = 47, length = 16)

Why are you calling ToArray here and elsewhere? These collections are already indexable, aren't they? My concern is that you might be unnecessarily allocating memory. Can you either confirm that these calls aren't allocating or remove them in preference of indexing directly into the IList<Rectangle>? Comment applies here and elsewhere of course.

lgolding: I didn't know you could [index a list](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.item?view=netframework-4.7.2). I'll change it. But isn't it `O(n^2)` in time to iterate over a list that way? Or does .NET optimize storage of small lists to allow efficient random access?

---
In reply to: [239513818](https://github.com/Microsoft/sarif-sdk/pull/1149#discussion_r239513818) [](ancestors = 239513818)

lgolding: TIL, thanks Michael.

---
In reply to: [239524038](https://github.com/Microsoft/sarif-sdk/pull/1149#discussion_r239524038) [](ancestors = 239524038,239513818)

michaelcfanning: btw - i did eventually locate the Linq ToArray source and can confirm this code allocates. So we'll save memory w/the change. 

---
In reply to: [239528433](https://github.com/Microsoft/sarif-sdk/pull/1149#discussion_r239528433) [](ancestors = 239528433,239524038,239513818)

michaelcfanning:
LGTM after converting to IList for collection access (will require updating .Length access to IList.Count)

PR #1158, opened 2018-12-06
Remove 'open' from list of valid rule configuration default values.

michaelcfanning:
@EasyRhinoMSFT @lgolding @cfaucon


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -116,12 +116,45 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+
+            JObject rules = (JObject)resources["rules"];
+            if (rules == null ) { return modifiedResources; }
+
+            foreach (JProperty rule in rules.Values<JProperty>())
+            {
+                JObject configuration = (JObject)rule.Value["configuration"];
+                if (configuration == null) { continue; }
+
+                if ("open".Equals((string)configuration["defaultLevel"]))
```
lgolding: >Equals [](start = 27, length = 6)

Ordinal comparison

PR #1159, opened 2018-12-06
Finish validator functional test upgrade

lgolding:
We upgrade the last validator rule (`UriBaseIdRequiresRelativeUri`) to the new test design. Then we remove support for the old design from `ValidationSkimmerTestBase`.

Also:
- As promised in the last PR, we rename the rule `EndTimeMustBeAfterStartTime` to `EndTimeMustNotBeBeforeStartTime`.
- We remove the obsolete rule `UseAbsolutePathsForNestedUriFragments`. The spec no longer permits a URI to directly specify a nested file (for example, by writing `file:///C:/example.zip#/src/file.c`).


src/Sarif.Multitool/Rules/RuleId.cs
```diff
@@ -6,10 +6,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
     public static class RuleId
     {
         public const string DoNotUseFriendlyNameAsRuleId = "SARIF1001";
-        public const string UseAbsolutePathsForNestedFileUriFragments = "SARIF1002";
         public const string UrisMustBeValid = "SARIF1003";
         public const string HashAlgorithmsMustBeUnique = "SARIF1006";
-        public const string EndTimeMustBeAfterStartTime = "SARIF1007";
+        public const string EndTimeMustNotBeBeforeStartTime = "SARIF1007";
```
michaelcfanning: >SARIF1007 [](start = 63, length = 9)

This rule id should be SARIF.3.17.7

:)

lgolding: I will make those changes all at once after the spec stabilizes.

---
In reply to: [239979221](https://github.com/Microsoft/sarif-sdk/pull/1159#discussion_r239979221) [](ancestors = 239979221)

michaelcfanning: Ok! But, just curious, will you think it's a good idea when you do that. Don't forget! If you fail to be persuaded an idea is valuable, you should feel free to delegate the work to someone who is convinced of that. I myself have invoked this principle in some code review contexts, as you know. ;)

Issue #57, opened 2016-01-11
We need a telemetry plan for a right-click/up or down vote issue features


michaelcfanning:
We want users to have an easy way to send telemetry that says, minimally, this is a good result vs. tool noise.

PR #1160, opened 2018-12-07
Emit column kind default explicitly for Windows SDK SARIF emit.

michaelcfanning:
@rtaket @ScottLouvau @lgolding 

This change is unusual. run.columnKind is a property that specifies the underlying construct that is used to derive column information in a region. For Windows, these are UTF16 code units. For *nix, these tend to be unicode code points. After discussion, we believe that unicode code points are the most reasonable default value for this property, and that change was approved in the spec.

This SDK change addresses two problems: 1) ensure that SARIF persisted by this SDK forces the appropriate Windows-specific column kind into new log files, 2) when processing existing SARIF, the SDK will assume those files were also previously produced by the Windows platform and will populate column kind explicitly with UTF16 code units if that property was previously absent from SARIF.

In the (unexpected) circumstance that this SDK processes pre-release SARIF produced on a platform where unicode code points were used to compute regions, the producer will need to provide that enum value explicitly, rather than depending on the property's default value.


src/Sarif/CodeGenHints.json
```diff
@@ -656,6 +656,14 @@
           "unicodeCodePoints"
         ]
       }
+    },
+    {
+      "kind": "AttributeHint",
```
michaelcfanning: >AttributeHint [](start = 15, length = 13)

Oops. test gap closed here by adding columnKind explicitly to our comprehensive test files.

src/Sarif/Autogenerated/Run.cs
```diff
@@ -138,6 +139,8 @@ public SarifNodeKind SarifNodeKind
         /// Specifies the unit in which the tool measures columns.
         /// </summary>
         [DataMember(Name = "columnKind", IsRequired = false, EmitDefaultValue = false)]
+        [JsonConverter(typeof(EnumConverter))]
+        [System.ComponentModel.DefaultValue(ColumnKind.UnicodeCodePoints)]
```
michaelcfanning: Jschema doesn't emit default values currently. We'll have to manually maintain them for now. [https://github.com/Microsoft/jschema/issues/57](https://github.com/Microsoft/jschema/issues/57)

src/Sarif/Autogenerated/Run.cs
```diff
@@ -151,6 +154,7 @@ public SarifNodeKind SarifNodeKind
         /// </summary>
         public Run()
         {
+            this.ColumnKind = ColumnKind.UnicodeCodePoints;
```
michaelcfanning: >this.ColumnKind = ColumnKind.UnicodeCodePoints [](start = 12, length = 46)

ok, this is an important apparent JSON.NET gap. in JSON.NET, you can provide an argument in a JSON convert settings object that specifies that default values should *populate* if missing on deserialization. the problem with that is that you need to explicitly create and pass the appropriate settings object (and we've attempted to make everything work with a single API call with no parameterization of this kind). 

to work around the problem, we provide a programmatic equivalent of populate by setting the default value in the ctor. when JSON.NET itself is deserialization, it will overwrite this property in cases where the JSON has an explicit value.  a fix to jschema #57 should provide this fix as well (and not just the default value attribute that elides property values from persisted JSON if they map to the default).

src/Sarif/Readers/EnumConverter.cs
```diff
@@ -21,7 +21,6 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             {
                 throw new ArgumentNullException(nameof(reader));
             }
-
             string value = (string)reader.Value;
```
lgolding: >s [](start = 12, length = 1)

Odd change. Doesn't our style require a blank line after a closing brace? See lines 34, 39, 53...

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +13,24 @@ public partial class Run
         private static Invocation EmptyInvocation = new Invocation();
         private static LogicalLocation EmptyLogicalLocation = new LogicalLocation();
 
+        public bool ShouldSerializeColumnKind()
```
lgolding: >ShouldSerializeColumnKind [](start = 20, length = 25)

Wow, I hadn't looked closely at this until now. This is the first time I wondered who was calling these, and I see that JSON.NET has a built-in mechanism for [conditional serialization](https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm).

lgolding:
Ship it. I am still studying your comment about the "JSON.NET gap", but I don't see any reason to block this merge.

lgolding:
If today

lgolding:
You're feeling fine

lgolding:
Please review

lgolding:
My #1159.

lgolding:
Burma Shave.

src/Sarif/Autogenerated/Run.cs
lgolding: I assume you are talking about [JsonSerializerSettings.DefaultValueHandling](https://www.newtonsoft.com/json/help/html/SerializationSettings.htm#DefaultValueHandling), whose `Include` value means that "The Json.NET deserializer will continue setting a field/property if the JSON value is the same as the default value."

But it goes on to say "DefaultValueHandling can also be customized on individual properties with JsonPropertyAttribute." So I think we can fix this by emitting
```
[JsonProperty DefaultValueHandling = DefaultValueHandling.Include)]
```
on every property.

---
In reply to: [239879240](https://github.com/Microsoft/sarif-sdk/pull/1160#discussion_r239879240) [](ancestors = 239879240)

lgolding: I added that to [https://github.com/Microsoft/jchema/issues/57](https://github.com/Microsoft/jchema/issues/57).

---
In reply to: [239910615](https://github.com/Microsoft/sarif-sdk/pull/1160#discussion_r239910615) [](ancestors = 239910615,239879240)

PR #1161, opened 2018-12-07
Fortify FPR converter improvements + unit tests

EasyRhinoMSFT:


src/Samples/SarifDeferredSample/Program.cs
```diff
@@ -27,7 +27,6 @@ private static void Main(string[] args)
             {
                 // Use 'SarifDeferredContractResolver' and 'JsonPositionedTextReader' to load a geferred version of the same object graph.
                 JsonSerializer serializer = new JsonSerializer();
-                serializer.ContractResolver = (deferred ? SarifDeferredContractResolver.Instance : SarifContractResolver.Instance);
```
michaelcfanning: >SarifDeferredContractResolver [](start = 58, length = 29)

I don't think you can delete this, otherwise the deferred contract resolver won't get invoked. What you need to do, I think, is only set ContractResolver for the deferred case (and not set it to SarifContractResolver in the other case).

src/Sarif.UnitTests/Core/StackTests.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public class StackTests
     {
-        [Fact]
+        [Fact(Skip ="Broken by new, unexpected framework behavior in StackTrace class")]
```
michaelcfanning: >Skip [](start = 14, length = 4)

Where's the issue on this? We should be referencing it in a code comment for any disabled test.

michaelcfanning:
Good stuff. Please make sure you didn't break the deferred contract resolver, then merge.

PR #1162, opened 2018-12-08
Remove 'package' as a documented logical location kind in the schema.…

michaelcfanning:
… Documentation only change.

@lgolding @EasyRhinoMSFT


lgolding:
I suppose this isn't strictly a breaking change. or even a change at all, since the string was always open-ended. So I guess you don't need to mention it in the history.

michaelcfanning:
Exactly, My thoughts exactly. Everyone is still free to emit 'package' in this open-ended string value. We simply no longer encourage people to do so for these container objects.

PR #1167, opened 2018-12-19
Add result.rank and ruleConfiguration.defaultRank

lgolding:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1306,6 +1306,13 @@
           ]
         },
 
+        "rank": {
+          "description": "A number representing the priority or importance of the result.",
```
michaelcfanning: >or impo [](start = 61, length = 7)

why do you need this? are you comfortable deleting 'or importance'? importance is a slippery slope that pulls in the severity of vulns, the level of exposure, etc. rank is a relative ordering. i think 'priority' sufficiently communicates what we're after.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1438,11 +1445,19 @@
         },
 
         "defaultLevel": {
-          "description": "Specifies the default severity level of the result.",
+          "description": "Specifies the default severity level for results generated by this rule.",
           "default": "warning",
           "enum": [ "note", "warning", "error" ]
         },
 
+        "defaultRank": {
```
michaelcfanning: >defaultRank [](start = 9, length = 11)

my apologies. I have, yet again, completely misunderstood your proposal in context of the TC. I never thought we were applying a defaultRank to a rule. This doesn't make sense: ranks are generally computed values that are applied to all results. Ranks are generally available or not. 

when you discussed a default rank, I thought you are asking 'what is the default value of result.rank' if it isn't explicit? in this case, i have argued that the default s/be 0. so what I expected in this case was that we would not have a new ruleConfiguration property but we would have another explanatory sentence in the spec.

lgolding:
@michaelcfanning As we discussed offline with @fishoak, the design is good as it stands. The only change is that the default should be -1. I filed oasis-tcs/sarif-spec#303 for that, and put it on the [agenda](https://github.com/oasis-tcs/sarif-spec/blob/master/Agendas/20190109.md) for TC <span>#</span>30.

Issue #1169, opened 2018-12-21
PREfast converter should Trim() message lines


michaelcfanning:
Lots of unexpected leading and trailing newlines are persisted to output.

PR #1170, opened 2018-12-21
Logical locations dictionaries to arrays

michaelcfanning:


src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -239,22 +239,28 @@ private Rule GetRuleFromResources(Result result, IDictionary<string, Rule> rules
...
             properties = properties ?? new Dictionary<string, SerializedPropertyInfo>();
 
             List<Result> newRunResults = new List<Result>();
             foreach (MatchedResults resultPair in results)
             {
-                newRunResults.Add(resultPair.CalculateNewBaselineResult(PropertyBagMergeBehavior));
+                Result result = resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior);
+                newRunResults.Add(resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior));
+
+                var logicalLocationIndexRemapper = new LogicalLocationIndexRemapper(resultPair.LogicalLocations, remappedLogicalLocations);
```
michaelcfanning: >LogicalLocationIndexRemapper [](start = 55, length = 28)

this visitor fixes up indices so that merged results (absent, shared and new) all have logical location index property values that correspond to the newly merged logical locations in the computed baseline.

src/Sarif/NotYetAutoGenerated/Location.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// </summary>
+        [DataMember(Name = "properties", IsRequired = false, EmitDefaultValue = false)]
+        internal override IDictionary<string, SerializedPropertyInfo> Properties { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Location" /> class.
+        /// </summary>
+        public Location()
+        {
+            LogicalLocationIndex = -1;
```
michaelcfanning: >LogicalLocationIndex =  [](start = 12, length = 23)

It is helpful for constructors to populate default values. JSON.NET won't do so except when deserializing. our current jschema code does not do this.

src/Sarif/NotYetAutoGenerated/Location.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// </summary>
+        [DataMember(Name = "fullyQualifiedLogicalName", IsRequired = false, EmitDefaultValue = false)]
+        public string FullyQualifiedLogicalName { get; set; }
+
+        /// <summary>
+        /// The index within the logicalLocations array that specifies the logical location associated with the result.
+        /// </summary>
+        [DataMember(Name = "logicalLocationIndex", IsRequired = false, EmitDefaultValue = false)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [System.ComponentModel.DefaultValue(-1)]
```
michaelcfanning: missing jschema functionality. we must manually apply these attributes.

src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
```diff
@@ -756,13 +756,11 @@ public virtual Run VisitRun(Run node)
 
                 if (node.LogicalLocations != null)
                 {
-                    var keys = node.LogicalLocations.Keys.ToArray();
-                    foreach (var key in keys)
+                    if (node.LogicalLocations != null)
```
michaelcfanning: >LogicalLocations [](start = 29, length = 16)

i have to manually update this visitor because our jschema code does not have the logical to properly visit all dictionaries (where the key value might be updated). the more dictionary to array conversion work we do, the less we need this feature and may be able to drop the request altogether.

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-022.xml.sarif
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-09-26",
-  "version": "2.0.0-csd.2.beta.2018-09-26",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-11-28",
```
michaelcfanning: >$schema [](start = 3, length = 7)

A small # of test files had some inconsistencies that would have required manual updates to resolve. Instead, I've simply converted them to the current version of SARIF. We have plenty of test content that continues to be transformed (in fact, the bulk of our test content is still pre-release v2). It is impressive that our transformer continues to allow all our testing to function.

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-050.xml
```diff
@@ -1,100 +1,279 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<DEFECTS><DEFECT><SFA><FILEPATH>c:\somepath\</FILEPATH><FILENAME>test050.cpp</FILENAME><LINE>7</LINE><COLUMN>12</COLUMN></SFA><DEFECTCODE>26011</DEFECTCODE><DESCRIPTION>Potential read underflow using expression 'a[i]'
-PREFAST_NEWLINE
-Buffer accessed is a
-PREFAST_NEWLINE
-Buffer is of length 10 elements (4 bytes/element) [size of variable]
-PREFAST_NEWLINE
-Accessing the element (4 bytes/element) at element offset i`4
-PREFAST_NEWLINE
+<DEFECTS>
```
michaelcfanning: >DEFECTS [](start = 1, length = 7)

Here, I've expressly expanded some PREfast log file content based on reformatting in VS. This has slightly increased the variability of the test content.

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -67,7 +67,12 @@ public int Run(TransformOptions transformOptions)
                     else
                     {
                         // We have a pre-release v2 file that we should upgrade to current. 
-                        string sarifText = PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(_fileSystem.ReadAllText(inputFilePath), formatting: formatting);
+                        PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(
```
michaelcfanning: >PrereleaseCompatibilityTransformer [](start = 24, length = 34)

This helper now returns the transformed contents as a SarifLog. We need to do this in the transformer in order to invoke the visitors that rewrite indices. There is no sense converting this log file back to text again, so that a consumer deserializes yet one more time. Instead, we have a rather cumbersome API that can be used to get the transformed text as an out arg or the transformed log file as a returned value. Note that the out argument is set to the original contents that are passed to the API, in failure cases and when the file doesn't need to be touched at all.

src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError_Expected.sarif
```diff
@@ -33,7 +33,7 @@
                   "uri": "file:///C:/Code/sarif-sdk/bld/bin/Sarif.Multitool.FunctionalTests/AnyCPU_Debug/netcoreapp2.0/TestData/DeserializationError.sarif"
                 },
                 "region": {
-                  "startLine": 15,
+                  "startLine": 13,
```
michaelcfanning: >13 [](start = 31, length = 2)

Larry, I can't account for this line change. Does it make sense to you?

michaelcfanning:
The core change here is the transformation of logical locations from dictionaries (which are keyed, generally, by the fully qualified name of a logical location) to an array of logical locations. Result locations now reference logical locations by a logical location index. This changes removes the necessity of resolving key name collisions for logical locations that differ only by type (a namespace that collides with the fully qualified name of a type being the classic example).

In addition to making the core change, we have also authored a transformation that converts existing pre-release SARIF v2 files to the new design. We accomplish this by creating dictionaries, with value type comparison for keys, that are keyed by logical locations. This processing requires that any parent keys already exist in the array (so that a logical location can populate its parent logical location index, if any).

In addition to the core functionality and any transformation of individual log files, result matching presents special complications. In a result matching scenario, the logical location index of a result is not relevant to its identify: only the contents of the logical location this index points to are relevant. Furthermore, when merging a baseline file (which can contain results that are exclusive to a single log file within the matching domain), logical location indices are subject to change and must be updated.

For this scenario and at least one other, we use a visitor pattern to update indices. The reason is that locations are pervasive in the format and the visitor provides a convenient mechanism to put common location processing logical. This visitor uses puts additional pressure on the transformation logic, as it entails additional deserialization of the JSON. With more time/effort, we could have exhaustively updated all locations using the JParser/JObject/etc. API domain. Oh well. 

Finally, we must update the logical that transforms v1 to v2 and vice versa.

Whew. If that was not already sufficiently intrusive, this work revealed some minor flaws in various converters (the ones that handle logical locations): AndroidStudio, FxCop and PREfast.

This change is complex but valuable. Logical locations are now expressed as coherent objects in their table. In the main, I have preferred to leave result.fullyQualifiedName populated (in addition to result.logicalLocationIndex, to support easily looking up matching logical locations).

lgolding:
>         /// <param name="logicalLocationDictionary">

Update parameter name and text of doc comment.

---
Refers to: src/Sarif/IResultLogWriter.cs:37 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

src/Sarif/Autogenerated/LocationEqualityComparer.cs
```diff
@@ -38,6 +38,11 @@ public bool Equals(Location left, Location right)
                 return false;
             }
 
+            if (left.LogicalLocationIndex != right.LogicalLocationIndex)
+            {
+                return false;
+            }
```
lgolding: >} [](start = 12, length = 1)

This is relevant to your comment in the PR description: that two locations with different indices in different log files might actually turn out to be the same. As I continue the review I expect to see how we avoid this potentially incorrect generated behavior.

src/Sarif/Autogenerated/LogicalLocation.cs
```diff
@@ -51,10 +52,12 @@ public SarifNodeKind SarifNodeKind
         public string DecoratedName { get; set; }
 
         /// <summary>
-        /// Identifies the key of the immediate parent of the construct in which the result was detected. For example, this property might point to a logical location that represents the namespace that holds a type.
+        /// Identifies the indev of the immediate parent of the construct in which the result was detected. For example, this property might point to a logical location that represents the namespace that holds a type.
```
lgolding: >indev [](start = 27, length = 5)

"indev" => "index"

src/Sarif/Autogenerated/LogicalLocationEqualityComparer.cs
```diff
@@ -43,7 +43,7 @@ public bool Equals(LogicalLocation left, LogicalLocation right)
                 return false;
             }
 
-            if (left.ParentKey != right.ParentKey)
+            if (left.ParentIndex != right.ParentIndex)
             {
                 return false;
             }
```
lgolding: >} [](start = 12, length = 1)

Again, not necessarily correct when comparing logical location objects from different runs. We shall see how this plays out.

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
```diff
@@ -756,13 +756,11 @@ public virtual Run VisitRun(Run node)
 
                 if (node.LogicalLocations != null)
                 {
-                    var keys = node.LogicalLocations.Keys.ToArray();
-                    foreach (var key in keys)
+                    if (node.LogicalLocations != null)
```
lgolding: >if (node.LogicalLocations != null) [](start = 20, length = 34)

No need to repeat the test from Line 757. I see that this is a copy of the hand-coded file in the NotYetAutogenerated directory.

src/Sarif/NotYetAutoGenerated/Location.cs
lgolding: No, we have `AttributeHint` for that. See CodeGenHints.json for examples. In fact, you used it extensively to apply the `JsonConverter` attribute when you removed the contract resolver :-)

---
In reply to: [243699993](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243699993) [](ancestors = 243699993)

src/Sarif/NotYetAutoGenerated/LogicalLocation.cs
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft.  All Rights Reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
```
lgolding: >using [](start = 0, length = 5)

If this file too is here just for the sake of applying the `DefaultValue` attribute, again, you don't need it -- use `AttributeHint`.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -82,6 +84,9 @@ public Result CalculateNewBaselineResult(DictionaryMergeBehavior propertyBagMerg
             {
                 ResultMatchingProperties.Add(MatchedResults.MatchResultMetadata_RunKeyName, PreviousResult.OriginalRun.Id.InstanceGuid);
             }
+
+            LogicalLocations = PreviousResult.OriginalRun.LogicalLocations;
```
lgolding: >LogicalLocations [](start = 58, length = 16)

At first I worried that you were duplicating the entire run's worth of logical location array elements into each matched result, but now I see it's just a reference.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -800,9 +807,11 @@
           "type": "string"
         },
 
-        "parentKey": {
-          "description": "Identifies the key of the immediate parent of the construct in which the result was detected. For example, this property might point to a logical location that represents the namespace that holds a type.",
-          "type": "string"
+        "parentIndex": {
+          "description": "Identifies the indev of the immediate parent of the construct in which the result was detected. For example, this property might point to a logical location that represents the namespace that holds a type.",
```
lgolding: >indev [](start = 41, length = 5)

"indev" => " index"

src/Sarif/Autogenerated/LogicalLocation.cs
lgolding: Oh, right, this comes from the schema. I made the comment there.

---
In reply to: [243719265](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243719265) [](ancestors = 243719265)

src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
lgolding: 1. I don't actually understand this comment. It seems to be saying something about how the code that JSchema generates treats dictionaries, but logicalLocations is now an array -- Oh! you are saying that if only _all_ the dictionaries were arrays, we wouldn't need this hand-coded visitor, and this is a step in the right direction. Got it.

2. No need to repeat the test from line 757.

---
In reply to: [243700311](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243700311) [](ancestors = 243700311)

src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError_Expected.sarif
lgolding: There are two things happening:
1. We've changed Newtonsoft versions since the last time we regenerated this file, and
2. The test is not sensitive to `startLine`.

The error JSON1002 is "Required property is missing.". The arguments (Line 23) tell us that the required `frames` property of the `stack` object is missing, which is true. The `startLine` is where Newtonsoft thinks the offending object lives. And in fact, (13, 13) (the position immediately before the open brace of the empty `frame` object) is a better answer than (15, 13) (a terrible answer -- past the end of the line containing the closing square brace of the `frames` array).

---
In reply to: [243702036](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243702036) [](ancestors = 243702036)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-022.xml.sarif
```diff
@@ -12,46 +12,55 @@
           "mimeType": "text/x-cpp"
         }
       },
-      "logicalLocations": {
-        "f": {
+      "logicalLocations": [
+        {
+          "fullyQualifiedName": "f",
```
lgolding: >fullyQualifiedName [](start = 11, length = 18)

FWIW, the `name` property is required. `fullyQualifiedName` is required for non-top-level locations, optional for top-level locations (because then it is the same as `name`). See §3.27.3,  §3.27.4.

Rationale:
1. Before the dictionary-to-array transformation, you always had FQLN available because it was the dictionary key. We require the `fQLN` property, now, to get parity with that functionality.
2. Without `fQLN` being present, a consumer would have to (a) walk the parent tree to get the sequence of components that comprise the FQLN, and (b) know the language-specific syntax for composing the components into the FQLN string.
3. `name` is required because without it, a consumer would have to know the language-specific way to parse it out of the FQLN.

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-022.xml.sarif
```diff
@@ -12,46 +12,55 @@
...
           "decoratedName": "?foo@@YADXZ"
         },
-        "bar": {
+        {
+          "fullyQualifiedName": "bar",
           "decoratedName": "?bar@@YAXPAHII@Z"
         },
-        "ShiftTest::f": {
+        {
           "name": "f",
```
lgolding: >, [](start = 21, length = 1)

So yes, you did the right thing by supplying both `name` and `fQLN` for non-top-level locations. It's just that for top-level locations, it's `name` that's required, and `fQLN` is optional.

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-049.xml.sarif
```diff
@@ -12,20 +12,24 @@
           "mimeType": "text/x-cpp"
         }
       },
-      "logicalLocations": {
-        "": {
+      "logicalLocations": [
+        {
+          "fullyQualifiedName": "",
           "decoratedName": ""
         },
```
lgolding: The spec doesn't prohibit empty names, but... how did PREfast produce this?

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-050.xml.sarif
```diff
@@ -12,21 +12,21 @@
...
+        {
+          "fullyQualifiedName": "f",
           "decoratedName": "?f@@YAXH@Z"
         },
-        "f-0": {
-          "name": "f",
+        {
           "fullyQualifiedName": "f",
           "decoratedName": "?f@@YAXPAUS@@@Z"
         }
```
lgolding: >} [](start = 8, length = 1)

I wonder whether we need some sort of "readable name", like `f()` or `f(int, string)` to help understand the `decoratedName`. (Here we have two `f`s that differ by their decorated name due to different signatures. Or perhaps a tool that has the information can just use that readable name for `name` (and the `fQLN` would be `Namespace1::Namespace2::f(int, string)`

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-050.xml.sarif
```diff
@@ -101,7 +99,7 @@
...
-            "text": "Overflow using expression 'a[a[0]]'\nBuffer accessed is a\nBuffer is of length 10 elements (4 bytes/element) [size of variable]\nAccessing the element (4 bytes/element) at element offset 10\n\nValues of variables:\na[0] = 10\n\nThere are other instances of this error:\nPotential read overflow using expression 'a[i]' at line 7\nPotential overflow using expression 'a[a[1]]' at line 11\nOverflow using expression 'a[a[0]]' at line 15\nPotential overflow using expression 'a[a[2]]' at line 18\n"
+            "text": "\n      Overflow using expression 'a[a[0]]'\n            Buffer accessed is a\n            Buffer is of length 10 elements (4 bytes/element) [size of variable]\n            Accessing the element (4 bytes/element) at element offset 10\n      \n            Values of variables:\n            a[0] = 10\n      \n            There are other instances of this error:\n            Potential read overflow using expression 'a[i]' at line 7\n            Potential overflow using expression 'a[a[1]]' at line 11\n            Overflow using expression 'a[a[0]]' at line 15\n            Potential overflow using expression 'a[a[2]]' at line 18\n          "
```
lgolding: >\n       [](start = 21, length = 8)

As a result of your reformatting the PREfast XML files, all the messages have this leading white space (and multiline messages have the white space before each line). Is this worth it? What was the rationale for the reformatting?

lgolding: I mean, I saw your comment that says "VS did it", but still...

---
In reply to: [243852678](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243852678) [](ancestors = 243852678)

src/Sarif.Multitool/TransformCommand.cs
lgolding: IMO the out parameter should start out as `null` to handle the failure case. Later on, if no transform was necessary, it can be set to the input text.

---
In reply to: [243701919](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243701919) [](ancestors = 243701919)

lgolding:
>                         _fileSystem.WriteAllText(fileName, _fileSystem.ReadAllText(inputFilePath));

FWIW we should have `IFileSystem.CopyFile` that calls through to `System.IO.File.Copy`,

---
Refers to: src/Sarif.Multitool/TransformCommand.cs:83 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

lgolding:
>             tools = tools.Distinct().ToList();

FWIW this would suffice:
```
List<string> tools = runsByToolPrevious.Keys.Union(runsByToolCurrent.Keys).ToList();
```

---
Refers to: src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs:63 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -239,22 +239,28 @@ private Rule GetRuleFromResources(Result result, IDictionary<string, Rule> rules
...
+            var remappedLogicalLocations = new Dictionary<LogicalLocation, int>();
+
             properties = properties ?? new Dictionary<string, SerializedPropertyInfo>();
 
             List<Result> newRunResults = new List<Result>();
             foreach (MatchedResults resultPair in results)
             {
-                newRunResults.Add(resultPair.CalculateNewBaselineResult(PropertyBagMergeBehavior));
+                Result result = resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior);
+                newRunResults.Add(resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior));
```
lgolding: >resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior) [](start = 34, length = 62)

Is there a reason you didn't pass the temporary `result` you calculated on the previous line?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -87,20 +118,60 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+                    if (logicalLocations != null)
+                    {
+                        logicalLocationToIndexMap = new Dictionary<LogicalLocation, int>(LogicalLocation.ValueComparer);
+                        var jObjectToIndexMap = new Dictionary<JObject, int>();
+
+                        run["logicalLocations"] =
+                            ConstructLogicalLocationsArray(
+                                logicalLocations,
+                                jObjectToIndexMap,
+                                logicalLocationToIndexMap,
```
lgolding: >logicalLocationToIndexMap [](start = 32, length = 25)

BUG: We pass this empty dictionary into `ConstructLogicalLocationsArray`. But that function immediately overwrites the parameter with a new empty dictionary. So the passed-in dictionary is never modified, and you always pass the empty dictionary to the `LogicalLocationIndexRemapper` on Line 164.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -87,20 +118,60 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+
+                    if (logicalLocations != null)
+                    {
+                        logicalLocationToIndexMap = new Dictionary<LogicalLocation, int>(LogicalLocation.ValueComparer);
+                        var jObjectToIndexMap = new Dictionary<JObject, int>();
+
+                        run["logicalLocations"] =
+                            ConstructLogicalLocationsArray(
+                                logicalLocations,
+                                jObjectToIndexMap,
```
lgolding: >jObjectToIndexMap [](start = 32, length = 17)

NIT: We never use this local variable after passing the empty dictionary into `ConstructLogicalLocationsArray`, so we may as well let that method create the dictionary.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: This has got to indicate a test hole. I don't see how the call to `Visit` on Line 166 can work if the remapper always holds an empty array.

---
In reply to: [243855820](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243855820) [](ancestors = 243855820)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -87,20 +118,60 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+                                logicalLocationIndexRemapper.Visit(sarifResult);
+                                rewrittenResults = rewrittenResults ?? new JArray();
+                                rewrittenResults.Add(JToken.Parse(JsonConvert.SerializeObject(sarifResult)));
+                            }
+                        }
+
+                        if (rewrittenResults != null)
+                        {
+                            run["results"] = rewrittenResults;
                         }
```
lgolding: >} [](start = 24, length = 1)

BUG: Set `modifiedLog = true` after overwriting `run["results"]1.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -131,13 +201,128 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+
+            if (logicalLocations == null) { return null; }
+
+            logicalLocationToIndexMap = new Dictionary<LogicalLocation, int>(LogicalLocation.ValueComparer);
+
+            fullyQualifiedLogicalNameToIndexMap = new Dictionary<string, int>();
+
+            foreach (JProperty logicalLocationEntry in logicalLocations.Properties())
+            {
+                // This condition may occur if we've already procssed a parent logical
```
lgolding: >procssed [](start = 61, length = 8)

"processd" => "processed"

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -131,13 +201,128 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+        private static Dictionary<string, int> CreateLogicalLocationArrayEntry(
+            JObject logicalLocationsDictionary,
+            string keyName,
+            JObject logicalLocation,
+            Dictionary<LogicalLocation, int> logicalLocationToIndexMap,
+            Dictionary<JObject, int> jObjectToIndexMap,
+            Dictionary<string, int> keyToIndexMap)
+        {
+            int index;
+            keyToIndexMap = keyToIndexMap ?? new Dictionary<string, int>();
```
lgolding: >keyToIndexMap = keyToIndexMap ?? new Dictionary<string, int>(); [](start = 12, length = 63)

The caller is passing a non-null dictionary. IMO you should just update it in place and not bother to return a "new" one (which is only necessary if the input had been `null`). I'd replace this with an assertion that `keyToIndexMap` is non-null, update it in place, and return `void`.

And the method name is a bit confusing: a method `CreateThing` should return a `Thing`. This method doesn't do that, whether or not you adopt my suggestion around `keyToIndexMap`. It should be `AddEntryToFullyQualifiedNameToIndexMap`, and the signature IMO should be (putting "the thing we're adding" first in the argument list, and the thing we're modifying last):

```
private static void AddEntryToFullyQualifiedNameToIndexMap(
    string fullyQualifiedLogicalName,
    JObject logicalLocationObject,
    Dictionary<LogicalLocation, int> logicalLocationToIndexMap,
    Dictionary<JObject, int> jObjectToIndexMap,
    Dictionary<string, int> fullyQualifiedLogicalNameToIndexMap);
```

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -131,13 +201,128 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+                if (!keyToIndexMap.TryGetValue(parentKey, out index))
+                {
+                    // The parent hasn't been processed yet. We must force its creation
+                    // determine its index in our array. This code path results in 
+                    // an array order that does not precisely match the enumeration
+                    // order of the logical locations dictionary.
+                    keyToIndexMap = CreateLogicalLocationArrayEntry(
+                        logicalLocationsDictionary,
+                        parentKey,
+                        (JObject)logicalLocationsDictionary[parentKey],
```
lgolding: >parentKey [](start = 60, length = 9)

If the SARIF file is invalid, in that it contains a `logicalLocation` object with a `parentKey`, but the parent isn't in the dictionary, then you'll get some generic "Key not found" exception. Consider checking for this, and throwing an exception with a message like "The logical location 'X::Y' specifies that its parent location is 'X', but the `run.logicalLocations` dictionary does not have an entry with key 'X'.".

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -131,13 +201,128 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+
+            if (parentKey == null)
+            {
+                // No parent key? We are a root location. 
+                logicalLocation["parentIndex"] = -1;
+            }
+            else
+            {
+                // Have we processed our parent yet? If so, retrieve the parentIndex and we're done
+                if (!keyToIndexMap.TryGetValue(parentKey, out index))
```
lgolding: >index [](start = 62, length = 5)

Rather than reusing `index` (which forces you to declare `index` at the top of the method, increasing the reader's cognitive load), consider using a local with a more specific name:
```
if (!keyToIndexMap.TryGetValue(parentKey, out int parentIndex))
```
... using the C# 7 out parameter declaration feature.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -131,13 +201,128 @@ private static bool ApplyChangesFromTC25ThroughTC28(JObject sarifLog)
...
+                        logicalLocationsDictionary,
+                        parentKey,
+                        (JObject)logicalLocationsDictionary[parentKey],
+                        logicalLocationToIndexMap,
+                        jObjectToIndexMap,
+                        keyToIndexMap);
+
+                    index = keyToIndexMap[parentKey];
+                }
+                logicalLocation["parentKey"] = null;
```
lgolding: >null [](start = 47, length = 4)

Consider: instead of setting the property to `null`, remove it:
```
logicalLocation.Remove("parentKey");
```

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: Ok, I see that this method can be called recursively, so it has to start off with the dictionary object.

---
In reply to: [243855979](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243855979) [](ancestors = 243855979)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -0,0 +1,33 @@
...
+            _fullyQualifiedNameToIndexMap = fullyQualifiedNameToIndexMap;
+        }
+
+        public override Location VisitLocation(Location node)
+        {
+            int index;
+
+            if (_fullyQualifiedNameToIndexMap != null && !string.IsNullOrEmpty(node.FullyQualifiedLogicalName))
+            {
+                if (_fullyQualifiedNameToIndexMap.TryGetValue(node.FullyQualifiedLogicalName, out index))
```
lgolding: >out [](start = 94, length = 3)

`out int index`, so you don't have to declare the variable separately.

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -0,0 +1,33 @@
...
+        public UpdateIndicesVisitor(IDictionary<string, int> fullyQualifiedNameToIndexMap)
+        {
+            _fullyQualifiedNameToIndexMap = fullyQualifiedNameToIndexMap;
+        }
+
+        public override Location VisitLocation(Location node)
+        {
+            int index;
+
+            if (_fullyQualifiedNameToIndexMap != null && !string.IsNullOrEmpty(node.FullyQualifiedLogicalName))
```
lgolding: >IsNullOrEmpty [](start = 65, length = 13)

`NullOrEmpty` -- seems reasonable, but there is a PREfast test file where I commented that you actually had a logical location with an empty name.

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v2/NotificationExceptionWithStack.sarif
```diff
@@ -76,10 +78,11 @@
                       "location": {
                         "physicalLocation": {
                           "fileLocation": {
-                            "uri": "file:///C:/src/foobar.cs"
+                            "uri": "file:///C:/src/testtarget.cs"
                           }
                         },
-                        "fullyQualifiedLogicalName": "ExecutionEngine.Engine.FooBar"
+                        "fullyQualifiedLogicalName": "ExecutionEngine.Engine.TestTarget",
+                        "logicalLocationIndex": 0
```
lgolding: >logicalLocationIndex [](start = 25, length = 20)

`ExecutionEngine.Engine.TestTarget` is at index 2, not 0.

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v2/OneRunWithLogicalLocations.sarif
```diff
@@ -22,7 +22,56 @@
...
+      "results": [
+        {
+          "ruleId": "TEST1001",
+          "message": {
+            "text": "A result fired against 'collections::list'."
+          },
+          "locations": [
+            {
+              "fullyQualifiedLogicalName": "collections::list",
+              "logicalLocationIndex": 1
```
lgolding: >logicalLocationIndex [](start = 15, length = 20)

? This is a 2018-10-10 file, so `run.logicalLocations` is a dictionary, so ` location` object shouldn't have a `logicalLocationIndex` property.

src/Sarif/Visitors/LogicalLocationIndexRemapper.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            _previousLogicalLocations = previousLogicalLocations;
+            RemappedLogicalLocations = remappedLogicalLocations;
+        }
+
+        public IDictionary<LogicalLocation, int> RemappedLogicalLocations { get; private set; }
+
+        public override Location VisitLocation(Location node)
+        {
+            if (_previousLogicalLocations == null) { return node; }
```
lgolding: >_previousLogicalLocations [](start = 16, length = 25)

If `_previoiusLogicalLocations` is `null`, there's no point visiting any of the locations. So you could cut the visit off early, like:
```
{
    if (_previousLogicalLocations == null) { return node; } // ... and remove this line from VisitLocation.

    base.VisitResult(node);
}
```

src/Sarif/Visitors/LogicalLocationIndexRemapper.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            LogicalLocation logicalLocation = _previousLogicalLocations[node.LogicalLocationIndex];
```
lgolding: >LogicalLocationIndex [](start = 77, length = 20)

Check that `node.LogicalLocationIndex` is not -1 before indexing.

lgolding:
>     /// The logic for populating the logicalLocations dictionary is encapsulated

"dictionary" => "array", or just "for populating logicalLocations".

---
Refers to: src/Sarif.Converters.UnitTests/ToolFileConverterBaseTests.cs:12 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

src/Sarif.Converters/ToolFileConverterBase.cs
```diff
@@ -16,68 +17,34 @@ public abstract class ToolFileConverterBase
...
         protected ToolFileConverterBase()
         {
-            LogicalLocationsDictionary = new Dictionary<string, LogicalLocation>();
+            _logicalLocationToIndexMap = new Dictionary<LogicalLocation, int>(LogicalLocation.ValueComparer);
+            LogicalLocations = new List<LogicalLocation>();
         }
 
         public abstract void Convert(Stream input, IResultLogWriter output, OptionallyEmittedData dataToInsert);
 
         // internal as well as protected it can be exercised by unit tests.
```
lgolding: >it [](start = 41, length = 2)

While we're here, "it" => "so it"

src/Sarif.Converters/PREFastConverter.cs
```diff
@@ -112,16 +115,21 @@ private Result CreateResult(Defect defect)
             return result;
         }
 
-        private void AddLogicalLocation(string name, string decoratedName)
+        private int AddLogicalLocation(string name, string decoratedName)
         {
-            // Use the FQLN so the correct dictionary key is generated by AddLogicalLocation
+            string fullyQualifiedName = name;
+
+            name = name.Contains("::") ? name.Substring(name.LastIndexOf("::") + "::".Length) : null;
```
lgolding: >"::" [](start = 33, length = 4)

While we're here, consider:
```
const string ScopeOperator = "::";
```

lgolding: >name [](start = 12, length = 4)

As I noted in a comment on one of the test files, this logic is actually backwards. `name` is required, and `fullyQualifiedName` is required unless it is the same as `name` (a top-level location).
```
string fullyQualifiedName = name;

int lastScopeOperatorPosition = name.Substring(name.LastIndexOf(ScopeOperator));
if (lastScopeOperatorPosition != -1)
{
    // This is a nested logical location, so "name" is the rightmost component.
    name = name.Substring(lastScopeOperatorPosition + ScopeOperator.Length)
}
else
{
    // This is a top-level logical location; fullyQualifiedName is the same as name, so
    // it's not necessary to emit it.
    fullyQualifiedName =null;
}
```

src/Sarif.Converters.UnitTests/FxCopConverterTests.cs
```diff
@@ -328,6 +328,15 @@ public void FxCopLogReader_Context_RefineProjectToExceptionIssue()
 
     public class FxCopConverterTests
     {
+        private static void ValidateLogicalLocations(IList<LogicalLocation> expectedLogicalLocations, IList<LogicalLocation> actualLogicalLocations)
+        {
+            for (int i = 0; i < expectedLogicalLocations.Count; i++)
+            {
+                expectedLogicalLocations[i].ValueEquals(actualLogicalLocations[i]).Should().BeTrue();
+            }
+            actualLogicalLocations.Count.Should().Be(expectedLogicalLocations.Count);
```
lgolding: >actualLogicalLocations [](start = 12, length = 22)

Consider: validating the count first. If the count doesn't match there's no reason to compare the elements.

lgolding:
>         public void FxCopConverter_CreateIssue_FakeContext_NoModule_Member()

While we're in here: Issue => Result

---
Refers to: src/Sarif.Converters.UnitTests/FxCopConverterTests.cs:424 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

src/Sarif.Converters.UnitTests/AndroidStudioConverterTests.cs
```diff
@@ -213,6 +213,15 @@ public void AndroidStudioConverter_ConvertSarifResult_DoesNotRecordTopLevelEntry
             loc.PhysicalLocation.Should().BeNull();
         }
 
+        private static void ValidateLogicalLocations(IList<LogicalLocation> expectedLogicalLocations, IList<LogicalLocation> actualLogicalLocations)
```
lgolding: >ValidateLogicalLocations [](start = 28, length = 24)

If you derive `AndroidStudioConverterTests` and `FxCopConverterTests` from a new `ToolFileConverterTestsBase` class (not to be confused with `ToolFileConverterBaseTests`!), you can DRY this out by moving it to the base class.

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -150,51 +152,73 @@ public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
...
+                //}
             }
 
             result.Locations = new List<Location> { location };
 
             return result;
         }
 
-        private string CreateFullyQualifiedLogicalName(AndroidStudioProblem problem)
+        private string CreateFullyQualifiedLogicalName(AndroidStudioProblem problem, out int index)
```
lgolding: >CreateFullyQualifiedLogicalName [](start = 23, length = 31)

This method actually does three things, and although it's hard to come up with a good method name that conveys that [*], it's at least worth a summary comment:
1. It adds an entry to the LogicalLocations array for each nested level of the logical location where the problem occurs.
2. It builds up and returns the fully qualified name of the most deeply nested location.
3. It sets index to the index within LogicalLocations of the newly created entry for the most deeply nested location.

[*] Usually a code smell that the method should be refactored, but let that be for another day.

lgolding: On further reflection, I think a good method name is `AddLogicalLocationAndAncestors`. It returns information about the (most deeply nested) logical location.

---
In reply to: [243866307](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243866307) [](ancestors = 243866307)

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -150,51 +152,73 @@ public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
...
 
-            return parentLogicalLocationKey;
+            return fullyQualifiedName;
         }
 
-        private string AddLogicalLocation(string parentKey, string value, string kind, string delimiter = @"\")
+        private int AddLogicalLocation(int parentIndex, ref string fullyQualifiedName, string value, string kind, string delimiter = ".")
         {
-            if (!String.IsNullOrEmpty(value))
+            fullyQualifiedName = fullyQualifiedName + delimiter + value;
```
lgolding: >fullyQualifiedName [](start = 33, length = 18)

How does this work on the first call, where `fullyQualifiedName` comes in as `null`? I could see this working if it was initialized to `string.Empty` on Line 175.

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -150,51 +152,73 @@ public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
...
+
+            // Need to decide which item gets preference, name vs. fully qualified name
+            // if both match. The behaviors of various API in the SDK differs on this point.
```
lgolding: >match [](start = 23, length = 5)

As noted elsewhere: the spec says that if `name` and `fullyQualifiedName` match (that is, if this is a top-level logical location), the `name` is required (as it _always_ is), and `fullyQualifiedLogicalName` is optional.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -50,7 +50,6 @@ internal AnnotatedCodeLocationVersionOne CreateAnnotatedCodeLocation(Location v2
                 {
                     Annotations = v2Location.Annotations?.Select(CreateAnnotation).ToList(),
                     FullyQualifiedLogicalName = v2Location.FullyQualifiedLogicalName,
-                    LogicalLocationKey = v2Location.FullyQualifiedLogicalName,
```
lgolding: >LogicalLocationKey [](start = 20, length = 18)

Why was this removed?

It's true that if you're going to keep it, you have to use `v2Location.FullyQualifiedLogicalName ?? v2Location.Name`, because `fullyQualifiedLogicalName` is optional if it's the same as `name`. But I'd've expected to see that fix, rather than to see the line removed.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -305,27 +304,32 @@ internal LocationVersionOne CreateLocation(Location v2Location)
                     ResultFile = CreatePhysicalLocation(v2Location.PhysicalLocation)
                 };
 
-                if (!string.IsNullOrWhiteSpace(v2Location.FullyQualifiedLogicalName) &&
-                    _currentV2Run.LogicalLocations?.ContainsKey(v2Location.FullyQualifiedLogicalName) == true)
+                if (!string.IsNullOrWhiteSpace(v2Location.FullyQualifiedLogicalName))
                 {
-                    location.DecoratedName = _currentV2Run.LogicalLocations[v2Location.FullyQualifiedLogicalName].DecoratedName;
+                    location.DecoratedName = _currentV2Run.LogicalLocations[v2Location.LogicalLocationIndex].DecoratedName;
```
lgolding: >LogicalLocationIndex [](start = 87, length = 20)

BUG: You have to test that the index isn't -1.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -305,27 +304,32 @@ internal LocationVersionOne CreateLocation(Location v2Location)
...
 
         internal LogicalLocationVersionOne CreateLogicalLocation(LogicalLocation v2LogicalLocation)
-        {
+        {            
             LogicalLocationVersionOne logicalLocation = null;
+            string parentKey = null;
+
+            if (_currentV2Run.LogicalLocations != null && v2LogicalLocation.ParentIndex > -1)
+            {
+                parentKey = _currentV2Run.LogicalLocations[v2LogicalLocation.ParentIndex].FullyQualifiedName;
```
lgolding: >FullyQualifiedName [](start = 90, length = 18)

BUG: Again, since FQLN is optional if it's the same as name, you need FQLN ?? Name.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -891,6 +895,22 @@ internal RunVersionOne CreateRun(Run v2Run)
...
+        private IDictionary<string, LogicalLocationVersionOne> CreateLogicalLocationsDictionary(IList<LogicalLocation> logicalLocations)
+        {
+            Dictionary<string, LogicalLocationVersionOne> logicalLocationsVersionOne = null;
+
+            if (logicalLocations != null)
+            {
+                logicalLocationsVersionOne = new Dictionary<string, LogicalLocationVersionOne>();
+                foreach (LogicalLocation logicalLocation in logicalLocations)
+                {
+                    logicalLocationsVersionOne[logicalLocation.FullyQualifiedName] = CreateLogicalLocation(logicalLocation);
```
lgolding: >FullyQualifiedName [](start = 63, length = 18)

Again FQLN ?? Name.

You know, this happens often enough that I suggest we add a partial class for `LogicalLocation` that defines a property... I don't know what to call it, "SafeName" or something... that returns FQLN ?? Name, and use that in all the places I mentioned.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -930,11 +950,10 @@ internal StackFrameVersionOne CreateStackFrame(StackFrame v2StackFrame)
                     string fqln = location.FullyQualifiedLogicalName;
 
                     if (_currentV2Run.LogicalLocations != null &&
-                        _currentV2Run.LogicalLocations.ContainsKey(fqln) &&
-                        !string.IsNullOrWhiteSpace(_currentV2Run.LogicalLocations[fqln].FullyQualifiedName))
+                        !string.IsNullOrWhiteSpace(_currentV2Run.LogicalLocations[location.LogicalLocationIndex].FullyQualifiedName))
```
lgolding: >LogicalLocationIndex [](start = 91, length = 20)

Test for -1. DRY out code by computing this expression only once: `_currentV2Run.LogicalLocations[location.LogicalLocationIndex].FullyQualifiedName` (and as usual, it's FQLN ?? Name.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -332,20 +337,22 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
                     FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
                     PhysicalLocation = CreatePhysicalLocation(v1Location.ResultFile),
-                    Properties = v1Location.Properties
+                    Properties = v1Location.Properties,
                 };
 
                 if (!string.IsNullOrWhiteSpace(location.FullyQualifiedLogicalName))
                 {
-                    if (_currentRun.LogicalLocations?.ContainsKey(location.FullyQualifiedLogicalName) == true)
+                    if (_v1KeyToV2LogicalLocationMap.TryGetValue(location.FullyQualifiedLogicalName, out LogicalLocation logicalLocation))
```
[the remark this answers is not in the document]
lgolding: Alright, looking over the file as a whole, I see that it's initialized in `CreateRun`, which happens before any of the subsidiary objects are created. Still initializing it in the ctor makes it so much easier for the reader to convince the reader that it's correct. The micro-optimization of not creating this empty array until/unless it's needed will never pay back the time I spent pondering this ;-)

---
In reply to: [243867544](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867544) [](ancestors = 243867544)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -332,20 +337,22 @@ internal Location CreateLocation(LocationVersionOne v1Location)
                 {
                     FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
```
lgolding: >v1Location.LogicalLocationKey [](start = 48, length = 29)

I don't think this was ever right. `LogicalLocationKey` was a synthetic key, like `N1::N2-1`, that was used when two FQLNs were identical. It would never have been right to assign that value to a v2 FQLN.

In fact, this logic is backwards. It should be:
```
FullyQualifiedLogicalName = v1Location.FullyQualifiedLogicalName ?? v1Location.LogicalLocationKey,
```
v1 FQLN is guaranteed to be the right answer if it's there; if it's absent, it's because it matched the key.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: >FullyQualifiedLogicalName [](start = 74, length = 25)

This is not the right lookup key. In the event of a collision, you'd have to do the lookup by means of the v1 key. So you need to do:
```
string v1Key = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName;
```
... and pass `v1Key` to `TryGetValue`.

The value of `v1Key` is the string that was incorrectly used as the value of v2 FQLN on Line 338. Turns out we need that string after all, but for something else :-)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -332,20 +337,22 @@ internal Location CreateLocation(LocationVersionOne v1Location)
...
-                    {
-                        LogicalLocation logicalLocation = CreateLogicalLocation(location.FullyQualifiedLogicalName,
-                                                                                decoratedName: v1Location.DecoratedName);
-                        location.FullyQualifiedLogicalName = AddLogicalLocation(logicalLocation);
+                        _logicalLocationToIndexMap.TryGetValue(logicalLocation, out int index);
+
+                        if (!string.IsNullOrEmpty(logicalLocation.DecoratedName))
+                        {
+                            logicalLocation.DecoratedName = v1Location.DecoratedName;
+                            _logicalLocationToIndexMap[logicalLocation] = index;
```
lgolding: >index [](start = 74, length = 5)

I don't get why the population of `_logicalLocationToIndexMap` depends on whether there's a decorated name. If this code is correct, it needs a comment.

lgolding:
>                     FullyQualifiedLogicalName = v1AnnotatedCodeLocation.LogicalLocationKey ?? v1AnnotatedCodeLocation.FullyQualifiedLogicalName,

The same issue as above with the role reversal between the key and the FQLN, here and starting on Line 383.

---
Refers to: src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs:377 in 37b9568. [](commit_id = 37b9568f74eb44d7d86e0f0628f62df7b3fe17d4, deletion_comment = False)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -412,21 +428,20 @@ internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCode
...
-                _currentRun.LogicalLocations.TryGetValue(logicalLocationKey, out logicalLocation))
-            {
-                logicalLocation.FullyQualifiedName = fullyQualifiedLogicalName;
-                logicalLocation.Name = GetLogicalLocationName(fullyQualifiedLogicalName);
-                location.FullyQualifiedLogicalName = logicalLocationKey;
-            }
-            else if (!string.IsNullOrWhiteSpace(fullyQualifiedLogicalName))
+            logicalLocation.FullyQualifiedName = fullyQualifiedLogicalName;
+            logicalLocation.Name = GetLogicalLocationName(fullyQualifiedLogicalName);
+            location.FullyQualifiedLogicalName = logicalLocationKey;
```
lgolding: >logicalLocationKey [](start = 49, length = 18)

Again, the key/FQLN reversal.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -990,9 +959,48 @@ internal Run CreateRun(RunVersionOne v1Run)
...
+        private void PopulateLogicalLocation(
+            Run v2Run, 
+            IDictionary<string, LogicalLocationVersionOne> v1LogicalLocations, 
+            string fullyQualifiedName, 
+            LogicalLocationVersionOne v1LogicalLocation, 
+            HashSet<string> populatedKeys)
+        {
+            // We saw and populated this one previously, because it was a parent to 
+            // a logical location that we encountered earlier
+            if (populatedKeys.Contains(fullyQualifiedName)) { return; }
```
lgolding: >fullyQualifiedName [](start = 39, length = 18)

Given how you called this method on Line 920, it's clear that the thing you're calling FQLN here is actually the v1 dictionary key. So, given the massive potential for confusion between the key and the FQLN, manifested several places above, it's important to name this `v1Key`, and make sure not to use it in a place where you really do need the FQLN.

lgolding:
Although I had a great many comments, it's not as bad as it seems.

Many of the comments are nits. There are also some bugs, but for the most part they are simple point fixes that I wouldn't hold up the approval for.

The reason I'm marking this PR "Code not Ready" is because of a pervasive problem in SarifVersionOneToCurrentVisitor.cs. There are many places in that code where the v1 dictionary key is used in a place where the FQLN should be used, and vice versa. There was already at least one existing bug of that kind, which I commented on. But the new code contains several other instances of this problem. That file needs to be looked at carefully before we merge (and we are clearly missing some test coverage around converting v1 files with logical location key collisions.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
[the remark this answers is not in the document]
lgolding: And ok, it's not the ctor, it's `VisitSarifLogVersionOne`. Still, it's the top-level method, it's right at the top of the file, and it's the place where you initialize `_logicalLocationToIndexMap`, so I'm going to move it.

---
In reply to: [243867847](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867847) [](ancestors = 243867847,243867544)

michaelcfanning:
Really appreciate the review! Perhaps we can take the small changes now, make note of additional changes and check-in. I will work on the files array conversion and perhaps you can clean up the VersionOneToCurrent issues?

---
In reply to: [449791980](https://github.com/Microsoft/sarif-sdk/pull/1170#issuecomment-449791980) [](ancestors = 449791980)

src/Sarif.Multitool/TransformCommand.cs
lgolding: I looked at `PrereleaseCompatibilityTransformer.UpdateToCurrentVersion` and contrary to this comment, it _does_ initially set the out parameter to `null`, as I suggested it should, so there's nothing to do here.

---
In reply to: [243852926](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243852926) [](ancestors = 243852926,243701919)

michaelcfanning:
Got it, proceeding as you say.

---
In reply to: [450013084](https://github.com/Microsoft/sarif-sdk/pull/1170#issuecomment-450013084) [](ancestors = 450013084,450011808,449791980)

src/Sarif/NotYetAutoGenerated/Location.cs
michaelcfanning: Documented gap in the following issue:
[https://github.com/Microsoft/jschema/issues/57](https://github.com/Microsoft/jschema/issues/57)

---
In reply to: [243699926](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243699926) [](ancestors = 243699926)

src/Sarif/NotYetAutoGenerated/Location.cs
michaelcfanning: Right. Still would be helpful to have more automatic code gen for this. Note that System.ComponentModel.DefaultValue is a standard attribute. A more important gap, mentioned in another comment thread: we need to initialize in ctor. I've updated comments in the issue below.

[https://github.com/Microsoft/jschema/issues/57](https://github.com/Microsoft/jschema/issues/57)

---
In reply to: [243719649](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243719649) [](ancestors = 243719649,243699993)

src/Sarif.Converters/PREFastConverter.cs
michaelcfanning: Yes. [https://github.com/Microsoft/sarif-sdk/issues/1173](https://github.com/Microsoft/sarif-sdk/issues/1173) 

---
In reply to: [243865028](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243865028) [](ancestors = 243865028)

src/Sarif/Autogenerated/LocationEqualityComparer.cs
michaelcfanning: Someone should explicitly review and confirm this in todo: [https://github.com/Microsoft/sarif-sdk/issues/1174](https://github.com/Microsoft/sarif-sdk/issues/1174)

---
In reply to: [243719244](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243719244) [](ancestors = 243719244)

src/Sarif/Autogenerated/LogicalLocationEqualityComparer.cs
michaelcfanning: It is critical that result matching travels the complete parent chain. It is also critical that ParentIndex itself is not part of our value type comparison semantics (but it is today!)
[https://github.com/Microsoft/sarif-sdk/issues/1175](https://github.com/Microsoft/sarif-sdk/issues/1175)
[https://github.com/Microsoft/sarif-sdk/issues/1176](https://github.com/Microsoft/sarif-sdk/issues/1176)

---
In reply to: [243719340](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243719340) [](ancestors = 243719340)

src/Sarif/NotYetAutoGenerated/LogicalLocation.cs
michaelcfanning: [https://github.com/Microsoft/sarif-sdk/issues/1177](https://github.com/Microsoft/sarif-sdk/issues/1177)

---
In reply to: [243719690](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243719690) [](ancestors = 243719690)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-022.xml.sarif
michaelcfanning: TODO: [https://github.com/Microsoft/sarif-sdk/issues/1173](https://github.com/Microsoft/sarif-sdk/issues/1173)

---
In reply to: [243851854](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243851854) [](ancestors = 243851854)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-049.xml.sarif
michaelcfanning: TODO. We should look at the logical for all logical location processing things: android studio, fxcop and prefast converters. Note that updating test files for these is tricky. If we change the code logical, new files won't match old. If we universally update all these older test files, we will no longer use the prerelease transformation logic to verify them. And so I suggest we defer these issues until we're ready for that wholesale conversion.

[https://github.com/Microsoft/sarif-sdk/issues/1171](https://github.com/Microsoft/sarif-sdk/issues/1171)

---
In reply to: [243852202](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243852202) [](ancestors = 243852202)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-050.xml.sarif
michaelcfanning: Yes. Presumably, this collision exists simply because someone hasn't sufficiently qualified the 'name' property. These collisions can occur in C++ due to failure to express the presence of some qualifiers (such as 'const') on various components. Same problem: resolves to unique linkable thing, has a distinct decorated name, but a shared undecorated name.

---
In reply to: [243852449](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243852449) [](ancestors = 243852449)

src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-050.xml.sarif
michaelcfanning: The truth is that this reformatting wasn't necessary. I happened to reformat just for readability. On doing so, however, I observed we have this undesired introduction of newlines, etc., that flow from the original file to SARIF. I left these artifacts so that we'd drive code improvements. Tracking item below. I suggest deferring all converter improvements until we get past the files array change (when we'll be ready to do more of a wholesale conversion of prerelease sarif v2 to current). Right now we need all the old content to help validate the ongoing prerelease transformation work.

[https://github.com/Microsoft/sarif-sdk/issues/1169](https://github.com/Microsoft/sarif-sdk/issues/1169)

---
In reply to: [243852699](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243852699) [](ancestors = 243852699,243852678)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: These are both good suggestions. 

---
In reply to: [243857089](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243857089) [](ancestors = 243857089)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Actually, the semantics of this JSON dictionary is that a null value will be returned. I can easily add a null check at the beginning of the method.

---
In reply to: [243857356](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243857356) [](ancestors = 243857356)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
michaelcfanning: s/be cleaned up by other work. i don't think we'll want to serialize those empty values.

---
In reply to: [243858172](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243858172) [](ancestors = 243858172)

src/Sarif.Converters/AndroidStudioConverter.cs
michaelcfanning: tracking item created as noted

---
In reply to: [243866740](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243866740) [](ancestors = 243866740)

src/Sarif.Converters/AndroidStudioConverter.cs
michaelcfanning: .NET string.concat happily accepts null and treats these as, effectively, string.empty instances.

---
In reply to: [243866667](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243866667) [](ancestors = 243866667)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: broken scenario. opened issue, marked 'urgent'

[https://github.com/Microsoft/sarif-sdk/issues/1180](https://github.com/Microsoft/sarif-sdk/issues/1180)

---
In reply to: [243867007](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867007) [](ancestors = 243867007)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: [https://github.com/Microsoft/sarif-sdk/issues/1181](https://github.com/Microsoft/sarif-sdk/issues/1181) marked urgent. suggest just creating a simple helper for constructing the key name

---
In reply to: [243867117](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867117) [](ancestors = 243867117)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: I think a helper function probably makes most sense. an extension method off LogicalLocation perhaps. LogicalLocation.GetSarifVersionOneLogicalLocationsDictionaryKey. whew

---
In reply to: [243867203](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867203) [](ancestors = 243867203)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: resolve as part of [https://github.com/Microsoft/sarif-sdk/issues/1181](https://github.com/Microsoft/sarif-sdk/issues/1181)

---
In reply to: [243867281](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243867281) [](ancestors = 243867281)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: [https://github.com/Microsoft/sarif-sdk/issues/1182](https://github.com/Microsoft/sarif-sdk/issues/1182) marked urgent

---
In reply to: [243868044](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243868044) [](ancestors = 243868044)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: [https://github.com/Microsoft/sarif-sdk/issues/1182](https://github.com/Microsoft/sarif-sdk/issues/1182)

---
In reply to: [243868340](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243868340) [](ancestors = 243868340)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Reactivated for the sake of adding the -1 check (otherwise I'll forget). The rest of the change is covered by #1181.

---
In reply to: [244054845](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r244054845) [](ancestors = 244054845,243867281)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: something looks odd here. also note that when we can't retrieve the logical location, we're setting the index to 0 which is wrong. 

[https://github.com/Microsoft/sarif-sdk/issues/1183](https://github.com/Microsoft/sarif-sdk/issues/1183)

---
In reply to: [243868420](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243868420) [](ancestors = 243868420)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: covered elsewhere

---
In reply to: [243868533](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243868533) [](ancestors = 243868533)

src/Sarif.Converters.UnitTests/AndroidStudioConverterTests.cs
michaelcfanning: Yes. But I generally do not introduce a new layer in the class hierarchy to share a single helper. YMMV. I've converted this code to at least utilize a shared static helper.

---
In reply to: [243865843](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243865843) [](ancestors = 243865843)

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v2/NotificationExceptionWithStack.sarif
michaelcfanning: This indicates a very serious problem somewhere. Issue marked as 'urgent'.

[https://github.com/Microsoft/sarif-sdk/issues/1179](https://github.com/Microsoft/sarif-sdk/issues/1179)


---
In reply to: [243858528](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r243858528) [](ancestors = 243858528)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: Fixed by Michael in PR #1184.

---
In reply to: [244051988](https://github.com/Microsoft/sarif-sdk/pull/1170#discussion_r244051988) [](ancestors = 244051988,243857089)

Issue #1183, opened 2018-12-26
Logical in CreateLocation(LocationVersionOne) looks weird/broken


michaelcfanning:
Two issues. We shouldn't be setting logical location index to 0 in cases where we can't find the logical location in the index map (we should be setting -1 in this case or possibly erroring out). also, the conditional around decorated name doesn't make obvious sense.

                        _logicalLocationToIndexMap.TryGetValue(logicalLocation, out int index);

                        if (!string.IsNullOrEmpty(logicalLocation.DecoratedName))
                        {
                            logicalLocation.DecoratedName = v1Location.DecoratedName;
                            _logicalLocationToIndexMap[logicalLocation] = index;
                        }

                        location.LogicalLocationIndex = index;

Issue #1179, opened 2018-12-26
Bad test output in NotificationExceptionWithStack


michaelcfanning:
Indicates a serious issue resolving logical location indices. Logical location index in test output for ExecutionEngine.Engine.TestTarget s/be 2 not zero.

PR #1185, opened 2018-12-27
Logical locations notes

michaelcfanning:
Resolves:

#1183: In which we didn't set -1 properly as logical location index in case of absent logical location. This code also mishandled flowing the v1 location.decoratedName to v1 logicalLocation.decoratedName

#1182: The definitive fully qualified name for a v1 result refers to location.fullyQualifiedName first then, if absent, consults location.logicalLocationKey. The definitive logical location dictionary key reverses this process. 

#1179: In which we failed to set logical location index properly for constructed stack frames (one of three v1 constructs, along with locations and annotated locations, that reference logical locations).

@lgolding

### DIFF
--- a/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
+++ b/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
@@ -22,6 +22,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
         private Run _currentRun;
         private RunVersionOne _currentV1Run;
         private int _threadFlowLocationNestingLevel;
+        private IDictionary<string, string> _v1KeyToFullyQualifiedNameMap;
         private Dictionary<LogicalLocation, int> _logicalLocationToIndexMap;
         private Dictionary<string, LogicalLocation> _v1KeyToV2LogicalLocationMap;
 
@@ -336,7 +337,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             {
                 location = new Location
                 {
-                    FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
+                    FullyQualifiedLogicalName = v1Location.FullyQualifiedLogicalName ?? v1Location.LogicalLocationKey,
                     PhysicalLocation = CreatePhysicalLocation(v1Location.ResultFile),
                     Properties = v1Location.Properties
                 };
@@ -381,9 +382,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                     Properties = v1AnnotatedCodeLocation.Properties
                 };
 
-                if (!string.IsNullOrWhiteSpace(location.FullyQualifiedLogicalName))
+                string logicalLocationKey = v1AnnotatedCodeLocation.LogicalLocationKey ?? v1AnnotatedCodeLocation.FullyQualifiedLogicalName;
+
+                if (!string.IsNullOrWhiteSpace(logicalLocationKey))
                 {
-                    if (_v1KeyToV2LogicalLocationMap.TryGetValue(location.FullyQualifiedLogicalName, out LogicalLocation logicalLocation))
+                    if (_v1KeyToV2LogicalLocationMap.TryGetValue(logicalLocationKey, out LogicalLocation logicalLocation))
                     {
                         _logicalLocationToIndexMap.TryGetValue(logicalLocation, out int index);
                         location.LogicalLocationIndex = index;
@@ -429,19 +432,21 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                 Message = CreateMessage(message)
             };
 
-            LogicalLocation logicalLocation = new LogicalLocation();
+            logicalLocationKey = logicalLocationKey ?? fullyQualifiedLogicalName;
 
-            logicalLocation.FullyQualifiedName = fullyQualifiedLogicalName;
-            logicalLocation.Name = GetLogicalLocationName(fullyQualifiedLogicalName);
-            location.FullyQualifiedLogicalName = logicalLocationKey;
+            // Retrieve logical location so that we can acquire the index
+            LogicalLocation logicalLocation;
+            _v1KeyToV2LogicalLocationMap.TryGetValue(logicalLocationKey, out logicalLocation);            
 
-            if (_logicalLocationToIndexMap.TryGetValue(logicalLocation, out int index))
+            location.FullyQualifiedLogicalName = fullyQualifiedLogicalName ?? logicalLocation?.FullyQualifiedName;
+
+            int logicalLocationIndex;
+            if (logicalLocation == null || !_logicalLocationToIndexMap.TryGetValue(logicalLocation, out logicalLocationIndex))
             {
-                index = _logicalLocationToIndexMap.Count;
-                _logicalLocationToIndexMap[logicalLocation] = index;
+                logicalLocationIndex = -1;
             }
 
-            location.LogicalLocationIndex = index;
+            location.LogicalLocationIndex = logicalLocationIndex;
 
             if (uri != null)
             {
@@ -479,17 +484,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             }
 
             return logicalLocation;
-        }
-
-        internal LogicalLocation CreateLogicalLocation(string fullyQualifiedLogicalName, string parentKey = null, string decoratedName = null, string kind = null)
-        {
-            return new LogicalLocation
-            {
-                DecoratedName = decoratedName,
-                FullyQualifiedName = fullyQualifiedLogicalName,
-                Name = GetLogicalLocationName(fullyQualifiedLogicalName),
-                Kind = kind
-            };
         }
 
         internal string GetLogicalLocationName(string fullyQualifiedLogicalName)
@@ -909,6 +903,16 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                         }
                     }
 
+                    // Pass 1 over results. In this phase, we're simply collecting fully qualified names that
+                    // may be duplicated in the logical locations dictionary. We're doing this so that we 
+                    // can properly construct the v2 logical instances in the converted array (i.e., we can't
+                    // populate the v2 logicalLocation.FullyQualifiedName property in cases where the 
+                    // v1 key is a synthesized value and not actually the fully qualified name)
+                    var visitor = new VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor();
+                    visitor.VisitRunVersionOne(v1Run);
+
+                    _v1KeyToFullyQualifiedNameMap = visitor.LogicalLocationKeyToFullyQualifiedNameMap;
+
                     if (v1Run.LogicalLocations != null)
                     {
                         run.LogicalLocations = new List<LogicalLocation>();
@@ -918,7 +922,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                         {
                             PopulateLogicalLocation(
                                 run, 
-                                v1Run.LogicalLocations, 
+                                v1Run.LogicalLocations,
+                                _v1KeyToFullyQualifiedNameMap,
                                 pair.Key, 
                                 pair.Value,
                                 populatedKeys);
@@ -965,23 +970,34 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 
         private void PopulateLogicalLocation(
             Run v2Run, 
-            IDictionary<string, LogicalLocationVersionOne> v1LogicalLocations, 
-            string fullyQualifiedName, 
+            IDictionary<string, LogicalLocationVersionOne> v1LogicalLocations,
+            IDictionary<string, string> keyToFullyQualifiedNameMap,
+            string logicalLocationKey, 
             LogicalLocationVersionOne v1LogicalLocation, 
             HashSet<string> populatedKeys)
         {
             // We saw and populated this one previously, because it was a parent to 
             // a logical location that we encountered earlier
-            if (populatedKeys.Contains(fullyQualifiedName)) { return; }
+            if (populatedKeys.Contains(logicalLocationKey)) { return; }
 
             if (v1LogicalLocation.ParentKey != null && !populatedKeys.Contains(v1LogicalLocation.ParentKey))
             {
                 // Ensure that any parent has been populated 
                 PopulateLogicalLocation(
                     v2Run, v1LogicalLocations,
+                    keyToFullyQualifiedNameMap,
                     v1LogicalLocation.ParentKey,
                     v1LogicalLocations[v1LogicalLocation.ParentKey],
                     populatedKeys);
+            }
+
+            string fullyQualifiedName;
+
+            if (!keyToFullyQualifiedNameMap.TryGetValue(logicalLocationKey, out fullyQualifiedName))
+            {
+                // If we don't find a remapping, the dictionary key itself comprises
+                // the fully qualified name.
+                fullyQualifiedName = logicalLocationKey;
             }
 
             // Create the logical location from the v1 version
@@ -993,11 +1009,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             // Store the old v1 look-up key for the new logical location
             // We will use this to generate the index when we walk results
             // v1 key -> logical location -> logical location index
-            _v1KeyToV2LogicalLocationMap[fullyQualifiedName] = logicalLocation;
+            _v1KeyToV2LogicalLocationMap[logicalLocationKey] = logicalLocation;
 
             v2Run.LogicalLocations.Add(logicalLocation);
 
-            populatedKeys.Add(fullyQualifiedName);
+            populatedKeys.Add(logicalLocationKey);
         }
 
         internal Stack CreateStack(StackVersionOne v1Stack)
--- a/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
+++ b/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
@@ -370,7 +370,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             Location location = null;
 
             if (v1AnnotatedCodeLocation != null)
-            {
+            {                
                 location = new Location
                 {
                     Annotations = v1AnnotatedCodeLocation.Annotations?.SelectMany(a => a.Locations,
@@ -379,7 +379,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                                                                                                          a.Message))
                                                                       .Where(r => r != null)
                                                                       .ToList(),
-                    FullyQualifiedLogicalName = v1AnnotatedCodeLocation.LogicalLocationKey ?? v1AnnotatedCodeLocation.FullyQualifiedLogicalName,
+                    FullyQualifiedLogicalName = v1AnnotatedCodeLocation.FullyQualifiedLogicalName,
                     Message = CreateMessage(v1AnnotatedCodeLocation.Message),
                     PhysicalLocation = CreatePhysicalLocation(v1AnnotatedCodeLocation.PhysicalLocation),
                     Properties = v1AnnotatedCodeLocation.Properties
--- a/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
+++ b/src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
@@ -25,7 +25,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
         private IDictionary<string, string> _v1KeyToFullyQualifiedNameMap;
         private IDictionary<LogicalLocation, int> _v2LogicalLocationToIndexMap;
         private IDictionary<string, LogicalLocation> _v1KeyToV2LogicalLocationMap;
-        private IDictionary<string, string> _v1fullyQualifiedNameToDecoratedNameMap;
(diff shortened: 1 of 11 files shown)


src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -336,7 +338,7 @@ internal Location CreateLocation(LocationVersionOne v1Location)
             {
                 location = new Location
                 {
-                    FullyQualifiedLogicalName = v1Location.LogicalLocationKey ?? v1Location.FullyQualifiedLogicalName,
+                    FullyQualifiedLogicalName = v1Location.FullyQualifiedLogicalName ?? v1Location.LogicalLocationKey,
```
lgolding: >FullyQualifiedLogicalName [](start = 59, length = 25)

This isn't right. According to the spec (I looked as far back as the original v1 "working draft" from which the TC started its work), `LogicalLocationKey` is never guaranteed to be a valid FQLN. It is only guaranteed to match one of the property names in the `run.logicalLocations` dictionary.

If `v1.FQLN` is missing, your only choice is to omit `v2.FQLN`, so the correct code here is:
```
location = new Location
{
    FullyQualifiedLogicalName = v1Location.FullyQualifiedLogicalName, // Works whether or not it's present in the v1 file
    ...
```

It doesn't even help to use the key to find the `logicalLocation` object in `run.logicalLocations` dictionary. In v1, a `logicalLocation` object doesn't have a `fullyQualifiedName` property, so you'd have to walk the parent chain and glue the FQLN together yourself (and you'd have to know the language syntax to do it).

So there's no good answer. Just copy the FQLN from v1 to v2 and call it a day.

lgolding: I think I might be contradicting what I commented in your previous PR. If so, my previous comment was wrong. You just can't get away with using a key as an FQLN.

---
In reply to: [244237410](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244237410) [](ancestors = 244237410)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -381,11 +383,13 @@ internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCode
                     Properties = v1AnnotatedCodeLocation.Properties
                 };
 
-                if (!string.IsNullOrWhiteSpace(location.FullyQualifiedLogicalName))
+                string logicalLocationKey = v1AnnotatedCodeLocation.LogicalLocationKey ?? v1AnnotatedCodeLocation.FullyQualifiedLogicalName;
```
lgolding: >logicalLocationKey [](start = 23, length = 18)

Yes, good.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -429,19 +433,21 @@ internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCode
                 Message = CreateMessage(message)
             };
 
-            LogicalLocation logicalLocation = new LogicalLocation();
+            logicalLocationKey = logicalLocationKey ?? fullyQualifiedLogicalName;
```
lgolding: >logicalLocationKey [](start = 12, length = 18)

Hold on there, bucko :-)

If FQLN is present but key is absent, then the lookup on Line 440 isn't guaranteed to succeed. `run.logicalLocation` might not even exist. The `logicalLocation` will be `null`, and from Line 442 on it will all end in tears.

You have to check the result of the lookup, and if it fails, synthesize your own `LogicalLocation` object as you did on Line 432 of the left-hand side.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -429,19 +433,21 @@ internal Location CreateLocation(AnnotatedCodeLocationVersionOne v1AnnotatedCode
...
+            location.FullyQualifiedLogicalName = fullyQualifiedLogicalName ?? logicalLocation?.FullyQualifiedName;
 
-            if (_logicalLocationToIndexMap.TryGetValue(logicalLocation, out int index))
+            int logicalLocationIndex;
+            if (logicalLocation == null || !_v2LogicalLocationToIndexMap.TryGetValue(logicalLocation, out logicalLocationIndex))
             {
-                index = _logicalLocationToIndexMap.Count;
-                _logicalLocationToIndexMap[logicalLocation] = index;
+                logicalLocationIndex = -1;
             }
```
lgolding: >} [](start = 12, length = 1)

This is right but I wonder if this is clearer:
```
int logicalLocationIndex = -1;
if (logicalLocation != null && _v2LogicalLocationToIndexMap.TryGetValue(logicalLocation, out int logicalLocationIndexFromMap)))
{
    logicalLocationIndex = logicalLocationIndexFromMap;
}
```

lgolding: Why I think it's clearer:
1. Doesn't require an uninitialized variable.
2. Doesn't use a temporary, and so...
3. One line shorter.
4. `if` statement uses "positive logic".

YMMV.

---
In reply to: [244239471](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244239471) [](ancestors = 244239471)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    /// <summary>
+    /// This class visits all locations in a SARIF v1 log file. On observing use of result.logicalLocationKey,
+    /// which is used to disambiguate logical locations in the result.logicalLocations dictionary that have a 
+    /// common fully qualified name (but which are different types), the visitor creates a mapping between
+    /// the logical location key and its associated fully qualified name. This allows the v2 transformation
+    /// in particular to more easily populate its logical location equivalents. Additionally, the visitor 
+    /// stores a mapping from fully qualified name to decorated name, if one exists. These data (the
```
lgolding: >exists [](start = 77, length = 6)

This is the thing I worried about in `SarifVersionOneToCurrentVisitor.cs`. I would expect a mapping from key to decorated name, not from FQLN to decorated name, since there could be two logical locations with the same FQLN but different decorated names. In fact, come to think of it, one of the test files has this situation.

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// object in v2.
+    /// </summary>
+    public class VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor : SarifRewritingVisitorVersionOne
+    {
+        public VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor()
+        {
+            FullyQualifiedNameToDecoratedNameMap = new Dictionary<string, string>();
+            LogicalLocationKeyToFullyQualifiedNameMap = new Dictionary<string, string>();
+        }
+        public IDictionary<string, string> FullyQualifiedNameToDecoratedNameMap { get; set; }
```
lgolding: >set [](start = 87, length = 3)

Setters not necessary for these two properties.

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public IDictionary<string, string> LogicalLocationKeyToFullyQualifiedNameMap { get; set; }
+
+        public override LocationVersionOne VisitLocationVersionOne(LocationVersionOne node)
+        {
+            if (!string.IsNullOrEmpty(node.LogicalLocationKey) &&
+                !string.IsNullOrEmpty(node.FullyQualifiedLogicalName) &&
+                !node.FullyQualifiedLogicalName.Equals(node.LogicalLocationKey))
```
lgolding: >( [](start = 54, length = 1)

`StringComparison.Ordinal`.

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                LogicalLocationKeyToFullyQualifiedNameMap[node.LogicalLocationKey] = node.FullyQualifiedLogicalName;
+            }
+
+            if (!string.IsNullOrEmpty(node.DecoratedName))
+            {
+                string fullyQualifiedName = node.FullyQualifiedLogicalName ?? node.LogicalLocationKey;
```
lgolding: >LogicalLocationKey [](start = 83, length = 18)

Just use the key as the index, for reason given above.

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
```diff
@@ -0,0 +1,77 @@
...
+        public override AnnotatedCodeLocationVersionOne VisitAnnotatedCodeLocationVersionOne(AnnotatedCodeLocationVersionOne node)
+        {
+            if (!string.IsNullOrEmpty(node.LogicalLocationKey) &&
+                !string.IsNullOrEmpty(node.FullyQualifiedLogicalName) &&
+                !node.FullyQualifiedLogicalName.Equals(node.LogicalLocationKey))
+            {
+                LogicalLocationKeyToFullyQualifiedNameMap[node.LogicalLocationKey] = node.FullyQualifiedLogicalName;
+            }
+
+            // v1 stack frame does not reference a decorated name
```
lgolding: >stack frame [](start = 18, length = 11)

annotated code location.

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
lgolding: Crud. This test occurs three times, but you can't DRY it out to `static bool NodeHasAmbiguousLogicalLocation` because the node is of three different types. Still yet another validation of the design improvement where we moved these properties into the `Location` object.

lgolding:
Two substantive bugs and several nits.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: I like where you're taking this. However, you neglect a key point, when TryGetValue fails, 'logicalLocationIndexFromMap' is initialized to the default of 0, which is not the actual default for logical location index (it is -1 as you know).

And so another subtlety in the format, our default values that do not correspond to C# primitive defaults require additional work in cases like this.

---
In reply to: [244239715](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244239715) [](ancestors = 244239715,244239471)

lgolding: Oh, right you are.

---
In reply to: [244372604](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244372604) [](ancestors = 244372604,244239715,244239471)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -918,7 +927,9 @@ internal Run CreateRun(RunVersionOne v1Run)
                         {
                             PopulateLogicalLocation(
                                 run, 
-                                v1Run.LogicalLocations, 
+                                v1Run.LogicalLocations,
+                                _v1fullyQualifiedNameToDecoratedNameMap,
+                                _v1KeyToFullyQualifiedNameMap,
```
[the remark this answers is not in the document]
michaelcfanning: This is a constantly confounding question. In general, I try to parameterize method calls with data that is relevant to its general operation. The use of instance data within a helper of this kind should represent data that 'slips' out of the core function. IOW, we are either utilizing some global initialization set previously (which is configurable) or we are opportunistically preserving some data that will be used later. 

Realistically, it's hard to maintain a mental model of the code to know that these 'rules' are being followed. Also, this kind of coding in general complicates unit testability, requiring creation/initialization in some cases to test isolated helpers. A more predictably testable pattern is to author as many helpers as possible as static methods. The downside here, of course, is that if you do a good job factoring methods into small pieces, you may find a need to 'plumb' instance members through several layers in order to make it available to some low level helper.

Anyway, come in soon and have lunch (on me) and we'll discuss this specific topic thoroughly. It's been on my mind for precisely the reason you added this comment here.

---
In reply to: [244240108](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240108) [](ancestors = 244240108)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
lgolding: Sorry, that comment is wrong. But so it the code. :-) You're trying to figure out what to use as the key, so the variable name should be `key`, and the code is:
```
string key = node.LogicalLocationKey ?? node.FullyQualifiedLogicalName;
KeyToDecoratedNameMap[key] = node.DecoratedName;
```


---
In reply to: [244240704](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240704) [](ancestors = 244240704)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -909,6 +907,17 @@ internal Run CreateRun(RunVersionOne v1Run)
...
+                    // Pass 1 over results. In this phase, we're simply collecting fully qualified names that
+                    // may be duplicated in the logical locations dictionary. We're doing this so that we 
+                    // can properly construct the v2 logical instances in the converted array (i.e., we can't
+                    // populate the v2 logicalLocation.FullyQualifiedName property in cases where the 
+                    // v1 key is a synthesized value and not actually the fully qualified name)
+                    var visitor = new VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor();
+                    visitor.VisitRunVersionOne(v1Run);
+
+                    _v1KeyToFullyQualifiedNameMap = visitor.LogicalLocationKeyToFullyQualifiedNameMap;
+                    _v1fullyQualifiedNameToDecoratedNameMap = visitor.FullyQualifiedNameToDecoratedNameMap;
```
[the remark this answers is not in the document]
michaelcfanning: nice catch, thanks.

---
In reply to: [244239959](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244239959) [](ancestors = 244239959)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
michaelcfanning: Yes, this contradicts what we discussed previously but what you say here makes perfect sense.

---
In reply to: [244237488](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244237488) [](ancestors = 244237488,244237410)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
michaelcfanning: Very nice catch. The PREfast samples have this condition. I believe all these samples, however, have been converted to pre-release v2. I will add a test for this case.

---
In reply to: [244240540](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240540) [](ancestors = 244240540)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
michaelcfanning: Indeed. Almost left a comment for you on precisely this point (my inability to DRY due to the 'duck typing' of our previous design that inlined FQN and logical location keys. :)

---
In reply to: [244240968](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240968) [](ancestors = 244240968)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
lgolding: (This is the fix to the bug I was worried about.)

---
In reply to: [244373857](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244373857) [](ancestors = 244373857,244240704)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
michaelcfanning: This comparison (string.Equals) is one of the places in .NET where Ordinal is the default. This is true for other primitives such as the numerics. I don't make this explicit in code (for reasons of readability and maintainability) because I consider it a sort of essential fact about .NET that is helpful to know. (just as I suggest to people that they avoid unnecessary initialization of instance members). YMMV. :)

[https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings)
"The most common methods that test for equality, String.Equals and String.Equality, use a case-sensitive ordinal comparison."

---
In reply to: [244240657](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240657) [](ancestors = 244240657)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
michaelcfanning: Your statement isn't quite right, I mention the intent in the class comment. This code only works in cases where disambiguation is created, i.e., when a logical location key is used to resolve a collision. We don't need to persist all the other cases, because generally we have our hands on a fully qualified name already on those instances. Still open to making a change here if you see one.

---
In reply to: [244373857](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244373857) [](ancestors = 244373857,244240704)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
michaelcfanning: cut-n-paste ftw

---
In reply to: [244240816](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244240816) [](ancestors = 244240816)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
lgolding: [I was not aware of that.](https://www.youtube.com/watch?v=Cd7tEsAuspA)

---
In reply to: [244376530](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244376530) [](ancestors = 244376530,244240657)

src/Sarif/Visitors/VersionOneLogicalLocationKeyToFullyQualifiedNameMappingVisitor.cs
lgolding: I still don't think this is right. You say that this code works "when a logical location key is used to resolve a collision." So consider the case:

```
location1 = {
    "fullyQualifiedLogicalName": "ambiguousFqln",
    "logicalLocationKey": "key1",
    "decoratedName": "dn1"
};

location2  = {
    "fullyQualifiedLogicalName": "ambiguousFqln",
    "logicalLocationKey": "key2",
    "decoratedName": "dn2"
};
```

`fullyQualifiedLogicalName` is the same, and non-null, for both locations. So when you visit `location1`, you will execute.
```
FullyQualifiedNameToDecoratedNameMap["ambiguousFqln"] = "dn1";
```
... and when you visit `location2`, you will execute
```
FullyQualifiedNameToDecoratedNameMap["ambiguousFqln"] = "dn2";
```
... losing `dn1` entirely. With my suggested code, you will execute
```
KeyNameToDecoratedNameMap["key1"] = "dn1";
...
KeyNameToDecoratedNameMap["key2"] = "dn2";
```

... which is what you need.

---
In reply to: [244377445](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244377445) [](ancestors = 244377445,244373857,244240704)

michaelcfanning: I'd already fixed this. I was confused by your statement in stand-up, thought we were discussing the other mapping scenario (to retrieve FQN in cases where there is a key collision)

---
In reply to: [244383433](https://github.com/Microsoft/sarif-sdk/pull/1185#discussion_r244383433) [](ancestors = 244383433,244377445,244373857,244240704)

michaelcfanning:
@lgolding thanks for your diligence cleaning this area up. I explicitly added tests that exhaustively test v1 key collisions for stack frames, annotated code locations, and locations. Everything now working.

PR #1186, opened 2018-12-27
Add fileIndex property to file object

### LINKED ISSUE #1177: Emit DefaultValue and JsonProperty values via code gen hints
And remove any files we can from the NotYetAutogenerated location. Realistically, we need to leave these files behind now, however, because they include some initialization in ctors that can't be accomplished via a code gen hint or jschema's code emit.

michaelcfanning:
@lgolding. Resolves #1177 to use code gen for logicalLocationIndex attribute hints. Also adds the new fileIndex property only, as part of work to convert files object from a dictionary to an array.


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -468,6 +468,13 @@
           "type": "string"
         },
 
+        "fileIndex": {
+          "description": "The index within the run files array that specifies the file object associated with the file location.",
```
lgolding: >that specifies [](start = 63, length = 14)

"that specifies" => "of"

lgolding:
Just once suggested wording change.

PR #1191, opened 2018-12-29
Reduced files array build

michaelcfanning:
@lgolding @EasyRhinoMSFT 

Trying to factor this large change into pieces. As a result, the overall build will not function. Tests will not pass. Once I get the overall build functioning, test updates will start to come online and new tests.

This first change restores compilation of the Sarif and Sarif.Converters binaries. It demonstrates a core pattern for working with file data objects: there is a helper on the Run object that will retrieve the requested file data object on the file location which serves as the effective key. This helper understands that it should only use the Uri and UriBaseId members to locate the associated file data object. This helper opportunistically updates the file location instance passed to it, populating its fileIndex property and also normalizing the Uri representation.

[btw - this approach of placing a helper on the run object may be useful for logical locations as well, will ponder this.]

While dropping in required updates to the converters code, I finally was compelled to DRY out all the shared results writing code into the base class. Having completed this, there is sufficient work to review.

Finally, as is inevitable, I ran into some existing rough edges that are either fixed or documented in bugs. We have a long-standing desire to zap IRule completely (it only exists due to being entangled with other interfaces backing the Driver extensibility model). IResultLogWriter has a bug, WriteToolNotifications s/be replaced by WriteInvocations (the invocations object now contains both tool and configuration notifications). The FortifyFpr converter attempts to persist tool notifications. I suspect successfully traveling this code path today would produce invalid SARIF, though (since the writer hangs the data off the run object.

@EasyRhinoMSFT, your attention particularly appreciated on the FPR converter work. Do we have a data file that contains tool notifications we can test?


src/Sarif/Sarif.csproj
```diff
@@ -24,7 +24,15 @@
...
 
   <ItemGroup>
+    <Compile Remove="Baseline\**" />
     <Compile Remove="NotYetAutoGenerated\**" />
+    <EmbeddedResource Remove="Baseline\**" />
+    <None Remove="Baseline\**" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="Visitors\SarifCurrentToVersionOneVisitor.cs" />
```
michaelcfanning: >SarifCurrentToVersionOneVisitor [](start = 30, length = 31)

Temporarily removed so I can isolate the v1/v2 transformation into a discrete change

src/Sarif/SarifUtilities.cs
```diff
@@ -207,5 +208,22 @@ public static int GetByteLength(string s, Encoding encoding)
             byte[] bytes = encoding.GetBytes(s);
             return bytes.Length;
         }
+
+        internal static bool UnitTesting = false;
+
+        internal static void DebugAssert(bool conditional)
```
michaelcfanning: >DebugAssert [](start = 29, length = 11)

Random thing I'm throwing out there. Debug.Assert statements are great, except that they aren't handled well while unit testing (the tests resolve as inconclusive). This pattern would converter a failed assert into an exception, assuming tests set the UnitTesting conditional. Note that this would also allow us to effectively inject the assert into retail builds (Debug.Assert will compile away entirely of course). This is a nice value, unit testing can provoke some assertion coverage in release testing that would otherwise never get executed.

Want to start a general conversation on this, as we'll all want to be in sync on this approach assuming we proceed.

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -70,30 +69,14 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
             };
 
-            var fileInfoFactory = new FileInfoFactory(null, dataToInsert);
-            Dictionary<string, FileData> fileDictionary = fileInfoFactory.Create(results);
-
             var run = new Run()
             {
-                Tool = tool,
-                ColumnKind = ColumnKind.Utf16CodeUnits
+                Tool = new Tool { Name = ToolFormat.AndroidStudio },
```
lgolding: >Tool [](start = 27, length = 4)

Did you also mean to remove the local variable `tool`?

src/Sarif.Converters/ToolFileConverterBase.cs
```diff
@@ -75,5 +75,54 @@ internal static string GetLogicalLocationName(string parentKey, string fullyQual
 
             return logicalName;
         }
+
+        protected static Run PersistResults(IResultLogWriter output, IList<Result> results, string toolName)
+        {
+            var run = new Run()
+            {
+                Tool = new Tool { Name = toolName },
+                ColumnKind = ColumnKind.Utf16CodeUnits
```
lgolding: >ColumnKind.Utf16CodeUnits [](start = 29, length = 25)

Since this occurs several times, maybe there's a good place for a global constant `ColumnKind DefaultColumnKind = ColumnKind.Utf16CodeUnits`.

src/Sarif.Converters/ToolFileConverterBase.cs
```diff
@@ -75,5 +75,54 @@ internal static string GetLogicalLocationName(string parentKey, string fullyQual
...
+            {
+                output.WriteLogicalLocations(run.LogicalLocations);
+            }
+
+            output.OpenResults();
+            output.WriteResults(run.Results);
+            output.CloseResults();
+
+            // TODO: should we move these to be emitted at the beginning of the file? What order fo we prefer generaally
+            if (run.Resources?.Rules != null)
```
lgolding: >if [](start = 12, length = 2)

A tool that emits results as it produces them would naturally put `files` and `logicalLocations` at the end, because doesn't know until the end the set of locations it will encounter. Such a tool would also put `resources` at the end if it chose to emit only those rules that actually triggered and those messages it actually emitted; otherwise it could put `resources` anywhere.

Also, a user probably wants to get to the results ASAP, without paging through dozens thousands of files before getting to the good stuff.

So IMO all of `files`, `logicalLocations`, and `resources` should go after `results`.

src/Sarif.Converters/ClangAnalyzerConverter.cs
```diff
@@ -58,27 +57,7 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
-                };
-
-                output.Initialize(run);
-
-                if (fileDictionary != null && fileDictionary.Count > 0) { output.WriteFiles(fileDictionary); }
-
-                output.OpenResults();
-                output.WriteResults(results);
-                output.CloseResults();
+                PersistResults(output, results, ToolFormat.ClangAnalyzer);
```
lgolding: >ClangAnalyzer [](start = 59, length = 13)

I don't think you should pass a `ToolFormat` to the `toolName` parameter. There are tools like FxCop and Fortify that can produce multiple formats. I think this confuses two concepts, the tool and its output format.

src/Sarif.Converters/FortifyConverter.cs
```diff
@@ -102,16 +93,10 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
                         Text = runDescription
                     }
                 },
-                Tool = tool
+                Tool = new Tool {  Name = "Fortify" }
```
lgolding: >Fortify [](start = 43, length = 7)

Right, here's a case where you can't use the format for the tool name, because there are two of them. Since you can't do it always, I think you should never do it; it's more or less an accident if the format name happens to match the tool name. Besides, the tool's preferred name might not even be a valid identifier (_e.g._, "Clang Analyzer").

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -116,7 +114,8 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
                     InstanceId = _automationId + "/"
                 },
                 Tool = tool,
-                Invocations = new[] { _invocation }
+                Invocations = new[] { _invocation },
+                Resources = new Resources {  Rules = _ruleDictionary}
```
lgolding: >Rules [](start = 45, length = 5)

Doesn't this mean you'll emit
```
"resources": {
  "rules": {}
}
```
if the dictionary is empty, rather than omitting the "rules" object? (In fact, rather than omitting `"resources"` entirely?) How about:
```
var run = new Run
{
    ...
};

if (_ruleDictionary.Any())
{
    run.Resources = new Resources { Rules = _ruleDictionary };
}
```

src/Sarif.Converters/ToolFileConverterBase.cs
```diff
@@ -75,5 +75,54 @@ internal static string GetLogicalLocationName(string parentKey, string fullyQual
 
             return logicalName;
         }
+
+        protected static Run PersistResults(IResultLogWriter output, IList<Result> results, string toolName)
```
lgolding: >PersistResults [](start = 29, length = 14)

A big +1 for DRYing this out.

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -53,46 +52,27 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
-            }
-
-            output.OpenResults();
-            output.WriteResults(results);
-            output.CloseResults();
-
             if (rules.Count > 0)
             {
-                IDictionary<string, IRule> rulesDictionary = new Dictionary<string, IRule>();
+                IDictionary<string, Rule> rulesDictionary = new Dictionary<string, Rule>();
```
lgolding: >Rule [](start = 36, length = 4)

Yes! Step by step...

src/Sarif.Converters/PREFastConverter.cs
```diff
@@ -57,22 +43,15 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
-                var fileInfoFactory = new FileInfoFactory(MimeType.DetermineFromFileExtension, dataToInsert);
-                Dictionary<string, FileData> fileDictionary = fileInfoFactory.Create(results);
-
-                if (fileDictionary?.Any() == true)
+                var run = new Run()
                 {
-                    output.WriteFiles(fileDictionary);
-                }
+                    Tool = new Tool { Name = ToolFormat.PREfast, FullName = "PREfast Code Analysis" },
+                    ColumnKind = ColumnKind.Utf16CodeUnits
```
lgolding: >ColumnKind = ColumnKind.Utf16CodeUnits [](start = 20, length = 38)

You could DRY this out by adding
```
static Run Create()
{
    return new Run { ColumnKind = ColumnKind.Utf16CodeUnits; }
}
```
... to the Run.cs partial class in Sarif\Core.

lgolding: ... It's true that then you couldn't use property initialization syntax to fill the rest of the properties, so... whatever you like.

---
In reply to: [244521253](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521253) [](ancestors = 244521253)

src/Sarif.Converters/PREFastConverter.cs
```diff
@@ -57,22 +43,15 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
+                };
 
-                if (LogicalLocations != null && LogicalLocations.Any())
-                {
-                    output.WriteLogicalLocations(LogicalLocations);
-                }
+                run.LogicalLocations = LogicalLocations;
```
lgolding: >LogicalLocations [](start = 39, length = 16)

Any reason not to include this in the property initialization syntax above?

src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -82,7 +79,18 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
-			output.WriteFiles(fileDictionary);
+            var visitor = new AddFileReferencesVisitor();
+            visitor.VisitRun(run);
+
+            foreach (Result result in results)
+            {
+                visitor.VisitResult(result);
+            }
+
+            if (run.Files != null && run.Files.Count > 0)
```
lgolding: >run.Files != null && run.Files.Count > 0 [](start = 16, length = 40)

`if (run.Files?.Count > 0)` or `if (run.Files?.Any())` (my favorite expression for this)...

... but the base class already does this.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -40,17 +40,25 @@ public class SarifLogger : IDisposable, IAnalysisLogger
...
-                    run.Files[fileDataKey] = fileData;
+                    var fileLocation = new FileLocation
+                    {
+                        Uri = uri
+                    };
+
+                    fileData.FileLocation = fileLocation;
+
+                    // This call will insert the file object into run.Files if not already present
+                    fileData.FileLocation.FileIndex = run.GetFileIndex(fileData.FileLocation, addToFilesTableIfNotPresent: true);
```
lgolding: >addToFilesTableIfNotPresent [](start = 94, length = 27)

Could we avoid this by starting out with a hash set? Well, I suppose that won't work because our ValueComparer includes `fileIndex`...

Hey! I could add a new type of code gen hint that would let you omit properties from the value comparer, what do you think?

src/Sarif/Core/Message.cs
```diff
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Linq;
```
lgolding: >Linq [](start = 13, length = 4)

Always happy to see this. Ultra nit: extra blank line.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1543,8 +1543,9 @@
 
         "files": {
           "description": "An array of file objects relevant to the run.",
```
lgolding: >array [](start = 29, length = 5)

Hah! The comment _used_ to be wrong and is now right.

src/Sarif/Sarif.csproj
lgolding: Ok.

---
In reply to: [244507727](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244507727) [](ancestors = 244507727)

src/Sarif/Core/Run.cs
```diff
@@ -32,13 +136,13 @@ public bool ShouldSerializeColumnKind()
             return true;
         }
 
-        public bool ShouldSerializeFiles() { return this.Files != null && this.Files.Values.Any(); }
+        public bool ShouldSerializeFiles() { return this.Files.Any(); }
```
lgolding: >. [](start = 62, length = 1)

Here and on Line 145, you're assuming that these array-valued properties will be non-null (that is, that they will always be present, even if empty). That means that we'll unnecessarily serialize empty arrays, right? Shouldn't we let them be null, and the use `this.TheArrayValuedProperty?.Any()`?

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
         private static Invocation EmptyInvocation = new Invocation();
         private static LogicalLocation EmptyLogicalLocation = new LogicalLocation();
 
+        private IDictionary<FileLocation, int> _fileToIndexMap;
+
+        public int GetFileIndex(
+            FileLocation fileLocation, 
+            bool addToFilesTableIfNotPresent = true,
+            OptionallyEmittedData dataToInsert = OptionallyEmittedData.None,
+            System.Text.Encoding encoding = null)
```
lgolding: >System.Text [](start = 12, length = 11)

Just curious how you decide when to inline this and when to add a `using`. I suppose your criterion is "only used once." And maybe that's better than my practice of always adding a `using`. It fits with the goal of reducing cognitive load ("Where is `Encoding` defined?")

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+        {
+            if (fileLocation == null) throw new ArgumentNullException(nameof(fileLocation));
```
lgolding: >throw [](start = 38, length = 5)

I am absolutely fine with this construct, but I think you've told me before that you prefer braces even if it's a single statement on the same line.

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+
+            if (this.Files == null || this.Files.Count == 0)
```
lgolding: >this.Files == null || this.Files.Count == 0 [](start = 16, length = 43)

`this.Files?.Any()`... BUT! If `Files` is missing and `addToFilesTableIfNotPresent` is `true`, then wouldn't this be a good time to create `Files`?

lgolding: Well, I guess I see what you're after -- if the producer doesn't initialize `run.Files` with the list of scanned files, maybe that means the producer didn't _want_ a files array, so we shouldn't force it on him here. But in that case, maybe the producer just shouldn't call this method?

---
In reply to: [244521702](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521702) [](ancestors = 244521702)

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+                var fileLocation = new FileLocation
+                {
+                    Uri = fileData.FileLocation.Uri,
+                    UriBaseId = fileData.FileLocation.UriBaseId
+                };
+
+                _fileToIndexMap[fileLocation] = i;
+
+                // For good measure, we'll explicitly populate the file index property
+                this.Files[i].FileLocation.FileIndex = i;
```
lgolding: >FileIndex [](start = 43, length = 9)

This could be in the property initializer list.

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -21,41 +20,18 @@ public TSLintConverter()
         
         public override void Convert(Stream input, IResultLogWriter output, OptionallyEmittedData dataToInsert)
         {
-            input = input ?? throw new ArgumentNullException(nameof(input));
-
+            input = input   ?? throw new ArgumentNullException(nameof(input));
```
lgolding: >    [](start = 25, length = 3)

Accidental extra space?

"throw expression" FTW. Nice language feature.

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+            // all cases. For now, we cover the Windows happy path, which assumes that
+            // most URIs in log files are file paths (which are case-insensitive)
+            //
+            // Tracking item for an improved comparer:
+            // https://github.com/Microsoft/sarif-sdk/issues/973
+
+            // When we perform a files table look-up, only the uri and uriBaseId
+            // are relevant; these properties together comprise the unique identity
+            // of the file object. The file index, of course, does not relate to the
+            // file identity. We consciously exclude the properties as well.
```
lgolding: >. We consciously exclude the properties as well [](start = 28, length = 47)

"..., so we exclude it" ? Or "..., so we exclude that property?" Not sure what to make of the plural here.

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+            };
+
+            int fileIndex;
+            if (!_fileToIndexMap.TryGetValue(filesTableKey, out fileIndex))
+            {
+                if (addToFilesTableIfNotPresent)
+                {
+                    fileIndex = this.Files.Count;
+
+                    string mimeType = Writers.MimeType.DetermineFromFileExtension(filesTableKey.Uri.ToString());
```
lgolding: >ToString() [](start = 100, length = 10)

`Uri.AbsolutePath` is probably faster, and makes it clear where you're getting the extension from.

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+                        mimeType: mimeType,
+                        encoding);
+
+                    fileData.FileLocation = fileLocation;
+
+                    this.Files.Add(new FileData
+                    {
+                        MimeType = mimeType,
+                        FileLocation = filesTableKey
+                    });
```
lgolding: Shouldn't you be adding the `var fileData` that you just created?

src/Sarif/Core/Run.cs
```diff
@@ -13,6 +14,109 @@ public partial class Run
...
+                }
+                else
+                {
+                    // We did not find the item. The call was not configured to add the entry.
+                    // Return the default value that indicates the item isn't present.
+                    fileIndex = -1;
+                }
+            }
+
+            fileLocation.FileIndex = fileIndex;
```
lgolding: >fileIndex [](start = 37, length = 9)

If you didn't enter the `if` block on Line 62, then this is uninitialized. I'm surprised this compiles.

src/Sarif/SarifUtilities.cs
lgolding: I like the idea of getting better unit test results. I suggest adding a `message` or `reason` string argument. You can use it as the message argument to `InvalidOperartionException` and as the second argument to `Debug.Assert`.

---
In reply to: [244507788](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244507788) [](ancestors = 244507788)

lgolding:
I had several comments, but nothing that I'd push back on whatever you decide (except for that whole "tool format" _vs._ "tool name" thing -- I feel pretty strongly about that. So, "approved", and do as ye will.

src/Sarif.Converters/ToolFileConverterBase.cs
michaelcfanning: Just to be clear, we're only discussing the converters scenario, not tools that emits results as they produce them. Now that we've DRYed out this code, we have the possibility of making all converter behaviors consistent. I'll put everything afterwards as you say. 

---
In reply to: [244520670](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244520670) [](ancestors = 244520670)

michaelcfanning: but btw one consequence of this is that our next converter update will produce wildly divergent content that will be impossible to diff effectively manually. Even if you are Laurence Golding, esq.

---
In reply to: [244545376](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244545376) [](ancestors = 244545376,244520670)

src/Sarif/SarifUtilities.cs
michaelcfanning: A fantastic idea. 

---
In reply to: [244542519](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244542519) [](ancestors = 244542519,244507788)

src/Sarif.Converters/FortifyFprConverter.cs
michaelcfanning: our deserialization logic can elide the resources object in cases where its contained strings and rules are empty, is my thought. this keeps the initialization logic in one place. eventually, we should consider having the OM initialize core objects automatically, to make object creation more concise. will require building out the ShouldSerializeXXX helpers exhaustively. All this is one reason I built that helper that automatically hydrates the OM with all possible items. This will help comprehensively test the helpers described above.

---
In reply to: [244521205](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521205) [](ancestors = 244521205)

src/Sarif.Converters/PREFastConverter.cs
michaelcfanning: We should make a pass over converter run creation in the future to ensure we fully hydrate things like full name.


---
In reply to: [244521261](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521261) [](ancestors = 244521261,244521253)

src/Sarif.Converters/SemmleQlConverter.cs
michaelcfanning: run.Files.Any() works because the extension handles the null 'this' case. still, this is unnecessary as you've pointed out, we should be calling the base helper. Simply failed to DRY here.

---
In reply to: [244521298](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521298) [](ancestors = 244521298)

src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -67,9 +67,6 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
 
             var results = GetResultsFromStream(input);
 
-			var fileInfoFactory = new FileInfoFactory(MimeType.DetermineFromFileExtension, dataToInsert);
-			Dictionary<string, FileData> fileDictionary = fileInfoFactory.Create(results);
-
             var tool = new Tool
             {
                 Name = "Semmle QL"
```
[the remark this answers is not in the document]
michaelcfanning: 'Pons' means bridge, 'facere' means 'to make' and so a pontificator is a bridge builder. YMMV being on the receiving end of it.

---
In reply to: [244521397](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521397) [](ancestors = 244521397)

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: I don't see how your comment is relevant here. 'addToFilesTableIfNotPresent' simply tells the underlying helper whether we are simply querying for the existing index or if it should be added. That underlying code maintains a dictionary of key to the relevant index. So not sure how a hash set helps.

However, yes, the general improvement you're pointing to is quite interesting: how do we capture relevant identity semantics in our complex format in a way that's broadly helpful. What we need here is a special value comparer that's used strictly for the files table. The code gen could then usefully emit a backing Dictionary<FileLocation, int> item that we'd consult in order to manage/populate this thing. We could perform similar tricks for rules and logical locations table (that latter table should have a unique array as well, it's a bug IMO that it isn't expressed this way today)

---
In reply to: [244521472](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521472) [](ancestors = 244521472)

src/Sarif/Core/Run.cs
michaelcfanning: hm, my assumption may be wrong on looking at relevant source. will test and update appropriately. handling a null 'this' pointer is a cute trick but non-obvious is you're not tracking that the relevant method is an extension method. so it may be verboten in CLR API as breaking principle of least surprise. will follow up.

---
In reply to: [244604442](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244604442) [](ancestors = 244604442,244521595)

src/Sarif/Core/Run.cs
michaelcfanning: My personal rule: inlined fully qualified type name is permissible when used a single time or when required to resolve a type name collision. A using statement is fine as well (except in cases when it requires creating an alias, in which case I generally prefer the fully qualified name).

Somewhere, there is a collision with the 'Encoding' term. I assumed I'd provoke it here but actually did not. Reverted to 'using.

---
In reply to: [244521675](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521675) [](ancestors = 244521675)

src/Sarif/Core/Run.cs
michaelcfanning: it's not my preference, it's my attempt to create a standard that balances the preferences of many. 

i.e., i do not set the coding guidelines in this code base according to my own preferences but instead attempt to help negotiate a set of standards that usefully encompasses valid concerns raised by the range of developers who contribute.

---
In reply to: [244521685](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521685) [](ancestors = 244521685)

src/Sarif/Core/Run.cs
michaelcfanning: No. This is just a bad bug you've identified that the unit test authoring has closed. In the case of files being null and 'addFileToTableIfNotPresent' == true, that boolean value would not be honored. Basically, this method should never return -1 if that value is true

---
In reply to: [244521733](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521733) [](ancestors = 244521733,244521702)

src/Sarif.Converters/TSLintConverter.cs
michaelcfanning: was actually intentional but you'd have to be in the code editor to see why i lined things up for readability. we don't have a lot of this sort of formatting as a standard practice (it's pervasive in most C++ code bases I've worked on). reverted.

---
In reply to: [244521769](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521769) [](ancestors = 244521769)

src/Sarif/Core/Run.cs
michaelcfanning: yes. arg. 

---
In reply to: [244521883](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521883) [](ancestors = 244521883)

src/Sarif/Core/Run.cs
michaelcfanning: variable declared on line 61. variable is initialized on call on line 62 (where it is used as an 'out' param). the only way line 62 isn't entered is in the event of an exception, in which case line 93 will not execute. now, if this code was executing in a 'finally' clause, you'd see the compilation error.

---
In reply to: [244521899](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521899) [](ancestors = 244521899)

src/Sarif/Core/Run.cs
michaelcfanning: we exclude the properties _bag_.

---
In reply to: [244521805](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521805) [](ancestors = 244521805)

src/Sarif/Core/Run.cs
michaelcfanning: Uri.AbsolutePath throws in cases where the URI isn't in fact, absolute. Uri.ToString() gets us where we want to be. Uri.OriginalString might also be sufficient though.

---
In reply to: [244521867](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244521867) [](ancestors = 244521867)

src/Sarif.Converters/TSLintConverter.cs
lgolding: Oh, right, lines up with "output". If I'd noticed that I wouldn't have commented.

---
In reply to: [244605570](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244605570) [](ancestors = 244605570,244521769)

src/Sarif.Converters/AndroidStudioConverter.cs
lgolding: This is marked "Resolved" but the local variable `tool` on Line 67 is still there.

---
In reply to: [244519901](https://github.com/Microsoft/sarif-sdk/pull/1191#discussion_r244519901) [](ancestors = 244519901)

PR #1194, opened 2019-01-01
External property files

michaelcfanning:
Update schema to conform to current spec in re: external property files.

@lgolding


src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// Rewriting visitor for the Sarif object model.
     /// </summary>
     [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.58.0.0")]
-    public abstract class SarifRewritingVisitor
+    public abstract partial class SarifRewritingVisitor
```
michaelcfanning: >SarifRewritingVisitor [](start = 34, length = 21)

Created this class to help minimize constant manual updates for generated file. There are still at least three manual changes required for emitted file.

src/Sarif/Core/SarifRewritingVisitor.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System;
+using System.CodeDom.Compiler;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Rewriting visitor for the Sarif object model.
```
lgolding: >model [](start = 47, length = 5)

Add a comment explaining that this is file contains the bulk of the changes that need to be "manually" applied to the class, and that it will go away once the class can be fully auto-generated.

src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
lgolding: Add a code comment to that effect.

---
In reply to: [244650147](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244650147) [](ancestors = 244650147)

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
```diff
@@ -278,13 +236,31 @@ public virtual ExceptionData VisitExceptionData(ExceptionData node)
...
+            return node;
+        }
+
+        public virtual ExternalPropertyFiles VisitExternalPropertyFiles(ExternalPropertyFiles node)
         {
             if (node != null)
             {
                 node.Conversion = VisitNullChecked(node.Conversion);
-                node.Files = VisitNullChecked(node.Files);
                 node.Graphs = VisitNullChecked(node.Graphs);
```
lgolding: >Graphs [](start = 21, length = 6)

You need to add a line for `node.Properties`.

lgolding:
One bug with a point fix, one request for a comment, otherwise LGTM.

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
michaelcfanning: there is no code emit for visiting properties anywhere currently, this is a TODO if we think it's useful

---
In reply to: [244650481](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244650481) [](ancestors = 244650481)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -254,63 +254,99 @@
...
+        "resources": {
+          "description": "An external property file containing a run.resources object to be merged with the root log file.",
+          "$ref": "#/definitions/externalPropertyFile"
         },
 
-        "invocations": {
-          "description": "An array of locations of files containing arrays of run.invocation objects to be merged with the root log file.",
+        "properties": {
+          "description": "An external property file containing a run.properties object to be merged with the root log file.",
+          "$ref": "#/definitions/externalPropertyFile"
```
michaelcfanning: >externalPropertyFile [](start = 33, length = 20)

@lgolding, note that our spec actually says the "properties" value is an external property file. this item doesn't actually have a property bag. i assume that users can populate run.properties if they need anything specific around this container.

lgolding: I am absolutely fine with the `externalPropertyFiles` object not itself having a property bag. I think Jim objected to that but I am ok with it, and I agree that `run.properties` will service.

But see my other comment about this issue, in the visitor code.

---
In reply to: [244831292](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244831292) [](ancestors = 244831292)

src/Sarif/Autogenerated/SarifRewritingVisitor.cs
lgolding: `externalPropertyFiles` has a property `properties`. But unlike every other property named `properties` in the spec, _this_ one is not of type `propertyBag`; it's of type `externalPropertyFile`. So it does have to be visited.

Am I missing something? I wonder if the visitor code gen omits any property whose name happens to be "properties". Checking...

---
In reply to: [244825443](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244825443) [](ancestors = 244825443,244650481)

lgolding: The codegen emits a visit for any object defined in the schema with a `"$ref"` rather than a `"type"`. The `"$ref`" tells the generator that the property's type is defined elsewhere in the schema, so it needs to be visited.

The schema says that `externalPropertyFiles.properties` is `"$ref": "#/definitions/externalPropertyFile"`, so the code generator should have produced a visit. Is it possible you missed it, or intentionally omitted it (because you thought it was a property bag), when you edited the not-yet-autogenerated file?

---
In reply to: [244835332](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244835332) [](ancestors = 244835332,244825443,244650481)

michaelcfanning: yes, i assumed this was a property bag, in which case there's no visitor bug. now that i've fixed that, yes, there is a visitor bug, thanks, will resolve.

---
In reply to: [244839002](https://github.com/Microsoft/sarif-sdk/pull/1194#discussion_r244839002) [](ancestors = 244839002,244835332,244825443,244650481)

PR #1201, opened 2019-01-03
Array scrub part 1: everything but anyOf-or-null properties.

lgolding:
NOTE: For explicitness, I added schema attributes even when they matched the JSON schema defaults, namely: `"minItems": 0` and `"uniqueItems": false`.

### DIFF
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -658,7 +658,6 @@
           "type": "array",
           "minItems": 0,
           "uniqueItems": false,
-          "default": [],
           "items": {
             "type": "string"
           }
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -20,7 +20,8 @@
     "runs": {
       "description": "The set of runs contained in this log file.",
       "type": "array",
-      "minItems": 1,
+      "minItems": 0,
+      "uniqueItems": false,
       "items": {
         "$ref": "#/definitions/run"
       }
@@ -55,8 +56,9 @@
         "regions": {
           "description": "An array of regions of interest within the attachment.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": true,
+          "default": [],
           "items": {
             "$ref": "#/definitions/region"
           }
@@ -65,8 +67,9 @@
         "rectangles": {
           "description": "An array of rectangles specifying areas of interest within the image.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": true,
+          "default": [],
           "items": {
             "$ref": "#/definitions/rectangle"
           }
@@ -96,7 +99,7 @@
           "description": "An array of one or more unique threadFlow objects, each of which describes the progress of a program through a thread of execution.",
           "type": "array",
           "minItems": 1,
-          "uniqueItems": true,
+          "uniqueItems": false,
           "items": {
             "$ref": "#/definitions/threadFlow"
           }
@@ -130,8 +133,9 @@
         "analysisToolLogFiles": {
           "description": "The locations of the analysis tool's per-run log files.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": true,
+          "default": [],
           "items": {
             "$ref": "#/definitions/fileLocation"
           }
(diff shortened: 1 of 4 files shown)


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -20,7 +20,8 @@
     "runs": {
       "description": "The set of runs contained in this log file.",
       "type": "array",
-      "minItems": 1,
+      "minItems": 0,
+      "uniqueItems": false,
```
michaelcfanning: >uniqueItems [](start = 7, length = 11)

basically now, any array with minItems == 0 which are not 'anyOf + null' as this one is, will have a default value of "[]". Yes?

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1628,6 +1679,8 @@
           "description": "An array of logical locations such as namespaces, types or functions.",
           "type": "array",
           "minItems": 0,
+          "uniqueItems": false,
```
michaelcfanning: >false [](start = 25, length = 5)

[https://github.com/oasis-tcs/sarif-spec/issues/304](https://github.com/oasis-tcs/sarif-spec/issues/304). let's definitely mark this issue as one that needs resolving. I don't understand why we aren't requiring each logical location entry to be unique. There should only be one entry in the array for each presumptive logical location. Allowing duplication will be problematic in doing things like rewriting indices after results matching.

btw - one idea I had for tracking spot changes of this kind that the TC hasn't review is to create a PR for them and to leave them open in the SDK. For this one, we'd simply create a branch off develop, adjust uniqueItems to 'true' and leave it as a TODO. the PR would need to be built out in some cases with test fixes, etc., once the TC approves.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1768,7 +1824,8 @@
         "frames": {
           "description": "An array of stack frames that represents a sequence of calls, rendered in reverse chronological order, that comprise the call stack.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
+          "uniqueItems": false,
```
michaelcfanning: >uniqueItems [](start = 11, length = 11)

default on this one s/be the empty array

michaelcfanning:
Thank you for completing this critical change. I reviewed it carefully and found a couple of issues. As part of the process, I also made some improvements to our definitive table used to drive the work at the link below. I put the table in the same order (alphabetical) as the schema, for example, which is a help in review. Finally, in addition to reviewing your explicit changes, I reviewed all the table entries that did not result in a code diff to ensure they are in conformance with the table.

[https://github.com/oasis-tcs/sarif-spec/issues/270](https://github.com/oasis-tcs/sarif-spec/issues/270)

src/Sarif/Schemata/sarif-schema.json
lgolding: Yes, I can't think of any exceptions to that.

---
In reply to: [245121606](https://github.com/Microsoft/sarif-sdk/pull/1201#discussion_r245121606) [](ancestors = 245121606)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -636,6 +656,9 @@
         "arguments": {
           "description": "An array of strings, containing in order the command line arguments passed to the tool from the operating system.",
           "type": "array",
+          "minItems": 0,
+          "uniqueItems": false,
+          "default": [],
```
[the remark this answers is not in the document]
lgolding: You are correct. Fixed.

---
In reply to: [245122913](https://github.com/Microsoft/sarif-sdk/pull/1201#discussion_r245122913) [](ancestors = 245122913)

src/Sarif/Schemata/sarif-schema.json
lgolding: It was `false` in the table in Issue #270, but I agree, I think we got that one wrong. The poster child for "non-unique" logical locations is ones that differ by `kind`, so they really _should_ be unique.

---
In reply to: [245125272](https://github.com/Microsoft/sarif-sdk/pull/1201#discussion_r245125272) [](ancestors = 245125272)

lgolding: NOT changing this because, as you say, the TC approved `false` here.

---
In reply to: [245137305](https://github.com/Microsoft/sarif-sdk/pull/1201#discussion_r245137305) [](ancestors = 245137305,245125272)

src/Sarif/Schemata/sarif-schema.json
lgolding: There is no default because `frames` is required.

---
In reply to: [245126076](https://github.com/Microsoft/sarif-sdk/pull/1201#discussion_r245126076) [](ancestors = 245126076)

PR #1202, opened 2019-01-03
Update supporting test code for processing checked in files.

michaelcfanning:
@lgolding @EasyRhinoMSFT 

This change contains the latest iteration of the SDK testing model for checked in files. The test pattern has these qualities:

1) Input files and expected sarif output text are compiled into assemblies as embedded resources. 
2) Input files are transformed into actual sarif text output and compared to expected output.
3) In the event of failure, actual and expected files are written to directories dedicated to each category
4) Test machinery then produces a batch file that compares these directories if any suite member fails
5) Developers can override a rebaselining property to automatically persist updated expected files text.

btw - the generated script is required because "\n" (literally the backslash followed by the n character) is transformed into a newline character in the test output window. This breaks any directory reporting for subdirectories that start with 'n' (which includes '\netcoreapp2.0'.

I've created new test suite for the prerelease transformers and updated the FPR converter as a proof point. I will eventually try to migrate most testing to this mechanism.

As always, I remain,
Unusually Finicky About Minimizing Unit Test Maintenance Costs


src/Sarif.Multitool.UnitTests/TransformCommandTests.cs
```diff
@@ -131,6 +131,7 @@ private static void RunTransformationToV2Test(string logFileContents)
             sarifLog.Version.Should().Be(SarifVersion.Current);
         }
 
+
```
EasyRhinoMSFT: Nit

PR #1203, opened 2019-01-08
Fix v1->v2 hash transformation

EasyRhinoMSFT:
CreateHash must be called to handle algorithm names that aren't in our translation table. Also updated a unit test to cover this case.


michaelcfanning:
oops

Issue #88, opened 2016-01-25
Write functional test to detect breaking schema changes


lgolding:
- [ ] Acquire a set of sample SARIF files produced by various tools.
- [ ] Create a test that validates them against the current schema

The idea is that if the test fails, we report the breaking change to our partners so they can update their tools to produce the new format version.

@michaelcfanning, @srivatsn: FYI. I entered this item to track the work I mentioned in today's standup.

PR #89, opened 2016-01-27
Add more information to error list, full message and friendly rule na…

michaelcfanning:
…mes.

PR #1204, opened 2019-01-08
Integrate jschema 0.61.0 into SDK

michaelcfanning:


src/Sarif/NotYetAutoGenerated/Run.cs
```diff
@@ -0,0 +1,447 @@
...
+        [DataMember(Name = "properties", IsRequired = false, EmitDefaultValue = false)]
+        internal override IDictionary<string, SerializedPropertyInfo> Properties { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Run" /> class.
+        /// </summary>
+        public Run()
+        {
+            RichMessageMimeType = "text/markdown;variant=GFM";
+            ColumnKind = ColumnKind.Utf16CodeUnits;
```
michaelcfanning: >ColumnKind.Utf16CodeUnits [](start = 25, length = 25)

Here's your column kind issue. For us to maintain reasonable compatibility, we need to instantiate this type by default with Utf16CodeUnits (which is not the literal default given by the schema, instead it's one that imposed by our Windows SDK to dig a 'pit of success' for this platform. 

By doing this, however, we also need to make sure to strip the other default handling. Otherwise we would never round-trip the alternate value (and actual default UnicodeCodePoints). That bad scenario would look like this:

1) user creates a log file, we init to utf16codeunits. user explicitly resets to unicodecodepoints.
2) user persist log file, and unicodecodepoints is elided (because it is the default)
3) user reopens log file, ctor inits to utf16codeunits and the unicodecodepoints setting is lost.

Instead, we basically overwrite everything such that run.columnKind is always persisted (by our SDK). If we happen to open a file without this value set, our SDK will treat it as Utf16CodeUnits.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -1081,7 +1081,7 @@ internal Tool CreateTool(ToolVersionOne v1Tool)
                 {
                     DottedQuadFileVersion = v1Tool.FileVersion,
                     FullName = v1Tool.FullName,
-                    Language = v1Tool.Language,
+                    Language = v1Tool.Language ?? "en-US",
```
michaelcfanning: >"en-US" [](start = 50, length = 7)

This value (and other terms) could usefully be stored as const values somewhere. I propose that jschema create (and itself refer to) these.

[https://github.com/Microsoft/jschema/issues/90](https://github.com/Microsoft/jschema/issues/90)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
```diff
@@ -24,7 +24,7 @@ private void VerifyFortifyFprConversionFromResource(string fprResourceName, stri
 
             byte[] fprData = GetResourceBytes(fprResourceName);
             string expectedSarifText = GetResourceText(expectedSarifResourceName);
-            PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(expectedSarifText, forceUpdate: false, formatting: Newtonsoft.Json.Formatting.None, out expectedSarifText);
+            PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(expectedSarifText, forceUpdate: false, formatting: Newtonsoft.Json.Formatting.Indented, out expectedSarifText);
```
michaelcfanning: change assists in debugging scenarios as all inputs are formatted.

src/Sarif.UnitTests/JsonTests.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     public abstract class JsonTests
     {
         protected static readonly Run DefaultRun = new Run();
-        protected static readonly Tool DefaultTool = new Tool();
+        protected static readonly Tool DefaultTool = new Tool() { Name = "DefaultTool" };
```
michaelcfanning: >Name = "DefaultTool"  [](start = 66, length = 21)

With the jschema 0.61.0 update, the validator now enforces the "required" element. And so there are several updates where tests did not previously provided required values.

src/Sarif.UnitTests/Readers/UriConverterTests.cs
```diff
@@ -104,8 +104,9 @@ private void TestConverter(string inputUri, string expectedUri)
   ""runs"": [
     {
       ""tool"": {
-        ""name"": null
+        ""name"": ""DefaultTool""
       },
+      ""columnKind"": ""utf16CodeUnits"",
```
michaelcfanning: >""columnKind"": ""utf16CodeUnits"" [](start = 6, length = 34)

It might not longer interest or surprise you to hear that I don't like updating inlined literals when maintaining tests.

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -169,15 +169,7 @@ public void SarifLogger_WritesSarifLoggerVersion()
...
-                    string ruleId = "RuleId";
-                    var rule = new Rule() { Id = ruleId };
-
-                    var result = new Result()
-                    {
-                        RuleId = ruleId
-                    };
-
-                    sarifLogger.Log(rule, result);
+                    LogDefaultResult(sarifLogger);
```
michaelcfanning: >LogDefaultResult [](start = 20, length = 16)

when i say that i aspire to make the world a DRY'er place, I'm not sure I like the way that sounds. You may be the only person in my circle who understands.

src/Sarif.ValidationTests/DefaultObjectPopulatingVisitor.cs
```diff
@@ -109,9 +109,10 @@ public override Node VisitNode(Node node)
             return node;
         }
 
-        private void PopulateInstanceWithDefaultMemberValues(ISarifNode node)        {
-           
+        private void PopulateInstanceWithDefaultMemberValues(ISarifNode node)
```
michaelcfanning: >PopulateInstanceWithDefaultMemberValues [](start = 21, length = 39)

FYI, these tests were very helpful in understanding our serialization/deserialization behaviors. They operate by producing data dynamically, by comprehensively walking/reflecting over the OM and poking values to various things. This is my nirvana for unit-testing: no inlined code, no data files driving testing (with unknown comprehensiveness).

michaelcfanning:
@lgolding, your jschema update, sir.

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -130,7 +127,7 @@ public void Initialize(Run run)
                 _serializer.Serialize(_jsonWriter, run.DefaultFileEncoding);
             }
 
-            if (run.RichMessageMimeType != null)
+            if (run.RichMessageMimeType != null && run.RichMessageMimeType != "text/markdown;variant=GFM")
```
michaelcfanning: >"text/markdown;variant=GFM [](start = 78, length = 26)

here's a place where it would be nice to reference a jschema-generated constant value.

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/ResultLocations.sarif
```diff
@@ -4,7 +4,8 @@
   "runs": [
     {
       "tool": {
-        "name": "CodeScanner"
+        "name": "CodeScanner",
+        "language": "en-US"
```
michaelcfanning: >"language": "en-US" [](start = 8, length = 19)

our SARIF v2 default now flows to all v1 transforms

src/Sarif/CodeGenHints.json
```diff
@@ -335,7 +335,7 @@
       "arguments": {
         "namespaceName": "Newtonsoft.Json",
         "typeName": "JsonConverter",
-        "arguments": [ "typeof(DateTimeConverter)" ]
+        "arguments": [ "typeof(Microsoft.CodeAnalysis.Sarif.Readers.DateTimeConverter)" ]
```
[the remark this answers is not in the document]
lgolding: Agreed.

---
In reply to: [246156914](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246156914) [](ancestors = 246156914)

src/Sarif/NotYetAutoGenerated/Result.cs
```diff
@@ -0,0 +1,511 @@
...
+        /// </summary>
+        [DataMember(Name = "baselineState", IsRequired = false, EmitDefaultValue = false)]
+        [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
+        public BaselineState BaselineState { get; set; }
+
+        /// <summary>
+        /// A number representing the priority or importance of the result.
+        /// </summary>
+        [DataMember(Name = "rank", IsRequired = false, EmitDefaultValue = false)]
+        [DefaultValue(0.0)]
```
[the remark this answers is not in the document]
lgolding: Agreed. Will fix along with #89.

---
In reply to: [246157762](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246157762) [](ancestors = 246157762)

src/Sarif.UnitTests/Readers/UriConverterTests.cs
lgolding: I've started the journey. Please see my JSchema PR [https://github.com/Microsoft/jschema/pull/91](https://github.com/Microsoft/jschema/pull/91). Would you please review this so when I fix #88 and #89 I can build on the new infrastructure.

---
In reply to: [246164297](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246164297) [](ancestors = 246164297)

michaelcfanning: Will take a look. The most recent SARIF SDK FileDiffing base class provides an end-to-end pattern for this, did you happen to look at it?

src/Sarif/Writers/ResultLogJsonWriter.cs
lgolding: Like, if a property `MyProp` of type `T` has default `X`, then generate this in the class:
```
public const T DefaultMyProp = X;
```
In this example, `Run.DefaultRichMessageType` of type `string`.

That is a great idea (and I see you already filed jschema#90 for it).

---
In reply to: [246175862](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246175862) [](ancestors = 246175862)

src/Sarif.UnitTests/Readers/UriConverterTests.cs
lgolding: I didn't. I was going by your description in email about what you were after. But I will take a look.

I expect the bulk of my change to survive that examination because it's mostly about ripping literals out into files and renaming a few variables, but I might get a surprise.

---
In reply to: [246508060](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246508060) [](ancestors = 246508060)

lgolding:
Ok, this looks good, thanks. I am building my small changes now and then will push and merge.

---
In reply to: [452476061](https://github.com/Microsoft/sarif-sdk/pull/1204#issuecomment-452476061) [](ancestors = 452476061)

src/Sarif/NotYetAutoGenerated/Run.cs
michaelcfanning: Having a knob to avoid emitting a schema-specified default would be helpful but is an incomplete solution, because we follow up with this by providing an alternate OM initialization only default. Not sure how to express this subtlety in jschema or if it even warrants doing so.

Note that all this would be much easier if this platform-specific data was required. We rejected doing this, however, because it seemed unnecessarily burdensome for tools that don't emit column details.

---
In reply to: [246527802](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246527802) [](ancestors = 246527802,246159777)

src/Sarif.UnitTests/Readers/UriConverterTests.cs
lgolding: You may well be right. Two things: please take a quick look and let me know whether the PR even moves in the right direction. If not, I'll discard it. In either case, I accept your offer to rough out the file diff solution -- whether or not the PR is a useful base for you to do that.

Finally, if we have infrastructure shared across 5 repos, we really need a separate repo to hold it, to include as a submodule in the 5 clients.

---
In reply to: [246550880](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246550880) [](ancestors = 246550880,246525111,246508060)

src/Sarif/NotYetAutoGenerated/Run.cs
lgolding: Another option is to have the hint supply the overridden value.

---
In reply to: [246546734](https://github.com/Microsoft/sarif-sdk/pull/1204#discussion_r246546734) [](ancestors = 246546734,246527802,246159777)

PR #1207, opened 2019-01-12
Start work on v1 <-> v2 transformation

lgolding:
Restore some commented out files; comment out only the few lines that don't work. Uncomment the tests of the v2 -> v1 transformation to observe the failures (6/10 succeed; the failure cases are the ones that need the files array).


lgolding:
@michaelcfanning Sorry, targeted the wrong branch. Will resubmit.

PR #1209, opened 2019-01-12
Original uri base id fixups

michaelcfanning:
@lgolding 

This change updates the visitor that expands uriBaseIds in the presence of originalUriBaseIds. Turns out there was technical debt here, because out tests provides originalUriBaseId equivalents in the property bag (because we had no official SARIF slot for them). I did not notice this gap when we made the schema change to add originalUriBaseIds. 

There are some confusing things in this change I can't account for currently. I've flagged them for research. Maybe you understand what's happening off the top of your head. Commenting now...


lgolding:
Good catch :-)

PR #1211, opened 2019-01-14
Get v2 -> v1 transform working with files array

lgolding:
Test failure count is down to 32; will be 28 when you merge your fix.

There is not -- and never was -- a test case for fileLocations that use uriBaseId (never was one). I know for a fact that there is no code to support that case. You’ll see a comment to that effect in the code. I will take care of that next. Then I will move on to v1 -> v2 transform.


src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/OneRunWithBasicInvocation.sarif
```diff
@@ -27,9 +27,7 @@
         }
       },
       "files": {
-        "collections.rsp": {
-          "uri": "collections.rsp"
-        }
+        "collections.rsp": {}
```
lgolding: The transformer now refrains from emitting `uri` if it matches the key, as permitted by the v1 spec. We could go further and omit the whole object if it's empty but that's more work that I'm willing to do, and would cause problems if the empty object were in a nested file chain.

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/OneRunWithFiles.sarif
```diff
@@ -27,7 +27,7 @@
           }
         },
         "file:///home/buildAgent/bin/app.zip#/docs/intro.docx": {
-          "uri": "file:///docs/intro.docx",
```
lgolding: >file [](start = 18, length = 4)

I fixed this existing bug in the transformer.

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/Regions.sarif
```diff
@@ -11,16 +11,13 @@
       "invocation": {},
       "files": {
         "file:///src/base/driver.cs": {
-          "uri": "file:///src/base/driver.cs",
```
lgolding: Again, the transformer now doesn't bother to emit these.

src/Sarif.UnitTests/Visitors/SarifCurrentToVersionOneVisitorTests.cs
```diff
@@ -69,9 +69,7 @@ private void VerifyCurrentToVersionOneTransformationFromResource(string v2InputR
                 File.WriteAllText(expectedFilePath, v1ExpectedLogText);
                 File.WriteAllText(actualFilePath, v1ActualLogText);
 
-
-                string errorMessage = string.Format(@"V2 conversion from V1 produced unexpected diffs for test: '{0}'.", v2InputResourceName);
```
lgolding: >errorMessage [](start = 23, length = 12)

This variable was never used, so this message never appeared.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -904,6 +903,105 @@ internal RunVersionOne CreateRun(Run v2Run)
...
+        //    2. There might be two files with the same URI but difference uriBaseIds.
+        //
+        // To deal with Case #1, we synthesize a key as follows:
+        //
+        //    <root file URI>#<nested file path>/<level 2 nested file path>/...
+        //
+        // To deal with Case #2, we would follow the chain or uriBaseIds, prepending
+        // the value of each one to the URI.
+        //
+        // WE DO NOT YET HANDLE CASE #2.
```
lgolding: This is the missing code and test case.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -600,13 +604,15 @@ private Stream GetContentStream(Uri uri, out Encoding encoding)
...
+            IList<FileData> files = _currentV2Run.Files;
 
-            if (uri != null && _currentV2Run.Files != null)
+            if (uri != null && files != null && _fileKeyToIndexDictionary != null)
             {
-#if FILES_ARRAY_WORKS
-                FileData fileData;
-                if (_currentV2Run.Files.TryGetValue(uri.OriginalString, out fileData))
+                if (_fileKeyToIndexDictionary.TryGetValue(uri.OriginalString, out int index)
+                    && index < files.Count)
```
michaelcfanning: > && index < files.Count [](start = 19, length = 23)

when would this index < files.Count not be true? that seems like an adverse conition

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -22,6 +22,10 @@ public class SarifCurrentToVersionOneVisitor : SarifRewritingVisitor
         private Run _currentV2Run = null;
         private RunVersionOne _currentRun = null;
 
+        // To understand the purpose of these fields, see the comment on CreateFileKeyIndexMappings.
+        private IDictionary<string, int> _fileKeyToIndexDictionary;
```
michaelcfanning: >_fileKeyToIndexDictionary [](start = 41, length = 25)

can you specify the version-ness of these things in the name? e.g, v1FileKeytoV2Index dictionary? Helps with clarity. For both members please.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -904,6 +903,105 @@ internal RunVersionOne CreateRun(Run v2Run)
...
+        {
+            var sb = new StringBuilder(v2File.FileLocation.Uri.OriginalString);
+            while (v2File.ParentIndex != -1)
+            {
+                FileData parentFile = v2Files[v2File.ParentIndex];
+
+                // The convention for building the key is as follows:
+                // The root file URI is separated from the chain of nested files by '#';
+                // The nested file URIs are separated from each other by '/'.
+                string separator = parentFile.ParentIndex == -1 ? "#" : "/";
```
michaelcfanning: >string separator = parentFile.ParentIndex [](start = 16, length = 41)

are you sure this code doesn't double slashes at the point where we jump from a nested archive to an element in it? 

i.e., need a test case for myarchive.zip#myarchive2.zip/foo.cpp

in this three-part file object, the most nested item will have a URL of /foo.cpp

does your code double that slash? like so myarchive.zip#myarchive2.zip//foo.cpp?

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -904,6 +903,105 @@ internal RunVersionOne CreateRun(Run v2Run)
...
+
+            if (_fileKeyToIndexDictionary != null)
+            {
+                filesVersionOne = new Dictionary<string, FileDataVersionOne>();
+                foreach (KeyValuePair<string, int> entry in _fileKeyToIndexDictionary)
+                {
+                    string key = entry.Key;
+                    int index = entry.Value;
+                    FileData v2File = _currentV2Run.Files[index];
+                    FileDataVersionOne fileDataVersionOne = CreateFileData(v2File);
```
michaelcfanning: >CreateFileData [](start = 60, length = 14)

Again prefer to version these things, e.g., CreateFileDataVersionOne

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/OneRunWithFiles.sarif
michaelcfanning: yes good catch, can you please add another test to this file that nests a second zip in this one, and another file object that is a nested file within that nested zip? this guarantees completeness in nested files testing.

---
In reply to: [247374942](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247374942) [](ancestors = 247374942)

src/Sarif.UnitTests/Visitors/SarifCurrentToVersionOneVisitorTests.cs
michaelcfanning: ok. this code needs to be converted to actually use the helpers in the filediff base class.

---
In reply to: [247374990](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247374990) [](ancestors = 247374990)

michaelcfanning:
I am mostly authoring docs tomorrow. Please update this PR based on my feedback according to your judgment and merge. After your update, I am it will LGTM.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: You're right: it's guaranteed by the way the dictionary is constructed. Removed this condition.

---
In reply to: [247376623](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247376623) [](ancestors = 247376623)

lgolding:
Ok, thanks for the review.

---
In reply to: [453900063](https://github.com/Microsoft/sarif-sdk/pull/1211#issuecomment-453900063) [](ancestors = 453900063)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Fixed.

---
In reply to: [247376669](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247376669) [](ancestors = 247376669)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Yes, I agree: I always found this code confusing for that reason, and _all_ the `CreateXxx` methods follow this bad pattern. I'll change them all.

---
In reply to: [247376992](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247376992) [](ancestors = 247376992)

lgolding: Fixed.

---
In reply to: [247565505](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247565505) [](ancestors = 247565505,247376992)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Good catch! I forgot that nested file URIs already start with a slash. I'll add the test case you asked for to expose this bug, and then fix it.

---
In reply to: [247376946](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247376946) [](ancestors = 247376946)

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/v1/OneRunWithFiles.sarif
lgolding: Done.

---
In reply to: [247377083](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247377083) [](ancestors = 247377083,247374942)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: Fixed.

---
In reply to: [247570514](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247570514) [](ancestors = 247570514,247376946)

src/Sarif.UnitTests/Visitors/SarifCurrentToVersionOneVisitorTests.cs
lgolding: I'll do it now, while I'm in here.

---
In reply to: [247377138](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247377138) [](ancestors = 247377138,247374990)

lgolding: Done! This SarifCurrentToVersionOneVisitorTests.cs is no 65 lines long :-) Take a look at the changes to FileDiffingTests.cs to see how it works.

---
In reply to: [247582997](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247582997) [](ancestors = 247582997,247377138,247374990)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: I will handle this in a separate PR, after lunch, after I review your PR.

---
In reply to: [247375067](https://github.com/Microsoft/sarif-sdk/pull/1211#discussion_r247375067) [](ancestors = 247375067)

PR #1212, opened 2019-01-14
Update insert optional data tests and update indices visitor.

michaelcfanning:


src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
```diff
@@ -1,13 +1,16 @@
...
+using Microsoft.CodeAnalysis.Sarif.TestUtilities;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.CodeAnalysis.Sarif.Converters
 {
-    public class FortifyFprConverterTests : FileDiffingTests
+    public class FortifyFprConverterTests : FileDiffingTests, IClassFixture<FortifyFprConverterTests.FortifyFprConverterTestsFixture>
     {
+        public class FortifyFprConverterTestsFixture : DeletesOutputsDirectoryOnClassInitializationFixture { }
```
michaelcfanning: >DeletesOutputsDirectoryOnClassInitializationFixture [](start = 55, length = 51)

This fixture deletes the outputs directory once per test session. Previously, we deleted this directory on executing every test case. Oops. This means that only the very last test failure was persisted to the directory.

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -53,37 +48,38 @@ public FileDiffingTests(ITestOutputHelper outputHelper)
 
         protected virtual string TestLogResourceNameRoot => "Microsoft.CodeAnalysis.Sarif.UnitTests.TestData." + TypeUnderTest;
 
-        protected virtual string ConstructTestOutputFromInputResource(string inputResource) { return string.Empty; }
+        protected virtual string ConstructTestOutputFromInputResource(string inputResourceName) { return string.Empty; }
 
         protected virtual bool RebaselineExpectedResults => false;
 
-        protected virtual void RunTest(string resourceName)
+        protected virtual void RunTest(string inputResourceName, string expectedOutputResourceName = null)
```
michaelcfanning: >RunTest [](start = 31, length = 7)

Here's how we handle one-to-many tests. Create an individual test referencing the common input, but generate an alternate output name. See the optional data visitor tests for a specific example.

michaelcfanning:
>             logs.All(

I changed some whitespace in this file, hence it is included. Oops. But observe this construct, which is currently failing. It is impossible to debug this test failure, need to improve this test for maintainability. 

---
Refers to: src/Sarif.UnitTests/Processors/Log/SarifLogExtensionTests.cs:114 in e39fb52. [](commit_id = e39fb525a55d944cea7019faf8832acc74d4a833, deletion_comment = False)

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/NestedFiles.sarif
```diff
@@ -46,6 +46,22 @@
             "fileIndex": 3
           },
           "mimeType": "text/plain"
```
michaelcfanning: >mimeType [](start = 11, length = 8)

This is the test update that somehow got clobbered recently.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -3,138 +3,111 @@
...
 using Microsoft.CodeAnalysis.Sarif.Writers;
 using Newtonsoft.Json;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.CodeAnalysis.Sarif.Visitors
 {
-    public class InsertOptionalDataVisitorTests : FileDiffingTests
+
+    public class InsertOptionalDataVisitorTests : FileDiffingTests, IClassFixture<InsertOptionalDataVisitorTests.InsertOptionalDataVisitorTestsFixture>
```
michaelcfanning: >InsertOptionalDataVisitorTests [](start = 17, length = 30)

converted this test suite to current pattern. suggest just looking at the new file rather than the diffs. will be most informative.

lgolding:
On it now.

---
In reply to: [453899975](https://github.com/Microsoft/sarif-sdk/pull/1212#issuecomment-453899975) [](ancestors = 453899975)

lgolding:
>             logs.All(

Oh good heavens. It's elegant, I'll give it that ;-)

---
In reply to: [453898287](https://github.com/Microsoft/sarif-sdk/pull/1212#issuecomment-453898287) [](ancestors = 453898287)

---
Refers to: src/Sarif.UnitTests/Processors/Log/SarifLogExtensionTests.cs:114 in e39fb52. [](commit_id = e39fb525a55d944cea7019faf8832acc74d4a833, deletion_comment = False)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
lgolding: Oh, nice, I will add that to SarifCurrentToVersionOneVisitorTests.

---
In reply to: [247381989](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247381989) [](ancestors = 247381989)

src/Sarif.TestUtilities/FileDiffingTests.cs
lgolding: So if an operation (such as a JSchema code gen from a schema) generates multiple outputs, we actually have to run the operation as many times as there are output files, even though all the outputs are generated each time? I'm not wild about that.

---
In reply to: [247382063](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247382063) [](ancestors = 247382063)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
michaelcfanning: Just want to be clear you understand this pattern. A fixture is called on a class with a public parameterless constructor. it is instantiated a single time on xunit identifying a class with one or more [Fact] or other attribute-derived tests. The test class itself cannot be the fixture class (as xUnit supports only one public ctor on a test class, due to the diversity of parameterizations of these things, I guess. Our fixture derives some computed data (i.e., the test output directory) from our test organization scheme. From the base class. I have recapped this minimal data computation and chosen to embed these fixtures as nested types with the actual test class (since they are not generally repurposable. The fixture must be named identically to the test class with the term 'Fixture' appended to it.

If you can think of anything cleaner/more elegant, let me know. The fixture itself can be provided, btw, to the test class ctor, with the result that it can be additionally parameterized when it is passed to each test case. This parameterization can be used most easily for test cleanup. You could imagine us simply placing a bool property on this thing, though ('HaveIDeletedTheOutputDirectoryYet' and having the first test case note that it's false, perform the delete itself, and set that bool to true. The test fixture is effectively just a bundle of static data with this approach. It would eliminate the requirement to define a nested type for each test (because we could put this standard operation into the file diffing base type). Thoughts?


---
In reply to: [247657934](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247657934) [](ancestors = 247657934,247381989)

michaelcfanning:
>             logs.All(

Indeed. Gratifyingly compact and the code is actually fairly readable. Unfortunately, it is impossible to debug on the failure case. :) 

---
In reply to: [454160497](https://github.com/Microsoft/sarif-sdk/pull/1212#issuecomment-454160497) [](ancestors = 454160497,453898287)

---
Refers to: src/Sarif.UnitTests/Processors/Log/SarifLogExtensionTests.cs:114 in e39fb52. [](commit_id = e39fb525a55d944cea7019faf8832acc74d4a833, deletion_comment = False)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
lgolding: I'll read up on XUnit's "fixture" support and get back to you.

---
In reply to: [247661342](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247661342) [](ancestors = 247661342,247657934,247381989)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
     public class UpdateIndicesVisitor : SarifRewritingVisitor
     {
         private IDictionary<string, int> _fullyQualifiedLogicalNameToIndexMap;
-        private IDictionary<string, int> _fileLocationKeyToIndexMap;
+        private IDictionary<FileLocation, int> _fileLocationToIndexMap;
 
-        public UpdateIndicesVisitor(IDictionary<string, int> fullyQualifiedLogicalNameToIndexMap, IDictionary<string, int> fileLocationKeyToIndexMap)
+        public UpdateIndicesVisitor(IDictionary<string, int> fullyQualifiedLogicalNameToIndexMap, IDictionary<FileLocation, int> fileLocationToIndexMap)
```
lgolding: >fileLocationToIndexMap [](start = 129, length = 22)

The one place this class is instantiated in product code, you now set this parameter to `null`. Is this a work in progress? Because as the product code stands in this branch, `VisitFileLocation` never does anything.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -80,7 +80,7 @@ public static class PrereleaseCompatibilityTransformer
             {
                 Debug.Assert(modifiedLog == true);
                 transformedSarifLog = JsonConvert.DeserializeObject<SarifLog>(sarifLog.ToString());
-                var indexUpdatingVisitor = new UpdateIndicesVisitor(fullyQualifiedLogicalNameToIndexMap, fileLocationKeyToIndexMap);
+                var indexUpdatingVisitor = new UpdateIndicesVisitor(fullyQualifiedLogicalNameToIndexMap, null);
```
lgolding: >null [](start = 105, length = 4)

See my comment in `UpdateIndicesVisitor.cs`.

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
lgolding: I read the XUnit documentation of [shared context](https://xunit.github.io/docs/shared-context). Now I know enough to ask:

Why do you derive a class from `DeletesOutputsDirectoryOnClassInitializationFixture`? I could understand it if the derived class added some functionality, but it doesn't. Why not just:

```
FortifyFprConverterTests : FileDiffingTests, IClassFixture<DeletesOutputsDirectoryOnClassInitializationFixture>
{
    ...
}
```

---
In reply to: [247662456](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247662456) [](ancestors = 247662456,247661342,247657934,247381989)

src/Sarif.TestUtilities/DeletesOutputsDirectoryOnClassInitializationFixture.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif.TestUtilities
+{
+    /// <summary>
+    /// This class is invoked a single time on initializing an xunit class for testing. It deletes
+    /// any existing test output files that may have been produced by the previous run. This fixture
+    /// is required because individual tests in test classes result in an object instantiation for
+    /// each test case. The FileDiffingTests pattern, by design outputs a bundle of outputs to
```
lgolding: >FileDiffingTests [](start = 28, length = 16)

Since this fixture is intended to support classes that derive from `FileDiffingTests`, why not do this:
```
class FileDiffingTests: IClassFixture<DeletesOutputsDirectoryOnClassInitializationFixture>
{
    ...
}

class FortifyFprConverterTests: FileDiffingTests
{
    ...
}
```

... rather than marking each derived class as implementing `IClassFixture`. Does XUnit's reflection not find the interface if it's on a base class?

lgolding: This might relate to the comment I made on `FortifyFprConverterTests`, where I asked why you bothered to derive a class from `DeletesOutputsDirectoryOnClassInitializationFixture`. It's true that if you want to leave the flexibility for each derived test class to have its own fixture, then you can't put the marker interface on the base class.

If that was your reasoning, it's worth a comment here.

---
In reply to: [247725332](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247725332) [](ancestors = 247725332)

lgolding: Ok, I see now that `DeletesOutputsDirectoryOnClassInitializationFixture` has some virtual properties that derived classes can override, so you really do intend each derived class to have a shot at it. I've been reworking this comment, so I'll mention that.

---
In reply to: [247725737](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247725737) [](ancestors = 247725737,247725332)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
lgolding: Ok, I'm bouncing back and forth between commenting on the fixture and commenting on this class :-). I now see that the fixture has virtual properties that a `FileDiffingTests`-derived class can override in a bespoke derived fixture class. But in cases like this class, that do _not_ override those properties or do anything else, I still think you should just mark them with `IClassFixture<DeletesOutputsDirectoryOnClassInitializationFixture>`.

---
In reply to: [247724801](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247724801) [](ancestors = 247724801,247662456,247661342,247657934,247381989)

src/Sarif.TestUtilities/DeletesOutputsDirectoryOnClassInitializationFixture.cs
lgolding: Ok, I updated this comment with my accustomed crystalline authorial voice :-)

---
In reply to: [247729832](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247729832) [](ancestors = 247729832,247725737,247725332)

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -191,14 +191,6 @@ public void RebaseUriVisitor_VisitFileData_RebasesAllTheThings()
             string uriRootText = "file:///home/buildAgent/";
             string toolsRootBaseId = "TOOLS_ROOT";
 
-            visitor.FileDataParentKeys.Count.Should().Be(4);
-            visitor.FileDataParentKeys.Where(k => k == null).Count().Should().Be(3);
-            visitor.FileDataParentKeys.Where(k => k != null && k.StartsWith(uriRootText)).Count().Should().Be(1);
-
-            visitor.FileDataKeys.Count.Should().Be(4);
-            visitor.FileDataKeys.Where(k => k != null && k.StartsWith(uriRootText)).Count().Should().Be(3);
-            visitor.FileDataKeys.Where(k => k != null && k.StartsWith("#" + toolsRootBaseId + "#")).Count().Should().Be(1);
```
lgolding: >visitor [](start = 12, length = 7)

This test still fails. Was this change intended to fix it?

src/Sarif.UnitTests/Writers/SarifLoggerTests.cs
```diff
@@ -463,7 +463,7 @@ public void SarifLogger_ScrapesFilesFromResult()
             {
                 string fileName = @"file" + i + ".cpp";
                 string fileDataKey = "file:///" + fileName;
-                sarifLog.Runs[0].Files.Where(f => f.FileLocation.Uri.AbsolutePath.Contains(fileDataKey)).Any().Should().BeTrue();
+                sarifLog.Runs[0].Files.Where(f => f.FileLocation.Uri.AbsoluteUri.ToString().Contains(fileDataKey)).Any().Should().BeTrue();
```
lgolding: >AbsoluteUri [](start = 69, length = 11)

This change apparently fixes this test. But in your branch, three other SarifLoggerTests went from failing to passing (see the table I sent you), but I don't see any changes that would account for that.

lgolding:
I have a couple of substantive comments and questions, about why some tests are now passing, and why one test is still failing, and why one product code class apparently isn't doing anything. :-) I also have a suggestion about the use of `DeletesOutputDirectoryOnClassInitializationFixture` (and another suggestion about its name).

I've merged `files-array` into your branch and pushed back to GitHub, so you should be able to pull and keep on trucking.

src/Sarif.TestUtilities/DeletesOutputsDirectoryOnClassInitializationFixture.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// by a previous run. It is designed for use on test classes that derive from FileDiffingTests.
+    /// It is required because FileDiffingTests emits the outputs from each test to a common directory,
+    /// to allow diffing an entire directory of failing tests. If each test case deleted this directory,
+    /// then at the end it would contain only the output from the last failing test.
+    ///
+    /// Each class that derives from FileDiffingTests can declare its own derived fixture class if
+    /// it wants to override the virtual TypeUnderTest or OutputFolderPath properties, but there seems
+    /// no good reason to do this.
+    /// </summary>
+    public abstract class DeletesOutputsDirectoryOnClassInitializationFixture
```
lgolding: >DeletesOutputsDirectoryOnClassInitializationFixture [](start = 26, length = 51)

This is an xUnit class fixture (which, as we now both know, does work on class initialization (and sometimes on class disposal). I suggest the name `DeletesOutputDirectoryClassFixture` (or even `OutputDirectoryDeletingClassFixture`).

michaelcfanning:
On it, thanks for the merge.

---
In reply to: [454234368](https://github.com/Microsoft/sarif-sdk/pull/1212#issuecomment-454234368) [](ancestors = 454234368)

src/Sarif.Converters.UnitTests/FortifyFprConverterTests.cs
michaelcfanning: No, this won't work. These properties are initialized based on class naming convention. A single class can only represent the intended output location of a single suite. I'm just going to change the pattern entirely, stay tuned. Not this change.

---
In reply to: [247732570](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247732570) [](ancestors = 247732570,247724801,247662456,247661342,247657934,247381989)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
michaelcfanning: Yes. The VisitFileLocation in this visitor only applies to result matching. So this functionality is still pending. The prerelease transformer doesn't need this update to function.

---
In reply to: [247716347](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247716347) [](ancestors = 247716347)

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
michaelcfanning: No, just an incidental change to allow for compilation: there are no more file data keys persisted by this visitor

---
In reply to: [247734541](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247734541) [](ancestors = 247734541)

src/Sarif.TestUtilities/DeletesOutputsDirectoryOnClassInitializationFixture.cs
michaelcfanning: Our fixture only deletes on class initialization, so would prefer to retain the name. In any case, this fixture is going to be changed, I'm switching to a different pattern as this one is non-obvious and cumbersome to implement.

---
In reply to: [247737869](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247737869) [](ancestors = 247737869)

src/Sarif.TestUtilities/FileDiffingTests.cs
michaelcfanning: glad to discuss further.

---
In reply to: [247659167](https://github.com/Microsoft/sarif-sdk/pull/1212#discussion_r247659167) [](ancestors = 247659167,247382063)

michaelcfanning:
@lgolding, in fact, you were right to worry about the update indices visitor. I broke that test. Fixing separately, thanks for catching this.

PR #1214, opened 2019-01-15
Convert debug assert to exception to make test execution more determi…

michaelcfanning:
…nistic.

@lgolding


michaelcfanning:
@lgolding, I'm checking this in. Please review. Despite the name of the branch, i do not actually remove any debug.assert, instead I fixed the issue that resulted in the assert firing. we need to do a scrub for debug.assert later.

lgolding:
These are all fine changes. I still don't understand how a `Debug.Assert` can be any less deterministic than an exception (Debug vs Release aside, I mean), assuming the rest of code is deterministic. But I'm glad it's fixed.

PR #1215, opened 2019-01-16
The multiool, core sarif, and validation test binaries now all pass

michaelcfanning:


src/Sarif/Autogenerated/Rule.cs
```diff
@@ -43,7 +43,7 @@ public SarifNodeKind SarifNodeKind
         /// An array of stable, opaque identifiers by which this rule was known in some previous version of the analysis tool.
         /// </summary>
         [DataMember(Name = "deprecatedIds", IsRequired = false, EmitDefaultValue = false)]
-        public IList<string> DeprecatedIds { get; set; }
+        public virtual IList<string> DeprecatedIds { get; set; }
```
michaelcfanning: >IList [](start = 23, length = 5)

Needs to be virtual due to converting skimmer base to derive from Rule (as part of ongoing effort to extirpate IRule)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -32,15 +32,27 @@ public override Location VisitLocation(Location node)
 
         public override FileLocation VisitFileLocation(FileLocation node)
         {
-            if (_fileLocationToIndexMap != null)
+
+            if (_fileLocationKeyToIndexMap != null)
```
michaelcfanning: >_fileLocationKeyToIndexMap [](start = 16, length = 26)

I removed necessary functionality earlier. The prerelease transformer depends on the old prerelease v2 file location key to perform some fixups. The result matching update will require index updates that match current v2 file locations. We'll need to update this class further then (or provide better separation of purpose for v2 update vs. current v2 matching.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -80,7 +80,7 @@ public static class PrereleaseCompatibilityTransformer
             {
                 Debug.Assert(modifiedLog == true);
                 transformedSarifLog = JsonConvert.DeserializeObject<SarifLog>(sarifLog.ToString());
-                var indexUpdatingVisitor = new UpdateIndicesVisitor(fullyQualifiedLogicalNameToIndexMap, null);
+                var indexUpdatingVisitor = new UpdateIndicesVisitor(fullyQualifiedLogicalNameToIndexMap, fileLocationKeyToIndexMap);
```
michaelcfanning: >fileLocationKeyToIndexMap [](start = 105, length = 25)

your bug. nice catch, thanks.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -93,7 +93,10 @@ public static class PrereleaseCompatibilityTransformer
                 // above this call). We are required to regenerate it, however, in order to properly 
                 // elide default values, etc. I could not find a way for the JToken driven
                 // ToString()/text-generating mechanism to honor default value ignore/populate settings.
-                updatedLog = JsonConvert.SerializeObject(transformedSarifLog);
+                if (forceUpdate || modifiedLog)
```
michaelcfanning: >forceUpdate [](start = 20, length = 11)

real issue here is that we weren't honoring the formatting param. added the forceUpdate/modifiedLog checks just to minimize code execution for certain cases

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -7,17 +7,12 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
-    public abstract class SkimmerBase<TContext>  : PropertyBagHolder, ISkimmer<TContext>
+    public abstract class SkimmerBase<TContext>  : Rule, ISkimmer<TContext>
```
michaelcfanning: >Rule [](start = 51, length = 4)

Rule derives from PropertyBag. The big problem with not using interfaces, of course, is that you need a single, sensible base class chain. We barely are able to accomplish this to eliminate IRule (thanks to the SARIF design change to make everything have a property bag).

src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs
```diff
@@ -87,15 +88,31 @@ protected void Verify(string testFileName, bool disablePrereleaseCompatibilityTr
         // Verify that those detected result locations match the expected locations.
         private void Verify(Run run, ExpectedValidationResults expectedResults)
```
michaelcfanning: >Verify [](start = 21, length = 6)

I couldn't debug this test easily. So made a slight improvement where the relevant locations of unexpected conditions were isolated/more obvious.

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif
```diff
@@ -25,11 +25,12 @@
           ]
         }
       ],
-      "files": {
-        "fi%le:///c:/src/file.c": {
+      "files": [
```
michaelcfanning: >files [](start = 7, length = 5)

for this file, had to spot-inject the v2 pattern, otherwise this non-transformed file doesn't deserialize. what a frankenfile! or i guess it's actually frankenstein's file

src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif
```diff
@@ -98,54 +67,89 @@
...
+          "revisionId": "de67ef7",
+          "mappedTo": {
+            "uri": "file:///C:/src/file.c",
+            "uriBaseId": "%SRCROOT%"
+          }
+        }
+      ],
+      "files": [
+        {
+          "fileLocation": {
```
michaelcfanning: >fileLocation [](start = 11, length = 12)

I had to update this entire file to SARIF v2 because otherwise the prerelease transformer would manufacture a new (correct) uri for the file. Rather than try to special-case the helper just to provoke testing of invalid content, just updated the file.

src/Sarif.UnitTests/Processors/Log/SarifLogExtensionTests.cs
```diff
@@ -70,30 +71,30 @@ public void RebaseUri_WorksAsExpected()
         public void AbsoluteUri_ReversesRebasedURIs()
```
michaelcfanning: >AbsoluteUri_ReversesRebasedURIs [](start = 20, length = 31)

made this thing slightly more debuggable. we now iterate over each log and block on discovering problem.

this test is just brain dead, though. it produces two kinds of uris only, for files table and results. the randomness is simply around the count of logs, runs, files, results. Dumb. let's just do a better job of writing this test later so that it operates against comprehensive data. this thought is why i didn't push a test pattern improvement elsewhere, just updated the failing test (which now passes).

michaelcfanning:
>         public void RebaseUriVisitor_VisitFileData_RebasesAllTheThings()

Horrible-to-maintain test champion. For the record, I authored this thing. Nerts. Trust me, I hand validated all the constant value changes below. We need a better test strategy here. It will be simple, actually, just need to produce the transformed sarif to disk, diff as usual. i'll do this soon.

---
Refers to: src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs:178 in bd80b54. [](commit_id = bd80b54d281999c8f36b91364404fa264c14a9db, deletion_comment = False)

michaelcfanning:
@lgolding, the favor of your review is requested.

src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs
```diff
@@ -87,15 +88,31 @@ protected void Verify(string testFileName, bool disablePrereleaseCompatibilityTr
...
         {
-            string[] detectedResultLocations = run.Results.Select(r => r.Message.Arguments[0]).OrderBy(loc => loc).ToArray();
-            string[] expectedResultLocations = expectedResults.ResultLocationPointers.OrderBy(loc => loc).ToArray();
+            HashSet<string> actualResultLocations = new HashSet<string>(run.Results.Select(r => r.Message.Arguments[0]).OrderBy(loc => loc));
```
lgolding: >.OrderBy(loc => loc) [](start = 119, length = 20)

Since you're making a `HashSet`, there's no reason to order the list.

michaelcfanning: Got it. FYI, i have updated this PR to merge files-array, will push an update in a moment.

src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs
```diff
@@ -87,15 +88,31 @@ protected void Verify(string testFileName, bool disablePrereleaseCompatibilityTr
...
+                    unexpectedNewIssues.Add(location);
+                }
+            }
+
+            // No new issues
+            unexpectedNewIssues.Count.Should().Be(0);
+
+            // No unexpectedly absent issues. If we found everything we expected,
+            // then our expected results should be empty
+            unexpectedlyAbsentResultLocations.Count.Should().Be(0);
```
lgolding: >unexpectedlyAbsentResultLocations [](start = 12, length = 33)

This whole thing can be done with a couple of LINQ calls. Also, a suggested variable rename for consistency:
```
IEnumerable<string> unexpectedNewResultLocations = actualResultLocations.Except(expectedResults.ResultLocationPointers);
unexpectedNewResultLocations.Count().Should().Be(0);

IEnumerable<string> unexpectedlyAbsentResultLocations = expectedResults.ResultLocationPointers.Except(actualResultLocations);
unexpectedlyAbsentResultLocations.Count().Should().Be(0)
```;

File: src/Sarif/Autogenerated/Rule.cs (hunk 1 above)
Comment (lgolding) [reply src/Sarif/Autogenerated/Rule.cs]: That makes sense. But I'm surprised this the code generator is not marking this property as virtual. Every other property is virtual. What's special about this one? (he asked rhetorically).

---
In reply to: [248136339](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248136339) [](ancestors = 248136339)

File: src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif (hunk 8 above)
Comment (lgolding) [reply src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif]: If I understand what's going on, the problem centers on the call to `PrereleaseCompatibilityTransformer.UpdateToCurrentVersion` in `FileDiffingTests.RunTest`. This call was _fixing_ the invalid `uri` property in this `fileLocation` object. By updating this file to the most current pre-release version, you avoided the file update altogether -- and that works because `RunTest` calls `PrereleaseCompatibilityTransformer.UpdateToCurrentVersion` with `forceUpdate: false` -- right?

---
In reply to: [248137397](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248137397) [](ancestors = 248137397)

File: src/Sarif/Autogenerated/Rule.cs (hunk 1 above)
Comment (michaelcfanning) [reply src/Sarif/Autogenerated/Rule.cs]: It could very well be for all I know, Rule.cs is getting picked up from NotYetAutogenerated for some other reason (or unnecessarily). I will take a look and get you a good answer, won't block here.

---
In reply to: [248368566](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248368566) [](ancestors = 248368566,248136339)

File: src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif (hunk 8 above)
Comment (michaelcfanning) [reply src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif]: Exactly right. 

---
In reply to: [248374537](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248374537) [](ancestors = 248374537,248137397)

Comment (lgolding) [reply src/Sarif.Multitool.FunctionalTests/TestData/UriBaseIdRequiresRelativeUri/Invalid.sarif]: In that case I'm going to have to revisit one of the fixes in my branch. One of the unit tests of `SarifVersionOneToCurrentVisitor` was failing because the pre-release transformer was _not_ updating one of the test files -- and I fixed it in my branch by changing `forceUpdate` to `true` in `PrereleaseCompatibilityTransformer.UpdateToCurrentVersion`! (Remember I sent you a message last night saying that I'd fixed a "bug" in `PrereleaseCompatibilityTransformer.UpdateToCurrentVersion ` :-))

So I'll change it back and change the test file instead.

---
In reply to: [248377880](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248377880) [](ancestors = 248377880,248374537,248137397)

File: src/Sarif/Autogenerated/Rule.cs (hunk 1 above)
Comment (michaelcfanning) [reply src/Sarif/Autogenerated/Rule.cs]: code gen isn't emitting these as virtual. we may have broken the connection between IRule/Rule that would result in this (as part of our partial extraction process). in which case, ok, expected. or maybe there's a jschema issue. i won't file a bug, because we will remember to go examine and attempt to review the notyetgenerated content as part of fixing other bugs, and we'll fix the blame then where it needs to be.

---
In reply to: [248377693](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248377693) [](ancestors = 248377693,248368566,248136339)

File: src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif (hunk 7 above)
Comment (lgolding) [reply src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif]: Yes, that's similar to the issue I mentioned in my comment on the other Invalid.sarif file that you changed. The "expected" file has a couple of instances of `"message": "blah"` instead of `"message": { "text": "blah" }`. And the `PreReleaseCompatibilityTransformer` is skipping the core transformations because it's called with `forceUpdate: false`, and so those bad messages are left alone, and so the file won't deserialize as v2.

---
In reply to: [248137193](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248137193) [](ancestors = 248137193)

File: src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs (hunk 11 above)
Comment (michaelcfanning) [reply src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs]: I considered this but I don't like leaving O(N^2) comparisons hanging around. The code as authored is much less compact/elegant but is clear and performant. 

---
In reply to: [248362142](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248362142) [](ancestors = 248362142)

Comment (lgolding) [reply src/Sarif.Multitool.FunctionalTests/Rules/ValidationSkimmerTestsBase.cs]: Would you at least do the variable rename?

---
In reply to: [248384220](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248384220) [](ancestors = 248384220,248362142)

File: src/Sarif.UnitTests/Processors/Log/SarifLogExtensionTests.cs [hunk 12]
```diff
@@ -70,30 +71,30 @@ public void RebaseUri_WorksAsExpected()
...
+
+                log.Runs.All(
+                    run =>
+                        run.Results == null ||
+                        run.Results.All(
+                            result =>
+                                result.Locations == null ||
+                                result.Locations.All(
+                                    location =>
+                                        location.PhysicalLocation.FileLocation.Uri.IsAbsoluteUri
```
lgolding: >l [](start = 40, length = 1)

I get that by eliminating the LINQ implicit loop over logs you're making it a little easier to find the problem. But why did you reverse (actually, not quite reverse, because it's still `&&` instead of `||`) the sense of this condition?

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -147,8 +147,8 @@ public void RebaseUriVisitor_VisitRun_DoesNotPatchFileDictionaryKeysWhenNotABase
         [Fact]
         public void RebaseUriVisitor_VisitFileData_PatchesParentUri()
         {
-            Uri rootfileUri = new Uri(@"file://C:/src/root/blah.zip#/stuff.doc");
-            Uri childFileUri = new Uri(@"/stuff.doc");
+            Uri rootfileUri = new Uri(@"file://C:/src/root/blah.zip");
+            Uri childFileUri = new Uri(@"/stuff.doc", UriKind.RelativeOrAbsolute);
```
lgolding: >RelativeOrAbsolute [](start = 62, length = 18)

Isn't it just plain relative?

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -158,7 +158,7 @@ public void RebaseUriVisitor_VisitFileData_PatchesParentUri()
                 {
                     new FileData { FileLocation = new FileLocation { Uri = rootfileUri, FileIndex = -0 }, ParentIndex = -1 },
```
lgolding: >- [](start = 100, length = 1)

While we're in here, "-0" => "0"

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -169,7 +169,7 @@ public void RebaseUriVisitor_VisitFileData_PatchesParentUri()
 
             run = rebaseUriVisitor.VisitRun(run);
 
-            run.Files[0].FileLocation.Uri.Should().Be(rootUriBaseId);            
+            run.Files[0].FileLocation.Uri.Should().Be("blah.zip");
```
lgolding: >blah [](start = 55, length = 4)

Let's also assert that `run.Files[0].FileLocation.UriBaseId.Should().Be("SRCROOT")`.

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
```diff
@@ -158,7 +158,7 @@ public void RebaseUriVisitor_VisitFileData_PatchesParentUri()
                 {
                     new FileData { FileLocation = new FileLocation { Uri = rootfileUri, FileIndex = -0 }, ParentIndex = -1 },
                     new FileData { FileLocation = new FileLocation { Uri = childFileUri, FileIndex = 1 }, ParentIndex = 0 },
-                    new FileData { FileLocation = new FileLocation { Uri = childFileUri, FileIndex = -1 }, ParentIndex = -1 }
+                    new FileData { FileLocation = new FileLocation { Uri = childFileUri, FileIndex = 2 }, ParentIndex = -1 }
```
lgolding: Why did you add this line?

michaelcfanning: I corrected the file index, which was wrong. have we addressed the matter of how the file index is populated for a file location within the files table itself? i.e., that it can either accurately reflect the actual index of its container or that it can be -1 (which isn't valid, the default, meaning it isn't expressed). or is this sufficiently clear in the spec, you think?

---
In reply to: [248386981](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248386981) [](ancestors = 248386981)

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
michaelcfanning: Not according to .NET. :)  which says, 'i can't make a determination'. leading slashes are problematic, in implementation in .NET, if not in the URI spec. 

---
In reply to: [248385717](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248385717) [](ancestors = 248385717)

src/Sarif.UnitTests/Visitors/RebaseUriVisitorTests.cs
lgolding: Sorry, I misread the diff, didn't realize this line was already there.

---
In reply to: [248387734](https://github.com/Microsoft/sarif-sdk/pull/1215#discussion_r248387734) [](ancestors = 248387734,248386981)

PR #1217, opened 2019-01-16
Fix Sarif.Driver and Sarif.Functional tests.

michaelcfanning:
@lgolding. After this, only five failing tests. And only one remaining failing test binary, Sarif.Converters.UnitTests.

### DIFF
--- a/src/Sarif.Converters/AndroidStudioConverter.cs
+++ b/src/Sarif.Converters/AndroidStudioConverter.cs
@@ -82,7 +82,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
         /// </returns>
         private IList<Result> ProcessAndroidStudioLog(XmlReader xmlReader)
         {
-            var results = new List<Result>();
+            var results = new HashSet<Result>(Result.ValueComparer);
 
             int problemsDepth = xmlReader.Depth;
             xmlReader.ReadStartElement(_strings.Problems);
@@ -98,7 +98,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             xmlReader.ReadEndElement(); // </problems>
 
-            return results;
+            return new List<Result>(results);
         }
 
         public Result ConvertProblemToSarifResult(AndroidStudioProblem problem)
--- a/src/Sarif.Driver.UnitTests/SimpleTestRule.cs
+++ b/src/Sarif.Driver.UnitTests/SimpleTestRule.cs
@@ -12,7 +12,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     [Export(typeof(IRule)), Export(typeof(IOptionsProvider)), Export(typeof(ISkimmer<TestAnalysisContext>))]
     internal class SimpleTestRule : TestRuleBase, IOptionsProvider
     {
-        public new string Id => "TEST002";
+        public override string Id => "TEST002";
 
         public override Message FullDescription { get { return new Message { Text = String.Empty }; } }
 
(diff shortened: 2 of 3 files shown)


src/Sarif.Driver.UnitTests/SimpleTestRule.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     [Export(typeof(IRule)), Export(typeof(IOptionsProvider)), Export(typeof(ISkimmer<TestAnalysisContext>))]
     internal class SimpleTestRule : TestRuleBase, IOptionsProvider
     {
-        public new string Id => "TEST002";
+        public override string Id => "TEST002";
```
michaelcfanning: >override [](start = 15, length = 8)

oops, use of new compromised ability of this code to perform correctly when access through an IRule reference.

src/Sarif.FunctionalTests/v2/ConverterTestData/AndroidStudio/AndroidStudio_issueLog5_raw.xml.sarif
```diff
@@ -653,13 +667,40 @@
...
+              "logicalLocationIndex": 0
+            }
+          ],
+          "properties": {
+            "severity": "TYPO",
+            "attributeKey": "TYPO"
+          }
+        },
+        {
+          "ruleId": "Typo",
```
michaelcfanning: >"ruleId": "Typo [](start = 10, length = 15)

Here's one of the duplicated issues.

src/Sarif.FunctionalTests/v2/ConverterTestData/AndroidStudio/AndroidStudio_issueLog5_raw.xml.sarif
```diff
@@ -1,50 +1,12 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-09-26",
-  "version": "2.0.0-csd.2.beta.2018-09-26",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-11-28",
+  "version": "2.0.0-csd.2.beta.2018-11-28",
```
michaelcfanning: >version [](start = 3, length = 7)

OK, mysterious. It turns out that this file, of all the converter test files, is the _only one_ that contains multiple results that are identical. This report file apparently finds multiple occurrences of typos on the same line. Only the line # is reported. Previously, our code only emitted a single result for this case. Now we're getting two. I have verified the source raw android studio has this duplication. I can't yet account for how/why this duplicated result wasn't getting added to the log file. Looking.

michaelcfanning: OK, figured it out. We made a bad change in the android studio converter, which was previously eliminating issue duplicated via a hashset. We now pass a list directly to some loggers, so this conversion to list allowed multiple identical issues to be generated. The results array, in fact, s/be unique. What could or should have happened here is that we added a second, duplicated location to the common result. Showing, we have one result, two occurrences which have identical locations.

I will open an SDK issue about preferring hashsets over lists for persisting results. 

---
In reply to: [248454813](https://github.com/Microsoft/sarif-sdk/pull/1217#discussion_r248454813) [](ancestors = 248454813)

src/Sarif.Converters/AndroidStudioConverter.cs
```diff
@@ -82,7 +82,7 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
         /// </returns>
         private IList<Result> ProcessAndroidStudioLog(XmlReader xmlReader)
         {
-            var results = new List<Result>();
+            var results = new HashSet<Result>(Result.ValueComparer);
```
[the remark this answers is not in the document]
michaelcfanning: yes, will do.

---
In reply to: [248475233](https://github.com/Microsoft/sarif-sdk/pull/1217#discussion_r248475233) [](ancestors = 248475233)

PR #1221, opened 2019-01-17
Update SarifVersionOneToCurrentVisitor for run.files array

lgolding:


lgolding:
>         internal IList<FileLocation> CreateResponseFilesList(IDictionary<string, string> responseFileToContentsDictionary)

Read this function as new code. Heavily reworked, only partially because of the files array. I think the new ordering of the logic is clearer.

---
Refers to: src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs:563 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -223,14 +221,22 @@ public static class PrereleaseCompatibilityTransformer
 
                     JObject tool = (JObject)run["tool"];
                     modifiedLog |= RenameProperty(tool, previousName: "fileVersion", newName: "dottedQuadFileVersion");
-                    if (tool != null && tool["language"] == null) { tool["language"] = "en-US"; }
+                    if (tool != null && tool["language"] == null)
+                    {
+                        tool["language"] = "en-US";
+                        modifiedLog = true;
```
lgolding: >modifiedLog [](start = 24, length = 11)

I noticed we missed a few of these. Maybe in future we can decorate the log with some sort of observer that notes any changes.

src/Sarif.UnitTests/Sarif.UnitTests.csproj
```diff
@@ -68,34 +68,34 @@
...
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\NotificationExceptionWithStack.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithBasicInvocation.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithFiles.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithInvocationAndNotifications.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithLogicalLocations.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithNotificationsButNoInvocations.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\OneRunWithRules.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\RestoreFromPropertyBag.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\v2\TwoResultsWithFixes.sarif" />
+    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\BasicResult.sarif" />
```
lgolding: >Inputs [](start = 59, length = 6)

Migrate to `FileDiffingTests` pattern.

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -75,7 +75,7 @@ protected virtual void RunTest(string inputResourceName, string expectedOutputRe
             bool passed;
             if (_testProducesSarifCurrentVersion)
             {
-                PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(expectedSarifText, forceUpdate: false, Formatting.Indented, out expectedSarifText);
+                PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(expectedSarifText, forceUpdate: true, Formatting.Indented, out expectedSarifText);
```
lgolding: >true [](start = 106, length = 4)

Some tests were failing with deserialization errors because without this, the core transformations didn't happen.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -21,6 +21,8 @@ public class SarifVersionOneToCurrentVisitor : SarifRewritingVisitorVersionOne
         private Run _currentRun;
         private RunVersionOne _currentV1Run;
         private int _threadFlowLocationNestingLevel;
+        private IDictionary<string, int> _v1FileKeytoV2IndexMap;
+        private IDictionary<Tuple<Uri, string>, int> _v2FileToIndexMap;
```
lgolding: >Tuple [](start = 28, length = 5)

I introduce this `Tuple` (which holds a `uri`, `uriBaseId` pair), rather than using `FileLocation` itself as the dictionary key. This is because when I create a new `FileLocation` object and try to decide if it's present in `run.files`, the `fileIndex` properties will never match. I just want to compare `uri` and `uriBaseId`. I could also have created a comparer.

I will add a comment about this.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -564,35 +570,53 @@ internal IList<FileLocation> CreateResponseFilesList(IDictionary<string, string>
...
-                        }
+                            Uri = new Uri(key, UriKind.RelativeOrAbsolute)
+                        };
+                    }
 
-#if FILES_ARRAY_WORKS
-                        if (!_currentRun.Files.ContainsKey(key))
+                    // If this response file has contents, add it to Run.Files, if it
+                    // isn't already there.
+                    if (!string.IsNullOrWhiteSpace(responseFileToContentsDictionary[key]))
```
michaelcfanning: >(responseFileToContentsDictionary[key] [](start = 50, length = 38)

you retrieve this content twice. why not cache, then set it later? see line 615

lgolding: Fixed.

---
In reply to: [248877992](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248877992) [](ancestors = 248877992)

michaelcfanning:
>   "$schema": "http://json.schemastore.org/sarif-1.0.0",

have we created copies of all these files? if so, would be nice perhaps to update the file diffing thing to allow for picking up inputs from some shared location.

---
Refers to: src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/Inputs/BasicResult.sarif:2 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
lgolding: Possibly. Let me think about it.

---
In reply to: [248879799](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248879799) [](ancestors = 248879799,248877564)

michaelcfanning:
>                     string.Equals(RemoveLineEndings(expected), RemoveLineEndings(assertion.Subject), StringComparison.OrdinalIgnoreCase)

this looks dangerous actually, i wonder if we pass tests without it. the problem is that many things are or s/be case-sensitive, e.g., file locations on unix platforms. property names in JSON. suggest changing this and seeing what breaks (i can take a look since this is a side issue)

---
Refers to: src/Sarif.TestUtilities/AssertionExtensions.cs:27 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

lgolding:
>   "$schema": "http://json.schemastore.org/sarif-1.0.0",

I didn't create copies in the course of this work, if that's what you're asking. If you're asking whether we could DRY out some test files, then possibly. Not in this change, for sure.

---
In reply to: [455370894](https://github.com/Microsoft/sarif-sdk/pull/1221#issuecomment-455370894) [](ancestors = 455370894)

---
Refers to: src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v1/BasicResult.sarif:2 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: or we could actually unit test the damn thing

---
In reply to: [248875783](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248875783) [](ancestors = 248875783)

lgolding:
>                     string.Equals(RemoveLineEndings(expected), RemoveLineEndings(assertion.Subject), StringComparison.OrdinalIgnoreCase)

Note that I did not change the code here; this is a method you introduced in one of your previous PRs. I just added the comment because I wanted to remember what it did.

---
In reply to: [455371193](https://github.com/Microsoft/sarif-sdk/pull/1221#issuecomment-455371193) [](ancestors = 455371193)

---
Refers to: src/Sarif.TestUtilities/AssertionExtensions.cs:27 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -223,14 +221,22 @@ public static class PrereleaseCompatibilityTransformer
...
+                    }
 
                     JObject conversion = (JObject)run["conversion"];
                     if (conversion != null)
```
michaelcfanning: > if (conversion != null) [](start = 19, length = 24)

this code and above could be DRY'ed out. i.e., author a little method PopulateValueIfNotAlreadyPresent(tool, "language", "en-us");

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: We could, but you can't test your way out of a fragile design ;-) The design requires us to remember to set this flag on every change. I sure don't want to get into this now, it's just something I noticed and spackled over while I was debugging through this code.

---
In reply to: [248881102](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248881102) [](ancestors = 248881102,248875783)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: ok.

---
In reply to: [248881391](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248881391) [](ancestors = 248881391)

src/Sarif.UnitTests/Visitors/SarifVersionOneToCurrentVisitorTests.cs
```diff
@@ -1,194 +1,69 @@
-﻿//// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT        
-//// license. See LICENSE file in the project root for full license information.
```
michaelcfanning: >//// [](start = 0, length = 4)

NOTE TO SELF: prefer multi-line C style code commenting in future to minimize code diffs when re-enabled. :P

src/Sarif.UnitTests/Visitors/SarifVersionOneToCurrentVisitorTests.cs
```diff
@@ -1,194 +1,69 @@
...
+            SarifLogVersionOne v1Log = JsonConvert.DeserializeObject<SarifLogVersionOne>(v1LogText, SarifTransformerUtilities.JsonSettingsV1Indented);
+            var transformer = new SarifVersionOneToCurrentVisitor();
+            transformer.VisitSarifLogVersionOne(v1Log);
+
+            SarifLog v2Log = transformer.SarifLog;
+            return JsonConvert.SerializeObject(v2Log, SarifTransformerUtilities.JsonSettingsIndented);
+        }
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_RestoreFromPropertyBag() => RunTest("RestoreFromPropertyBag.sarif");
```
michaelcfanning: > => RunTest("RestoreFromPropertyBag.sarif"); [](start = 76, length = 44)

Just my 2 cents, I actually don't like the collapsing you've done here. It's hard for my eye to separate the unique name of the test file in these single line constructs (which can get lengthy. I'd find a small method that isolates the call to RunTest much more readable. I assume use of this pattern won't complicate setting a breakpoint. it's certainly true that it will be more difficult to introduce one-off code changes to an individual test with this pattern. final point, it's only 67 lines of length, this file, still small after expansion. So I don't understand what the emphasis on compact code buys us.

src/Sarif.UnitTests/Visitors/SarifVersionOneToCurrentVisitorTests.cs
```diff
@@ -1,194 +1,69 @@
...
+        public void SarifTransformerTests_ToCurrent_OneRunWitRules() => RunTest("OneRunWithRules.sarif");
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_OneRunWithBasicInvocation() => RunTest("OneRunWithBasicInvocation.sarif");
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_OneRunWithInvocationAndNotifications() => RunTest("OneRunWithInvocationAndNotifications.sarif");
+
+        [Fact]
+        public void SarifTransformerTests_ToCurrent_OneRunWithNotificationsButNoInvocations() => RunTest("OneRunWithNotificationsButNoInvocations.sarif");
```
michaelcfanning: >public void SarifTransformerTests_ToCurrent_OneRunWithNotificationsButNoInvocations() => RunTest("OneRunWithNotificationsButNoInvocations.sarif"); [](start = 8, length = 146)

this line of code exceeds 130 chars in length as a result of use of this pattern. boo.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: whatever your thoughts on the design (agree there's a conversation here), it would be useful for us to be testing individual transformations that we add to this code with discrete test cases. that is a simple discipline that can avoid adding additional complexity. particularly for a potentially short-lived type (as we originally perceived this thing to be). 

---
In reply to: [248881460](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248881460) [](ancestors = 248881460,248881102,248875783)

michaelcfanning: fairly simple change might achieve the value you have in mind, simply create a helper for every mutation operation, and set the variable there. not much of an actual design change required.

---
In reply to: [248883476](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248883476) [](ancestors = 248883476,248881460,248881102,248875783)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -161,11 +160,10 @@ public static class PrereleaseCompatibilityTransformer
                     // Files are now persisted to an array. We will persist a mapping from
                     // previous file key to file index. We will associate the index with
                     // each result when we iterate over run.results later.
-                    var files = run["files"] as JObject;                    
 
-                    if (files != null)
+                    if (run["files"] is JObject files)
```
michaelcfanning: >run["files"] is JObject files [](start = 24, length = 29)

NIT: if you're going to introduce utilization of a new language pattern here, it is a common courtesy to scrub the code for other opportunities, rather than leaving things inconsistent. :) if you're not completely under the gun of course.

michaelcfanning:
>                     string.Equals(RemoveLineEndings(expected), RemoveLineEndings(assertion.Subject), StringComparison.OrdinalIgnoreCase)

Yes, that is clear in my diff utility which clearly shows you did not introduce a relevant change. ;) that is why i framed this as a 'side issue', by which I meant: not something relevant to the actual change you've introduced.

Do me a favor, though, while we're here: please change this value to Ordinal and then update the code comment to accurately reflect this change. I verified tests should continue to pass. I don't want to leave a case-insensitive comparison behind, that could be disastrous. in future, we can provide for case-sensitive and case-insensitive helpers if/when they are useful.

---
In reply to: [455371617](https://github.com/Microsoft/sarif-sdk/pull/1221#issuecomment-455371617) [](ancestors = 455371617,455371193)

---
Refers to: src/Sarif.TestUtilities/AssertionExtensions.cs:27 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

michaelcfanning:
>           "contents": "VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw==",

Given the fact that we're testing files array conversion, it seems to me our test content is awfully light on run.files constructs. I suggest you go find the following v2 SARIF file and use this as an expected output for your test suite (i.e., go author the speculative v1 file that will transform into this output). the v2 to v1 transform should also be tested with this content (which contains doubly nested things)

src\Sarif.UnitTests\TestData\PrereleaseCompatibilityTransformer\ExpectedOutputs\NestedFiles.sarif

---
Refers to: src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/Inputs/OneRunWithFiles.sarif:10 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif.TestUtilities/FileDiffingTests.cs
michaelcfanning: ok. suggest resetting this value to false. there are three inputs that are failing. go update their version to 10-10, which will provoke the transform. i think this is better than having test files that profess to be 11-28 but which clearly aren't

---
In reply to: [248880334](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248880334) [](ancestors = 248880334,248876087)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: I introduced `PopulatePropertyIfAbsent` which gets us part of the way there.

---
In reply to: [248884175](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248884175) [](ancestors = 248884175,248883476,248881460,248881102,248875783)

lgolding:
>                     string.Equals(RemoveLineEndings(expected), RemoveLineEndings(assertion.Subject), StringComparison.OrdinalIgnoreCase)

Fixed.

---
In reply to: [455375229](https://github.com/Microsoft/sarif-sdk/pull/1221#issuecomment-455375229) [](ancestors = 455375229,455371617,455371193)

---
Refers to: src/Sarif.TestUtilities/AssertionExtensions.cs:27 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

src/Sarif.TestUtilities/FileDiffingTests.cs
lgolding: Good idea. Two of the files actually were already 10-10 and needed to be downgraded to 09-26 to provoke the core transform, but all tests now pass with `forceUpdate: false`.

---
In reply to: [248887247](https://github.com/Microsoft/sarif-sdk/pull/1221#discussion_r248887247) [](ancestors = 248887247,248880334,248876087)

lgolding:
>           "contents": "VGhlIHF1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZw==",

Doing this now.

---
In reply to: [455376073](https://github.com/Microsoft/sarif-sdk/pull/1221#issuecomment-455376073) [](ancestors = 455376073)

---
Refers to: src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v1/OneRunWithFiles.sarif:10 in d94dbca. [](commit_id = d94dbcae5c6fbdede87956762e6269243a9e0a38, deletion_comment = False)

PR #1223, opened 2019-01-18
Basic rules transformation (except for v1 <-> v2 conversion)

michaelcfanning:
@lgolding, core rules array transformation work is done. I haven't made my own scan of this PR yet, wanted you to know that this is definitely landing today. Note that I've disabled the v1 <-> v2 testing for the change.


src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Base class for objects that can hold properties of arbitrary types.
     /// </summary>
-    public abstract class PropertyBagHolder : IPropertyBagHolder
+    public class PropertyBagHolder : IPropertyBagHolder
```
michaelcfanning: >class [](start = 11, length = 5)

not strictly a required change. i am observing odd cases where serialization isn't eliding default values, thought this might help. i don't see a need to retain this class as abstract, though, so left this in.

src/Sarif/Core/Rule.cs
```diff
@@ -13,5 +13,12 @@ public string Format(string messageId, IEnumerable<string> arguments)
         {
             return string.Format(CultureInfo.CurrentCulture, this.MessageStrings[messageId], arguments.ToArray());
         }
+
+        public bool ShouldSerializeDeprecatedIds()
```
michaelcfanning: >ShouldSerializeDeprecatedIds [](start = 20, length = 28)

the absence of this helper caused a test to fail. somehow the conversion from dictionary to array interacts with json.net default value handling. to be investigated further.

src/Sarif/Core/SarifRewritingVisitor.cs
```diff
@@ -1,85 +0,0 @@
...
-
-using System;
-using System.Linq;
-
-namespace Microsoft.CodeAnalysis.Sarif
-{
-    /// <summary>
-    /// Rewriting visitor for the Sarif object model.
-    /// </summary>
-    public abstract partial class SarifRewritingVisitor
```
michaelcfanning: >SarifRewritingVisitor [](start = 34, length = 21)

yay! all dictionaries converted to arrays, no need for this key visiting thing. _unless_ we decide it's required in future for things like graphs dictionary

michaelcfanning:
>                 // OriginalUriBaseIds are directories, not files. We'll disable this visit until the

btw - this is the only code that is requiring us to maintain a NotYetGenerated copy

---
Refers to: src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs:724 in 1707f72. [](commit_id = 1707f7277c660a7c4f2a8118e7f186152572b059, deletion_comment = False)

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/ExpectedOutputs/CoreTests_TextFiles.sarif
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-11-28",
-  "version": "2.0.0-csd.2.beta.2018-11-28",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
```
michaelcfanning: >http://json.schemastore.org/sarif-2.0.0 [](start = 14, length = 39)

need a version that forces the prerelease transform

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -27,9 +27,9 @@ public static string GetTestDirectory(string subdirectory = "")
 
         // Retrieving the source path of the tests is only used in developer ad hoc
         // rebaselining scenarios. i.e., this path won't be consumed by AppVeyor.
-        public static string GetProductTestDataDirectory(string subdirectory = "")
+        public static string GetProductTestDataDirectory(string testBinaryName, string subdirectory = "")
         {
-            return Path.GetFullPath(Path.Combine(@"..\..\..\..\..\src\Sarif.UnitTests\TestData", subdirectory));
+            return Path.GetFullPath(Path.Combine($@"..\..\..\..\..\src\{testBinaryName}\TestData", subdirectory));
```
michaelcfanning: >testBinaryName [](start = 72, length = 14)

baseline file location was previously always getting redirected to sarif.unit tests location.

src/Sarif.UnitTests/TestData/InsertOptionalDataVisitor/ExpectedOutputs/CoreTests_TextFiles.sarif
lgolding: Got it.

---
In reply to: [249130020](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249130020) [](ancestors = 249130020)

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/ComprehensiveFileProperties.sarif
```diff
@@ -29,7 +29,7 @@
       ],
       "results": [
         {
-          "ruleId": "C1",
+          "ruleIndex": 0,
```
lgolding: >ruleIndex [](start = 11, length = 9)

I don't think this is the desired outcome. I think you should always include `ruleId` for readability. As your "rule id collisions" test demonstrates, you do mention both `ruleId` and `ruleIndex` when there is a collision. I think you should do it even if there is no collision.

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/ComprehensiveFileProperties.sarif
```diff
@@ -69,13 +69,13 @@
         }
       ],
       "resources": {
-        "rules": {
-          "C1": {
+        "rules": [
+          {
```
lgolding: >{ [](start = 10, length = 1)

And now the `rule` object doesn't have an `id` property. In v1 the key was the id, unless there was a collision, so we should have `"id": "C1"` here.

src/Sarif/Core/PropertyBagHolder.cs
lgolding: Sure, fine.

---
In reply to: [249128719](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249128719) [](ancestors = 249128719)

lgolding:
>                 // OriginalUriBaseIds are directories, not files. We'll disable this visit until the

Progress!

---
In reply to: [455630304](https://github.com/Microsoft/sarif-sdk/pull/1223#issuecomment-455630304) [](ancestors = 455630304)

---
Refers to: src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs:724 in 1707f72. [](commit_id = 1707f7277c660a7c4f2a8118e7f186152572b059, deletion_comment = False)

src/Sarif.TestUtilities/FileDiffingTests.cs
lgolding: Nice generalization.

---
In reply to: [249130208](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249130208) [](ancestors = 249130208)

src/Sarif.Multitool.FunctionalTests/TestData/UriMustBeAbsolute/Invalid.sarif
```diff
@@ -40,7 +40,7 @@
             "runs[0].tool.downloadUri",
             "runs[0].results[0].workItemUris[0]",
             "runs[0].originalUriBaseIds.SRCROOT",
-            "runs[0].resources.rules.TST0001.helpUri",
+            "runs[0].resources.rules[0].helpUri",
```
lgolding: >runs[0].resources.rules[0].helpUri [](start = 13, length = 34)

I hope you are enjoying the easy maintainability :-)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -915,38 +924,37 @@ private void AddMessagesToResults()
         {
             foreach (Result result in _results)
             {
-                Rule rule;
-                if (_ruleDictionary.TryGetValue(result.RuleId, out rule))
-                {
-                    Message message = rule.ShortDescription ?? rule.FullDescription;
-                    Dictionary<string, string> replacements = null;
+                int ruleIndex = _ruleIdToIndexMap[result.RuleId];
```
lgolding: >RuleId [](start = 57, length = 6)

Don't you need a test similar to the old code, where you do `TryGetValue` on `_ruleIdToIndexMap`?

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -59,16 +59,9 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
-
-                foreach (Rule rule in rules)
-                {
-                    rulesDictionary[rule.Id] = rule;
-                }
-
                 run.Resources = new Resources
                 {
-                    Rules = rulesDictionary
+                    Rules = rules
```
lgolding: >rules [](start = 28, length = 5)

Well, _that's_ a nice simple change!

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -654,9 +654,10 @@ private void Visit(Resources resources, string resourcesPointer)
         {
             string rulesPointer = resourcesPointer.AtProperty(SarifPropertyName.Rules);
 
-            foreach (string key in resources.Rules.Keys)
+
+            for (int i = 0; i < resources.Rules.Count; ++i)
```
lgolding: Remove extra blank line.

src/Sarif.Converters/FortifyFprConverter.cs
michaelcfanning: don't believe so. i think the previous test was unnecessary, as the fortify report consistently contains rules metadata. will verify.

---
In reply to: [249143447](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249143447) [](ancestors = 249143447)

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/ComprehensiveFileProperties.sarif
michaelcfanning: hm, good catch

---
In reply to: [249136902](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249136902) [](ancestors = 249136902)

src/Sarif.Multitool.FunctionalTests/SarifMultitoolTestBase.cs
```diff
@@ -46,38 +46,38 @@ private static string MakeQualifiedFilePath(string testDirectory, string testFil
 
         private static void SelectiveCompare(SarifLog actualLog, SarifLog expectedLog)
         {
-            Result[] actualResults = SafeListToArray(actualLog.Runs[0].Results);
```
lgolding: >SafeListToArray [](start = 37, length = 15)

I don't think you can remove this. The spec says that `Runs` can be `null` if (for example) the tool couldn't start, or the database query that was supposed to return a set of runs was malformed. Removing this from _test_ code relies on us never having a test with a missing `Runs` array.

lgolding: How about `actualLog.Runs?[0].Results` -- checking to see if that's legal C#.

---
In reply to: [249150331](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249150331) [](ancestors = 249150331)

lgolding: I'm sorry, I'm confused, but I still have a point :-). You don't need the helper, because you're not converting a possibly-`null` list to an array. BUT! `Runs` might still be `null`. I guess it's ok to leave. On the day we add such a test, that will be time enough to test for that.

---
In reply to: [249150739](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249150739) [](ancestors = 249150739,249150331)

src/Sarif.Multitool.FunctionalTests/SarifMultitoolTestBase.cs
```diff
@@ -124,20 +124,14 @@ private static void SelectiveCompare(IDictionary<string, Rule> actualRules, IDic
...
                 actualRules.Count.Should().Be(expectedRules.Count);
 
-                foreach (string key in actualRules.Keys)
+                for (int i = 0; i < actualRules.Count; ++i)
                 {
-                    Rule actualRule = actualRules[key];
-                    Rule expectedRule;
-                    expectedRules.TryGetValue(key, out expectedRule).Should().BeTrue();
+                    Rule actualRule = actualRules[i];
+                    Rule expectedRule = expectedRules[i];
```
lgolding: >[i] [](start = 53, length = 3)

Ah, I see. Assuming I wrote this code, it was all based on my misconception that you couldn't index into a list. Bleah.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -308,7 +308,16 @@ public void Log(IRule rule, Result result)
 
             // TODO: we need to finish eliminating the IRule interface from the OM
             // https://github.com/Microsoft/sarif-sdk/issues/1189
-            Rules[result.RuleId] = (Rule)rule;
+            if (!RuleToIndexMap.TryGetValue((Rule)iRule, out int ruleIndex))
```
lgolding: >if [](start = 12, length = 2)

These 8 lines (311-318) occur twice. Consider DRYing them out with a help `LogRule(Rule rule, out int ruleIndex)`

lgolding:
>             _serializer.Serialize(_jsonWriter, rules);

Ha ha! The serializer didn't care, no change.

---
Refers to: src/Sarif/Writers/ResultLogJsonWriter.cs:234 in 1707f72. [](commit_id = 1707f7277c660a7c4f2a8118e7f186152572b059, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1248,10 +1248,13 @@
           }
         },
 
+
         "rules": {
```
lgolding: >rules [](start = 9, length = 5)

Specify `"default: []`

src/Sarif.Multitool.FunctionalTests/TestData/UriMustBeAbsolute/Invalid.sarif
michaelcfanning: indeed! much improved. :)

---
In reply to: [249139939](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249139939) [](ancestors = 249139939)

src/Sarif.Converters/FxCopConverter.cs
michaelcfanning: :)

---
In reply to: [249144017](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249144017) [](ancestors = 249144017)

src/Sarif.UnitTests/Visitors/UpdateIndicesVisitorTests.cs
```diff
@@ -41,7 +41,7 @@ public void UpdateIndicesVisitor_FunctionsWithNullMaps()
         {
             var result = _result.DeepClone();
 
-            var visitor = new UpdateIndicesVisitor(null, null);
+            var visitor = new UpdateIndicesVisitor(null, null, null);
```
lgolding: >UpdateIndicesVisitor [](start = 30, length = 20)

Consider: use named parameter syntax so we know what the `null`s are.

michaelcfanning:
>             _serializer.Serialize(_jsonWriter, rules);

Yes! if you elide the type argument, magic just happens,

---
In reply to: [455654495](https://github.com/Microsoft/sarif-sdk/pull/1223#issuecomment-455654495) [](ancestors = 455654495)

---
Refers to: src/Sarif/Writers/ResultLogJsonWriter.cs:234 in 1707f72. [](commit_id = 1707f7277c660a7c4f2a8118e7f186152572b059, deletion_comment = False)

src/Sarif.Multitool.FunctionalTests/TestData/MessagesShouldEndWithPeriod/Invalid.sarif
```diff
@@ -253,12 +253,12 @@
             "runs[0].invocations[0].configurationNotifications[0].message.richText",
             "runs[0].invocations[0].toolNotifications[0].message.text",
             "runs[0].invocations[0].toolNotifications[0].message.richText",
-            "runs[0].resources.rules.TST0001.shortDescription.text",
-            "runs[0].resources.rules.TST0001.shortDescription.richText",
-            "runs[0].resources.rules.TST0001.fullDescription.text",
-            "runs[0].resources.rules.TST0001.fullDescription.richText",
-            "runs[0].resources.rules.TST0001.messageStrings.Default",
-            "runs[0].resources.rules.TST0001.richMessageStrings.Default"
+            "runs[0].resources.rules[0].shortDescription.text",
```
lgolding: >rules[0] [](start = 31, length = 8)

I assume this path has an array, even though the `rules` property above is a dictionary, because this file goes through the `PrereleaseCompatibilityTransformer`.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -163,14 +163,15 @@ public override Message VisitMessage(Message node)
...
             {
-                Rule rule = null;
                 string formatString = null;
-                if (_ruleId != null &&
-                    _run.Resources?.Rules.TryGetValue(_ruleId, out rule) == true)
+                Rule rule = _ruleIndex != -1 ? _run.Resources.Rules[_ruleIndex] : null;
+
+                if (rule != null && rule.MessageStrings.TryGetValue(node.MessageId, out formatString))
                 {
+                    // Otherwise, pick it up from the
```
lgolding: >the [](start = 50, length = 3)

Unfinished comment.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -163,14 +163,15 @@ public override Message VisitMessage(Message node)
             if ((node.Text == null || _dataToInsert.HasFlag(OptionallyEmittedData.OverwriteExistingData)) &&
                 _dataToInsert.HasFlag(OptionallyEmittedData.FlattenedMessages))
             {
-                Rule rule = null;
                 string formatString = null;
-                if (_ruleId != null &&
-                    _run.Resources?.Rules.TryGetValue(_ruleId, out rule) == true)
+                Rule rule = _ruleIndex != -1 ? _run.Resources.Rules[_ruleIndex] : null;
+
+                if (rule != null && rule.MessageStrings.TryGetValue(node.MessageId, out formatString))
```
lgolding: >.T [](start = 55, length = 2)

Careful of `null`: `rule.MessageStrings?.TryGetValue`.

src/Sarif/Core/Rule.cs
```diff
@@ -13,5 +13,12 @@ public string Format(string messageId, IEnumerable<string> arguments)
...
+        {
+            return this.DeprecatedIds != null &&
+                (this.DeprecatedIds.Where((e) => { return e != null; }).Count() == this.DeprecatedIds.Count);
```
lgolding: >Where [](start = 36, length = 5)

`Count` takes a predicate (and my goodness, so many extra keystrokes in that lambda :-)):
```
this.DeprecatedIds.Count(e => e != null) == this.DeprecatedIds.Count()
```

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -163,19 +172,36 @@ public static class PrereleaseCompatibilityTransformer
...
                             ConstructFilesArray(
                                 files,
-                                keyToIndexMap);
+                                fileKeyToIndexMap);
 
-                        modifiedLog |= keyToIndexMap.Count > 0;
+                        modifiedLog |= fileKeyToIndexMap.Count > 0;
                     }
 
+                    if (run["resources"] is JObject resources)
```
lgolding: >resources [](start = 52, length = 9)

Yay!

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -163,19 +172,36 @@ public static class PrereleaseCompatibilityTransformer
...
+                    {
+                        // Remove 'open' from rule configuration default level enumeration
+                        // https://github.com/oasis-tcs/sarif-spec/issues/288
+                        modifiedLog |= RemapRuleDefaultLevelFromOpenToNote(resources);
```
lgolding: >RemapRuleDefaultLevelFromOpenToNote [](start = 39, length = 35)

_Love_ the method name!

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -163,19 +172,36 @@ public static class PrereleaseCompatibilityTransformer
...
+
+                        ruleKeyToIndexMap = new Dictionary<string, int>();
 
-                    var results = (JArray)run["results"];
-                    if (results != null)
+                        if (resources["rules"] is JObject rules)
+                        {
+                            resources["rules"] =
+                                ConstructRulesArray(
```
lgolding: >ConstructRulesArray [](start = 32, length = 19)

How about `ConvertRulesDictionaryToArray`. I had to do a quick glance upward to remember that variable `rules` is a dictionary at this point.

lgolding: For that matter, same thing for `ConstructFilesArray`.

---
In reply to: [249161968](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249161968) [](ancestors = 249161968)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -477,6 +532,7 @@ private static bool RemapRuleDefaultLevelFromOpenToNote(JObject resources)
                     // reasonable level to associate with this class of report, if it is emitted. In 
                     // practice, we don't expect that a current producer exists who is in this condition.
                     configuration["defaultLevel"] = "note";
+                    modifiedResources = true;
```
lgolding: >modifiedResources [](start = 20, length = 17)

*Biting my tongue* ;-)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
lgolding: Oh, it already existed.

---
In reply to: [249161656](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249161656) [](ancestors = 249161656)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -247,6 +256,52 @@ public static class PrereleaseCompatibilityTransformer
...
+                int index = keyValuePair.Value;
+                JObject updatedFileData = keyValuePair.Key;
+                updatedArrayElements[index] = updatedFileData;
+            }
+
+            return new JArray(updatedArrayElements);
+        }
+
+
+        private static void AddEntryToRuleToIndexMap(JObject rulesDictionary, string key, JObject file, Dictionary<JObject, int> jObjectToIndexMap, Dictionary<string, int> ruleKeyToIndexMap)
```
lgolding: >file [](start = 98, length = 4)

`file` => `rule`

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -247,6 +256,52 @@ public static class PrereleaseCompatibilityTransformer
...
+            {
+                AddEntryToRuleToIndexMap(
+                    rules,
+                    ruleEntry.Name,
+                    (JObject)ruleEntry.Value,
+                    jObjectToIndexMap,
+                    ruleKeyToIndexMap);
+            }
+
+            var updatedArrayElements = new JObject[jObjectToIndexMap.Count];
```
lgolding: >updatedArrayElements [](start = 16, length = 20)

Why not just `rulesArray`? This name makes it sounds like there were some array elements before, and now they're different.

src/Sarif/Core/Rule.cs
lgolding: Ok

---
In reply to: [249129151](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249129151) [](ancestors = 249129151)

lgolding:
Wow, reading my comments over I see I'm in full-on snark mode today ;-)

Anyway, most of my comments are easy point fixes. The only functional change I requested was always to populate `result.ruleId` (my comment in ComprehensiveFileProperties.sarif). Please let me know if you disagree.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -163,14 +163,15 @@ public override Message VisitMessage(Message node)
             if ((node.Text == null || _dataToInsert.HasFlag(OptionallyEmittedData.OverwriteExistingData)) &&
                 _dataToInsert.HasFlag(OptionallyEmittedData.FlattenedMessages))
             {
-                Rule rule = null;
                 string formatString = null;
-                if (_ruleId != null &&
-                    _run.Resources?.Rules.TryGetValue(_ruleId, out rule) == true)
+                Rule rule = _ruleIndex != -1 ? _run.Resources.Rules[_ruleIndex] : null;
```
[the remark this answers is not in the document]
lgolding: Darn it, I don't even think this works. There are messages in sub-objects of a result that are not rule messages. You will incorrectly take this branch, for example, for a code flow location message.

---
In reply to: [249157484](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249157484) [](ancestors = 249157484)

lgolding:
Re-reading my comments, I noticed a bug in `InsertOptionalDataVisitor` that's going to be tricky to fix.

---
In reply to: [455665596](https://github.com/Microsoft/sarif-sdk/pull/1223#issuecomment-455665596) [](ancestors = 455665596)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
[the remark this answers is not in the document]
lgolding: One possible fix: As soon as you see `_ruleIndex` is not -1, then use it _and immediately reset it_ to -1. Don't wait to come back from `base.VisitResult`.

---
In reply to: [249166259](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249166259) [](ancestors = 249166259,249157484)

michaelcfanning:
Thanks for the snark alert, will prepare myself. result.ruleId absolutely should be populated for readability, agree

---
In reply to: [455667726](https://github.com/Microsoft/sarif-sdk/pull/1223#issuecomment-455667726) [](ancestors = 455667726,455665596)

src/Sarif/Core/Rule.cs
lgolding: But is this the right test? You could just
```
return this.DeprecatedIds?.Count(e => e != null) > 0;
```
... and then persist the non-`null` ones.

---
In reply to: [249160403](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249160403) [](ancestors = 249160403)

src/Sarif.Converters/FortifyFprConverter.cs
michaelcfanning: verified this data s.be present, no need for check

---
In reply to: [249147574](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249147574) [](ancestors = 249147574,249143447)

src/Sarif.UnitTests/Visitors/UpdateIndicesVisitorTests.cs
michaelcfanning: right, when every argument is null in a test named 'FunctionsWithNullMap', i don't see the need for this readability.  the need for naming comes during actual utilization. 

things that need to be readable need to be readable. otherwise, they should be compact. still don't care enough about this to push back.

---
In reply to: [249155287](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249155287) [](ancestors = 249155287)

michaelcfanning: I took a look at the fix and it actually, to me, compromises the readability of this test. the test is intended to comprehensively validate a call where all args are null. the call without named parameters makes it easiest to chunk the information allowing this condition to be validated. 

---
In reply to: [249187800](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249187800) [](ancestors = 249187800,249155287)

src/Sarif.Multitool.FunctionalTests/TestData/MessagesShouldEndWithPeriod/Invalid.sarif
michaelcfanning: that's right. i didn't care to tinker partially with this file. better to tackle all these comprehensively later when appropriate.

---
In reply to: [249156600](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249156600) [](ancestors = 249156600)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: nice catch. what a fantastic corner case. adding unit test.

---
In reply to: [249157825](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249157825) [](ancestors = 249157825)

src/Sarif/Core/Rule.cs
michaelcfanning: yes, greatly  reduced, like it.

---
In reply to: [249182441](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249182441) [](ancestors = 249182441,249160403)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Prudent. And still your point is made. :) We'll get this fixed.

---
In reply to: [249162790](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249162790) [](ancestors = 249162790)

michaelcfanning: Fixed later, I mean. :)

---
In reply to: [249219447](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249219447) [](ancestors = 249219447,249162790)

src/Sarif/Core/Rule.cs
```diff
@@ -13,5 +13,10 @@ public string Format(string messageId, IEnumerable<string> arguments)
         {
             return string.Format(CultureInfo.CurrentCulture, this.MessageStrings[messageId], arguments.ToArray());
         }
+
+        public bool ShouldSerializeDeprecatedIds()
+        {
+            return this.DeprecatedIds?.Count(e => e != null) > 1;
```
lgolding: >1 [](start = 63, length = 1)

Shouldn't this be `> 0` ?

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
lgolding: I don't think this is a sufficient fix. It assumes that a message id that points into a rule's message strings will never be the same as a message id that points into `run.resources.messageStrings`. It's easy to imagine message ids like `"default"` or `"success"` that might occur in both places (and even if it weren't easy to imagine, it might still happen).

What do you think of the fix I suggested above (reset `_ruleIndex` to -1 as soon as you use it)?

---
In reply to: [249194811](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249194811) [](ancestors = 249194811,249157825)

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -10,11 +10,40 @@ public class UpdateIndicesVisitor : SarifRewritingVisitor
...
+        {
+            if (_ruleKeyToIndexMap != null)
+            {
+                if (_ruleKeyToIndexMap.TryGetValue(node.RuleId, out int ruleIndex))
+                {
+                    node.RuleIndex = ruleIndex;
+
+                    // We need to update the rule id, as it previously referred to a synthesized 
+                    // key that resolved some collision in the resources.rules collection.
+                    node.RuleId = _resources.Rules[ruleIndex].Id;
```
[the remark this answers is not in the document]
lgolding: ok, that sounds like a reasonable workflow to require.

---
In reply to: [249204481](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249204481) [](ancestors = 249204481,249159879)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
lgolding: See my comment in the unit tests, where I describe the failure case I'm worried about.

---
In reply to: [249229167](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249229167) [](ancestors = 249229167,249194811,249157825)

michaelcfanning: I don't understand your suggestion. When we are visiting a result with a corresponding resultIndex, we retain it. If we reset this to -1 on first use, as you suggest, no other message visiting in context of the result would have access to this data. This seems like entirely the wrong thing to so. Remember: stacks, codeflows, graphs, fixes, etc. might dip into the rule resources to get format string.

We explicitly designed string look-up to refer to the rule first, then fall back to the global. Maybe your objections will be clearer to me when i look at your unit test remarks.

[https://github.com/oasis-tcs/sarif-spec/issues/216](https://github.com/oasis-tcs/sarif-spec/issues/216)
"in the presence of a non-null message.messageId, consult resources.rules by rule id, to attempt to locate the string. if not present, fall back to resources for look-up."

---
In reply to: [249229167](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249229167) [](ancestors = 249229167,249194811,249157825)

src/Sarif/Core/Rule.cs
michaelcfanning: Oops. All changes introduce change of regression, even small ones. :)

---
In reply to: [249228972](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249228972) [](ancestors = 249228972)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,6 +110,103 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+                        }
+                    }
+                }
+            };
+
+            var visitor = new InsertOptionalDataVisitor(OptionallyEmittedData.FlattenedMessages);
+            visitor.Visit(run);
+
+            run.Results[0].Message.Text.Should().Be(globalMessageValue);
+        }
```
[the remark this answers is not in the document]
michaelcfanning: I don't agree. A result is strongly associated with a rule. Every string look-up in context of that result should refer to the rules table first. The global table is intended to hold strings that apply to all rules. A fix may utilize a global string, sure, as anything might. But it is more likely that the fix itself has user-facing text that applies to the rule (every rule remediation is likely to differ in some aspects).

It looks like the spec may not be accurate on this. The bug I filed on it in September describes the correct look-up semantics. I will file a new issue

---
In reply to: [249229879](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249229879) [](ancestors = 249229879)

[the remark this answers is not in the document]
michaelcfanning: [https://github.com/oasis-tcs/sarif-spec/issues/310](https://github.com/oasis-tcs/sarif-spec/issues/310)

---
In reply to: [249290961](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249290961) [](ancestors = 249290961,249229879)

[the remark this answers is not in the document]
lgolding: Regardless of whether we agree on the intended outcome, we can agree on two things this is the test case we want.

You say "A fix may utilize a global string.". How would it do that?

---
In reply to: [249290961](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249290961) [](ancestors = 249290961,249229879)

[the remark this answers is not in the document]
michaelcfanning: Gotcha and agree. :) my remark on fix is a general comment, yes, nearly anything could use a global string. to do so, there must not be a collision between strings in the current rule (i.e., the rule associated with the current result) and the global table. 

in any case, you're right, this is a test we need, will take care of it.

---
In reply to: [249291338](https://github.com/Microsoft/sarif-sdk/pull/1223#discussion_r249291338) [](ancestors = 249291338,249290961,249229879)

PR #1226, opened 2019-01-20
Add files array tests for nested files and uriBaseIds

lgolding:


src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -214,9 +208,7 @@ internal FileLocation CreateFileLocation(Uri uri, string uriBaseId)
 
             if (uri != null)
             {
-                var fileToIndexKey = new Tuple<Uri, string>(uri, uriBaseId);
-
-                if (_v2FileToIndexMap.TryGetValue(fileToIndexKey, out int fileIndex))
+                if (_v1FileKeytoV2IndexMap.TryGetValue(uri.OriginalString, out int fileIndex))
```
michaelcfanning: >_v1FileKeytoV2IndexMap [](start = 20, length = 22)

i see. basically use the v1 as a key to the new indices, makes sense

PR #1228, opened 2019-01-21
Implement v2->v1 conversion for rules dictionary

lgolding:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1274,7 +1281,7 @@
       "properties": {
 
         "ruleId": {
-          "description": "The stable, unique identifier of the rule (if any) to which this notification is relevant. This member can be used to retrieve rule metadata from the rules dictionary, if it exists.",
```
lgolding: >( [](start = 68, length = 1)

This is the only place we used parens instead of commas on "if any".

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -23,8 +23,11 @@ public class SarifCurrentToVersionOneVisitor : SarifRewritingVisitor
         private RunVersionOne _currentRun = null;
 
         // To understand the purpose of these fields, see the comment on CreateFileKeyIndexMappings.
-        private IDictionary<string, int> _v1FileKeyToV2IndexDictionary;
-        private IDictionary<int, string> _v2FileIndexToV1KeyDictionary;
+        private IDictionary<string, int> _v1FileKeyToV2IndexMap;
```
lgolding: For consistency with v1->v2 transformer. You preferred Map to Dictionary.

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/ExpectedOutputs/ResultLocations.sarif
```diff
@@ -101,14 +101,14 @@
         }
       ],
       "rules": {
-        "WEB1079.AttributeValueIsNotQuoted": {
+        "WEB1079": {
```
lgolding: >WEB1079 [](start = 9, length = 7)

When we were converting a dictionary of v2 rules to a dictionary of v1 rules, we could preserve the v2 keys in the v1 dictionary. Now that the `PreReleaseCompatibilityTransformer` is converting the dictionary to an array, the keys are lost before the v2-to-v1 transformer is invoked.

In fact, the v2 format has no place for this "friendly compound id". If you want a place for it, we could add it to v2 rule.

michaelcfanning: Oh, right. These terms should be persisted to rule.name, can you go ahead and do that? still an open question whether we document using rule.id and rule.name in this way to compose a compound id (with an opaque stable and instable readable component in well known locations).

---
In reply to: [249572234](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249572234) [](ancestors = 249572234,249559914)

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/ExpectedOutputs/TwoResultsWithFixes.sarif
```diff
@@ -156,14 +156,14 @@
...
       "rules": {
-        "WEB1079.AttributeValueIsNotQuoted": {
+        "WEB1079": {
           "id": "WEB1079",
           "shortDescription": "The attribute value is not quoted.",
           "messageFormats": {
             "default": "The  value of the '{0}' attribute is not quoted. Wrap the attribute value in single or double quotes."
           }
         },
-        "WEB1066.TagNameIsNotLowercase": {
```
michaelcfanning: >TagNameIsNotLowercase [](start = 17, length = 21)

same note, move the readable names to rule.name: NIT

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithRules.sarif
```diff
@@ -9,25 +9,29 @@
       },
       "results": [
         {
-          "ruleId": "C2009",
+          "ruleId": "C2001",
+          "ruleIndex":  0,
```
michaelcfanning: >ruleIndex [](start = 11, length = 9)

why couldn't the prerelease transformer fix things up for you? i.e., was there a reason you had to introduce the ruleIndex object? generally, i think we should avoid piecemeal upgrade of these files to current format.

src/Sarif.UnitTests/Visitors/SarifCurrentToVersionOneVisitorTests.cs
```diff
@@ -44,23 +44,21 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
         [Fact]
         public void SarifTransformerTests_ToVersionOne_OneRunWithFiles() => RunTest("OneRunWithFiles.sarif");
 
-#if TRANSFORM_CODE_AUTHORED
```
michaelcfanning: >TRANSFORM_CODE_AUTHORED [](start = 4, length = 23)

yay!

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -1032,6 +1014,81 @@ private static string CreateFileDictionaryKey(FileData v2File, IList<FileData> v
...
+                {
+                    Rule rule = rules[i];
+                    if (rule.Id != null)
+                    {
+                        ruleIdToCountMap[rule.Id] = ruleIdToCountMap.ContainsKey(rule.Id)
+                            ? ruleIdToCountMap[rule.Id] + 1
+                            : 1;
+
+                        v2RuleIndexToV1KeyMap[i] = ruleIdToCountMap[rule.Id] == 1
+                            ? null
```
michaelcfanning: why store null in some cases? instead of just persisting the rule id? then look-up just works exclusively against this look-up. or perhaps you use the null case to indicate something special later?

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -1032,6 +1014,81 @@ private static string CreateFileDictionaryKey(FileData v2File, IList<FileData> v
...
+
+            if (rules != null)
+            {
+                // Keep track of how many distinct rules have each id.
+                var ruleIdToCountMap = new Dictionary<string, int>();
+
+                for (int i = 0; i < rules.Count; ++i)
+                {
+                    Rule rule = rules[i];
+                    if (rule.Id != null)
```
michaelcfanning: >if (rule.Id != null) [](start = 20, length = 20)

ok, so in old SARIF, if you had no rule.Id, you had no look-up key, so there won't be any coupling between results and rule metadata for these checks.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -1032,6 +1014,81 @@ private static string CreateFileDictionaryKey(FileData v2File, IList<FileData> v
...
+
+            return v2RuleIndexToV1KeyMap;
+        }
+
+        private static string GetV1RuleKeyFromV2Index(
+            int ruleIndex,
+            IDictionary<int, string> v2RuleIndexToV1KeyMap)
+        {
+            string ruleKey = null;
+            if (0 <= ruleIndex && ruleIndex < v2RuleIndexToV1KeyMap?.Count)
```
michaelcfanning: >0 <= ruleIndex [](start = 16, length = 14)

can you just make this ruleIndex > -1 perhaps? your comparison here is confusing. if you retain it, please comment the code.

michaelcfanning: >ruleIndex < v2RuleIndexToV1KeyMap [](start = 34, length = 33)

wait, are you sure about this? v2RuleIndexToV1 is a map, and we skip rule entries with no ids. So I don't think a comparison to the map count is right.

Consider a rules array with 49 checks with no ids. the 50th does have a rule idea and we have a single SARIF result pointing to it with ruleIndex == 49. your v2RuleIndexToV1KeyMap will only have a single entry right? and so we'll fail this test. 

I don't understand all these hijinks in any case. why can't you just look for ruleIndex and try to find the key? it will be set to null in all failures (which is right). All your code assumes that the map may have values in it for things like ruleIndex <0 (which we want to avoid retrieving). why would that be?

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/ExpectedOutputs/ResultLocations.sarif
lgolding: That's a good idea and makes sense for this file. The problem is that the dictionary key is not _always_ a friendly name (consider `TST0001-2`) and there's no way to tell whether it was intended as a friendly name. Thoughts?

---
In reply to: [249572657](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249572657) [](ancestors = 249572657,249572234,249559914)

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/ExpectedOutputs/TwoResultsWithFixes.sarif
lgolding: See my response to your other comment on this issue.

---
In reply to: [249572746](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249572746) [](ancestors = 249572746)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: I've got a principled objection to treating a sentinel value numerically just because it happens to be ordered properly. `ruleIndex != -1` ("it's not the sentinel!") is tempting, but it lets `-2` slip through. I like `ruleIndex >= 0` because it states the validity condition independent of what we happened to choose as the sentinel.

Or were you objecting to the idiom `min <= x && x < limit` (which I occasionally use to mimic the math notation `min < x < limit`)?

---
In reply to: [249574859](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249574859) [](ancestors = 249574859)

lgolding: You're right that this is a bug. And you're right that there's no need for the condition at all. I don't know what I was thinking.

Then the question is whether to get rid of this helper altogether. It has the advantage of returning its value rather than using an `out` parameter. I think I'll keep it.

---
In reply to: [249575749](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249575749) [](ancestors = 249575749)

lgolding: Marking this one "By Design" (because I really do like the comparison the way I wrote it), but marking the next one "Resolved" because you're right, we don't need the comparison at all.

---
In reply to: [249613504](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249613504) [](ancestors = 249613504,249574859)

src/Sarif.UnitTests/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithRules.sarif
lgolding: `FileDiffingTests` runs the prerelease transformer on the _expected output_ (except in the case of tests that, like this one, produce SARIF v1), not on the input.

---
In reply to: [249573084](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249573084) [](ancestors = 249573084)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
lgolding: BUT! I found what I think is a clearer way to express this logic. Now, I _do_ set the map entry to `ruleId` rather than `null`, and I _don't_ need the `null` check (and its comment) in `ConvertRulesArrayToDictionary`. Instead, I do the `null` check right at the point where we care about its effect: at the point where I assign `result.RuleKey`.

See the last commit.

---
In reply to: [249616588](https://github.com/Microsoft/sarif-sdk/pull/1228#discussion_r249616588) [](ancestors = 249616588,249615764,249573985)

PR #1229, opened 2019-01-21
Notification rule index

michaelcfanning:
@lgolding  As discussed, notification messages pull purely from the global strings namespace in run.resources

### DIFF
--- a/src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
+++ b/src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
@@ -110,101 +110,241 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                 OptionallyEmittedData.FlattenedMessages);
         }
 
-        [Fact]
-        public void InsertOptionalDataVisitorTests_FlattensGlobalMessageString()
-        {
-            string ruleId = nameof(ruleId);
-            string globalMessageId = nameof(globalMessageId);
-            string globalMessageValue = nameof(globalMessageValue);
+        private const int RuleIndex = 0;
+        private const string RuleId = nameof(RuleId);
+        private const string NotificationId = nameof(NotificationId);
 
+        private const string SharedMessageId = nameof(SharedMessageId);
+        private const string SharedKeyRuleMessageValue = nameof(SharedKeyRuleMessageValue);
+        private const string SharedKeyGlobalMessageValue = nameof(UniqueGlobalMessageValue);
+
+        private const string UniqueRuleMessageId = nameof(UniqueRuleMessageId);
+        private const string UniqueRuleMessageValue = nameof(UniqueRuleMessageValue);
+
+        private const string UniqueGlobalMessageId = nameof(UniqueGlobalMessageId);
+        private const string UniqueGlobalMessageValue = nameof(UniqueGlobalMessageValue);
+
+        private static Run CreateBasicRunForMessageStringLookupTesting()
+        {
+            // Returns a run object that defines unique string instances both
+            // for an individual rule and in the global strings object. Also
+            // defines values for a key that is shared between the rule object
+            // and the global table. Used for evaluating string look-up semantics.
             var run = new Run
             {
-                Results = new List<Result>
-                {
-                    new Result
-                    {
-                        RuleId = ruleId,
-                        RuleIndex = 0,
-                        Message = new Message
-                        {
-                             MessageId = globalMessageId
-                        }
-                    }
-                },
+                Results = new List<Result> { }, // add non-null collections for convenience
+                Invocations = new List<Invocation> { new Invocation { } },
                 Resources = new Resources
                 {
                     MessageStrings = new Dictionary<string, string>
                     {
-                        [globalMessageId] = globalMessageValue
+                        [UniqueGlobalMessageId] = UniqueGlobalMessageValue,
+                        [SharedMessageId] = SharedKeyGlobalMessageValue
                     },
                     Rules = new List<Rule>
                     {
                         new Rule
                         {
-                            Id = ruleId
+                            Id = RuleId,
+                            MessageStrings = new Dictionary<string, string>
+                            {
+                                [UniqueRuleMessageId] = UniqueRuleMessageValue,
+                                [SharedMessageId] = SharedKeyRuleMessageValue
+                            }
                         }
                     }
                 }
             };
 
+            run.Invocations[0].ToolNotifications = new List<Notification>();
+            run.Invocations[0].ConfigurationNotifications = new List<Notification>();
+
+            return run;
+        }
+
+        [Fact]
+        public void InsertOptionalDataVisitorTests_FlattensMessageStringsInResult()
+        {
+            Run run = CreateBasicRunForMessageStringLookupTesting();
+
+            run.Results.Add(
+                new Result
+                {
+                    RuleId = RuleId,
+                    RuleIndex = RuleIndex,
+                    Message = new Message
+                    {
+                        MessageId = UniqueGlobalMessageId
+                    }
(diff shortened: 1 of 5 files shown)


src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
-            string ruleId = nameof(ruleId);
-            string globalMessageId = nameof(globalMessageId);
-            string globalMessageValue = nameof(globalMessageValue);
+        private const int RuleIndex = 0;
+        private const string RuleId = nameof(RuleId);
+        private const string NotificationId = nameof(NotificationId);
+
+        private const string SharedMessageId = nameof(SharedMessageId);
+        private const string SharedKeyRuleMessageValue = nameof(SharedKeyRuleMessageValue);
+        private const string SharedKeyGlobalMessageValue = nameof(UniqueGlobalMessageValue);
```
lgolding: >UniqueGlobalMessageValue [](start = 66, length = 24)

I assume this is a copy/paste error and you intended `nameof(SharedKeyGlobalMessageValue)`. Now I have to figure out why it doesn't make your tests fail.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+
+        private const string UniqueRuleMessageId = nameof(UniqueRuleMessageId);
+        private const string UniqueRuleMessageValue = nameof(UniqueRuleMessageValue);
+
+        private const string UniqueGlobalMessageId = nameof(UniqueGlobalMessageId);
+        private const string UniqueGlobalMessageValue = nameof(UniqueGlobalMessageValue);
 
+        private static Run CreateBasicRunForMessageStringLookupTesting()
+        {
+            // Returns a run object that defines unique string instances both
```
lgolding: >Returns [](start = 15, length = 7)

This is a good comment.

Ultra-nit: IMO it is a method summary comment so it belongs outside the method body. (I am _not_ suggested XML docs.)

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+
+            var visitor = new InsertOptionalDataVisitor(OptionallyEmittedData.FlattenedMessages);
+            visitor.Visit(run);
+
+            run.Results[0].Message.Text.Should().Be(UniqueGlobalMessageValue);
+            run.Results[1].Message.Text.Should().Be(UniqueRuleMessageValue);
+
+            // Prefer rule-specific value in the event of a message id collision
+            run.Results[2].Message.Text.Should().Be(SharedKeyRuleMessageValue);
+        }
```
lgolding: >} [](start = 8, length = 1)

Yes, perfect test.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+            // Shared message id with no overriding rule id
+            toolNotifications.Add(
+                new Notification
+                {
+                    Id = NotificationId,
+                    Message = new Message {  MessageId = SharedMessageId}
+                });
+            configurationNotifications.Add(toolNotifications[0]);
+
+            // Shared message id with an overriding rule id. This message
```
lgolding: >Shared message id with an overriding rule id. [](start = 15, length = 45)

How about this for the first sentence: "Notification that refers to a rule that contains a message with the same id as the specified notification message id." The second sentence still works.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+            toolNotifications.Add(
+                new Notification
+                {
+                    Id = NotificationId,
+                    RuleIndex = RuleIndex,
+                    Message = new Message { MessageId = SharedMessageId }
+                });
+            configurationNotifications.Add(toolNotifications[1]);
+
+            toolNotifications.Add(
```
lgolding: >Add [](start = 30, length = 3)

Comment: "Notification that refers to a rule that does _not_ contain a message with the same id as the specified notification id. In this case it is no surprise that the message comes from the global string table."

... and then I think the test reads most naturally if you reverse the order of notifications 1 and 2.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
-            run.Results[0].Fixes[0].Description.Text.Should().Be(globalMessageValue);
+            run.Results[0].Fixes[0].Description.Text.Should().Be(UniqueGlobalMessageValue);
+            run.Results[0].Fixes[1].Description.Text.Should().Be(UniqueRuleMessageValue);
+
+            // Prefer rule-specific value in the event of a message id collision
+            run.Results[0].Fixes[2].Description.Text.Should().Be(SharedKeyRuleMessageValue);
+
+            // Prefer global value in the event of no rules metadata
+            run.Results[1].Fixes[0].Description.Text.Should().Be(SharedKeyGlobalMessageValue);
         }
```
lgolding: >} [](start = 8, length = 1)

Another perfect test.

lgolding:
Really nice. A few ultra-nits. Merge at will.

src/Sarif.UnitTests/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -110,101 +110,230 @@ public void InsertOptionalDataVisitor_PersistsAll()
...
+                                [UniqueRuleMessageId] = UniqueRuleMessageValue,
+                                [SharedMessageId] = SharedKeyRuleMessageValue
+                            }
                         }
                     }
                 }
             };
 
+            run.Invocations[0].ToolNotifications = new List<Notification>();
+            run.Invocations[0].ConfigurationNotifications = new List<Notification>();
```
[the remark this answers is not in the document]
michaelcfanning: Doing this introduces 6 new lines of vertical space, which is why I avoided it. Why doesn't this provoke your 'chunking' need? :) I did take the note, btw

---
In reply to: [249608580](https://github.com/Microsoft/sarif-sdk/pull/1229#discussion_r249608580) [](ancestors = 249608580)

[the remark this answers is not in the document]
lgolding: Everything is a tradeoff. The question is, what do I get back from those 6 lines? For me, in this case, the answers are:

1. Conceptual grouping: The object initialization happens entirely within the object initializer braces. There are no surprises later ("Oh! It wasn't completely initialized after all!")
2. Locality of reference: As the code stands, we create the `Invocation` object on L136, but we don't initialize it until LL159-160.
3. DRY: We don't want to repeat the prefix `run.Invocations[0]`.
4. Readability: The reader doesn't have to parse the nested member access expressions to discover that they really _are_ just part of the object initialization.

---
In reply to: [249836127](https://github.com/Microsoft/sarif-sdk/pull/1229#discussion_r249836127) [](ancestors = 249836127,249608580)

PR #1232, opened 2019-01-23
Implement v1->v2 conversion for rules dictionary

lgolding:


lgolding:

src/Sarif/Visitors/UpdateIndicesVisitor.cs
```diff
@@ -27,7 +27,7 @@ public override Result VisitResult(Result node)
         {
             if (_ruleKeyToIndexMap != null)
             {
-                if (_ruleKeyToIndexMap.TryGetValue(node.RuleId, out int ruleIndex))
+                if (node.RuleId != null &&_ruleKeyToIndexMap.TryGetValue(node.RuleId, out int ruleIndex))
```
lgolding: >node.RuleId [](start = 20, length = 11)

One of the test cases has a result with a `ruleKey` but no `ruleId`. That is valid SARIF v1, and we need to protect against it.

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/TwoResultsWithFixes.sarif
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-11-28",
-  "version": "2.0.0-csd.2.beta.2018-11-28",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0",
+  "version": "2.0.0",
```
lgolding: >"2.0.0" [](start = 13, length = 7)

Force pre-release compatibility transformation.

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -68,14 +71,13 @@
               "text": "Rent internal rebellion competence biography photograph."
             },
             "configuration": {
-              "defaultLevel": "note"
+              "defaultLevel": "note",
+              "defaultRank": 0.0
             }
-          },
-          "C2002-1": {
```
lgolding: >"C2002-1" [](start = 10, length = 9)

The v1 file has a `result` with a `ruleKey` that does not exist in the `rules` dictionary. The visitor code was constructing an empty `rule` object for that key to refer to. I don't see the point of that. The new code just sets the v2 `ruleIndex` to -1 in this case. See how on the right-hand side of this diff, the `result` starting on L28 has no `ruleIndex`. The old code would have given it a synthetic one.

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -45,18 +47,19 @@
...
+          {
             "id": "C2002",
             "fullDescription": {
               "text": "Catfish season continuous hen lamb include dose copy grant."
             },
             "configuration": {
               "enabled": true,
-              "defaultLevel": "error"
+              "defaultLevel": "error",
+              "defaultRank": 0.0
```
lgolding: >defaultRank [](start = 15, length = 11)

I do not know why the serializer is emitting `defaultRank` with its default value, but I wouldn't hold this PR to look at it. Considering this occurs in the v1->v2 transform, I don't think we should _ever_ take the time to look at it. If something similar shows up in direct production, that will be the time to look at it.

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-10-10",
-  "version": "2.0.0-csd.2.beta.2018-10-10",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-11-28",
+  "version": "2.0.0-csd.2.beta.2018-11-28",
```
lgolding: >28 [](start = 39, length = 2)

The prerelease compatibility transformer, although it was performing correctly, masked some of the behavior this test needed to demonstrate. We can talk about the details offline (assuming I haven't forgotten them by then).

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -732,51 +732,11 @@ internal Result CreateResult(ResultVersionOne v1Result)
...
 
-                if (v1Result.RuleKey == null)
-                {
-                    result.RuleId = v1Result.RuleId;
-                }
-                else
-                {
-                    if (v1Result.RuleId == null)
-                    {
-                        result.RuleId = v1Result.RuleKey;
```
lgolding: >result [](start = 24, length = 6)

This was never right. It was never valid to assign a `ruleKey` to a `ruleId`. The situation "`ruleId == null; ruleKey != null`" meant "This rule has no id (which is perfectly valid, both in v1 and v2). If you want to know more about this rule (like its description), you can find the rule metadata here."

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -10,30 +10,32 @@
...
           "ruleId": "C2001",
+          "ruleIndex": 0,
           "message": {
             "arguments": [ "anotherVariable" ],
             "messageId": "default"
           }
         },
         {
-          "ruleId": "C2002-1",
+          "ruleId": "C2002",
```
lgolding: >C2002 [](start = 21, length = 5)

See the comment in the visitor code. It was never right to assign a `ruleKey` to a `ruleId`.

lgolding: In this case, in the input, the `ruleKey` did not point to any entry in the `rules` dictionary, so this `result` has no `ruleIndex` property (it defaults to -1).

---
In reply to: [250300944](https://github.com/Microsoft/sarif-sdk/pull/1232#discussion_r250300944) [](ancestors = 250300944)

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -10,30 +10,32 @@
...
           "message": { "text": "Some testing occurred." }
         },
         {
-          "ruleId": "C2003",
+          "ruleIndex": 2,
```
lgolding: >ruleIndex [](start = 11, length = 9)

In the input file, this result has no `ruleId`, and again it was never right to assign a `ruleKey` to a `ruleId` -- so the v2 `result` has no `ruleId` property. On the other hand, the v1 `ruleKey` _did_ point to a real `rule` object, so this v2 `result` _does_ have a `ruleIndex` property.

src/Sarif/Autogenerated/Notification.cs
```diff
@@ -41,13 +41,13 @@ public SarifNodeKind SarifNodeKind
         public string Id { get; set; }
 
         /// <summary>
-        /// The stable, unique identifier of the rule (if any) to which this notification is relevant. This member can be used to retrieve rule metadata from the rules dictionary, if it exists.
+        /// The stable, unique identifier of the rule, if any, to which this notification is relevant.
```
lgolding: >if any [](start = 55, length = 6)

This was lost in the schema merge.

michaelcfanning: Sure. Would materialize on next BuildAndTest gesture. It's important to perform this step before finalizing a push to get relevant changes in autogen code, sorry for omitting this step.

---
In reply to: [250303030](https://github.com/Microsoft/sarif-sdk/pull/1232#discussion_r250303030) [](ancestors = 250303030)

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
michaelcfanning: ok. side issue: what does v2 say about no rule id and ruleIndex == -1. Seems like an invalid condition. A rule (especially for a converted log file) might not have the SARIF notion of a rule id but in this case, should have a name, so the rule object should exist.

IOW, suggest we say in the spec that either rule id or ruleIndex (with a corresponding rule object) SHOULD be present.

---
In reply to: [250286632](https://github.com/Microsoft/sarif-sdk/pull/1232#discussion_r250286632) [](ancestors = 250286632)

lgolding: We discussed this in person, but for the record: This condition is valid. If a tool does not provide opaque, stable rule ids as required by the spec, `ruleId` must be absent. If the log file does not provide rule metadata (or at least, does not provide metadata for the rule implicated by this result), then `ruleIndex` must be absent/-1.

---
In reply to: [250389698](https://github.com/Microsoft/sarif-sdk/pull/1232#discussion_r250389698) [](ancestors = 250389698,250286632)

PR #1234, opened 2019-01-28
Files array results matching

michaelcfanning:
After this change, we can (and should) merge back to 'develop'. Whew.

@lgolding


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -323,7 +323,7 @@ private int LogRule(IRule iRule)
                 ruleIndex = _ruleToIndexMap.Count;
                 _ruleToIndexMap[rule] = ruleIndex;
                 _run.Resources = _run.Resources ?? new Resources();
-                _run.Resources.Rules = _run.Resources.Rules ?? new List<Rule>();
+                _run.Resources.Rules = _run.Resources.Rules ?? new OrderSensitiveValueComparisonList<Rule>(Rule.ValueComparer);
```
michaelcfanning: >OrderSensitiveValueComparisonList [](start = 67, length = 33)

All this change does is enable Resources.Rules.Contains to have value, not reference, semantics. It's not strictly required for this change. Moving forward, we should figure out how/whether we should utilize this class across the SDK.

src/Sarif.UnitTests/RandomSarifLogGenerator.cs
```diff
@@ -105,7 +106,8 @@ public static IList<Result> GenerateFakeResults(Random random, List<string> rule
                             {
                                 FileLocation = new FileLocation
                                 {
-                                    Uri = filePaths[random.Next(filePaths.Count)]
+                                    Uri = filePaths[fileIndex],
+                                    FileIndex = fileIndex
```
michaelcfanning: >FileIndex = fileIndex [](start = 36, length = 21)

The results matcher now only persists files that are observed via results. I will fix this later, the results matcher should retain all files from current run. Implementation as it stands exists to ensure we flow property bags associated with baseline files, when appropriate. It's easier for me to just fix the gap then to explain much more. :)

[https://github.com/Microsoft/sarif-sdk/issues/1235](https://github.com/Microsoft/sarif-sdk/issues/1235)

PR #1238, opened 2019-01-28
v1 transformer analysis target region persistence fix.

michaelcfanning:
@EasyRhinoMSFT @lgolding


michaelcfanning:
@lgolding for your awareness, please observe the bad hack in master around trimming ".1" off the semantic version (required in order not to alter the actual SARIF schema version utilized in validation testing).

We shouldn't conflate these things, i.e., the SARIF package semantic version may be updated out of sync with the schema version. I don't have time to make a good fix here, we'll clean things up moving forward.

PR #1241, opened 2019-01-29
Baseline states

michaelcfanning:
Add 'updated' state. Rename 'existing' to 'unchanged'.

@lgolding, take a look. I was not able to give this a careful review before heading out, will check in later. Please note that we need some clean-up in the TC30 branch to reflect that this is our ship condition as of the 1/09 TC (e.g., by updating the schema version emitted by the SDK). Sorry for any loose ends, will update later.

### DIFF
--- a/src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
+++ b/src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
@@ -252,7 +252,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
                 Result result = resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior);
 
                 IList<FileData> files = 
-                    (PropertyBagMergeBehavior.HasFlag(DictionaryMergeBehavior.InitializeFromOldest) && result.BaselineState == BaselineState.Existing)
+                    (PropertyBagMergeBehavior.HasFlag(DictionaryMergeBehavior.InitializeFromOldest) &&
+                    (result.BaselineState == BaselineState.Unchanged || result.BaselineState == BaselineState.Updated)) 
                     ? resultPair.PreviousResult.OriginalRun.Files
                     : resultPair.Run.Files;
 
--- a/src/Sarif/Visitors/SarifTransformerUtilities.cs
+++ b/src/Sarif/Visitors/SarifTransformerUtilities.cs
@@ -187,7 +187,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                 case BaselineStateVersionOne.Absent:
                     return BaselineState.Absent;
                 case BaselineStateVersionOne.Existing:
-                    return BaselineState.Existing;
+                    return BaselineState.Unchanged;
                 case BaselineStateVersionOne.New:
                     return BaselineState.New;
                 default:
@@ -200,7 +200,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             {
                 case BaselineState.Absent:
                     return BaselineStateVersionOne.Absent;
-                case BaselineState.Existing:
+                case BaselineState.Unchanged:
+                case BaselineState.Updated:
                     return BaselineStateVersionOne.Existing;
                 case BaselineState.New:
                     return BaselineStateVersionOne.New;
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -47,16 +47,22 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
 
             switch (version)
             {
-                case "2.0.0-csd.2.beta.2018-11-28":
+                case "2.0.0-csd.2.beta.2019-01-24":
                 {
-                    // SARIF TC28. Nothing to do.
+                    // SARIF TC31. Nothing to do.
                     break;
+                }
+
+                case "2.0.0-csd.2.beta.2019-01-09":
+                {
+                        modifiedLog |= ApplyChangesFromTC31(sarifLog);
+                        break;
                 }
 
                 case "2.0.0-csd.2.beta.2018-10-10":
                 {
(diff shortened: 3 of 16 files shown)


src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -200,7 +200,8 @@ public static BaselineStateVersionOne CreateBaselineStateVersionOne(BaselineStat
             {
                 case BaselineState.Absent:
                     return BaselineStateVersionOne.Absent;
-                case BaselineState.Existing:
+                case BaselineState.Unchanged:
+                case BaselineState.Updated:
```
lgolding: So we lose information on the round trip (v2 Updated => v1 Existing => v2 Unchanged), but oh well.

lgolding:
>                     break;

Don't we also want to apply the TC 31 changes to this case, and the default case below?

---
Refers to: src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs:70 in 83e80d6. [](commit_id = 83e80d63801bac6965e01999b742fb8713eed2f4, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -109,7 +115,35 @@ public static class PrereleaseCompatibilityTransformer
...
 
-        private static bool ApplyChangesFromTC25ThroughTC28(
+        private static bool ApplyChangesFromTC31(JObject sarifLog)
+        {
+            bool modifiedLog = UpdateSarifLogVersion(sarifLog);
+
+            var runs = (JArray)sarifLog["runs"];
+
+            if (runs != null)
+            {
```
lgolding: Replace LL121-123 with:
```
if (sarifLog["runs"] is JArray runs)
```

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -109,7 +115,35 @@ public static class PrereleaseCompatibilityTransformer
...
+                foreach (JObject run in runs)
+                {
+                    if (run["results"] is JArray results)
```
lgolding: >results [](start = 49, length = 7)

Ah! You did use that construct here.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -109,7 +115,35 @@ public static class PrereleaseCompatibilityTransformer
...
+                    {
+                        foreach (JObject result in results)
+                        {
+                            string baselineState = (string)result["baselineState"];
+
+                            if ("existing".Equals(baselineState))
```
lgolding: I suggest `StringComparison.Ordinal`

lgolding:
@michaelcfanning One possible bug, a couple of nits, otherwise LGTM.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -252,7 +252,8 @@ private Rule GetRuleFromResources(Result result, IDictionary<string, Rule> rules
                 Result result = resultPair.CalculateBasedlinedResult(PropertyBagMergeBehavior);
 
                 IList<FileData> files = 
-                    (PropertyBagMergeBehavior.HasFlag(DictionaryMergeBehavior.InitializeFromOldest) && result.BaselineState == BaselineState.Existing)
+                    (PropertyBagMergeBehavior.HasFlag(DictionaryMergeBehavior.InitializeFromOldest) &&
+                    (result.BaselineState == BaselineState.Unchanged || result.BaselineState == BaselineState.Updated))
```
lgolding: >Updated [](start = 110, length = 7)

Need a unit test that shows updated files are included.

michaelcfanning: Not yet! No matcher actually sets that enum value (to indicate, yes, found a match but details were update). That's a good improvement for the heuristics manager to set, potentially. All this area needs a significant update, however, which Scott will do in the future. So deferring for now.

---
In reply to: [252002401](https://github.com/Microsoft/sarif-sdk/pull/1241#discussion_r252002401) [](ancestors = 252002401)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Ordinal is what's used for String.Equals. This is an important fact about .NET which is my argument for preferring not to make it explicit.

---
In reply to: [252001522](https://github.com/Microsoft/sarif-sdk/pull/1241#discussion_r252001522) [](ancestors = 252001522)

michaelcfanning:
>                     break;

yes, of course, good catch.

---
In reply to: [458689467](https://github.com/Microsoft/sarif-sdk/pull/1241#issuecomment-458689467) [](ancestors = 458689467)

---
Refers to: src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs:70 in 83e80d6. [](commit_id = 83e80d63801bac6965e01999b742fb8713eed2f4, deletion_comment = False)

src/Sarif/Visitors/SarifTransformerUtilities.cs
michaelcfanning: That's true. We could persist something in a property bag, of course. But we don't foresee need for detailed round-tripping of this level of precision currently. Can revisit if need be (don't think either of us would put odds on it).

---
In reply to: [251997378](https://github.com/Microsoft/sarif-sdk/pull/1241#discussion_r251997378) [](ancestors = 251997378)

michaelcfanning:
Thanks, Larry, appreciate the look.

---
In reply to: [458692710](https://github.com/Microsoft/sarif-sdk/pull/1241#issuecomment-458692710) [](ancestors = 458692710)

michaelcfanning:
@lgolding FYI, will take your notes in the next change. Is there any chance you could cherry pick the updates to the release notes and package build into this branch?

Should we just go ahead and ship 1/09 in your opinion?

PR #1249, opened 2019-01-31
Result level

michaelcfanning:
@lgolding, this change isn't where it needs to be, some tests are failing for example. Please take a look if you can, though, as it's a large, intrusive change. 

A core thing to keep in mind: any result with a kind of Fail should have a FailureLevel or note, warning or error. Note the use of the term 'FailureLevel' to make this association explicit, let me know what you think. FailureLevel of None means there is no 'failedness' associated with the message

With this scheme, you can now emit a pure trace message in a tool by using both a kind of none and a failure level of none. Should we add these values to the schema therefore?

Please see new hijinks in the result object to keep the level/kind data in sync via their respective property setters.


src/Sarif/ConsoleLogger.cs
```diff
@@ -94,40 +94,47 @@ public void Log(IRule rule, Result result)
...
                 result.RuleId,
                 message);
         }
 
-        private void WriteToConsole(ResultLevel level, Uri uri, Region region, string ruleId, string message)
+        private void WriteToConsole(FailureLevel level, ResultKind kind, Uri uri, Region region, string ruleId, string message)
         {
+            // If we are not in an actual failure case, the message level
+            // is not relevant so we'll explicitly set it to None.
+            if (kind != ResultKind.Fail)
```
michaelcfanning: >kind [](start = 16, length = 4)

We could insist that callers maintain this invariant instead of correcting the data ourselves. ?

src/Sarif/Autogenerated/Result.cs
```diff
@@ -48,15 +48,51 @@ public SarifNodeKind SarifNodeKind
...
+        [DataMember(Name = "kind", IsRequired = false, EmitDefaultValue = false)]
+        [DefaultValue(ResultKind.Fail)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
+        public ResultKind Kind
+        {
+            get { return _kind; }
+            set
+            {
+                _kind = value;
```
michaelcfanning: >_kind [](start = 16, length = 5)

of particular interest

src/Sarif.Converters.UnitTests/PylintConverterTests.cs
```diff
@@ -56,7 +56,8 @@ private static Result CreateResult()
             {
                 RuleId = "C0412(testSymbol)",
                 Message = new Message { Text = "testMessage" },
-                Level = ResultLevel.Warning,
+                Level = FailureLevel.Warning,
+                Kind = ResultKind.Fail,
```
michaelcfanning: these pairs shouldn't literally be required after fixing up the property setters in Result.cs. i.e., simply by setting level to warning, kind will be updated properly. I'll fix these unnecessary inits later (it was my unease about requiring all these pairs to be kept accurate that led me to make the property changes)

src/Sarif/CodeGenHints.json
```diff
@@ -329,22 +329,13 @@
...
       "kind": "EnumHint",
       "arguments": {
-        "typeName": "NotificationLevel",
-        "description": "Values specifying the level of a notification.",
+        "typeName": "FailureLevel",
+        "description": "Values specifying the level of a notification or result.",
+        "zeroValueName": "None",
         "memberNames": [
           "Warning",
           "Note",
```
lgolding: >Note [](start = 11, length = 4)

Since this is now clearly a severity, I suggest we order the enum values by severity: `{ None = 0, Note, Warning, Error }`. I will do that in my branch.

lgolding: While I'm here, since again I think this now maps nicely to what people think of as "severity", did you consider calling the type `SeverityLevel` or even just `Severity`? Did you reject the idea because "`FailureLevel.None` ("no failure") reads better to you than "`Severity.None`" ("no severity"???)?

---
In reply to: [252888798](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252888798) [](ancestors = 252888798)

src/Sarif/CodeGenHints.json
```diff
@@ -506,6 +496,29 @@
       }
     }
   ],
+  "Result.Level": [
+    {
+      "kind": "EnumHint",
+      "arguments": {
+        "typeName": "FailureLevel",
```
lgolding: >FailureLevel [](start = 21, length = 12)

Wow. Is this the first time we've had two properties need the same generated enum type, and had the problem that now we have to keep them in sync? Oh well. Ideally the code hint processor would provide a way for an `EnumHint` to indirectly specify a common enum type, but today is not that day.

src/Sarif/CodeGenHints.json
```diff
@@ -480,19 +471,18 @@
...
         "memberNames": [
           "NotApplicable",
           "Pass",
-          "Note",
-          "Warning",
-          "Error",
+          "Fail",
+          "Review",
           "Open"
         ]
```
lgolding: >] [](start = 8, length = 1)

I do like this factoring out of the concepts of "kind" and "severity/failure level".

src/Sarif/CodeGenHints.json
lgolding: Oh wow, `RuleConfiguration.DefaultLevel` below uses the same type -- it now appears three times in this file. Good luck to us.

---
In reply to: [252889661](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252889661) [](ancestors = 252889661)

src/Sarif/ConsoleLogger.cs
```diff
@@ -94,40 +94,47 @@ public void Log(IRule rule, Result result)
 
             WriteToConsole(
                 result.Level,
+                result.Kind,
```
lgolding: >Kind [](start = 23, length = 4)

I suggest putting `kind` before `Level` in the argument list; it's the more general concept.

src/Sarif/ConsoleLogger.cs
lgolding: Yes, I think so. If `kind` is not `fail`, but `level` is (say) `error`, then how do we know what's right? Did they mean to set `kind` to `fail`? or did they mean to set `level` to `none`? I'd throw `ArgumentException` if they're inconsistent.

---
In reply to: [252549059](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252549059) [](ancestors = 252549059)

src/Sarif/ConsoleLogger.cs
```diff
@@ -136,10 +143,16 @@ private void WriteToConsole(ResultLevel level, Uri uri, Region region, string ru
             Region region,
             string ruleId,
             string message,
-            ResultLevel resultLevel)
+            FailureLevel resultLevel,
```
lgolding: >resultLevel [](start = 25, length = 11)

I would either consistently name these variables `level` and `kind` (as you did on L104), or consistently name them for their types (`failureLevel` and `resultKind`).

lgolding: I prefer the latter. We have too many "kinds" (and possibly "levels", I don't remember) floating around.

---
In reply to: [252893421](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252893421) [](ancestors = 252893421)

src/Sarif/IResultLogWriter.cs
```diff
@@ -111,19 +111,11 @@ public interface IResultLogWriter
         void WriteResults(IEnumerable<Result> results);
 
         /// <summary>
-        /// Write a set of notifications relevant to the operation of the tool to the log.
+        /// Write a set of invocations relevant to the operation of the tool to the log.
```
lgolding: >invocations relevant to the operation of the tool [](start = 27, length = 49)

IMO, just "a set of tool invocations"

src/Sarif/Autogenerated/Result.cs
```diff
@@ -48,15 +48,51 @@ public SarifNodeKind SarifNodeKind
...
+                if (_kind != ResultKind.None && _kind != ResultKind.Fail)
```
lgolding: >_kind != ResultKind.None [](start = 20, length = 24)

I don't agree with the first part of the condition. If somebody writes `myResult.Kind = ResultKind.None` -- if they don't even know the _kind_ of the result -- it's fair to say they don't know the failure level, or even if it _is_ a failure. I'd just leave the second half of the condition.

Having said that, I'm uncomfortable with this side effect; I can't decide whether if violates the principle of least surprise.

src/Sarif/Autogenerated/Result.cs
```diff
@@ -48,15 +48,51 @@ public SarifNodeKind SarifNodeKind
...
+                if (_level!= FailureLevel.None)
+                {
+                    _kind = ResultKind.Fail;
+                }
+            }
+        }
 
         /// <summary>
-        /// A message that describes the result. The first sentence of the message only will be displayed when visible space is limited.
+        /// A message tha_lket describes the result. The first sentence of the message only will be displayed when visible space is limited.
```
lgolding: >tha_lket [](start = 22, length = 8)

oops

src/Sarif.Multitool.FunctionalTests/ValidateCommandTests.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
     {
-        [Fact(DisplayName = nameof(ValidateCommand_ReportsJsonSyntaxError))]
+        [Fact(DisplayName = nameof(ValidateCommand_ReportsJsonSyntaxError), Skip = "JSchema requires an update.")]
         public void ValidateCommand_ReportsJsonSyntaxError()
         {
             Verify("SyntaxError.sarif", disablePreleaseCompatibilityTransform: true);
         }
 
-        [Fact(DisplayName = nameof(ValidateCommand_ReportsDeserializationError))]
+        [Fact(DisplayName = nameof(ValidateCommand_ReportsDeserializationError), Skip = "Jschema requires an update.")]
```
lgolding: >Jschema [](start = 89, length = 7)

If I read the rest of the PR will I learn what needs to happen to JSchema? If not please respond here.

src/Sarif.TestUtilities/FileDiffingTests.cs
```diff
@@ -201,13 +201,15 @@ public static bool AreEquivalent<T>(string actualSarif, string expectedSarif, IC
...
 
-            expectedSarif = expectedSarif ?? "{}";
-
             JsonSerializerSettings settings = new JsonSerializerSettings()
             {
                 ContractResolver = contractResolver,
+                Formatting = Formatting.Indented
             };
 
+            // Round-trip expected content to make it realie various default values
```
lgolding: >realie [](start = 54, length = 6)

realize? reify?

lgolding:
Mostly done. Will finish later this evening.

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -60,95 +60,130 @@ public static string CreateDisambiguatedName(string baseName, int index)
...
-                    return ResultLevel.Warning;
+                    return FailureLevel.Warning;
                 case ResultLevelVersionOne.NotApplicable:
-                    return ResultLevel.NotApplicable;
+                    return FailureLevel.None;
                 default:
-                    return ResultLevel.Default;
+                    return FailureLevel.Warning;
+            }
+        }
```
lgolding: >} [](start = 8, length = 1)

Add blank line.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -286,15 +287,14 @@ private static bool ExtractCallerAndCallee(string text, out string caller, out s
...
         {
             switch (sdvLevel)
             {
-                case "Error": return ResultLevel.Error;
+                case "Error": return FailureLevel.Error;
             }
 
-            Debug.Assert(false);
-            return ResultLevel.Default;
+            throw new InvalidOperationException($"Unknown SDV level encountered: {sdvLevel}");
```
lgolding: >InvalidOperationException [](start = 22, length = 25)

Better!

src/Sarif.Converters/TSLintConverter.cs
```diff
@@ -41,21 +41,22 @@ internal Result CreateResult(TSLintLogEntry entry)
...
                     break;
                 case "WARN":
                 case "WARNING":
-                    result.Level = ResultLevel.Warning;
+                    result.Level = FailureLevel.Warning;
                     break;
                 case "DEFAULT":
                 default:
-                    result.Level = ResultLevel.Note;
+                    result.Level = FailureLevel.Note;
```
lgolding: >Note [](start = 48, length = 4)

But if we set `Level` to `Note`, then we should reset `Kind` and not leave it as `Fail`

src/Sarif.Converters.UnitTests/PylintConverterTests.cs
lgolding: Yeah, I did comment on that. Not sure what's best, I leave it to you.

---
In reply to: [252549912](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252549912) [](ancestors = 252549912)

lgolding:
Looks good. A few micro-nits, and one BUG with an obvious fix. Other than that, looks solid. TTY you on Tuesday!

src/Sarif.Multitool.FunctionalTests/ValidateCommandTests.cs
michaelcfanning: Sorry to have neglected this. Jschema will no longer load the validator because it can't find the ResultConfigurationDefaultLevel object. It seems clear that JSchema is picking up some newly compiled dependency rather than its own notion of the SARIF SDK core. To be chased down later...

---
In reply to: [252899117](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252899117) [](ancestors = 252899117)

src/Sarif/CodeGenHints.json
michaelcfanning: I took the ordering change. I left it FailureLevel for now, strictly to help emphasize the direct relationship with ResultKind.Fail (which isn't ResultKind.HasSeverity)

---
In reply to: [252889280](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252889280) [](ancestors = 252889280,252888798)

src/Sarif/VersionConstants.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class VersionConstants
     {
-        public const string Prerelease = "csd.2.beta.2019-01-24";
+        public const string Prerelease = "csd.2.beta.2018-11-28";
```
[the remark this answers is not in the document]
michaelcfanning: bad merge! thanks

---
In reply to: [252895428](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252895428) [](ancestors = 252895428)

src/Sarif/Autogenerated/Result.cs
michaelcfanning: No, you're right. If ResultKind is not Fail, None is the only valid value for FailureLevel. Good catch.

---
In reply to: [252898602](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252898602) [](ancestors = 252898602)

src/Sarif.Converters/TSLintConverter.cs
michaelcfanning: No sir. 'Note' is actually a fail. :)

---
In reply to: [252922438](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252922438) [](ancestors = 252922438)

src/Sarif/ConsoleLogger.cs
michaelcfanning: agree

---
In reply to: [252892432](https://github.com/Microsoft/sarif-sdk/pull/1249#discussion_r252892432) [](ancestors = 252892432,252549059)

Issue #1253, opened 2019-01-31
FortifyFpr: empty result.locations causes exception in AddSnippetsToResults


EasyRhinoMSFT:
We should wait to create the locations list until we know we have one. See ParseLocationsFromTraces:337. Should move down to line 480.

Issue #1252, opened 2019-01-31
FortifyFpr: exception when Node.SourceLocation doesn't have a snippet


EasyRhinoMSFT:
In AddNodeLocationsToThreadFlowLocations, we should make sure we got a snippetId out of _locationToSnippetIdDictionary before we try to use it to get the regions.

PR #1254, opened 2019-02-01
Fix #1251 #1252 #1253

### LINKED ISSUE #1251: FortifyFpr: SourceBasePath from run on Linux causes exception
SourceBasePath will look like "/root/blah/blah" which causes a format exception to be thrown by the Uri constructor. We should add file: to the path string if it's from Linux.

EasyRhinoMSFT:
Fixes for issues uncovered by Katrina's Fortify FPR converter usage.


src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -124,11 +124,19 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
...
             if (!string.IsNullOrWhiteSpace(_originalUriBasePath))
             {
-                var uri = new Uri(_originalUriBasePath);
-                run.OriginalUriBaseIds = new Dictionary<string, FileLocation>
+                if (_originalUriBasePath.StartsWith("/") &&
+                    _invocation.GetProperty("Platform") == "Linux")
                 {
-                    { FileLocationUriBaseId, new FileLocation { Uri = uri } }
-                };
+                    _originalUriBasePath = "file:/" + _originalUriBasePath;
```
lgolding: >/ [](start = 49, length = 1)

Reading RFC 3986, I think this should be just `file:`, not `file:/`.

src/Sarif.Converters.UnitTests/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
```diff
@@ -11,7 +11,7 @@
       },
       "originalUriBaseIds": {
         "SRCROOT": {
-          "uri": "file:///C:/projects/myproject/src/"
+          "uri": "file://root/projects/myproject/src/"
```
lgolding: >/ [](start = 23, length = 1)

Again, doesn't `file:/root...` work? If not, don't bother to change it.

lgolding:
Before you provide a private package to Katrina, you need to change the version number to `...2019-01-09` in four places:

- build.props
- VersionConstants.cs
- sarif-schema.json
    - title
    - description

lgolding:
Include that as part of this PR.

---
In reply to: [459772298](https://github.com/Microsoft/sarif-sdk/pull/1254#issuecomment-459772298) [](ancestors = 459772298)

src/Sarif.Converters/FortifyFprConverter.cs
EasyRhinoMSFT: Tried that, Uri ctor throws.

---
In reply to: [253101090](https://github.com/Microsoft/sarif-sdk/pull/1254#discussion_r253101090) [](ancestors = 253101090)

src/Sarif.Converters.UnitTests/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
EasyRhinoMSFT: Meaning keep it as a relative URI?

---
In reply to: [253102144](https://github.com/Microsoft/sarif-sdk/pull/1254#discussion_r253102144) [](ancestors = 253102144)

EasyRhinoMSFT: I'm going to check in with this change, we can revert later after we discuss.

---
In reply to: [253125417](https://github.com/Microsoft/sarif-sdk/pull/1254#discussion_r253125417) [](ancestors = 253125417,253102144)

PR #1258, opened 2019-02-04
Generalize message metadata. Break apart rule configuration from rule…

michaelcfanning:
… description. Move relevant data from run object to invocations. @lgolding @EasyRhinoMSFT 

This change shows the schema update only, as an early sanity check.


michaelcfanning:
run.resources object removed and decomposed into tool.globalMessageStrings, tool.notificationMetadata and tool.rulesMetadata

Metadata for notification and rules metadata generalized into a common 'messageDescriptor' object. These descriptor have a 'defaultConfiguration' property that provides default state for checks,

Invocation.rulesConfiguration provides an array of ruleConfiguration objects describing a specific tool invocation.

michaelcfanning:
closing. this branch doesn't look quite right, will recreate

PR #1259, opened 2019-02-04
Tool message metadata changes.

michaelcfanning:
run.resources object removed and decomposed into tool.globalMessageStrings, tool.notificationMetadata and tool.rulesMetadata

Metadata for notification and rules metadata generalized into a common 'messageDescriptor' object. These descriptor have a 'defaultConfiguration' property that provides default state for checks,

Invocation.rulesConfiguration provides an array of ruleConfiguration objects describing a specific tool invocation.

@lgolding @EasyRhinoMSFT


src/Sarif/IResultLogWriter.cs
```diff
@@ -41,15 +41,15 @@ public interface IResultLogWriter
...
+        /// observed during analysis, might not be known until all results have been
+        /// generated. A Sarif file may also contain only message metadata.
         /// </summary>
-        /// <param name="rules">
-        /// An array whose elements comprise rules relevant to the analysis.
+        /// <param name="tool">
+        /// The tool object relevant to the current run.
         /// </param>
-        void WriteRules(IList<Rule> rules);
+        void WriteTool(Tool tool);
```
michaelcfanning: >WriteTool [](start = 13, length = 9)

In general, the SARIF design has changed sufficiently that the approach of this logger isn't ideal (it wasn't a great design to begin with). Moving forward, I think this logger could usefully work as follow: initialize the logger with a run object, user then writes results. logger accumulates file objects, etc., to the run object associated with the log. On Dispose, the log would write all non-null run objects, for logical locations, etc. Consumer could still update these run objects on the fly if useful. By using this approach, the logger interface could be greatly reduced (i.e., no need to maintain helpers to write all the various tables).

src/Sarif/Core/IRule.cs
```diff
@@ -1,9 +0,0 @@
-﻿// Copyright (c) Microsoft.  All Rights Reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace Microsoft.CodeAnalysis.Sarif
-{
-    public partial interface IRule : IPropertyBagHolder
```
michaelcfanning: >IRule [](start = 29, length = 5)

It's gone. Whew. :)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
```diff
@@ -22,7 +41,31 @@ public void PrereleaseCompatibilityTransformer_UpgradesPrereleaseTwoZeroZero()
             PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(sarifText, forceUpdate: false, formatting: Formatting.None, out sarifText);
 
             SarifLog sarifLog = JsonConvert.DeserializeObject<SarifLog>(sarifText);
-            JsonConvert.SerializeObject(sarifLog);            
+            JsonConvert.SerializeObject(sarifLog);
+        }
+
+        [Fact]
+        public void PrereleaseCompatibilityTransformer_NestedFiles()
```
michaelcfanning: >PrereleaseCompatibilityTransformer_NestedFiles [](start = 20, length = 46)

these tests moved from a deleted file. I put a separate prerelease transformer tests file in the wrong location (visitor dir rather than writers)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
```diff
@@ -22,7 +41,31 @@ public void PrereleaseCompatibilityTransformer_UpgradesPrereleaseTwoZeroZero()
...
+        }
+
+        [Fact]
+        public void PrereleaseCompatibilityTransformer_RuleIdCollisions()
+        {
+            RunTest("RuleIdCollisions.sarif");
+        }
+
+        [Fact]
+        public void PrereleaseCompatibilityTransformer_RunResources()
```
michaelcfanning: this is actually the new relevant test

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/OneRunWithRules.sarif
```diff
@@ -70,7 +70,8 @@
               "text": "Rent internal rebellion competence biography photograph."
             },
             "configuration": {
-              "defaultLevel": "note",
+              "enabled" :  false,
```
michaelcfanning: >"enabled" :  false, [](start = 14, length = 19)

bug in transformer logic didn't always flow enabled == false state properly, now fixed

src/Sarif.UnitTests/RandomSarifLogGenerator.cs
```diff
@@ -47,18 +47,21 @@ public static SarifLog GenerateSarifLogWithRuns(Random randomGen, int runCount)
...
-
-            run.Tool = new Tool() { Name = "Test", Version = "1.0", };
-            run.Resources.Rules = GenerateRules(ruleIds);
-            run.Files = GenerateFiles(filePaths);
-
             int results = resultCount == null ? random.Next(100) : (int)resultCount;
-            run.Results = GenerateFakeResults(random, ruleIds, filePaths, results);
 
-            return run;
+            return new Run()
```
michaelcfanning: >Run [](start = 23, length = 3)

just cleaned up object creation to use the standard syntax we prefer these days.

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif
```diff
@@ -5,7 +5,13 @@
     {
       "tool": {
         "name": "CodeScanner",
-        "downloadUri": "ht%tp://www.example.com/tools/codescanner/download.html"
+        "downloadUri": "ht%tp://www.example.com/tools/codescanner/download.html",
+        "rulesMetadata": [
```
michaelcfanning: >rulesMetadata [](start = 9, length = 13)

as you recall, certain invalid test files _must_ be kept in conformance with current SARIF format because they can't be put through the prerelease transformer without having their essential flaws (upon which testing depends) cleaned up. :)

src/Sarif.UnitTests/Readers/IsSuppressedInSourceConverterTests.cs
```diff
@@ -46,6 +46,8 @@ public void SuppressionStatus_SuppressedInSource()
                         SuppressionStates = SuppressionStates.SuppressedInSource
                     }
                 });
+
+                uut.CloseResults();
```
michaelcfanning: >CloseResults [](start = 20, length = 12)

this call not strictly required

src/Sarif.UnitTests/Readers/IsSuppressedInSourceConverterTests.cs
```diff
@@ -87,6 +89,9 @@ public void BaselineState_None()
                         BaselineState = BaselineState.None
                     }
                 });
+
+                uut.CloseResults();
```
michaelcfanning: >CloseResults [](start = 20, length = 12)

note this alternate pattern, currently supported by code, to explicitly close results and write a tool object, rather than allowing the log writer Dispose method to handle it.

src/Sarif.Multitool.FunctionalTests/ValidateCommandTests.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
     public class ValidateCommandTests : SarifMultitoolTestBase
     {
-        [Fact(DisplayName = nameof(ValidateCommand_ReportsJsonSyntaxError), Skip = "JSchema requires an update.")]
+        [Fact(Skip = "JSchema attempts to instantiate types from the compiled Sarif assembly that no longer exist")]
```
michaelcfanning: >JSchema [](start = 22, length = 7)

Here's what's happening: when the validation code loads during testing, our tested/in-flight SARIF.DLL is already loaded in the default app domain. JSchema therefore attempts to load Rule, IRule, RuleConfiguration and Result objects from it. it is very interesting that JSchema has been effectively upgrade to current SARIF dll without breaking previously, I was originally able to work around this problem by reproducing the rule factory code from JSchema (that builds a rule object from the jschema error identifiers). At some point, however, the divergence of the result objects broke as well. Easiest path forward: update Jschema then reenable all these tests.

michaelcfanning:
@lgolding ready for your attention and thanks for the review

src/Sarif/CodeGenHints.json
```diff
@@ -412,11 +442,6 @@
       }
     }
   ],
-  "Resources.MessageStrings": [
```
lgolding: >MessageStrings [](start = 13, length = 14)

So we've confirmed that there really never was a use for these "global message strings"? Or rather, I suppose, they were used _only_ in notifications?

lgolding: Oh, I see, this became `Tool.GlobalMessageStrings`.

---
In reply to: [255306444](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255306444) [](ancestors = 255306444)

src/Sarif/Core/Tool.cs
```diff
@@ -49,5 +50,16 @@ public static Tool CreateFromAssemblyData(string prereleaseInfo = null)
 
             return tool;
         }
+
+        public bool ShouldSerializeRulesMetadata()
+        {
+            return this.RulesMetadata != null &&
+                    this.RulesMetadata.Where((n) => { return n != null; }).Any();
+        }
```
lgolding: >} [](start = 8, length = 1)

Add blank line below.

src/Sarif/Core/Tool.cs
```diff
@@ -49,5 +50,16 @@ public static Tool CreateFromAssemblyData(string prereleaseInfo = null)
 
             return tool;
         }
+
+        public bool ShouldSerializeRulesMetadata()
+        {
+            return this.RulesMetadata != null &&
+                    this.RulesMetadata.Where((n) => { return n != null; }).Any();
```
lgolding: >Where((n) => { return n != null; }).Any() [](start = 39, length = 41)

The second line can simplify to `this.RulesMetadata.Any(n => n != null)`. Taking it a step further, the whole method can collapse to
```
    return this.RulesMetadata?.Any(n => n != null) == true;
```

src/Sarif/Core/Tool.cs
```diff
@@ -49,5 +50,16 @@ public static Tool CreateFromAssemblyData(string prereleaseInfo = null)
...
+        }
+        public bool ShouldSerializeNotificationsMetadata()
+        {
+            return this.NotificationsMetadata != null &&
+                    this.NotificationsMetadata.Where((n) => { return n != null; }).Any();
+        }
```
lgolding: >} [](start = 8, length = 1)

You can DRY these two methods out:
```
public bool ShouldSerializeRulesMetadata() => ShouldSerializeMessageDescriptors(this.RulesMetadata);
public bool ShouldSerializeNotificationsMetadata() => ShouldSerializeMessageDescriptors(this.NotificationsMetadata);

static private bool ShouldSerializeMessageDescriptors(IEnumerable<MessageDescriptor> messageDescriptors)
    =>messageDescriptors?.Any(n => n != null) == true;
```

lgolding:
>   "title": "Static Analysis Results Format (SARIF) Version 2.0.0-csd.2.beta.2018-11-28 JSON Schema",

2019-01-24 on this line and the next.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:3 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -715,6 +710,16 @@
           "type": "integer"
         },
 
+        "ruleConfiguration": {
```
lgolding: >ruleConfiguration [](start = 9, length = 17)

Ah, yes, I remember discussing how this is really a property of the invocation. Nice.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2084,6 +2053,36 @@
...
+          "minItems": 0,
+          "uniqueItems": true,
+          "default": [],
+          "items": {
+            "$ref": "#/definitions/messageDescriptor"
+          }
+        },
+
+        "rulesMetadata": {
+          "description": "An array of message descriptor objects relevant to the analysis provided by the tool.",
```
lgolding: >provided [](start = 90, length = 8)

provided => performed ?

lgolding: >provided [](start = 90, length = 8)

provided => performed?

src/Sarif/NotYetAutoGenerated/SarifRewritingVisitor.cs
```diff
@@ -720,11 +709,8 @@ public virtual Run VisitRun(Run node)
                         node.VersionControlProvenance[index_0] = VisitNullChecked(node.VersionControlProvenance[index_0]);
                     }
                 }
-
-                // OriginalUriBaseIds are directories, not files. We'll disable this visit until the
-                // schema can catch up with this reality.
```
lgolding: Don't remove this comment. It's important to know why files are in the `NotYetAutogenerated` directory.

lgolding:
Each class in the `NotYetAutogenerated` directory needs a summary comment that explains how it differs from what the code generator would have produced. Otherwise we'll never know/remember when we can get rid of them.

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
```diff
@@ -26,6 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Sarif.Multitool\Sarif.Multitool.csproj" />
```
lgolding: >Multitool [](start = 56, length = 9)

Why?

src/Sarif.Multitool.FunctionalTests/TestData/UrisMustBeValid/Invalid.sarif
lgolding: Got it.

---
In reply to: [255249407](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255249407) [](ancestors = 255249407)

lgolding:
>         public Run Run { get; set;  }

This property can be get-only.

---
Refers to: src/Sarif.TestUtilities/ResultLogObjectWriter.cs:20 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

lgolding:
>         private Tool _tool;

This field can be `readonly`.

---
Refers to: src/Sarif.TestUtilities/ResultLogObjectWriter.cs:13 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

lgolding:
>         public Run Run { get; set;  }

(I wonder why `Run` is a property but `_tool` is a field. They should probably both just be `readonly` fields.)

---
In reply to: [462066560](https://github.com/Microsoft/sarif-sdk/pull/1259#issuecomment-462066560) [](ancestors = 462066560)

---
Refers to: src/Sarif.TestUtilities/ResultLogObjectWriter.cs:20 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -303,11 +303,6 @@
           "$ref": "#/definitions/externalPropertyFile"
         },
 
-        "resources": {
```
lgolding: >resources [](start = 9, length = 9)

It's right to remove this -- but now I think you need to allow "tool" to be externalized. Either that, or externalize `tool.rulesMetadata`, `tools.notificationsMetadata`, and `tools.globalMessageStrings` separately -- then we would have to bite the bullet and specify the JPath for each of them (up to now, every externalized property has come directly from the `run` object).

src/Sarif/VersionOne/Readers/SarifContractResolverVersionOne.cs
```diff
@@ -53,9 +53,6 @@ protected override JsonContract CreateContract(Type objectType)
             else if (objectType == typeof(SuppressionStatesVersionOne))
                 contract.Converter = FlagsEnumConverter.Instance;
 
-            else if (objectType == typeof(Dictionary<string, IRuleVersionOne>))
```
lgolding: >IRuleVersionOne [](start = 61, length = 15)

This can't be right -- we didn't (or shouldn't have) removed `IRule` from version 1.

src/Sarif.Multitool.FunctionalTests/TestData/DoNotUseFriendlyNameAsRuleId/Invalid.sarif
```diff
@@ -26,8 +26,8 @@
       "properties": {
         "expectedResults": {
           "resultLocationPointers": [
-            "runs[0].resources.rules[0]",
-            "runs[0].resources.rules[1]"
+            "runs[0].tool.rulesMetadata[0]",
```
lgolding: >rulesMetadata [](start = 26, length = 13)

I get it, I get it. Once this file is pre-release-compatibility-transformed, that's where the `resources.rules` will end up. But boy is this confusing to read. I can't wait for us to update all the test files.

src/Sarif.Multitool.FunctionalTests/Rules/UriMustBeAbsoluteTests.cs
```diff
@@ -12,6 +12,9 @@ public void UriMustBeAbsolute_ReportsInvalidSarif()
         {
             // We need to disable compatibility transformations for any files that require
             // a malformed schema or malformed JSON, as this code fixes those things up
```
lgolding: >up [](start = 85, length = 2)

While you're in here, add the period at the end of this sentence.

src/Sarif.Multitool.FunctionalTests/Rules/UrisMustBeValidTests.cs
```diff
@@ -19,13 +19,20 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
...
-        public void UrisMustBeValid_ReportsInvalidSarif() =>
+        public void UrisMustBeValid_ReportsInvalidSarif()
+        {
             // We need to disable compatibility transformations for any files that require
             // a malformed schema or malformed JSON, as this code fixes those things up
+            //
+            // As a result, when this test fails, you must open the test file below
+            // and manually bring it into conformance with the current v2 spec.
             Verify("Invalid.sarif", disablePrereleaseCompatibilityTransform: true);
+        }
```
lgolding: >} [](start = 8, length = 1)

*sigh* Your code base, I'm just the hired help ;-)

src/Sarif.Multitool.FunctionalTests/ValidateCommandTests.cs
lgolding: You know, JSchema _hasn't_ always upgraded successfully from one version to the next. That's why we had that long discussion about refactoring the packages and building them in a particular order -- first JSchema + JSchemaToDotNet, then SARIF SDK, then JSchema validator, then the rest of the SARIF packages. And when I got around to doing that, the problem we'd observed -- whatever it was, who knows 18 months later -- was gone and I never bothered.

Yes, we can work around the problem again by disabling this tests, upgrading JSchema, and re-enabling the tests.

---
In reply to: [255250496](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255250496) [](ancestors = 255250496)

src/Sarif.UnitTests/Writers/PrereleaseCompatibilityTransformerTests.cs
```diff
@@ -3,15 +3,34 @@
...
 using System.IO;
-
+using Microsoft.CodeAnalysis.Sarif.TestUtilities;
 using Microsoft.CodeAnalysis.Sarif.Writers;
 using Newtonsoft.Json;
 using Xunit;
+using Xunit.Abstractions;
 
-namespace Microsoft.CodeAnalysis.Sarif.UnitTests.Writers
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
```
lgolding: >Visitors [](start = 39, length = 8)

The product code for `PrereleaseCompatibilityTransformer` is in `Sarif.Writers` namespace.

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/Inputs/RunResources.sarif
```diff
@@ -0,0 +1,47 @@
...
+        },
+        {
+          "ruleId": "SHARED-ID.2",
+          "message": { "messageId": "a" },
+          "locations": [ { "physicalLocation": { "fileLocation": { "uri": "file.png" } } } ]
+        }
+      ]
+    }
+  ]
+}
```
lgolding: I love how well thought out this example is -- two variants of a rule about "sensitive content", the graphics variant being disabled by default but an error if present, the "empty file" case using a generic message -- really nice.

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/Inputs/RunResources.sarif
```diff
@@ -0,0 +1,47 @@
...
+        "semanticVersion": "2.1.0"
+      },
+      "resources": {
+        "messageStrings": {"noContent": "This file has no content to review."},
+        "rules": {
+          "SHARED-ID.2": {
+            "id": "SHARED-ID",
+            "configuration": {
+              "defaultLevel": "error",
+              "enabled": false
```
lgolding: >false [](start = 25, length = 5)

This rule is disabled by default, yet there is a `result` that refers to it. Just for the sake of making this beautiful example self-consistent, there should be an `invocation` object whose `ruleConfiguration` enables this rule.

src/Sarif.UnitTests/Readers/IsSuppressedInSourceConverterTests.cs
```diff
@@ -78,7 +78,7 @@ public void BaselineState_None()
 }";
             string actual = GetJson(uut =>
             {
-                var run = new Run() { Tool = DefaultTool };
+                var run = new Run() {  };
```
lgolding: > {  } [](start = 35, length = 5)

Unnecessary braces.

src/Sarif.UnitTests/Readers/IsSuppressedInSourceConverterTests.cs
lgolding: That's fine. Just curious why you changed the test to use that pattern. If the reason was to get coverage on this code path, then your PR comment should become an in-source comment explaining that.

---
In reply to: [255249783](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255249783) [](ancestors = 255249783)

src/Sarif.UnitTests/RandomSarifLogGenerator.cs
lgolding: Lovely!

---
In reply to: [255249109](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255249109) [](ancestors = 255249109)

src/Sarif.Multitool.FunctionalTests/ValidateCommandTests.cs
lgolding: Really good explanation, BTW.

---
In reply to: [255250496](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255250496) [](ancestors = 255250496)

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -7,18 +7,18 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
-    public abstract class SkimmerBase<TContext>  : Rule, ISkimmer<TContext>
+    public abstract class Skimmer<TContext>  : MessageDescriptor
```
lgolding: >Skimmer [](start = 26, length = 7)

Since you renamed the class, rename the file as well.

src/Sarif.Driver/Sdk/SkimmerBase.cs
```diff
@@ -7,18 +7,18 @@
...
             this.Options = new Dictionary<string, string>();
         }
         private IDictionary<string, string> messageStrings;
         private IDictionary<string, string> richMessageStrings;
 
-        abstract protected ResourceManager ResourceManager { get; }
+        virtual protected ResourceManager ResourceManager => throw new NotImplementedException();
 
-        abstract protected IEnumerable<string> MessageResourceNames { get; }
+        virtual protected IEnumerable<string> MessageResourceNames => throw new NotImplementedException();
```
lgolding: >virtual [](start = 8, length = 7)

Sure, don't require people who don't need these to implement them.

src/Sarif.Driver/Sdk/ISkimmer.cs
```diff
@@ -1,43 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-
-namespace Microsoft.CodeAnalysis.Sarif.Driver
-{
-    public interface ISkimmer<TContext> : IRule
```
lgolding: >ISkimmer [](start = 21, length = 8)

Yay! Glad to see it go.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -829,6 +819,24 @@ private void Visit(ThreadFlowLocation threadFlowLocation, string threadFlowLocat
         private void Visit(Tool tool, string toolPointer)
         {
             Analyze(tool, toolPointer);
+
+            if (tool.NotificationsMetadata != null)
+            {
+                string notificationsPointer = toolPointer.AtProperty(SarifPropertyName.NotificationsMetadata);
+                for (int i = 0; i < tool.RulesMetadata.Count; ++i)
```
lgolding: >RulesMetadata [](start = 41, length = 13)

BUG! `NotificationsMetadata`. Copy/paste FTL. Apparently missing some test coverage.

lgolding:
>         public const string Resources = "resources";

Remove `Resources`.

---
Refers to: src/Sarif.Multitool/SarifPropertyName.cs:54 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

src/Sarif.Driver.UnitTests/TestRuleBase.cs
```diff
@@ -54,17 +54,14 @@ public override RuleConfiguration Configuration
 
         public override Message Help { get { return new Message() { Text = "[Empty]" }; } }
 
-        public abstract void Analyze(TestAnalysisContext context);
```
lgolding: >Analyze [](start = 29, length = 7)

Yes, now that we are an actual `Skimmer` instead of an `ISkimmer`, we already _have_ an abstract `Analyze` method. And that's why everything else is now an override. Nice.

src/Sarif/IResultLogWriter.cs
```diff
@@ -41,15 +41,15 @@ public interface IResultLogWriter
...
 
         /// <summary>
-        /// Write information about rules to the log. This information may appear
-        /// after the results, as the relevant set of rules might not be known until
-        /// all results have been generated. A Sarif file may also contain only rules
-        /// metadata.
+        /// Write information about the tool to the log. This information may appear
+        /// after the results, as some tool data, such as the relevant set of rule
+        /// observed during analysis, might not be known until all results have been
+        /// generated. A Sarif file may also contain only message metadata.
```
lgolding: >Sarif [](start = 25, length = 5)

SARIF

src/Sarif/IResultLogWriter.cs
```diff
@@ -41,15 +41,15 @@ public interface IResultLogWriter
...
         /// </summary>
-        /// <param name="rules">
-        /// An array whose elements comprise rules relevant to the analysis.
+        /// <param name="tool">
+        /// The tool object relevant to the current run.
```
lgolding: >relevant to the current run [](start = 28, length = 27)

I'd only use "X relevant to Y" when I can't find a more specific way to express the relationship between X and Y. How about "The tool that was executed in the current run".

src/Sarif/IResultLogWriter.cs
lgolding: I like that idea.

---
In reply to: [255236265](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255236265) [](ancestors = 255236265)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -821,11 +821,11 @@ internal ResultVersionOne CreateResultVersionOne(Result v2Result)
             return result;
         }
 
-        internal static RuleVersionOne CreateRuleVersionOne(IRule v2IRule)
+        internal static RuleVersionOne CreateRuleVersionOne(MessageDescriptor v2IRule)
         {
             RuleVersionOne rule = null;
 
-            var properties = v2IRule is Rule v2Rule ? v2Rule.Properties : null;
+            var properties = v2IRule.Properties;
```
lgolding: >var properties = v2IRule.Properties [](start = 12, length = 35)

No need for the temporary; just assign this to `RuleVersionOne.Properties` on L839.

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -821,11 +821,11 @@ internal ResultVersionOne CreateResultVersionOne(Result v2Result)
             return result;
         }
 
-        internal static RuleVersionOne CreateRuleVersionOne(IRule v2IRule)
+        internal static RuleVersionOne CreateRuleVersionOne(MessageDescriptor v2IRule)
```
lgolding: >v2IRule [](start = 78, length = 7)

`v2MessageDescriptor`

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -807,15 +807,11 @@ internal Rule CreateRule(RuleVersionOne v1Rule)
...
+                rule.DefaultConfiguration = new RuleConfiguration
                 {
-                    rule.Configuration = new RuleConfiguration
-                    {
-                        DefaultLevel = level,
-                        Enabled = v1Rule.Configuration == RuleConfigurationVersionOne.Enabled
-                    };
-                }
+                    Level = level,
+                    Enabled = v1Rule.Configuration != RuleConfigurationVersionOne.Disabled
```
lgolding: >!= [](start = 51, length = 2)

Yup. Missing means enabled.

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -245,10 +231,9 @@ public void OpenResults()
 
             _jsonWriter.WritePropertyName("results");
             _jsonWriter.WriteStartArray(); // Begin: results
-            _writeConditions = Conditions.ResultsInitialized;
+            _writeConditions |= Conditions.ResultsInitialized;
```
lgolding: >|= [](start = 29, length = 2)

Yikes!

lgolding: Test coverage hole?

---
In reply to: [255310688](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255310688) [](ancestors = 255310688)

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -228,12 +227,9 @@ public virtual void Dispose()
                     _run.Invocations[0].EndTimeUtc = DateTime.UtcNow;
                 }
 
-                // Note: we write out the backing rules
-                // to prevent the property accessor from populating
-                // this data with an empty collection.
```
lgolding: >collection [](start = 43, length = 10)

Good lord, I don't even want to _try_ to remember what that comment might once have meant.

michaelcfanning:
good idea. i am about 10 minutes from authoring a script or code that just rewrites the autogenerated stuff to make all this work. :P

---
In reply to: [462066291](https://github.com/Microsoft/sarif-sdk/pull/1259#issuecomment-462066291) [](ancestors = 462066291)

src/Sarif.Converters/ToolFileConverterBase.cs
```diff
@@ -108,21 +101,6 @@ protected static Run PersistResults(IResultLogWriter output, IList<Result> resul
...
-
-            if (run.LogicalLocations?.Count > 0)
-            {
-                output.WriteLogicalLocations(run.LogicalLocations);
-            }
-
-            if (run.Resources?.Rules != null)
-            {
-                output.WriteRules(run.Resources?.Rules);
-            }
```
lgolding: >} [](start = 12, length = 1)

Done now by `output`'s `IResultLogWriter.Dispose`, as you said.

src/Sarif.UnitTests/Readers/IsSuppressedInSourceConverterTests.cs
michaelcfanning: Add some comments. If we act on the design ideas mentioned here for this writer, all these calls will go away in future.

---
In reply to: [255309635](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255309635) [](ancestors = 255309635,255249783)

src/Sarif.Driver/Sdk/DefaultDriverOptions.cs
```diff
@@ -31,7 +31,7 @@ private DefaultDriverOptions()
                 @"Enabled state of rule. Valid values: Default, Disabled, Warning, Error. A rule in the 'Default' " + 
                 @"state will raise all issues as errors or warnings according to how the issue is logged.");
```
lgolding: Is this true?

src/Sarif/Core/Tool.cs
michaelcfanning: No sir, the Any extension is associated with bool not Nullable<bool>. Here's the most concise form I've come up with, building on your idea.

return this.RulesMetadata != null && this.RulesMetadata.Any(n => n != null);

---
In reply to: [255306805](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255306805) [](ancestors = 255306805)

src/Sarif.Driver/Sdk/ExportRulesMetadataCommandBase.cs
```diff
@@ -117,37 +117,17 @@ private void OutputSarifRulesMetada(string outputFilePath, ImmutableArray<IRule>
...
-            run.Resources = new Resources
-            {
-                Rules = new List<Rule>()
-            };
-
-            SortedDictionary<int, Rule> sortedRules = new SortedDictionary<int, Rule>();
-
-            foreach (IRule rule in skimmers)
-            {
-                var newRule = new Rule();
```
lgolding: >newRule [](start = 20, length = 7)

Weird. Do you remember what this was for?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -126,18 +126,17 @@ private static bool ApplyChangesFromTC31(JObject sarifLog)
...
-                            if ("existing".Equals(baselineState))
-                            {
-                                result["baselineState"] = "unchanged";
-                                modifiedLog = true;
-                            }
+                            // https://github.com/oasis-tcs/sarif-spec/issues/312
+                            modifiedLog |= UpdateBaselineExistingStateToUnchanged(result);
 
+                            // https://github.com/oasis-tcs/sarif-spec/issues/317
                             modifiedLog |= SetResultKindAndFailureLevel(result);
```
lgolding: >SetResultKindAndFailureLevel [](start = 43, length = 28)

So nice! isolating each issue-related change into a named method, even if it's only a couple of lines.

src/Sarif/Core/Tool.cs
michaelcfanning: For two one-liners only? Not super compelling. But! a generic helper for IEnumerable<T> s/be usable in lots of places. I see something that already exists for IList<string> that could be generalized. BUT! This code takes advantage of the trick that an extension method can be passed (and properly handle) a null 'this' instance. This avoids the need for the explicit null check in the caller but breaks the principle of least surprise. still might be worth it. take a look when i push.

---
In reply to: [255307231](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255307231) [](ancestors = 255307231)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -146,13 +145,83 @@ private static bool ApplyChangesFromTC31(JObject sarifLog)
             return modifiedLog;
         }
 
+        private static bool MoveRulesMetadataAndConfiguration(JObject run)
+        {
+            if (!(run["resources"] is JObject resources))
```
lgolding: >! [](start = 16, length = 1)

Here's a case where you _did_ use the new pattern in a negative case, and I think it's just fine.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -146,13 +145,83 @@ private static bool ApplyChangesFromTC31(JObject sarifLog)
...
+            {
+                return false;
+            }
+
+            JObject tool = (JObject)run["tool"];
+
+            // https://github.com/oasis-tcs/sarif-spec/issues/311
```
lgolding: >// [](start = 12, length = 2)

Make this a method-level comment?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -146,13 +145,83 @@ private static bool ApplyChangesFromTC31(JObject sarifLog)
...
+
+            // 1. 'run.resources.messageStrings' moves to 'run.tool.globalMessageStrings'
+            if (resources["messageStrings"] is JObject messageStrings)
+            {
+                tool["globalMessageStrings"] = messageStrings;
+                resources["messageStrings"] = null;
```
lgolding: >resources["messageStrings"] = null; [](start = 16, length = 35)

No need to do this since you're going to get rid of `resources` entirely.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -146,13 +145,83 @@ private static bool ApplyChangesFromTC31(JObject sarifLog)
...
+            return true;
+        }
+
+        private static bool UpdateBaselineExistingStateToUnchanged(JObject result)
+        {
+            bool modifiedLog = false;
+
+            // Rename 'existing' baseline state to 'unchanged'
+            // (as part of adding the 'updated' state, which 
+            // will not exist in any legacy SARIF logs.
```
lgolding: >. [](start = 54, length = 1)

Close paren. And again, I think this and the issue URI should be a method level comment.

src/Sarif/Core/Tool.cs
lgolding: Yes, even for two-liners. Consider:

1. I had a comment on your implementation. You had to change it in two places.
2. The reader has to convince himself that the two implementations _are_ the same. My formulation makes that explicit; no thought required.

---
In reply to: [255311391](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255311391) [](ancestors = 255311391,255307231)

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
michaelcfanning: Honestly can't account for it. I was required to add this reference when I took a dependency on the private Jschema code i checked into multitool. After removing that (and doing other work) I was unable to remove this reference, either from the IDE UI or by editing the csproj. I must have somehow introduced a new dependency on this project. Will look for the relevant code in other changes. I need a resolution of 'beats me' to close this comment, choosing another less precise option instead.

---
In reply to: [255307688](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255307688) [](ancestors = 255307688)

src/Sarif/VersionOne/Readers/SarifContractResolverVersionOne.cs
michaelcfanning: this is right! this converter existed due to a bug in JSON.NET v6 which we no longer have a dependency on. So this converter can disappear entirely.

---
In reply to: [255308073](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255308073) [](ancestors = 255308073)

src/Sarif/Readers/RuleDictionaryConverter.cs
```diff
@@ -1,70 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using System;
-using System.Collections.Generic;
-using Newtonsoft.Json;
-
-namespace Microsoft.CodeAnalysis.Sarif.Readers
-{
-    /// <summary>
-    /// This converter only exists because the version of Json.NET we're using (6.0.8)
```
michaelcfanning: >This converter only exists because the version of Json.NET we're using (6.0.8) [](start = 8, length = 78)

@larrygolding FYI.

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: ok, good catch. for now, I've externalized run.tool. this isn't a coherent solution because our schema requires some representation of tool to be a valid log file (presumably we could specify that the external property merges properly with it, dunno). anyway, i'm not worrying too much about this now, as we have a plan to hoist this thing to the root and drive it against JSON paths, as you noted.

---
In reply to: [255308008](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255308008) [](ancestors = 255308008)

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/Inputs/RunResources.sarif
michaelcfanning: the count of reviewers who would notice this subtlety is drastically smaller than non-members of the set. :)

---
In reply to: [255309492](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255309492) [](ancestors = 255309492)

src/Sarif.UnitTests/TestData/PrereleaseCompatibilityTransformer/Inputs/RunResources.sarif
michaelcfanning: yes! this data is weird. but we don't have a way to express the whole world. in the new world, we have a default enabled state _and_ a rule configuration on the invocation that may diverge. I added enabled == false here to get a non-default value for the bool. to make the example consistent, I need to add a new rule that is disabled (with no corresponding output). Will do that.

---
In reply to: [255309560](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255309560) [](ancestors = 255309560)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
michaelcfanning: Yes! I was going to add a test but there's no prerelease v2 notifications metadata equivalent, it's entirely new to TC31. Still, we can't leave this gap: I will convert a single file to current v2 so we can close it.

---
In reply to: [255309858](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255309858) [](ancestors = 255309858)

src/Sarif/Writers/ResultLogJsonWriter.cs
michaelcfanning: Yes there was. The introduction of Dispose handling has made sure we cover relevant code paths, so no new explicit test required. (This is, of course, how I found the bug, provoking relevant code paths due to writer changes...)

---
In reply to: [255310696](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255310696) [](ancestors = 255310696,255310688)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: Done and searched for other linger IRule references in names, none found.

---
In reply to: [255310270](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255310270) [](ancestors = 255310270)

michaelcfanning:
>         public Run Run { get; set;  }

This test class is not greatly and thinly built out. Still I will gut it further by cleaning some things up.

---
In reply to: [462066656](https://github.com/Microsoft/sarif-sdk/pull/1259#issuecomment-462066656) [](ancestors = 462066656,462066560)

---
Refers to: src/Sarif.TestUtilities/ResultLogObjectWriter.cs:20 in cde93c2. [](commit_id = cde93c292db9fafe708be64e91d088d744362abf, deletion_comment = False)

src/Sarif.Driver/Sdk/DefaultDriverOptions.cs
michaelcfanning: Yes, BUT! you have made me realize RuleEnabledState can now be entirely zapped in preference of FailureLevel. Splitting that out was really a good change, as two other enums have converged to it. i'm glad we gave this a second thought in the f2f

---
In reply to: [255311140](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255311140) [](ancestors = 255311140)

src/Sarif.Driver/Sdk/ExportRulesMetadataCommandBase.cs
michaelcfanning: Yes, note that the incoming param was an IRule. We had to convert to a concrete rule instance for some reason (to get access to downstream API). This horribleness drove the extirpation of IRule, now completely departed.

---
In reply to: [255311248](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255311248) [](ancestors = 255311248)

src/Sarif.ValidationTests/Sarif.ValidationTests.csproj
lgolding: If you remove it, what's the diagnostic on the build break?

---
In reply to: [255312082](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255312082) [](ancestors = 255312082,255307688)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: oh right, turns out i did convert all cases. only remaining ones relate to use of string type...

---
In reply to: [255311429](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255311429) [](ancestors = 255311429)

michaelcfanning:
good idea but i will do this next round (since i will need to regen anyway..)

---
In reply to: [462073597](https://github.com/Microsoft/sarif-sdk/pull/1259#issuecomment-462073597) [](ancestors = 462073597,462066291)

src/Sarif.Multitool.FunctionalTests/Rules/UrisMustBeValidTests.cs
michaelcfanning: hey! I only did this because the comments bothered me with no bracing. then i brought the single other test into conformance. you can revisit this point if you like. :) just revert it next time you're in here. i'm prepared to lose the argument.

---
In reply to: [255308227](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255308227) [](ancestors = 255308227)

src/Sarif/ExtensionMethods.cs
```diff
@@ -14,9 +14,15 @@ namespace Microsoft.CodeAnalysis.Sarif
...
     public static class ExtensionMethods
     {
-        public static bool HasAtLeastOneNonNullValue(this IList<string> list)
+        public static bool HasAtLeastOneNonNullValue<T>(this IEnumerable<T> collection)
         {
-            return list != null && list.Any((s) => s != null);
+            return collection != null && collection.Any((m) => m != null);
+        }
+
+        public static bool HasAtLeastOneNonDefaultValue<T>(this IEnumerable<T> collection, IEqualityComparer<T> comparer) where T : new()
```
michaelcfanning: >HasAtLeastOneNonDefaultValue [](start = 27, length = 28)

Note that we could eliminate the argument from this helper if ValueEquals was a member of ISarifNode. This would require a generic interface, however.

src/Sarif/Core/Run.cs
```diff
@@ -145,14 +145,14 @@ public bool ShouldSerializeColumnKind()
...
 
-        public bool ShouldSerializeInvocations() { return this.Invocations != null && this.Invocations.Any((e) => e != null && !e.ValueEquals(EmptyInvocation)); }
+        public bool ShouldSerializeInvocations() { return this.Invocations.HasAtLeastOneNonDefaultValue(Invocation.ValueComparer); }
 
-        public bool ShouldSerializeLogicalLocations() { return this.LogicalLocations?.Count > 0; }
+        public bool ShouldSerializeLogicalLocations() { return this.LogicalLocations.HasAtLeastOneNonNullValue(); }
 
-        public bool ShouldSerializeNewlineSequences() { return this.NewlineSequences != null && this.NewlineSequences.Any((s) => s != null); }
+        public bool ShouldSerializeNewlineSequences() { return this.NewlineSequences.HasAtLeastOneNonNullValue(); }
     }
```
lgolding: >} [](start = 4, length = 1)

Very nicely DRYed out.

src/Sarif.Multitool.FunctionalTests/TestData/UriMustBeAbsolute/Invalid.sarif
```diff
@@ -1,27 +1,18 @@
...
-          ]
-        }
-      ],
-      "originalUriBaseIds": {
-        "SRCROOT": { "uri": "Code/sarif-sdk/src" }
-      },
-      "resources": {
-        "rules": [
+        "downloadUri": "tools/codescanner/download.html",
+        "notificationsMetadata": [
```
lgolding: >notificationsMetadata [](start = 9, length = 21)

:-) this doesn't actually expose the copy/paste bug I found, which would have been triggered if the number of notification metadatas was different from the number of rule metadatas.

michaelcfanning: this change actually did provoke a test failure, i developed the test first to prove it, then took a fix. I've added a second notification, however. I also corrected the notification HTML page name to correspond to the id. :) This change so close to perfection, let's just nail it. 

---
In reply to: [255316074](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255316074) [](ancestors = 255316074)

michaelcfanning: remember, these tests are driven off JSON paths to the issue location. so after my original change, the test failed because we found an issue at "runs[0].tool.notificationsMetadata[0].helpUri". which wasn't documented in the expected results properties.

---
In reply to: [255346258](https://github.com/Microsoft/sarif-sdk/pull/1259#discussion_r255346258) [](ancestors = 255346258,255316074)

PR #1262, opened 2019-02-05
Revert unintentional change to BaselineState

lgolding:
The `develop` branch should match TC <span>#</span>30, but we inadvertently introduced a change from  TC <span>#</span>31: replacing `BaselineState.Existing` with `.Unchanged` and `Updated`.

I did not revert the entire change. Some things (like having AppVeyor build the `tc-31` branch instead of the defunct `files-array` branch, and some C# 7 updates to the `PrereleaseCompatibilityTransformer`) were good, and I kept them.

Also:
- Update the version to `2019-01-09` in preparation for merge to `master`.


lgolding:

lgolding:
@michaelcfanning I merged this. Feel free to comment _post hoc_. I'm now going to apply the changes from `master` (Bogdan's point fix) to `develop`, in preparation for merging `develop` to `master` and publishing 2019-01-09.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -47,28 +47,23 @@ public static class PrereleaseCompatibilityTransformer
...
                 {
-                        modifiedLog |= ApplyChangesFromTC31(sarifLog);
-                        break;
+                    // SARIF TC30. Nothing to do.
+                    break;
                 }
 
                 case "2.0.0-csd.2.beta.2018-10-10":
+                case "2.0.0-csd.2.beta.2018-10-10.1":
+                case "2.0.0-csd.2.beta.2018-10-10.2":
```
michaelcfanning: this is wrong. these aren't valid schema references from the JSON. all these versions should refer to the same JSON schema, the point updates are for the SDK backing them (i.e., the semver associated with the nuget package updates).

michaelcfanning:
@lgolding found a problem but in practice might not be worth fixing (since no one s/be producing SARIF files against the bogus versions referenced).

lgolding:
@michaelcfanning You're right. I will fix it as soon as we decide what branch we're in going forward :-)

PR #1264, opened 2019-02-06
Merge changes from develop to master, to publish TC #30 (2019-01-09).

lgolding:
After syncing to the latest `master`, I immediately branched from `master` to `users/lgolding/publish-2019-01-09`. Then I merged `develop` into `users/lgolding/publish-2019-01-09`. After resolving the merge conflicts (taking the version from `develop` in each case, after verifying its correctness), I had to make additional changes to fix unit tests that were broken because of changes that were logically incorrect but did not cause a merge conflict:

1. Several tests failed because the `PrereleaseCompatibilityTransformer` was creating SARIF v2 files with version `"…2019-01-"` instead of `"…2019-01-09"`. It turned out that the `SarifUtilities.SemanticVersion` property still had code to trim the last two characters from the version string (Michael added this code to deal with the trailing `".1"` when he made a point fix for Bogdan).

2. One test failed because, when I previously cherry-picked the Bogdan fix from `master` to `develop` in preparation for this merge, I fixed the white space in one of the `result` objects in one of the test data input files. The merged input file included _both_ results, the badly white-spaced one and the correctly white-spaced one. As a result, the test produced two `result` objects when it only expected one.

After fixing both these problems, all the tests passed in the `users/lgolding/publish-2019-01-09`, so I committed the merge to that branch, and now we're ready to merge it to `master`.


lgolding:

lgolding:

michaelcfanning:
@lgolding as per offline discussion, merging this change.

PR #1266, opened 2019-02-07
V2.0.0 csd.2.beta.2018 10 10.3

michaelcfanning:
@lgolding @EasyRhinoMSFT

### DIFF
--- a/src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v1/OneRunWithFiles.sarif
+++ b/src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v1/OneRunWithFiles.sarif
@@ -12,6 +12,10 @@
             {
               "algorithm": "sha256",
               "value": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592"
+            },
+            {
+              "algorithm": "md5",
+              "value": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592"
             }
           ]
         },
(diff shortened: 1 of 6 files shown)


michaelcfanning:
I didn't handle the branching correctly here. This PR compares to 10.10.2 purely for informational purposes. If you agree with the delta than the 10-10.3 branch is correct. What is should have done was to create a 10-103. branch, then branched off that to cherry-pick the relevant change. Please review and signoff. If you agree it's right, I'll create a pipeline to ship from the 10.3 branch. I will not, obviously, merge back into 10.10.2

src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/v1/OneRunWithFiles.sarif
```diff
@@ -12,6 +12,10 @@
             {
               "algorithm": "sha256",
               "value": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592"
+            },
+            {
+              "algorithm": "md5",
```
lgolding: >md5 [](start = 28, length = 3)

RIght, you add the test case to show that you can carry MD5 over.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -157,7 +157,7 @@ internal FileData CreateFileData(FileDataVersionOne v1FileData)
             {
                 fileData = new FileData
                 {
-                    Hashes = BuildHashesDictionary(v1FileData.Hashes),
+                    Hashes = v1FileData.Hashes?.Select(CreateHash).ToDictionary(p => p.Key, p => p.Value),
```
lgolding: >CreateHash [](start = 55, length = 10)

Because `CreateHash` does the `ToLowerInvariant` call that `BuildHashesDictionary` lacked.

lgolding: No, I see, it's because it handles values that are _not_ in `Utilities.AlgorithmKindNameMap` -- which "md5" is not.

---
In reply to: [254911070](https://github.com/Microsoft/sarif-sdk/pull/1266#discussion_r254911070) [](ancestors = 254911070)

lgolding:
Closing without merging. We want this on a separate branch.

PR #1267, opened 2019-02-07
Support for Sarif comparison

ScottLouvau:
- Re-enabling result matching mode in Sarif.Multitool.
- Checking in sample to convert Sarif to CSV to easily review compared results.

We should make a more comprehensive Sarif to CSV converter, but I wanted to check in this simple implementation as a sample before the work to make it more configurable and integrated with the SDK can be done.

### DIFF
--- a/src/Samples/SarifToCsv/Program.cs
+++ b/src/Samples/SarifToCsv/Program.cs
@@ -1,0 +1,86 @@
+﻿using AzureDevOpsCrawlers.Common.IO;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Newtonsoft.Json;
+using System;
+using System.Diagnostics;
+using System.Linq;
+
+namespace SarifToCsv
+{
+    /// <summary>
+    ///  SarifToCsv is a simple example of converting a few key SARIF result properties to CSV format
+    ///  to allow easily viewing results compared with the 'match-results-forward' Sarif Multitool mode.
+    /// </summary>
+    /// <remarks>
+    ///  Should be modified to:
+    ///    - Make the output columns configurable.
+    ///    - Enable writing more result properties.
+    ///    - Integrate as a Sarif Multitool mode.
+    /// </remarks>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length < 2)
+            {
+                Console.WriteLine("Usage: SarifToCsv [sarifFilePath] [csvFilePath]");
+                return;
+            }
+
+            try
+            {
+                string sarifFilePath = args[0];
+                string csvFilePath = args[1];
+
+                Console.WriteLine($"Converting \"{sarifFilePath}\" to \"{csvFilePath}\"...");
+                Stopwatch w = Stopwatch.StartNew();
+
+                SarifLog log = null;
+
+                // Read the SarifLog with the deferred reader
+                JsonSerializer serializer = new JsonSerializer();
+                serializer.ContractResolver = new SarifDeferredContractResolver();
+                using (JsonPositionedTextReader jtr = new JsonPositionedTextReader(sarifFilePath))
+                {
+                    log = serializer.Deserialize<SarifLog>(jtr);
+                }
+
+                // Convert the each result location into a row in a CSV
+                using (CsvWriter writer = new CsvWriter(csvFilePath))
+                {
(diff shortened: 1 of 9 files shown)


src/Samples/SarifToCsv/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    writer.SetColumns(new string[] { "RuleId", "BaselineState", "Location.Region.Snippet.Text", "Location.Region.StartLine", "Location.Region.StartColumn", "Location.Uri", "Message.Text", "Fingerprint" });
+
+                    foreach (var run in log.Runs)
+                    {
+                        foreach (var result in run.Results)
+                        {
+                            foreach (PhysicalLocation location in result.PhysicalLocations())
+                            {
+                                writer.Write(result.RuleId);
```
michaelcfanning: >writer [](start = 32, length = 6)

Interesting to think about how this could be made configurable. I.e., if we could specify desirable columns in the config file or on the command-line, how to associate data emit corresponding to these settings.

michaelcfanning:
> using AzureDevOpsCrawlers.Common.IO;

Need MIT header here and in some other files. Also, namespaces aren't sorted propertly.

---
Refers to: src/Samples/SarifToCsv/Program.cs:1 in a88fb1d. [](commit_id = a88fb1df4454fb32b6a4c080eb107070100db103, deletion_comment = False)

src/Samples/SarifToCsv/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+    ///  SarifToCsv is a simple example of converting a few key SARIF result properties to CSV format
+    ///  to allow easily viewing results compared with the 'match-results-forward' Sarif Multitool mode.
+    /// </summary>
+    /// <remarks>
+    ///  Should be modified to:
+    ///    - Make the output columns configurable.
+    ///    - Enable writing more result properties.
+    ///    - Integrate as a Sarif Multitool mode.
+    /// </remarks>
+    class Program
```
michaelcfanning: >class [](start = 4, length = 5)

another nit, make accessibility explicit

src/Samples/SarifToCsv/SdkExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+
+        public static string FileUri(this FileLocation fileLocation, Run run)
+        {
+            if (fileLocation == null)
+            {
+                return null;
+            }
+            else if (fileLocation.Uri != null)
```
michaelcfanning: >else if [](start = 12, length = 7)

you don't required chained ifs here since each conditional clause returns. would make more concise to eliminate them (as well as final else clause.

src/Samples/SarifToCsv/SdkExtensions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.CodeAnalysis.Sarif;
+using System.Collections.Generic;
+
+namespace SarifToCsv
+{
+    public static class SdkExtensions
+    {
+        public static IEnumerable<PhysicalLocation> PhysicalLocations(this Result result)
+        {
+            foreach (Location location in result.Locations)
```
michaelcfanning: >result.Locations [](start = 42, length = 16)

this code null derefs for results with no locations. this pattern (to make basic enumeration of nested things like locations more convenient) is an interesting idea for core sdk.

michaelcfanning:
This is all fine to check in. Noted some nits. But can you create a PR against master? @lgolding has argued persuasively for the abandonment of develop.

src/Samples/SarifToCsv/Program.cs
ScottLouvau: Fixed. This is the default in the current VS project templates.

src/Samples/SarifToCsv/SdkExtensions.cs
ScottLouvau: Fixed. Yeah, I think SDK users are likely to want to traverse all of various properties under the result which require intermediate foreach traversals. It'd be nice to put some helpers in to make that easier.

src/Samples/SarifToCsv/SdkExtensions.cs
ScottLouvau: I intended this pattern to help readability; the if.. else if... else if meaning to show we're only taking one of these paths.

ScottLouvau:
Abandoning in favor of PR 1278, which has all feedback here incorporated and will merge to master.

PR #1268, opened 2019-02-08
Initial version of the Contrast Security converter

lgolding:
Written my @michaelcfanning. Adapted to the 2019-01-09 version of the SARIF SDK by @lgolding.


lgolding:

lgolding:

lgolding:

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+    /// Converts exported Contrast Security XML report files to sarif format
+    /// </summary>
+    ///<remarks>
+    ///</remarks>
+    internal sealed class ContrastSecurityConverter : ToolFileConverterBase
+    {
+        private const string ContrastSecurityRulesData = "Microsoft.CodeAnalysis.Sarif.Converters.RulesData.ContrastSecurity.sarif";
+
+        private IDictionary<string, Rule> _rules;
+        private HashSet<FileData> _files;
```
lgolding: >HashSet [](start = 16, length = 7)

One of my changes to adapt to 2019-01-09. Used to be a dictionary.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+            using (var streamReader = new StreamReader(stream))
+            {
+                string prereleaseRuleDataLogText = streamReader.ReadToEnd();
+                PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(prereleaseRuleDataLogText, forceUpdate: true, Newtonsoft.Json.Formatting.Indented, out string currentRuleDataLogText);
+                sarifLog = JsonConvert.DeserializeObject<SarifLog>(currentRuleDataLogText);
+            }
+
+            // 2. Retain a pointer to the rules dictionary, which we will use to set rule severity
+            Run run = sarifLog.Runs[0];
+            _rules = run.Resources.Rules.ToDictionary(rule => rule.Id);
```
lgolding: >ToDictionary [](start = 41, length = 12)

The other change to adapt to 2019-01-09. `run.resources.rules` is an array, but `_rules` needs to be a dictionary because you look up default severities by rule id. Of course this wouldn't work if there were two elements in the rules array with the same id, but I assume that doesn't happen for Contrast Security.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+
+            // 1. First we initialize our global rules data from the SARIF we have embedded as a resource
+            Assembly assembly = typeof(ContrastSecurityConverter).Assembly;
+            SarifLog sarifLog;
+
+            using (var stream = assembly.GetManifestResourceStream(ContrastSecurityConverter.ContrastSecurityRulesData))
+            using (var streamReader = new StreamReader(stream))
+            {
+                string prereleaseRuleDataLogText = streamReader.ReadToEnd();
+                PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(prereleaseRuleDataLogText, forceUpdate: true, Newtonsoft.Json.Formatting.Indented, out string currentRuleDataLogText);
```
lgolding: >UpdateToCurrentVersion [](start = 51, length = 22)

The Contrast Security rules log is in 10-10 format -- containing a rules dictionary -- and it was easier to do this than to update the file. In a separate PR we can run Multitool to update it to 2019-01-09.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+            //}
+
+            output.OpenResults();
+            output.WriteResults(results);
+            output.CloseResults();
+        }
+
+        internal Result CreateResult(ContrastLogReader.Context context)
+        {
+            switch (context.RuleId)
```
lgolding: >RuleId [](start = 28, length = 6)

Oh, Harleen will have _lots_ of fun filling out the rest of the cases :-)

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+                    return ConstructWebApplicationDeployedinDebugModeResult(context.Properties);
+                }
+
+                default:
+                {
+                    return ConstructNotImplementedRuleResult(context.RuleId);
+                }
+            }
+
+            throw new InvalidOperationException();
```
lgolding: >InvalidOperationException [](start = 22, length = 25)

IMO this should be `ArgumentException`. The method was called with a `context` object containing an unrecognized `RuleId`. `InvalidOperationException` means "the object can't do what you asked in its current state".

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+        }
+
+        private Result ConstructNotImplementedRuleResult(string ruleId)
```
lgolding: >private [](start = 8, length = 7)

And now I have to confess that I didn't review the remaining 1000 lines. I suppose I really should, huh?

What happened was, I reviewed enough of it to understand what I had to do to make it work with 2019-01-09, and then I stopped because I thought it important to get this in to `master`. How about this -- we merge this, to give you and/or me or Harleen the ability to move forward, and I review the existing code _post hoc_, and submit a PR with my suggested changes.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+            var reader = new ContrastLogReader();
+            reader.FindingRead += (ContrastLogReader.Context current) => { results.Add(CreateResult(current)); };
+            reader.Read(context, input);
+
+            // 4. Construct the files array, based on all results returned
+            var fileInfoFactory = new FileInfoFactory(MimeType.DetermineFromFileExtension, dataToInsert);
+            HashSet<FileData> files = fileInfoFactory.Create(results);
+
+            // 5. Finally, complete the SARIF log file with various tables and then the results
+            output.Initialize(run);
```
lgolding: >Initialize [](start = 19, length = 10)

I think we should be calling `ToolFileConverterBase.PersistResults` here.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -0,0 +1,1364 @@
...
+
+            if (LogicalLocations?.Any() == true)
+            {
+                output.WriteLogicalLocations(LogicalLocations);
+            }
+
+            //if (_rules?.Any() == true)
+            //{
+            //    output.WriteRules(_rules.Values.ToList());
+            //}
```
lgolding: >//} [](start = 12, length = 3)

With this code uncommented, the rules were being persisted twice. I need to find out why.

lgolding:
As it stands, this PR accomplishes the goal of making the Contrast Security converter work on the 2019-01-09 version of the SDK. But there are several things I still want to do:

1. Review the remaining 1000-odd lines of `finding` converter code.
2. Add a functional test for the WebGoat findings.
3. DRY out the code by modifying the `Convert` method to call the base class's `PersistResults` method.

I suggest we merge this as is and let me do the rest of those things _post hoc_. That will unblock us to make further progress while I do that.

michaelcfanning:
agreed. LGTM. Let's merge for now. Please note that we have additional rules to cover from the contrast security rule set as well.

lgolding:
Yes, I thought that might be a good first task for Harleen. That's the "further progress" I was referring to in my comment above.

---
In reply to: [461940054](https://github.com/Microsoft/sarif-sdk/pull/1268#issuecomment-461940054) [](ancestors = 461940054)

PR #1269, opened 2019-02-08
Address vulnerability in dev dependency lodash 4.14.

lgolding:


michaelcfanning:
The actual issue flagged by githhub is the reference in TypeScript\Sarif.SDK\package-lock.json. it seems to me we should remove that reference.

lgolding:
Closing per offline conversation with @michaelcfanning. @EasyRhinoMSFT will take this on. Michael will communicate details.

PR #1276, opened 2019-02-11
Handle relative paths for uri base ids

michaelcfanning:
@lgolding @jeffersonking

It is useful to allow command-line args that set uri base id values to be rendered as relative paths (that are subsequently resolved to absolute paths by the tool). This makes it more convenient to script rewrites and transformations in build agent environments.


src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -655,6 +655,10 @@ private Stream GetContentStream(Uri uri, out Encoding encoding)
                             {
                                 failureReason = $"File '{uri.LocalPath}' could not be accessed: {ex.ToString()}";
                             }
+                            catch (UnauthorizedAccessException ex)
```
michaelcfanning: >UnauthorizedAccessException [](start = 35, length = 27)

IOException, SecurityException and UnauthorizedAccessException are the exceptions that need to be handled for System.IO API that actually touch the file system (some API, such as the path construction API might not actually need to interop with the file system). Someday we could write a little action/delegate helper that wraps a handler of this kind and returns any exception that's been raised to DRY out this pattern. Might be a nice small thing for Harleen.

lgolding: Consider DRYing this out with a C# 6.0 exception filter:
```
catch (Exception ex) when (
    ex is IOException ||
    ex is SecurityException ||
    ex is UnauthorizedAccessException)
{
    failureReason = $"File '{uri.LocalPath}' could not be accessed: {ex}"
}
```

(Also note that you don't need to call `ToString()`.)

---
In reply to: [255589393](https://github.com/Microsoft/sarif-sdk/pull/1276#discussion_r255589393) [](ancestors = 255589393)

src/Sarif.Driver/Sdk/DriverSdkExtensions.cs
```diff
@@ -16,7 +21,39 @@ public static class DriverSdkExtensions
...
+                        }
+                        catch (ArgumentException) { } // illegal file path characters throw this
+                        catch (PathTooLongException) { }
+                    }
+                }
+                catch (UriFormatException) { }
+
+                if (value == null)
+                {
+                    throw new InvalidOperationException("Could not construct absolute URI from specified value: " + tokens[1]);
```
lgolding: >absolute [](start = 77, length = 8)

If you get here you couldn't construct a relative URI either, right?

lgolding:
Let me make sure I understand how the fix helps. It allows you to specify a relative URI on the command line:
```
-u SRCROOT=..\..\..\..\..\ 
```

The code will transform this to an absolute path, which might differ depending on the file system layout of the machine where the script runs. So -- the SARIF file might differ depending on the machine, but that same _script_ can run on all machines.

Is that right?

michaelcfanning:
Yes. The point of a uri base id is to render the immediate absolute path. But scripting environments themselves often need to use relative paths to refer to things. So we allow these relative paths to be passed to our utilities where they will be resolved as absolute paths. this need identified by pipeline processing for our Semmle analysis results. Semmle log files don't contain snippets by default, neither do these log files populate originalUriBaseIds object. So our utility must do so in order to support post-processing for snippet injection.

---
In reply to: [462512599](https://github.com/Microsoft/sarif-sdk/pull/1276#issuecomment-462512599) [](ancestors = 462512599)

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
michaelcfanning: This code isn't the focus of the change, added UnauthorizedAccessException here for completeness. We'll clean this up later, sure.

---
In reply to: [255709280](https://github.com/Microsoft/sarif-sdk/pull/1276#discussion_r255709280) [](ancestors = 255709280,255589393)

PR #1283, opened 2019-02-13
Multiformat message strings

michaelcfanning:
@lgolding schema and release note changes only for your early review

PR #1287, opened 2019-02-14
Fix ComprehensiveToolProperties test.

lgolding:
We fix a test that should have failed as a result of the previous PR (#1283, MultiformatMessageStrings) but did not.

The input file was stamped with the wrong version, so the `PrereleaseCompatibilityTransformer` didn’t properly populate `tool.driver.ruleDescriptors`. The expected output file was checked in without noticing that this property was missing, so the test "passed".

We fix the `version` property in the input file. This causes the test to fail as expected. Then we update the expected output file, where `tool.driver.ruleDescriptors` is not present.


lgolding:
@michaelcfanning This is merged.

PR #1289, opened 2019-02-15
Conversion tool

michaelcfanning:
@lgolding a good catch on your part. Please review/refine/merge.

### DIFF
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -218,6 +218,24 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
 
             run["tool"] = tool;
 
+            // 6. Update some properties on run.conversion, if present. Note that the
+            // notion of reportingDescriptors associated with the conversion did not
+            // exist previously, so no transformation is required here.
+            if (run["conversion"] is JObject conversion)
+            {
+                driver = (JObject)conversion["tool"];
+                
+                tool = new JObject(new JProperty("language", driver["language"] ?? "en-US"));
+
+                driver.Remove("language");
+                driver.Remove("sarifLoggerVersion");
+
+                tool["driver"] = driver;
+                conversion.Remove("tool");
+                conversion["tool"] = tool;
+                run["conversion"] = conversion;
+            }
+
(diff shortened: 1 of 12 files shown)


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -218,6 +218,24 @@ private static void MoveToolPropertiesIntoDriverToolComponent(JObject run)
 
             run["tool"] = tool;
 
+            // 6. Update some properties on run.conversion, if present. Note that the
+            // notion of reportingDescriptors associated with the conversion did not
+            // exist previously, so no transformation is required here.
+            if (run["conversion"] is JObject conversion)
+            {
+                driver = (JObject)conversion["tool"];
```
michaelcfanning: >driver [](start = 16, length = 6)

I explicitly did not invest in DRY'ing this code out with earlier code processing run.tool. I prefer the detailed breakout above. I also wonder whether the conversion processing may benefit from a big more thought. Are we completely comfortable with providing all new tool object functionality for conversion.too (such as the new reporting descriptors)? 

I propose that we accept this for now and close on the remainder TC31 items as quickly as possible (there are two more schema changes). We can refine conversion.tool in a furthewr conversation if necessary. This change should literally bring us as far as we designed through TC31. 

Please refine/merge according to your inclination. My internet access is a bit spotty while i'm away.

lgolding: Ack. I was already doing this but hadn't pushed PR yet. I suggest we each look at the other's and then decide which one to pick. I did in fact DRY out the "conversion". I will push in a few minutes.

---
In reply to: [257347046](https://github.com/Microsoft/sarif-sdk/pull/1289#discussion_r257347046) [](ancestors = 257347046)

lgolding: I changed my mind. The DRYing out turns out to be trickier than I thought. I will review/approve/merge this one.

---
In reply to: [257370187](https://github.com/Microsoft/sarif-sdk/pull/1289#discussion_r257370187) [](ancestors = 257370187,257347046)

lgolding:
LGTM.

PR #1295, opened 2019-02-19
Rule extension index

michaelcfanning:
Rename result.extensionIndex to result.ruleExtensionIndex. I updated the OASIS spec issue with this as well. I believe this is the last change before we ship 01-24.

 https://github.com/oasis-tcs/sarif-spec/issues/179


lgolding:
LGTM

michaelcfanning:
@lgolding looks like a transient appveyor failure, so merging.

Issue #1300, opened 2019-02-20
Rename result.kind 'none' value to 'debug'


michaelcfanning:
See "result" definition in the schema, for the 'kind' property. Rename 'none' value to 'debug'. Be sure to look at CodeGenHints to see if you need to specify the zero value (which debug is).

michaelcfanning:
Please update the release notes, this change will go into 2019-01-24.1. Also might as well add a note around charOffset default value change. :)

PR #1302, opened 2019-02-21
Renaming result.kind 'none' value to 'debug'

harleenkohli:
#1300


harleenkohli:
note to self: need to update release notes about the changes going in for 2019-01-24.1

PR #1311, opened 2019-02-27
adding taxonomies

harleenkohli:


michaelcfanning:
Can you double-check this change? I don't see that you've added the ruleDescriptor.taxonomyTags, etc.

harleenkohli:
abadoning this one, i am including these changes in a different PR shortly

PR #1315, opened 2019-02-28
additional props added

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2107,6 +2107,26 @@
...
+          "description": "A product suite to which the tool belongs.",
+          "type": "string"
+        },
+
+        "shortDescription": {
+          "description": "A brief, multiformat description of the tool.",
+          "$ref": "#/definitions/multiformatMessageString"
+        },
+
+        "longDescription": {
```
lgolding: >longDescription [](start = 9, length = 15)

Sorry, the issue was wrong. s/b `fullDescription` to match the corresponding property name in `reportingDescriptor`.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2107,6 +2107,26 @@
...
+          "type": "string"
+        },
+
+        "product": {
+          "description": "A product suite to which the tool belongs.",
+          "type": "string"
+        },
+
+        "shortDescription": {
+          "description": "A brief, multiformat description of the tool.",
```
lgolding: >multiformat [](start = 35, length = 11)

I don't think the word "multiformat" is needed in the description. Just "A brief description of the tool".

lgolding:
Approval subject to my two comments.

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: good catch

---
In reply to: [261382346](https://github.com/Microsoft/sarif-sdk/pull/1315#discussion_r261382346) [](ancestors = 261382346)

PR #1316, opened 2019-02-28
Schema only change for SARIF https://github.com/oasis-tcs/sarif-spec/issues/324

michaelcfanning:


michaelcfanning:
eballot is closed now.

PR #1318, opened 2019-03-01
Schema only changes for SARIF https://github.com/oasis-tcs/sarif-spec/issues/314

michaelcfanning:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1838,6 +1879,17 @@
           "default": "unicodeCodePoints"
         },
```
harleenkohli: >}, [](start = 8, length = 2)

side note: i forgot to update releasehistory.md

michaelcfanning:
Oasis eballot is closed now.

PR #1319, opened 2019-03-01
adding inline externalpropertyfiles : [https://github.com/oasis-tcs/sarif-spec/issues/321]

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -27,6 +27,16 @@
       }
     },
 
+    "inlineExternalPropertyFiles": {
+      "description": "References to external property files that share data between runs.",
```
lgolding: >References [](start = 22, length = 10)

@michaelcfanning @harleenkohli

I'm writing the change draft for this now. This property is not an array of _references_ to external property files; it is an array each element of which contains the _contents_ of an external property file. It's as if you took an actual external property file and copy/pasted its contents into one of the array elements.

How about this for the description:

"An array of externalPropertyFile objects containing data that is shared among the runs in the log file."

PR #1320, opened 2019-03-01
Caching mechanism for duped codeflow data : [https://github.com/oasis-tcs/sarif-spec/issues/320]

harleenkohli:
Add `run.threadFlowLocations`, an array of `threadFlowLocation` objects cached at the run level, analogous to `run.files` and `run.logicalLocation`.
Add `threadFlowLocation.index`, an index into `run.threadFlowLocations`.
Add `externalPropertyFiles.threadFlowLocations`, allowing cached data to be stored in a separate file.

### DIFF
--- a/src/Sarif/CodeGenHints.json
+++ b/src/Sarif/CodeGenHints.json
@@ -805,7 +805,7 @@
       "arguments": {
         "namespaceName": "Newtonsoft.Json",
         "typeName": "JsonConverter",
-        "arguments": [ "typeof(EnumConverter)" ]
+        "arguments": [ "typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter)" ]
       }
     }
   ],
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -522,7 +522,7 @@
         },
 
         "threadFlowLocations": {
-          "description": "An array of threadFlowLocation objects cached in external file.",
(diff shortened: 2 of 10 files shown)


src/Sarif/CodeGenHints.json
```diff
@@ -805,7 +805,7 @@
       "arguments": {
         "namespaceName": "Newtonsoft.Json",
         "typeName": "JsonConverter",
-        "arguments": [ "typeof(EnumConverter)" ]
+        "arguments": [ "typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter)" ]
```
harleenkohli: >[ "typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter)" ] [](start = 21, length = 64)

this was needed bcuz now we have ambiguous reference in the autogenerated\ThreadFlowLocation.cs for "EnumConverter" due to addition of System.ComponentModel namespace.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -521,6 +521,17 @@
           }
         },
 
+        "threadFlowLocations": {
+          "description": "An array of threadFlowLocation objects cached in external file.",
```
harleenkohli: >"An array of threadFlowLocation objects cached in external file." [](start = 25, length = 65)

please help improve this (and other descriptions), we also need to update the not-yet-autogen files if applicable.

Issue #1312, opened 2019-02-27
Validator can fail in pre-release compatibility transformer


lgolding:
The Sarif.Multitool `validate` command performs JSON schema validation before it runs the SARIF-specific validation rules. However, `ValidateCommand` _first_ runs the `PrereleaseCompatibilityTransformer` (which upgrades a downlevel v2 file to the current v2 schema), and only then performs schema validation against the current schema.

That makes sense: you can't validate a file authored to an old schema against a newer schema. The problem is that the `PrereleaseCompatibilityTransformer` implicitly assumes that certain aspects of the input file conform to the schema. For example:

```
var tool = (JObject)run["tool"];
```

This line assumes that the `tool` property of the `run` object is a JSON object, and it crashes if, for example, its value in the input document is a string:

```
"tool": "CodeScanner"
```

We could fix this by:

1. Adding a command line option `--no-compatibilty-transform` to the `validate` command. But that would only work if the input document only uses SARIF constructs that conform to the latest schema.

2. Avoid the use of casts such as `(JObject)run["tool"]`, and instead always use the `is` operator:

```
if (run["tool"] is JObject toolObject) {
    ...
}
```

If you do this, then it won't crash, and mistakes like setting `run.tool` to a string will be left behind for the schema validator to catch.

3. Perform the schema validation _before_ the compatibility transform. That would require you to ensure that the schema file specified by the `--json-schema` option matches the input document's `sarifLog.version` property. This option has the drawback that it won't work for downlevel files that erroneously declare there version as `2.0.0` (omitting the pre-release identifier).

@michaelcfanning, @harleenkohli FYI

lgolding:
FYI I think option 2 is best (has none of the drawbacks of options 1 or 3).

PR #1331, opened 2019-03-05
336 tool component deltas

harleenkohli:
delta changes from 336: https://github.com/oasis-tcs/sarif-spec/issues/336
items already in master:
- In the tool object:
    * Add a organization property of type string: the company or organization that produced the tool
    * Add a product property of type string: a product suite to which the tool belongs,
    * Add a shortDescription property of type multiformatMessageString: a brief, multiformat description of the tool
    * Add a fullDescription property of type multiformatMessageString: a comprehensive multiformat description of the tool

in this pr:
- In the tool object:
    * Convert artifactIndex to an array of artifact locations named artifactIndices


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -138,11 +138,61 @@ private static bool ApplyChangesFromTC32(JObject sarifLog)
...
             return true;
         }
 
+        private static void UpdateAllToolComponentProperties(JObject run)
+        {
+            // Access and modify run.tool
+            if (run["tool"] is JObject tool)
+            {
+                UpdateToolObjectToolComponentProperties(tool);
+            }
```
harleenkohli: >} [](start = 12, length = 1)

we talked about the unsafe cast in the context of a different issue (#1312), so i have added a condition check here. If tool doesnt exist, we can add throwing an error here to fail faster, i wasnt sure if i should just go ahead and do it hence leaving a comment for now :)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -138,11 +138,61 @@ private static bool ApplyChangesFromTC32(JObject sarifLog)
...
+            {
+                UpdateToolObjectToolComponentProperties(tool2);
+            }
+        }
+
+        private static void UpdateToolObjectToolComponentProperties(JObject tool)
+        {
+            // Access and modify tool.driver
+            if (tool["driver"] is JObject driver)
+            {
```
harleenkohli: this pattern is repeat of another set of methods which also navigate to toolComponent, i have not combined the two methods for keeping it simple but i am adding another issue (#1332) to write a general purpose to navigate to the input path as a separate item, so please consider this code temporary.

PR #1333, opened 2019-03-05
Provide --inline argument for RebaseUriCommand

michaelcfanning:


michaelcfanning:
@jeffersonking

src/Sarif.Multitool/RebaseUriCommand.cs
```diff
@@ -37,7 +37,11 @@ public int Run(RebaseUriOptions rebaseOptions)
 
                 var sarifFiles = GetSarifFiles(rebaseOptions);
 
-                Directory.CreateDirectory(rebaseOptions.OutputFolderPath);
+                if (!rebaseOptions.Inline)
```
jeffersonking: Can the mutual exclusivity be expressed high-up? Say the at the commandOptions level or at least as an assert for this command?

PR #1334, opened 2019-03-06
Fix bad type in external property files object.

michaelcfanning:
@harleenkohli


src/Sarif/Autogenerated/ReportingDescriptor.cs
```diff
@@ -229,6 +246,42 @@ private void Init(string id, string name, IEnumerable<string> deprecatedIds, Mul
                 Help = new Message(help);
             }
 
+            if (taxonomyReferences != null)
```
harleenkohli: >if (taxonomyReferences != null) [](start = 12, length = 31)

i merged from master and post build, this file popped up. I think in the merge madness yesterday we missed checking in this file along with the taxonomies PR.

PR #1341, opened 2019-03-08
Default rewriting visitor

michaelcfanning:


src/Sarif/Autogenerated/ReportingDescriptor.cs
```diff
@@ -184,7 +175,7 @@ public ReportingDescriptor(ReportingDescriptor other)
                 throw new ArgumentNullException(nameof(other));
             }
 
-            Init(other.Id, other.Guid, other.Name, other.DeprecatedIds, other.ShortDescription, other.FullDescription, other.MessageStrings, other.DefaultConfiguration, other.HelpUri, other.Help, other.TaxonomyReferences, other.OptionalTaxonomyReferences, other.Properties);
```
harleenkohli: >other.Guid, other.Name, [](start = 27, length = 23)

is this intentional? this looks like a bad merge

harleenkohli: ah, we should just rebuild so that these autogen files are all in sync with actual schema

---
In reply to: [263966681](https://github.com/Microsoft/sarif-sdk/pull/1341#discussion_r263966681) [](ancestors = 263966681)

PR #1346, opened 2019-03-10
Defer Directory.Exists check for performance reasons when determining…

michaelcfanning:
… file mime type. @ScottLouvau, FYI

### DIFF
--- a/src/Sarif/Writers/MimeType.cs
+++ b/src/Sarif/Writers/MimeType.cs
@@ -20,11 +20,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 throw new ArgumentNullException(nameof(path));
             }
 
-            if (System.IO.Directory.Exists(path))
-            {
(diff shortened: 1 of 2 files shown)


src/Sarif/Writers/MimeType.cs
```diff
@@ -39,6 +34,12 @@ public static string DetermineFromFileExtension(string path)
                     }
                 }
             }
+
+            if (System.IO.Directory.Exists(path))
```
ScottLouvau: Great! This should help quite a bit by itself.

Did you also defer the check in FileData.Create when we didn't request any dataToInsert? That was the other half of the performance issue.

ScottLouvau: Cool, that works. =)

michaelcfanning:
@ScottLouvau did you have a chance to look at this?

ScottLouvau:
Looks good; I reviewed.

PR #1349, opened 2019-03-15
general purpose method for parsing sarif nodes

harleenkohli:
per title. please see comments

### DIFF
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -23,6 +23,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
         private const string SchemaPropertyPattern = @"""\$schema""\s*:\s*""[^""]+""";
         private static readonly Regex s_SchemaRegex = new Regex(SchemaPropertyPattern, RegexOptions.Compiled);
 
+        private delegate void ActionOnJObject(JObject jObject);
+        private const string arrayIndicatorSymbol = "[]";
+        private const char nodeDelimiterSymbol = '.';
+
         public static SarifLog UpdateToCurrentVersion(
             string prereleaseSarifLog, 
             Formatting formatting, 
@@ -455,7 +459,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 "conversion.tool.extensions[].ruleDescriptors[]"
             };
 
-            ApplyActionOnNodeIfPathExists(reportingDescriptorPathsToUpdate, run, UpdateReportingDescriptorPropertyTypes);
+            PerformActionOnLeafNodeIfExists(reportingDescriptorPathsToUpdate, run, UpdateReportingDescriptorPropertyTypes);
 
         }
 
@@ -1719,84 +1723,60 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
             return modifiedNotification;
         }
 
-        private delegate void ActionOnJObject(JObject jObject);
-
-        private static void ApplyActionOnNodeIfPathExists(string[] possibleNodePaths, JObject rootNode, ActionOnJObject action)
+        private static void PerformActionOnLeafNodeIfExists(string[] possiblePathsToLeafNode, JObject rootNode, ActionOnJObject action)
         {
-            var nodesToUpdate = FindAllNodes(possibleNodePaths, rootNode);            
-
-            foreach (JObject node in nodesToUpdate)
+            foreach(string nodePath in possiblePathsToLeafNode)
             {
-                action(node);
+                PerformActionOnLeafNodeIfExists(nodePath, rootNode, action);
             }
-
         }
 
-        private static List<JObject> FindAllNodes(string[] fullPaths, JObject rootlog)
+        private static void PerformActionOnLeafNodeIfExists(string possiblePathToLeafNode, JObject rootNode, ActionOnJObject action)
         {
-            var nodes = new List<JObject>();
-
-            foreach (string path in fullPaths)
+            if (possiblePathToLeafNode == null)
             {
-                nodes.AddRange(GetAllNodesAtPath(path, rootlog));
+                action(rootNode);
+                return;
             }
 
-            return nodes;
-        }
+            (string currentNodeName, string childPath) = SplitCurrentNodeNameAndRemainingLeafNodePath(possiblePathToLeafNode);
 
-        private static List<JObject> GetAllNodesAtPath(string fullPath, JObject rootLog)
-        {
-            if (fullPath == null)
+            if (currentNodeName.EndsWith(arrayIndicatorSymbol))
             {
-                return new List<JObject>() { rootLog };
-            }
+                currentNodeName = currentNodeName.TrimEnd(arrayIndicatorSymbol.ToCharArray());
 
-            (string currentNodeName, string childPath) = SplitCurrentNodeNameAndChildPath(fullPath);
-
-            var leafNodes = new List<JObject>();
-
-            if (currentNodeName.EndsWith("[]"))
-            {
-                currentNodeName = currentNodeName.TrimEnd("[]".ToCharArray());
-
-                if (rootLog[currentNodeName] is JArray currentLogs)
+                if (rootNode[currentNodeName] is JArray currentArray)
                 {
-                    foreach (JObject currentLog in currentLogs)
+                    foreach (JObject currentLog in currentArray)
                     {
-                        leafNodes.AddRange(GetAllNodesAtPath(childPath, currentLog));
+                        PerformActionOnLeafNodeIfExists(childPath, currentLog, action);
                     }
                 }
             }
             else
             {
-                if (rootLog[currentNodeName] is JObject currentLog)
+                if (rootNode[currentNodeName] is JObject currentNode)
                 {
-                    leafNodes.AddRange(GetAllNodesAtPath(childPath, currentLog));
+                    PerformActionOnLeafNodeIfExists(childPath, currentNode, action);
                 }
             }
-
-            return leafNodes;
-
         }
 
-        private static (string, string) SplitCurrentNodeNameAndChildPath(string fullPath)
+        private static (string, string) SplitCurrentNodeNameAndRemainingLeafNodePath(string fullPath)
         {
-            if (string.IsNullOrWhiteSpace(fullPath))
+            char[] delimiter = { nodeDelimiterSymbol };
+
+            string[] splitItems = fullPath.Split(separator: delimiter, count: 2);
+
+            string currentNodeName = splitItems[0];
+            string remainingLeafNodePath = null;
+
+            if (splitItems.Length == 2)
             {
-                throw new ArgumentNullException(nameof(fullPath));
+                remainingLeafNodePath = splitItems[1];
             }
 
-            char[] delimiter = { '.' };
-
-            string[] splittedItems = fullPath.Split(delimiter, 2);
-
-            if(splittedItems.Length == 1)
-            {
-                return (splittedItems[0], null);
-            }
-
-            return (splittedItems[0], splittedItems[1]);
+            return (currentNodeName, remainingLeafNodePath);
         }
-
     }
 }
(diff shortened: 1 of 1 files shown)


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -409,90 +409,58 @@ private static void RemoveToolLanguage(JObject run)
 
         private static void UpdateAllReportingDescriptorPropertyTypes(JObject run)
         {
-            // Access and modify run.tool
-            JObject tool = (JObject)run["tool"];
-            UpdateToolReportingDescriptorPropertyTypes(tool);
-
-            // Access and modify run.conversion.tool
-            if (run["conversion"] is JObject conversion)
+            string[] reportingDescriptorPathsToUpdate =
```
harleenkohli: >reportingDescriptorPathsToUpdate [](start = 21, length = 32)

note: i picked up only a simple method for now and converted it to demonstrate usage (and get tests to do their job).

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -409,90 +409,58 @@ private static void RemoveToolLanguage(JObject run)
...
-            if (toolComponent["ruleDescriptors"] is JArray ruleDescriptors)
+            if (reportingDescriptor["shortDescription"] is JObject message2)
             {
-                UpdateReportingDescriptorPropertyTypes(ruleDescriptors);
-            }
-        }
-
-        private static void UpdateReportingDescriptorPropertyTypes(JArray reportingDescriptors)
-        {
-            foreach (JObject reportingDescriptor in reportingDescriptors)
```
harleenkohli: >foreach [](start = 12, length = 7)

the parser method will return all objects so it will take care of arrays as well as jobjects so this iteration in not needed.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+
+            foreach (string path in fullPaths)
+            {
+                nodes.AddRange(GetAllNodesAtPath(path, rootlog));
+            }
+
+            return nodes;
+        }
+
+        private static List<JObject> GetAllNodesAtPath(string fullPath, JObject rootLog)
```
harleenkohli: >GetAllNodesAtPath [](start = 37, length = 17)

one drawback - this algorithm will not handle recursive object references. (e.g. innerExceptions > exception) Perhaps we can introduce a special symbol for recursive references so tht we can keep looping recursively till all objects are found? i am thinking about it while this version is being reviewed.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+            var nodesToUpdate = FindAllNodes(possibleNodePaths, rootNode);            
+
+            foreach (JObject node in nodesToUpdate)
+            {
+                action(node);
+            }
+
+        }
+
+        private static List<JObject> FindAllNodes(string[] fullPaths, JObject rootlog)
```
harleenkohli: >FindAllNodes [](start = 37, length = 12)

another thought i had was if i should move these up in a helper class - decisions like this make me uncomfortable so i am eager to hear thoughts of more experienced folks here :)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -409,90 +409,58 @@ private static void RemoveToolLanguage(JObject run)
...
+            };
 
-        private static void UpdateToolReportingDescriptorPropertyTypes(JObject tool)
-        {
-            // Access and modify tool.driver
-            if (tool["driver"] is JObject driver)
-            {
-                UpdateToolComponentReportingDescriptorPropertyTypes(driver);
-            }
+            ApplyActionOnNodeIfPathExists(reportingDescriptorPathsToUpdate, run, UpdateReportingDescriptorPropertyTypes);
```
harleenkohli: >ApplyActionOnNodeIfPathExists [](start = 12, length = 29)

am i over optimizing here? unsure. Maybe a simpler approach would be to have the actual node modification logic in this method itself. at top level caller level (updateTC..) we will have only one single method call which does the job. i like the thought that any new change of this sort would introduce only one new method, not two.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+        private static void ApplyActionOnNodeIfPathExists(string[] possibleNodePaths, JObject rootNode, ActionOnJObject action)
+        {
+            var nodesToUpdate = FindAllNodes(possibleNodePaths, rootNode);            
+
+            foreach (JObject node in nodesToUpdate)
+            {
+                action(node);
+            }
+
+        }
```
michaelcfanning: >  } [](start = 6, length = 3)

NIT: remove unnecessary vertical whitespace

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: Write a test that thoroughly breaks our current approach due to this limitation, then mark it [Skip] and open a tracking to address the problem.

---
In reply to: [265813535](https://github.com/Microsoft/sarif-sdk/pull/1349#discussion_r265813535) [](ancestors = 265813535)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+        private static (string, string) SplitCurrentNodeNameAndChildPath(string fullPath)
+        {
+            if (string.IsNullOrWhiteSpace(fullPath))
+            {
+                throw new ArgumentNullException(nameof(fullPath));
+            }
+
+            char[] delimiter = { '.' };
+
+            string[] splittedItems = fullPath.Split(delimiter, 2);
```
michaelcfanning: >splittedItems [](start = 21, length = 13)

s/be 'split' not 'splitted'

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+                {
+                    leafNodes.AddRange(GetAllNodesAtPath(childPath, currentLog));
+                }
+            }
+
+            return leafNodes;
+
+        }
+
+        private static (string, string) SplitCurrentNodeNameAndChildPath(string fullPath)
```
michaelcfanning: >string, string [](start = 24, length = 14)

please see if this feature provides for naming

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+            char[] delimiter = { '.' };
+
+            string[] splittedItems = fullPath.Split(delimiter, 2);
+
+            if(splittedItems.Length == 1)
+            {
+                return (splittedItems[0], null);
+            }
+
+            return (splittedItems[0], splittedItems[1]);
```
michaelcfanning: >splittedItems[0] [](start = 20, length = 16)

need better naming for readability

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+            if (string.IsNullOrWhiteSpace(fullPath))
+            {
+                throw new ArgumentNullException(nameof(fullPath));
+            }
+
+            char[] delimiter = { '.' };
+
+            string[] splittedItems = fullPath.Split(delimiter, 2);
+
+            if(splittedItems.Length == 1)
```
michaelcfanning: >if [](start = 12, length = 2)

introduce a space after 'if' NIT

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+            return leafNodes;
+
+        }
+
+        private static (string, string) SplitCurrentNodeNameAndChildPath(string fullPath)
+        {
+            if (string.IsNullOrWhiteSpace(fullPath))
+            {
+                throw new ArgumentNullException(nameof(fullPath));
+            }
```
michaelcfanning: suggest removing this and verifying that all callers are correct

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+                return new List<JObject>() { rootLog };
+            }
+
+            (string currentNodeName, string childPath) = SplitCurrentNodeNameAndChildPath(fullPath);
+
+            var leafNodes = new List<JObject>();
+
+            if (currentNodeName.EndsWith("[]"))
+            {
+                currentNodeName = currentNodeName.TrimEnd("[]".ToCharArray());
```
michaelcfanning: >"[]" [](start = 58, length = 4)

add a const string at the beginning of the method for this literal.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+
+        private delegate void ActionOnJObject(JObject jObject);
+
+        private static void ApplyActionOnNodeIfPathExists(string[] possibleNodePaths, JObject rootNode, ActionOnJObject action)
+        {
+            var nodesToUpdate = FindAllNodes(possibleNodePaths, rootNode);            
+
+            foreach (JObject node in nodesToUpdate)
+            {
+                action(node);
```
michaelcfanning: >action(node); [](start = 16, length = 13)

I'd suggest applying the action on each possible path. This approach maximizes memory consumption.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: >2 [](start = 63, length = 1)

use a named argument here

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -1716,5 +1684,85 @@ private static bool ConvertExceptionMessageFromStringToMessageObject(JObject exc
...
+
+            var leafNodes = new List<JObject>();
+
+            if (currentNodeName.EndsWith("[]"))
+            {
+                currentNodeName = currentNodeName.TrimEnd("[]".ToCharArray());
+
+                if (rootLog[currentNodeName] is JArray currentLogs)
+                {
+                    foreach (JObject currentLog in currentLogs)
```
michaelcfanning: >currentLogs [](start = 51, length = 11)

currentLogs is not the correct name, yes? maybe currentArray

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: great idea! let's wait for a second use.

---
In reply to: [265813845](https://github.com/Microsoft/sarif-sdk/pull/1349#discussion_r265813845) [](ancestors = 265813845)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
harleenkohli: with named params, i think its clearer, having local vars defined with the same name seems overkill (i tried it but hated it). This version looks readable to me :)

---
In reply to: [266093407](https://github.com/Microsoft/sarif-sdk/pull/1349#discussion_r266093407) [](ancestors = 266093407)

PR #1350, opened 2019-03-15
[abandon this, see 1351 instead] 'reportingDesriptor.help' should be a multiformatMessageString, not  message

michaelcfanning:
The reporting descriptor help is not a parameterized string, so it should have been converted to a multiformatMessageString (like short and full description) when we created that type.

@harleenkohli 
@lgolding I added this to the tracking list of bugs we've encountered (i.e., changes that were in the spirit of previous TC decisions but not correctly or comprehensively implemented).

### DIFF
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -141,6 +141,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 {
                     // https://github.com/oasis-tcs/sarif-spec/issues/337
                     ConvertToolToDriverInExternalPropertyFiles(run);
+
+                    UpdateAllReportingDescriptorPropertyTypes(run, UpdateReportingDescriptorHelp);
                 }
             }
             return true;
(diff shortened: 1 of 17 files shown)


michaelcfanning:
>           "$ref": "#/definitions/multiformatMessageString"

We could consider adding a test file and a transformer.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:1424 in ad2f13e. [](commit_id = ad2f13ee47e2eebee96248ca4cbd3746efd4f3f5, deletion_comment = False)

harleenkohli:
>           "$ref": "#/definitions/multiformatMessageString"

i am picking this pr up and adding the tansformer now.

---
In reply to: [473361214](https://github.com/Microsoft/sarif-sdk/pull/1350#issuecomment-473361214) [](ancestors = 473361214)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:1424 in ad2f13e. [](commit_id = ad2f13ee47e2eebee96248ca4cbd3746efd4f3f5, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -141,6 +141,8 @@ private static bool ApplyChangesFromTC33(JObject sarifLog)
                 {
                     // https://github.com/oasis-tcs/sarif-spec/issues/337
                     ConvertToolToDriverInExternalPropertyFiles(run);
+
+                    UpdateAllReportingDescriptorPropertyTypes(run, UpdateReportingDescriptorHelp);
```
harleenkohli: >run, UpdateReportingDescriptorHelp [](start = 62, length = 34)

Mike, wat do u think about this pattern? i didnt want to duplicate the code which creates the list of paths. I had to rename some stuff, see if it makes sense, thanks :)

harleenkohli:
this branch should be abandoned, we have pushed some undesirable changes and a clean revert didnt happen. I have a rasied a new PR with only desired changes.

PR #1352, opened 2019-03-18
Update Microsoft.Json assemblies (jschema) to 0.63.0. Re-enable previ…

michaelcfanning:
…ously skipped tests.

@lgolding, updated SDK to latest Microsoft.Json bits. Have a moment to look?

### DIFF
--- a/src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError.sarif
+++ b/src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError.sarif
@@ -1,10 +1,10 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2019-01-09",
-  "version": "2.0.0-csd.2.beta.2019-01-09",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2019-02-20",
+  "version": "2.0.0-csd.2.beta.2019-02-20",
   "runs": [
     {
       "tool": {
-        "name": "CodeScanner"
+        "driver": { "name": "CodeScanner" }
       },
       "results": [
         {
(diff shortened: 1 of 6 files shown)


src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError.sarif
```diff
@@ -1,10 +1,10 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2019-01-09",
-  "version": "2.0.0-csd.2.beta.2019-01-09",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2019-02-20",
+  "version": "2.0.0-csd.2.beta.2019-02-20",
   "runs": [
     {
       "tool": {
-        "name": "CodeScanner"
+        "driver": { "name": "CodeScanner" }
```
michaelcfanning: >driver [](start = 9, length = 6)

These required changes are what prompted us to disable jschema previously. as you recall, i'm sure, the validator 'works' even w/breaking SARIF changes but only as long as it isn't required to return any errors. After updating, I was able to see and fix errors from the test files.

src/Sarif.Multitool.FunctionalTests/TestData/DeserializationError_Expected.sarif
```diff
@@ -83,7 +83,7 @@
               "message": {
                 "text": "Analyzing 'DeserializationError.sarif'..."
               },
-              "level": "note",
+              "level": "none",
```
michaelcfanning: >none [](start = 24, length = 4)

These are trace messages and in no way a problem report, so "none" is the correct level.

PR #1355, opened 2019-03-20
340 inline logical location changes

harleenkohli:


harleenkohli:
>   "version": "2.0.0-csd.2.beta.2019-02-20",

we should regen files with upcoming version? i need to learn how to do this. i hace added that in the issues list to keep track

---
Refers to: src/Sarif.FunctionalTests/v2/ConverterTestData/PREfast/Expected-049.xml.sarif:3 in 283708a. [](commit_id = 283708a89254a18c4de38fba9f7035ebb88a0b87, deletion_comment = False)

PR #1360, opened 2019-03-21
341 instance guid to guid

harleenkohli:


src/Sarif/NotYetAutoGenerated/Run.cs
```diff
@@ -102,21 +102,21 @@ public SarifNodeKind SarifNodeKind
...
-        [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
-        public RunAutomationDetails Id { get; set; }
+        [DataMember(Name = "automationDetails", IsRequired = false, EmitDefaultValue = false)]
+        public RunAutomationDetails AutomationDetails { get; set; }
 
         /// <summary>
         /// Automation details that describe the aggregate of runs to which this run belongs.
         /// </summary>
-        [DataMember(Name = "aggregateIds", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "RunAggregates", IsRequired = false, EmitDefaultValue = false)]
```
harleenkohli: >RunAggregates [](start = 28, length = 13)

the ballot text didnt actually mention that we must update this specific property, however it felt like the right thing to do, considering that we are renaming id as well. I checked with Larry offline and he suggested to use the name "RunAggregates". @Mikefan, please validate if this change is ok and/or requires a different name.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -141,11 +141,106 @@ private static bool ApplyChangesFromTC33(JObject sarifLog)
...
+                }
+
+                run.Remove("aggregateIds");
+                run.Add("runAggregates", aggregateIds);
+            }
+
+            // run.results[].instanceGuid -> run.results[].guid
+            // run.results[].resultProvenance.firstDetectionRunInstanceGuid -> run.results[].resultProvenance.firstDetectionRunGuid
+            // run.results[].resultProvenance.lastDetectionRunInstanceGuid -> run.results[].resultProvenance.lastDetectionRunGuid
+
```
harleenkohli: oops, this needs to be removed, will fix

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2112,24 +2112,24 @@
...
-        "aggregateIds": {
-          "description": "Automation details that describe the aggregate of runs to which this run belongs.",
-          "type": "array",
-          "minItems": 0,
-          "uniqueItems": true,
-          "default": [],
-          "items": {
-            "$ref": "#/definitions/runAutomationDetails"
-          }
+        "RunAggregates": {
```
lgolding: >RunAggregates [](start = 9, length = 13)

Yes, sorry, Harleen and I discussed this. I felt that if "id" was wrong then so was "aggregateIds" and I suggested "runAggregates" meaning the set of aggregates to which this run belongs. Having said that, I never made that change in the "guid" change draft. I'll go with whatever you want here, up to and including leaving "aggregateIds" alone.

src/Sarif/NotYetAutoGenerated/Run.cs
michaelcfanning: Sounds good, I updated #341 with this.

---
In reply to: [267608983](https://github.com/Microsoft/sarif-sdk/pull/1360#discussion_r267608983) [](ancestors = 267608983)

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: It's fine. I updated #341 with this suggestion, let's just make sure the text is accurate when it comes to it. :)

---
In reply to: [268453384](https://github.com/Microsoft/sarif-sdk/pull/1360#discussion_r268453384) [](ancestors = 268453384)

PR #1363, opened 2019-03-21
326 update graphs objects

harleenkohli:
impementation for https://github.com/oasis-tcs/sarif-spec/issues/326


src/Sarif.Multitool.FunctionalTests/TestData/MessagesShouldEndWithPeriod/Invalid.sarif
```diff
@@ -110,48 +110,6 @@
...
-              "edgeTraversals": [
-                {
-                  "edgeId": "e",
-                  "message": {
-                    "text": "edgeTraversal message without period",
-                    "richText": "_Rich_ edgeTraversal message without period"
-                  }
-                }
-              ]
-            }
```
harleenkohli: ok, this is a controversial choice i made! in order for the associated test cases to pass afaik i had two options: 1. write a prerelease transformer, 2. remove these from our samples since we do not actually have a customer who uses this feature yet. 

If theres a third option which can help us preserve these cases, please let me know, i would be happy to implement :)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -480,170 +480,173 @@
...
+          "$ref": "#/definitions/conversion"
+        },
+
+        "graphs": {
+          "description": "An array of graph objects that will merged with a separate run.",
+          "type": "array",
+          "minItems": 0,
+          "default": [],
+          "uniqueItems": true,
+          "items": {
```
harleenkohli: according to ballot text, we need to ensure this property is "optional". I wasnt sure how to actually enforce that, isnt everything optional unless we add the "required" property?

If yes, then follow up question - there are items in the ballot text which do not mention the word optional - does that mean they are required? 

text for quick reference:

In the run object:

Change the type of the graphs property from object with graph-valued properties to graph[], unique, minItems: 0, default: empty array.

In the externalProperties object:

Rename graph property to graphs and change its type from object with graph-valued properties to graph[], optional, unique, minItems: 0, default: empty array.

src/Sarif.Multitool.FunctionalTests/TestData/MessagesShouldEndWithPeriod/Invalid.sarif
```diff
@@ -110,48 +110,6 @@
               ]
             }
           ],
-          "graphs": {
```
michaelcfanning: >graphs [](start = 11, length = 6)

You have removed all this test content because we decided not to author a transformer. Did you consider just manually updating all this content to conform to the new standard instead? shouldn't be hard.

src/Sarif.Multitool.FunctionalTests/TestData/MessagesShouldEndWithPeriod/Invalid.sarif
michaelcfanning: the third option would be to preserve the test content but update it to the new standard. :)

PR #1364, opened 2019-03-21
324 updating ReportingDescriptorReference object

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1719,12 +1719,24 @@
       "properties": {
 
         "id": {
-          "description": "A notification identifier.",
+          "description": "The Id of the descriptor.",
```
michaelcfanning: >Id [](start = 30, length = 2)

do not capitalize

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1719,12 +1719,24 @@
...
+          "default": -1,
+          "minimum": -1
+        },
+
+        "toolComponentGuid": {
+          "description": "A guid that uniquely identifies the tool component associated with the descriptor.",
+          "type": "string"
+        },
+
+        "descriptorGuid": {
```
michaelcfanning: >descriptorGuid [](start = 9, length = 14)

call this guid. also we need to check in with @lgolding, who i believe is proposing that we have a toolComponentReference object in this type.

PR #1367, opened 2019-03-24
346 reporting descriptor deprecated names and deprecated guids

michaelcfanning:
@lgolding @harleenkohli


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1542,13 +1542,24 @@
...
+
         "guid": {
           "description": "A unique identifer for the reporting descriptor in the form of a GUID.",
           "type": "string"
         },
 
-        "deprecatedIds": {
-          "description": "An array of stable, opaque identifiers by which this report was known in some previous version of the analysis tool.",
+        "deprecatedGuids": {
+          "description": "An array of unique identifies in the form of a GUID by which this report was known in some previous version of the analysis tool.",
```
harleenkohli: >identifies [](start = 45, length = 10)

identifiers?

src/Sarif/Autogenerated/Location.cs
```diff
@@ -98,9 +98,9 @@ public Location()
         /// <param name="properties">
         /// An initialization value for the <see cref="P:Properties" /> property.
         /// </param>
-        public Location(Address address, PhysicalLocation physicalLocation, LogicalLocation logicalLocation, Message message, IEnumerable<Region> annotations, IDictionary<string, SerializedPropertyInfo> properties)
```
harleenkohli: >n(Address address, Phys [](start = 22, length = 23)

this seems like a bad change?

harleenkohli: oops, no this is correct! thanks for catching this!

---
In reply to: [268851294](https://github.com/Microsoft/sarif-sdk/pull/1367#discussion_r268851294) [](ancestors = 268851294)

PR #1370, opened 2019-03-26
Not yet autogen tests and reverting changes in result.cs 

harleenkohli:


src/Sarif/Autogenerated/Location.cs
```diff
@@ -33,12 +33,6 @@ public SarifNodeKind SarifNodeKind
             }
         }
 
-        /// <summary>
-        /// The address of the location.
-        /// </summary>
-        [DataMember(Name = "address", IsRequired = false, EmitDefaultValue = false)]
-        public Address Address { get; set; }
```
harleenkohli: this is just a minor change in an auto-gen file we missed earlier.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -113,6 +114,17 @@ private void ValidateDefaultDocument(IDictionary<Type, DefaultObjectPopulatingVi
             SarifLog clonedLog = sarifLog.DeepClone();
             clonedLog.ValueEquals(sarifLog).Should().BeTrue();
 
+            JObject jObjectSarifLog = JObject.Parse(toValidate);
+
+            // The following validations ensure NotYetAutoGenerated fields work as expected.
```
harleenkohli: >// The following validations ensure NotYetAutoGenerated fields work as expected. [](start = 12, length = 80)

@mikefan i tried two things here for "Run.ColumnKind" validation:

1. use local SarifLog object for the validation:
        newSarifLog.Runs[0].ColumnKind.Should().Be(ColumnKind.Utf16CodeUnits);
 however, this fails because our logic populates the default value, which according to our schema is ColumnKind.UnicodeCodePoints.

2. i added code to initialize a new run object (like you suggested in the first place, but it took me while to understand why #1 wont work :)). However i changed my mind about it and reverted it because of two reasons:
      a. this code is run for a plethora of files, it felt weird adding that test here which actually has no relation to the current file.
      b. removal of not-yet-autogen/run file will end up failing many converter test cases so this will be hard to miss.

Perhaps there s a simple UTC we should add instead (like you suggested at one point) and now i agree :).

harleenkohli: >// The following validations ensure NotYetAutoGenerated fields work as expected. [](start = 12, length = 80)

@mikefan i tried two things here for "Run.ColumnKind" validation:

1. use local SarifLog object for the validation:
        newSarifLog.Runs[0].ColumnKind.Should().Be(ColumnKind.Utf16CodeUnits);
 however, this fails because our logic populates the default value, which according to our schema is ColumnKind.UnicodeCodePoints.

2. i added code to initialize a new run object (like you suggested in the first place, but it took me while to understand why #1 wont work :)).
Something like this:

            // see .src/sarif/NotYetAutoGenerated/Run.cs (columnKind property)
            var newSarifLog = new SarifLog
            {
                Runs = new Run[]
                {
                    new Run()
                }
            };
            newSarifLog.Runs[0].ColumnKind.Should().Be(ColumnKind.Utf16CodeUnits);

However i changed my mind about it and reverted it because of these reasons:
      a. It felt weird adding that test here since default document is indeed valid with UnicodeCodePoints value.
      b. Having to add a new sarifLog object just to test this here will introduce this disconnected piece of validation in three different TCs, which will all fail (ValidateAllTheThings etc.)

Perhaps there s a simple UTC we should add instead (like you suggested at one point) and now i agree :). 

Thoughts?

Note: removal of not-yet-autogen/run file will end up failing many converter test cases so this will be hard to miss.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -113,6 +114,17 @@ private void ValidateDefaultDocument(IDictionary<Type, DefaultObjectPopulatingVi
...
+
+            // see .src/sarif/NotYetAutoGenerated/ReportingConfiguration.cs (Rank property)
+            // see .src/sarif/NotYetAutoGenerated/Result.cs (Rank property)
+            bool isRankSerialized = RecursivePropertySearch(jObjectSarifLog, "rank");
+            isRankSerialized.Should().BeFalse();
+
+            // TODO: Add check for .src/sarif/NotYetAutoGenerated/Result.cs (resultKind and failureLevel properties)
+
```
harleenkohli: this test is needed, our current tests wont catch a change, lets me chat with you about this offline.

src/Sarif.ValidationTests/SarifValidatorTests.cs
```diff
@@ -195,5 +207,46 @@ private string FailureReason(Result[] errors)
 
             return sb.ToString();
         }
+
+        private static bool RecursivePropertySearch(JObject parentObject, string searchKey)
```
harleenkohli: >string searchKey [](start = 74, length = 16)

i think i should change this to accept string[] so that we can test for other fields like rank as well - i.e. the ones which should not be serialized.

src/Sarif.ValidationTests/SarifValidatorTests.cs
michaelcfanning: i agree

---
In reply to: [269355722](https://github.com/Microsoft/sarif-sdk/pull/1370#discussion_r269355722) [](ancestors = 269355722)

PR #1371, opened 2019-03-27
proposed changes for item #314 - define result taxonomies

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -147,7 +147,11 @@
               "addedFile",
               "deletedFile",
               "renamedFile",
-              "uncontrolledFile"
+              "uncontrolledFile",
+              "taxon",
+              "extension",
+              "translation",
+              "reportingConfigurationOverride"
```
michaelcfanning: >reportingConfigurationOverride [](start = 15, length = 30)

call this 'policy', add 'driver' and remove 'toolComponent'

PR #1375, opened 2019-03-27
PR 2 of 2 for 338 Localization and Policy updates

harleenkohli:
Phew!


harleenkohli:
>         },

Per spec text, these two must default to the semanticVersionproperty of the component. Is there a way to enforce this?

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2741 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

harleenkohli:
>         "language": {

per spec text: required for translations; otherwise optional. I am guessing there is no way to enforce this currently but please confirm. Also, should we update text?

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2710 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

harleenkohli:
>         },

also, should we update description text to indicate this?

---
In reply to: [477389626](https://github.com/Microsoft/sarif-sdk/pull/1375#issuecomment-477389626) [](ancestors = 477389626)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2741 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

michaelcfanning:
>         "language": {

there should be a default value here, of "en-us". we will need a validator rule to enforce this, file a work item.

---
In reply to: [477389836](https://github.com/Microsoft/sarif-sdk/pull/1375#issuecomment-477389836) [](ancestors = 477389836)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2710 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

michaelcfanning:
>         },

default means that if it is not present, we pick up the value from the component version. one illegal thing would be if the minimum required version exceeded the tool component version. :) good rule for a validator.

---
In reply to: [477389900](https://github.com/Microsoft/sarif-sdk/pull/1375#issuecomment-477389900) [](ancestors = 477389900,477389626)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2741 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

harleenkohli:
>         },

oh! i see :) thanks!

---
In reply to: [477392400](https://github.com/Microsoft/sarif-sdk/pull/1375#issuecomment-477392400) [](ancestors = 477392400,477389900,477389626)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2741 in a73d6cc. [](commit_id = a73d6cc5e25463df1da2538be6278d2e536f90e2, deletion_comment = False)

PR #1378, opened 2019-03-28
314 - define result taxonomies

harleenkohli:


harleenkohli:
This PR turned out to be much simpler because most of the additions required already made it to master because of #338.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2751,6 +2752,28 @@
...
+          "type": "array",
+          "minItems": 0,
+          "uniqueItems": true,
+          "default": [],
+          "items": {
+            "$ref": "#/definitions/toolComponentReference"
+          }
+        },
+
+        "reportingConfigurationOverrides": {
```
michaelcfanning: >reportingConfigurationOverrides [](start = 9, length = 31)

I think we should remove this. reportingConfigurationOverrides are now provided by 'policy' objects that simply recapitulate the core toolComponent and make changes to its reporting descriptor defaultConfiguration values.

src/Sarif/Autogenerated/ArtifactRoles.cs
```diff
@@ -30,6 +30,7 @@ public enum ArtifactRoles
         Extension = 8192,
         Translation = 16384,
         Taxonomy = 32768,
-        Policy = 65536
+        Policy = 65536,
+        ReportingConfigurationOverride = 131072
```
michaelcfanning: Let's use 'referencedOnCommandLine' instead. Let's ask Larry if we can drop 'responseFile'.

PR #1379, opened 2019-03-28
removing result.ruleExtensionIndex

harleenkohli:


michaelcfanning:
>         "rule": {

please go move this to be close to ruleId and ruleIndex as they are related!!

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2027 in 18bfed3. [](commit_id = 18bfed38aa9ca50a599a7e0e32680635d8cfec2d, deletion_comment = False)

PR #336, opened 2016-06-06
Resolved CA 2213: Closed a disposable field

jericahuang:
Closed the disposable field _jsonTextWriter

@lgolding @michaelcfanning @Rolstenhouse @kschecht

PR #302, opened 2016-05-23
VSIX: Add a collection to visibility converter to hide tabs which do not contain data

rtaket:
Add a readme to explain the debugging settings for the VSIX project.

@michaelcfanning

PR #1385, opened 2019-03-29
end game sweep 3

harleenkohli:
issues addressed:

- add invocation.ruleConfigurationOverrides & invocation.notificationConfigurationOverrides #324
- addresses should be externalizable #302
- add artifact.roles.memoryContents #302
- add toolComponent.releaseDateUtc, toolComponent.productSuite #179 & #336

### DIFF
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -152,7 +152,8 @@
               "translation",
               "taxonomy",
               "policy",
-              "referencedOnCommandLine"
+              "referencedOnCommandLine",
+              "memoryContents"
             ]
           }
         },
@@ -661,6 +662,17 @@
           }
         },
 
+        "addresses": {
+          "description": "Addresses that will be merged with a separate run.",
+          "type": "array",
+          "minItems": 0,
+          "uniqueItems": false,
+          "default": [],
+          "items": {
+            "$ref": "#/definitions/address"
+          }
+        },
+
         "properties": {
           "description": "Key/value pairs that provide additional information about the external properties.",
           "$ref": "#/definitions/propertyBag"
@@ -711,7 +723,8 @@
         "graphs": {
           "description": "An array of external property files containing a run.graphs object to be merged with the root log file.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
+          "default": [],
           "uniqueItems": true,
           "items": {
             "$ref": "#/definitions/externalPropertyFileReference"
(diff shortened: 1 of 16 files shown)


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -711,7 +723,8 @@
         "graphs": {
           "description": "An array of external property files containing a run.graphs object to be merged with the root log file.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
+          "default": [],
```
michaelcfanning: this is wrong. it doesn't make sense to allow for a non-null and empty graphs collection. this is just empty data.

minItems:1 is right.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1015,8 +1028,19 @@
           "type": "integer"
         },
 
-        "configurationOverrides": {
-          "description": "An array of configurationOverride objects that describe runtime reporting behavior.",
+        "ruleConfigurationOverrides": {
+          "description": "An array of configurationOverride objects that describe rules related runtime overrides.",
+          "type": "array",
+          "minItems": 0,
+          "default": [],
```
michaelcfanning: this is wrong, minItems s/be 1 in both this case and notificationConfigurationOverrides.

Ask yourself: what value does an empty collection provide? this is very rare. mostly, it shouldn't be present at all, or there should be at least one item in the array.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1015,8 +1028,19 @@
...
+          "items": {
+            "$ref": "#/definitions/configurationOverride"
+          }
+        },
+
+        "notificationConfigurationOverrides": {
+          "description": "An array of configurationOverride objects that describe notifications related runtime overrides.",
           "type": "array",
           "minItems": 0,
           "default": [],
```
michaelcfanning: minItems s/be 1, no default

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1669,7 +1693,8 @@
         "taxa": {
           "description": "An array of references used to locate a set of taxonomy reporting descriptors that are always applicable to a result.",
           "type": "array",
-          "minItems": 1,
+          "default": [],
+          "minItems": 0,
```
michaelcfanning: nope. I will now stop providing this comment. minItems == 1, no default.

michaelcfanning: please fix all minItems then merge.

---
In reply to: [270602535](https://github.com/Microsoft/sarif-sdk/pull/1385#discussion_r270602535) [](ancestors = 270602535)

PR #1394, opened 2019-04-02
Streamline Fortify FPR converter + fix #1293

### LINKED ISSUE #1293: Update Fortify converter to emit well-known thread flow location kinds
e.g., we emit InCall today. SARIF has the following well known kinds, is there an equivalent to in call?

acquire, release, enter, exit, call, return, branch, implicit, false, true, caution, danger, unknown, unreachable, taint, function, handler, lock, memory, resource, and scope

EasyRhinoMSFT:


src/Sarif.Converters.UnitTests/TestData/FortifyFprConverter/Inputs/OneResultBasic.fvdl
```diff
@@ -116,5 +117,17 @@ SqlDataReader objReader = query.ExecuteReader();</Text>
     <Username>johndoe</Username>
     <Platform>Linux</Platform>
   </MachineInfo>
+  <RuleInfo>
+    <Rule id="31D4607A-A3FF-447C-908A-CA2BBE4CE4B7">
+      <MetaInfo>
+        <Group name="package">C Core</Group>
+        <Group name="altcategorySTIG4.8">APSC-DV-002560 CAT I</Group>
```
ScottLouvau: Should these reference IDs make it into the Rule.Properties somewhere?
Is one of these IDs a better choice as the RuleID than the GUID? Which one?

EasyRhinoMSFT: I think it should be in rule.guid. I think Michael mentioned this last week.

EasyRhinoMSFT: But then I don't know what to use for the ruleId

ScottLouvau: Maybe we could ask Katerina in a thread.
Also, I think I was unclear: I think the secondary IDs (APSC-DV-002560 CAT I; CWE ID 22, CWE ID 73, etc) should also be in the rule in SARIF somewhere, so we know all of those alternative rule identities.

EasyRhinoMSFT: Yeah I just noticed I'm not emitting the CWE references. Looking now...

EasyRhinoMSFT: OK fixed. I just sent mail to Katrina so I'll wait for a reply before pushing.

PR #1395, opened 2019-04-02
Get CS converter working

lgolding:
This change fixes a set of bugs that prevent the partially implemented Contrast Security converter from running to completion. Now that these bugs are fixed, we have a working converter to which we can add new rule conversions one at a time.

The bugs are all failures to check whether a rule's `DefaultConfiguration` is non-`null` before dereferencing it to obtain its `Level` property. If `DefaultConfiguration` is missing, `Level` should default to `FailureLevel.Warning`. So wrap up the null-checking and defaulting logic in a new helper method `FailureLevel GetRuleFailureLevel(string ruleId)`, and call it everywhere (including `ConstructNotImplementedRuleResult`).

Also:
- Fix some nits I ran across while studying the code.

### DIFF
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -255,7 +255,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             var result = new Result
             {
-                Level = _rules[ContrastSecurityRuleIds.AntiCachingControlsMissing].DefaultConfiguration.Level,
+                Level = GetFailureLevel(ContrastSecurityRuleIds.AntiCachingControlsMissing),
                 RuleId = ContrastSecurityRuleIds.AntiCachingControlsMissing,
                 Locations = locations,
                 Message = new Message
@@ -352,7 +352,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             var result = new Result
             {
-                Level = _rules[ContrastSecurityRuleIds.PagesWithoutAntiClickjackingControls].DefaultConfiguration.Level,
+                Level = GetFailureLevel(ContrastSecurityRuleIds.PagesWithoutAntiClickjackingControls),
                 RuleId = ContrastSecurityRuleIds.PagesWithoutAntiClickjackingControls,
                 Locations = locations,
                 Message = new Message
@@ -389,7 +389,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
             result = new Result
             {
-                Level = _rules[ContrastSecurityRuleIds.CrossSiteScripting].DefaultConfiguration.Level,
+                Level = GetFailureLevel(ContrastSecurityRuleIds.CrossSiteScripting),
                 RuleId = ContrastSecurityRuleIds.CrossSiteScripting,
                 Locations = new List<Location>()
                 {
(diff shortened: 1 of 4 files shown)


src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1120,21 +1126,6 @@ public ContrastLogReader()
 
         public void Read(Context context, Stream input)
         {
-            XmlSchemaSet schemaSet = new XmlSchemaSet();
-            Assembly assembly = typeof(ContrastLogReader).Assembly;
-            var settings = new XmlReaderSettings
-            {
-                DtdProcessing = DtdProcessing.Ignore,
-                XmlResolver = null
-            };
```
lgolding: >}; [](start = 12, length = 2)

These local variables are never used.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1120,21 +1126,6 @@ public ContrastLogReader()
...
-
-            //using (var stream = assembly.GetManifestResourceStream(ContrastLogReader.ContrastSecurityReportSchema))
-            //using (var reader = XmlReader.Create(stream, settings))
-            //{
-            //    XmlSchema schema = XmlSchema.Read(reader, new ValidationEventHandler(ReportError));
-            //    schemaSet.Add(schema);
-            //}
```
lgolding: >} [](start = 14, length = 1)

This was FxCop-specific code (the ContrastSecurity converter was based on the FxCop converter). Not needed here.

lgolding:

lgolding:

harleenkohli:
>             // default : The

possible typo? i know older code, but looks funny :)

---
Refers to: src/Sarif.Converters/ContrastSecurityConverter.cs:696 in 2d9358d. [](commit_id = 2d9358dc0a18a8acf85e150ce8e62405186041e8, deletion_comment = False)

lgolding:
>             // default : The

Hah :-) Now idea what that comment was supposed to be; will remove.

---
In reply to: [479264371](https://github.com/Microsoft/sarif-sdk/pull/1395#issuecomment-479264371) [](ancestors = 479264371)

---
Refers to: src/Sarif.Converters/ContrastSecurityConverter.cs:696 in 2d9358d. [](commit_id = 2d9358dc0a18a8acf85e150ce8e62405186041e8, deletion_comment = False)

lgolding:
>             // default : The

Oh I see what those comments are. They replicate the `rule.messageStrings` property. The `default` is the key into the `messageStrings` dictionary. I'll fix this one up.

---
In reply to: [479265400](https://github.com/Microsoft/sarif-sdk/pull/1395#issuecomment-479265400) [](ancestors = 479265400,479264371)

---
Refers to: src/Sarif.Converters/ContrastSecurityConverter.cs:696 in 2d9358d. [](commit_id = 2d9358dc0a18a8acf85e150ce8e62405186041e8, deletion_comment = False)

lgolding:
>             // default : The

Pushed fix.

---
In reply to: [479265903](https://github.com/Microsoft/sarif-sdk/pull/1395#issuecomment-479265903) [](ancestors = 479265903,479265400,479264371)

---
Refers to: src/Sarif.Converters/ContrastSecurityConverter.cs:696 in 2d9358d. [](commit_id = 2d9358dc0a18a8acf85e150ce8e62405186041e8, deletion_comment = False)

PR #1399, opened 2019-04-03
address.baseAddress & address..offset to int

harleenkohli:


src/Sarif.UnitTests/TestData/SarifVersionOneToCurrentVisitor/ExpectedOutputs/NotificationExceptionWithStack.sarif
```diff
@@ -13,9 +13,6 @@
         {
           "toolExecutionNotifications": [
             {
-              "descriptor": {
```
michaelcfanning: >descriptor [](start = 15, length = 10)

in "notification" definition, move "descriptor" closer to the top.

PR #1405, opened 2019-04-04
Fix Multitool ValidateCommand

EasyRhinoMSFT:


src/Sarif.Multitool/ValidateCommand.cs
```diff
@@ -65,12 +65,12 @@ protected override SarifValidationContext CreateContext(ValidateOptions options,
...
                 // (which can happen, for example, if a property required by the schema is
                 // missing. In that case, again, there's no point in going on.
-                string sarifText = FileSystem.ReadAllText(context.TargetUri.OriginalString);
+                context.InputLogContents = FileSystem.ReadAllText(context.TargetUri.OriginalString);
 
                 PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(
                     context.InputLogContents,
                     formatting: Formatting.None,
-                    out sarifText);
+                    out string sarifText);
```
michaelcfanning: >string [](start = 24, length = 6)

I think you can just pass content.InputLogContents here and get rid of the sarifText var use.

michaelcfanning:
We need an entry in ReleaseHistory.md for this. Note that the fix will be included in v2.0.0-csd.2.beta.2019.04-03.3. You can change the suffix in src\build.props to reflect this if you want. Finally, we need to investigate why the tests in  E:\src\sarif\src\Sarif.Multitool.FunctionalTests\ValidateCommandTests.cs did not protect us. I recently disabled these tests (because we need a new JSchema update) but they were running as of a week ago. How did they succeed? Anyway, let's close the test gap. I'll fix Jschema for you tomorrow.

E:\src\sarif\src\Sarif.Multitool.FunctionalTests\TestData\

EasyRhinoMSFT:
Issue #1406

PR #1409, opened 2019-04-07
CS: populate threadFlow.immutableState from headers.

lgolding:
Also:
- Avoid hard-coding the rule ids repeatedly by using the value that already exists in the context. This has the nice side effect of making all the rule handler signatures the same (they all need to take the context now, whereas before some of them could get away with passing only the properties dictionary).

I'll continue to add to this PR, adjusting the title and description as required, until one of you reviews this.

### DIFF
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -391,7 +391,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             string untrustedData = BuildSourcesString(context.Sources);
             string page = context.RequestUri;
             string caller = context.PropagationEvents[context.PropagationEvents.Count - 1].Stack.Frames[0].Location.LogicalLocation?.FullyQualifiedName;
-            string controlID = context.Properties.ContainsKey(nameof(controlID)) ? context.Properties[nameof(controlID)] : null;
+            string controlId = context.Properties.ContainsKey(nameof(controlId)) ? context.Properties[nameof(controlId)] : null;
 
             const string RuleId = ContrastSecurityRuleIds.CrossSiteScripting;
             var result = new Result
@@ -407,7 +407,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 }
             };
 
-            if (controlID == null)
+            if (controlId == null)
             {
                 result.Message = new Message
                 {
@@ -428,7 +428,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     {                  // A cross-site scripting vulnerability was seen as untrusted data
                         untrustedData, // '{0}' on 
                         page,          // '{1}' was accessed within 
-                        controlID      // '{2}' control and observed going into the HTTP response without validation or encoding.
+                        controlId      // '{2}' control and observed going into the HTTP response without validation or encoding.
                     }
                 };
             }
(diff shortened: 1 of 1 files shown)


src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -101,11 +101,6 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
                 output.WriteLogicalLocations(LogicalLocations);
             }
 
-            //if (_rules?.Any() == true)
-            //{
-            //    output.WriteRules(_rules.Values.ToList());
```
lgolding: >WriteRules [](start = 25, length = 10)

This method no longer exists. The rules are serialized with the tool.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -396,13 +390,12 @@ private Result ConstructCrossSiteScriptingResult(ContrastLogReader.Context conte
             string untrustedData = BuildSourcesString(context.Sources);
             string page = context.RequestUri;
             string caller = context.PropagationEvents[context.PropagationEvents.Count - 1].Stack.Frames[0].Location.LogicalLocation?.FullyQualifiedName;
-            string controlID = context.Properties.ContainsKey(nameof(controlID)) ? context.Properties[nameof(controlID)] : null;
+            string controlId = context.Properties.ContainsKey(nameof(controlId)) ? context.Properties[nameof(controlId)] : null;
```
lgolding: >controlId [](start = 19, length = 9)

Rename for style conformance.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -882,8 +874,8 @@ private IList<CodeFlow> CreateCodeFlows(ContrastLogReader.Context context)
                          {
                              new ThreadFlow
                              {
-                                  Locations = context.PropagationEvents
-                                  // TODO: Populate ImmutableState from the headers.
+                                  Locations = context.PropagationEvents,
+                                  ImmutableState = context.Headers
```
lgolding: >ImmutableState [](start = 34, length = 14)

Michael's favorite feature! Request headers, enabled by `ThreadFlow.ImmutableState`.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -117,12 +112,12 @@ internal Result CreateResult(ContrastLogReader.Context context)
             {
                 case ContrastSecurityRuleIds.AntiCachingControlsMissing:
                 {
-                    return ConstructAntiCachingControlsMissingResult(context.Properties);
+                    return ConstructAntiCachingControlsMissingResult(context);
```
lgolding: >context [](start = 69, length = 7)

All of these changes involve passing in the entire context so we can use the `RuleId` it contains, rather than hard-coding the rule id in each handler.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -256,11 +252,10 @@ private Result ConstructAntiCachingControlsMissingResult(IDictionary<string, str
 
             string pageCount = locations.Count.ToString();
 
-            const string RuleId = ContrastSecurityRuleIds.AntiCachingControlsMissing;
```
lgolding: >RuleId [](start = 25, length = 6)

An example of not having to hard-code the rule id.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -256,11 +252,10 @@ private Result ConstructAntiCachingControlsMissingResult(IDictionary<string, str
...
             return new Result
             {
-                RuleId = RuleId,
-                Level = GetRuleFailureLevel(RuleId),
+                RuleId = context.RuleId,
```
lgolding: >context [](start = 25, length = 7)

... and getting it from the context instead.

PR #1412, opened 2019-04-08
361 states to multiformat

harleenkohli:


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -132,6 +143,54 @@ public static class PrereleaseCompatibilityTransformer
             return transformedSarifLog;
         }
 
+        private static bool ApplyChangesFromTC34(JObject sarifLog)
+        {
+            // TODO: update sarif version
```
michaelcfanning: >TODO: update [](start = 15, length = 12)

remove this comment

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -132,6 +143,54 @@ public static class PrereleaseCompatibilityTransformer
...
+            // https://github.com/oasis-tcs/sarif-spec/issues/361
+            ConvertAllStateStringsToMultiFormatMessageStrings(sarifLog);
+            return true;
+        }
+
+        private static void ConvertAllStateStringsToMultiFormatMessageStrings(JObject run)
+        {
+            string[] statePathsToUpdate =
+            {
+                "inlineExternalProperties[].results[].graphTraversals[].edgeTraversals[].finalState",
```
michaelcfanning: >inlineExternalProperties [](start = 17, length = 24)

replace this with a new helper that accepts a hashset of strings, each of which is a property name like finalState. then make a single pass over the JSON, and call your callback with any property value where the property name is included in the set.

src/Test.UnitTests.Sarif.Driver/Sdk/DriverSdkExtensionsTests.cs
```diff
@@ -31,7 +31,7 @@ public void ConstructUriBaseIdsDictionary_RetrievesAbsoluteFilePath()
             uriBaseIds.Values.First().Uri.LocalPath.Should().Be(path);
         }
 
-        [Fact]
+        [Fact(Skip = "If run concurrently with tests for other libraries, this test fails. To be investigated.")]
```
harleenkohli: > [Fact(Skip = "If run concurrently with tests for other libraries, this test fails. To be investigated.")] [](start = 7, length = 106)

opened issue #213 for tracking this seperately. I tried the .substring trick but it didnt work intermittently.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
harleenkohli: tracking item #214

---
In reply to: [273270880](https://github.com/Microsoft/sarif-sdk/pull/1412#discussion_r273270880) [](ancestors = 273270880)

PR #1414, opened 2019-04-09
CS: Populate request object

lgolding:
We populate the new SARIF `request` object with the contents of the CS `<request>` element.

Also:
- DRY out the core result creation logic into a helper method `CreateResultCore`.
- Fix a bug that prevented `<p>` elements from being processed.


lgolding:
For some reason appveyor is failing with the message "AppVeyor was unable to build non-mergeable pull request", even though testing on my desktop shows that my branch merges cleanly into master. I'm going to close this PR and open another one.

PR #1417, opened 2019-04-09
Update schema to reflect '-beta.0' condition.

michaelcfanning:
@harleenkohli

### DIFF
--- a/src/Sarif/SarifUtilities.cs
+++ b/src/Sarif/SarifUtilities.cs
@@ -79,7 +79,10 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         public static Uri ConvertToSchemaUri(this SarifVersion sarifVersion)
         {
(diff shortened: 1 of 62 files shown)


src/Sarif/SarifUtilities.cs
```diff
@@ -79,7 +79,10 @@ public static string ConvertToText(this SarifVersion sarifVersion)
 
         public static Uri ConvertToSchemaUri(this SarifVersion sarifVersion)
         {
-            return new Uri(SarifSchemaUriBase + sarifVersion.ConvertToText(), UriKind.Absolute);
+            return new Uri(
+                    SarifSchemaUriBase +
+                    sarifVersion.ConvertToText() +
+                    (sarifVersion == SarifVersion.Current ? "-beta.0" : ""), UriKind.Absolute);
```
harleenkohli: >"-beta.0" [](start = 60, length = 9)

i will add a tracking item to see if this can be picked from our build constants instead, so that release process gets simpler.

PR #1424, opened 2019-04-11
366-TC feedback

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1878,7 +1883,7 @@
         "kinds": {
           "description": "A set of distinct strings that categorize the relationshship. Well-known kinds include canPrecede, canFollow, willPrecede, willFollow, superset, subset, equal, disjoint, relevant, and incomparable.",
           "type": "array",
-          "minItems": 1,
+          "default": [ "relevant" ],
```
harleenkohli: >default": [ "relevant" ], [](start = 11, length = 25)

this is currently breaking the validator. i will take a look at it tomorrow with a fresh mind. Raising this PR as a draft for awareness.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2954,14 +2971,14 @@
           }
         },
 
-        "artifactIndices": {
-          "description": "The indices within the run artifacts array of the artifact objects associated with the tool component.",
+        "artifacts": {
```
michaelcfanning: >artifacts [](start = 9, length = 9)

we got this wrong. the type here s/be artifactLocation, not artifact. the name of the property as 'artifacts' is correct. though.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -105,6 +105,18 @@
       "type": "object",
       "properties": {
 
+        "index": {
+          "description": "The index within the run artifacts array of the artifact object.",
+          "type": "integer",
+          "default": -1,
+          "minimum": -1
+        },
```
michaelcfanning: we don't need this index as it is not useful within the artifacts table itself (only useful if the artifact would be inlined elsewhere, as we actually do not plan to do, see my other comment).

the description property should remain, though.

michaelcfanning: please more 'index' and then merge.

---
In reply to: [275451263](https://github.com/Microsoft/sarif-sdk/pull/1424#discussion_r275451263) [](ancestors = 275451263)

harleenkohli: oops! i missed this change, will update!

---
In reply to: [275541609](https://github.com/Microsoft/sarif-sdk/pull/1424#discussion_r275541609) [](ancestors = 275541609,275451263)

PR #1426, opened 2019-04-13
Bugbash outcomes 000

harleenkohli:
Bugs fixed in this PR:
- remove run.markdownMessageMimeType
- externalPropertyFilerefence.itemCount minimum value shall be 1
- Add toolComponent.informationUri
- toolComponent.IsComprehensive default value should be false.
-  externalPropertyFileReferences - all instances should have uniform attributes - default : [ ], unique : true, minItems : 0
- MultiFormatMessageString.text is required
- remove extra space in "translationMetadata.informationUri
- sarifLog.inlineExternalProperties should be unique
- in artifactLocation object: anyOf {uri, index} is required
- externalPropertyFileReference.itemCount min should be 1
- run.results should be empty array [] by default (@harleenkohli -- it's actually `externalPropertyFileReferences.results` whose default is `[]`, not `run.results`. But the code you wrote is correct. -- Larry)


src/Test.FunctionalTests.Sarif/DefaultObjectPopulatingVisitor.cs
```diff
@@ -195,7 +195,8 @@ private void PopulatePropertyWithGeneratedDefaultValue(ISarifNode node, Property
             object propertyValue = null;
             Type propertyType = property.PropertyType;
 
-            bool isRequired = PropertyIsRequiredBySchema(node.GetType().Name, property.Name);
+            bool isRequired = PropertyIsRequiredBySchema(node.GetType().Name, property.Name) || 
+                PropertyIsAnyOfRequiredBySchema(node.GetType().Name, property.Name);
```
lgolding: >PropertyIsAnyOfRequiredBySchema [](start = 16, length = 31)

This logic doesn't look right. If a property appears in an `anyOf`, as `b` does here...
```
"someObject": {
  "required": [ "a" ],
  "anyOf": [
    {
     "required": [ "b" ]
    },
    {
      "required": [ "c" ]
    }
  ]
}
```
... you wouldn't conclude that `b` is required.

I see your comment below that this is a "quick fix" for something -- that at least for now, you must treat something as required if it appears in any `anyOf` element -- but why? Maybe I'll find out later...

lgolding: Anyway, I'm not really supposed to be reviewing SDK code, I should only be looking at the schema, so, over to that...

---
In reply to: [275125189](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275125189) [](ancestors = 275125189)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2388,6 +2399,7 @@
           "type": "array",
           "minItems": 0,
           "uniqueItems": false,
+          "default": [],
```
michaelcfanning: This is definitely wrong. The default is null. A non-empty array is an explicit indicator that a run succeeded, but there were no results. A null value indicates that the analysis was not able to be completed for some reason. @lgolding, did you notice this during your review? Harlene says the spec sets this default (empty array). That seems wrong to me.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2414,12 +2426,6 @@
           "type": "string"
         },
 
-        "markdownMessageMimeType": {
```
michaelcfanning: >markdownMessageMimeType [](start = 9, length = 23)

@lgolding, can you recall for me why we deleted this? simply can't remember the rationale (i do recall the decision)

src/Test.FunctionalTests.Sarif/DefaultObjectPopulatingVisitor.cs
```diff
@@ -276,16 +277,43 @@ private void PopulatePropertyWithGeneratedDefaultValue(ISarifNode node, Property
...
             string jsonPropertyName = GetJsonNameFor(propertyName);
+            JsonSchema propertySchema = GetJsonSchemaForObject(objectTypeName);
+
+            if (propertySchema.AnyOf != null && propertySchema.AnyOf.Count > 0)
+            {
+                // TODO: Add additional logic to randomly select one or more required subsets and populate only those.
+
+                // As a quick fix, we will populate all properties which are in any required subset.
+                // This means this method must return true for all properties in any required subset in the list.
+                foreach (var item in propertySchema.AnyOf)
```
michaelcfanning: >var [](start = 25, length = 3)

provide an explicit type here, please

src/Test.FunctionalTests.Sarif/DefaultObjectPopulatingVisitor.cs
```diff
@@ -276,16 +277,43 @@ private void PopulatePropertyWithGeneratedDefaultValue(ISarifNode node, Property
...
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        private JsonSchema GetJsonSchemaForObject(string objectTypeName)
+        {
+            // SarifLog.Version will be converted to sarifLog.version
```
michaelcfanning: >SarifLog [](start = 15, length = 8)

make it clear you're providing an example here

src/Sarif/Schemata/sarif-schema.json
lgolding: Closing loop: Because we only support Markdown, we specify GFM, and CommonMark is a subset of GFM.

---
In reply to: [275446870](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275446870) [](ancestors = 275446870)

src/Test.FunctionalTests.Sarif/DefaultObjectPopulatingVisitor.cs
lgolding: Closing loop: Harleen explained this to me. The variable name `isRequired` might be better, but I understand that the idea is to ensure that you don't end up generating a SARIF file that's missing a required property (or that has _none_ of a set of alternative required properties).

---
In reply to: [275125217](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275125217) [](ancestors = 275125217,275125189)

src/Sarif/Schemata/sarif-schema.json
lgolding: I did notice it but I got confused. In the PR's summary comment, I wrote:

- run.results should be empty array [] by default (@harleenkohli -- it's actually `externalPropertyFileReferences.results` whose default is `[]`, not `run.results`. But the code you wrote is correct. -- Larry)

So you see I knew that `run.results` shouldn't default to empty array. But when I said "the code you wrote is correct", it's because I thought this line _was_ in the `externalPropertyFileReferences` object. I was wrong; it's in the `run` object, and it needs to be moved to the `externalPropertyFileReferences` object.

Now as to what the spec says -- I'll check, and bug fix it if it says the default is `[]`.

---
In reply to: [275445302](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275445302) [](ancestors = 275445302)

michaelcfanning:
>         "results": {

Larry's right, this is actually the place where the default:[] array should be.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:635 in 7b91139. [](commit_id = 7b91139bdfa7f24ca33c9109f60699135125ff37, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
lgolding: Closing the loop: the spec does not specify a default for `run.results`, nor does it for any other object-valued property.

---
In reply to: [275452387](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275452387) [](ancestors = 275452387,275445302)

src/Test.FunctionalTests.Sarif/DefaultObjectPopulatingVisitor.cs
harleenkohli: Thanks Larry, as discussed, i am going to add a comment to clearly explain the intention. I like the way you described it so i might plagiarize this.

---
In reply to: [275450923](https://github.com/Microsoft/sarif-sdk/pull/1426#discussion_r275450923) [](ancestors = 275450923,275125217,275125189)

harleenkohli:
>         "results": {

dang!! i messed this up! thanks for the expnanation! now i see what i did!!

---
In reply to: [483332770](https://github.com/Microsoft/sarif-sdk/pull/1426#issuecomment-483332770) [](ancestors = 483332770)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:635 in 7b91139. [](commit_id = 7b91139bdfa7f24ca33c9109f60699135125ff37, deletion_comment = False)

PR #1428, opened 2019-04-16
bugbash-002 - EZPZ fixes!

harleenkohli:
- artifact.offset minimum should   be 0
- artifact.length   minimum should be 0
- result.graphTraversals, array   minItems should be 1
- result.stacks,   array minItems should be 1
- result.suppressions, array items   should be unique and default to null
- result.fixes   -> array minItems should be 1
- physicalLocation.id ->   minimum should be 0
- edgeTraversal.stepOverEdgeCount   minimum should be 0
- threadFlowLocation.nestingLevel   minimum should be 0
- threadFlowLocation.executionOrder   minimum should be 0

### DIFF
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -124,12 +124,14 @@
 
         "offset": {
           "description": "The offset in bytes of the artifact within its containing artifact.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "length": {
           "description": "The length of the artifact in bytes.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "roles": {
@@ -504,7 +506,8 @@
 
         "stepOverEdgeCount": {
           "description": "The number of edge traversals necessary to return from a nested graph.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "properties": {
@@ -1512,7 +1515,8 @@
 
         "id": {
           "description": "Value that distinguishes this physical location from all other physical locations in this run object.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "address": {
@@ -2114,7 +2118,7 @@
         "stacks": {
           "description": "An array of 'stack' objects relevant to the result.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
           "uniqueItems": false,
           "default": [],
           "items": {
@@ -2147,7 +2151,7 @@
         "graphTraversals": {
           "description": "An array of one or more unique 'graphTraversal' objects.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
           "uniqueItems": true,
           "default": [],
           "items": {
@@ -2170,8 +2174,7 @@
           "description": "A set of suppressions relevant to this result.",
           "type": "array",
           "minItems": 0,
-          "uniqueItems": false,
-          "default": [],
+          "uniqueItems": true,
           "items": {
             "$ref": "#/definitions/suppression"
           }
@@ -2231,9 +2234,8 @@
         "fixes": {
           "description": "An array of 'fix' objects, each of which represents a proposed fix to the problem indicated by the result.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
           "uniqueItems": true,
-          "default": [],
           "items": {
             "$ref": "#/definitions/fix"
           }
@@ -2789,12 +2791,14 @@
 
         "nestingLevel": {
           "description": "An integer representing a containment hierarchy within the thread flow.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "executionOrder": {
           "description": "An integer representing the temporal order in which execution reached this location.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "executionTimeUtc": {
@@ -2979,7 +2983,7 @@
         "locations": {
           "description": "An array of the artifactLocation objects associated with the tool component.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
           "default": [],
           "items": {
             "$ref": "#/definitions/artifactLocation"
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -2118,7 +2118,7 @@
         "stacks": {
           "description": "An array of 'stack' objects relevant to the result.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": false,
           "default": [],
           "items": {
@@ -2151,7 +2151,7 @@
         "graphTraversals": {
           "description": "An array of one or more unique 'graphTraversal' objects.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": true,
           "default": [],
           "items": {
@@ -2797,8 +2797,7 @@
 
         "executionOrder": {
           "description": "An integer representing the temporal order in which execution reached this location.",
-          "type": "integer",
-          "minimum": 0
+          "type": "integer"
         },
 
         "executionTimeUtc": {
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -130,8 +130,7 @@
 
         "length": {
           "description": "The length of the artifact in bytes.",
-          "type": "integer",
-          "minimum": 0
+          "type": "integer"
         },
 
         "roles": {
@@ -2234,8 +2233,9 @@
         "fixes": {
           "description": "An array of 'fix' objects, each of which represents a proposed fix to the problem indicated by the result.",
           "type": "array",
-          "minItems": 1,
+          "minItems": 0,
           "uniqueItems": true,
+          "default": [],
           "items": {
             "$ref": "#/definitions/fix"
           }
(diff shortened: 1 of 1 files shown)


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2114,7 +2118,7 @@
         "stacks": {
           "description": "An array of 'stack' objects relevant to the result.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
```
michaelcfanning: this s/be zero. this construct is no different in this regard than codeFlows or graphs. @lgolding

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2147,7 +2151,7 @@
         "graphTraversals": {
           "description": "An array of one or more unique 'graphTraversal' objects.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
```
michaelcfanning: >minItems [](start = 11, length = 8)

same not, s/be 0

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2789,12 +2791,14 @@
...
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
         },
 
         "executionOrder": {
           "description": "An integer representing the temporal order in which execution reached this location.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
```
michaelcfanning: >"minimum": 0 [](start = 10, length = 12)

how does this constraint help us? there's nothing in the spec that says these values are sequential. why should they be positive? what's the value?

src/Sarif/Schemata/sarif-schema.json
lgolding: +1. I don't remember this change from Harlene's spreadsheet -- is it possible this change belongs somewhere else? There's nothing special about "stacks".

---
In reply to: [276013428](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276013428) [](ancestors = 276013428)

src/Sarif/Schemata/sarif-schema.json
lgolding: Yes 0 is right. Also note that the default for both of these is `[]` -- which is right -- and the default shouldn't violate the `minItems` constraint.

---
In reply to: [276013562](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276013562) [](ancestors = 276013562)

src/Sarif/Schemata/sarif-schema.json
lgolding: It's true, they don't have to be sequential. All the spec says is that they have to increase with time (that is, if two tFLs both have `executionOrder`, then the one that occurs at the later time must have the larger `executionOrder`.

---
In reply to: [276014542](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276014542) [](ancestors = 276014542)

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: good point! Harlene, let's double-check the spec to make sure you didn't flag something in external property files or something.

---
In reply to: [276015729](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276015729) [](ancestors = 276015729,276013428)

src/Sarif/Schemata/sarif-schema.json
michaelcfanning: so what does the spec say? does it require this value to be positive? I don't see why it should.

---
In reply to: [276016264](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276016264) [](ancestors = 276016264,276014542)

lgolding: The spec actually does say they have to be positive (and if we believed that, the constraint would be "minimum: 1", not "minimum: 0"), but there's no reason for that constraint.

> A `threadFlowLocation` object **MAY** contain a property named `executionOrder` whose value is a positive integer that represents the temporal order in which execution reached this location, across all `threadFlowLocation` objects within all `threadFlow` objects belonging to a single codeFlow (§3.34).

src/Sarif/Schemata/sarif-schema.json
harleenkohli: Thanks, i will revert this and mention this as spec issue in excel sheet.

---
In reply to: [276015944](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276015944) [](ancestors = 276015944,276013562)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: thanks, going to remove this.

---
In reply to: [276017630](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276017630) [](ancestors = 276017630)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: marked it as a spec issue in excel, will revert it!

---
In reply to: [276016620](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276016620) [](ancestors = 276016620,276015729,276013428)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -124,12 +124,14 @@
...
         },
 
         "length": {
           "description": "The length of the artifact in bytes.",
-          "type": "integer"
+          "type": "integer",
+          "minimum": 0
```
michaelcfanning: >minimum [](start = 11, length = 7)

Is this right? How do we differentiate a zero-byte length artifact from one that hasn't had a length specified? Note that this note doesn't apply to 'offset', we can assume that 0 is correct in the default case (i.e., that the artifact isn't offset from a parent). @lgolding

michaelcfanning: To be clear, I wonder whether there shouldn't be a default value of -1 here indicating (not specified). With a min of -1.

---
In reply to: [276421274](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276421274) [](ancestors = 276421274)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2231,9 +2234,8 @@
         "fixes": {
           "description": "An array of 'fix' objects, each of which represents a proposed fix to the problem indicated by the result.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
```
michaelcfanning: this is wrong, it should be minItems 0 with a default of empty array.

michaelcfanning:
>           "uniqueItems": true,

where's the default value on this? should be empty array i think.

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2222 in 3319b8b. [](commit_id = 3319b8b893494a4ab34428677c3d85345e7724df, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2979,7 +2982,7 @@
         "locations": {
           "description": "An array of the artifactLocation objects associated with the tool component.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
```
michaelcfanning: this is not correct, s/be minItems of 0. note the discrepancy between this value and the default value of empty array provided.

src/Sarif/Schemata/sarif-schema.json
harleenkohli: ok, @larrygolding@comcast.net please confirm.

3.26.30 fixes property
A result object MAY contain a property named fixes whose value is an array of one or more unique (§3.7.3) fix objects (§3.53).


---
In reply to: [276422386](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276422386) [](ancestors = 276422386)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: spec text for reference: 

3.24.5 length property
An artifact object MAY contain a property named length whose value is a non-negative integer specifying the length of the artifact in bytes.


---
In reply to: [276421529](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276421529) [](ancestors = 276421529,276421274)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: you are right, the spec does say that it should be 0. Will correct this.

---
In reply to: [276423146](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276423146) [](ancestors = 276423146)

harleenkohli:
>           "uniqueItems": true,

spec for reference:

3.26.27 workItemUris property
A result object MAY contain a property named workItemUris whose value is either null or an array of zero or more unique (§3.7.3) strings each of which contains the absolute URI [RFC3986] of a work item associated with this result.
If workItemUris is absent, it SHALL default to null.


---
In reply to: [484250866](https://github.com/Microsoft/sarif-sdk/pull/1428#issuecomment-484250866) [](ancestors = 484250866)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2222 in 3319b8b. [](commit_id = 3319b8b893494a4ab34428677c3d85345e7724df, deletion_comment = False)

michaelcfanning:
>           "uniqueItems": true,

you're right!

---
In reply to: [484256796](https://github.com/Microsoft/sarif-sdk/pull/1428#issuecomment-484256796) [](ancestors = 484256796,484250866)

---
Refers to: src/Sarif/Schemata/sarif-schema.json:2222 in 3319b8b. [](commit_id = 3319b8b893494a4ab34428677c3d85345e7724df, deletion_comment = False)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: Sorry, i have reverted this change from this specific PR, because its a bit costly, we need to make fixes in some tests. I have this item in the excel sheet, lets do it in a separate PR

---
In reply to: [276427257](https://github.com/Microsoft/sarif-sdk/pull/1428#discussion_r276427257) [](ancestors = 276427257,276421529,276421274)

PR #1429, opened 2019-04-16
Bugbash outcomes 003 (more ezpz)

harleenkohli:
fixes: 

- threadFlowLocation.importance   should default to "important"
- request.index should have   default: -1, minimum: -1
- response.index   should have default: -1, minimum: -1
- resultProvenance.conversionSources   array minItems should be 1
- reportingDescriptor.id is required
- externalProperties.version is not a required property if its not root element

### DIFF
--- a/src/Sarif/Schemata/sarif-schema.json
+++ b/src/Sarif/Schemata/sarif-schema.json
@@ -739,8 +739,7 @@
           "description": "Key/value pairs that provide additional information about the external properties.",
           "$ref": "#/definitions/propertyBag"
         }
-      },
-      "required": [ "version" ]
+      }
     },
 
     "externalPropertyFileReference": {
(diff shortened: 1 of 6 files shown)


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2306,7 +2311,7 @@
         "conversionSources": {
           "description": "An array of physicalLocation objects which specify the portions of an analysis tool's output that a converter transformed into the result.",
           "type": "array",
-          "minItems": 0,
+          "minItems": 1,
```
michaelcfanning: >minItems [](start = 11, length = 8)

i don;t think this is right. please refer to the issue below for our driving principles. conversion sources is not an object that, if not populated, invalidates the containing object. therefore, in order not to be overly prescriptive, we allow it to be non-null and empty.

[https://github.com/oasis-tcs/sarif-spec/issues/270](https://github.com/oasis-tcs/sarif-spec/issues/270)

harleenkohli: Thanks! updating

---
In reply to: [276015852](https://github.com/Microsoft/sarif-sdk/pull/1429#discussion_r276015852) [](ancestors = 276015852)

PR #1434, opened 2019-04-17
bugbash: `artifact.length` should have `default: -1, minimum: -1`

harleenkohli:


src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -174,7 +174,7 @@ internal Artifact CreateFileData(FileDataVersionOne v1FileData, string key)
                 fileData = new Artifact
                 {
                     Hashes = v1FileData.Hashes?.Select(CreateHash).ToDictionary(p => p.Key, p => p.Value),
-                    Length = v1FileData.Length,
+                    Length = v1FileData.Length == 0 ? -1 : v1FileData.Length,
```
michaelcfanning: >v1FileData.Length == 0  [](start = 29, length = 23)

this means knowledge of zero length files from v1 is lost. that's ok!

PR #1436, opened 2019-04-18
Hoist physical location

harleenkohli:


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -157,24 +157,49 @@ private static bool ApplyChangesFromTC35(JObject sarifLog)
...
             return modifiedLog;
         }
 
-        private static bool RenameDefaultFileEncodingToDefaultEncoding(JObject sarifLog)
+        private static bool HoistIdsFromPhysicalLocationToLocation(JObject run)
         {
-            bool isModified = false;
+            string[] locationPathsToUpdate =
+            {
+                "results[].locations",
```
harleenkohli: >"results[].locations", [](start = 16, length = 22)

we have few more location object paths, are they dont care for us? 

eg. all instances of 

..threadFlowLocation.location,
..stackFrame.location

harleenkohli:
@michaelcfanning i cant vote on this, since i raised this PR, i have left one comment which i thought was worth cross checking

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: No known SARIF producer emits ids for any other locations! That's why i short-handed this. :)

---
In reply to: [276493698](https://github.com/Microsoft/sarif-sdk/pull/1436#discussion_r276493698) [](ancestors = 276493698)

PR #1439, opened 2019-04-18
artifact-roles-renames

harleenkohli:


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -170,9 +170,82 @@ private static bool ApplyChangesFromTC35(JObject sarifLog)
...
+                                    isModified = true;
+                                    break;
+                                }
+                            case "renamedFile":
+                                {
+                                    role.Value = "renamed";
+                                    isModified = true;
+                                    break;
+                                }
+                            case "uncontrolledFile":
```
michaelcfanning: >uncontrolledFile [](start = 34, length = 16)

you could have put all of these in a single case that rename the item simply by trimming the original file by its "File" suffix.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -170,9 +170,82 @@ private static bool ApplyChangesFromTC35(JObject sarifLog)
...
+                "runs[].artifacts[]"
+            };
+
+            bool actionOnLeafNode(JObject artifact)
+            {
+                if (artifact["roles"] is JArray roles)
+                {
+                    bool isModified = false;
+
+                    foreach(JValue role in roles)
```
michaelcfanning: >foreach [](start = 20, length = 7)

missing whitespace after foreach

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -170,9 +170,82 @@ private static bool ApplyChangesFromTC35(JObject sarifLog)
...
+                                    break;
+                                }
+                            case "addedFile":
+                                {
+                                    role.Value = "added";
+                                    isModified = true;
+                                    break;
+                                }
+                            case "deletedFile":
+                                {
```
michaelcfanning: >{ [](start = 32, length = 1)

your braces aren't indented to our style, you need to change your VS settings. these should be left aligned with the 'case' keyword. otherwise the contents of the case is effectively tabbed twice, which is a lot of right shifting

michaelcfanning: i've approved but feel free to fix the little things i commented on :)

---
In reply to: [277023975](https://github.com/Microsoft/sarif-sdk/pull/1439#discussion_r277023975) [](ancestors = 277023975)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
harleenkohli: oh! correct! thanks!

---
In reply to: [277023544](https://github.com/Microsoft/sarif-sdk/pull/1439#discussion_r277023544) [](ancestors = 277023544)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
harleenkohli: will do!

---
In reply to: [277024249](https://github.com/Microsoft/sarif-sdk/pull/1439#discussion_r277024249) [](ancestors = 277024249,277023975)

PR #1442, opened 2019-04-18
suppressions - separate kind and state (371)

harleenkohli:


src/Sarif/Autogenerated/Suppression.cs
```diff
@@ -46,6 +47,14 @@ public SarifNodeKind SarifNodeKind
         [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
         public SuppressionKind Kind { get; set; }
 
+        /// <summary>
+        /// A string that indicates the state of the suppression. Well-known values are `accepted`, `underReview` and `rejected`.
+        /// </summary>
+        [DataMember(Name = "state", IsRequired = false, EmitDefaultValue = false)]
+        [DefaultValue("accepted")]
```
rtaket: >accepted [](start = 23, length = 8)

A default value of "accepted" is odd to me (I realize it's specified in the bug.) But "accepted" implies that someone looked at the issue and declared it "accepted". Shouldn't the default be empty or "undefined".

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2698,13 +2698,17 @@
...
-            "suppressedExternally",
-            "underReview",
-            "suppressionRejected"
+            "inSource",
+            "external"
           ]
         },
 
+        "state": {
+          "description": "A string that indicates the state of the suppression. Well-known values are `accepted`, `underReview` and `rejected`.",
```
michaelcfanning: >description [](start = 11, length = 11)

shouldn't this be an enum? not an open ended string?

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -243,6 +246,36 @@ private static bool ConvertRunRedactionTokenToArray(JObject run)
...
 
+        private static bool RenameSuppressionKindValues(JObject run)
+        {
+            string suppressionsPathToUpdate = "results[].suppressions[]";
+
+            bool actionOnLeaf(JObject suppression)
+            {
+                if (suppression["kind"] is JValue kind && kind.Value is string)
+                {
+                    if (kind.Value as string == "suppressedInSource" )
```
michaelcfanning: >kind [](start = 24, length = 4)

you know these are strings already, so just use a cast. also you could capture inReview here and then create a new state object to hold it. we don't do that because it's just not worth the cost though, yes? i.e., no one is populating 'inreview'. i think that's fair

michaelcfanning: convert these to casts and the go right ahead and merge!

---
In reply to: [277075058](https://github.com/Microsoft/sarif-sdk/pull/1442#discussion_r277075058) [](ancestors = 277075058)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: the language in spec issue text seems to indicate an open ended string, but i could be wrong. I will check with Larry!

---
In reply to: [277074605](https://github.com/Microsoft/sarif-sdk/pull/1442#discussion_r277074605) [](ancestors = 277074605)

src/Sarif/Autogenerated/Suppression.cs
harleenkohli: thanks Ryley - leaving this PR open till we have a  final decision on the above.

Larry's suggestion was

Define two values, unknown and notReviewed. The default is notReviewed, which makes sense for direct producers. However, in case of converters (where we dont know what the original intention was), it can set the value to unknown.


@mikefan@microsoft.com what are you thoughts on this?


---
In reply to: [277044665](https://github.com/Microsoft/sarif-sdk/pull/1442#discussion_r277044665) [](ancestors = 277044665)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
harleenkohli: thanks! based on the other comment, i am going to wait on this to confirm if we actually need to fill the state value here. will work on the cast issue :)

---
In reply to: [277075132](https://github.com/Microsoft/sarif-sdk/pull/1442#discussion_r277075132) [](ancestors = 277075132,277075058)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: This should be enum!

---
In reply to: [277089440](https://github.com/Microsoft/sarif-sdk/pull/1442#discussion_r277089440) [](ancestors = 277089440,277074605)

Issue #373, opened 2016-06-24
Provide helper method to format messages


lgolding:
@bobschumaker wrote:

Is there a method to generate the formatted message? That way I wouldn’t have to do this:

```C#
      if (rule != null && rule.MessageFormats != null && rule.MessageFormats.ContainsKey(issue.FormattedRuleMessage.FormatId))
      {
                var format = rule.MessageFormats[issue.FormattedRuleMessage.FormatId];
                fields[6] = String.Format(format, issue.FormattedRuleMessage.Arguments.ToArray());
      }
```

lgolding:
There is such a helper method. It's now documented on the [How-To](https://github.com/Microsoft/sarif-sdk/blob/master/docs/how-to.md) page, under [Format a result message](https://github.com/Microsoft/sarif-sdk/blob/master/docs/how-to.md#format-a-result-message).

PR #1444, opened 2019-04-19
#373 & adding some other missing TC items in release history file.

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -2699,6 +2699,11 @@
           ]
         },
 
+        "justification": {
```
michaelcfanning: >justification [](start = 9, length = 13)

when was this proposed? i did not participate in the conversation. :) i don't want this data in the log file, personally.

lgolding:
Approving after discussion with and at request of @mikefan@microsoft.com

harleenkohli:
thanks so much Larry!

---
In reply to: [485005470](https://github.com/Microsoft/sarif-sdk/pull/1444#issuecomment-485005470) [](ancestors = 485005470)

src/Sarif/Schemata/sarif-schema.json
harleenkohli: marking this by design based on chat conversation.

---
In reply to: [277075812](https://github.com/Microsoft/sarif-sdk/pull/1444#discussion_r277075812) [](ancestors = 277075812)

PR #1451, opened 2019-04-23
region.byteoffset default to -1

harleenkohli:


src/Sarif.Converters/SemmleQlConverter.cs
```diff
@@ -221,6 +221,11 @@ private IList<Location> NormalizeRawMessage(string rawMessage, out string normal
                         };
                     }
 
+                    if (physicalLocation.Region.ByteOffset == 0)
```
michaelcfanning: >ByteOffset [](start = 48, length = 10)

this doesn't look right to me. isn't it possible that Semmle sets a byteOffset to 0? I see parsing occurring above. Since -1 is already the default, why do you have to set it here? I'd expect byteOffset to be -1 on instantiating the region. if the code above replaces it, then the default is overridden for a useful purpose. Ask Larry his opinion here.

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -715,7 +715,7 @@ internal Region CreateRegion(RegionVersionOne v1Region)
                 region = new Region
                 {
                     ByteLength = v1Region.Length,
-                    ByteOffset = v1Region.Offset,
+                    ByteOffset = v1Region.Offset == 0 ? -1 : v1Region.Offset,
```
michaelcfanning: >v1Region.Offset [](start = 33, length = 15)

what if byteLength > 0? in this case you should be preserving a byteOffset of 0, isn't that true?

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -759,7 +764,7 @@ internal Region CreateRegion(int startColumn, int startLine, int endColumn = 0,
                 region = new Region
                 {
                     ByteLength = length,
-                    ByteOffset = offset,
+                    ByteOffset = offset == 0 ? -1 : offset,
```
harleenkohli: >yteOffset = offset == 0 ? -1 : offset [](start = 21, length = 37)

DANG! i missed removing this :'(

src/Sarif.Converters/SemmleQlConverter.cs
harleenkohli: i updated this to use ByteLength instead. The reason i need to set it here is that incoming object has a default value of zero and if we dont override it, we end up getting zero valued ByteOffsets. I think having the check against ByteLength is a good check.

---
In reply to: [277867847](https://github.com/Microsoft/sarif-sdk/pull/1451#discussion_r277867847) [](ancestors = 277867847)

PR #1452, opened 2019-04-23
OneOf constraint on reportingDescriptor {shortDescription, fullDescription}

harleenkohli:


src/Sarif/Schemata/sarif-schema.json
```diff
@@ -1819,7 +1819,11 @@
           "$ref": "#/definitions/propertyBag"
         }
       },
-      "required": [ "id" ]
+      "required": [ "id" ],
+      "anyOf": [
+        { "required": [ "shortDescription" ] },
+        { "required": [ "fullDescription" ] }
+      ]
```
michaelcfanning: larry, is this a reasonable requirement? why wouldn't it be sufficient for someone to simply provide a readable name, for example?

I am comfortable with the spec language that says you SHOULD provide a description, but i don't think we should have SHALL (of either short or full description).

harleenkohli:
abandoning this change since we have agreed to remove this constraint

PR #1454, opened 2019-04-23
Enable Sarif.SDK .NET 4.5 Builds

ScottLouvau:
We want to consume the SARIF SDK from services which must use .NET 4.5 and should only consume CLS-compliant binaries.

It *looks like* the SARIF SDK can actually be built fairly easily to target .NET 4.5. If we target the corresponding version of .NET Standard (1.1), there are many missing things (SerializableAttribute, FileAttributes, HybridDictionary, etc).

However, it looks like we can ask to target .NET Standard 2.0 and .NET 4.5 both. In that case, I just had to change one "return multiple values" function. To be CLS Compliant, I just had to change two uint-based enums. 

If these changes look ok, I'd like to get the merged and publish a new NuGet release as soon as possible, so that we can start consuming them from our services. I'll need help with validating the NuGet package looks good and has the right content for 4.5 targeting in it.


src/Sarif/RuntimeConditions.cs
```diff
@@ -49,14 +49,14 @@ public enum RuntimeConditions : uint
...
-        TargetNotValidToAnalyze   = 0x20000000,
-        OneOrMoreWarningsFired    = 0x40000000,
-        OneOrMoreErrorsFired      = 0x80000000,
+        UnassignedNonfatal        = 0x03F0000,
+        RuleWasExplicitlyDisabled = 0x0400000,
+        RuleCannotRunOnPlatform   = 0x0800000,
+        RuleNotApplicableToTarget = 0x1000000,
+        TargetNotValidToAnalyze   = 0x2000000,
+        OneOrMoreWarningsFired    = 0x4000000,
+        OneOrMoreErrorsFired      = 0x8000000,
```
michaelcfanning: >0x8000000 [](start = 36, length = 9)

the idea previously here was to reserve the most significant bits for the non-fatal conditions. Can you please reset the enum values starting with 0x07000000? And set the nonfatal mask to 7FF?

I realize you are just preserving the readability of the previous values. But since we're having to change the literal underlying enum values in any case, would prefer to preserve the convention of reserving 'leftmost' bits for this class of data.

src/Sarif/OptionallyEmittedData.cs
```diff
@@ -70,6 +70,6 @@ public enum OptionallyEmittedData : uint
         // information in the SARIF log file. In the absence of this setting, any existing
         // data that would otherwise have been overwritten by the insert operation will
         // be preserved.
-        OverwriteExistingData = 0x80000000
+        OverwriteExistingData = 0x8000000
```
michaelcfanning: >0x8000000 [](start = 32, length = 9)

Same thing, please use 0x40000000 for this value...

michaelcfanning: feel free to merge after taking feedback

---
In reply to: [277882305](https://github.com/Microsoft/sarif-sdk/pull/1454#discussion_r277882305) [](ancestors = 277882305)

src/Sarif/RuntimeConditions.cs
ScottLouvau: I think you mean 0x40000000, which is the next bit down. We could use the high bit with unchecked casts, but that's confusing. I've now shifted the values and masks over one bit instead of four bits.

PR #1467, opened 2019-04-29
401 address object redesign

harleenkohli:


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -188,9 +188,88 @@ private static bool ApplyBeta3Changes(JObject sarifLog)
...
+            modifiedLog |= RenameBaseAddressAndOffsetInAddressObject(sarifLog);
+
             return modifiedLog;
         }
 
+        private static bool RenameBaseAddressAndOffsetInAddressObject(JObject sarifLog)
+        {
+            string[] addressPathsToUpdate =
+            {
+                "inlineExternalProperties[].addresses[]",
```
harleenkohli: >"inlineExternalProperties[].addresses[]", [](start = 16, length = 41)

we have all these address usages! the reason why i didnt use the simple rename function is because we have another object ("artifact") with a property named "offset". If we know for a fact that some of these objects are not being used, i can comment them out to avioud unneeded scans (e.g. all the inlineExternalProperties?)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -188,9 +188,88 @@ private static bool ApplyBeta3Changes(JObject sarifLog)
             // https://github.com/oasis-tcs/sarif-spec/issues/399
             modifiedLog |= ConvertInvocationToolExecutionSuccessfulToExecutionSuccessful(sarifLog);
 
+            // https://github.com/oasis-tcs/sarif-spec/issues/401
+            modifiedLog |= RenameBaseAddressAndOffsetInAddressObject(sarifLog);
+
             return modifiedLog;
         }
 
+        private static bool RenameBaseAddressAndOffsetInAddressObject(JObject sarifLog)
```
michaelcfanning: >RenameBaseAddressAndOffsetInAddressObject [](start = 28, length = 41)

don't do this! just delete it.

src/Sarif/Schemata/sarif-schema.json
```diff
@@ -48,18 +48,34 @@
...
+          "minimum": -1,
+          "default": -1
+
+        },
+
+        "relativeAddress": {
+          "description": "The address expressed as a byte offset from the absolute address of the top-most parent object.",
+          "type": "integer",
+          "minimum": -1,
+          "default": -1
```
michaelcfanning: remove these!

also, these will need to be nullable types in the SDK.

lgolding: She has done so.

---
In reply to: [279567287](https://github.com/Microsoft/sarif-sdk/pull/1467#discussion_r279567287) [](ancestors = 279567287)

PR #1474, opened 2019-05-02
Multitool: Enable cross-platform "dotnet tool" support

dbjorge:
Since .NET Core 2.1, the `dotnet` CLI tools have supported a concept called [Global Tools](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools), which allows a NuGet package containing a console app to be easily installed/distributed in a manner which is consistent with other tools and compatible across multiple platforms/CI systems.

This PR adds support for installing in this fashion to the Multitool. Usage (from any dev/CI environment with v2.1+ of the .NET Core SDK installed) would look like:

```
dotnet tool install --tool-path ./bin Sarif.Multitool
./bin/sarif-multitool <multitool_args>
```

This involved adding a netcoreapp2.1 TargetFramework to the Multitool. I presumed that adding this as a built/supported/published target implied you'd want the test projects to build/run for that framework like they do for the other supported frameworks, so this also adds netcoreapp2.1 targets to all the test projects and updates appveyor configuration appropriately to run them. It also updates appveyor to use a 2.1 .NET Core SDK version.

Here is an example of using the multitool from the standard `dotnet/core/sdk docker` linux container using nupkgs produced with these changes:

```
PS> type .\dockerfile
FROM mcr.microsoft.com/dotnet/core/sdk:2.2 AS build

# User experience: Install tool from nuget.org
# RUN dotnet tool install --tool-path ./bin Sarif.Multitool

# Local testing experience: Install tool from locally build nupkg
ARG sarif_nupkg_dir=./nupkg
ARG package_version=2.1.0-danielbj.0
COPY ${sarif_nupkg_dir}/* ./nupkg/
RUN dotnet tool install --tool-path ./bin --add-source ./nupkg --version ${package_version} Sarif.Multitool

# Run tool
ENTRYPOINT [ "./bin/sarif-multitool", "--help" ]

PS> docker build -t multitool-helloworld .
Sending build context to Docker daemon  3.846MB
Step 1/6 : FROM mcr.microsoft.com/dotnet/core/sdk:2.2 AS build
 ---> bbae085fa7eb
Step 2/6 : ARG sarif_nupkg_dir=./nupkg
 ---> Using cache
 ---> daca0c15ccc2
Step 3/6 : ARG package_version=2.1.0-danielbj.0
 ---> Using cache
 ---> 1084dbf25a99
Step 4/6 : COPY ${sarif_nupkg_dir}/* ./nupkg/
 ---> Using cache
 ---> d0a3352dc20c
Step 5/6 : RUN dotnet tool install --tool-path ./bin --add-source ./nupkg --version ${package_version} Sarif.Multitool
 ---> Using cache
 ---> a8e37cf4f3b6
Step 6/6 : ENTRYPOINT [ "./bin/sarif-multitool", "--help" ]
 ---> Using cache
 ---> fadcbd027889
Successfully built fadcbd027889
Successfully tagged multitool-helloworld:latest
SECURITY WARNING: You are building a Docker image from Windows against a non-Windows Docker host. All files and directories added to build context will have '-rwxr-xr-x' permissions. It is recommended to double check and reset permissions for sensitive files and directories.

PS> docker run multitool-helloworld
SARIF Multitool 2.1.0
© Microsoft Corporation. All rights reserved.

  validate                 Validate a SARIF file against the schema and against
                           additional correctness rules.

  convert                  Convert a tool output log to SARIF format.

  rewrite                  Transform a SARIF file to a reformatted version.

  transform                Transform a SARIF log to a different version.

  merge                    Merge multiple SARIF files into one.

  rebaseuri                Rebase the URIs in one or more sarif files.

  absoluteuri              Turn all relative Uris into absolute URIs (to be
                           used after rebaseUri is run)

  match-results-forward    Track results run over run by persisting IDs and
                           then matching them forward

  help                     Display more information on a specific command.

  version                  Display version information.

```


appveyor.yml
```diff
@@ -10,10 +10,8 @@ branches:
     - tc-31
     - sarif-v1
 
-platform: Any CPU
```
dbjorge: Dropping this [recommended by AppVeyor](https://help.appveyor.com/discussions/problems/10117-dotnet-test-with-xunit-fails-because-of-the-wrongly-resolved-path) for `dotnet test` compatibility

scripts/Run-Tests.ps1
```diff
@@ -34,27 +34,15 @@ Import-Module -Force $PSScriptRoot\Projects.psm1
 
 $ScriptName = $([io.Path]::GetFileNameWithoutExtension($PSCommandPath))
 
-$ReporterOption = $null
-if ($AppVeyor) {
-    $ReporterOption = "-appveyor"
```
dbjorge: This is dropped without replacement because xunit automatically detects this nowadays (see xunit help parameters). I verified that appveyor was still picking up tests as they ran without this.

scripts/Run-Tests.ps1
```diff
@@ -34,27 +34,15 @@ Import-Module -Force $PSScriptRoot\Projects.psm1
...
         Write-Information "Running tests in ${project}: $framework..."
-        Push-Location $BinRoot\${Platform}_$Configuration\$project\$framework
-        $dll = "$project" + ".dll"
-        if ($framework -eq "netcoreapp2.0") {
-            & dotnet ${TestRunnerRootPath}netcoreapp2.0\xunit.console.dll $dll $ReporterOption
-        } else {
-            & ${TestRunnerRootPath}net452\xunit.console.exe $dll $ReporterOption
-        }
+        Push-Location $SourceRoot\$project
+        & dotnet test --no-build --framework $framework --configuration $Configuration
```
dbjorge: This was required to enable netcoreapp2.1 tests because the entry point has to be netcoreapp2.1 cognizant and xunit.console only releases netcoreapp2.0 DLLs.

This update brings the script more in line with the current recommendations from both [dotnet](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test) and [xunit](https://xunit.net/docs/getting-started/netcore/cmdline). It handles correctly detecting the xunit runner based on test projects' `PackageReference`s and correctly handles the net461 case.

src/Sarif.Multitool/DotnetToolSettings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<DotNetCliTool Version="1">
+    <Commands>
+        <Command Name="sarif-multitool" EntryPoint="Sarif.Multitool.dll" Runner="dotnet" />
```
dbjorge: I picked "sarif-multitool" as the command name somewhat arbitrarily based on the most common looking conventions among [other .NET Core tools](https://github.com/natemcmaster/dotnet-tools). If you'd prefer "Sarif.Multitool" or just "sarif", I think those would both also be reasonable.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -14,7 +19,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <ItemGroup>
-    <PackageReference Include="CommandLineParser" Version="2.2.1" />
+    <PackageReference Include="CommandLineParser" Version="2.5.0" />
```
dbjorge: This was required to test the change in docker; the multitool breaks in docker using 2.2.1 due to https://github.com/commandlineparser/commandline/issues/153

src/Sarif.Multitool/DotnetToolSettings.xml
michaelcfanning: I think 'sarif' is probably right here. This utility is basically the swiss army knife of all SARIF processing operations, and so the concise form sounds right. There won't be another utility competing for this name.

---
In reply to: [280277257](https://github.com/Microsoft/sarif-sdk/pull/1474#discussion_r280277257) [](ancestors = 280277257)

PR #1475, opened 2019-05-02
skeleton work for poliCheck converter + test files

harleenkohli:


src/Test.UnitTests.Sarif.Converters/PoliCheckConverterTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    public class PoliCheckConverterTests
+    {
+        [Fact]
+        public void PoliCheckConverter_Convert_NullInput()
+        {
+            Assert.Throws<ArgumentNullException>(() => new PoliCheckConverter().Convert(null, null, OptionallyEmittedData.None));
```
harleenkohli: i should pass valid output here so that the test is targeted to input null check only

src/Test.FunctionalTests.Sarif/Test.FunctionalTests.Sarif.csproj
```diff
@@ -3551,6 +3551,9 @@
     <None Update="v2\ConverterTestData\FxCop\SarifSdkTest.xml.sarif">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="v2\ConverterTestData\PoliCheck\PoliCheck5.9.1_Sample001.xml">
```
harleenkohli: i could not figure out how the other test files (not *.sarif) are magically appearing in build output subfolder. do tell!

PR #1476, opened 2019-05-06
SarifToCsv sample updates for Sarif 2.0 RTM standard.

ScottLouvau:
A few small updates to make the SarifToCsv sample compatible with the current Sarif SDK object model.

### DIFF
--- a/src/Samples/SarifToCsv/Program.cs
+++ b/src/Samples/SarifToCsv/Program.cs
@@ -118,7 +118,7 @@ namespace SarifToCsv
             writers["CorrelationGuid"] = (writer, result, pLoc) => { writer.Write(result.CorrelationGuid ?? ""); };
             writers["Fingerprints"] = (writer, result, pLoc) => { writer.Write(String.Join("; ", result.Fingerprints?.Values ?? Array.Empty<string>())); };
             writers["HostedViewerUri"] = (writer, result, pLoc) => { writer.Write(result.HostedViewerUri?.ToString() ?? ""); };
-            writers["InstanceGuid"] = (writer, result, pLoc) => { writer.Write(result.InstanceGuid ?? ""); };
+            writers["Guid"] = (writer, result, pLoc) => { writer.Write(result.Guid ?? ""); };
             writers["Level"] = (writer, result, pLoc) => { writer.Write(result.Level.ToString()); };
             writers["Message.Text"] = (writer, result, pLoc) => { writer.Write(result.Message?.Text ?? ""); };
             writers["OccurrenceCount"] = (writer, result, pLoc) => { writer.Write(result.OccurrenceCount); };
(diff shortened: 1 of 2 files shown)


src/Samples/SarifToCsv/Program.cs
```diff
@@ -150,6 +149,13 @@ public static void ConvertSarifLog(JsonSerializer serializer, string sarifFilePa
             writers["Location.Region.StartLine"] = (writer, result, pLoc) => { writer.Write(pLoc.Region?.StartLine ?? -1); };
             writers["Location.Region.Tags"] = (writer, result, pLoc) => { writer.Write(String.Join("; ", ((IEnumerable<string>)pLoc.Region?.Tags) ?? Array.Empty<string>())); };
 
+            writers["Properties.MatchState"] = (writer, result, pLoc) => 
+            {
+                string state;
+                if(!result.TryGetProperty("matchState", out state)) { state = null; }
```
michaelcfanning: >{ state = null; } [](start = 68, length = 17)

this assignment isn't necessary. state will be set to null if TryGetProperty fails. you can't return from a method with an out parameter without initializing it.

src/Samples/SarifToCsv/Program.cs
```diff
@@ -150,6 +149,13 @@ public static void ConvertSarifLog(JsonSerializer serializer, string sarifFilePa
             writers["Location.Region.StartLine"] = (writer, result, pLoc) => { writer.Write(pLoc.Region?.StartLine ?? -1); };
             writers["Location.Region.Tags"] = (writer, result, pLoc) => { writer.Write(String.Join("; ", ((IEnumerable<string>)pLoc.Region?.Tags) ?? Array.Empty<string>())); };
 
+            writers["Properties.MatchState"] = (writer, result, pLoc) =>
```
michaelcfanning: >Properties.MatchState [](start = 21, length = 21)

it doesn't make sense to me that we would emit some private match state from properties and yet not provide suppressions information from the core SARIF types.

src/Samples/SarifToCsv/Program.cs
michaelcfanning: >matchState [](start = 43, length = 10)

since this isn't standard SARIF utilization, you should add a comment noting what one-off tool it originates with.

ScottLouvau: True. I like to be explicit in cases like this so one doesn't have to look up what the method returns (though out is a logical default). It also looks a little weird when one calls "TryXX" and then ignores the return value.

src/Samples/SarifToCsv/Program.cs
ScottLouvau: I'll take this out. This was to support my EngineTest comparisons quickly. I need to add support for collection properties in general in order to really handle the remaining result properties.

src/Samples/SarifToCsv/Program.cs
ScottLouvau: I just removed it instead.

PR #1480, opened 2019-05-07
CS: path-traversal rule.

lgolding:


src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -762,10 +788,18 @@ private string BuildSourcesString(HashSet<Tuple<string, string>> sources)
             var sb = new StringBuilder();
             foreach (Tuple<string, string> tuple in sources)
             {
-                if (tuple.Item1 == null || tuple.Item2 == null) { continue; }
```
lgolding: >tuple [](start = 20, length = 5)

This code didn't work when there was no name for the source, for example, when the tuple was `( , body)` or `( , other)`.

It also didn't format well if there was more than one source (it squished the sources together with no white space). With this change, the two sources above (which are real data from the `path-traversal` rule) format together as `body, other`.

lgolding:
With the changes in this PR, the first `path-traversal` result in the SARIF file looks like this:

```
        {
          "ruleId": "path-traversal",
          "level": "error",
          "message": {
            "id": "default",
            "arguments": [
              "body, other",
              "/webgoat/Content/UploadPathManipulation.aspx"
            ]
          },
          "locations": [
            {
              "logicalLocation": {
                "fullyQualifiedName": "OWASP.WebGoat.NET.UploadPathManipulation.btnUpload_Click"
              }
            }
          ],
          "codeFlows": [
            ...
          ],
          "webRequest": [
            ...
          ]
        }
```

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -685,7 +685,33 @@ private Result ConstructPathTraversalResult(ContrastLogReader.Context context)
...
+            };
+
+            string page = context.RequestTarget;
+            string sources = BuildSourcesString(context.Sources);
+
+            result.Message = new Message
+            {
+                Id = "default",
+                Arguments = new List<string>
+                {              // Attacker-controlled path traversal was observed from
```
michaelcfanning: >Attacker [](start = 34, length = 8)

Suggest adding an article, 'An attacker-controlled path...'

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -762,10 +788,18 @@ private string BuildSourcesString(HashSet<Tuple<string, string>> sources)
             var sb = new StringBuilder();
             foreach (Tuple<string, string> tuple in sources)
             {
-                if (tuple.Item1 == null || tuple.Item2 == null) { continue; }
+                if (sb.Length > 0) { sb.Append(", "); }
+
                 // Item1 is the name, Item2 is the source type, e.g., parameter
-                sb.Append(tuple.Item1 + "(" + tuple.Item2 + ")");
+                if (!string.IsNullOrWhiteSpace(tuple.Item1)) { sb.Append(tuple.Item1 + ": "); }
```
michaelcfanning: >string.IsNullOrWhiteSpace [](start = 21, length = 25)

you probably don't need this check. verify by trying to execute this code (sb.Append(null + ":"). the compiled string.concat call handles null gracefully, i believe

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -941,7 +975,37 @@ private PhysicalLocation CreatePhysicalLocation(string uri, Region region = null
                 Region = region,
                 ContextRegion = region
             };
+        }
+
+        private static readonly Regex LogicalLocationRegex =
+            new Regex(
+                @"
+                  ([^\s]*\s+)?         # Skip over an optional leading blank-terminated return type name such as 'void '
```
michaelcfanning: >         # Skip over an optional leading blank-terminated return type name such as 'void ' [](start = 30, length = 90)

wow are comments supported in these regexes? if so, looks like once you introduce a # character the remainder of the line is treated as a comment? TIL

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -941,7 +975,37 @@ private PhysicalLocation CreatePhysicalLocation(string uri, Region region = null
...
+                  (?<fqln>[^(]+)       # Take everything up to the opening parenthesis.
+                ",
+                RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.IgnorePatternWhitespace);
```
michaelcfanning: >IgnorePatternWhitespace [](start = 84, length = 23)

oh! is this setting what enables that feature?

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: Fixed. Glad I have leave to copy-edit the messages (which I assume came directly from CS). In that case, also "'{1}' page" => "page '{1}'".

---
In reply to: [282149268](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282149268) [](ancestors = 282149268)

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -762,10 +788,18 @@ private string BuildSourcesString(HashSet<Tuple<string, string>> sources)
...
-                if (tuple.Item1 == null || tuple.Item2 == null) { continue; }
+                if (sb.Length > 0) { sb.Append(", "); }
+
                 // Item1 is the name, Item2 is the source type, e.g., parameter
-                sb.Append(tuple.Item1 + "(" + tuple.Item2 + ")");
+                if (!string.IsNullOrWhiteSpace(tuple.Item1)) { sb.Append(tuple.Item1 + ": "); }
+
+                string sourceType = !string.IsNullOrWhiteSpace(tuple.Item2)
+                    ? tuple.Item2
+                    : "<unknown source type>";
```
michaelcfanning: >"<unknown source type>" [](start = 22, length = 23)

you created a method scoped const string below for the System prefix but not here. is there a method you use to determine when it's appropriate to do this?

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: Yes. Super-helpful for complicated regexes.

---
In reply to: [282150310](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282150310) [](ancestors = 282150310)

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: Yes!

---
In reply to: [282150462](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282150462) [](ancestors = 282150462)

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: No, I was just lazy here, will fix.

---
In reply to: [282150892](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282150892) [](ancestors = 282150892)

lgolding: Fixed.

---
In reply to: [282152025](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282152025) [](ancestors = 282152025,282150892)

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: I don't want "" either.

---
In reply to: [282149859](https://github.com/microsoft/sarif-sdk/pull/1480#discussion_r282149859) [](ancestors = 282149859)

PR #1481, opened 2019-05-07
Fix #1479: PrereleaseCompatibilityTransformer did not move id from physLoc to location

### LINKED ISSUE #1479: PhysicalLocation.id property getting lost during 2.1.0 transformation.
Reported by Rusty Scrivens. 
Expected behavior: Id should be moved from `PhysicalLocation` to the parent `Location` object during 2.1.0 transformation.
Actual behavior: Id property is getting removed.

Original text from Rusty's mail:
Just a heads up I ran into this bug while updating 2.0.0 sarif files via the transform function of the latest published multitool. The Id isn’t getting moved from PhysicalLocation to the parent Location. Which is breaking the embedded links matching in messages. Below is the two files, original(2.0.0) on the left, transformed(2.1.0) on the right:

location-id-removed

Looks like it’s from this spec change:
https://github.com/oasis-tcs/sarif-spec/issues/375

I found this behavior when transforming this file:  https://github.com/microsoft/sarif-sdk/blob/master/src/Test.FunctionalTests.Sarif/v2/SpecExamples/Comprehensive.sarif
It also happens with a version that’s more up to date("2.0.0-csd.2.beta.2019-02-20"):  https://github.com/microsoft/sarif-vscode-extension/blob/master/samples/Comprehensive.sarif

Thanks,
Rusty

lgolding:


lgolding:

lgolding:

PR #1482, opened 2019-05-08
Add functional test for Contrast Security converter.

lgolding:


lgolding:
The "file move" is because I renamed directory v2/ConverterTestData/Contrast to .../ContrastSecurity.

PR #1486, opened 2019-05-09
CS: crypto-bad-mac: Use message text closer to CS's text.

lgolding:
This ties up a loose end. When I converted the `crypto-bad-mac` rule, I didn't have an account that gave me access to CS's message text, so I made up a message based on the `crypto-bad-cypher` rule. It turns out the actual text is even closer to the text of the `crypto-bad-cypher` rule.


lgolding:
Merged with admin superpowers.

PR #1488, opened 2019-05-09
Fix #1487: .NET Fx version of Multitool crashes on startup

### LINKED ISSUE #1487: Multitool crashes on launch: Can't find Command.Line.dll
```
c:\dev\sarif\sarif-sdk>bld\bin\AnyCPU_Release\Sarif.Multitool\net461\Sarif.Multitool.exe

Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly
'CommandLine, Version=2.2.1.0, Culture=neutral, PublicKeyToken=de6f01bd326f8c32'
or one of its dependencies. The located assembly's manifest definition does not match
the assembly reference. (Exception from HRESULT: 0x80131040)
```

lgolding:


lgolding:

lgolding:

src/build.props
```diff
@@ -11,7 +11,7 @@
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
     
     <!-- VersionPrefix denotes the current Semantic Version for the SDK. Must be updated before every nuget drop. -->
-    <VersionPrefix>2.1.0</VersionPrefix>
+    <VersionPrefix>2.1.1</VersionPrefix>
```
lgolding: >VersionPrefix [](start = 5, length = 13)

If this now denotes the "Semantic Version for the SDK" it should be called `SdkSemanticVersion`.

src/build.props
```diff
@@ -29,9 +29,9 @@
     place. These properties are actually used by the PowerShell script that
     hides the previous package versions on nuget.org.
     -->
-    <PreviousVersionPrefix>2.0.0-csd.2.beta.2019-04-03</PreviousVersionPrefix>
+    <PreviousVersionPrefix>2.0.0</PreviousVersionPrefix>
```
lgolding: >PreviousVersionPrefix [](start = 5, length = 21)

What is the difference between this and `PreviousPackageVersion` below? Is this still used?

Even if this is still used, it is apparently not a "prefix", since its previous value was `2.0.0-csd.2.beta.2019-04-03`. What is this?

lgolding:
NOTE: I am **not** doing the test in this PR so you can publish the fixed SDK ASAP. I will do another PR with the test.

src/build.props
harleenkohli: You are right, i named it similarly yesterday but that broke local build! Unfortunately there is a .net core dependency to call it "versionPrefix" :/ If we dont have this explicitly set, it is defaulted to 1.0.0. So, we added the comment instead.  (I should have added this note as well i.e. the name has a dependency!)

---
In reply to: [282603197](https://github.com/microsoft/sarif-sdk/pull/1488#discussion_r282603197) [](ancestors = 282603197)

src/build.props
harleenkohli: oops, this is wrong, thanks for catching this! I forgot to rename it. It should be "PreviousStableSarifVersion". I dont think we have an explicit usage for this  variable today but lets have the right name here!

---
In reply to: [282603779](https://github.com/microsoft/sarif-sdk/pull/1488#discussion_r282603779) [](ancestors = 282603779)

harleenkohli: reg "PreviousVersionPrefix"  - hmm... we have to interpret it as "PreviousSdkSemanticVersion"

---
In reply to: [282609970](https://github.com/microsoft/sarif-sdk/pull/1488#discussion_r282609970) [](ancestors = 282609970,282603779)

lgolding: The **previous** PR should have changed this to `2.1.0-rtm.0`.

---
In reply to: [282609970](https://github.com/microsoft/sarif-sdk/pull/1488#discussion_r282609970) [](ancestors = 282609970,282603779)

PR #1491, opened 2019-05-13
FortifyFprConverter huge file support

ScottLouvau:
- FortifyFprConverter now two-pass to write results as-we-go to avoid out of memory.
- FortifyFprConverter now uses Run.ThreadFlowLocations to reuse Fortify UnifiedNodePool nodes (Sarif becomes much smaller).
- FortifyFprConverter has properties to allow omitting ContextRegions, all Snippets, and all ThreadFlowLocations (stacks).
- ResultLogJsonWriter fixes to include all current Run properties.
- SarifLogger/ResultLogJsonWriter duplication elimination.


src/Samples/SarifToCsv/Program.cs
```diff
@@ -23,13 +23,24 @@ namespace SarifToCsv
     /// </summary>
     public class Program
     {
+        public class WriteContext
```
ScottLouvau: I didn't mean for the SarifToCsv work to be in this PR. I made a branch from the in progress branch, merged another, and then pushed, and it updated the branch from the PR. :/

src/Sarif.Converters/ToolFormatConverter.cs
```diff
@@ -55,6 +56,8 @@ public ToolFormatConverter(AssemblyLoadFileDelegate assemblyLoadFileDelegate = n
 
             // FileMode settings here will results in an exception being raised if the input 
             // file does not exist, and that an existing output file will be overwritten
+            Stopwatch w = Stopwatch.StartNew();
```
ScottLouvau: I will remove - debuggability while I was coding.

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/TwoResultsWithNodeRefs.sarif
```diff
@@ -391,13 +266,153 @@
...
+              "contextRegion": {
+                "startLine": 79,
+                "endLine": 85,
+                "snippet": {
+                  "text": "                                Logger.Error(LogMessage.RETRY_ATTEMPTS_EXCEEDED, new object[] { tool, server});\n                                continue;\n                            }\n                            Logger.Error(ex.ToString());\n                        }\n                    }\n\n"
+                }
+              }
+            },
+            "message": {
+              "text": "Error(0)"
```
michaelcfanning: >"text": "Error(0)" [](start = 14, length = 18)

Is this really the message? Seems like there's something further to resolve here. Probably not a gap (if one) that you introduced. open issue for it?

src/Sarif.Converters/ToolFormatConverter.cs
michaelcfanning: Time to remove now? :)

---
In reply to: [283578926](https://github.com/microsoft/sarif-sdk/pull/1491#discussion_r283578926) [](ancestors = 283578926)

src/Samples/SarifToCsv/Program.cs
```diff
@@ -85,70 +98,109 @@ public static void ConvertSarifLog(JsonSerializer serializer, string sarifFilePa
...
-            writers["RuleId"] = (writer, result, pLoc) => { writer.Write(result.RuleId ?? ""); };
-            writers["RuleIndex"] = (writer, result, pLoc) => { writer.Write(result.RuleIndex); };
-            writers["SuppressionStates"] = (writer, result, pLoc) => { writer.Write(result.SuppressionStates.ToString()); };
-            writers["Tags"] = (writer, result, pLoc) => { writer.Write(String.Join("; ", ((IEnumerable<string>)result.Tags) ?? Array.Empty<string>())); };
-            writers["WorkItemUris"] = (writer, result, pLoc) => { writer.Write(String.Join("; ", result.WorkItemUris?.Select((uri) => uri.ToString()) ?? Array.Empty<string>())); };
+            context.Writer.NextRow();
+            if(context.Writer.RowCountWritten % 1000 == 0) { Console.Write("."); }
+        }
+
+        public static Dictionary<string, Action<WriteContext>> BuildWriters()
```
michaelcfanning: >Action [](start = 41, length = 6)

Still no suppressions data? Seems like we should have this, since it is a concept we exercise in the pattern matcher. Fine to defer.

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -71,63 +74,25 @@ public void Initialize(Run run)
...
-                _jsonWriter.WritePropertyName("defaultEncoding");
-                _serializer.Serialize(_jsonWriter, run.DefaultEncoding);
-            }
-
-            if (run.RedactionTokens != null)
-            {
-                _jsonWriter.WritePropertyName("redactionTokens");
-                _serializer.Serialize(_jsonWriter, run.RedactionTokens);
-            }
+            SerializeIfNotNull(_run.AutomationDetails, "automationDetails");
```
michaelcfanning: >automationDetails [](start = 56, length = 17)

Where did your property ordering come from? May be preferable to keep it in sync with the schema since (as a result of our code gen ordering) this will drive log property ordering when using standard JSON.NET serialization.

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/TwoResultsWithNodeRefs.sarif
ScottLouvau: This comes from the <Action> in the <Node> in the stack. It looks like for this rule they are the function name and argument, but it varies. It looks like this is a fallback path for when there isn't an explicit "NodeLabel" in the FVDL to use instead.

It's from here in the FVDL we converted:
`
<UnifiedNodePool>
  <Node id="4">
    <SourceLocation path="Program.cs" line="82" lineEnd="82" colStart="0" colEnd="0" contextId="2" snippet="F64E01C1BF6395D7221768BE7CC7BBB9#Program.cs:82:82"/>
    <Action type="InCall">Error(0)</Action>
    <Knowledge>
      <Fact primary="false" type="Call">Direct : Helpers.Logger.Error</Fact>
    </Knowledge>
  </Node>
`

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -409,25 +433,7 @@ private void ParseVulnerability()
                 {
                     // Get the rule GUID from the ClassId element.
                     string ruleId = _reader.ReadElementContentAsString();
```
michaelcfanning: >ruleId [](start = 27, length = 6)

The guid isn't really appropriate as a rule id. Furthermore, I'm not sure we're setting this value on the appropriate reportingDescriptor.guid property (where it belongs)? Not your problem, maybe we can file an issue?

src/Sarif/Writers/ResultLogJsonWriter.cs
ScottLouvau: I tried to:
- Maintain the current split between InitializeRun and CompleteRun
- Continue to write the "special" components in the order the code was before.
- Write everything else alphabetically to make it easy to check for missing properties relative to the full Run list.

I can make it match the schema order, but only within the before and after results segments.

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -1043,11 +1063,29 @@ private void ParseMachineInfo()
...
 
+        private ReportingDescriptor FindOrCreateRule(string ruleId, out int ruleIndex)
+        {
+            if (_ruleIdToIndexMap.TryGetValue(ruleId, out ruleIndex))
+            {
+                return _rules[ruleIndex];
+            }
+            else
+            {
+                ReportingDescriptor rule = new ReportingDescriptor() { Guid = ruleId };
```
michaelcfanning: >Guid = ruleId [](start = 71, length = 13)

Ah, we are in fact setting the guid.

src/Sarif.Converters/FortifyFprConverter.cs
ScottLouvau: It looks like Chris had a discussion with Yekaterina about this and ended up committing this way. See https://github.com/Microsoft/sarif-sdk/pull/1394. Do you want an issue in the sarif-sdk to investigate further?

src/Sarif.Converters/ToolFormatConverter.cs
ScottLouvau: Fixed.

src/Samples/SarifToCsv/Program.cs
ScottLouvau: Added Suppressions and came up with a model to add arbitrary Properties. The other collections looked much more complex to decide how to serialize.

src/Sarif/Writers/ResultLogJsonWriter.cs
ScottLouvau: Changed to move everything down to "CompleteRun" and to make everything ordered per schema except Results. 

This still won't be consistent with serializing from the whole OM because of the ordering of Results itself, and also will be inconsistent if the API user calls one of the ResultLogJsonWriter methods to write another collection early (ex: WriteArtifacts).

PR #1499, opened 2019-05-23
remove schema from id value

michaelcfanning:
@harleenkohli


michaelcfanning:
@harleenkohli reviewed this in person.

PR #1500, opened 2019-05-23
Draft: Sarif Paging and Json Map

ScottLouvau:
This is a draft PR for Sarif Paging. I still need to add tests for paging overall and may tweak the design.

This adds a Sarif.Multitool "page" verb to extract a slice of results from another Sarif file. This is an easy way to turn a huge Sarif file into manageable chunks for tools which can't use DeferredReader.

Parsing a huge file still takes a long time - ~90 seconds for the 7.15GB minified, converted huge Fortify sample. To avoid having to do this repeatedly to get different pages, I define a "JsonMap" concept which compactly describes the large file so that we can easily cut out chunks of it. I designed JsonMap to be potentially usable in other contexts.

I fixed a few bugs and performance issues in LineMappingStreamReader to make this work.
Two bad things: Json.net returns the Line and Char of the last char in a token, so the closest to the start we can get for literals is "previous token end + 2", which is only after the comma separator if there's no whitespace. It's quite expensive to really check and the comma might not be in the buffer depending on how the read happens.

Also, JsonTextReader uses int for LinePosition, and for minified files over 2GB, it overflows. I had to make a class to watch this happening and correct for it.

Ideally we'd probably write our own very minimal JsonScanner to walk the Json and identify exact token start byte offsets to avoid all of the mapping and cleanup. That's kind of crazy, though.


src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -0,0 +1,226 @@
...
+        private const long ArrayStartSizeEstimateBytes = 5;    // Bytes for one ArrayStart location, if under 10KB.
+
+        public double MaxFileSizePercentage { get; private set; }
+        public int MinimumSizeForNode { get; private set; }
+
+        /// <summary>
+        ///  Construct a JsonMapBuilder for the given target output size.
+        ///  (ex: 0.01 means 1% of the input file size)
+        /// </summary>
+        /// <param name="maxFileSizePercentage">Maximum Size of the Map relative to the source file (0.10 means 10%)</param>
```
michaelcfanning: >ximum Size of the Map [](start = 50, length = 21)

NIT: add casing here in this comment

src/Test.Utilities.Sarif/ResourceExtractor.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.IO;
```
michaelcfanning: >System [](start = 6, length = 6)

needs copyright header

src/Test.Utilities.Sarif/ResourceExtractor.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Reflection;
+
+namespace Microsoft.CodeAnalysis.Test.Utilities.Sarif
```
michaelcfanning: we'll need to discuss standard namespace conventions for test code at some point, won't randomize this PR with it.

src/Test.UnitTests.Sarif/Map/LongArrayDeltaConverterTests.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Fact]
+        public void LongArrayDeltaConverter_Basics()
+        {
+            RoundTrip(null);
+            RoundTrip(new long[] { });
+            RoundTrip(new long[] { -1 });
+            RoundTrip(new long[] { 0, 0, 0, 0, 0 });
+            RoundTrip(new long[] { 100, 101, 102, 103, 104, 105 });
+            RoundTrip(new long[] { 100, -100, 200, -200, 300, -300 });
```
michaelcfanning: >RoundTrip [](start = 12, length = 9)

add a test that incorporates long max and min values, perhaps?

src/Test.UnitTests.Sarif/Map/LongArrayDeltaConverterTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        private static void RoundTrip(long[] values)
+        {
+            Container before = new Container(); 
+            if(values != null)
```
michaelcfanning: >if [](start = 12, length = 2)

whitespace

src/Sarif.Multitool/PageOptions.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    [Verb("page", HelpText = "Extract a subset of results from a source SARIF file.")]
+    internal class PageOptions
```
michaelcfanning: >PageOptions [](start = 19, length = 11)

are 'inline', 'force', 'pretty-print' useful for this command? they tend to be standard for other log file processing/producing things.

src/Sarif.Multitool/PageCommand.cs
```diff
@@ -0,0 +1,160 @@
...
+            Func<Stream> inputStreamProvider = () => _fileSystem.OpenRead(options.InputFilePath);
+            long firstResultStart = results.FindArrayStart(options.Index, inputStreamProvider);
+            long lastResultEnd = results.FindArrayStart(options.Index + options.Count + 1, inputStreamProvider) - 1;
+
+            // Build the Sarif Log subset
+            long lengthWritten = 0;
+            byte[] buffer = new byte[64 * 1024];
+
+            using (FileStream output = File.Create(options.OutputFilePath))
+            using (FileStream source = File.OpenRead(options.InputFilePath))
```
michaelcfanning: >FileStream [](start = 19, length = 10)

i see, no pretty print. :)

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -0,0 +1,226 @@
...
+        public JsonMapBuilder(double maxFileSizePercentage)
```
michaelcfanning: >JsonMapBuilder [](start = 15, length = 14)

probably want to ensure maxFilePercentage range is good here?

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Fixed.

src/Test.Utilities.Sarif/ResourceExtractor.cs
ScottLouvau: Changed to match surrounding classes.

src/Test.Utilities.Sarif/ResourceExtractor.cs
ScottLouvau: Ran CodeFormatter

src/Test.UnitTests.Sarif/Map/LongArrayDeltaConverterTests.cs
ScottLouvau: Formatted.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Added verification it's > 0 and < 100. I can't say that values within that range aren't potentially useful. I use 20x in some tests and am trying 0.0001 for the huge file.

src/Test.UnitTests.Sarif/Map/LongArrayDeltaConverterTests.cs
ScottLouvau: Added, along with comments warning about the real range limits.

src/Sarif.Multitool/PageCommand.cs
ScottLouvau: Right. We're copying stream subsets, so we'll echo the formatting in the input and omit indenting between boundaries we're writing.

src/Sarif.Multitool/PageOptions.cs
ScottLouvau: Added force, with different default and explanation.
Pretty-Print can't be implemented here.
Inline seems like a bad idea for Page because the map would be built for an immediately invalidated file and all data outside the desired range would be lost.

src/Sarif.Multitool/PageOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+        public bool Force { get; set; } = true;
+
+        [Option(
+            'o',
+            "output",
+            HelpText = "A file path of the SARIF subset output file.",
+            Required = true)]
+        public string OutputFilePath { get; set; }
+
+        [Option('a',
```
harleenkohli: >a [](start = 17, length = 1)

why 'a'? maybe 'm' would be more intuitive?

src/Sarif.Multitool/PageCommand.cs
```diff
@@ -0,0 +1,195 @@
...
+            w.Stop();
+            Console.WriteLine($"Done in {w.Elapsed.TotalSeconds:n1}s.");
+
+            return root;
+        }
+
+        private void ExtractPage(PageOptions options, JsonMapNode root)
+        {
+            Stopwatch w = Stopwatch.StartNew();
+            Console.WriteLine($"Extracting Page [{options.Index}, {options.Index + options.Count}) from \"{options.InputFilePath}\" into \"{options.OutputFilePath}\"...");
```
harleenkohli: >) [](start = 97, length = 1)

this doesn't match with the open square bracket " [ ",

src/Sarif.Multitool/PageCommand.cs
```diff
@@ -0,0 +1,195 @@
...
+            if (!_fileSystem.FileExists(options.InputFilePath)) { throw new FileNotFoundException($"Input file \"{options.InputFilePath}\" not found."); }
+
+            if (options.Force == false && _fileSystem.FileExists(options.OutputFilePath))
+            {
+                Console.WriteLine($"Output file \"{options.OutputFilePath}\" already exists. Stopping.");
+                return;
+            }
+
+            // Load the JsonMap, if previously built and up-to-date, or rebuild it
+            string mapPath = Path.ChangeExtension(options.InputFilePath, ".map.json");
```
harleenkohli: >string mapPath = Path.ChangeExtension(options.InputFilePath, ".map.json"); [](start = 12, length = 74)

NIT: we could move this processing inside the method since its an internal information.. 


unless... the idea here was to ensure we reusability of the method if someone chooses a different path for storing the map file, in which case, i relent :)

just curious!

src/Sarif.Multitool/PageOptions.cs
ScottLouvau: Good suggestion - the result of an earlier rename.
I've removed the short name for this argument, as I think it'll be rarely used and isn't especially intuitive.

src/Sarif.Multitool/PageCommand.cs
ScottLouvau: Yep, good suggestion. This was another facet of a prior factoring.

src/Sarif.Multitool/PageCommand.cs
ScottLouvau: This was formal mathematical notation. I'm writing and naming the arguments separately now to be clearer:
$"Extracting {options.Count:n0} results from index {options.Index:n0}\r\n  from \"{options.InputFilePath}\"\r\n  into \"{options.OutputFilePath}\"..."

PR #1505, opened 2019-05-28
CS: Implement rule http-only-disabled

lgolding:
Also:

- When the XML contains a `snippet` property, generate `physicalLocation.region`, but not `contextRegion`. There is no need to generate `contextRegion` when it is the same as `region`.
- Bug fix: We we not clearing the web request properties from the context after each `finding`. As a result, we were emitting "stale" `webRequest` objects on findings that don't have a request at all (such as `http-only-disabled`, which is how I noticed this).
- Bug fix: Don't emit an empty `webRequest` object if the XML does not define a request. (Again, `http-only-disabled` was doing this -- as were all the "NYI" rules.)

### DIFF
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -325,6 +325,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private Result ConstructInsecureHashAlgorithmsResult(ContrastLogReader.Context context)
         {
+            // crypto-bad-mac : Insecure Hash Algorithms
+
             Result result = CreateResultCore(context);
 
             Stack stack = context.MethodEvent.Stack;
@@ -370,7 +372,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private Result ConstructPagesWithoutAntiClickjackingControlsResult(ContrastLogReader.Context context)
         {
-            // cache-controls-missing : Anti-Caching Controls Missing
+            // clickjacking-control-missing : Pages Without Anti-Clickjacking Controls
 
             // <properties name="/webgoat/Content/EncryptVSEncode.aspx">1536704253063</properties>
             // <properties name="/webgoat/WebGoatCoins/MainPage.aspx">1536704186306</properties>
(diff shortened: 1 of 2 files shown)


src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -918,21 +950,22 @@ private Result CreateResultCore(ContrastLogReader.Context context)
...
-                Headers = context.Headers,
-                Parameters = context.Parameters,
-                Body = string.IsNullOrEmpty(context.RequestBody)
-                    ? null
-                    : new ArtifactContent
-                    {
-                        Text = context.RequestBody
-                    }
-            };
+            return context.HasRequest() ?
```
lgolding: >HasRequest [](start = 27, length = 10)

This change prevents the creation of empty `webRequest` objects.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1088,10 +1120,19 @@ internal class Context
...
+                // it will always have a "protocol" attribute. (The same is true of "method",
+                // but there's no need to check both.)
+                return RequestProtocol != null;
+            }
+
             internal void RefineFinding(string ruleId)
             {
                 RuleId = ruleId;
                 ClearProperties();
+                ClearRequest();
```
lgolding: >ClearRequest [](start = 16, length = 12)

This fix prevents us from outputting "stale" `webRequest` objects.

lgolding:

lgolding:

harleenkohli:
>             }

was this a bug? (i see we now have code for cleaning up certain properties, so most likely this change is by design) flagging it just in case :)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:9622 in 91fe22a. [](commit_id = 91fe22a0c7903f8b06bac56f967ac04de9bb2025, deletion_comment = True)

harleenkohli:
>             "id": "default",

very cool to see stuff light up in test, thanks! its a nice idea to have TCs pre-written! :)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:11271 in 91fe22a. [](commit_id = 91fe22a0c7903f8b06bac56f967ac04de9bb2025, deletion_comment = False)

harleenkohli:
>             }

never mind, clearly a bug after looking at the whole file!

---
In reply to: [496717647](https://github.com/microsoft/sarif-sdk/pull/1505#issuecomment-496717647) [](ancestors = 496717647)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:9622 in 91fe22a. [](commit_id = 91fe22a0c7903f8b06bac56f967ac04de9bb2025, deletion_comment = True)

lgolding:
>             }

Yes (and it was my bug, because I'm the one who implemented `WebRequest` in the converter). Take a look at the `finding` in WebGoat.xml starting on line 3336. You see it has no request information. The `webRequest` property here was left over in the context from a previous `finding`. That's what the change on line 1135 here fixes.

---
In reply to: [496717647](https://github.com/microsoft/sarif-sdk/pull/1505#issuecomment-496717647) [](ancestors = 496717647)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:9622 in 91fe22a. [](commit_id = 91fe22a0c7903f8b06bac56f967ac04de9bb2025, deletion_comment = True)

Issue #1503, opened 2019-05-25
notification.time is being zapped during v2 to v1 transformation


harleenkohli:
refer to the test case: OneRunWithInvocationAndNotification for reproing this behavior. "time" value should be produced but its not.

PR #1509, opened 2019-05-29
notification.time getting removed bug

harleenkohli:
fix for: https://github.com/microsoft/sarif-sdk/issues/1503

### DIFF
--- a/src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithNotificationTime.sarif
+++ b/src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithNotificationTime.sarif
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-10-10",
-  "version": "2.0.0-csd.2.beta.2018-10-10",
+  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-09-26",
+  "version": "2.0.0-csd.2.beta.2018-09-26",
   "runs": [
     {
       "tool": {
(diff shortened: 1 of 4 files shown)


src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithInvocationAndNotifications.sarif
```diff
@@ -31,7 +31,7 @@
               },
               "level": "error",
               "threadId": 52,
-              "time": "2016-07-16T14:18:43.119Z",
```
harleenkohli: >time" [](start = 15, length = 5)

we noted this bug in a recent PR. we incorrectly assumed that this was a transformer bug. Actually, its not! This was an incorrect schema issue in the test.

we have code for time -> timeUtc transformation. However, that transformation applies prior to "2018-10-10" release. So, i have fixed this test to be a valid "2018-10-10" schema, where the field is called timeUtc, not time.

But to cross check that our transformation correctly works, i have added another test case (with an older schema version).

src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/Inputs/OneRunWithNotificationTime.sarif
```diff
@@ -0,0 +1,61 @@
...
+                "fileLocation": {
+                  "uri": "file:///home/buildAgent/src/crypto/hash.cpp"
+                }
+              },
+              "message": {
+                "text": "Exception evaluating rule \"C2152\". Rule disabled; run continues."
+              },
+              "level": "error",
+              "threadId": 52,
+              "time": "2016-07-16T14:18:43.119Z",
```
harleenkohli: >"time" [](start = 14, length = 6)

this will internally transform to utctime during prerelease transformation (not visible in output because the test case is really for v2 -> v1 conversion. so, internally, first this file is upgraded to rtm schema version, then converted to v1. The fact that we see the time field in output (v1) file, is proof that transformation was applied correctly.

PR #1512, opened 2019-05-29
CS: Implement rule session-rewriting.

lgolding:
This is the last WebGoat-related rule.


lgolding:

PR #1513, opened 2019-05-30
Do not emit an artificial absolute URI base

lgolding:
I’d like you both to look at this PR, not because it has anything to do with Contrast Security, but because it corrects a misuse of the `uriBaseId` mechanism. This is generally useful guidance.

The CS converter does three things wrong:

1. It artificially prepends "E:\src\WebGoat" to the `uri` property of every `artifactLocation` object it creates. This is wrong. In general `artifactLocation.uri` should be a relative reference. (There are exceptions; this isn’t one.)
2. It also emits a `run.originalUriBaseIds` property that assigns that same prefix to `SITE_ROOT`. But the converter has no way of knowing this value; it’s not in the XML file.
3. Having set `SITE_ROOT`, the converter doesn’t actually _use_ it as the value for `uriBaseId` in the `artifactLocation` objects it creates. 

The right thing to do is:

1. Leave the `artifactLocation.uri` values as relative references.
2. Do not supply the `uri` in `run.originalUriBaseIds`, since you don’t know its value.
3. _Do_ provide a `description` in `run.originalUriBaseIds`, allowing a viewer to display a hint to the user when prompting for the value.
4. Add `"uriBaseId": "SITE_ROOT"` to every `artifactLocation` object.

### DIFF
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -69,11 +69,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             Run run = sarifLog.Runs[0];
             _rules = run.Tool.Driver.Rules.ToDictionary(rule => rule.Id);
 
-            run.OriginalUriBaseIds = new Dictionary<string, ArtifactLocation>
-            {
-                {  "SITE_ROOT", new ArtifactLocation { Uri = new Uri(@"E:\src\WebGoat.NET") } } 
-            };
-
             // 3. Now, parse all the contrast XML to create the complete results set.
             var results = new List<Result>();
             var reader = new ContrastLogReader();
@@ -1090,13 +1085,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private PhysicalLocation CreatePhysicalLocation(string uri, Region region = null)
         {
-            uri = @"E:\src\WebGoat.NET" + uri.Replace(@"/", @"\");
-
             return new PhysicalLocation
             {
                 ArtifactLocation = new ArtifactLocation
                 {
-                    Uri = new Uri(uri, UriKind.Absolute)
+                    Uri = new Uri(uri, UriKind.RelativeOrAbsolute)
                 },
                 Region = region
             };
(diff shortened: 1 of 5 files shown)


lgolding:

lgolding:

lgolding:

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -597,7 +617,7 @@ private Result ConstructFormsWithoutAutocompletePreventionResult(ContrastLogRead
                     {
                         ArtifactLocation = new ArtifactLocation
                         {
-                            //UriBaseId = "RuntimeGenerated",
```
lgolding: >RuntimeGenerated [](start = 43, length = 16)

@mikefan@microsoft.com or @harleenkohli, whichever one of you wrote or commented this line -- _did_ you intend a different `uriBaseId` than `SITE_ROOT` here?

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -12,6 +12,22 @@
...
+    </EmbeddedResource>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Update="ConverterResources.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>ConverterResources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
```
lgolding: >ItemGroup [](start = 4, length = 9)

Years ago (literally), when we converted these projects to .NET Core, we removed all the .cs files. But it turns out you still need the incantation for handling .resx files. You don't notice the problem until you try to add a new resource string and you find the .Designer.cs file doesn't get updated and your build fails.

We've run into this and fixed it in other projects (see for example SdkResources.resx in Sarif.csproj. There might or might not be more of these left to fix.

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: No, I looked at the output from this rule. `SITE_ROOT` is correct.

---
In reply to: [289119180](https://github.com/microsoft/sarif-sdk/pull/1513#discussion_r289119180) [](ancestors = 289119180)

src/Sarif.Converters/Sarif.Converters.csproj
harleenkohli: thanks for explanation, good to know!

---
In reply to: [289120150](https://github.com/microsoft/sarif-sdk/pull/1513#discussion_r289120150) [](ancestors = 289120150)

PR #1515, opened 2019-05-31
Fix #1514: location.logicalLocation convenience setter mishandles null

### LINKED ISSUE #1514: location.logicalLocation convenience setter mishandles null
If you set `location.LogicalLocation` to `null`, the convenience property sets `location.LogicalLocations` to an array with a single `null` element, which serializes as `"logicalLocations": [ {} ]`. The right answer is for it to set `location.LogicalLocations` to `null`, which would be omitted from the serialization

lgolding:
If you set `location.LogicalLocation` to `null`, the convenience property sets `location.LogicalLocations` to an array with a single `null` element, which serializes as `"logicalLocations": [ {} ]`. The right answer is for it to set `location.LogicalLocations` to `null`, which would be omitted from the serialization.

Add unit tests for the convenience property.

### DIFF
--- a/src/Sarif/Core/Location.cs
+++ b/src/Sarif/Core/Location.cs
@@ -10,7 +10,16 @@ namespace Microsoft.CodeAnalysis.Sarif
         public LogicalLocation LogicalLocation
         {
             get { return LogicalLocations?[0]; }
-            set { LogicalLocations = new List<LogicalLocation> { value }; }
+            set {
+                if (value != null)
+                {
+                    LogicalLocations = new List<LogicalLocation> { value };
+                }
+                else
+                {
+                    LogicalLocations = null;
+                }
+            }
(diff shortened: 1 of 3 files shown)


lgolding:

lgolding:

src/Sarif/Core/Location.cs
```diff
@@ -10,7 +10,16 @@ public partial class Location
         public LogicalLocation LogicalLocation
         {
             get { return LogicalLocations?[0]; }
-            set { LogicalLocations = new List<LogicalLocation> { value }; }
```
harleenkohli: >set { LogicalLocations = new List<LogicalLocation> { value }; } [](start = 12, length = 63)

yuck! this was my doing! thanks for fixing this :)

PR #1516, opened 2019-05-31
Update CS rules file to final SARIF 2.1.0 schema

lgolding:


src/Sarif.Converters/RulesData/ContrastSecurity.sarif
```diff
@@ -1,749 +1,604 @@
 {
-  "$schema": "http://json.schemastore.org/sarif-2.0.0-csd.2.beta.2018-10-10",
-  "version": "2.0.0-csd.2.beta.2018-10-10",
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.1.json",
```
lgolding: >$schema [](start = 3, length = 7)

I just ran "multitool transform" on it.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -62,9 +62,8 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
             using (var stream = assembly.GetManifestResourceStream(ContrastSecurityRulesData))
             using (var streamReader = new StreamReader(stream))
             {
-                string prereleaseRuleDataLogText = streamReader.ReadToEnd();
-                PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(prereleaseRuleDataLogText, Newtonsoft.Json.Formatting.Indented, out string currentRuleDataLogText);
```
harleenkohli: >PrereleaseCompatibilityTransformer. [](start = 16, length = 35)

if we have a future schema version released (and i say 'if' :)), this means we would need to remember to update the rulesData to that upgraded version or we risk that this file goes out of sync. Having a transformer call was solving that problem transparently for us.

Without looking at the code, i ask: do we have any test already that fails if we forget to upgrade the RulesData?

lgolding: That's a good point. To answer your question, if we changed the schema and forgot to update the rules data file, then the CS converter's functional test would fail (because it would attempt to deserialize the rules data file and would fail).

The usefulness of the compatibility transform depends on how often we update the schema. During spec development, when we were updating the schema constantly, it was very helpful. If we never updated the schema again, there'd be no point to it at all.

Somewhere in between is the world we're actually living in. It's a real coin flip for me. I like to have the absolute minimum amount of code; OTOH, some degree of future proofing is sometimes sensible.

I'll revert this change.

---
In reply to: [289584481](https://github.com/microsoft/sarif-sdk/pull/1516#discussion_r289584481) [](ancestors = 289584481)

PR #1520, opened 2019-06-04
Multitool Query Mode

ScottLouvau:
This change adds Multitool 'Query' mode, which takes a Result expression (ex: "BaselineState != Unchanged AND Risk > 1.5") and evaluates it against an input log. It logs the result count and runtime by default, and can also log the matching results, write the filtered log to a file, return the count as an exit code, and return an exit code for count above a threshold.

Query mode only supports simple Result properties for now, and only filtering Results for now. It was designed so that we can support a broader property and entity set in the future without breaking compatibility.

- Implement Expression grammar and parser in Sarif (.Query)
- Implement ExpressionEvaluator helper classes for Double, Long, Bool, String, and Enum.
- Implement ExpressionEvaluator factory for simple properties in Result.
- Add Multitool query command itself
- Fix multiple multitool arguments used the same short form.
- Added test to prevent short form duplication in the future.
- Added OM method to get a ToolComponent from a ToolComponentReference.
- Added OM methods to get a RuleDescriptor from a RuleDescriptorReference.
- Added IFileSystem support for opening as stream.
- Made CommandBase Read/Write Sarif more memory efficient (Stream, not separate full string)
- Simplified TransformCommand via stream IFileSystem changes.


src/Sarif/Core/Result.cs
```diff
@@ -10,5 +12,63 @@ public partial class Result
         public bool ShouldSerializeWorkItemUris() { return this.WorkItemUris != null && this.WorkItemUris.Any((s) => s != null); }
 
         public bool ShouldSerializeLevel() { return this.Level != FailureLevel.Warning; }
+
+        /// <summary>
+        ///  Look up the ReportingDescriptor for this Result.
+        /// </summary>
+        /// <param name="run">Run instance containing this Result</param>
+        /// <returns>ReportingDescriptor for Result Rule, if available</returns>
+        public ReportingDescriptor GetReportingDescriptor(Run run)
```
rtaket: >GetReportingDescriptor [](start = 35, length = 22)

The method name is ambiguous since it could refer to the Rule or the Taxa. Perhaps it should be GetRule().

src/Test.UnitTests.Sarif.Multitool/GenericCommandTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace Sarif.Multitool.UnitTests
+{
+    public class GenericCommandTests
+    {
+        [Fact]
+        public void CommandsDoNotHaveConflictingShortNames()
```
harleenkohli: >CommandsDoNotHaveConflictingShortNames() [](start = 20, length = 40)

thanks for adding this test!

src/Sarif/Query/Evaluators/ResultEvaluators.cs
```diff
@@ -0,0 +1,39 @@
...
+                case "correlationguid":
+                    return new StringEvaluator<Result>(r => r.CorrelationGuid, term, StringComparison.OrdinalIgnoreCase);
+                case "guid":
+                    return new StringEvaluator<Result>(r => r.Guid, term, StringComparison.OrdinalIgnoreCase);
+                case "hostedvieweruri":
+                    return new StringEvaluator<Result>(r => r.HostedViewerUri?.ToString(), term, StringComparison.OrdinalIgnoreCase);
+                case "kind":
+                    return new EnumEvaluator<Result, ResultKind>(r => r.Kind, term);
+                case "level":
+                    return new EnumEvaluator<Result, FailureLevel>(r => r.Level, term);
```
rtaket: It's a bit weird that the severity level is an EnumEvaluator. But I guess there's no choice since Level is defined as an enum in SARIF?

src/Sarif/Query/ExpressionParser.cs
```diff
@@ -0,0 +1,354 @@
...
+            foreach (Literal<CompareOperator> op in CompareOperators)
+            {
+                if (text.StartsWith(op.Text, StringComparison.OrdinalIgnoreCase))
+                {
+                    text = text.Substring(op.Text.Length);
+                    return op.Value;
+                }
+            }
+
+            throw new QueryParseException("compare operator", text);
```
rtaket: >text [](start = 62, length = 4)

Should the exception print the StringSlice as it entered the method? i.e. it looks like this method modifies the StringSlice, so it's unclear if we want the original text or the updated text in the exception.

src/Sarif/Query/ExpressionParser.cs
```diff
@@ -0,0 +1,354 @@
...
+            {
+                if (!IsWhitespace(text[whitespaceCount])) { break; }
+            }
+
+            text = text.Substring(whitespaceCount);
+        }
+
+        private static bool IsWhitespace(char c)
+        {
+            return c == ' ';
```
rtaket: >' ' [](start = 24, length = 3)

Only space?

src/Sarif/Core/Result.cs
ScottLouvau: Good idea - renamed.

src/Sarif/Query/Evaluators/ResultEvaluators.cs
ScottLouvau: Yep. The evaluator for each property is dictated by the type in the code. Enum => EnumEvaluator.

src/Sarif/Query/ExpressionParser.cs
ScottLouvau: The exception should write the slice, which should start with the thing the parser had trouble with. Each place in the parser which throws should not consume the objectionable thing. In this case, we only throw if none of the operators in the loop were recognized, so whatever was first will not have been consumed.

src/Sarif/Query/ExpressionParser.cs
ScottLouvau: Yeah. I wanted to stick to just space as whitespace for now (JSON allows space, tab, carriage return, newline) but factored a method out in case we start accepting multiline expressions and wanted to expand it later.

src/Test.UnitTests.Sarif.Multitool/GenericCommandTests.cs
ScottLouvau: No problem. =)

I fixed the one case and then thought there would likely be more and they would be easy to add in the future.

Issue #1526, opened 2019-06-10
Transformation from SARIF 1.0 to 2.x throws Argument-out-of-range exception if result.locations is an empty array


harleenkohli:
Originally reported by Nathan Davidson of MS.

PR #1527, opened 2019-06-10
BUGFIX: Transformation from SARIF 1.0 to 2.x throws Argument-Out-of-Range exception if result.locations is an empty array

harleenkohli:
* BUGFIX: Transformation from SARIF 1.0 to 2.x throws Argument-Out-of-Range exception if `result.locations` is an empty array. https://github.com/microsoft/sarif-sdk/issues/1526


src/ReleaseHistory.md
```diff
@@ -390,4 +390,5 @@
...
 * BUGFIX: `location.logicalLocation` convenience setter mishandles null. https://github.com/microsoft/sarif-sdk/issues/1514
 * BUGFIX: Upgrade schemas to latest version (remove `draft-04` from `$schema` property and change `id` to `$id`). This is necessary because the schemas use the `uri-reference` format, which was not defined in draft-04. https://github.com/microsoft/sarif-sdk/issues/1521
-* API BREAKING: The `Init` methods in the Autogenerated SARIF object model classes are now `protected virtual`. This enables derived classes to add additional properties without having to copy the entire code of the `Init` method.
\ No newline at end of file
+* API BREAKING: The `Init` methods in the Autogenerated SARIF object model classes are now `protected virtual`. This enables derived classes to add additional properties without having to copy the entire code of the `Init` method.
+* BUGFIX: Transformation from SARIF 1.0 to 2.x throws Argument-Out-of-Range exception if `result.locations` is an empty array. https://github.com/microsoft/sarif-sdk/issues/1526
```
lgolding: >Argument-Out-of-Range exceptio [](start = 54, length = 30)

May as well just say the type name `ArgumentOutOfRangeException`.

src/Test.UnitTests.Sarif/TestData/SarifVersionOneToCurrentVisitor/Inputs/WebScout.sarif
```diff
@@ -0,0 +1,577 @@
...
+                "scanConfigurationId": 128483,
+                "rootUrls": [
+                    "https://example.com/"
+                ]
+            }
+        },
+      "results": [
+        {
+          "ruleId": "WS099",
+          "level": "error",
```
harleenkohli: > "level": "error", [](start = 9, length = 18)

Question: ok, i manually edited this before uploading the file. The version which was sent to us had this value as "5". This failed for me when i plugged it as a unit test case. I just want to understand how this works for our customers? where does the enum conversion happen?

src/ReleaseHistory.md
harleenkohli: i agree! why didnt i!

---
In reply to: [292205908](https://github.com/microsoft/sarif-sdk/pull/1527#discussion_r292205908) [](ancestors = 292205908)

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
```diff
@@ -50,9 +50,11 @@
     <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\MinimumWithLanguage.sarif" />
     <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\NestedInnerExceptionsInNotifications.sarif" />
     <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\OneRunWithAllReportingDescriptors.sarif" />
+    <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\WebScout.sarif" />
```
lgolding: >WebScout [](start = 75, length = 8)

This is an internal tool. I suggest:

1. Rename it to something like V1ZeroLengthLocationsArray.sarif
2. Strip it down to the absolute minimum file that demonstrates the problem. One result, no optional properties... just what you need to trigger the bug.

harleenkohli: Ok! will update!

---
In reply to: [292206492](https://github.com/microsoft/sarif-sdk/pull/1527#discussion_r292206492) [](ancestors = 292206492)

PR #1528, opened 2019-06-11
Fortify severity fix based on Micros focus feedback

harleenkohli:


src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -121,6 +121,10 @@ public override void Convert(Stream input, IResultLogWriter output, OptionallyEm
             _nodeDictionary.Clear();
             _snippetDictionary.Clear();
 
+            // Uncomment following Line to fetch FVDL content from any FPR file. This is not needed for conversion.
+            // However, we keep a copy of FVDL for each test file for debugging purposes.
+            // string fvdlContent = ExtractFvdl(OpenAuditFvdlReader(input));
```
harleenkohli: >// string fvdlContent = ExtractFvdl(OpenAuditFvdlReader(input)); [](start = 12, length = 64)

added this comment (and a simple method) for fvdl extraction from the fpr file. This is handy for new files.

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -449,13 +466,6 @@ private void ParseVulnerability(IResultLogWriter output)
                 {
                     rule.SetProperty(_strings.Subtype, _reader.ReadElementContentAsString());
                 }
-                else if (AtStartOfNonEmpty(_strings.InstanceSeverity))
-                {
-                    if (double.TryParse(_reader.ReadElementContentAsString(), out double rank))
-                    {
-                        result.Rank = rank;
-                    }
-                }
```
harleenkohli: this should be removed because:

- according to Katrina, Instance severity is a deprecated property.
- we are not mapping it correctly to rank in any way since rank's range is [0,100] but InstanceSeverity ranges from [0,5] so this is wrong.

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -477,6 +487,41 @@ private void ParseVulnerability(IResultLogWriter output)
...
+            // Low impact - ["1.0", "0,0"] => Note
+            // Any other value (invalid input) or no value => None
+
+            if (rule.TryGetProperty("Impact", out string impactValue))
+            {
+                if (float.TryParse(impactValue, out float impact))
+                {
+                    if (impact > 5.0)
+                    {
+                        return FailureLevel.None;
```
harleenkohli: >FailureLevel.None; [](start = 31, length = 18)

to use a very uncouth american phrase, i pulled it out my... no i cant say it i am sorry.

I just randomly decided that any invalid value (ie. not [0 - 5.0] should be considered "None". I think the alternate is that this value should be left unset -> which means it would get the default value of "Warning". So, my intent here is to say - NO, we didnt know what the severity of this issue is because it wasnt provided!


..just tell me. Thank you.

src/Test.UnitTests.Sarif.Converters/Test.UnitTests.Sarif.Converters.csproj
```diff
@@ -13,6 +13,12 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
+  <ItemGroup>
+    <None Remove="TestData\FortifyFprConverter\ExpectedOutputs\SdkScan.sarif" />
```
harleenkohli: >SdkScan [](start = 63, length = 7)

aah, i need to rename this file to a more appropriate name e.g. scanWithSeverityInfo etc.

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
```diff
@@ -6,6 +6,7 @@
       "results": [
         {
           "ruleIndex": 0,
+          "level": "none",
```
harleenkohli: >"level": "none", [](start = 10, length = 16)

i know you hate it (possibly!) see my other comment in the cs file

harleenkohli:
>           "ruleIndex": 37,

missing result.level here means - rule 37 violation is a warning! (i cross checked)

Also, side note: is it not ok to add rule guid here OR rule ID in the rules array? It was really hard to manually verify this without having a a quick visual for mapping!

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:8 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

harleenkohli:
>           "codeFlows": [

i apologize for such a big file, i tried to cut it short but its not easy since the real source of this the FPR file which has a very weird format - i didnt want to spend too much time trying to edit it. I know its possible, if the huge size bothers you, i can give it another try.

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:34 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

harleenkohli:
>           "rules": [

more fun starts here - i decided that it might be a good idea to surface the severity related matrices somewhere. Hence, i added the three called out by Katrina - Accuracy, Impact and Probability in the property bag.

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:47525 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

harleenkohli:
>                 "Probability": "3.0",

here!

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:47551 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

src/ReleaseHistory.md
```diff
@@ -391,4 +391,5 @@
 * BUGFIX: `location.logicalLocation` convenience setter mishandles null. https://github.com/microsoft/sarif-sdk/issues/1514
 * BUGFIX: Upgrade schemas to latest version (remove `draft-04` from `$schema` property and change `id` to `$id`). This is necessary because the schemas use the `uri-reference` format, which was not defined in draft-04. https://github.com/microsoft/sarif-sdk/issues/1521
 * API BREAKING: The `Init` methods in the Autogenerated SARIF object model classes are now `protected virtual`. This enables derived classes to add additional properties without having to copy the entire code of the `Init` method.
-* BUGFIX: Transformation from SARIF 1.0 to 2.x throws ArgumentOutOfRangeException, if `result.locations` is an empty array. https://github.com/microsoft/sarif-sdk/issues/1526
\ No newline at end of file
+* BUGFIX: Transformation from SARIF 1.0 to 2.x throws ArgumentOutOfRangeException, if `result.locations` is an empty array. https://github.com/microsoft/sarif-sdk/issues/1526
+* BUGFIX: Improve Result Severity calculation for Fortify Converter based on MicroFocus feedback.
```
lgolding: >Severity [](start = 25, length = 8)

You could possibly justify capitalizing "Result" because it shows up in the SDK that way (although I personally wouldn't). But "Severity" isn't an SDK term (we use "Level"), so don't capitalize it.

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
lgolding: You're right, I do :-). I'm going to look very carefully at this!

---
In reply to: [292263379](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292263379) [](ancestors = 292263379)

lgolding:
>           "rules": [

Yes, it's legitimate for a converter to use the property bag to hold "native" values for which SARIF has no direct equivalent.

---
In reply to: [500671794](https://github.com/microsoft/sarif-sdk/pull/1528#issuecomment-500671794) [](ancestors = 500671794)

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:47525 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

lgolding:
>                 "Probability": "3.0",

What are you calling my attention to here?

---
In reply to: [500671973](https://github.com/microsoft/sarif-sdk/pull/1528#issuecomment-500671973) [](ancestors = 500671973)

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:47551 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

lgolding:
>           "ruleIndex": 37,

Yes, missing `level` means: Use the `defaultConfiguration.level` value from the rule metadata, if present; otherwise use `"warning"`.

I'm not sure what you're asking in your second paragraph. If you are asking "Is it ok not to supply `result.rule.guid`?", the answer is "yes." `result.rule` is optional, and `reportingDescriptor.guid` is optional. If you're asking whether it is ok not to supply `tool.driver.rules[0].id`, the answer is "no." `reportingDescriptor.id` is required.

Have I misunderstood the question?

---
In reply to: [500670733](https://github.com/microsoft/sarif-sdk/pull/1528#issuecomment-500670733) [](ancestors = 500670733)

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:8 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

src/Test.UnitTests.Sarif.Converters/FortifyFprConverterTests.cs
```diff
@@ -40,5 +42,59 @@ public void FortifyFprConverter_Convert_TwoResultsWithNodeRefs()
...
         }
+
+        [Fact]
+        public void FortifyFprConverter_Convert_ScanWithSeverityData()
+        {
+            RunTest("SdkScan.fpr");
+        }
+
+        [Fact]
+        public void FortifyFprConverter_GetFailureLevelFromRuleMetadata_MissingImpactProperty_ReturnsNone()
```
lgolding: >MissingImpactProperty_ReturnsNone [](start = 72, length = 33)

Thanks for such a well-named test!

I don't agree with what the test says. If `Impact` is missing from a Fortify rule definition, that feels to me like a SARIF rule definition that doesn't specify `defaultConfiguration.level`. So I think it should get the same treatment: it should correspond to a SARIF `result.level` of `"warning"` -- in which case you can omit it from the SARIF `result` object.

src/Sarif.Converters/FortifyFprConverter.cs
lgolding: I don't agree with this. The result would be that if somebody made a mistake in the definition of a Fortify rule, that end users would tend to ignore all results from that rule, even if the rule is pointing out a serious problem! (Automated systems might filter out results with level "none", and then end users might not see them at all!)

- I think you should treat impact > 5.0 as "error" (maybe the rule author forgot that 5.0 was the max, and they turned it up to "11" :-))

- I think you should treat impact < 0.0 as "warning". At least get the result on the end-user's radar.

- If impact is missing, I think you should also treat that as "warning". See my comment in the unit test code for my reasoning.

---
In reply to: [292262904](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292262904) [](ancestors = 292262904)

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -751,7 +796,18 @@ private void ParseRuleInfo()
                             string groupName = _reader.GetAttribute(_strings.NameAttribute);
                             switch (groupName)
                             {
+                                case "Accuracy":
+                                case "Impact":
+                                case "Probability":
+                                {
+                                    _reader.Read();
+                                    string nodeValue = _reader.Value;
+                                    rule.SetProperty<string>(groupName, nodeValue);
```
lgolding: ><string> [](start = 52, length = 8)

You don't have to specify the generic type parameter name. The compiler can deduce it from the type of the argument `nodeValue`, a string.

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
lgolding: Ok, I looked, and I don't agree. See my comments in the converter file and the converter tests file.

---
In reply to: [292523811](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292523811) [](ancestors = 292523811,292263379)

src/ReleaseHistory.md
harleenkohli: you are right! i will rephrase it to mention so

---
In reply to: [292522153](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292522153) [](ancestors = 292522153)

src/Sarif.Converters/FortifyFprConverter.cs
harleenkohli: Thanks Larry! 

---
In reply to: [292551572](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292551572) [](ancestors = 292551572,292262904)

src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/OneResultBasic.sarif
harleenkohli: hah, thanks. i will make the change! 

---
In reply to: [292553684](https://github.com/microsoft/sarif-sdk/pull/1528#discussion_r292553684) [](ancestors = 292553684,292523811,292263379)

harleenkohli:
>           "ruleIndex": 37,

Thanks, i typed a long explanation but then thought, boy this is a side conversation that doesnt really impact my current PR. I will ask you offline.

---
In reply to: [500917257](https://github.com/microsoft/sarif-sdk/pull/1528#issuecomment-500917257) [](ancestors = 500917257,500670733)

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:8 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

harleenkohli:
>                 "Probability": "3.0",

haha, the new values in the property bag! 

---
In reply to: [500905154](https://github.com/microsoft/sarif-sdk/pull/1528#issuecomment-500905154) [](ancestors = 500905154,500671973)

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/FortifyFprConverter/ExpectedOutputs/SdkScan.sarif:47551 in 21b6344. [](commit_id = 21b6344818369a1726db663a8a639f15e9e9696f, deletion_comment = False)

PR #1531, opened 2019-06-12
Fix #1530: CS converter doesn't produce threadFlowLocation.location.

### LINKED ISSUE #1530: CS converter doesn't produce threadFlowLocation.location
When the CS converter translates a sequence of `propagation-event` elements into a `threadFlow`, it populates each `threadFlowLocation.stack` from the `propagation-event`'s `stack` sub-element. The converter also takes the `propagation-event`'s `signature` sub-element and places it at the top of the stack.

But the converter does not populate `threadFlowLocation.location`, which confuses the VS Code Viewer, which expects `location` to be present. Sure, we could "fix" the viewer, but really, `signature` _does_ specify the location, and we _should_ populate it.

@michaelcfanning @rscrivens FYI

lgolding:
When the CS converter translates a sequence of `propagation-event` elements into a `threadFlow`, it populates each `threadFlowLocation.stack` from the `propagation-event`'s `stack` sub-element. The converter also takes the `propagation-event`'s `signature` sub-element and places it at the top of the stack.

But the converter does not populate `threadFlowLocation.location`, which confuses the VS Code Viewer, which expects `location` to be present. Sure, we could "fix" the viewer, but really, `signature` _does_ specify the location, and we _should_ populate it.

While looking at the code I found a couple more issues:
- The WebGoat.xml.sarif test file included `run.language` = `"en-US"`, which is wrong because it's the default. I removed it. I have a fix in another branch for the test infrastructure bug that failed to detect this.
- "Fully qualified logical names" were not being handled consistently. Now, they _never_ include the return type of the method, and the _never_ include the argument signature.  As a result, there are hundreds of line changes in WebGoat.xml.sarif test file. The important changes, for purposes of fixing the bug, is that every `threadFlowLocation` object now has `location` property in addition to the `stack` property.

### DIFF
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -1536,10 +1536,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         private static readonly Regex LogicalLocationRegex =
             new Regex(
-                @"
-                  ([^\s]*\s+)?         # Skip over an optional leading blank-terminated return type name such as 'void '
-                  (?<fqln>[^(]+)       # Take everything up to the opening parenthesis.
-                ",
+                @"^
+                  ([^\s]*\s+)?         # Skip over an optional leading blank-terminated return type name such as 'void '.
+                  (?<fqln>.*)          # Take everything else.
+                $",
                 RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.IgnorePatternWhitespace);
 
         private static string RemoveReturnTypeFrom(string signature)
--- a/src/Sarif.Converters/ContrastSecurityConverter.cs
+++ b/src/Sarif.Converters/ContrastSecurityConverter.cs
@@ -1121,14 +1121,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             };
         }
 
-        private static readonly Regex LogicalLocationRegex =
-            new Regex(
-                @"
-                  ([^\s]*\s+)?         # Skip over an optional leading blank-terminated return type name such as 'void '
-                  (?<fqln>[^(]+)       # Take everything up to the opening parenthesis.
-                ",
-                RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.IgnorePatternWhitespace);
-
         // Find the user code method call closest to the top of the stack. This is
         // the location we should report as being responsible for the result.
         private static string GetUserCodeLocation(Stack stack)
@@ -1138,14 +1130,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             foreach (StackFrame frame in stack.Frames)
             {
                 string fullyQualifiedLogicalName = frame.Location.LogicalLocation.FullyQualifiedName;
-                Match match = LogicalLocationRegex.Match(fullyQualifiedLogicalName);
-                if (match.Success)
+                if (!fullyQualifiedLogicalName.StartsWith(SystemPrefix))
                 {
-                    fullyQualifiedLogicalName = match.Groups["fqln"].Value;
-                    if (!fullyQualifiedLogicalName.StartsWith(SystemPrefix))
-                    {
-                        return fullyQualifiedLogicalName;
-                    }
+                    return fullyQualifiedLogicalName;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


lgolding:

lgolding:

lgolding:
@rscrivens is added to the review. #Closed

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1121,14 +1121,6 @@ private PhysicalLocation CreatePhysicalLocation(string uri, Region region = null
             };
         }
 
-        private static readonly Regex LogicalLocationRegex =
```
lgolding: >LogicalLocationRegex [](start = 38, length = 20)

Moved down, right before the new method `RemoveReturnTypeFrom`.

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1138,14 +1130,9 @@ private static string GetUserCodeLocation(Stack stack)
             foreach (StackFrame frame in stack.Frames)
             {
                 string fullyQualifiedLogicalName = frame.Location.LogicalLocation.FullyQualifiedName;
-                Match match = LogicalLocationRegex.Match(fullyQualifiedLogicalName);
```
lgolding: >fullyQualifiedLogicalName [](start = 57, length = 25)

There's no need to strip off the return type here, in this special case, because now we do it on _every_ stack frame when we create it (see below).

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -1617,6 +1625,8 @@ private static void ReadStack(SparseReader reader, object parent)
             Debug.Assert(context.CurrentThreadFlowLocation == null);
             context.CurrentThreadFlowLocation = new ThreadFlowLocation
             {
+                Location = context.Signature.Location,
```
lgolding: >Location [](start = 16, length = 8)

After all the fuss and bother, this one-liner is the fix for #1530. The rest was just making sure the fully qualified logical name was in the right form.

lgolding:
>                       "location": {

The result of the fix: every `threadFlowLocation` object now has a `location` property. Rusty, note that you only get a logical location. The viewer needs to handle that case.

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:134 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = False)

lgolding:
>               "System.Web.HttpRequest.get_Form",

The first of hundreds of places where we stripped off the return type and the method signature from the fully qualified logical name, leaving just the namespace-qualified method name.

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:113 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = False)

michaelcfanning:
>                                   "fullyQualifiedName": "System.Web.Hosting.PipelineRuntime.ProcessRequestNotification()"

why is this a good change? Removing the parens suggests we aren't providing params data, are we not?

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:3049 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = True)

michaelcfanning:
>                             "fullyQualifiedName": "System.String.Concat"

this isn't a fully qualified name, there are many string.concat overloads, so which one is this>? you need to populate this before doing the stack frame stripping, i think

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:12246 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = False)

src/Sarif.Converters/ContrastSecurityConverter.cs
michaelcfanning: i don't like what we've done to FQN, we've elided the information that qualifies one method overload over the other. this isn't heplful and will compromise result matching.

---
In reply to: [293025496](https://github.com/microsoft/sarif-sdk/pull/1531#discussion_r293025496) [](ancestors = 293025496)

lgolding:
>                             "fullyQualifiedName": "System.String.Concat"

It was the same change that stripped off the return type. I'll restore the signature (including empty parens when there are no arguments).

---
In reply to: [501905176](https://github.com/microsoft/sarif-sdk/pull/1531#issuecomment-501905176) [](ancestors = 501905176)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:12246 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = False)

lgolding:
>                                   "fullyQualifiedName": "System.Web.Hosting.PipelineRuntime.ProcessRequestNotification()"

It's not. I got overaggressive when I stripped off the return type. Will fix.

---
In reply to: [501904892](https://github.com/microsoft/sarif-sdk/pull/1531#issuecomment-501904892) [](ancestors = 501904892)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:3049 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = True)

src/Sarif.Converters/ContrastSecurityConverter.cs
lgolding: Same as the other two comments -- I will restore the signature to the FQN.

---
In reply to: [293604799](https://github.com/microsoft/sarif-sdk/pull/1531#discussion_r293604799) [](ancestors = 293604799,293025496)

lgolding:
>               "System.Web.HttpRequest.get_Form",

Per feedback, we now strip just the return type, and leave the argument signature.

---
In reply to: [501370995](https://github.com/microsoft/sarif-sdk/pull/1531#issuecomment-501370995) [](ancestors = 501370995)

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif:113 in 6a4b5d6. [](commit_id = 6a4b5d647f2fa6dc21603bed508fbd1959124380, deletion_comment = False)

PR #1535, opened 2019-06-14
Fix #1529: CS converter emits incomplete artifact objects

### LINKED ISSUE #1529: CS converter emits incomplete artifact objects
The CS converter emits some `artifact` objects with only a `mimeType` property, and others with only a `contents` property. In neither case is `location` present.

lgolding:
There were two problems:

1. The `ContrastSecurityConverter`, alone among all converters, used the method `FileInfoFactory.Create` to populate `run.artifacts`, and it didn't work properly. All other converters populate `run.artifacts` by creating an `AddFileReferencesVisitor`.

    The `FileInfoFactory` class is `internal`, and is used nowhere else in the SDK, so I removed it.

    The `ContrastSecurityConverter` was also the only remaining converter to call `ResultLogJsonWriter.{Open,Write,Close}Results` directly. I modified it so that, like the other converters, it calls `ToolFileConverterBase.PersistResults`, which does all of that _and_ invokes the `AddFileReferencesVisitor`.

2. The method `ConstructFormsWithoutAutocompletePreventionResult` was manually adding `artifact` objects for each `properties` element. It did that in order to take the opportunity to populate `artifact.contents`. But the string the it used to populate `artifact.contents` was _not_, in fact, the entire file contents. It was just a snippet. So the right place to put it is on `result.locations[0].physicalLocation.region.snippet`.

    Now that we don't need to populate `artifact.contents`, we no longer need to manually create an `artifact`, because we can rely on the `AddFileReferencesVisitor` to do it when it encounters the result location.

Also:
- Changed the MIME type for `.ascx` and `.aspx` from `text/xml` to `text/html`. They are HTML pages (or fragments of pages) with embedded directives and code. The markup they contain (even ignoring the directives and code) is not valid XML (for example, `<p>Paragraph 1<p>Paragraph 2`).


lgolding:

lgolding:

lgolding:
@rscrivens is added to the review. #Closed

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -592,24 +576,14 @@ private Result ConstructFormsWithoutAutocompletePreventionResult(ContrastLogRead
...
                 string jsonValue = properties[key];
                 JObject root = JObject.Parse(jsonValue);
-                string html = root["html"].Value<string>();
-                string fileDataKey = "#RuntimeGenerated#=" + key;
+                string snippet = root["html"].Value<string>();
 
-                fileDataKey = key;
-
-                byte[] htmlBytes = Encoding.UTF8.GetBytes(html);
-                string encoded = System.Convert.ToBase64String(htmlBytes);
```
lgolding: >encoded [](start = 23, length = 7)

No need to save the encoded (binary) form of the snippet.

src/ReleaseHistory.md
```diff
@@ -382,7 +382,7 @@
 * BUGFIX: Multitool crashes on launch: Can't find CommandLine.dll. https://github.com/microsoft/sarif-sdk/issues/1487
 
 ## **v2.1.2** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.2) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.2) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.2) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.2)
-* API BREAKING: Change result.logicalLocation to logicalLocations array. https://github.com/oasis-tcs/sarif-spec/issues/414
+* API BREAKING: Change location.logicalLocation to logicalLocations array. https://github.com/oasis-tcs/sarif-spec/issues/414
```
harleenkohli: >location.l [](start = 23, length = 10)

fml, thanks!

src/Sarif.Converters/ContrastSecurityConverter.cs
```diff
@@ -579,6 +577,11 @@ private Result ConstructFormsAuthenticationSSLResult(ContrastLogReader.Context c
 
         private Result ConstructFormsWithoutAutocompletePreventionResult(ContrastLogReader.Context context)
         {
+            // The maximum length of a snippet that Contrast Security embeds in their HTML file.
+            // If the actual <form> element is longer than this, they tack on an ellipsis, which
+            // we don't want to include in the SARIF file.
+            const int MaxSnippetLength = 1000;
```
harleenkohli: >const int MaxSnippetLength = 1000; [](start = 12, length = 34)

does it make sense that this be moved to a private constant at the class level? I know its only being used in this method today, this is mostly for furutre: 

- would be easier to spot and update in future.
- used in other functions if needed.

lgolding: It's a subtle question. The two bullet points are distinct:

To your first point : If this constant continues to be used _only in this method_, and it needs to be updated, then there will be no problem finding it _in this method_.

The real concern is your second point: What if I find I need a similar constant in a different method? Will I remember that it already exists in this method?

To be honest, the answer is "probably not". So the trade-off is between current understandability and future-proofing. There's this principle that says you should reduce the scope of symbols as much as possible to reduce "cognitive load" on the reader. Suppose I define this symbol at the top of the file (500 lines earlier). Then someone reading the file from the top has to hold this symbol in their heads, even though they might never read the one method that uses it. OTOH, someone who reads just this method and encounters a _usage_ of the symbol has to jump _out_ of the method to find out what it means.

So there is definitely a trade-off, but I almost always come down on the side of reduced scope, unless it is clear to me that we're going to use it elsewhere in the immediate future. It's just another example of the YAGNI principle.

---
In reply to: [294483628](https://github.com/microsoft/sarif-sdk/pull/1535#discussion_r294483628) [](ancestors = 294483628)

PR #1536, opened 2019-06-15
Multitool bug: null reference exception during creation of absolute URI for result location

harleenkohli:
fix for multitool issue - null reference exception during creation of absolute URI for result location IF top level original uri base id doesn't have a uri defined.


src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -33,21 +33,30 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
...
+                visitor.Visit(actualLog.Runs[0]);
 
-            var visitor = new InsertOptionalDataVisitor(_currentOptionallyEmittedData);
-            visitor.Visit(actualLog.Runs[0]);
+                // Restore the remanufactured URI so that file diffing matches
+                actualLog.Runs[0].OriginalUriBaseIds["TESTROOT"] = new ArtifactLocation { Uri = originalUri };
+            }
+            else
+            {
+                var visitor = new InsertOptionalDataVisitor(_currentOptionallyEmittedData);
```
harleenkohli: >var visitor = new InsertOptionalDataVisitor(_currentOptionallyEmittedData); [](start = 16, length = 75)

the new test case doesnt actually have access to the original file, so i simply created an else case - i know this code could have been refactored better, i was rabit-holing a bit, so pushing this one in the interest of time. If you have suggestions on improvements, i want to hear :)

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -122,6 +131,13 @@ public void InsertOptionalDataVisitor_PersistsAll()
                 OptionallyEmittedData.FlattenedMessages);
         }
 
+        [Fact]
+        public void InsertOptionalDataVisitorTests_ContextRegionSnippets_DoesNotFail_TopLevelOriginalUriBaseIdUriMissing()
```
harleenkohli: >DoesNotFail [](start = 73, length = 11)

the test exactly replicates the issue we saw with webgoat.xml.sarif.

src/Sarif/Core/ArtifactLocation.cs
```diff
@@ -20,6 +20,11 @@ public partial class ArtifactLocation
         /// <returns></returns>
         public bool TryReconstructAbsoluteUri(IDictionary<string, ArtifactLocation> originalUriBaseIds, out Uri resolvedUri)
         {
+            resolvedUri = null;
+
+            // If this artifactLocation represents a top level originalUriBaseId, then the value of the URI may be absent.
+            if (this.Uri == null) { return false; }
+
             resolvedUri = this.Uri.IsAbsoluteUri ? this.Uri : null;
```
lgolding: >resolvedUri [](start = 12, length = 11)

If `this.Uri.IsAbsoluteUri`, then we are done, regardless of what `originalUriBaseIds` does or does not say. So replace this line with:
```
if (this.Uri.IsAbsoluteUri)
{
    resolvedUri = this.Uri;
    return true;
}
```

lgolding: >resolvedUri [](start = 12, length = 11)

If `this.Uri.IsAbsoluteUri`, then we are done, regardless of what `originalUriBaseIds` does or does not say. So replace this line with:
```
if (this.Uri.IsAbsoluteUri)
{
    resolvedUri = this.Uri;
    return true;
}
```

lgolding:
>             if (originalUriBaseIds == null) { return this.Uri.IsAbsoluteUri; }

If you do what I suggest at line 28, then at this point you know the URI is relative, so if `originalUriBaseIds` is missing, you can just `return false;`.

---
Refers to: src/Sarif/Core/ArtifactLocation.cs:32 in 2b41052. [](commit_id = 2b41052967076fd475c04d2bbabcd4af285759d6, deletion_comment = False)

lgolding:
>             if (!string.IsNullOrEmpty(this.UriBaseId) &&

The problem here is that it only looks up one level. The correct logic would have say "While I haven't yet found an absolute URI, keep looking up the chain." If I have time this weekend, I'll write this out and propose an alternate PR.

---
Refers to: src/Sarif/Core/ArtifactLocation.cs:36 in 2b41052. [](commit_id = 2b41052967076fd475c04d2bbabcd4af285759d6, deletion_comment = False)

lgolding:
See my comments in ArtifactLocation.cs. I'm approving this because it unblocks your test pipeline and allows me to validate the Contrast Security converter output. But really, the existing logic was incomplete, and this whole method needs to be reworked. I'll try to get to this tomorrow.

lgolding:
>     public class ArtifactTests

The file was previously named FileDataTests.cs. I missed the file name change when we did the "file" =>"artifact" changeover (although I did remember to change the class name from `FileDataTests` to `ArtifactTests`.

---
Refers to: src/Test.UnitTests.Sarif/Core/ArtifactTests.cs:15 in ec612b4. [](commit_id = ec612b4d5c6a834f197a6397c20f346eaf73523a, deletion_comment = False)

lgolding:
Ok, I fixed up the logic and added the unit tests. Go ahead and merge when ready.

---
In reply to: [502390338](https://github.com/microsoft/sarif-sdk/pull/1536#issuecomment-502390338) [](ancestors = 502390338)

lgolding:
>             if (!string.IsNullOrEmpty(this.UriBaseId) &&

Rather than pushing an alternate PR, I made the changes on your branch, same PR.

---
In reply to: [502390244](https://github.com/microsoft/sarif-sdk/pull/1536#issuecomment-502390244) [](ancestors = 502390244)

---
Refers to: src/Sarif/Core/ArtifactLocation.cs:36 in 2b41052. [](commit_id = 2b41052967076fd475c04d2bbabcd4af285759d6, deletion_comment = False)

PR #1537, opened 2019-06-19
Introduce web message parsers

lgolding:
We introduce methods to parse `WebRequest` and `WebResponse` objects from the raw web message strings.

Also:
- Add codegen hints to fix bugs in generated `WebRequest` and `WebResponse` objects:
    - `WebRequest.Headers` must be a dictionary.
    - `WebRequest.Parameters` must be a dictionary.
    - `WebResponse.Headers` must be a dictionary.

### DIFF
--- a/src/Sarif.Converters/VisualStudioBuildLogConverter.cs
+++ b/src/Sarif.Converters/VisualStudioBuildLogConverter.cs
@@ -82,7 +82,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
             \s*:\s*
             (?<message>.*)
             $";
-        private static readonly Regex s_errorLineRegex = RegexFromPattern(ErrorLinePattern);
+        private static readonly Regex s_errorLineRegex = SarifUtilities.RegexFromPattern(ErrorLinePattern);
 
         private Result GetResultFrom(string fullMessage)
         {
@@ -170,23 +170,23 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
 
         // (startLine)
         private const string StartLineStartOnlyPattern = @"^(?<startLine>\d+)$";
-        private static readonly Regex s_startLineOnlyRegex = RegexFromPattern(StartLineStartOnlyPattern);
+        private static readonly Regex s_startLineOnlyRegex = SarifUtilities.RegexFromPattern(StartLineStartOnlyPattern);
 
         // (startLine-endLine)
         private const string StartLineEndLinePattern = @"^(?<startLine>\d+)-(?<endLine>\d+)$";
-        private static readonly Regex s_startLineEndLineRegex = RegexFromPattern(StartLineEndLinePattern);
+        private static readonly Regex s_startLineEndLineRegex = SarifUtilities.RegexFromPattern(StartLineEndLinePattern);
 
         // (startLine,startColumn)
         private const string StartLineStartColumnPattern = @"^(?<startLine>\d+),(?<startColumn>\d+)$";
-        private static readonly Regex s_startLineStartColumnRegex = RegexFromPattern(StartLineStartColumnPattern);
+        private static readonly Regex s_startLineStartColumnRegex = SarifUtilities.RegexFromPattern(StartLineStartColumnPattern);
 
         // (startLine,startColumn-endColumn)
         private const string StartLineStartColumEndColumnPattern = @"^(?<startLine>\d+),(?<startColumn>\d+)-(?<endColumn>\d+)$";
-        private static readonly Regex s_startLineStartColumnEndColumnRegex = RegexFromPattern(StartLineStartColumEndColumnPattern);
+        private static readonly Regex s_startLineStartColumnEndColumnRegex = SarifUtilities.RegexFromPattern(StartLineStartColumEndColumnPattern);
 
         // (startLine,startColumn,endLine,endColumn)
         private const string StartLineStartColumEndLineEndColumnPattern = @"^(?<startLine>\d+),(?<startColumn>\d+),(?<endLine>\d+),(?<endColumn>\d+)$";
-        private static readonly Regex s_startLineStartColumnEndLineEndColumnRegex = RegexFromPattern(StartLineStartColumEndLineEndColumnPattern);
+        private static readonly Regex s_startLineStartColumnEndLineEndColumnRegex = SarifUtilities.RegexFromPattern(StartLineStartColumEndLineEndColumnPattern);
 
         private static Region GetRegionFrom(string regionString)
         {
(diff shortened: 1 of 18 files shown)


src/Sarif/SarifUtilities.cs
```diff
@@ -181,6 +181,11 @@ public static int GetByteLength(string s, Encoding encoding)
             return bytes.Length;
         }
 
+        public static Regex RegexFromPattern(string pattern) =>
```
lgolding: >RegexFromPattern [](start = 28, length = 16)

Generally useful function pulled out of the `VisualStudioBuildLogConverter`.

src/Sarif.Converters/VisualStudioBuildLogConverter.cs
```diff
@@ -278,10 +278,5 @@ private static Region GetRegionFrom(string regionString)
 
             return region;
         }
-
-        private static Regex RegexFromPattern(string pattern) =>
```
lgolding: >RegexFromPattern [](start = 29, length = 16)

Now used in multiple places; moved to `SarifUtilities`.

src/Test.UnitTests.Sarif/Core/ArtifactTests.cs
```diff
@@ -5,12 +5,13 @@
...
 using System.Text;
 using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.CodeAnalysis.Sarif.Writers;
 using Moq;
 using Newtonsoft.Json;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Sarif
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Core
```
lgolding: >Core [](start = 54, length = 4)

The unit tests are in a hodge-podge of namespaces, which makes it hard to find them in the VS test explorer. Since I let VS choose the namespace for the test classes I introduced in the PR, I took the opportunity to move several other test classes into the same default namespace.

src/Test.UnitTests.Sarif/Core/RunTests.cs
```diff
@@ -75,8 +76,10 @@ public void Run_RetrievesExistingFileDataObject()
 
             // Now use a file location that has no url and therefore relies strictly on
             // the index in run.artifacts (i.e., _fileToIndexMap will not be used).
-            fileLocation = new ArtifactLocation();
-            fileLocation.Index = 0;
+            fileLocation = new ArtifactLocation
+            {
+                Index = 0
+            };
```
lgolding: >} [](start = 12, length = 1)

Satisfying an IDE note.

src/Sarif/Core/WebMessageUtilities.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// spent.
+    /// </remarks>
+    internal static class WebMessageUtilities
+    {
+        // When including the contents of a web request or response in an exception message, truncate
+        // it to this length:
+        private const int MaxExceptionMessageStringLength = 200;
+
+        internal const string CRLF = "\r\n";
+        internal const string TokenPattern = "[!#$%&'*+._`|~0-9a-zA-Z^-]+";
```
lgolding: >internal [](start = 8, length = 8)

The patterns are based on the grammar in RFC 7230, "Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing".

lgolding:

lgolding:

src/Sarif/Core/WebResponse.cs
```diff
@@ -0,0 +1,87 @@
...
+    public partial class WebResponse
+    {
+        [JsonIgnore]
+        public string HttpVersion { get; private set; }
+
+        public static WebResponse Parse(string responseString)
+        {
+            var webResponse = new WebResponse();
+
+            ParseStatusLine(
```
harleenkohli: >ParseStatusLine [](start = 12, length = 15)

why not parse the string directly into the webResponse object and return?

harleenkohli: (+ similar pattern in webRequest class)

---
In reply to: [295566231](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295566231) [](ancestors = 295566231)

src/Sarif/Core/WebMessageUtilities.cs
harleenkohli: thanks! could we mention so in a comment in code?

---
In reply to: [295491099](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295491099) [](ancestors = 295491099)

src/Sarif/Core/WebResponse.cs
lgolding: You are right. Code is *so* much cleaner that way! Great comment!

---
In reply to: [295566746](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295566746) [](ancestors = 295566746,295566231)

src/Sarif/Core/WebMessageUtilities.cs
lgolding: Yes, will do. In fact whenever I add a PR comment I ask myself "Would this be better as a code comment?" -- and I missed this one. Good find.

---
In reply to: [295567359](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295567359) [](ancestors = 295567359,295491099)

src/Test.UnitTests.Sarif/Core/ArtifactTests.cs
harleenkohli: thanks this makes sense. i think the mismatched namespaces were a result of these files moving around at a point. thanks for fixing this, the test explorer bothered me too!

---
In reply to: [295397930](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295397930) [](ancestors = 295397930)

src/Sarif/Core/WebResponse.cs
lgolding: Saved 45 lines of code.

---
In reply to: [295568125](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295568125) [](ancestors = 295568125,295566746,295566231)

src/Sarif/Core/WebMessageUtilities.cs
lgolding:
```
        // These patterns are taken from the grammar defined in RFC 7230, "Hypertext Transfer Protocol (HTTP/1.1):
        // Message Syntax and Routing". Yes, tokens (such as message header field names) really can have all those
        // weird characters.
        //
        // We are a little looser than the RFC in a couple of places. In the request, we don't verify that the
        // "target" field on the request line is a valid URI. Nor do we verify that the request body or response
        // body contains only "VCHAR"s (visible characters as opposed to control characters), as required by
        // the RFC.
```

---
In reply to: [295568281](https://github.com/microsoft/sarif-sdk/pull/1537#discussion_r295568281) [](ancestors = 295568281,295567359,295491099)

PR #1539, opened 2019-06-19
Result Matching v2

ScottLouvau:
- Added Result Matching v2, unit tests, and Multitool end-to-end test mode (result-matching-test).
- Tweaked ExtractedResult and IResultMatcher interfaces so V2 could use them.
- Left old Result Matching in case users need old algorithm (add an option and plumb through to the boolean argument on GetDefaultResultMatchingBaseliner if needed).

Sorry, big review. Many changed files are refactoring, but WhatComparer, WhereComparer, and V2ResultMatcher are big.

See the algorithm description in the internal OneNote.

Use the Multitool mode to test quickly against large numbers of real examples.


src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -15,7 +12,65 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
...
+        }
+
+        /// <summary>
+        ///  Match the 'Category' of two ExtractedResults (Tool and RuleId)
+        /// </summary>
+        /// <param name="other">ExtractedResult to match</param>
+        /// <returns>True if Category is identical, False otherwise</returns>
+        public bool MatchesCategory(ExtractedResult other)
+        {
+            return this.RuleId == other.RuleId;
```
rtaket: >return [](start = 12, length = 6)

Is this also supposed to check the Tool?

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -15,7 +12,65 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
...
+            // Tool (checked on Runs beforehand)
+            // Consider: CorrelationGuid, Level, Rank?
+        }
+
+        /// <summary>
+        ///  Match the 'What' properties of two ExtractedResults (Fingerprint and Snippet)
+        /// </summary>
+        /// <param name="other">ExtractedResult to match</param>
+        /// <returns>True if *any* 'What' property matches, False otherwise</returns>
+        public bool MatchesWhat(ExtractedResult other)
```
rtaket: >MatchesWhat [](start = 20, length = 11)

It would be nice if the name of MatchesWhat and MatchesWhere indicated if it used 'any' or 'all' semantics.

rtaket:
>             }

Do you want to add the previous GUID to the absent result?

---
Refers to: src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs:92 in 0dd4711. [](commit_id = 0dd4711297f2ffd622193793ef68201bcca82205, deletion_comment = False)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -14,14 +14,20 @@ public class MatchedResults
     {
         private const string MatchResultMetadata_RunKeyName = "Run";
         private const string MatchResultMetadata_FoundDateName = "FoundDate";
+        private const string MatchResultMetadata_PreviousGuid = "PreviousGuid";
 
         public ExtractedResult PreviousResult { get; set; }
```
rtaket: >set [](start = 53, length = 3)

Since you added a constructor that sets PreviousResult and CurrentResult, should the setters be removed? It seems like an odd pattern that this class can basically be 'reused' by re-setting PreviousResult and CurrentResult and then calling one of the Construct() methods. Or is the intent that the instance be resued?

src/Sarif.Multitool/ResultMatchingTestOptions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    [Verb("result-matching-test", HelpText = "Match Results across runs in the same Group in a folder.")]
```
rtaket: >Verb [](start = 5, length = 4)

I believe the Verb attribute has a property named Hidden prevents the verb from being displayed in the command line help, while still being callable.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            MatchingIndexFromBefore = new int[Before.Count];
+            Fill(MatchingIndexFromBefore, -1);
+
+            MatchingIndexFromAfter = new int[After.Count];
+            Fill(MatchingIndexFromAfter, -1);
+        }
+
+        public IList<MatchedResults> Match()
+        {
+            if (Before.Count == 1 && After.Count == 1)
```
rtaket: >if [](start = 12, length = 2)

Is this common enough to special case?

src/Sarif/Baseline/V2/WhereComparer.cs
```diff
@@ -0,0 +1,197 @@
...
+        public static int CompareTo(IList<Location> left, Run leftRun, IList<Location> right, Run rightRun)
+        {
+            if (left == null && right == null) { return 0; }
+            if (left == null) { return -1; }
+            if (right == null) { return 1; }
+
+            int commonLength = Math.Min(left.Count, right.Count);
+            for (int i = 0; i < commonLength; ++i)
+            {
+                int cmp = CompareTo(left[i], leftRun, right[i], rightRun);
```
rtaket: >left [](start = 36, length = 4)

Are the Left and Right lists sorted? i.e. doesn't the comparison rely on the lists being sorted?

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            // At this point, no high confidence properties matched or failed to match
+            string leftMessage = left.Result.GetMessageText(left.Result.GetRule(left.OriginalRun));
+            string rightMessage = right.Result.GetMessageText(right.Result.GetRule(right.OriginalRun));
+
+            string leftSnippet = GetFirstSnippet(left);
+            string rightSnippet = GetFirstSnippet(right);
+
+            return string.Equals(leftMessage, rightMessage) && string.Equals(leftSnippet, rightSnippet);
```
rtaket: >string.Equals(leftMessage, rightMessage) && string.Equals(leftSnippet, rightSnippet); [](start = 19, length = 85)

I wouldn't have expected the message and snippets to be in the what comparison.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: +1

---
In reply to: [295535056](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r295535056) [](ancestors = 295535056)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -160,5 +171,21 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
...
+                return $"= {PreviousResult}";
+            }
+            else if (PreviousResult == null)
+            {
+                return $"+ {CurrentResult}";
+            }
+            else
+            {
+                return $"- {PreviousResult}";
+            }
```
lgolding: >} [](start = 12, length = 1)

This suggests that the ctor should verify that at least one of `current` and `previous` is non-`null`.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -69,7 +75,7 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
         }
 
         private Result ConstructAbsentResult(
-            Dictionary<string, object> ResultMatchingProperties, 
+            Dictionary<string, object> ResultMatchingProperties,
```
lgolding: >R [](start = 39, length = 1)

camelCase for parameter names.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
ScottLouvau: Tool is part of the category, but SarifLogMatcher only compares runs which have the same tool, so it's intentionally omitted. I made the comment clearer.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
ScottLouvau: Good idea - renamed to "MatchesAllWhere" and "MatchesAnyWhat".

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
ScottLouvau: Yeah, that's a good idea. Fixed.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
ScottLouvau: Good idea. Added.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
ScottLouvau: Fixed all of these for this class.

src/Sarif.Multitool/ResultMatchingTestOptions.cs
ScottLouvau: After more thought, I think we might want this mode for real use, so I renamed it "match-result-set" and am going to leave it visible.

ScottLouvau: [ex: Get Clean scan comparisons over commit history]

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: Just one result is pretty common in Nathan's examples.
The issue is that if there's only one result, the "match before" and "match after" logic doesn't fire for them, so this is the only way they'll actually get compared.

src/Sarif/Baseline/V2/WhatComparer.cs
ScottLouvau: Yeah, some Result properties could arguably belong to "what" or "where". I wanted to restrict where to the location and not the content, so that very simple renames or code changes would still result in a where match. The Message may be static for the rule, but will sometimes describe what is wrong (this class didn't properly implement interface ISomethingOrOther), in which case it would describe what the problem is. The Snippet hopefully includes all of the code encompassing the problem, when it's not a "this is missing" problem, so it also can potentially describe what the problem is.

I do think the "what" and "where" sets and the strictness of matching we require is something we'll likely iterate on. For example, if the line numbers stay but the column numbers move (format document indenting), should we still consider two Results a where match? Probably.

src/Sarif/Baseline/V2/WhereComparer.cs
ScottLouvau: Hmm. They aren't, and they should be. I don't know how to easily sort them, because an IComparer<Location> doesn't have the accompanying run to look up locations which just have an index. I think we need to figure out how to deal with that in the OM before we can really improve this.

ScottLouvau:
> > ```
> >         }
> > ```
> 
> Do you want to add the previous GUID to the absent result?
> 
> Refers to: src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs:92 in [0dd4711](https://github.com/microsoft/sarif-sdk/commit/0dd4711297f2ffd622193793ef68201bcca82205). [](commit_id = [0dd4711](https://github.com/microsoft/sarif-sdk/commit/0dd4711297f2ffd622193793ef68201bcca82205), deletion_comment = False)

The GUID only needs to be copied when the result is unchanged, because then the result in the output log is a copy of the newer one with the newer GUID. If the result is new or absent, the GUID on the result is the GUID from the unmatched original result either way.

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                string fileName = Path.GetFileName(filePath);
+                string currentGroup = GetGroupName(fileName);
+
+                try
+                {
+                    currentLog = ReadSarifFile<SarifLog>(_fileSystem, filePath);
+
+                    // Compare each log with the previous one in the same group
+                    if (currentGroup.Equals(previousGroup) && currentLog?.Runs[0].Results.Count != 0 && previousLog?.Runs[0].Results.Count != 0)
```
lgolding: >currentLog?.Runs[0].Results.Count != 0 [](start = 62, length = 38)

Is this test right? If the previous run had results but the current one doesn't, don't we want to produce a log where all the results are marked "absent"? If the current run has results but the previous one didn't, don't we want to produce a log where all the results are marked "new"?

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+                        {
+                            string outputFilePath = Path.Combine(options.OutputFolderPath, fileName);
+                            WriteSarifFile(_fileSystem, result, outputFilePath, formatting);
+                        }
+                    }
+                }
+                catch (Exception ex) when (!Debugger.IsAttached)
+                {
+                    Console.WriteLine(ex.ToString());
+                }
```
lgolding: >} [](start = 16, length = 1)

Suppose we have MyGroup-001.sarif, MyGroup-002.sarif, MyGroup-003.sarif. We read 001 successfully, but we throw an exception while attempting to process 002 and compare it to 001. Now, says the code, we keep going and attempt to compare 003 to 002. Is that the right thing to do?

Alternatively, we could:
- Give up entirely.
- Give up on 002, but now try to recover by comparing 003 to 001.

In the latter case, we would put lines 75-77 below _inside_ the `try` block, so we only advance the "previous log" if we processed it successfully.

Thoughts?

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                try
+                {
+                    currentLog = ReadSarifFile<SarifLog>(_fileSystem, filePath);
+
+                    // Compare each log with the previous one in the same group
+                    if (currentGroup.Equals(previousGroup) && currentLog?.Runs[0].Results.Count != 0 && previousLog?.Runs[0].Results.Count != 0)
+                    {
+                        Console.WriteLine();
+                        Console.WriteLine($"{previousFileName} -> {fileName}:");
```
lgolding: >-> [](start = 63, length = 2)

I'm not wild about this console output, which suggests that we're transforming the previous file into the current file. Should we say instead: `"Comparing current file '{fileName}' to previous file '{previousFileName}'..."`?

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+                        SarifLog result = matcher.Match(new[] { previousLog }, new[] { currentLog }).First();
```
lgolding: >result [](start = 33, length = 6)

Mostly I won't comment on the nits I find -- I'll just fix them and raise a PR -- but here I'll point out that "result" is an unfortunate variable name in a SARIF context, where the word means something specific. In my PR I propose `comparisonResultLog`; feel free to suggest other names.

lgolding: Oh, I see that when he passes `result` to `WriteDifferences`, the parameter name in `WriteDifferences` is "`mergedLog`". I like that. I'll use it.

---
In reply to: [296050568](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r296050568) [](ancestors = 296050568)

src/Sarif.Multitool/ResultMatchSetCommand.cs
lgolding: (Of course if this is the first run, we want to leave it alone.)

---
In reply to: [296047688](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r296047688) [](ancestors = 296047688)

lgolding:
>         public ReportingDescriptor GetRule(Run run)

I've never reviewed this code, so I can't vouch for its correctness. I can say that if it _is_ correct, its usage in `ResolvedRuleId` above is correct. :-)

---
Refers to: src/Sarif/Core/Result.cs:31 in 38873d1. [](commit_id = 38873d1bcbf35ce9cc616e161d47b216dc243ce6, deletion_comment = False)

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -15,7 +12,68 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
...
+        /// </summary>
+        /// <param name="other">ExtractedResult to match</param>
+        /// <returns>True if *any* 'What' property matches, False otherwise</returns>
+        public bool MatchesAnyWhat(ExtractedResult other)
+        {
+            return WhatComparer.MatchesWhat(this, other);
+        }
+
+        /// <summary>
+        ///  Match all of the 'Where' properties of two ExtractedResults (FileUri, StartLine/Column, EndLine/Column)
```
lgolding: >Column [](start = 109, length = 6)

Region matching is complicated. In addition to being specified by line/column properties, a region in a text file can also be specified by `CharOffset`+`CharLength`, and a region in a binary file is specified by `ByteOffset` +`ByteLength`. I will check the `WhereComparer` to make sure it does the right comparison.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: Actually, if you do the assignment in the ctor, you don't even need the `private set`. I'll take care of that in my "nit fix" PR.

---
In reply to: [295888531](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r295888531) [](ancestors = 295888531)

src/Sarif/Baseline/V2/V2ResultMatcher.cs
lgolding: I added a comment explaining this.

---
In reply to: [295892427](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r295892427) [](ancestors = 295892427)

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
lgolding: Verified that the `WhereComparer` compares all appropriate `Region` properties. Will clarify this comment.

---
In reply to: [296319818](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r296319818) [](ancestors = 296319818)

src/Sarif/Baseline/V2/WhatComparer.cs
lgolding: I agree with Scott that snippet and message are more appropriately "what" than "where".

---
In reply to: [295894088](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r295894088) [](ancestors = 295894088)

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Sarif.Baseline
+{
+    internal class WhatComponent : IEquatable<WhatComponent>
+    {
+        public string Category { get; set; }
```
lgolding: >set; [](start = 38, length = 4)

Setters aren't needed since the properties are initialized in the ctor.

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return Equals(this as WhatComponent);
+        }
+
+        public override int GetHashCode()
+        {
+            int hashCode = 0;
+
+            hashCode *= 13;
+            hashCode = this.Category?.GetHashCode() ?? 0;
```
lgolding: >0 [](start = 55, length = 1)

Bug: We keep resetting `hashCode` to 0 when we encounter a `null` property. In my PR, I reimplemented this to match the (generated) `GetHashCode` implementations in the OM classes.

src/Test.UnitTests.Sarif/TestData/Baseline/elfie-arriba.sarif
```diff
@@ -0,0 +1,237 @@
...
+          ],
+          "partialFingerprints": {
+            "SecretHash/v1": "nHX0oMUmpq3IgCxodLjnV4F3ipb+Sj3SazMpzYQNZ9Y="
+          },
+          "properties": {
+            "matchState": "NotSuppressed"
+          }
+        }
+      ],
+      "automationDetails": {},
```
lgolding: >"automationDetails": {}, [](start = 6, length = 24)

The SDK (and all SARIF producers over which we have control) avoids emitting objects with no properties. How was this file generated?

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            // Match Guid
+            if (left.Result.Guid != null)
+            {
+                if (string.Equals(left.Result.Guid, right.Result.Guid))
+                {
+                    return true;
+                }
+
+                // Non-Match doesn't force false, as GUIDs are often re-generated by tools.
```
lgolding: >GUIDs are often re-generated by tools [](start = 53, length = 37)

I suppose we can't stop tools from doing that, but they're not supposed to. A GUID is supposed to identify an "instance" of a result for all time.

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            // Match Properties (50% must match)
+            if (left.Result.Properties != null && right.Result.Properties != null)
+            {
+                int correspondingPropertyCount = 0;
+                int matchCount = 0;
+
+                foreach (var property in left.Result.Properties)
+                {
+                    if (left.Result.TryGetProperty(property.Key, out string leftValue)
```
lgolding: >out string leftValue [](start = 65, length = 20)

This looks wrong -- property bag elements don't have to be strings; they can be any object. Let me look into this...

lgolding: Yes, this is a bug. If the property bag has an entry with the specified key, but the corresponding value is not a string, the SDK will `throw new InvalidOperationException(SdkResources.CallGenericGetProperty)`.

I have to think about what to do here.

---
In reply to: [296411642](https://github.com/microsoft/sarif-sdk/pull/1539#discussion_r296411642) [](ancestors = 296411642)

src/Test.UnitTests.Sarif/Baseline2/WhereComparerTests.cs
```diff
@@ -0,0 +1,190 @@
...
+            Assert.True(WhereComparer.CompareTo(left, right) < 0);
+
+            // Right EndColumn > Left EndColumn
+            right = new Region(left) { EndColumn = left.EndColumn + 1 };
+            Assert.True(WhereComparer.CompareTo(left, right) < 0);
+
+            left.ByteOffset = 100;
+            left.ByteLength = 10;
+
+            // Right Offset bigger, supercedes others
```
lgolding: >supercedes [](start = 36, length = 10)

It's not really a matter of superseding. The spec distinguishes three types of region property sets:
- line/column properties
- character offset/length properties
- byte offset/length properties

At least one set must be present. If more than one set is present, the **MUST**  refer to exactly the same range of bytes in the file. So you are entitled to look at any set you wish, and if another set gives a different answer, that's just too bad; the SARIF file is invalid.

So here, you are perfectly entitled to look at the byte offset/length properties to decide which region is "smaller", but the fact that this works, given that your test `Region` object is inconsistent, is an artifact of the fact that you happen to look at those properties first.

As long as you keep your implementation the same there's nothing wrong with this.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+
+        public void LinkAdjacentSimilarResults()
+        {
+            // Walk down, matching similar, previously unlinked Results after already linked pairs
+            for (int beforeIndex = 0; beforeIndex < Before.Count - 1; ++beforeIndex)
+            {
+                int afterIndex = MatchingIndexFromBefore[beforeIndex];
+                if (afterIndex == -1 || afterIndex + 1 >= After.Count) { continue; }
```
lgolding: >afterIndex + 1 >= After.Count [](start = 40, length = 29)

How could this happen?

PR #1540, opened 2019-06-21
Fix schema cross-references in external property files schema.

lgolding:
When we updated our SDK schema version numbers from `rtm.1` to `rtm.2`, we missed the cross-references from the external properties file schema back to the log file schema.

There's no need to update the release history: this is not a fix from a previous SDK version; it's the correction of incomplete work we did within the current version.


lgolding:

lgolding:

PR #1542, opened 2019-06-21
Fix nits in VS result matching

lgolding:
This PR is a result of my taking a long time to study the result matching code. I find I absorb code best when I literally have my hands on it, so I took the opportunity to fix nits such as:

- Removing unused private methods.
- Improving (IMHO) some variable and method names.
- Adding additional null conditional operators, to harden and in some cases to simplify the code.
- Additional uses of LINQ to eliminate hard-coded loops.
- Renaming files to match the classes they contain.
- Putting user-facing strings in resources.
- Improving unit testability by adding methods to injected interfaces.
- Putting periods at the ends of sentences.

I'm not quite done studying yet, so be aware I might still push more changes.

@ScottLouvau -- I added you just because it seemed so wrong not to, but I don't expect you to look. Just a courtesy in case you want to see who's scribbling on your code.

@rtaket -- I'd most appreciate your review since you did the most thorough review on Scott's original PR.


src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -57,10 +57,6 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
                 // Baseline result is present, current result is missing => absent.
                 result = ConstructAbsentResult(resultMatchingProperties, out originalResultMatchingProperties);
             }
-            else
-            {
-                throw new InvalidOperationException("Cannot generate a result for a new baseline where both results are null.");
-            }
```
lgolding: Not necessary since the ctor ensures this.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -111,13 +107,13 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
...
 
-            if (CurrentResult.OriginalRun.AutomationDetails?.Guid != null)
+            if (CurrentResult.OriginalRun?.AutomationDetails?.Guid != null)
             {
                 resultMatchingProperties.Add(MatchedResults.MatchResultMetadata_RunKeyName, CurrentResult.OriginalRun.AutomationDetails.Guid);
             }
 
             // Potentially temporary -- we persist the "originally found date" forward, and this sets it.
-            if (CurrentResult.OriginalRun?.Invocations?.Any() == true && CurrentResult.OriginalRun.Invocations[0].StartTimeUtc != null)
+            if (CurrentResult.OriginalRun?.Invocations?[0]?.StartTimeUtc != null)
```
lgolding: >?[0]? [](start = 54, length = 5)

Avoids use of LINQ `Any` together with compound Boolean expression.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -161,16 +157,14 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
             Dictionary<string, object> firstPropertyBag,
             Dictionary<string, object> secondPropertyBag)
         {
-            Dictionary<string, object> result = firstPropertyBag;
+            Dictionary<string, object> mergedPropertyBag = firstPropertyBag;
 
-            foreach (string key in secondPropertyBag.Keys)
+            foreach (string key in secondPropertyBag.Keys.Except(firstPropertyBag.Keys))
```
lgolding: >Except [](start = 58, length = 6)

Avoids hand-coded loop.

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -25,16 +26,16 @@ public int Run(ResultMatchSetOptions options)
             options.OutputFolderPath = options.OutputFolderPath ?? Path.Combine(options.FolderPath, "Out");
 
             ISarifLogMatcher matcher = ResultMatchingBaselinerFactory.GetDefaultResultMatchingBaseliner();
-            Formatting formatting = (options.PrettyPrint ? Formatting.Indented : Formatting.None);
+            Formatting formatting = options.PrettyPrint ? Formatting.Indented : Formatting.None;
 
-            // Remove previous results
-            if (Directory.Exists(options.OutputFolderPath))
+            // Remove previous results.
+            if (_fileSystem.DirectoryExists(options.OutputFolderPath))
```
lgolding: >DirectoryExists [](start = 28, length = 15)

New methods in `IFileSystem` improve unit testability -- may as well, since we're already injecting the interface.

src/Sarif/Core/LogicalLocation.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public partial class LogicalLocation
+    {
+        public LogicalLocation Resolve(Run run)
```
lgolding: >Resolve [](start = 31, length = 7)

Moved from `WhereComparer`.

src/Sarif/Core/ArtifactLocation.cs
```diff
@@ -97,5 +97,12 @@ public static ArtifactLocation CreateFromFilesDictionaryKey(string key, string p
                 UriBaseId = uriBaseId
             };
         }
+
+        public ArtifactLocation Resolve(Run run)
```
lgolding: >Resolve [](start = 32, length = 7)

Moved from `WhereComparer`.

src/Sarif/Baseline/V2/WhereComparer.cs
```diff
@@ -159,39 +149,7 @@ public static int CompareTo(Uri left, Uri right)
 
         private static Uri ArtifactUri(ArtifactLocation loc, Run run)
         {
-            return loc?.Uri ?? Resolve(loc, run)?.Uri;
-        }
-
-        private static LogicalLocation Resolve(LogicalLocation loc, Run run)
```
lgolding: >private [](start = 8, length = 7)

Moved the following two generally useful methods into the Core API.

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -42,19 +42,30 @@ public override bool Equals(object obj)
 
         public override int GetHashCode()
         {
-            int hashCode = 0;
+            int hashCode = 17;
 
-            hashCode *= 13;
-            hashCode = this.Category?.GetHashCode() ?? 0;
```
lgolding: >0 [](start = 55, length = 1)

Bug: We were resetting `hashCode` to 0 each time we found a `null` property value.

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -42,19 +42,30 @@ public override bool Equals(object obj)
 
         public override int GetHashCode()
         {
-            int hashCode = 0;
+            int hashCode = 17;
```
lgolding: >17 [](start = 27, length = 2)

Reimplemented to match the (generated) `GetHashCode` implementations in the OM classes.

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -7,10 +7,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline
 {
     internal class WhatComponent : IEquatable<WhatComponent>
     {
-        public string Category { get; set; }
```
lgolding: >set [](start = 38, length = 3)

Setters not needed since these are initialized in the ctor.

lgolding:

lgolding:

lgolding:
@harkohli is added to the review. #Closed

lgolding:

src/Sarif/Baseline/V2/WhereComparer.cs
```diff
@@ -40,21 +40,15 @@ public static int CompareTo(IList<Location> left, Run leftRun, IList<Location> r
...
             if (left == null && right == null) { return 0; }
             if (left == null) { return -1; }
             if (right == null) { return 1; }
 
-            // Compare by Physical Location, if present
-            cmp = CompareTo(left.PhysicalLocation, leftRun, right.PhysicalLocation, rightRun);
-            if (cmp != 0) { return cmp; }
-
-            // Compare by 'primary' Logical Location, if present
-            cmp = CompareTo(left.LogicalLocation, leftRun, right.LogicalLocation, rightRun);
```
lgolding: >LogicalLocation [](start = 33, length = 15)

The scalar `LogicalLocation` property is actually a convenience property that returns `LogicalLocations[0]`. It exists for backward compatibility because we changes `LogicalLocations` to an array at the last minute, and we didn't want it to be a breaking change.

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -33,7 +33,7 @@ public static IEnumerable<WhatComponent> WhatProperties(this ExtractedResult res
             {
                 foreach (var fingerprint in result.Result.Fingerprints)
                 {
-                    yield return new WhatComponent(result.RuleId, "Fingerprint", fingerprint.Key, fingerprint.Value);
+                    yield return new WhatComponent(result.RuleId, "Fingerprint", "Fingerprint/" + fingerprint.Key, fingerprint.Value);
```
lgolding: >Key [](start = 110, length = 3)

In principle, elements of the `Fingerprints`, `PartialFingerprints`, and `Properties` dictionaries might have the same keys, so we disambiguate them.

src/Test.UnitTests.Sarif/Core/PropertyBagHolderTests.cs
```diff
@@ -4,11 +4,12 @@
 using System;
 using System.Collections.Generic;
 using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Sarif
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Core
```
lgolding: >Test [](start = 33, length = 4)

We're standardizing the test namespaces to what VS generates by default when you create a test class. As matters stand, our unit tests are in a hodgepodge of namespaces, which makes them difficult to find in the Test Explorer.

src/Test.UnitTests.Sarif/Baseline2/WhatComparerTests.cs
```diff
@@ -1,64 +1,70 @@
...
-            Result right = new Result() { RuleId = "Rule1", Message = new Message() { Text = "Two" } };
+            Result left = new Result { RuleId = "Rule1", Message = new Message() { Text = "One" } };
+            Result right = new Result { RuleId = "Rule1", Message = new Message() { Text = "Two" } };
             ExtractedResult eLeft = new ExtractedResult(left, null);
             ExtractedResult eRight = new ExtractedResult(right, null);
 
             // GUIDs
             // =====
-            Guid g1 = Guid.NewGuid(), g2 = Guid.NewGuid();
+            string g1 = Guid.NewGuid().ToString(), g2 = Guid.NewGuid().ToString();
```
lgolding: >ToString [](start = 39, length = 8)

Convert them once instead of each time they're used.

src/Test.UnitTests.Sarif/Baseline2/WhatComparerTests.cs
```diff
@@ -69,101 +75,111 @@ public void WhatComparer_Basics()
...
+            WhatComparer.MatchesWhat(eLeft, eRight).Should().BeTrue();
 
-            // Don't match when names are different
+            // Don't match when names are different.
             left.SetProperty("Fifth_Left", "0005");
             right.SetProperty("Fifth_Right", "0005");
-            Assert.False(WhatComparer.MatchesWhat(eLeft, eRight));
+            WhatComparer.MatchesWhat(eLeft, eRight).Should().BeFalse();
+
+            // Match when non-string properties are equal.
```
lgolding: >// Match when non-string properties are equal. [](start = 12, length = 46)

Test the new ability to compare non-string-valued properties. Before I added the new APIs to `PropertyBagHolder`, and modified `WhatComparer` to use them, these two tests would have crashed.

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -1,13 +1,18 @@
...
 using System.Linq;
 using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Baseline;
 using Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching;
 using Newtonsoft.Json;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Sarif.Baseline
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Baseline
```
lgolding: >.Test.UnitTests. [](start = 32, length = 16)

See my comment about namespaces in PropertyBagHolderTests.cs.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -116,10 +116,10 @@ private SarifLog BaselineSarifLogs(IEnumerable<Run> previous, IEnumerable<Run> c
         {
             // Spin out SARIF logs into MatchingResult objects.
             List<ExtractedResult> baselineResults = 
-                previous == null ? new List<ExtractedResult>() : GetMatchingResultsFromRuns(previous);
```
lgolding: >GetMatchingResultsFromRuns [](start = 65, length = 26)

I didn't like this name because we don't know whether the results we extract will match. Another possible name for `ExtractedResult` is `MatchableResult`, and then this method could be called `GetMatchableResultsFromRuns`.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -182,18 +182,6 @@ private List<ExtractedResult> GetMatchingResultsFromRuns(IEnumerable<Run> sarifR
 
             return results;
         }
-        
-        private ReportingDescriptor GetRuleFromResources(Result result, IDictionary<string, ReportingDescriptor> rules)
```
lgolding: >GetRuleFromResources [](start = 36, length = 20)

Unused.

src/Sarif/Baseline/ResultMatching/SarifLogMatcher.cs
```diff
@@ -319,14 +307,6 @@ private ReportingDescriptor GetRuleFromResources(Result result, IDictionary<stri
             };
         }
 
-        private void MergeDictionaryInto<T, S>(
```
lgolding: >MergeDictionaryInto [](start = 21, length = 19)

Unused.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -1,14 +1,17 @@
...
-
 namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
 {
     /// <summary>
-    /// Result metadata for matching a group of results from a 
-    /// set of runs (such as the current set of results) to a group of results 
-    /// from a different set of runs (such as the prior set of results).
+    /// This class associates a Result object with additional metadata that
+    /// makes it possible for a matching algorithm to determine if it matches
+    /// a result from another run.
```
lgolding: >run [](start = 30, length = 3)

As I mentioned in a comment on `SarifLogResultMatcher`, this class could also be called `MatchableResult`, and I think this comment conveys that better.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -12,16 +12,22 @@ public class V2ResultMatcher : IResultMatcher
...
             return state.Match();
         }
     }
 
     /// <summary>
-    ///  ResultMatchState contains all of the state needed to compute matches
+    ///  StatefulResultMatcher contains all of the state needed to compute matches
     ///  between two batches of Results.
     /// </summary>
-    internal class ResultMatchState
```
lgolding: >ResultMatchState [](start = 19, length = 16)

This class isn't a "state", it's a matcher that retains state as it works.

rtaket:
>             result.SetProperty(SarifLogResultMatcher.ResultMatchingResultPropertyName, resultMatchingProperties);

Is i not possible for 'result' to be null here?

---
Refers to: src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs:69 in 7f2bbdd. [](commit_id = 7f2bbdd098a7dd12a2feac74b01ff270cbb67cf4, deletion_comment = False)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
rtaket: Is there a perf loss in the new implementation? It looks like Except might create another list and the keys are enumerated more than once. So if the lists are large, there could be a perf problem. But I'm just guessing and I don't know if perf is a consideration for this method.

---
In reply to: [296356649](https://github.com/microsoft/sarif-sdk/pull/1542#discussion_r296356649) [](ancestors = 296356649)

src/Sarif/FileSystem.cs
```diff
@@ -239,5 +239,32 @@ public void SetAttributes(string path, FileAttributes fileAttributes)
...
+
+        /// <summary>
+        /// Deletes an empty directory from a specified path.
+        /// </summary>
+        /// <param name="path">
+        /// The name of the empty directory to remove. This directory must be writable and empty.
+        /// </param>
+        public void DeleteDirectory(string path, bool recursive = false)
+        {
+            Directory.Delete(path);
```
rtaket: >Directory.Delete(path); [](start = 12, length = 23)

Did you intend to use the 'recursive' input parameter?

src/Sarif/SdkResources.resx
```diff
@@ -243,4 +243,7 @@
   <data name="ERR997_AllRulesExplicitlyDisabled" xml:space="preserve">
     <value>All rules were explicitly disabled so there is no work to do.</value>
   </data>
+  <data name="ErrorMatchedResultsRequiresNonNullResult" xml:space="preserve">
+    <value>MatchedResults requires at least one non-null Result.</value>
```
rtaket: >MatchedResults [](start = 11, length = 14)

I wouldn't have expected this resource/exception string to be in a resx file since I'm not sure an end user would know what to do with it. i.e. it seems like an internal exception. ... I only ask in case there's cost to adding to the resx file; localization, etc.

src/Sarif/FileSystem.cs
lgolding: Oops :-)

---
In reply to: [296901139](https://github.com/microsoft/sarif-sdk/pull/1542#discussion_r296901139) [](ancestors = 296901139)

lgolding:
>             result.SetProperty(SarifLogResultMatcher.ResultMatchingResultPropertyName, resultMatchingProperties);

No, `result` was populated by one of the calls to `Construct{Existing,New,Absent}Result` above, and they construct the `Result` object by deep-cloning an input that is known non-null.

---
In reply to: [505158184](https://github.com/microsoft/sarif-sdk/pull/1542#issuecomment-505158184) [](ancestors = 505158184)

---
Refers to: src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs:69 in 7f2bbdd. [](commit_id = 7f2bbdd098a7dd12a2feac74b01ff270cbb67cf4, deletion_comment = False)

src/Sarif/SdkResources.resx
lgolding: You're right, this one isn't really user facing. I'm not concerned about the runtime cost of extracting a rarely-used string from resources, but there's no sense spending money for a translator to localize it.

---
In reply to: [296904795](https://github.com/microsoft/sarif-sdk/pull/1542#discussion_r296904795) [](ancestors = 296904795)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: Good question. I checked the source code. It creates a `Set` to perform the existence checking, but it only iterates once:

[https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Except.cs](https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Except.cs)



---
In reply to: [296895448](https://github.com/microsoft/sarif-sdk/pull/1542#discussion_r296895448) [](ancestors = 296895448,296356649)

michaelcfanning: Why would we want the set to be created in this case? Is there something that you find objectionable about the original code? Can you articulate it?

---
In reply to: [296915286](https://github.com/microsoft/sarif-sdk/pull/1542#discussion_r296915286) [](ancestors = 296915286,296895448,296356649)

PR #1545, opened 2019-06-24
Fix bugs in WebRequest.Parse related to parsing a URI's query and to header lines.

lgolding:


lgolding:

lgolding:

src/Sarif/Core/WebRequest.cs
```diff
@@ -99,35 +106,38 @@ private static Uri SynthesizeAbsoluteUriFrom(Uri uri)
...
-                )
-              )*";
+                  (?<name>[^=]+)                # parameter name (everything that's not an = sign),
+                  =                             # an equal sign, and
+                  (
+                    (?<value>[^&]*)             # the value (everything that's not an '&')...
+                    &?                          # and an '&' (except after the last one).
+                  )
+                )*
+                |                               # RFC 3986 does _not_ require a URI's query to
```
harleenkohli: >RFC 3986 does _not_ require a [](start = 50, length = 29)

thanks for this comment, seeing this made me go read a bit reg rfc3986, and that made me think, we are currently not handling the possible presence of #<fragment> post the query. do we care?

harleenkohli: Here's the definition for quick reference - i couldnt judge if its a useful parameter for us to store in sarif: The fragment identifier component of a URI allows indirect identification of a secondary resource by reference to a primary resource and additional identifying information.

---
In reply to: [296963117](https://github.com/microsoft/sarif-sdk/pull/1545#discussion_r296963117) [](ancestors = 296963117)

lgolding: That's a good question. The short answer is that RFC 7230 (HTTP) and RFC 3986 (URI) together prohibit a fragment here. The long answer is fun :-) :

RFC 7230 §3.1.1 says that a request message begins with a `request-line`:
```
request-line = method SP request-target SP HTTP-version CRLF
```
For example:
```
GET /hello.txt HTTP/1.1
```
We are concerned with the part in the middle, the `request-target`.

RFC 7230 §5.3 tells us that `request-target` can take one of four forms: `origin-form`, `absolute-form`, `authority-form`, and `asterisk-form`.

§5.3.1 says that `origin-form` is `absolute-path [ "?" query ]`. So, no fragment.

§5.3.2 says that `absolute-form` is `absolute-URI`, which we have to look up in RFC 3986, where we learn that it's `scheme ":" hier-part [ "?" query ]`. Again, no fragment.

§5.3.3 says that `authority-form` is just `authority`, which again we have to look up in RFC 3986, where we learn that it's `[ userinfo "@" ] host [ ":" port ]`. In full generality, it might look like `harleen@www.example.com:7700`. Usually it's just `www.example.com`.

§5.3.4 says that `asterisk-form` (§5.3.4) is just an asterisk: `*`.

---
In reply to: [296964466](https://github.com/microsoft/sarif-sdk/pull/1545#discussion_r296964466) [](ancestors = 296964466,296963117)

src/Sarif/Core/WebMessageUtilities.cs
```diff
@@ -53,8 +53,8 @@ internal static class WebMessageUtilities
         {
             var headers = new Dictionary<string, string>();
             totalLength = 0;
-             
-            do
+
+            while (!requestString.StartsWith(CRLF))     // An empty line signals the end of the headers.
```
lgolding: >while [](start = 12, length = 5)

Fix bug: There don't have to be any header lines.

src/Sarif/Core/WebRequest.cs
```diff
@@ -6,14 +6,27 @@
...
 using System.Text;
 using System.Text.RegularExpressions;
-using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
     public partial class WebRequest
     {
-        [JsonIgnore]
-        public string HttpVersion { get; private set; }
```
lgolding: >HttpVersion [](start = 22, length = 11)

I had added this convenience property, which the `Parse` method populated. I had marked it `JsonIgnore` because we can't serialize it because the SARIF spec doesn't define this property. The result is that when you serialize a `WebRequest` object containing this property, and then you deserialize the resulting string back into a `WebRequest` object, this property is lost. This made it confusing and not useful.

src/Test.UnitTests.Sarif/Core/WebResponseTests.cs
```diff
@@ -120,20 +118,19 @@ public void WebResponse_Parse_ThrowsIfResponseDoesNotEndWithABlankLine()
         }
 
         [Theory]
-        [InlineData("", false, null, null, null, -1, null, -1, "status line is empty")]
```
lgolding: >null [](start = 31, length = 4)

Remove the first `null` argument (corresponding to the now-removed `httpVersion` property) in each data line.

src/Sarif/Core/WebRequest.cs
harleenkohli: wow, thanks for the detailed explanation!

---
In reply to: [297001721](https://github.com/microsoft/sarif-sdk/pull/1545#discussion_r297001721) [](ancestors = 297001721,296964466,296963117)

PR #1547, opened 2019-06-27
Fortify converter ignores originalUriBaseId if root (e.g. C:)

harleenkohli:
in fvdl, if <SourceBasePath> is a root (e.g C:), this is ignored by fortify converter.
    <SourceBasePath>C:</SourceBasePath>

This is because Uri.TryCreate method does not treat "C:" as an absolute URI.

Fixing this by manually appending a "/" . We had a test case which covers this scenario already, hence rebaselined it to new expected result.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -410,4 +410,4 @@
 * Change schemas back to draft-04 to reenable Intellisense in the Visual Studio JSON editor.
 
 ## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
-* BUGFIX: Fortify converter does not populat originalUriBaseId if the Source is root (e.g. C:)
+* BUGFIX: Fortify FPR converter does not populate originalUriBaseIds if the source is a drive letter (e.g. C:)
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -408,3 +408,6 @@
 
 ## **v2.1.5** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.5) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.5) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.5) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.5)
 * Change schemas back to draft-04 to reenable Intellisense in the Visual Studio JSON editor.
+
+## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
+* BUGFIX: Fortify converter does not populat originalUriBaseId if the Source is root (e.g. C:)
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -223,7 +223,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                 if (originalUriBasePath.StartsWith("/") &&
                      platform == "Linux")
                 {
-                    originalUriBasePath = "file:/" + originalUriBasePath;
+                    originalUriBasePath = "file://" + originalUriBasePath;
                 }
 
                 if (!originalUriBasePath.EndsWith("/"))
--- a/src/Sarif.Converters/FortifyFprConverter.cs
+++ b/src/Sarif.Converters/FortifyFprConverter.cs
@@ -17,7 +17,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     internal class FortifyFprConverter : ToolFileConverterBase
     {
         private const string FortifyExecutable = "[REMOVED]insourceanalyzer.exe";
-        private const string FileLocationUriBaseId = "SRCROOT";
+        internal const string FileLocationUriBaseId = "SRCROOT";
         private const string ReplacementTokenFormat = "<Replace key=\"{0}\"/>";
         private const string EmbeddedLinkFormat = "[{0}](1)";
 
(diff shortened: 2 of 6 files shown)


src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -218,6 +218,11 @@ private Run CreateSarifRun()
                     _originalUriBasePath = "file:/" + _originalUriBasePath;
                 }
 
+                if (_originalUriBasePath.EndsWith(":"))
+                {
+                    _originalUriBasePath = _originalUriBasePath + "/";
+                }
```
lgolding: >} [](start = 16, length = 1)

Please add a unit test.

src/ReleaseHistory.md
```diff
@@ -407,4 +407,7 @@
 * API NON-BREAKING: Introduce `WebRequest.TryParse` and `WebResponse.TryParse` to accompany existing `Parse` methods.
 
 ## **v2.1.5** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.5) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.5) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.5) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.5)
-* Change schemas back to draft-04 to reenable Intellisense in the Visual Studio JSON editor.
\ No newline at end of file
+* Change schemas back to draft-04 to reenable Intellisense in the Visual Studio JSON editor.
+
+## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
+* BUGFIX: Fortify converter does not populat originalUriBaseId if the Source is root (e.g. C:)
```
lgolding: >is root [](start = 77, length = 7)

1. There are two Fortify converters, so "Fortify FPR converter" 
2. "is root" => "starts with a drive letter

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -218,6 +218,11 @@ private Run CreateSarifRun()
                     _originalUriBasePath = "file:/" + _originalUriBasePath;
                 }
 
+                if (_originalUriBasePath.EndsWith(":"))
+                {
+                    _originalUriBasePath = _originalUriBasePath + "/";
```
lgolding: >"/" [](start = 66, length = 3)

You know, the spec says (§3.14.14, originalUriBaseIds property):

> The values of the `uri` properties in the `artifactLocation` objects in `originalUriBaseIds`:
>   • SHALL end with a single forward slash.

It's true for _all_ the `uri` values, not just the ones that happen to look like `C:`. So replace these three lines with:
```
if (!originalUriBasePath.EndsWith("/")) { _originalUriBasePath += "/"; }
```

Once you've merged this PR, consider adding a utility method like `SarifUtilities.EnsureTrailingSlash`, and make sure this logic happens _everywhere_ we create a `run.originalUriBaseIds`. Or just file an issue :-)

src/ReleaseHistory.md
harleenkohli: thanks! you are right, i will update this

---
In reply to: [297937009](https://github.com/microsoft/sarif-sdk/pull/1547#discussion_r297937009) [](ancestors = 297937009)

src/Sarif.Converters/FortifyFprConverter.cs
harleenkohli: thanks, updated. and i extracted the method to bring under unit test, this includes linux platform related change as well, so i added that as another test case.

---
In reply to: [297943211](https://github.com/microsoft/sarif-sdk/pull/1547#discussion_r297943211) [](ancestors = 297943211)

harleenkohli: oh! i just noticed the second part of your suggestion reg moving this to a utility, thats a great idea! will add an issue.

---
In reply to: [297959359](https://github.com/microsoft/sarif-sdk/pull/1547#discussion_r297959359) [](ancestors = 297959359,297943211)

src/Test.UnitTests.Sarif.Converters/FortifyFprConverterTests.cs
```diff
@@ -85,7 +85,7 @@ public void FortifyFprConverter_GetFailureLevelFromRuleMetadata_ReturnsAppropria
 
             };
 
-            foreach(KeyValuePair<string,FailureLevel> keyValuePair in expectedInputOutputs)
+            foreach (KeyValuePair<string, FailureLevel> keyValuePair in expectedInputOutputs)
```
lgolding: >  [](start = 19, length = 1)

Thank you :)

src/Test.UnitTests.Sarif.Converters/FortifyFprConverterTests.cs
```diff
@@ -94,7 +94,58 @@ public void FortifyFprConverter_GetFailureLevelFromRuleMetadata_ReturnsAppropria
...
+        }
+
+        [Fact]
+        public void FortifyFprConverter_GetOriginalUriBaseIdsDictionary_sourceIsLinuxStyleAbsolutePath()
+        {
+            Dictionary<string, ArtifactLocation> originalUriBaseIdsDictionary = FortifyFprConverter.GetOriginalUriBaseIdsDictionary("/root/projects/myproject/src/", "Linux");
+
+            originalUriBaseIdsDictionary.Count.Should().Be(1);
+            originalUriBaseIdsDictionary.ContainsKey(FortifyFprConverter.FileLocationUriBaseId).Should().BeTrue();
+            originalUriBaseIdsDictionary[FortifyFprConverter.FileLocationUriBaseId].Uri.Should().Be(@"file://root/projects/myproject/src/");
```
lgolding: >// [](start = 107, length = 2)

Existing bug: this two-slash form isn't right. See [https://tools.ietf.org/html/rfc8089#appendix-B](https://tools.ietf.org/html/rfc8089#appendix-B) (RFC 8089 : The "file" URI Scheme). The URI for a local file is either
```
[file:///path/to/file](file:///path/to/file)
```
or
```
file:/path/to/file
```
With two slashes, the first component is interpreted as the "authority":
```
[file://host.example.com/path/to/file](file://host.example.com/path/to/file)
```
In your example, `[file://root/projects/myproject/src/`](file://root/projects/myproject/src/`) means "the file `/projects/myproject/src/` on host `root`.

harleenkohli: thanks! fixed!

---
In reply to: [298217975](https://github.com/microsoft/sarif-sdk/pull/1547#discussion_r298217975) [](ancestors = 298217975)

PR #1549, opened 2019-06-28
Bug fix: rebaseuri fails for indexed artifacts

harleenkohli:
BUGFIX: Multitool rebaseUri command throws null reference exception if results reference run.artifacts using indexes.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -411,3 +411,4 @@
 
 ## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
 * BUGFIX: Fortify FPR converter does not populate originalUriBaseIds if the source is a drive letter (e.g. C:)
+* BUGFIX: Multitool rebaseUri command throws null reference exception if results reference run.artifacts using indexes.
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -411,4 +411,5 @@
 
 ## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
 * BUGFIX: Fortify FPR converter does not populate originalUriBaseIds if the source is a drive letter (e.g. C:)
-* BUGFIX: Multitool rebaseUri command throws null reference exception if results reference run.artifacts using indexes.
+* BUGFIX: Multitool rebaseUri command throws null reference exception if results reference run.artifacts using the index property.
+* BUGFIX: Pre-release transformer does not upgrade schema uri if input version is higher than rtm.1.
--- a/src/Sarif.Multitool/AbsoluteUriCommand.cs
+++ b/src/Sarif.Multitool/AbsoluteUriCommand.cs
@@ -55,7 +55,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
         private IEnumerable<AbsoluteUriFile> GetSarifFiles(AbsoluteUriOptions absoluteUriOptions)
         {
             // Get files names first, as we may write more sarif logs to the same directory as we rebase them.
-            HashSet<string> fileNames = CreateTargetsSet(absoluteUriOptions.TargetFileSpecifiers, absoluteUriOptions.Recurse);
+            HashSet<string> fileNames = CreateTargetsSet(absoluteUriOptions.TargetFileSpecifiers, absoluteUriOptions.Recurse, _fileSystem);
             foreach (var file in fileNames)
             {
                 yield return new AbsoluteUriFile() { FileName = file, Log = ReadSarifFile<SarifLog>(_fileSystem, file) };
--- a/src/Sarif.Multitool/RebaseUriCommand.cs
+++ b/src/Sarif.Multitool/RebaseUriCommand.cs
@@ -64,10 +64,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
             return 0;
         }
         
-        private IEnumerable<RebaseUriFile> GetSarifFiles(RebaseUriOptions mergeOptions)
+        private IEnumerable<RebaseUriFile> GetSarifFiles(RebaseUriOptions rebaseUriOptions)
         {
             // Get files names first, as we may write more sarif logs to the same directory as we rebase them.
-            HashSet<string> fileNames = CreateTargetsSet(mergeOptions.TargetFileSpecifiers, mergeOptions.Recurse);
+            HashSet<string> fileNames = CreateTargetsSet(rebaseUriOptions.TargetFileSpecifiers, rebaseUriOptions.Recurse, _fileSystem);
             foreach(var file in fileNames)
             {
                 yield return new RebaseUriFile() { FileName = file, Log = ReadSarifFile<SarifLog>(_fileSystem, file) };
--- a/src/Sarif/Visitors/RebaseUriVisitor.cs
+++ b/src/Sarif/Visitors/RebaseUriVisitor.cs
@@ -34,6 +34,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 
         public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
         {
+            if (node.Uri == null)
+            {
+                return node;
+            }
+
             if (_currentRun == null || _currentRun.OriginalUriBaseIds == null || !_currentRun.OriginalUriBaseIds.Values.Contains(node))
             {
                 if (node.Uri.IsAbsoluteUri && _baseUri.IsBaseOf(node.Uri))
--- a/src/Sarif/Visitors/RebaseUriVisitor.cs
+++ b/src/Sarif/Visitors/RebaseUriVisitor.cs
@@ -39,7 +39,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
                 return node;
             }
 
-            if (_currentRun == null || _currentRun.OriginalUriBaseIds == null || !_currentRun.OriginalUriBaseIds.Values.Contains(node))
+            if (_currentRun?.OriginalUriBaseIds?.Values.Contains(node) != true)
             {
                 if (node.Uri.IsAbsoluteUri && _baseUri.IsBaseOf(node.Uri))
                 {
--- a/src/Test.UnitTests.Sarif.Multitool/MemoryStreamToStringBuilder.cs
+++ b/src/Test.UnitTests.Sarif.Multitool/MemoryStreamToStringBuilder.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    class MemoryStreamToStringBuilder : MemoryStream
+    {
+        private StringBuilder OutputTo { get; set; }
+
+        public MemoryStreamToStringBuilder(StringBuilder outputTo)
+        {
+            OutputTo = outputTo;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            OutputTo.Append(Encoding.UTF8.GetString(this.ToArray()));
+            base.Dispose(disposing);
+        }
+    }
+}
--- a/src/Test.UnitTests.Sarif.Multitool/PageCommandTests.cs
+++ b/src/Test.UnitTests.Sarif.Multitool/PageCommandTests.cs
@@ -8,7 +8,7 @@ using Microsoft.CodeAnalysis.Sarif.Multitool;
 using Newtonsoft.Json;
 using Xunit;
 
-namespace Sarif.Multitool.UnitTests
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
 {
     public class PageCommandTests
     {
--- a/src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
+++ b/src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
@@ -1,0 +1,88 @@
+﻿using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    public class RebaseUriCommandTests : FileDiffingUnitTests
+    {
+        private static ResourceExtractor Extractor = new ResourceExtractor(typeof(RebaseUriCommandTests));
+
+        public RebaseUriCommandTests(ITestOutputHelper outputHelper) : base(outputHelper) { }
+
+        [Fact]
+        public void RebaseUriCommand_RebaseRunWithArtifacts()
+        {
+            string testFilePath = "RunWithArtifacts.sarif";
+
+            var options = new RebaseUriOptions
+            {
+                BasePath = @"C:\vs\src\2\s\",
+                BasePathToken = "SRCROOT",
+                Inline = true,
+                SarifOutputVersion = SarifVersion.Current,
+                PrettyPrint = true
+            };
+
+            RunAndValidateRebaseUriCommand(testFilePath, options);
+        }
+
+        private void RunAndValidateRebaseUriCommand(string testFilePath, RebaseUriOptions options)
+        {
+            string inputSarifLog = Extractor.GetResourceText($"RebaseUriCommand.Inputs.{testFilePath}");
+            string expectedOutput = Extractor.GetResourceText($"RebaseUriCommand.ExpectedOutputs.{testFilePath}");
+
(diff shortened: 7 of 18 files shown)


src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -34,6 +34,11 @@ public RebaseUriVisitor(string uriBaseId, Uri baseUri, bool rebaseRelativeUris =
 
         public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
         {
+            if (node.Uri == null)
+            {
+                return node;
+            }
+
```
harleenkohli: the "fix". Everything else you see in this pr is tests and related infra

harleenkohli: more explanation: ArtifactLocation could be just have a single property "index" wich indexes into run.artifacts.

---
In reply to: [298444180](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298444180) [](ancestors = 298444180)

src/Test.UnitTests.Sarif.Multitool/TransformCommandTests.cs
```diff
@@ -180,21 +180,5 @@ private static string RunTransformationCore(string logFileContents, SarifVersion
...
-            {
-                OutputTo = outputTo;
-            }
-
-            protected override void Dispose(bool disposing)
-            {
-                OutputTo.Append(Encoding.UTF8.GetString(this.ToArray()));
-                base.Dispose(disposing);
-            }
-        }
```
harleenkohli: moved this out to its own class!

src/Test.UnitTests.Sarif.Multitool/MemoryStreamToStringBuilder.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    class MemoryStreamToStringBuilder : MemoryStream
```
harleenkohli: >lass MemoryStreamToStringBuilder : MemoryStream [](start = 5, length = 47)

moved out of TransformCommandTests.cs to allow reusablity

src/ReleaseHistory.md
```diff
@@ -410,4 +410,5 @@
 * Change schemas back to draft-04 to reenable Intellisense in the Visual Studio JSON editor.
 
 ## **v2.1.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.6)
-* BUGFIX: Fortify FPR converter does not populate originalUriBaseIds if the source is a drive letter (e.g. C:)
\ No newline at end of file
+* BUGFIX: Fortify FPR converter does not populate originalUriBaseIds if the source is a drive letter (e.g. C:)
+* BUGFIX: Multitool rebaseUri command throws null reference exception if results reference run.artifacts using indexes.
```
lgolding: >indexes [](start = 111, length = 7)

"indexes" => "indices" if you want the English plural, or "the index property" (my preference).

src/Sarif/Visitors/RebaseUriVisitor.cs
```diff
@@ -34,6 +34,11 @@ public RebaseUriVisitor(string uriBaseId, Uri baseUri, bool rebaseRelativeUris =
...
             if (_currentRun == null || _currentRun.OriginalUriBaseIds == null || !_currentRun.OriginalUriBaseIds.Values.Contains(node))
```
lgolding: >if [](start = 12, length = 2)

With the C# 6 null conditional operator, this simplifies to
```
if (_currentRun?.OriginalUriBaseIds?.Values.Contains(node) != true)
```
The explicit test is necessary because the return type of the operator is `bool?` (`Nullable<bool>`), and there's no implicit conversion from `bool?` to the `bool` that the `if` conditional requires.

lgolding:
>             newRun.OriginalUriBaseIds = newRun.OriginalUriBaseIds ?? new Dictionary<string, ArtifactLocation>();

Wouldn't it be nice if there were a `??=` operator! Then we'd have
```
newRun.OriginalUriBaseIds ??= new Dictionary<string, ArtifactLocation>();
```
But there isn't.

---
Refers to: src/Sarif/Visitors/RebaseUriVisitor.cs:64 in c350778. [](commit_id = c350778f63b7fca10dc70a1c1daaed834faa77ed, deletion_comment = False)

lgolding:
>             newRun.OriginalUriBaseIds = newRun.OriginalUriBaseIds ?? new Dictionary<string, ArtifactLocation>();

I submitted it as a C# language suggestion :) [https://github.com/dotnet/csharplang/issues/2621](https://github.com/dotnet/csharplang/issues/2621)

---
In reply to: [506787061](https://github.com/microsoft/sarif-sdk/pull/1549#issuecomment-506787061) [](ancestors = 506787061)

---
Refers to: src/Sarif/Visitors/RebaseUriVisitor.cs:64 in c350778. [](commit_id = c350778f63b7fca10dc70a1c1daaed834faa77ed, deletion_comment = False)

src/Sarif/Visitors/RebaseUriVisitor.cs
lgolding: That's so bizarre. I wasted time because CodeFlow colored everything from here down to the end of the file. It's just these three lines that are new. Change looks good.

---
In reply to: [298444310](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298444310) [](ancestors = 298444310,298444180)

src/Sarif.Multitool/RebaseUriCommand.cs
```diff
@@ -64,10 +64,10 @@ public int Run(RebaseUriOptions rebaseOptions)
             return 0;
         }
         
-        private IEnumerable<RebaseUriFile> GetSarifFiles(RebaseUriOptions mergeOptions)
+        private IEnumerable<RebaseUriFile> GetSarifFiles(RebaseUriOptions rebaseUriOptions)
```
lgolding: >rebaseUriOptions [](start = 74, length = 16)

LOL. Copy/paste FTL.

src/Sarif.Multitool/AbsoluteUriCommand.cs
```diff
@@ -55,7 +55,7 @@ public int Run(AbsoluteUriOptions absoluteUriOptions)
         private IEnumerable<AbsoluteUriFile> GetSarifFiles(AbsoluteUriOptions absoluteUriOptions)
         {
             // Get files names first, as we may write more sarif logs to the same directory as we rebase them.
-            HashSet<string> fileNames = CreateTargetsSet(absoluteUriOptions.TargetFileSpecifiers, absoluteUriOptions.Recurse);
+            HashSet<string> fileNames = CreateTargetsSet(absoluteUriOptions.TargetFileSpecifiers, absoluteUriOptions.Recurse, _fileSystem);
```
lgolding: >_fileSystem [](start = 126, length = 11)

It's such a pity that every derived command class accepts, and stores separately, a `FileSystem` object, and then has to pass it down to a bunch of `static` methods on the base class.

What we should have done was (using `RebaseUriCommand` as an example):
```
internal class RebaseUriCommand : CommandBase
{
    public RebaseUriCommand(IFileSystem fileSystem = null) : base(fileSystem)
    {
        ...
    }

    public int Run(RebaseUriOptions rebaseOptions)
    {
        ...
        //  NO NEED TO PASS _fileSystem DOWN TO THE BASE CLASS.
        WriteSarifFile(sarifLog.Log, outputName, formatting);
    }

    ...
}

public abstract class CommandBase
{
    // STORED ONCE, IN BASE CLASS; AVAILABLE TO DERIVED CLASSES IF NEEDED
    protected IFileSystem FileSystem { get; }

    public CommandBase(IFileSystem fileSystem)
    {
        FileSystem = fileSystem;
    }

    // NO LONGER TAKES AN IFileSystem ARGUMENT; NO LONGER STATIC.
    public T ReadSarifFile<T>(string filePath, IContractResolver contractResolver = null)
    {
        var serializer = new JsonSerializer() { ContractResolver = contractResolver };

        // USES THE NEW FileSystem PROPERTY.
        using (JsonTextReader reader = new JsonTextReader(new StreamReader(FileSystem.OpenRead(filePath))))
        {
            return serializer.Deserialize<T>(reader);
        }
    }
    ...
}
```

Of course it would be a breaking change to remove the existing public static methods on the base class. But we could keep the static methods, _add_ new same-named instance methods, and have the instance methods delegate to the static methods.

I don't recommend we do this now. Now that I've thought it through, I'll file an issue.

src/Test.UnitTests.Sarif.Multitool/PageCommandTests.cs
```diff
@@ -8,7 +8,7 @@
 using Newtonsoft.Json;
 using Xunit;
 
-namespace Sarif.Multitool.UnitTests
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
```
lgolding: >Multitool [](start = 54, length = 9)

Yay!

lgolding:
> using FluentAssertions;

Add copyright notice.

---
Refers to: src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs:1 in c350778. [](commit_id = c350778f63b7fca10dc70a1c1daaed834faa77ed, deletion_comment = False)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
```diff
@@ -0,0 +1,88 @@
+﻿using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
```
lgolding: System usings at the top (our convention, that's all).

lgolding:
>             newRun.OriginalUriBaseIds = newRun.OriginalUriBaseIds ?? new Dictionary<string, ArtifactLocation>();

Hah! It already exists in C# 8 (currently in preview). [https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8)

---
In reply to: [506788487](https://github.com/microsoft/sarif-sdk/pull/1549#issuecomment-506788487) [](ancestors = 506788487,506787061)

---
Refers to: src/Sarif/Visitors/RebaseUriVisitor.cs:64 in c350778. [](commit_id = c350778f63b7fca10dc70a1c1daaed834faa77ed, deletion_comment = False)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            Assert.True(
+                AreEquivalent<SarifLog>(actualOutput, expectedOutput),
+                GetErrorMessageWithDiffFileDetails(testFilePath, expectedOutput, actualOutput)
+                );
+        }
+
+        private static string GetErrorMessageWithDiffFileDetails(string testFilePath, string expectedOutput, string actualOutput)
+        {
+            // TODO: Rebaseline test output functionality is not integrated for these tests yet.
```
lgolding: >yet [](start = 92, length = 3)

Is this a big job? It seems like something that should be part of this PR. If you add it, it will be easy to review, just as the diff between this iteration and the next.

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            var rebaseUriCommand = new RebaseUriCommand(mockFileSystem.Object);
+
+            int returnCode = rebaseUriCommand.Run(options);
+            string actualOutput = transformedContents.ToString();
+
+            returnCode.Should().Be(0);
+
+            Assert.True(
+                AreEquivalent<SarifLog>(actualOutput, expectedOutput),
+                GetErrorMessageWithDiffFileDetails(testFilePath, expectedOutput, actualOutput)
```
lgolding: >GetErrorMessageWithDiffFileDetails [](start = 16, length = 34)

This will be called even in the happy path where the test passes, which means we'll write two files to the file system even when the test passes. Is this a consequence of not having the rebaseline functionality? If so, it strengthens my suggestion (below) that you add rebaselining in this PR.

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    public class RebaseUriCommandTests : FileDiffingUnitTests
```
lgolding: Wow, the rebaselineuri command had no tests before! Glad we're adding them.

src/Sarif.Multitool/AbsoluteUriCommand.cs
lgolding: I filed [https://github.com/microsoft/sarif-sdk/issues/1550](https://github.com/microsoft/sarif-sdk/issues/1550), "Store IFileSystem object in protected property of CommandBase".

---
In reply to: [298665212](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298665212) [](ancestors = 298665212)

harleenkohli:
>             newRun.OriginalUriBaseIds = newRun.OriginalUriBaseIds ?? new Dictionary<string, ArtifactLocation>();

wow!! i love it! also TIL how to submit suggestions to c# people!

---
In reply to: [506799521](https://github.com/microsoft/sarif-sdk/pull/1549#issuecomment-506799521) [](ancestors = 506799521,506788487,506787061)

---
Refers to: src/Sarif/Visitors/RebaseUriVisitor.cs:64 in c350778. [](commit_id = c350778f63b7fca10dc70a1c1daaed834faa77ed, deletion_comment = False)

src/Sarif/Visitors/RebaseUriVisitor.cs
harleenkohli: oh! strange! thanks though for bonus inputs :) i will make the changes

---
In reply to: [298658405](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298658405) [](ancestors = 298658405,298444310,298444180)

harleenkohli: PS: i see the strange codeflow behaviour as well (will show it to Birendra)

---
In reply to: [298689012](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298689012) [](ancestors = 298689012,298658405,298444310,298444180)

src/Test.UnitTests.Sarif.Multitool/PageCommandTests.cs
harleenkohli: i knew you'd like this!

---
In reply to: [298666308](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298666308) [](ancestors = 298666308)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
harleenkohli: Si! i will further refine this in future to see if we can generalize this to run file diffing tests for all commands, i see this need. we dont have such tests for other commands that i have been using e.g. AbsoluteUri.

---
In reply to: [298673342](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298673342) [](ancestors = 298673342)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
harleenkohli: i completely agree!! let me give it another shot today with a fresh mind!

---
In reply to: [298672159](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298672159) [](ancestors = 298672159)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
harleenkohli: oh, thats good feedback, now that i think about it yes - it does perform the unnecessary steps. This statement is gone because we are now using the base method for this goodness, but i will keep this is in mind for future.

---
In reply to: [298673152](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298673152) [](ancestors = 298673152)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
harleenkohli: done! all i need was to override the function which reads the expected output file. I did have to go change its access modifier though 

---
In reply to: [298692734](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298692734) [](ancestors = 298692734,298672159)

lgolding: Very nice solution.

---
In reply to: [298764040](https://github.com/microsoft/sarif-sdk/pull/1549#discussion_r298764040) [](ancestors = 298764040,298692734,298672159)

src/Test.UnitTests.Sarif.Multitool/RebaseUriCommandTests.cs
```diff
@@ -0,0 +1,82 @@
...
+                BasePathToken = "SRCROOT",
+                Inline = true,
+                SarifOutputVersion = SarifVersion.Current,
+                PrettyPrint = true
+            };
+
+            RunTest(testFilePath);
+        }
+
+        protected override string ConstructTestOutputFromInputResource(string testFilePath)
```
lgolding: >ConstructTestOutputFromInputResource [](start = 34, length = 36)

Right, so when you use this, the base class calls this overridden method and then executes the comparisons.

Issue #1554, opened 2019-06-30
[eslint-formatter-sarif] Artifacts without messages should be included in the artifacts element as well


mhuguet:
Currently the logic only includes artifacts that have messages. According to the ESLint documentation for [custom formatters](https://eslint.org/docs/developer-guide/working-with-custom-formatters), the results object can contain a child result with a null messages child. The list of artifacts should include all files included in the analysis, not just those with messages.
```javascript
if (result.messages.length > 0) {
    if (typeof sarifFiles[result.filePath] === "undefined") {

        let contentsUtf8;

        // Create a new entry in the files dictionary.
        sarifFiles[result.filePath] = {
            artifactLocation: {
                uri: result.filePath
            }
        };
```

Issue #1553, opened 2019-06-30
[eslint-formatter-sarif] Update to the latest version of the schema (currently emitting 2.1.0-beta.2)


mhuguet:
This also includes artifact.artifactLocation being replaced with artifact.location

Issue #1552, opened 2019-06-30
[eslint-formatter-sarif] Required property 'name' not found in JSON. Path 'runs[0].tool.driver'


mhuguet:
Encountered this exception when attempting to open a generated SARIF file from ESLint. It is currently generating an older version of the format and after choosing to upgrade the format to a temporary location this occurred. Need to update the custom formatter to include the required properties for the current version of the standard.

**Exception with call stack:**
Sarif upgrade failed with error:
        Newtonsoft.Json.JsonSerializationException: Required property 'name' not found in JSON. Path 'runs[0].tool.driver', line 87, position 9.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EndProcessProperty(Object newObject, JsonReader reader, JsonObjectContract contract, Int32 initialDepth, JsonProperty property, PropertyPresence presence, Boolean setDefaultValue)
...

PR #1555, opened 2019-06-30
Ensured tool driver name is always set, updated to sarif-2.1.0-rtm.1, alwa…

mhuguet:
…ys including artifacts regardless of whether or not it has a message or not


mhuguet:
fix #1553
fix #1552 
fix #1554

Issue #1557, opened 2019-07-01
SarifCurrentToVersionOneVisitor emits an empty run.stableId


lgolding:
Urgent. This breaks an internal customer. @michaelcfanning @harleenkohli FYI.

Issue #1556, opened 2019-07-01
SarifCurrentToVersionOneVisitor drops partialFingerprints


lgolding:
Urgent. This breaks an internal customer. @michaelcfanning @harleenkohli FYI.

PR #1558, opened 2019-07-01
Fix bugs in SarifCurrentToVersionOneVisitor

lgolding:
- #1556: `SarifCurrentToVersionOneVisitor` drops partial fingerprints.
- #1557: `SarifCurrentToVersionOneVisitor` drops `run.id` and emits empty `run.stableId`.

### DIFF
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -75,7 +75,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (!instanceId.Contains("/"))
             {
-                return String.Empty;
+                return null;
             }
 
             return instanceId.Substring(0, instanceId.LastIndexOf('/'));
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -45,9 +45,8 @@
     <Content Include="Schemata\sarif-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
     <Content Include="Schemata\sarif-external-property-file-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
 
-    <!-- Link=false: Don't display this item in Solution Explorer. -->
     <!-- This item will be added to the NuGet package, however. -->
-    <Content Include="..\ReleaseHistory.md" Link="false" />
+    <Content Include="..\ReleaseHistory.md" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
+++ b/src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
@@ -901,7 +901,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 
                     run.BaselineId = v2Run.BaselineGuid;
                     run.Files = CreateFileDataVersionOneDictionary();
-                    run.Id = v2Run.AutomationDetails?.Guid;
+                    run.Id = v2Run.AutomationDetails?.Id;
                     run.AutomationId = v2Run.RunAggregates?.FirstOrDefault()?.Id;
 
                     run.StableId = v2Run.AutomationDetails?.InstanceIdLogicalComponent();
--- a/src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
+++ b/src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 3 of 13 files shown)


lgolding:

lgolding:

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -886,7 +901,7 @@ internal RunVersionOne CreateRunVersionOne(Run v2Run)
 
                     run.BaselineId = v2Run.BaselineGuid;
                     run.Files = CreateFileDataVersionOneDictionary();
-                    run.Id = v2Run.AutomationDetails?.Guid;
+                    run.Id = v2Run.AutomationDetails?.Id;
```
lgolding: >Id [](start = 54, length = 2)

I don't think any other customer is both producing v1 and using automation details. I think this is the right value for `run.Id` given the decision below for `run.stableId`.

src/Sarif/ExtensionMethods.cs
```diff
@@ -75,7 +75,7 @@ public static string InstanceIdLogicalComponent(this RunAutomationDetails runAut
 
             if (!instanceId.Contains("/"))
             {
-                return String.Empty;
+                return null;
```
lgolding: >null [](start = 23, length = 4)

This is what cause the empty `stableId` that our customer saw. Just omit it.

src/Sarif/Sarif.csproj
```diff
@@ -45,9 +45,8 @@
     <Content Include="Schemata\sarif-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
     <Content Include="Schemata\sarif-external-property-file-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
 
-    <!-- Link=false: Don't display this item in Solution Explorer. -->
     <!-- This item will be added to the NuGet package, however. -->
-    <Content Include="..\ReleaseHistory.md" Link="false" />
```
lgolding: >false [](start = 50, length = 5)

The effect of this was that when you edited ReleaseHistory.md, the tab in the VS editor showed the file name "false".

harleenkohli: oh wow, thanks for fixing this!

---
In reply to: [299229228](https://github.com/microsoft/sarif-sdk/pull/1558#discussion_r299229228) [](ancestors = 299229228)

src/Sarif/ExtensionMethods.cs
michaelcfanning: When are we going to get started on unit testing? This is a spot fix, I'd expect a spot unit test. Using the functional test harness is a getting a bit heavyweight for this sort of detail bug fix. For this case, for example, we could have simply tested/documented the purpose and behavior of this method in a dedicated unit test. Something to keep in mind moving forward. I don't want to see us add multiple files of tests for spot fixes, this gets very burdensome over time.

---
In reply to: [299228991](https://github.com/microsoft/sarif-sdk/pull/1558#discussion_r299228991) [](ancestors = 299228991)

src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/ExpectedOutputs/OneRunWithAutomationDetails.sarif
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-1.0.0.json",
+  "version": "1.0.0",
+  "runs": [
```
michaelcfanning: >runs [](start = 3, length = 4)

i think i would have investigated introducing these patterns into some other test file rather than creating a detail test file for these fixes. just a gut call.

src/Test.UnitTests.Sarif/TestData/SarifCurrentToVersionOneVisitor/Inputs/ResultWithTwoPartialFingerprints.sarif
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.3",
+  "version": "2.1.0",
+  "runs": [
```
michaelcfanning: >runs [](start = 3, length = 4)

again, wouldn't have created individual files for these, would minimally have tended to merge the cases with one or two partial fingerprints. not sure how it helps to have them factored differently. the value of knowing the exact condition that failed (because it is encoded in the file name) is not worth the direction we're headed towards: analyzing hundreds of files in order to validate lots of detailed runtime conditions.

PR #1560, opened 2019-07-02
upgrading (insertoptionaldata) coretests file to rtm.3

harleenkohli:


harleenkohli:
this was only a test pull request to repro an appveyor specifc issue

PR #1562, opened 2019-07-02
Fix sample how to write a SARIF file

pascalberger:
Fix invalid syntax in sample how to write a SARIF file.


harleenkohli:
Thanks very much for your contribution @pascalberger

PR #1564, opened 2019-07-05
Fix result matching bug

lgolding:
TODO: Once we understand it, file it.


lgolding:
We concluded that there is no bug. The customer was using the result matching API in a way that wasn't supported (by providing multiple baseline runs from the same tool on the same code base). The customer now understands the correct usage pattern.

PR #1565, opened 2019-07-08
Schema: add missing additionalProperties: false constraints

lgolding:
Also:
- Add missing object descriptions and improve others.

These changes are per oasis-tcs/sarif-spec#420.

**UPDATE** The TC approved the changes so I've marked this PR as "Ready for review".

### DIFF
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -63,38 +63,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 {
                     switch (schemaSubVersion)
                     {
-                        case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json":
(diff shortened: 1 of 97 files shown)


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -63,9 +63,10 @@ public static class PrereleaseCompatibilityTransformer
                 {
                     switch (schemaSubVersion)
                     {
+                        case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json":
                         case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.3.json":
```
harleenkohli: >case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.3.json": [](start = 24, length = 81)

this case will require a slight modification in schema ==> which is just that we need to upgrade the schema version from rtm.3 to rtm.4.

this is a recent bug we had for rtm.2 and 1 as well, if you see the code in function ApplyRtm2andRtm3Changes, thats all it does. In retrospect, i should have named that function better so that spotting this potential bug is obvious.

PR #1570, opened 2019-07-11
Fix #1571: FxCop converter doesn't emit run.logicalLocations + add validation rule

### LINKED ISSUE #1571: FxCop converter produces logicalLocation.index but not run.logicalLocations
@harleenkohli discovered an FxCop SARIF file where `theResult.locations[0].logicalLocations[0].index` was populated, but `theRun.logicalLocations` was absent. This is invalid by §3.33.3:

>If `thisObject` is an element of `theRun.logicalLocations`, then `index` **MAY** be present. ...
>
> Otherwise, if `theRun.logicalLocations` is absent, or if it does not contain a cached object for `thisObject`, then `index` **SHALL NOT** be present.

lgolding:
@harleenkohli discovered an FxCop SARIF file where `theResult.locations[0].logicalLocations[0].index` was populated, but `theRun.logicalLocations` was absent. This is invalid by §3.33.3:

>If `thisObject` is an element of `theRun.logicalLocations`, then `index` **MAY** be present. ...
>
> Otherwise, if `theRun.logicalLocations` is absent, or if it does not contain a cached object for `thisObject`, then `index` **SHALL NOT** be present.

The reason is that the FxCop converter accumulates the list of logical locations but then never writes them out. It's a one-line fix.

Since this is an in-the-wild occurrence of invalid SARIF that we did not detect, I added a validation rule. **The rule is complicated because there are _many_ places we need to do similar array index validations.**

Notes:
- The validation rule needs to know what run it's looking at, so I added `SarifValidationContext.CurrentRun`.

Also:
- If there is neither physical nor logical location available information, don't emit `result.locations`. This happens, for example, on an engine exception (CA0001) or an assembly load error (CA0055).
- To aid debugging, implement `Region.ToString` (returns, for example, `"(24,10)"`).
- Update `SarifValidationSkimmerBase.HelpUri` to always point to the latest spec.
- Prevent the validator from emitting empty rule help messages.


lgolding:

lgolding:

src/Sarif/RuleUtilities.cs
```diff
@@ -32,7 +33,7 @@ public static Result BuildResult(FailureLevel level, IAnalysisContext context, R
             return BuildResult(level, kind, context, region, ruleMessageId, arguments);
         }
 
-        public static Result BuildResult(FailureLevel level, ResultKind kind, IAnalysisContext context, Region region, string ruleMessageId, params string[] arguments)
+        public static Result BuildResult(FailureLevel level, ResultKind kind, IAnalysisContext context, Region region, string ruleMessageId, params object[] arguments)
```
michaelcfanning: >object [](start = 148, length = 6)

Not sure I like this change. it's preferable to require callers to convert to strings, so that they can apply culture insensitive/sensitive transformations during ToString. we can't make an appropriate choice from the object.ToString() helper.

src/Sarif.Multitool/Rules/RuleId.cs
```diff
@@ -16,5 +16,6 @@ public static class RuleId
         public const string UriBaseIdRequiresRelativeUri = "SARIF1014";
         public const string UriMustBeAbsolute = "SARIF1015";
         public const string ContextRegionRequiresRegion = "SARIF1016";
+        public const string InvalidLogicalLocationIndex = "SARIF1017";
```
michaelcfanning: >InvalidLogicalLocationIndex [](start = 28, length = 27)

As always, I'm concerned this rule is overly granular, create one rule for any invalid index.

lgolding: Working on it. There are a total of 15 array-index-valued properties, so it will take a couple of hours:

- `logicalLocation.index`
- `logicalLocation.parentIndex`
- `artifactLocation.index`
- `artifact.parentIndex`
- `result.ruleIndex`
- `address.index`
- `address.parentIndex`
- `threadFlowLocation.index`
- `graphTraversal.resultGraphIndex`
- `graphTraversal.runGraphIndex`
- `webRequest.index`
- `webResponse.index`
- `resultProvenance.invocationIndex`
- `reportingDescriptorReference.index`
- `toolComponentReference.index`

---
In reply to: [303553749](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r303553749) [](ancestors = 303553749)

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -138,6 +140,11 @@ internal Result CreateResult(FxCopLogReader.Context context)
                 sourceFile = string.IsNullOrWhiteSpace(sourceFile) ? targetFile : sourceFile;
             }
 
+            // Don't emit a location if neither physical location nor logical location information
+            // is present. This is the case for CA0001, (unexpected error in analysis tool).
```
michaelcfanning: >, [](start = 54, length = 1)

remove this comma?

src/Sarif.Converters/FxCopConverter.cs
```diff
@@ -138,6 +140,11 @@ internal Result CreateResult(FxCopLogReader.Context context)
...
+            // https://docs.microsoft.com/en-us/visualstudio/code-quality/ca0001?view=vs-2019
+            bool emitLocation = false;
```
michaelcfanning: >false [](start = 32, length = 5)

NIT you could make this assignment the return value of !string.IsNullOrWhitespace. Then below, simply test 'emitLocation' to enter the conditional. Eliminates the second assignment to emitLocation.

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
```diff
@@ -0,0 +1,209 @@
...
+            ValidateArrayIndex(
+                graphTraversal.ResultGraphIndex,
+                Context.CurrentResult.Graphs,
+                graphTraversalPointer,
+                "graphTraversal",
+                "resultGraphIndex",
+                $"runs[{Context.CurrentRunIndex}].results[{Context.CurrentResultIndex}].graphTraversals");
+        }
+
+        protected override void Analyze(LogicalLocation logicalLocation, string logicalLocationPointer)
```
michaelcfanning: >LogicalLocation [](start = 40, length = 15)

where's the check for logicalLocation.parentIndex?

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
```diff
@@ -0,0 +1,209 @@
...
+                Context.CurrentRun.Artifacts,
+                artifactLocationPointer,
+                "artifactLocation",
+                "index",
+                $"runs[{Context.CurrentRunIndex}].artifacts");
+        }
+
+        protected override void Analyze(GraphTraversal graphTraversal, string graphTraversalPointer)
+        {
+            ValidateArrayIndex(
```
michaelcfanning: >ValidateArrayIndex [](start = 12, length = 18)

please double check you've covered all 15 identified cases. Am I correct that I find only 13 here? logicalLocations.parentIndex certainly seems missing, can't find the other, assuming there is another.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -59,42 +56,53 @@ protected void LogResult(string jPointer, string formatId, params string[] args)
...
 
         protected virtual void Analyze(Graph graph, string graphPointer)
         {
         }
 
+        protected virtual void Analyze(GraphTraversal graphTraversal, string graphTraversalPointer)
+        {
+        }
+
         protected virtual void Analyze(Invocation invocation, string invocationPointer)
```
michaelcfanning: >Analyze [](start = 31, length = 7)

If we were ninja, there would be a test that reflects over this type and parses the schema and attempts to ensure that we find an Analyze method for all types. Looks like graphTraversal was missing for example.

michaelcfanning: Format is complete now, so not sure it's worth the time to stop. If we touch this file again, I'll do it.:)

---
In reply to: [305429964](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305429964) [](ancestors = 305429964)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -186,7 +214,7 @@ internal static string JsonPointerToJavaScript(string pointerString)
             var pointer = new JsonPointer(pointerString);
             foreach (string token in pointer.ReferenceTokens)
             {
-                if (int.TryParse(token, out int index))
+                if (int.TryParse(token, out _))
```
michaelcfanning: >_ [](start = 44, length = 1)

NIT can you use 'unusedIndex' or something rather than an underscore? i find it jarring to see this where I expect a variable name. I understand you're trying to indicate it is not consumed.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -554,6 +662,36 @@ private void Visit(Region region, string regionPointer)
...
+                Visit(reportingDescriptor.ShortDescription, reportingDescriptorPointer.AtProperty(SarifPropertyName.ShortDescription));
+            }
+
+            if (reportingDescriptor.FullDescription != null)
+            {
+                Visit(reportingDescriptor.FullDescription, reportingDescriptorPointer.AtProperty(SarifPropertyName.FullDescription));
+            }
+        }
+
+        private void Visit(ReportingDescriptorReference reportingDescriptorReference, string reportingDescriptorReferencePointer)
```
michaelcfanning: >Visit [](start = 21, length = 5)

this code never should have reauthored a comprehensive visitor in the SDK but should have used our rewriting visitor. all this churn emphasizes the point to me.

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
lgolding: I have not finished all the cases yet. And some of the cases have sub-cases: for instance, you have to validate `ReportingDescriptorReference.Index`, but `ReportingDescriptorReferences` occur in several different contexts -- and you need tests for all of them to ensure you've got error message arguments right in each case.

---
In reply to: [305428389](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305428389) [](ancestors = 305428389)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
lgolding: Actually this is a language feature. "`_`" is special -- note how I didn't have to specify the type of the output parameter.

---
In reply to: [305430798](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305430798) [](ancestors = 305430798)

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
lgolding: Absolutely. I have a list of cases to cover. Here's the current status (which I think includes work done after you did your review):
```
[ ] One rule for invalid index
    [x] logicalLocation.parentIndex
    [x] artifactLocation.index
    [x] artifact.parentIndex
    [x] result.ruleIndex
    [x] address.index
    [x] address.parentIndex
    [x] threadFlowLocation.index
    [x] graphTraversal.resultGraphIndex
    [x] graphTraversal.runGraphIndex
    [x] webRequest.index
        [x] result.webRequest
        [x] threadFlowLocation.webRequest
    [x] webResponse.index
        [x] result.webResponse
        [x] threadFlowLocation.webResponse
    [x] resultProvenance.invocationIndex
    [ ] reportingDescriptorReference.index
        [x] result.rule
          [ ] Cover the case where there is a toolComponent reference that affects the array this points into.
        [x] configurationOverride.descriptor
            [x] invocation.ruleConfigurationOverrides
            [x] invocation.notificationConfigurationOverrides
        [ ] reportingDescriptorRelationship.target
        [ ] notification.descriptor
        [ ] notification.associatedRule
        [ ] Use in Taxa
        [ ] Use in ReportingDescriptorRelationship
    [ ] toolComponentReference.index
        [ ] toolComponent.associatedComponent
        [ ] reportingDescriptorReference.toolComponent
```

---
In reply to: [305429284](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305429284) [](ancestors = 305429284)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
lgolding: I don't remember how we got to this state. I think I discovered at some point that the generated `RewritingVisitor` wasn't fit for this purpose, so I started work in a branch (which I still have hanging around called) `users/lgolding/generate-visitor`, where I was going to generate _this_ visitor. But I never got time to finish it, and it turns out there are aspects of this code that are hard to auto-generate. I'm not opposed to taking another look, but boy, I can't imagine this rising to the top of my stack.

---
In reply to: [305431532](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305431532) [](ancestors = 305431532)

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
lgolding: Having said that, I just looked at my checklist (I pasted it into another comment), and I see that I put an `x` in the `logicalLocation.parentIndex` box -- and that was a mistake. Good catch! I'll uncheck it and implement it (and then re-check it ;-)).

---
In reply to: [305525258](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305525258) [](ancestors = 305525258,305428389)

src/Sarif.Multitool/Rules/SARIF1017.InvalidIndex.cs
michaelcfanning: Great. Resetting my review to 'reviewing' pending your completion of the work. Carry on! Good stuff.

---
In reply to: [305525771](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305525771) [](ancestors = 305525771,305429284)

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
lgolding: There are places in this code where I have to set context to establish whether I'm looking at a notification descriptor, or a taxon descriptor, in order to produce the correct error message. (See, for example, the method `Visit(Notification, string)` in this class, and in SARIF1017.InvalidIndex.cs, see the overload of `Analyze` that takes a `ReportingDescriptorReference`.)

That would make it hard to auto-generate this class.

---
In reply to: [305526382](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305526382) [](ancestors = 305526382,305431532)

src/ReleaseHistory.md
```diff
@@ -420,4 +420,8 @@
...
-* BUGFIX: Add missing `"additionalProperties": false` constraints to schema; add missing object descriptions and improve other object descriptions in schema; update schema version to -rtm.4.
\ No newline at end of file
+* BUGFIX: Add missing `"additionalProperties": false` constraints to schema; add missing object descriptions and improve other object descriptions in schema; update schema version to -rtm.4.
+
+## **v2.1.9** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.9) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.9) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.9) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.9)
+* BUGFIX: FxCop converter produced logicalLocation.index but did not produce the run.logicalLocations array. https://github.com/microsoft/sarif-sdk/issues/1571
+* FEATURE: Add validation rule to ensure that `logicalLocation.index` is not present unless `run.logicalLocations` is present.
```
harleenkohli: >`logicalLocation.index` is not present unless `run.logicalLocations` [](start = 46, length = 68)

this text can be changed to cover all index based validations

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
harleenkohli: aah, beautiful!! TIL. I always found it annoying thing about TryParses! most of the times we dont care about the out param, so this makes me very happy! :)

---
In reply to: [305525416](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r305525416) [](ancestors = 305525416,305430798)

src/ReleaseHistory.md
lgolding: Thanks, fixed.

---
In reply to: [312603755](https://github.com/microsoft/sarif-sdk/pull/1570#discussion_r312603755) [](ancestors = 312603755)

PR #1574, opened 2019-07-12
Provide a results caching mechanism in cases where hashing is enabled.

michaelcfanning:


michaelcfanning:
The core change is to cache results by file has when file hashing is enabled in the driver framework. When --hashes exists on the command-line, the driver framework now comprehensively produces the set of file hashes in a multi-threaded operation. After this, when scanning a target, the framework consults a cache (by file hash) that determines whether this file has been scanned before. If it has, the framework grabs the cached results and 'replays' them. As a consequence, log files produced when --hashes exists on the command-line or does not should be identical. The value is a significant performance gain for some tool (BinSkim, for example, acquires and parses PDBs for each file copy currently, a very expensive operation).

Essential files:
 ResultCachingLogger.cs: new logger that caches results by encountered file hash
 MSBuildConverter.cs: renamed converter that can now round-trip all SARIF driver console messages
 ConsoleLogger.cs: now captures emitted output, for test purposes
 AnalyzeCommandBase.cs: integrates results caching with the analysis driver
 AnalyzeCommandBaseTests.cs: testing for new funcitonality

In order to produce and effectively test this functional change, I've also made several other improvements:

1) The ConsoleLogger can be configured to capture all output that passes through it. This allows us to test scenarios where a user has not configured a persisted file (i.e., we can capture the output in order to validate the presence of errors/warnings/etc).
2) Console messages (formatted per MSBUILD conventions) now produce the complete range of SARIF failure levels and result kinds, i.e., error, warning but also open, review, etc.
3) Console messages now use the SARIF 'note' level for failures with that level. We preserve 'info' for SARIF 'informational' messages for back-compat w/previous implementation. This discrepancy (info vs. informational) is the only delta between output and SARIF terms.
4) Renamed VisualStudioBuildLogConverter to MSBuildConverter.
5) We use the MSBuildConverter to turned captured console output back into an actual log file. This round-tripping is a useful scenario: users who have captured console-only output from a SARIF driver tool can now rehydrate that output into a SARIF log file, if useful.
6) The MSBuildConverter now understands all SARIF failure levels and result kinds.
7) In the absence of a file location, the ConsoleLogger now outputs a tool name. This is required for the MSBuildConverter to find/parse the result. This brings the console logger into closer conformance with other MSBUILD message emitting tools (i.e., it was a functional gap).
8) The HashUtilities helper is now plumbed to use an injected IFileSystem object (so that we can replace with a mock during unit testing)
9) Added a 'verbose' flag to the MSBuild converter. When passed, the converter no longer coalesces duplicated messages to a single result. This is required for the console logger scenario, where we want to recover duplicative messages (because these aren't coalesced when persisting to a log file).
10) Fixed issues where 'analyzing 'blah.dll' messages were emitted to log in non-verbose mode (and emit of empty help markdown).

src/Sarif.Driver/Properties/AssemblyInfo.cs
```diff
@@ -5,5 +5,4 @@
...
-[assembly: InternalsVisibleTo("Sarif.Multitool, PublicKey=0024000004800000940000000602000000240000525341310004000001000100433fbf156abe9718142bdbd48a440e779a1b708fd21486ee0ae536f4c548edf8a7185c1e3ac89ceef76c15b8cc2497906798779a59402f9b9e27281fb15e7111566cdc9a9f8326301d45320623c5222089cf4d0013f365ae729fb0a9c9d15138042825cd511a0f3d4887a7b92f4c2749f81b410813d297b73244cf64995effb1")]
-
+[assembly: InternalsVisibleTo("Sarif.Multitool, PublicKey=0024000004800000940000000602000000240000525341310004000001000100433fbf156abe9718142bdbd48a440e779a1b708fd21486ee0ae536f4c548edf8a7185c1e3ac89ceef76c15b8cc2497906798779a59402f9b9e27281fb15e7111566cdc9a9f8326301d45320623c5222089cf4d0013f365ae729fb0a9c9d15138042825cd511a0f3d4887a7b92f4c2749f81b410813d297b73244cf64995effb1")][assembly: InternalsVisibleTo("Test.FunctionalTests.Sarif, PublicKey=0024000004800000940000000602000000240000525341310004000001000100433fbf156abe9718142bdbd48a440e779a1b708fd21486ee0ae536f4c548edf8a7185c1e3ac89ceef76c15b8cc2497906798779a59402f9b9e27281fb15e7111566cdc9a9f8326301d45320623c5222089cf4d0013f365ae729fb0a9c9d15138042825cd511a0f3d4887a7b92f4c2749f81b410813d297b73244cf64995effb1")]
```
lgolding: >][ [](start = 380, length = 2)

Typo: you lost a line break here.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -22,12 +22,7 @@ public abstract class SarifValidationSkimmerBase : Skimmer<SarifValidationContex
 
         public override Uri HelpUri => _defaultHelpUri;
 
-        private readonly MultiformatMessageString _emptyHelpMessage = new MultiformatMessageString
-        {
-            Text = string.Empty
-        };
-
-        public override MultiformatMessageString Help => _emptyHelpMessage;
+        public override MultiformatMessageString Help => null;
```
lgolding: >null [](start = 57, length = 4)

I've made the same change in my pending PR #1570.

src/Sarif/Core/MultiformatMessageString.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public partial class MultiformatMessageString
+    {
+        public bool ShouldSerializeMarkdown() { return !string.IsNullOrEmpty(this.Markdown); }
```
lgolding: >IsNullOrEmpty [](start = 63, length = 13)

Consider: `IsNullOrWhiteSpace`

src/Sarif/ConsoleLogger.cs
```diff
@@ -10,17 +10,46 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+
+        public string CapturedOutput
+        {
+            get
+            {
+                if (_capturedOutput != null)
+                {
+                    return _capturedOutput.ToString();
+                }
+                return null;
```
lgolding: Or just `return _capturedOutput?.ToString();`

src/Sarif/ConsoleLogger.cs
```diff
@@ -10,17 +10,46 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+
         public bool Verbose { get; set; }
 
+        private void WriteLineToConsole(string text = null)
+        {
+            Console.WriteLine(text);
+
+            if (CaptureOutput)
+            {
+                _capturedOutput = _capturedOutput ?? new StringBuilder();
```
lgolding: >_capturedOutput [](start = 16, length = 15)

Can't wait for C#8 where this is just `_capturedOutput ??= new StringBuilder();`

src/Sarif/ConsoleLogger.cs
lgolding: Or write the whole thing on one line (saves 10 lines!):
```
    public string CapturedOutput => _capturedOutput?.ToString();
```

---
In reply to: [303172007](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303172007) [](ancestors = 303172007)

src/Sarif/ConsoleLogger.cs
```diff
@@ -219,9 +251,10 @@ private static void ValidateKindAndLevel(ResultKind kind, FailureLevel level)
                 location = region.FormatForVisualStudio();
             }
 
-            string result = (path != null ? (path + location + ": ") : "") +
-                   issueType + (!string.IsNullOrEmpty(ruleId) ? " " : "") +
-                   (!string.IsNullOrEmpty(ruleId) ? (ruleId + ": ") : "") +
+            string result = 
+                   (path != null ? (path + location) : toolName) + ": " +
+                   issueType + (!string.IsNullOrEmpty(ruleId) ? " " : "")  +
```
lgolding: >   [](start = 73, length = 2)

Intentionally double-spaced?

src/Sarif/ConsoleLogger.cs
```diff
@@ -219,9 +251,10 @@ private static void ValidateKindAndLevel(ResultKind kind, FailureLevel level)
                 location = region.FormatForVisualStudio();
             }
 
-            string result = (path != null ? (path + location + ": ") : "") +
-                   issueType + (!string.IsNullOrEmpty(ruleId) ? " " : "") +
-                   (!string.IsNullOrEmpty(ruleId) ? (ruleId + ": ") : "") +
+            string result =
```
lgolding: >result [](start = 19, length = 6)

While we're in here, suggested variable name: `messageText` (esp. since method is `GetMessageText`).

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1001.SyntaxError.sarif
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json",
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.2.json",
   "version": "2.1.0",
```
michaelcfanning: >toolExecutionNotifications [](start = 11, length = 26)

Here's an example of the 'analyzing file' verbose message that had snuck into our log files, now removed.

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
michaelcfanning: ack

---
In reply to: [303171587](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303171587) [](ancestors = 303171587)

src/Sarif/ExtensionMethods.cs
```diff
@@ -75,7 +75,7 @@ public static string InstanceIdLogicalComponent(this RunAutomationDetails runAut
 
             if (!instanceId.Contains("/"))
             {
-                return null;
+                return String.Empty;
```
lgolding: >  [](start = 8, length = 1)

Extra space introduced

src/Sarif/ConsoleLogger.cs
michaelcfanning: nice!

---
In reply to: [303172128](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303172128) [](ancestors = 303172128)

src/Sarif/HashUtilities.cs
```diff
@@ -2,21 +2,71 @@
...
+            {
+                _fileSystem = _fileSystem ?? new FileSystem();
+                return _fileSystem;
+            }
+
+            set
+            {
+                _fileSystem = value;
+            }
+        }
```
lgolding: I don't think you need all of this. Since the static ctor initializes the property, it's guaranteed to be set by the time anyone invokes the getter, so
```
internal static IFileSystem FileSystem { get; set; }
```
will suffice.

src/Sarif/HashUtilities.cs
```diff
@@ -2,21 +2,71 @@
...
+
+        public static IDictionary<string, HashData> MultithreadedComputeTargetFileHashes(IEnumerable<string> analysisTargets)
+        {
+            if (analysisTargets == null) { return null; }
+
+            Console.WriteLine("Computing file hashes...");
```
lgolding: >Console.WriteLine("Computing file hashes..."); [](start = 12, length = 46)

Did you mean to leave this in? I don't think you want console output from a utility function.

src/Sarif/HashUtilities.cs
michaelcfanning: this doesn't provide for the pattern where a user wants to remove their file system mock, so we set the property to null. this allows tests to set the mock, then set to null, after which the code will naturally rehydrate the standard file system object.

---
In reply to: [303174647](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303174647) [](ancestors = 303174647)

src/Sarif/HashUtilities.cs
michaelcfanning: It's been useful to understand when hashing is occurring but it isn't strictly necessary. to 'fix' this, i'd need to plumb through a logger object to some relevant caller. I'll go look and see if there's a natural place to do this. Agree it's weird code for a utility.

---
In reply to: [303174902](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303174902) [](ancestors = 303174902)

src/Sarif/HashUtilities.cs
lgolding: That is compelling and requires what you just said to be a code comment :-)

---
In reply to: [303175028](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303175028) [](ancestors = 303175028,303174647)

src/Sarif/Core/Run.cs
```diff
@@ -31,7 +32,8 @@ public Uri ExpandUrisWithUriBaseId(string key, string currentValue = null)
             ArtifactLocation fileLocation,
             bool addToFilesTableIfNotPresent = true,
             OptionallyEmittedData dataToInsert = OptionallyEmittedData.None,
-            Encoding encoding = null)
+            Encoding encoding = null,
+            HashData hashData = null)
```
lgolding: >DetermineFromFileExtension [](start = 55, length = 26)

Yeah, good riddance.

src/Sarif/ResultsCachingLogger.cs
```diff
@@ -0,0 +1,103 @@
...
+        public void AnalyzingTarget(IAnalysisContext context)
+        {
+            this.currentFileHash = context.Hashes.Sha256;
+
+            if (HashToResultsMap.ContainsKey(currentFileHash))
+            {
+                // We already have results cached for this file. This means that we shouldn't collect
+                // any data in calls to Log. These calls are the driver framework 'replaying' the
+                // cached results via the aggregated logger instance
+                cacheLoggingData = false;
```
lgolding: >cacheLoggingData [](start = 16, length = 16)

Suggested variable name: `targetAlreadyAnalyzed`.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -216,11 +228,20 @@ private static bool ValidateInvocationPropertiesToLog(TContext context, IEnumera
...
             var logger = new AggregatingLogger();
 
             if (!analyzeOptions.Quiet)
             {
-                logger.Loggers.Add(new ConsoleLogger(analyzeOptions.Verbose));
+                _consoleLogger = new ConsoleLogger(analyzeOptions.Verbose, _tool.Driver.Name) { CaptureOutput = _captureConsoleOutput };
+                logger.Loggers.Add(_consoleLogger);
+            }
+
+            if (analyzeOptions.ComputeFileHashes)
```
lgolding: >ComputeFileHashes [](start = 31, length = 17)

This option (command line option `--hashes`) is obsolete, replaced by `analyzeOptions.DataToInsert` (command line option `--insert`). I suggest that you move the following code:
```
                        OptionallyEmittedData dataToInsert = analyzeOptions.DataToInsert.ToFlags();

                        // This code is required in order to support the obsoleted ComputeFileHashes argument
                        // on the analyze command-line;
                        if (analyzeOptions.ComputeFileHashes) { dataToInsert |= OptionallyEmittedData.Hashes; }
```
... which currently exists in the `InitializeOutputFile` method, to the top of the `Run` method, store the result in an instance file `_dataToInsert`, and use it wherever you've added code to use `ComputeFileHashes`.

You can also do what we do in `SarifLogger`, namely, introduce a property `AnalysisCommandBase.ComputeFileHashes` that pulls out the appropriate flag.

It's a shame to duplicate that code -- ideally it would be an extension method on `AnalyzeOptionsBase` -- but `AnalyzeOptionsBase` already has a property `ComputeFileHashes`. Maybe you could rename that to `ComputeFileHashesObsolete`.

src/Sarif/ResultsCachingLogger.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        public void Log(ReportingDescriptor rule, Result result)
+        {
+            if (!cacheLoggingData) { return; }
+
+            switch (result.Level)
+            {
+                // These result types are optionally emitted.
```
lgolding: >These [](start = 19, length = 5)

It's a pity you have to remember that to write this code correctly. It duplicates logic elsewhere. If we introduced more verbosity levels ("only errors", for example), we'd have to remember to change this code that decides whether to cache the result.

Is there any harm in just always caching it, regardless of verbosity level?

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
lgolding: In fact I notice in your PR description that you expect this feature to be driven by `--hashes`. I think you might have forgotten about `--insert`.

---
In reply to: [303215000](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303215000) [](ancestors = 303215000)

src/Sarif/ResultsCachingLogger.cs
lgolding: Or perhaps introduce introduce a utility method `Utilities.ResultShouldBeLogged(level, verbosity)` and call it in both places.

---
In reply to: [303215190](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303215190) [](ancestors = 303215190)

src/Sarif/HashUtilities.cs
lgolding: Reactivating this for your further consideration. At the very least, if you're not persuaded by this argument, please add the code comment to explain the scenario you're enabling.

I don't think the test code pattern you're enabling is a good one:
```
HashUtilities.FileSystem = mockFileSystem;

// Do stuff

HashUtilities.FileSystem = null;

// Good news! The standard file system will return on the next get access to this property. But I doubt the client knows that.
```

I think that behavior violates least surprise. It could even cause a bug if (because this is test code, after all) the value of this property was _not_ a standard `FileSystem` object before you got your hands on it. I prefer the "leave things as you found them" pattern:
```
IFileSystem previousFileSystem = HashUtilitles.FileSystem;
HashUtilities.FileSystem = mockFileSystem;

// Do stuff

HashUtilities.FileSystem = previousFileSystem;

// Good news! You know for a fact you've done no harm.
```

---
In reply to: [303175558](https://github.com/microsoft/sarif-sdk/pull/1574#discussion_r303175558) [](ancestors = 303175558,303175028,303174647)

PR #1576, opened 2019-07-13
Updating dependency lodash in ESLint.formatter to patch a vulnerability

rscrivens:
all 9 tests passed after updating the dependency

### DIFF
--- a/src/ESLint.Formatter/package.json
+++ b/src/ESLint.Formatter/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "chai": "^4.2.0",
     "jschardet": "latest",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.14",
     "mocha": "^5.2.0",
     "utf8": "^3.0.0"
   }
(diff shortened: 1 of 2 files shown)


src/ESLint.Formatter/package.json
```diff
@@ -31,7 +31,7 @@
   "dependencies": {
     "chai": "^4.2.0",
     "jschardet": "latest",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.14",
```
michaelcfanning: >^4.17.1 [](start = 15, length = 7)

Do we have a release notes document for this component that we can/should update?

lgolding: We don't. @rscrivens: As a separate PR, would you please:

1. Place a ReleaseHistory.md in the src\ESLint.Formatter\ directory, modeled on the one in the src\ directory? (Just start with this release, fixing the vuln, but we won't publish it yet.)
2. Add it to package.json.

The next time we publish (the time _after_ this vuln-fixing release), it will include the release history.

PR #1579, opened 2019-07-19
WorkItemFiler accepts and validates a SARIF log file

lgolding:


lgolding:

lgolding:

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -1,9 +1,100 @@
...
+            _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+        }
+
+        /// <summary>
+        /// Files work items from the results in a SARIF log file.
+        /// </summary>
+        /// <param name="path">
+        /// The path to the SARIF log file.
+        /// </param>
+        public void FileWorkItems(string path)
```
harleenkohli: >path [](start = 41, length = 4)

could we name this something like logFilePath for quicker context?

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -1,9 +1,100 @@
...
+        {
+            if (path == null) { throw new ArgumentNullException(nameof(path)); }
+
+            ValidateLogFileAgainstSarifSchema(path);
```
harleenkohli: >ValidateLogFileAgainstSarifSchema [](start = 12, length = 33)

suggestion: what about a name like "EnsureIsValidSarif" etc?

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: Good rename. Will do in next PR.

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: Good rename. Will do in next PR.

PR #1580, opened 2019-07-20
Introduce filing target abstraction and only file new bugs.

lgolding:
This change does three things:

1. It introduces a "filing target" abstraction ("target" meaning "the place the bugs are going") that will be implemented as `GitHubFilingTarget` and `AzureDevOpsFilingTarget`.

    We represent the abstraction with an abstract base class `FilingTarget` (not `FilingTargetBase` or anything like that) rather than as an interface `IFilingTarget`. I can explain this choice offline if you're interested; I know Michael and I have discussed this before.

2. The `WorkItemFiler` filters out all but the "new" results before passing them on to the filing target.

    The filtering logic ("'new' results only") is currently hard-coded in the `WorkItemFiler` class. You'll see a TODO comment that calls out the possible need to introduce a results filtering strategy abstraction.

3. It makes the `FileWorkItems` method on the `WorkItemFiler` async. This affects the unit tests.

    It turns out to be hard to use FluentAssertions's `Should().Throw<T>` and `Should().NotThrow()` with async, so I did some searching and borrowed some code from the intertubes to do that. It's in the Test.Utilities.SarifProject in AssertionExtensions.cs.

Some things we haven't done yet:

- We have not yet introduced a bucketing strategy object. You'll see a TODO comment that calls out where it fits in. To be clear: unlike the "filtering" I mentioned above, where we do have filtering logic but it's hard-coded, we have not yet implemented _any_ bucketing logic.
- The async methods do not support timeouts, or retries, or progress reporting, or cancellation. One step at a time...


lgolding:

lgolding:

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    [Verb("file-work-items", HelpText = "Send SARIF results to a work item tracking system such as GitHub or Azure DevOps")]
+    public class FileWorkItemsOptions
+    {
+        [Option(
+            't',
+            "target-type",
+            HelpText = "The type of system to which work items are to be filed. Must be one of 'github' and 'azureDevOps'.",
+            Required = true)]
+        public string TargetType { get; internal set; }
```
harleenkohli: >TargetType [](start = 22, length = 10)

Mike's comment during the call made me think - since this is exposed to the user, yes this could be interpreted as work item type (e.g. bug/task/issue). Here are some other suggestions:

- filing-target
- version-control-type
- bug-manager
- issue-reporter

lgolding: I think you're on the right track! We already use "filing target" to mean "the place the bugs are going" but how about `filing-target-type`?

---
In reply to: [306055034](https://github.com/microsoft/sarif-sdk/pull/1580#discussion_r306055034) [](ancestors = 306055034)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -24,35 +31,64 @@ public class WorkItemFiler
...
+            if (logFilePath == null) { throw new ArgumentNullException(nameof(logFilePath)); }
+
+            string logFileContents = _fileSystem.ReadAllText(logFilePath);
+
+            EnsureValidSarifLogFile(logFileContents, logFilePath);
+
+            SarifLog sarifLog = JsonConvert.DeserializeObject<SarifLog>(logFileContents);
+
+            // CONSIDER: Multiple runs?
+            if (sarifLog.Runs[0]?.Results?.Count > 0)
```
harleenkohli: >sarifLog.Runs[0]? [](start = 16, length = 17)

i know this is a future todo based on the comment - we almost certainly would need to merge the sarifs into a single one before the bug filing step is executed. So, we will certainly need to iterate through multiple run objects.

src/Sarif.WorkItemFiling/OneResultPerWorkItemGroupingStrategy.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
+{
+    /// <summary>
+    /// A grouping strategy that files each SARIF result as a separate work item.
+    /// </summary>
+    public class OneResultPerWorkItemGroupingStrategy : IGroupingStrategy
```
harleenkohli: >OneResultPerWorkItemGroupingStrategy [](start = 17, length = 36)

OMG. i loved this name when i was reviewing earlier!

`one result` per `workitem` grouping strategy - yay! so clear!

and then a few hours later - i came back to it and woah!!

the second time i read it as:

 `One result` per `workitemGrouping` strategy  -- which in my mind meant - one result for one work item group ---- which is always true isnt it?

SO.. 

may i recommmend we name this as: `OneResultPerWorkItemStrategy`

src/Sarif.Multitool/FileWorkItemsOptions.cs
harleenkohli: you have my vote on filing-target-type :)

---
In reply to: [306060353](https://github.com/microsoft/sarif-sdk/pull/1580#discussion_r306060353) [](ancestors = 306060353,306055034)

src/Sarif.WorkItemFiling/OneResultPerWorkItemGroupingStrategy.cs
lgolding: I hear you, but since it is a "grouping strategy", I'd like to stick with this name.

---
In reply to: [306064261](https://github.com/microsoft/sarif-sdk/pull/1580#discussion_r306064261) [](ancestors = 306064261)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: Ok, I will make that change now.

---
In reply to: [306062380](https://github.com/microsoft/sarif-sdk/pull/1580#discussion_r306062380) [](ancestors = 306062380)

PR #1582, opened 2019-07-23
Define "filtering strategy" and two implementations; use it in the filer

lgolding:
Also:
- Push filtering and grouping algorithms and tests into subdirectories.

### DIFF
--- a/src/Sarif.WorkItemFiling/Filtering/IFilteringStrategy.cs
+++ b/src/Sarif.WorkItemFiling/Filtering/IFilteringStrategy.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering
+{
+    /// <summary>
+    /// Interface that defines a strategy for selecting only those SARIF results that should
+    /// be filed as work items.
+    /// </summary>
+    public interface IFilteringStrategy
+    {
+        IList<Result> FilterResults(IEnumerable<Result> results);
+    }
+}
--- a/src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
+++ b/src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
@@ -10,9 +10,9 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering
     /// <summary>
     /// A filtering strategy that selects only new results.
     /// </summary>
-    public class NewResultsFilteringStrategy : IFilteringStrategy
+    public class NewResultsFilteringStrategy : FilteringStrategy
     {
-        public IList<Result> FilterResults(IEnumerable<Result> results)
+        public override IList<Result> FilterResults(IEnumerable<Result> results)
         {
             if (results == null) { throw new ArgumentNullException(nameof(results)); }
 
--- a/src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
+++ b/src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
@@ -5,19 +5,19 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
+namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering
 {
     /// <summary>
-    /// A grouping strategy that files each SARIF result as a separate work item.
+    /// A filtering strategy that selects only new results.
     /// </summary>
-    public class OneResultPerWorkItemGroupingStrategy : IGroupingStrategy
+    public class NewResultsFilteringStrategy : IFilteringStrategy
     {
-        public IList<ResultGroup> GroupResults(IEnumerable<Result> results)
+        public IList<Result> FilterResults(IEnumerable<Result> results)
         {
             if (results == null) { throw new ArgumentNullException(nameof(results)); }
 
             return results
-                .Select(r => new ResultGroup { Results = new List<Result> { r } })
+                .Where(r => r.BaselineState == BaselineState.New)
                 .ToList();
         }
     }
--- a/src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
+++ b/src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
 
 namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.GitHub
 {
--- a/src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
+++ b/src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
@@ -9,7 +9,7 @@ using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
 namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.GitHub
 {
     /// <summary>
-    /// Represents a GitHub project to which work items can be filed.
+    /// Represents a GitHub project in which work items can be filed.
     /// </summary>
     public class GitHubFilingTarget : FilingTarget
     {
--- a/src/Sarif.WorkItemFiling/WorkItemFiler.cs
+++ b/src/Sarif.WorkItemFiling/WorkItemFiler.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
 using Microsoft.Json.Schema;
 using Microsoft.Json.Schema.Validation;
 using Newtonsoft.Json;
--- a/src/Sarif.WorkItemFiling/WorkItemFiler.cs
+++ b/src/Sarif.WorkItemFiling/WorkItemFiler.cs
@@ -22,9 +22,11 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
         private static readonly Validator s_logFileValidator = CreateLogFileValidator();
 
         private readonly FilingTarget _filingTarget;
-        private readonly IFilteringStrategy _filteringStrategy;
-        private readonly IGroupingStrategy _groupingStrategy;
-        private readonly IFileSystem _fileSystem;
+        private readonly FilteringStrategy _filteringStrategy;
+        private readonly GroupingStrategy _groupingStrategy;
+
+        // internal to expose to unit tests.
+        internal IFileSystem FileSystem { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WorkItemFiler"> class.</see>
@@ -39,14 +41,14 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
         /// </param>
         public WorkItemFiler(
             FilingTarget filingTarget,
-            IFilteringStrategy filteringStrategy,
-            IGroupingStrategy groupingStrategy,
-            IFileSystem fileSystem)
+            FilteringStrategy filteringStrategy = null,
+            GroupingStrategy groupingStrategy = null,
+            IFileSystem fileSystem = null)
         {
             _filingTarget = filingTarget ?? throw new ArgumentNullException(nameof(filingTarget));
             _filteringStrategy = filteringStrategy ?? throw new ArgumentNullException(nameof(filteringStrategy));
             _groupingStrategy = groupingStrategy ?? throw new ArgumentNullException(nameof(groupingStrategy));
-            _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+            FileSystem = fileSystem ?? new FileSystem();
         }
 
         /// <summary>
@@ -62,7 +64,7 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
         {
             if (logFilePath == null) { throw new ArgumentNullException(nameof(logFilePath)); }
 
-            string logFileContents = _fileSystem.ReadAllText(logFilePath);
+            string logFileContents = FileSystem.ReadAllText(logFilePath);
 
             EnsureValidSarifLogFile(logFileContents, logFilePath);
 
@@ -75,9 +77,8 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
                 if (sarifLog.Runs[i]?.Results?.Count > 0)
                 {
                     IList<Result> filteredResults = _filteringStrategy.FilterResults(sarifLog.Runs[i].Results);
-
-                    // TODO: Consider whether to await each run in turn, or to do them in parallel.
                     IList<ResultGroup> groupedResults = _groupingStrategy.GroupResults(filteredResults);
+
                     IEnumerable<ResultGroup> filedResultsForRun = await _filingTarget.FileWorkItems(groupedResults);
 
                     // TODO: Consider whether to return one batch of "filed results", or one batch per run.
--- a/src/Sarif.WorkItemFiling/WorkItemFiler.cs
+++ b/src/Sarif.WorkItemFiling/WorkItemFiler.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
 using Microsoft.Json.Schema;
 using Microsoft.Json.Schema.Validation;
@@ -19,9 +20,9 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
     public class WorkItemFiler
     {
         private static readonly Validator s_logFileValidator = CreateLogFileValidator();
-        private static readonly Task<IEnumerable<ResultGroup>> s_noFiledResults = Task.FromResult(new List<ResultGroup>().AsEnumerable());
 
         private readonly FilingTarget _filingTarget;
+        private readonly IFilteringStrategy _filteringStrategy;
         private readonly IGroupingStrategy _groupingStrategy;
         private readonly IFileSystem _fileSystem;
 
@@ -36,9 +37,14 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
         /// An object that represents the system (for example, GitHub or Azure DevOps)
         /// to which the work items will be filed.
         /// </param>
-        public WorkItemFiler(FilingTarget filingTarget, IGroupingStrategy groupingStrategy, IFileSystem fileSystem)
+        public WorkItemFiler(
+            FilingTarget filingTarget,
+            IFilteringStrategy filteringStrategy,
+            IGroupingStrategy groupingStrategy,
+            IFileSystem fileSystem)
         {
             _filingTarget = filingTarget ?? throw new ArgumentNullException(nameof(filingTarget));
+            _filteringStrategy = filteringStrategy ?? throw new ArgumentNullException(nameof(filteringStrategy));
             _groupingStrategy = groupingStrategy ?? throw new ArgumentNullException(nameof(groupingStrategy));
             _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
         }
@@ -68,8 +74,7 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
             {
                 if (sarifLog.Runs[i]?.Results?.Count > 0)
                 {
-                    // TODO: Extract this filtering logic into some sort of "filtering strategy" object.
-                    IList<Result> filteredResults = FilterResults(sarifLog.Runs[i].Results);
+                    IList<Result> filteredResults = _filteringStrategy.FilterResults(sarifLog.Runs[i].Results);
 
                     // TODO: Consider whether to await each run in turn, or to do them in parallel.
                     IList<ResultGroup> groupedResults = _groupingStrategy.GroupResults(filteredResults);
--- a/src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
+++ b/src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
+using Microsoft.CodeAnalysis.Test.Utilities.Sarif;
 using Xunit;
 
 namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling.Filtering
@@ -40,54 +41,57 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling.Filtering
             {
                 new Result
                 {
-                    RuleId = "TST0001",
+                    RuleId = TestConstants.RuleIds.Rule1
                 },
 
                 new Result
                 {
-                    RuleId = "TST0002",
+                    RuleId = TestConstants.RuleIds.Rule2,
                     BaselineState = BaselineState.Absent
                 },
 
                 new Result
                 {
-                    RuleId = "TST0003",
+                    RuleId = TestConstants.RuleIds.Rule3,
                     BaselineState = BaselineState.New
                 },
 
                 new Result
                 {
-                    RuleId = "TST0004",
+                    RuleId = TestConstants.RuleIds.Rule4,
                     BaselineState = BaselineState.None
                 },
 
                 new Result
                 {
-                    RuleId = "TST0005",
+                    RuleId = TestConstants.RuleIds.Rule5,
                     BaselineState = BaselineState.Unchanged
                 },
 
                 new Result
                 {
-                    RuleId = "TST0006",
+                    RuleId = TestConstants.RuleIds.Rule6,
                     BaselineState = BaselineState.Updated
                 },
 
                 new Result
                 {
-                    RuleId = "TST0007",
+                    RuleId = TestConstants.RuleIds.Rule7,
                     BaselineState = BaselineState.New
                 }
             };
 
             IList<Result> filteredResults = strategy.FilterResults(results);
 
-            int numResults = results.Length;
-            filteredResults.Count.Should().Be(numResults);
-            for (int i = 0; i < numResults; ++i)
-            {
-                filteredResults[i].RuleId.Should().Be("TST000" + (i + 1).ToString());
-            }
+            filteredResults.Count.Should().Be(results.Length);
+
+            filteredResults[0].RuleId.Should().Be(TestConstants.RuleIds.Rule1);
+            filteredResults[1].RuleId.Should().Be(TestConstants.RuleIds.Rule2);
+            filteredResults[2].RuleId.Should().Be(TestConstants.RuleIds.Rule3);
+            filteredResults[3].RuleId.Should().Be(TestConstants.RuleIds.Rule4);
+            filteredResults[4].RuleId.Should().Be(TestConstants.RuleIds.Rule5);
+            filteredResults[5].RuleId.Should().Be(TestConstants.RuleIds.Rule6);
+            filteredResults[6].RuleId.Should().Be(TestConstants.RuleIds.Rule7);
         }
     }
 }
--- a/src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
+++ b/src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling.Filtering
+{
+    public class AllResultsFilteringStrategyTests
+    {
+        [Fact]
+        public void AllResultsFilteringStrategy_RequiresResults()
+        {
+            var strategy = new AllResultsFilteringStrategy();
+
+            Action action = () => strategy.FilterResults(results: null);
+
+            action.Should().Throw<ArgumentNullException>();
+        }
+
+        [Fact]
+        public void AllResultsFilteringStrategy_AcceptsEmptyResults()
+        {
+            var strategy = new AllResultsFilteringStrategy();
+
+            IList<Result> filteredResults = strategy.FilterResults(new Result[0]);
+
+            filteredResults.Count.Should().Be(0);
+        }
+
+        [Fact]
+        public void AllResultsFilteringStrategy_SelectsAllResults()
+        {
+            var strategy = new AllResultsFilteringStrategy();
+            var results = new Result[]
+            {
+                new Result
+                {
+                    RuleId = "TST0001",
+                },
+
+                new Result
+                {
+                    RuleId = "TST0002",
+                    BaselineState = BaselineState.Absent
+                },
+
+                new Result
+                {
+                    RuleId = "TST0003",
+                    BaselineState = BaselineState.New
+                },
+
+                new Result
+                {
+                    RuleId = "TST0004",
+                    BaselineState = BaselineState.None
+                },
+
+                new Result
+                {
+                    RuleId = "TST0005",
+                    BaselineState = BaselineState.Unchanged
+                },
+
+                new Result
+                {
+                    RuleId = "TST0006",
+                    BaselineState = BaselineState.Updated
+                },
+
+                new Result
+                {
+                    RuleId = "TST0007",
+                    BaselineState = BaselineState.New
+                }
+            };
+
+            IList<Result> filteredResults = strategy.FilterResults(results);
+
+            int numResults = results.Length;
+            filteredResults.Count.Should().Be(numResults);
+            for (int i = 0; i < numResults; ++i)
+            {
+                filteredResults[i].RuleId.Should().Be("TST000" + (i + 1).ToString());
+            }
+        }
+    }
+}
--- a/src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
+++ b/src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
@@ -10,6 +10,7 @@ using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling;
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
+using Microsoft.CodeAnalysis.Test.Utilities.Sarif;
 using Moq;
 using Xunit;
 
@@ -33,6 +34,8 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling
         {
             WorkItemFiler filer;
             Action action = () => filer = new WorkItemFiler(filingTarget: CreateMockFilingTarget(), filteringStrategy: null, groupingStrategy: CreateGroupingStrategy(), fileSystem: CreateMockFileSystem());
+
+            action.Should().Throw<ArgumentNullException>();
         }
 
         [Fact]
@@ -45,12 +48,11 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling
         }
 
         [Fact]
-        public void WorkItemFiler_RequiresAFileSystem()
+        public void WorkItemFiler_UsesARealFileSystemByDefault()
         {
-            WorkItemFiler filer;
-            Action action = () => filer = new WorkItemFiler(filingTarget: CreateMockFilingTarget(), filteringStrategy: CreateFilteringStrategy(), groupingStrategy: CreateGroupingStrategy(), fileSystem: null);
+            WorkItemFiler filer = new WorkItemFiler(filingTarget: CreateMockFilingTarget(), filteringStrategy: CreateFilteringStrategy(), groupingStrategy: CreateGroupingStrategy(), fileSystem: null);
 
-            action.Should().Throw<ArgumentNullException>();
+            filer.FileSystem.GetType().FullName.Should().Be(typeof(FileSystem).FullName);
         }
 
         [Fact]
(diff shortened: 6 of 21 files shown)


michaelcfanning:
>     public interface IGroupingStrategy

In general, I think your namespaces may be overly narrow. A good rule of thumb is that if you have <= 5 namespaces, you probably did not require the additional namespace. A single namespace with 40 - 50 files is perfectly fine, in my opinion/

---
Refers to: src/Sarif.WorkItemFiling/Grouping/IGroupingStrategy.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     public abstract class FilingTarget

This should have a name like FilingHost or FilingSystem (I prefer Host). Remember, that we may evolve this code base into general work item management activities (in addition to the single method you've declared below). And so I think WorkItemHost may be the most appropriate name.

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -9,6 +9,8 @@
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
```
michaelcfanning: Not sure I buy into these namespaces as opposed to a single namespace. This approach is very appealing when authoring the code, as it keeps things conceptually distinct. API consumers are less enthusiastic, as they are required to add multiple namespaces to their source files (already cluttered with namespaces) to make use of happy path scenarios.

src/Sarif.WorkItemFiling/Filtering/IFilteringStrategy.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering
+{
+    /// <summary>
+    /// Interface that defines a strategy for selecting only those SARIF results that should
+    /// be filed as work items.
+    /// </summary>
+    public interface IFilteringStrategy
```
michaelcfanning: >IFilteringStrategy [](start = 21, length = 18)

What is your rationale for preferring an interface in this case over an abstract in FilingTarget (soon to be WorkItemHost, I hope).

michaelcfanning:
>     public class AzureDevOpsFilingTarget : FilingTarget

Please zap this namespace. To get the sort of grouping you're looking for, the file/type name is perfectly sufficient. IE, we can use AzureDevOps (Filter, Grouping, whatever) to help organize this content sensibly in the IDE. Again, the greatest value for API consumers is to add a single namespace. Use the file names to facilitate developer grokking when actually working on this API implementation.

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     /// Represents an Azure DevOps project to which work items can be filed.

NIT: in which?

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -35,9 +37,14 @@ public class WorkItemFiler
...
         /// </param>
-        public WorkItemFiler(FilingTarget filingTarget, IGroupingStrategy groupingStrategy, IFileSystem fileSystem)
+        public WorkItemFiler(
+            FilingTarget filingTarget,
+            IFilteringStrategy filteringStrategy,
+            IGroupingStrategy groupingStrategy,
+            IFileSystem fileSystem)
         {
             _filingTarget = filingTarget ?? throw new ArgumentNullException(nameof(filingTarget));
+            _filteringStrategy = filteringStrategy ?? throw new ArgumentNullException(nameof(filteringStrategy));
```
michaelcfanning: >ArgumentNullException [](start = 64, length = 21)

why aren't we providing happy path values here? I suggest we consider it.

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -35,9 +37,14 @@ public class WorkItemFiler
...
             _groupingStrategy = groupingStrategy ?? throw new ArgumentNullException(nameof(groupingStrategy));
             _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
```
michaelcfanning: >fileSystem [](start = 26, length = 10)

certainly we should be instantiating a default file system here, i.e., the one that maps to System.IO.

src/Sarif.WorkItemFiling/WorkItemFiler.cs
```diff
@@ -67,8 +74,7 @@ public async Task<IEnumerable<ResultGroup>> FileWorkItems(string logFilePath)
             {
                 if (sarifLog.Runs[i]?.Results?.Count > 0)
                 {
-                    // TODO: Extract this filtering logic into some sort of "filtering strategy" object.
-                    IList<Result> filteredResults = FilterResults(sarifLog.Runs[i].Results);
+                    IList<Result> filteredResults = _filteringStrategy.FilterResults(sarifLog.Runs[i].Results);
 
                     // TODO: Consider whether to await each run in turn, or to do them in parallel.
```
michaelcfanning: >or to do them in pa [](start = 73, length = 19)

In general, .NET's web request API is throttled to two connections to the same endpoint. It is a neat trick to try to subvert this and increase parallelism. Since a single run is our happy path, suggest leaving this alone for now.

michaelcfanning:
>     public class ResultGroup

Please do not create this class yet. It is a very trivial change to add it. I very much prefer to leave the code in a focused condition, particularly when costs are very low to make the sort of change implied here. The reasons are twofold. First, perhaps YAGNI. Second, adding significant new files, namespaces, types, leads to downward pressure in taking core design changes in the very early phases of emerging work. Certainly a holder type that could be replaced by List<Result> doesn't meet my sniff test.

---
Refers to: src/Sarif.WorkItemFiling/Grouping/ResultGroup.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>             throw new System.NotImplementedException();

why is this type fully qualified?

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:25 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>         public IList<ResultGroup> GroupResults(IEnumerable<Result> results)

Please see my comments on ResultGroup. But also, I think we're missing something obvious here. We aren't simply breaking results apart, we actually required refactored SARIF files. If I capture a result, for example, I want to capture all the associated logical locations, artifacts, etc. that are relevant to that result (remember we are adding embedded files). With this in mind, You could define this GroupResults helper as returning IList<SarifLog> That is, the design provides a one-to-one association with a potentially refactored SARIF log file and a work item to be filed. Note that this makes the grouping strategy for 'file all results in a single bug' to be very straightforward, just return the input log file.

---
Refers to: src/Sarif.WorkItemFiling/Grouping/OneResultPerWorkItemGroupingStrategy.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     public interface IGroupingStrategy

My opinion is that you should define a log file refactoring strategy (that returns IList<SarifLog> rather than this pattern. The log file is a nice, natural driver for lots of things. It potentially holds lots of data relevant to the results within it, it can be attached to the bug that is filed, etc.

---
Refers to: src/Sarif.WorkItemFiling/Grouping/IGroupingStrategy.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     public class OneResultPerWorkItemGroupingStrategy : IGroupingStrategy

Consider a one bug for all log results refactoring strategy. Also, might as well create the placeholder for our actual default strategy: one bug per tool run per rule. That is, for each run (a specific invocation of a tool), create a bucket bug for each rule id and bucket all results associated with that id. This bucketing strategy is our internal default and will be used in our first scenarios.

---
Refers to: src/Sarif.WorkItemFiling/Grouping/OneResultPerWorkItemGroupingStrategy.cs:13 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public class NewResultsFilteringStrategy : IFilteringStrategy
+    {
+        public IList<Result> FilterResults(IEnumerable<Result> results)
+        {
+            if (results == null) { throw new ArgumentNullException(nameof(results)); }
+
+            return results
+                .Where(r => r.BaselineState == BaselineState.New)
+                .ToList();
```
michaelcfanning: >ToList [](start = 17, length = 6)

I know that you prefer the elegance of these Linq patterns but your tendency to do so (and to create Lists) tends to create performance problems that we have to solve later. Can we possibly think now about a design pattern that minimizes allocations? One approach, for example, might simply store a list of indices into a root file (rather than copying our very large result objects). A second approach might be to only refer to the filter on an as needed basis when performing grouping/refactoring (to see whether it's a candidate to be included in the grouping strategy).

src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Grouping;
```
michaelcfanning: >WorkItemFiling.Grouping [](start = 35, length = 23)

Here's a code smell I'm always skeptical of: no code changed in this file and you were required to add a namespace.

michaelcfanning:
>     public class WorkItemFilerTests

Really hoping to see some tests that try to demonstrate the happy path uses cases. By design, these tests should attempt to minimize test artifacts but should try to mimic as closely as possible a sequence of steps that an actual API client might take. Some of this is obvious from your tests, but it's good to see this in isolation. You can do this in code comments if it's not easy to implement a clean solution backed by mocks. These tests don't have to run, necessarily (although it's nice if they do!), they only need to compile. In my experience, early and frequent looks at how the API actually is consumed is very helpful in getting to a solid design. We have not followed this process much in the SARIF SDK and it shows. :P

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs:18 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
```diff
@@ -95,9 +104,9 @@ public async Task WorkItemFiler_FilesWorkItemsOnlyForNewResults()
 
             IEnumerable<ResultGroup> filedWorkItems = await filer.FileWorkItems(LogFilePath);
 
-            // The test file NewAndOldResults.sarif contains 5 results, but only 2 of them
-            // have "baselineState": "new".
-            filedWorkItems.Count().Should().Be(2);
+            // The test file NewAndOldResults.sarif contains 5 results, only 2 of them
+            // have "baselineState": "new", but the "all results" filter will take all of them.
+            filedWorkItems.Count().Should().Be(5);
```
michaelcfanning: >5 [](start = 47, length = 1)

These magic numbers are a poor practice, as you know, they are decoupled from the test artifacts and are difficult to maintain (requiring developers to manually understand/verify whether code changes have properly modified these values. A preferable approach would be to author some test utilities that construct the log file in a parameterized way. Alternately, you could glean your magic values from a separate helper that simply parsed the SarifLog file using alternate, well-established code (in this case, you'd deserialize and just cound the results). These kinds of tests are 'self-healing', i.e., as test inputs change, the validation values update in a reliable way.

michaelcfanning:
>             resultGroups[0].Results[0].RuleId.Should().Be("TST0001");

you should DRY these constants and might consider creating a class, shared location for various test data. 

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/Grouping/OneResultPerWorkItemGroupingStrategyTests.cs:56 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>         public abstract Task<IEnumerable<ResultGroup>> FileWorkItems(IEnumerable<ResultGroup> resultGroups);

Just curious, have you done some reading/web querying on creating sync/async patterns in API? e.g

[https://www.infoq.com/articles/Async-API-Design/](https://www.infoq.com/articles/Async-API-Design/) 
[http://channel9.msdn.com/Events/TechEd/Europe/2013/DEV-B318](http://channel9.msdn.com/Events/TechEd/Europe/2013/DEV-B318) 

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:25 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using System;
+using System.Collections.Generic;
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.WorkItemFiling.Filtering;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.WorkItemFiling.Filtering
+{
+    public class AllResultsFilteringStrategyTests
```
michaelcfanning: >AllResultsFilteringStrategyTests [](start = 17, length = 32)

Can I suggest that you please create some early negative test cases? It is important to understand API requirements in exceptional cases, for example. i.e., what does the code look like to handle a case where one of our grouping/refactoring strategies throws an exception? something I think is important to drive into the design early. I usually author concrete types that can be initialized via ctor with data that describes what behaviors they should engage in. In addition to unhandled exceptions, we should be considering some eventual machinery around hang conditions/time out. This is tricky to implement, as you know. We will likely need to address this at some point but perhaps not right away. We certainly should tackle the problem of exceptions in absolutely every possible component/layer in the design from the beginning, imo.

michaelcfanning:
Your code coverage so far is very good, nice work. Can you please help model using explicit code coverage data in order to keep our coverage >=95%? In your current implementation, at the expense of some tests that simply expect NotImplementedException on invoking the actual file work items method, you could hit this goal. Creating those shell tests is not a waste of time: you will expand them to more complex testing/validation later. Again, I am using this larger team opportunity for us to demonstrate greater maturity in our work than the SARIF SDK has consistently seen (we've been pretty decent with testing, admittedly <g>).

michaelcfanning:
I know that you've only created a shell command for file-work-items, would really loved to have seen some parallel shell tests alongside these. It is very easy to forget about testing the high level commands (which is unfortunate, since they often required some subtle design planning in order to be testable). Creating some preliminary tests will help set the stage for us, even if they are quite rudimentary or incomplete to begin with.

lgolding:
>             throw new System.NotImplementedException();

Because VS generated the method from the right-click, "Implement interface" gesture. But since I'm using `System` anyway here (I don't always), I'll change it.

---
In reply to: [514041840](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514041840) [](ancestors = 514041840)

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:25 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class AzureDevOpsFilingTarget : FilingTarget

Let's talk about namespaces when you get back. In various comments you gave various rationales for a single namespace, so I will collect my responses here and we can discuss them all at once. For now I am going to leave things where they are.

- **File names an adequate grouping mechanism**: They can be, in some cases. In other cases (filtering strategies, grouping strategies), the names do not share a common prefix.
- **"Five members"**: This is just a guideline. In the case of strategies, the whole point is to be able to add more of them. On the day we define the fifth grouping strategy, it will be too late to push them into a namespace. In fact, in another comment you suggested two more grouping strategies (one bug per tool, one bug per log) -- and then, along with the base class, the population of the `Grouping` namespace is already five!
- **Multiple namespaces a programmer burden**: True, but mitigated by IDE "fix it" gesture that adds the `using` directive for you.

---
In reply to: [514039756](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514039756) [](ancestors = 514039756)

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     /// Represents an Azure DevOps project to which work items can be filed.

Well, "to" made sense to me because you direct things "to" a target, but sure, "in" reads a little better. Fixed.

---
In reply to: [514039790](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514039790) [](ancestors = 514039790)

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public interface IGroupingStrategy

See my uber-response to your namespace comments in AzureDevOpsFilingTarget.cs. Let's discuss when you get back.

---
In reply to: [514038801](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514038801) [](ancestors = 514038801)

---
Refers to: src/Sarif.WorkItemFiling/IGroupingStrategy.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class WorkItemFilerTests

I'm not sure what you're getting at. We have exactly one way to call this:
```
IEnumerable<ResultGroup> filedWorkItems
    = await filer.FileWorkItems(LogFilePath);
```
There is no "sequence of steps" to call it, and this API call is demonstrated in several tests ("null results", "empty results", "new and old results", "multiple runs").

Let's discuss when you get back.

---
In reply to: [514044873](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514044873) [](ancestors = 514044873)

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs:18 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
lgolding: The problem with writing helpers to construct the test artifacts is that now you have to verify the helpers. I've found problems in such helpers before.

The problem with calculating the expected values is that you end up duplicating the product code logic, and then you're not testing anything. For example, suppose I deserialize the log and write this test code:
```
string logFileContents = ... // Extract text from resource.
var log =JsonConvert.Deserialize<SarifLog>(logFileContents);
int expectedCount = log.Runs[0].Results.Where(r => r.BaselineState == BaselineState.New);
...
filedWorkItems.Should().Be(expectedCount);
```
I just rewrote the product code.

The approach you're suggesting is valuable if there is an _independent_ way of deciding what outcome you expect. If you have your old copy of Code Complete lying around, see the section "Two Algorithms are Better Than One", pp. 33-38, or just skip to the punch line on p. 38, "Use a second algorithm to validate your results". Of course he's not writing about unit tests -- he's really talking about adding the second algorithm to the Debug version of your _product_ code! -- but the idea is the same.

We can talk about this more when you get back.

---
In reply to: [306120394](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306120394) [](ancestors = 306120394)

lgolding:
>         public abstract Task<IEnumerable<ResultGroup>> FileWorkItems(IEnumerable<ResultGroup> resultGroups);

I've read (and re-read the relevant portion of, for this project) the book "Concurrency in C# Cookbook" by Stephen Cleary. But I will also look at those two references, thanks.

---
In reply to: [514046440](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514046440) [](ancestors = 514046440)

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:25 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public abstract class FilingTarget

But then why is the top-level class called `WorkItemFiler` (or just `Filer` if I have my way)? If we want to future-proof in that direction, we should call it (for example) `WorkItemClient`.

This is the first I've heard of the possibility of evolving to support other work item management activities.

---
In reply to: [514039042](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514039042) [](ancestors = 514039042)

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: Oh! That explains why, in one of the code bases I looked at (I forget which one), they processed the work items in batches of two. Naturally, there was no comment explaining that ;-)

Ok, I will remove this comment and we'll leave it alone for now.

---
In reply to: [306116599](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306116599) [](ancestors = 306116599)

src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
lgolding: We're not copying `Result` objects; they're reference types. We're just copying references. (Will think more about this comment and respond in more detail later.)

---
In reply to: [306119043](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306119043) [](ancestors = 306119043)

src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
lgolding: This isn't a code smell. When I move a type from one namespace to another, client code changes. Once the type is in the right namespace, this doesn't happen again. This change is just "I made a mistake, I put a type in the wrong namespace".

---
In reply to: [306119188](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306119188) [](ancestors = 306119188)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: See my uber-response to your namespace comments in AzureDevOpsFilingTarget.cs. Let's discuss when you get back.

---
In reply to: [306115401](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306115401) [](ancestors = 306115401)

src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
michaelcfanning: You're right about that, of course, we will copy references only. Minimally you can allocate this collection with a predetermined count. Actual issue may be due to short-lived objects (maybe not for a large list, though).

---
In reply to: [306309402](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306309402) [](ancestors = 306309402,306119043)

michaelcfanning:
>     public class ResultGroup

ok, having thought a bit, I do see that you will need some sort of object that contains an array of objects (to be persisted as attachments) and some metadata for the bug. Something like WorkItemMetadata (i.e., a bundling of all the things that will be examined or utilized to actually file a bug). I suggest we decouple this thing from SARIF entirely, i.e., you should be able to file a simple bug without invoking a single SARIF concept. This class could still have a ctor or initialize from SARIF so that it can, for example, utilize some strategy for creating a bug title.

Object should contain, for now, minimally, Title, Tags, Area Path, Body and an array of objects/streams to add as attachments (have you started researching attachments yet?).

---
Refers to: src/Sarif.WorkItemFiling/Grouping/ResultGroup.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/Filtering/IFilteringStrategy.cs
lgolding: My rationale was poor. I was thinking of abstract base class as a place to put shared implementation helpers, and I couldn't imagine there being any such for the strategy classes. But:

1. The stronger rationale for abstract base classes is that it allows you to add new _members_ to the base, without breaking clients.
2. Uniformity is a good thing, and if we use abstract base classes everywhere, then we don't have readers of the code wondering "why here but not there?".

Fixed

---
In reply to: [306115581](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306115581) [](ancestors = 306115581)

michaelcfanning:
>     public class WorkItemFilerTests

In a real world, end-to-end scenario, a user would retrieve a SARIF log file, compare it to some other, choose the 'file new strategy, file bugs, handle exceptions, etc..

This single line is admirably brief but wouldn't be sufficient for a code/SDK sample.

---
In reply to: [514194451](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514194451) [](ancestors = 514194451,514044873)

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs:18 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class WorkItemFilerTests

What you're describing is a great SDK sample but I don't see it as a member of `WorkItemFilerTests`. I will file a work item to create a sample and put it in the Samples directory. And I agree that should happen soon -- but I don't agree it should happen this sprint, when we are under the gun to show end-to-end functionality with our pipeline. Come to think of it, our pipeline does demonstrate the end-to-end, so we can use it as the basis for our sample.

---
In reply to: [514209109](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514209109) [](ancestors = 514209109,514194451,514044873)

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs:18 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class ResultGroup

Resolving this comment per your new comment below (to which I will reply).

---
In reply to: [514041734](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514041734) [](ancestors = 514041734)

---
Refers to: src/Sarif.WorkItemFiling/ResultGroup.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class ResultGroup

Glad you agree with the need to associate filing metadata with the set of `Result`s (per my `<remarks>` in the summary comment above). And I agree with a `WorkItemMetadata` class... except... this is interesting; it seems you're proposing to invert the design, so that instead of
```
ResultGroup
    Result*
    WorkItemMetadata
        Title
        ...
```
we have
```
WorkItemMetadata
    Title
    ...
    Attachment* (which can be the SARIF result objects, or a log)
```
I don't want to do this in this PR, but will do this or something like it as part of the actual bug filing PR, which I will get to today.

I do know the attachment story for both targets:

- ADO supports them: [https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/attachments?view=azure-devops-server-rest-5.0](https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/attachments?view=azure-devops-server-rest-5.0)
- GitHub does _not_ have a web API for attachments: [https://github.com/isaacs/github/issues/1133](https://github.com/isaacs/github/issues/1133). You can drag/drop an attachment into a comment -- but only certain file types are supported (the feature was intended for images). We'd have to work around this by (for example) adding an issue comment containing the text of the attachment. In fact, see this response to the GitHub issue I mentioned: [https://github.com/isaacs/github/issues/1133#issuecomment-407345263](https://github.com/isaacs/github/issues/1133#issuecomment-407345263)

---
In reply to: [514208468](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514208468) [](ancestors = 514208468)

---
Refers to: src/Sarif.WorkItemFiling/ResultGroup.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/Filtering/NewResultsFilteringStrategy.cs
lgolding: I absolutely agree we should defer perf work until we have profiling data. Resolving.

---
In reply to: [306313524](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306313524) [](ancestors = 306313524,306309402,306119043)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: Fixed.

---
In reply to: [306116240](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306116240) [](ancestors = 306116240)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: I provided a default file system.

There is no obvious default for filtering strategy. If we use "all results", it surprises people who use result matching and don't want to see issues repeated. If we use "new results", it surprises people who don't use result matching and wonder why nothing got filed.

We could define `NewOrUnmarkedFilteringStrategy` that takes either `BaselineState.New` or `BaselineState.None`, because the spec requires that `result.baselineState` be "all or nothing" (and our result matchers implement it this way).

For grouping strategy, `OneResultPerWorkItemStrategy` is probably the least surprising, but it can have serious consequences (filing 100,000 bugs is no joke), so IMO the caller should have to make the choice explicitly.

So I think we should leave the other parameters without defaults for now.

---
In reply to: [306116112](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306116112) [](ancestors = 306116112)

lgolding:
>     public interface IGroupingStrategy

That makes a lot of sense. I won't do it in this PR because constructing a _valid_ log file from a subset of results is a delicate thing:

- You have to subset the `run.files`, `run.logicalLocations`, `run.threadFlowLocations`, `run.graphs`, `run.artifacts`, `run.webRequests`, and `run.webResponses` arrays to only those elements needed by the `results` subset (otherwise your attachment is potentially huge in a large number of work items).
- You have to adjust the array indices (and remember, there are 15 properties!) to the subsetted arrays.

We can't do this in time to meet our sprint goal.

---
In reply to: [514042765](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514042765) [](ancestors = 514042765)

---
Refers to: src/Sarif.WorkItemFiling/IGroupingStrategy.cs:12 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public abstract class FilingTarget

For that matter, why is our project/assembly called `WorkItemFiling`? We can discuss this when you get back but I won't change it now.

---
In reply to: [514201966](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514201966) [](ancestors = 514201966,514039042)

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>         public IList<ResultGroup> GroupResults(IEnumerable<Result> results)

See my comment in GroupingStrategy.cs (renamed from IGroupingStrategy.cs). We can't do that work this sprint.

---
In reply to: [514042567](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514042567) [](ancestors = 514042567)

---
Refers to: src/Sarif.WorkItemFiling/OneResultPerWorkItemGroupingStrategy.cs:15 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
lgolding: This comment actually applies to `WorkItemFilerTests`, where the question is "What happens to the filer when a strategy throws an exception?". I can easily test that by mocking up strategy objects that throw. The answer, of course, as the code stands, is "The exception escapes from the `WorkItemFiler`", and I'll write the tests that say that.

As for this class (`AllResultsFilteringStrategy`) we do have a test for the only exception it explicitly throws (`ArgumentNullException`).

Working on this now..

---
In reply to: [306121895](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306121895) [](ancestors = 306121895)

lgolding: Fixed.

---
In reply to: [306355604](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306355604) [](ancestors = 306355604,306121895)

lgolding:
>             resultGroups[0].Results[0].RuleId.Should().Be("TST0001");

Done. New class `TestConstants` in test utilities project. Sub-class `RuleIds`, so `TestContants.RuleIds,Rule1` etc.

---
In reply to: [514045727](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514045727) [](ancestors = 514045727)

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/OneResultPerWorkItemGroupingStrategyTests.cs:56 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
>     public class OneResultPerWorkItemGroupingStrategy : IGroupingStrategy

I will do the bug-per-log strategy after we modify the code so that the grouping strategy produces a log file.

I will do the per-tool-per-rule strategy after I get back from the dentist.

---
In reply to: [514043196](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514043196) [](ancestors = 514043196)

---
Refers to: src/Sarif.WorkItemFiling/OneResultPerWorkItemGroupingStrategy.cs:13 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

lgolding:
The only two NYI exceptions I see are in the `FileWorkItems` methods of the `{GitHub,AzureDevOp}FilingTarget`. Yes, it's a good idea to set up those test shells.

I can't actually measure code coverage because it doesn't work on the "preview" version of VS 2019 that I installed to do the Roslyn work. Tonight I'll uninstall it and install a production version and hopefully it will work.

Working on the test shells now...

---
In reply to: [514049845](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514049845) [](ancestors = 514049845)

lgolding:
Done.

---
In reply to: [514321065](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514321065) [](ancestors = 514321065,514049845)

lgolding:
Since I've already started the multitool work on another branch, I will do this as part of the next PR.

---
In reply to: [514050646](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514050646) [](ancestors = 514050646)

lgolding:
Now done in [https://github.com/microsoft/sarif-sdk/pull/1583](https://github.com/microsoft/sarif-sdk/pull/1583).

---
In reply to: [514329235](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514329235) [](ancestors = 514329235,514050646)

src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
michaelcfanning: You didn't just rewrite the product code. Instead, you utilize a separate area of the product in order to validate the function of fileWorkItems (which flows through an entirely different code path). Assuming that serialization itself isn't broken (which is very unlikely and which should be triggering other tests), you're ok. Now your test is self-healing: if you adjust the test file, a developer doesn't have to go adjust your magic values.

---
In reply to: [306303730](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306303730) [](ancestors = 306303730,306120394)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
michaelcfanning: I don't want an overly granular namespacing strategy for the API. Yes, we can talk when I get back. I'm glad to make this change myself if you don't want to do it. My strategy is to keep things flat and to make namespaces argue their way way.

---
In reply to: [306311176](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306311176) [](ancestors = 306311176,306115401)

michaelcfanning:
>     public abstract class FilingTarget

OK, sounds like a good naming discussion. We might want to convene a short meeting just on this topic, once we have sketched out all the broad concepts we think we'll need.

---
In reply to: [514233879](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514233879) [](ancestors = 514233879,514201966,514039042)

---
Refers to: src/Sarif.WorkItemFiling/FilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     public class AzureDevOpsFilingTarget : FilingTarget

I am most concerned about the readability of code. With a complex file (very common in today's OSS-driven world), adding five using statements from a single OSS component is not pleasant and clutters the client source file. We should prefer a single namespace only ('give me access to the SARIF work item management facility) unless/until there is some exotic bolt-on functionality.

---
In reply to: [514190907](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514190907) [](ancestors = 514190907,514039756)

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

michaelcfanning:
>     public class WorkItemFilerTests

The point here is to do this as an exercise as early as possible to drive the design. Very standard API design tool. I will put something together. I think you'll find it accelerates our ability to think about the end-to-end design.

---
In reply to: [514212754](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514212754) [](ancestors = 514212754,514209109,514194451,514044873)

---
Refers to: src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs:18 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
michaelcfanning: All results is a fine filtering strategy. After all, what's wrong with 'please assume all these results are relevant to the operation' as a default? The most nuanced default filtering strategy would look like so: examine the first result, if baselineState == None, then we have no baseline information in the log file and so, please pull in all results. If the first result is not None, then all results shall be decorated with a baseline state, in which case we should pull in any result that is either Updated or New. For updated items, the work item thingy will find the existing bug and update its contents, for New items we will file a new item. I suggest you implement this strategy and make it the default. For now, it's fine for Updated bugs to be a no-op.

---
In reply to: [306346682](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306346682) [](ancestors = 306346682,306116112)

src/Test.UnitTests.Sarif.WorkItemFiling/Filtering/AllResultsFilteringStrategyTests.cs
michaelcfanning: My larger point is that unhandled exceptions in all possible layers/components s/be explicitly tested before we declare victory. For asynchronous API, exception handling is critical, so that exceptions don't get swallowed/unreported/cause hangs when they occur.

---
In reply to: [306365826](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306365826) [](ancestors = 306365826,306355604,306121895)

src/Sarif.WorkItemFiling/GitHub/GitHubFilingTarget.cs
michaelcfanning: Gotcha.

---
In reply to: [306310265](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306310265) [](ancestors = 306310265,306119188)

lgolding:
>     public class AzureDevOpsFilingTarget : FilingTarget

A flat list of `using` directives that your eye skip right over (do you ever read `using` directives except to see if any of them are grayed out and can be removed?) does not IMO compromise the readability of the code.

Having said that, I have apparently failed to persuade you, so I have (with ill grace) removed the nested namespaces. I have _preserved_ the directory structure.

---
In reply to: [514411964](https://github.com/microsoft/sarif-sdk/pull/1582#issuecomment-514411964) [](ancestors = 514411964,514190907,514039756)

---
Refers to: src/Sarif.WorkItemFiling/AzureDevOps/AzureDevOpsFilingTarget.cs:14 in ff637a5. [](commit_id = ff637a50bddc07dfaccadd9f73a34abd0d7751b4, deletion_comment = False)

src/Sarif.WorkItemFiling/WorkItemFiler.cs
lgolding: See my response in AzureDevopsFilingTarget.cs I have removed the nested namespaces but preserved the directory structure.

---
In reply to: [306564912](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306564912) [](ancestors = 306564912,306311176,306115401)

src/Test.UnitTests.Sarif.WorkItemFiling/WorkItemFilerTests.cs
lgolding: No, the reason `FileWorkItems` arrived at the answer it did (2 filed work items, at the time I wrote my previous comment) is that it invoked the `NewResultsFilteringStrategy`, whose code includes precisely the expression `Where(r =>r.BaselineState == BaselineState.New)` that appears in the test code.

Now, having said that -- you could argue that we have separate unit tests for the `NewResultsFilteringStrategy`, so it's ok to rely on its correctness here. That kind of goes against the spirit of a unit test which is to test a method in isolation, independent of its dependencies.

And having said all that, I've now modified the test to use the trivial `AllResultsFilteringStrategy`, and even I am not so principled that I'm worried about its correctness. So I'll make the change in this test. I'm not convinced this is always the right thing to do.

---
In reply to: [306564691](https://github.com/microsoft/sarif-sdk/pull/1582#discussion_r306564691) [](ancestors = 306564691,306303730,306120394)

PR #1583, opened 2019-07-23
Define and validate options for file-work-items command.

lgolding:


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
             "filing-target-type",
             HelpText = "The type of system to which work items are to be filed. Must be one of 'github' and 'azureDevOps'.",
             Required = true)]
-        public string TargetType { get; internal set; }
+        public string FilingTargetType { get; internal set; }
+
+        [Option(
+            'p',
+            "project-uri",
+            HelpText = "The absolute URI of the project under which the work items are to be filed.",
```
harleenkohli: >The absolute URI of the project under which the work items are to be filed." [](start = 24, length = 76)

it might be useful to provide an example here? e.g. [https://dev.azure.com/{org}/{project](https://dev.azure.com/{org}/{project)} or [https://github.com/{org}/{project](https://github.com/{org}/{project)}

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+            'f',
+            "filtering-strategy",
+            HelpText = "Specifies the strategy for selecting which results to file. Must be one of 'new' and 'all'.",
+            Required = true)]
+        public string FilteringStrategy { get; internal set; }
+
+        [Option(
+            'g',
+            "grouping-strategy",
+            HelpText = "Specifies the strategy for grouping grouping SARIF results into sets that should each be filed together as a single work item. Must be 'perResult'.",
```
harleenkohli: >grouping grouping [](start = 51, length = 17)

typo :O

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+
+        // For an explanation of the difference between ProjectUriString and ProjectUri, see FileWorkItemsCommand.ValidateOptions.
+        public Uri ProjectUri { get; internal set; }
+
+        [Option(
+            'f',
+            "filtering-strategy",
+            HelpText = "Specifies the strategy for selecting which results to file. Must be one of 'new' and 'all'.",
+            Required = true)]
+        public string FilteringStrategy { get; internal set; }
```
harleenkohli: >FilteringStrategy [](start = 22, length = 17)

does it make sense have all these properties (filterstrategy, groupingStrategy, filing-target-type) as enums instead of strings?

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
     public class FileWorkItemsOptions
     {
         [Option(
-            'f',
+            't',
             "filing-target-type",
             HelpText = "The type of system to which work items are to be filed. Must be one of 'github' and 'azureDevOps'.",
             Required = true)]
-        public string TargetType { get; internal set; }
+        public string FilingTargetType { get; internal set; }
```
harleenkohli: >FilingTargetType [](start = 22, length = 16)

Ryley points out (and now that i think of it, i agree) we could potentially derive this information from the root url, so do we really need it?

src/Sarif.Multitool/MultitoolResources.resx
```diff
@@ -129,6 +129,21 @@
   <data name="ErrorNoGroupInFileName" xml:space="preserve">
     <value>Error: No "group" (the part of the file name before the dash) found in '{0}'. All file names must be of the form [Group]-[RunID] and each adjacent pair in the same group is compared.</value>
   </data>
+  <data name="ErrorUnknownFilingTargetType" xml:space="preserve">
+    <value>Error: The value of the '{0}' option is not a known filing target type.
```
harleenkohli: >Error: The value of the '{0}' option is not a known filing target type. [](start = 11, length = 71)

(ryley says) should we show the wrong value that user inputted in these messages for faster context?

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: It might. I will look into that.

---
In reply to: [306567622](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306567622) [](ancestors = 306567622)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Good idea.

---
In reply to: [306566043](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306566043) [](ancestors = 306566043)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: You are right -- at least for now. You can imagine a system where customers can stand up their own instances ([https://www.example.com/issues](https://www.example.com/issues), [https://www.contoso.com/issues](https://www.contoso.com/issues)). Then you can't tell. But as long as we only support ADO and GitHub, we don't need this.

---
In reply to: [306568347](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306568347) [](ancestors = 306568347)

src/Sarif.Multitool/MultitoolResources.resx
```diff
@@ -129,6 +129,21 @@
   <data name="ErrorNoGroupInFileName" xml:space="preserve">
     <value>Error: No "group" (the part of the file name before the dash) found in '{0}'. All file names must be of the form [Group]-[RunID] and each adjacent pair in the same group is compared.</value>
   </data>
+  <data name="ErrorUnknownFilingTargetType" xml:space="preserve">
```
harleenkohli: >ErrorUnknownFilingTargetType [](start = 14, length = 28)

since this is a common file for all potential "multitool" related messages, we should append the ones specific bug filing component with some special keyword e.g. "FileWorkItems_"


(Ryley says)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Fixed.

---
In reply to: [306570111](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306570111) [](ancestors = 306570111,306568347)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Fixed.

---
In reply to: [306569470](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306569470) [](ancestors = 306569470,306566043)

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: Yes, these s/be enums if we retain them. This provides for automatic command-line validation provided by our OSS helper (which isn't always super readable, but hey, it's free).

---
In reply to: [306569347](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306569347) [](ancestors = 306569347,306567622)

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+
+        [Option(
+            'p',
```
michaelcfanning: >p [](start = 13, length = 1)

You'll need to reserve 'p' for 'pretty-print' on this command. Also, please add 'pretty-print' for the output path. Also, add output file path. Also add a force option to overwrite existing outputs. See SingeFileOptions and the options command base for boilerplate arg definitions and if you see a DRYing opportunity, of course, grab it.

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: >under [](start = 56, length = 5)

NIT: first your code said 'project to which', then I suggested using 'project in which' and now you author 'project under which'. Language is flexible.

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+        public string ProjectUriString { get; internal set; }
+
+        // For an explanation of the difference between ProjectUriString and ProjectUri, see FileWorkItemsCommand.ValidateOptions.
+        public Uri ProjectUri { get; internal set; }
+
+        [Option(
+            'f',
+            "filtering-strategy",
+            HelpText = "Specifies the strategy for selecting which results to file. Must be one of 'new' and 'all'.",
+            Required = true)]
```
michaelcfanning: >Required = true [](start = 12, length = 15)

We should have reasonable defaults. See the other PR for the behavior I suggest (either pull in all results, if no baseline data, otherwise, process updated and new).

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+        public string FilteringStrategy { get; internal set; }
+
+        [Option(
+            'g',
+            "grouping-strategy",
```
michaelcfanning: >grouping-strategy [](start = 13, length = 17)

I prefer 'refactoring' as a term here and elsewhere because we will, in fact, be refactoring a log file (potentially) into many others. That's a bit more of an extreme operation than grouping (which suggests reorganzing a set to me)

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
+            HelpText = "Specifies the strategy for grouping grouping SARIF results into sets that should each be filed together as a single work item. Must be 'perResult'.",
+            Required = true)]
+        public string GroupingStrategy { get; internal set; }
+
+        [Value(0,
+            MetaName = "<inputFile>",
```
michaelcfanning: > MetaName = "<inputFile>", [](start = 11, length = 26)

Note to self and team: I don't know if we're properly utilizing MetaName across the multitool, w/be a nice clean-up if not. Great onboarding work for Surbhi or anyone else interested in ramping up over here.

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -2,15 +2,105 @@
...
+using System.Linq;
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
     public class FileWorkItemsCommand : CommandBase
     {
-        public int Run(FileWorkItemsOptions fileWorkItemsOptions)
+        private static readonly string[] s_knownFilingTargetTypes = new[] { "github", "azuredevops" };
+        private static readonly string[] s_knownFilteringStrategies = new[] { "new", "all" };
+        private static readonly string[] s_knownGroupingStrategies = new[] { "perresult" };
```
michaelcfanning: You will definitely want enums for these, I think. :)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: I'm gonna push back on that. "Refactoring" has a very specific meaning: a code change made with the intent of improving the maintainability or extensibility of a code base. I don't see it here. I think grouping is just fine.

---
In reply to: [306575297](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306575297) [](ancestors = 306575297)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Ok. I take it that the output file is the original SARIF file with the bug URIs added.

---
In reply to: [306574608](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306574608) [](ancestors = 306574608)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: IIRC the price is that these parameter values are case-sensitive. I will look into it.

---
In reply to: [306574126](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306574126) [](ancestors = 306574126,306569347,306567622)

lgolding: See what I did in validating the `ProjectUri` option. I found CommandLine's built-in validation inadequate and worked around it.

---
In reply to: [306577599](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306577599) [](ancestors = 306577599,306574126,306569347,306567622)

src/Sarif.Multitool/MultitoolResources.resx
michaelcfanning: yes we should. I don't like this Error: prefix convention, what's that all about? 

---
In reply to: [306569098](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306569098) [](ancestors = 306569098)

src/Test.UnitTests.Sarif.Multitool/FileWorkItemsCommandTests.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Collections.Generic;
+using FluentAssertions;
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    public class FileWorkItemsCommandTests
+    {
+        [Theory]
```
michaelcfanning: >Theory [](start = 9, length = 6)

No [Theory] usage please! If you'll recall, this convention tanked unit test performance previously and I had to go to considerable effort to undo them and make tests run in a reasonable time. There are other patterns that work just fine and which don't provoke whatever reflection costs this mechanism appears to.

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: I'd love to use `SingleFileOptions`. But it drags in `CommonOptions` which means I'd have to support `u` (uriBaseId) and `i` (insert optional data) and `v` (select SARIF output version). I'm going to copy the options I need.

---
In reply to: [306577540](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306577540) [](ancestors = 306577540,306574608)

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -2,15 +2,105 @@
...
-            return 1;
+            bool valid = true;
+
+            options.FilingTargetType = options.FilingTargetType.ToLowerInvariant();
+            if (!s_knownFilingTargetTypes.Contains(options.FilingTargetType))
+            {
+                string optionDescription = CommandUtilities.GetOptionDescription<FileWorkItemsOptions>(nameof(options.FilingTargetType));
+                Console.Error.WriteLine(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
```
michaelcfanning: This is a default and this operation is not particularly sensitive. You can leave in the explicit culture, fine, but not strictly necessary, particularly if you'd like code to be more concise.

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -2,15 +2,105 @@
...
     public class FileWorkItemsCommand : CommandBase
     {
-        public int Run(FileWorkItemsOptions fileWorkItemsOptions)
+        private static readonly string[] s_knownFilingTargetTypes = new[] { "github", "azuredevops" };
+        private static readonly string[] s_knownFilteringStrategies = new[] { "new", "all" };
+        private static readonly string[] s_knownGroupingStrategies = new[] { "perresult" };
+
+        private FileWorkItemsOptions _options;
+
+        public int Run(FileWorkItemsOptions options)
```
harleenkohli: >public int Run(FileWorkItemsOptions options) [](start = 8, length = 44)

ryley says: If options are required for the functionality, shouldn't the validation (and setting of _options) happen in a constructor?

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,10 +10,40 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
             Required = true)]
-        public string TargetType { get; internal set; }
+        public string FilingTargetType { get; internal set; }
+
+        [Option(
+            'p',
+            "project-uri",
+            HelpText = "The absolute URI of the project under which the work items are to be filed.",
+            Required = true)]
+        public string ProjectUriString { get; internal set; }
```
harleenkohli: >public string ProjectUriString { get; internal set; } [](start = 8, length = 53)

we have certain pieces of info being carried in "propertybag" of result e.g. the areapath. Should the Project uri also be embedded in there instead? What value are we bringing by having the user of the multitool pass this information. If the "area-path" doesnt exist in the result.PropertyBag, then we simply cannot log a bug. So, why split these two bound pieces of info being fed from two different places.

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: oh, it's very unfortunate if you don't support the insert option (and the 'remove' option which is coming. also, we do want uriBaseId.

Can you get your mind around utilizing SingleFileOption. Would be preferable I think. in general, we want all our verbs to provide similar processing capabilities.

---
In reply to: [306578651](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306578651) [](ancestors = 306578651,306577540,306574608)

src/Sarif.Multitool/FileWorkItemsCommand.cs
michaelcfanning: This is the Commandline OSS component pattern. By convention, it creates and throws away a single instance. The design Ryley mentions could work, except that a ctor, obviously, cannot return an exit code (and is dedicated to initialization only). So to implement his suggestion, commandline lib would need to find and invoke the ctor, then find and invoke the Run method. I think they thought it more straightforward to simply use a single method. Also, it is a fine design to create an object that provides a guarantee of allowing for coherent initialization/operation against a method like this. So the design can be justified on its own terms, imo.

---
In reply to: [306578717](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306578717) [](ancestors = 306578717)

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: This is a really good question. The basic reason, I think, is that a user might want to 'quarantine' issues in a test project by using a mirror. i.e., you might file bugs against your test project, which mirrors actual and then switch over. And so this has a practical use. Area paths have to be managed in some way, because there isn't necessarily a one-to-one mapping of a log file to an area path (i.e., you might want to group different results in different paths).

Having said the above, though, I wonder if we could get away with a single area path in our first scenario and then we'll side-step this weird intermediate data decoupling you've noticed. I think it's a good idea (ie, go ahead and add an area path arg and keep this out of SARIF for now. we will revisit when we double click into the more complex metadata universe). @lgolding?

---
In reply to: [306581368](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306581368) [](ancestors = 306581368)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Done. This means I'll have to add code to support the other options. I sure hope there's a single API I can call to parse out the `--insert` option and enrich the SARIF file with the things it asks for.

---
In reply to: [306583029](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306583029) [](ancestors = 306583029,306578651,306577540,306574608)

src/Sarif.Multitool/FileWorkItemsCommand.cs
lgolding: I always supply it, to remind the reader that they should always consider which value to use.

---
In reply to: [306578696](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306578696) [](ancestors = 306578696)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: (In an email to the SARIF TC this morning I used the word "refactoring" in precisely this sense: I suggested pulling the elements of the SARIF format that are common to static and dynamic analysis into a separate base format, the idea being to ease maintenance by allowing bugs in the common subset to be fixed in exactly one place.)

---
In reply to: [306576671](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306576671) [](ancestors = 306576671,306575297)

src/Sarif.Multitool/MultitoolResources.resx
lgolding: Messages modified to mention incorrect value; "Error:" removed.

---
In reply to: [306578099](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306578099) [](ancestors = 306578099,306569098)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Will do later. Not as important as demonstrating bug filing this week.

---
In reply to: [306575068](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306575068) [](ancestors = 306575068)

src/Sarif.Multitool/FileWorkItemsCommand.cs
lgolding: Will do later. Not as important as demonstrating bug filing this week.

---
In reply to: [306575692](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306575692) [](ancestors = 306575692)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: Resolving duplicate of another comment that also asked for the enums (which I am deferring in the interest of getting bug filing working this week).

---
In reply to: [306577730](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306577730) [](ancestors = 306577730,306577599,306574126,306569347,306567622)

src/Test.UnitTests.Sarif.Multitool/FileWorkItemsCommandTests.cs
lgolding: Will do later. Not as important as getting bug filing working this week.

---
In reply to: [306578562](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306578562) [](ancestors = 306578562)

src/Sarif.Multitool/FileWorkItemsOptions.cs
lgolding: 1. If project URI is going to be in a property bag, it should be in `run.properties`, not `result.properties`. It's the same for all results in the run.
2. I did not understand the first paragraph of Michael's response. Let's mark this Pending and discuss further.

---
In reply to: [306584166](https://github.com/microsoft/sarif-sdk/pull/1583#discussion_r306584166) [](ancestors = 306584166,306581368)

PR #1587, opened 2019-07-24
File bugs to ADO

lgolding:
We create bugs in ADO from the SARIF Multitool, and populate dummy data into the Title, Description, Tags, and Area Path.

There's lots more to do including:

- Starting to refactor in some very nice design ideas that we've been kicking around today.
- Populating the bugs with real data.

This PR includes the following change suggested in a previous PR:

- Declare the --filtering-strategy and --grouping-strategy command line arguments as enums instead of strings, and rely on the CommandLine package's validation. (This has the drawback of making those options case-sensitive).
- Replace `[Theory]` with loop over test cases.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

PR #1589, opened 2019-07-24
New work item api mockup

michaelcfanning:
@lgolding this is a draft PR that shows the design endpoint that we discussed. Ignore everything in ProgramStubs, that's all the scaffolding that simply allows this code to compile. 

Please help review this to assess whether you feel like this is a usable pattern.

@harleenkohli, FYI


src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+        private static SarifLog RetrieveBaselineSarif(string v)
+        {
+            string[] resourceNames = typeof(Program).Assembly.GetManifestResourceNames();
+            Console.WriteLine(String.Join("\r\n", resourceNames));
+
+            string sarifText = GetResourceText("Sarif.Sdk.Sample.SampleTestFiles.Baseline.sarif");
+            return JsonConvert.DeserializeObject<SarifLog>(sarifText);
+        }
+
+        public static string GetResourceText(string resourceName)
```
lgolding: >GetResourceText [](start = 29, length = 15)

FYI Scott wrote a nice `ResourceExtractor` class.

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+                {
+                    text = reader.ReadToEnd();
+                }
+            }
+            return text;
+        }
+
+        private interface ISarifLogBaseliner
+        {
+            SarifLog CreateBaselinedSarifLog(SarifLog baselineSarif, SarifLog currentSarif, bool assignGuidsToNewResults);
```
lgolding: >assignGuidsToNewResults [](start = 97, length = 23)

Could use default argument (`assignGuidsToNewResults = false`) rather than overload -- I've never thought about which is preferable.

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+            SarifLog CreateBaselinedSarifLog(SarifLog baselineSarif, SarifLog currentSarif);
+        }
+
+        private class SarifLog
+        {
+            public IList<Run> Runs { get; internal set; }
+
+            internal static IEnumerable<WorkItemFilingMetadata> CreateWorkItemFilingMetadataFromLogs(IEnumerable<SarifLog> sarifLogsToFile)
```
lgolding: >CreateWorkItemFilingMetadataFromLogs [](start = 64, length = 36)

So this is called as `SarifLog.CreateWorkItemFilingMetadataFromLogs(sarifLogsToFile)`. I think I like it better as an extension method on `IEnumerable<SarifLog>`, so it's called as `sarifLogsToFile.CreateWorkItemFilingMetadata()`.

For that matter I like it best of all as an extension method on a single `SarifLog`, so:
```
IEnumerable<WorkItemFilingMetadata> metadatas /* yeah, I know */ = sarifLogsToFile.Select(log => log.CreateWorkItemFilingMetadata());
```

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
lgolding: Opinions vary but I'm persuaded by the argument that defaults get confusing when your API evolves to have many of them. The counterargument goes that with defaults, the caller is guaranteed that there is exactly one implementation and doesn't have to worry about bugs in the delegation from one overload to the next.

---
In reply to: [307081676](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307081676) [](ancestors = 307081676)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+
+        private class Result
+        {
+            public string Guid { get; internal set; }
+            public IList<Uri> WorkItemUris { get; internal set; }
+        }
+
+        private class GuidMappingVisitor
```
lgolding: >GuidMappingVisitor [](start = 22, length = 18)

Derived from `SarifRewritingVisitor`, I assume -- worth stating in the code here.

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
-            var adoClient = new AzureDevOpsClient();
-            var filingTarget = new AzureDevOpsFilingTarget(adoClient);
+            var guidMappingVisitor = new GuidMappingVisitor();
+            guidMappingVisitor.Visit(matchedSarif);
 
-            var workItemFiler = new WorkItemFiler(
-                filingTarget,
-                filteringStrategy: new NewResultsFilteringStrategy(),
-                groupingStrategy: new OneResultPerWorkItemGroupingStrategy());
+            // 4. Now we need to refactor the log file into one or more log files,
```
lgolding: >refactor [](start = 33, length = 8)

factor. Or split.

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
-                filteringStrategy: new NewResultsFilteringStrategy(),
-                groupingStrategy: new OneResultPerWorkItemGroupingStrategy());
+            // 4. Now we need to refactor the log file into one or more log files, 
+            // each of which will drive the creation of a single bug.
 
-            string logFilePath = Path.GetTempFileName();
-            File.WriteAllText(logFilePath, JsonConvert.SerializeObject(matchedLog));
+            IEnumerable<SarifLog> sarifLogsToFile = SarifSplitter.Split(
+                sarifLog: matchedSarif,
+                filterPredicate: (result) => { return result.BaselineState == BaselineState.New; },
```
lgolding: >filterPredicate [](start = 16, length = 15)

We don't want users to have to write their own predicates (I mean they can, but they shouldn't have to). So we should supply a set of predicates (or strategies, see next comment) so they can write either
```
filteringPredicate = FilteringPredicates.NewResults    // a static property of delegate type
```
or
```
filteringStrategy = new NewResultsFilteringStrategy()
```
or
```
filterStrategy = FilteringStrategies.NewResults // a pre-built NewResultsFilteringStrategy object
```

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
+                splittingStrategy: SplittingStrategy.OneFilePerRunPerRule);
```
lgolding: >splittingStrategy [](start = 16, length = 17)

What was your reasoning in using a bare function for `filterPredicate`, but a strategy class (with a single method) for `splittingStrategy`?

I like the strategy class approach because you can add more methods or properties later on without breaking existing callers. (At least, you can do that if all your strategy classes derive from an abstract base class, as the ones in my implementation do.)

lgolding: Bottom line, I like:
```
filterStrategy: FilteringStrategies.NewResults,
splittingStrategy: SplittingStrategies.OneFilePerRunPerRule
```
And I like `FilteringStrategies`/`SpilttingStrategies` (plural) so I can reserve `FilteringStrategy`/`SplittingStrategy` as the name of the abstract base class.

OTOH, I suppose you could put those prebuilt objects _in_ the abstract base class, but I don't like the idea of making the ABC aware of all its descendants. Let someone else have that knowledge.

---
In reply to: [307084776](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307084776) [](ancestors = 307084776)

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
 
-            // IMPORTANT! There's no facility defined yet for creating/configuring titles, area paths, etc.
+            // 5. Now we have a set of SARIF files, each of which will drive creation of a single bug
+            IEnumerable<WorkItemFilingMetadata> workItemsToFile = SarifLog.CreateWorkItemFilingMetadataFromLogs(sarifLogsToFile);
```
lgolding: >CreateWorkItemFilingMetadataFromLogs [](start = 75, length = 36)

See my comments on this API in the stubs file.

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
-                    foreach (Result result in resultGroup.Results)
-                    {
-                        bool matched = guidToMatchedResultMap.TryGetValue(result.Guid, out Result originalResult);
-                        Debug.Assert(matched);
-                        originalResult.WorkItemUris = result.WorkItemUris;
-                    }
-                }
-            }
-            finally
+            AdoWorkItemFiler adoWorkItemFiler = new AdoWorkItemFiler()
```
lgolding: >AdoWorkItemFiler [](start = 52, length = 16)

I presume that this and a `GitHubWorkItemFiler` would derive from an ABC `WorkItemFiler`. Why did you choose this approach rather than injecting a `WorkItemHost` abstraction (which is my current design)?

End users would not have to do they injection. We would provide a factory:
```
public static class WorkItemFilerFactory
{
    public static WorkItemFiler CreateAdoWorkItemFiler() { return new WorkItemFiler(new AdoWorkItemHost()); }
    public static WorkItemFiler CreateGitHubWorkItemFiler() { return new WorkItemFiler(new GitHubWorkItemHost(); }
}

Now the caller does this:
```
WorkItemFiler filer = WorkItemFilerFactory.CreateAdoWorkItemFiler();
```
Callers only have to worry about one class instead of one per host type.

File: src/Samples/Sarif.WorkItems.Sample/Program.cs [hunk 10]
```diff
@@ -5,112 +5,83 @@
...
-        }
+                FileAttachments = true
+            };
 
-        private static SarifLog RetrieveBaselineSarif()
-        {
-            string[] resourceNames = typeof(Program).Assembly.GetManifestResourceNames();
-            Console.WriteLine(String.Join("\r\n", resourceNames));
+            // 6. Wire up an event handler that gets called when the bug is filed
+            //    This callback will retrieve the
```
lgolding: >retrieve the [](start = 37, length = 12)

unfinished...

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            internal async Task FileWorkItems(IEnumerable<WorkItemFilingMetadata> workItemsToFile)
+            {
+                throw new NotImplementedException();
+            }
+
+            public class WorkItemFiledOrUpdatedArgs
+            {
+                public WorkItemFilingMetadata Metadata;
+            }
```
lgolding: >} [](start = 12, length = 1)

YAGNI, but at some point we might find we want a flag to tell whether it was filed or updated.

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+            try
+            {
+                IEnumerable<ResultGroup> resultGroups = workItemFiler.FileWorkItems(logFilePath).Result;
+                foreach (ResultGroup resultGroup in resultGroups)
+                {
+                    foreach (Result result in resultGroup.Results)
+                    {
+                        bool matched = guidToMatchedResultMap.TryGetValue(result.Guid, out Result originalResult);
+                        Debug.Assert(matched);
+                        originalResult.WorkItemUris = result.WorkItemUris;
```
lgolding: >originalResult [](start = 24, length = 14)

Ok, reference equals one of the `Result` objects in the `matchedLog`. Worth a comment.

src/Samples/Sarif.WorkItems.Sample/Program.cs
```diff
@@ -5,112 +5,83 @@
...
 
-        public static string GetResourceText(string resourceName)
-        {
-            string text = null;
+                Debug.Assert(filedSarifLog.Runs.Count == 1);
+                foreach (Result result in filedSarifLog.Runs[0].Results)
+                {
+                    bool matched = guidToResultMap.TryGetValue(result.Guid, out Result originalResult);
+                    Debug.Assert(matched);
+                    originalResult.WorkItemUris = new Uri[] { metadata.Uri };
```
lgolding: >originalResult [](start = 20, length = 14)

It's worth a comment that this is reference-equal to a result in the `matchedLog`, so when you write it out below, the work item URIs go along for the ride.

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
```diff
@@ -0,0 +1,233 @@
...
+            public List<object> Attachments { get; internal set; }
+        }
+
+
+        private static void LogFailure(Exception exception)
+        {
+            throw new NotImplementedException();
+        }
+
+        private class AdoWorkItemFiler
```
lgolding: >AdoWorkItemFiler [](start = 22, length = 16)

See my comments in the program -- I'd prefer a plain `WorkItemFiler` with an injected `WorkItemHost`-derived object (`new AdoWorkItemHost()` or `new GitHubWorkItemHost()`.

lgolding:
Summary: I agree with the overall shape. I have several specific suggestions.

src/Samples/Sarif.WorkItems.Sample/Program.cs
lgolding: Yes, it has to be that way. If someone writes their own strategy and derives from the ABC, you don't want them to have to carry along the dependencies (if any) of all the out-of-the-box concrete classes. Ok, fine, they are probably dragging those dependencies along anyway because the same assembly that contains the ABC (e.g. `FilteringStrategy` also contains the static holder `FilteringStrategies`. So I guess that's not a strong argument. I just don't like the idea that a class knows its descendants.

---
In reply to: [307085292](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307085292) [](ancestors = 307085292,307084776)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: thanks for the information. none of the program stubs code is 'live', this is all throwaway.

---
In reply to: [307081180](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307081180) [](ancestors = 307081180)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: These API signatures are speculative, not crafted. I actually decided to remove this arg from the use case, which is what you should be looking at (i.e., the sample is the truth). Guids are something that s/be added or removed just as other data is handled.

---
In reply to: [307082989](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307082989) [](ancestors = 307082989,307081676)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: This is throwaway code, as mentioned in the description and at the top of this file, this content was not intended to be reviewed. Instead, my hope was that people would focus on the shape of the API as it is used. This is the focus of this exercise.

---
In reply to: [307083504](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307083504) [](ancestors = 307083504)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: Yes, there should be a vanilla work item filer, but we should also create an instance of this that injected the standard AdoWorkItemHost(). This is more usable for API consumers (again, this is a top-down exercise of trying to get the API use case in the best form for consumers). 

---
In reply to: [307088864](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307088864) [](ancestors = 307088864)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: Yes, an extension method had occurred to me as well. The API usability issue for this is typically one of discovery. IOW, if we do create this method, it should live in the core namespace (so that it appears in IntelliSense in expected situations)..

---
In reply to: [307082555](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307082555) [](ancestors = 307082555)

src/Samples/Sarif.WorkItems.Sample/Program.cs
michaelcfanning: yes.

---
In reply to: [307083636](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307083636) [](ancestors = 307083636)

src/Samples/Sarif.WorkItems.Sample/Program.cs
michaelcfanning: Filtering is a predicate because filtering is naturally a true/false operation (is this thing in or out)? Splitting is entirely different because you are reorganizing multiple sets into arbitrary other sets. I don't see filtering as benefiting from object inheritance. Someone could create a static predicate instance that bolts together other static predicate instances though. I need to think more about strategy injection into this splitter class. The current design simply provides an enum to drive the specific operation you want. Again, the reorganization that's implied here is so arbitrary that I can't see the need for layering. i.e., you basically want an object with a Split() method that accepts one log file and produces many. From this angle, the API should actually instantiate and invoke one of these concrete types, rather than parameterizing. Let me think a bit more.

---
In reply to: [307089548](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307089548) [](ancestors = 307089548,307085292,307084776)

src/Samples/Sarif.WorkItems.Sample/Program.cs
michaelcfanning: This all assumes there will be no difference in helpers provided for an ADO-specific work item filer and I predict this is not the case. And so (new AdoWorkItemFiler()) that provides some core functionality and additional content is very likely. ie. basic object inheritance.  Glad to discuss further. This sort of API evolution is simple to absorb (i.e., breaking a core type into more granular derived types), so I don't see it as an urgent topic.

---
In reply to: [307087340](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307087340) [](ancestors = 307087340)

src/Samples/Sarif.WorkItems.Sample/Program.cs
michaelcfanning: Yes.

---
In reply to: [307084390](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307084390) [](ancestors = 307084390)

src/Samples/Sarif.WorkItems.Sample/ProgramStubs.cs
michaelcfanning: Agreed. We haven't really dug into the contents of the metadata class because this API usability study is strictly around 'file all new bugs'. A next use case might add 'Updated' to compare/contrast. A further use case would consider 'close resolved items'. I actually think that latter case is most interesting to us in the short-term.

---
In reply to: [307087950](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307087950) [](ancestors = 307087950)

src/Samples/Sarif.WorkItems.Sample/Program.cs
michaelcfanning: Yes. Don't you have a comment, however, on the usability of this pattern? i.e., our machinery requires the user to handle the association between the original and split SARIF log files. this is a design weakness, in my opinion.

---
In reply to: [307088666](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307088666) [](ancestors = 307088666)

lgolding: Well yes, I suppose if the user's happy path code requires a comment then it's not as usable as it should be.

---
In reply to: [307391618](https://github.com/microsoft/sarif-sdk/pull/1589#discussion_r307391618) [](ancestors = 307391618,307088666)

PR #1591, opened 2019-07-26
Scrub core functionality of some obvious non-determinisms in output.

michaelcfanning:


michaelcfanning:
This change eliminates some undesirable verbose output from the core SDK. It is a necessary preparatory change to finalizing a large feature to cache results by file hash. Changes include:

1) Remove mime-type from log files. This information is exclusively gleaned from file names in all use cases to date. We don't need to inject it by default.
2) Eliminate output that provided an 'Analyzing blah.txt' message for each scan target. This is a verbose message that fired in non-verbose mode. It is arguably completely unnecessary in any case, as a verbose analysis in the default driver produces an explicit message for each scan target (i.e., non-failure cases result in explicit pass/other messages).
3) We already had a 'NondeterministicProperties' option in optional data. Now we have a '--remove' argument that honors this setting and eliminates timestamps when passed. We will build this out for other properties later. 

@lgolding

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -402,46 +411,6 @@ private void CaptureFile(ArtifactLocation fileLocation)
...
-
-            // Analyzing target '{0}'...
-
-            if (context == null)
-            {
-                throw new ArgumentNullException(nameof(context));
-            }
-
-            string message = string.Format(CultureInfo.InvariantCulture, 
-                SdkResources.MSG001_AnalyzingTarget,
```
lgolding: Can we get rid of this resource?

src/Sarif.Driver/Sdk/CommonOptionsBase.cs
```diff
@@ -31,6 +31,15 @@ public class CommonOptionsBase
             "and NondeterministicProperties.")]
         public IEnumerable<OptionallyEmittedData> DataToInsert { get; set; }
 
+        [Option(
+            "remove",
+            Separator = ';',
+            HelpText =
+            "Optionally present data, expressed as a semicolon-delimited list, that should be not be persisted to or which "  +
+            "should be removed from the log file. Valid values include Hashes, TextFiles, BinaryFiles, EnvironmentVariables, " +
```
lgolding: Love me some parallel construction!

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -62,7 +63,10 @@ public class SarifLogger : IDisposable, IAnalysisLogger
                 }
             }
 
-            var invocation = Invocation.Create(dataToInsert.HasFlag(OptionallyEmittedData.EnvironmentVariables), invocationPropertiesToLog);
+            var invocation = Invocation.Create(
+                emitMachineEnvironment: dataToInsert.HasFlag(OptionallyEmittedData.EnvironmentVariables),
+                emitTimestamps: !dataToRemove.HasFlag(OptionallyEmittedData.NondeterministicProperties),
```
lgolding: >emitTimestamps [](start = 16, length = 14)

Not for now, perhaps, but as you build this out, I wonder whether `Invocation.Create` should just take the `dataToInsert` and `dataToRemove` objects directly, rather than a growing set of individual flags. `Invocation.Create` can do the `HasFlag` checks internally.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -243,7 +251,8 @@ public void AnalysisStarted()
 
         public void AnalysisStopped(RuntimeConditions runtimeConditions)
         {
-            if (_run.Invocations != null && _run.Invocations.Count > 0)
+            if (_run.Invocations != null && _run.Invocations.Count > 0 &&
+                !_dataToRemove.HasFlag(OptionallyEmittedData.NondeterministicProperties))
```
lgolding: >NondeterministicProperties [](start = 61, length = 26)

Why do we assign `EndTimeUtc` in two different places? And if we have to do it, why is the code that decides whether to assign it different in the two places? Doesn't 
```
if (_run?.Invocations?.Count > 0 && !_dataToRemove.HasFlag(OptionallyEmittedData.NondeterministicProperties))
```
work in both cases? And if it does work in both places, and this code needs to exist in both places, we should DRY it out into `ShouldEmitEndTime`.

(In fact, I often DRY out logic like that into a helper even if it only exists once, for documentation purposes.)

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: No, I agree, we should do this. If we're breaking this helper, let's break it once.

---
In reply to: [307786117](https://github.com/microsoft/sarif-sdk/pull/1591#discussion_r307786117) [](ancestors = 307786117)

michaelcfanning: Oops, abort mission. Problem is this help has a mixture of mechanisms to specify what's included, the invocation properties log and these bools. Really should have one converged solution. Let's defer. Agree it needs improvement.

---
In reply to: [307853676](https://github.com/microsoft/sarif-sdk/pull/1591#discussion_r307853676) [](ancestors = 307853676,307786117)

PR #1596, opened 2019-07-29
Add simple file splitting to support work item filing.

michaelcfanning:
@harleenkohli @SurbhiMS


src/Sarif/Visitors/PerRunPerRuleSplittingVisitor.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public class PerRunPerRuleSplittingVisitor : SarifRewritingVisitor
```
michaelcfanning: >PerRunPerRuleSplittingVisitor [](start = 17, length = 29)

Note that our actual intended splitting strategy is per run per target per rule. That will come in the next change.

michaelcfanning:
This change works incrementally towards our final design. The newer design uses a visitor to perform the file splitting. This visitor accepts a predicate for general filtering. I've removed the command-line arguments so that we can get the core scenario working against defaults. When we have a finished design, we will re-add these arguments.

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -48,17 +49,27 @@ public int Run(FileWorkItemsOptions options, IFileSystem fileSystem)
 
             SarifLog sarifLog = JsonConvert.DeserializeObject<SarifLog>(logFileContents);
 
-            FilteringStrategy filteringStrategy = FilteringStrategyFactory.CreateFilteringStrategy(options.FilteringStrategy);
```
harleenkohli: >FilteringStrategy filteringStrategy [](start = 12, length = 35)

why did we decide to remove the support for user passing these values? My understanding for the design was: that the user would control the strategy and then splitter logic "splits" based on the user input.

michaelcfanning: Our design has a built-in default: we filter by 'new' and 'updated' and we split by 'per run per target per rule'. Our emerging pipeline will use these defaults. I've eliminated the arguments and some supporting files as many of them are obsoleted by the current design. my thought is to get the defaults working and then to figure out how to make them generally configurable. this work technique is designed to avoid YAGNI and other costs related to reworking things that aren't immediately required.

---
In reply to: [308471467](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308471467) [](ancestors = 308471467)

src/Sarif/Visitors/PerRunPerRuleSplittingVisitor.cs
harleenkohli: thanks for the clarification because it confused me. so current implementation will log a single bug for every unique issue we are seeing across all 150+ build definitions (for testing).

PS: i wonder if the word "perRun" is adding any value here? wouldn't the bugs be logged always on the per run basis? (assuming the filtering strategy filters out the new ones)

---
In reply to: [308470724](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308470724) [](ancestors = 308470724)

src/Sarif.Multitool/FileWorkItemsCommand.cs
harleenkohli: ok

---
In reply to: [308472672](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308472672) [](ancestors = 308472672,308471467)

src/Sarif/FilteringStrategies.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public static class FilteringStrategies
+    {
+        public static Func<Result, bool> NewOrUpdatedResults =
+            (result) =>
+            {
+                return result.BaselineState == BaselineState.New ||
+                       result.BaselineState == BaselineState.Updated;
```
harleenkohli: >BaselineState == BaselineState.Updated [](start = 30, length = 38)

I wonder why we decided to start with new and updated rather then just new. i know this is just one of the strategies - just curious - would an updation in the code warrant a new bug?

src/Sarif.WorkItemFiling/WorkItemFilingMetadata.cs
```diff
@@ -14,14 +14,7 @@ public class WorkItemFilingMetadata
...
-
-        /// <summary>
-        /// The results that should be filed as a single work item.
-        /// </summary>
-        /// <remarks>
-        /// TEMPORARY. WHAT WE REALLY WANT HERE IS AN ATTACHMENT, A SARIF FILE WITH THE RELEVANT RESULTS.
-        /// </remarks>
-        public IList<Result> Results { get; set; }
+        public Attachment Attachment { get; set; }
+        public object Tag { get; set; }
```
harleenkohli: >public object Tag [](start = 8, length = 17)

this is a bit confusing since we also have a concept of tags in workitems and in fact have another properties "tags" in this object which have a very specific meaning. we should have a better name for this and/or provide a description to understand what exactly this object is.

harleenkohli: based on the usage, it appears this is the SarifFile, right? why not name it so?

---
In reply to: [308475219](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308475219) [](ancestors = 308475219)

src/Sarif/Visitors/PerRunPerRuleSplittingVisitor.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        public PerRunPerRuleSplittingVisitor(Func<Result, bool> filteringStrategy = null)
```
harleenkohli: >PerRunPerRuleSplittingVisitor [](start = 15, length = 29)

i wonder if the word "perRun" is adding any value here? wouldn't the bugs be logged always on the per run basis? (assuming the filtering strategy filters out the new ones)

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -48,17 +49,27 @@ public int Run(FileWorkItemsOptions options, IFileSystem fileSystem)
...
+                if (sarifLog.Runs[runIndex]?.Results?.Count > 0)
                 {
-                    // TODO: THESE TWO LINES ARE REPLACED BY A CALL TO SarifSplitter.Split with the specified strategies (or predicates).
-                    // It would return a set of SarifLog objects, rather than a set of WorkItemFilingMetadata objects as I currently have it.
-                    IList<Result> filteredResults = filteringStrategy.FilterResults(sarifLog.Runs[i].Results);
-                    IList<WorkItemFilingMetadata> workItemMetadata = groupingStrategy.GroupResults(filteredResults);
+                    var visitor = new PerRunPerRuleSplittingVisitor();
+                    visitor.VisitRun(sarifLog.Runs[runIndex]);
+
+                    IList<WorkItemFilingMetadata> workItemMetadata = new List<WorkItemFilingMetadata>(visitor.SplitSarifLogs.Count);
```
harleenkohli: >visitor.SplitSarifLogs.Count [](start = 102, length = 28)

wow, TIL we could specify capacity!

src/Test.UnitTests.Sarif/Visitors/PerRunPerRuleSplittingVisitorTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        [Fact]
+        public void PerRunPerRuleSplittingVisitor_EmptyLog()
+        {
+            var visitor = new PerRunPerRuleSplittingVisitor();
+            visitor.VisitRun(new Run());
+            visitor.SplitSarifLogs.Count.Should().Be(0);
+        }
+
+        [Fact]
+        public void PerRunPerRuleSplittingVisitor_Simple()
```
harleenkohli: >PerRunPerRuleSplittingVisitor_Simple [](start = 20, length = 36)

these tests could be named more aptly - i.e. i cant tell quickly the difference between "simple" and "default".

src/Test.UnitTests.Sarif/Visitors/PerRunPerRuleSplittingVisitorTests.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            visitor.SplitSarifLogs[0].Runs[0].Results.Count.Should().Be(1);
+            visitor.SplitSarifLogs[0].Runs[0].Results[0].RuleId.Should().Be(TestConstants.RuleIds.Rule1);
+
+            visitor.SplitSarifLogs[1].Runs[0].Results.Count.Should().Be(1);
+            visitor.SplitSarifLogs[1].Runs[0].Results[0].RuleId.Should().Be(TestConstants.RuleIds.Rule2);
+        }
+
+        [Fact]
+        public void PerRunPerRuleSplittingVisitor_Default()
```
harleenkohli: >PerRunPerRuleSplittingVisitor_Default [](start = 20, length = 37)

we could add a test for multiple runs and same rule

src/Sarif/FilteringStrategies.cs
michaelcfanning: Agreed, let's keep this entirely focused on what's actually in play.

---
In reply to: [308474361](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308474361) [](ancestors = 308474361)

src/Sarif.WorkItemFiling/WorkItemFilingMetadata.cs
michaelcfanning: I will rename this as Object. the point here is for this to be entirely decoupled from SARIF. :) it's very confusing to have 'Tag' and 'Tags' on the class, that was just an oversight.

---
In reply to: [308475568](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308475568) [](ancestors = 308475568,308475219)

src/Sarif/Visitors/PerRunPerRuleSplittingVisitor.cs
michaelcfanning: you could imagine a single bug per log file. imagine a complete SDL run against a small project, where you file a single bucket bug. also, a log file could have multiple runs of the same tool, which could be aggregated. 

---
In reply to: [308476793](https://github.com/microsoft/sarif-sdk/pull/1596#discussion_r308476793) [](ancestors = 308476793)

PR #1599, opened 2019-07-31
Update default filtering

michaelcfanning:
@rtaket


src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -30,8 +30,14 @@ public static WorkItemFilingMetadata CreateWorkItemFilingMetadata(this SarifLog
             string areaPath = $@"{workItemProjectName}" + artifactLocation.GetProperty("AreaPath").Replace(@"\\", @"\");
             string buildDefinitionName = artifactLocation.GetProperty("BuildDefinitionName");
 
+            Result result = sarifLog.Runs[0].Results[0];
             string ruleName = sarifLog.Runs[0].Results[0].RuleId;
```
rtaket: >RuleId [](start = 58, length = 6)

For some reason I thought Result.RuleId wasn't required? Can't a Result alternatively specify Result.Rule.Id?

PR #1612, opened 2019-08-05
Fix #1680 (regex perf in WebRequest.TryParse) by avoiding backtracking.

lgolding:
The bug was exposed by an actual web request which left the `=` sign off of the last in a long list of query parameters:
```
/getSomething?firstName=John&lastName=Smith&...&Address2=&Address3
```
Because our regex for query parameters expects an `=` after each query parameter name, the lack of the final `=` caused the engine to backtrack all the way to the beginning. By turning off backtracking within the individual parameter names and values, and by eliminating an unnecessary alternation, we minimize the number of states the engine has to save.

### DIFF
--- a/src/Sarif/Core/WebRequest.cs
+++ b/src/Sarif/Core/WebRequest.cs
@@ -112,30 +112,37 @@ namespace Microsoft.CodeAnalysis.Sarif
             return uri;
         }
 
+        // In this query pattern, note how backtracking is disabled with (?>...) while parsing the
+        // parameter names and values. A parameter name can't contain a '=', and a parameter value
+        // can't contain a '&', so it's never necessary to backtrack over an individual parameter
+        // name or value.
         const string QueryPattern =
             @"^
-              \?                                # A literal '?', followed by zero or more
-              (?<query>
+              \?                        # A literal '?', followed by zero or more of...
+              (
+                (?>
+                  (?<name>[^=]+)        # a parameter name (everything that's not an = sign),
+                )
+                =                       # an equal sign, and
                 (
-                  (?>(?<name>[^=]+))            # parameter name (everything that's not an = sign),
-                  =                             # an equal sign, and
-                  (
-                    (?>(?<value>[^&]*))         # the value (everything that's not an '&')...
-                    &?                          # and an '&' (except after the last one).
+                  (?>
+                    (?<value>[^&]*)     # the value (everything that's not an '&')...
                   )
-                )*
-                |                               # RFC 3986 does _not_ require a URI's query to
-                (?<nonParameterizedQuery>.*)    # consist of key-value pairs. If it doesn't,
-                                                # don't fail.
-              )$";
+                  &?                    # and an '&' (except after the last one).
+                )
+              )*
+              $";
 
         private static readonly Regex s_queryRegex = SarifUtilities.RegexFromPattern(QueryPattern);
 
         private static IDictionary<string, string> ParseParametersFromQueryString(string query)
         {
-            Dictionary<string, string> parameterDictionary = null;
+            IDictionary<string, string> parameterDictionary = new Dictionary<string, string>();
 
             Match match = s_queryRegex.Match(query);
+
+            // RFC 3986 does _not_ require a URI's query to consist of key-value pairs. If it
+            // doesn't, don't fail; just return an empty parameter dictionary.
             if (match.Success)
             {
                 List<string> names = match.Groups["name"].Captures.Cast<Capture>().Select(c => c.Value).ToList();
--- a/src/Sarif/Core/WebRequest.cs
+++ b/src/Sarif/Core/WebRequest.cs
@@ -114,13 +114,13 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         const string QueryPattern =
             @"^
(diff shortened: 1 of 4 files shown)


src/Sarif/Core/WebRequest.cs
```diff
@@ -114,13 +114,13 @@ private static Uri SynthesizeAbsoluteUriFrom(Uri uri)
 
         const string QueryPattern =
             @"^
-              \?                              # A literal '?', followed by zero or more
+              \?                                # A literal '?', followed by zero or more
               (?<query>
                 (
-                  (?<name>[^=]+)                # parameter name (everything that's not an = sign),
+                  (?>(?<name>[^=]+))            # parameter name (everything that's not an = sign),
```
lgolding: >(?> [](start = 18, length = 3)

Disables backtracking in the enclosed subexpression.

src/Test.UnitTests.Sarif/Core/WebRequestTests.cs
```diff
@@ -184,5 +185,27 @@ public void WebRequest_TryParse_ReturnsFalseOnInvalidRequest()
...
+Referer: https://login.somedomain.com/77777777777777777777/oauth2/authorize?client_id=7777&response_mode=form_post&response_type=code+id_token&scope=openid+profile&state=somestate
+Accept-Encoding: gzip, deflate, br
+Accept-Language: en-US,en;q=0.9
+Cookie: ARRAffinity=somecode; .AspNet.Cookies=somecode
+
+";
+            Action action = () => WebRequest.TryParse(RequestString, out _);
+
+            // On my machine this takes about 7 msec. Leaving a 5x safety factor.
+            action.ExecutionTime().Should().BeLessOrEqualTo(35.Milliseconds());
```
lgolding: >ExecutionTime [](start = 19, length = 13)

FluentAssertions FTW.

src/Test.UnitTests.Sarif/Core/WebRequestTests.cs
```diff
@@ -184,5 +185,27 @@ public void WebRequest_TryParse_ReturnsFalseOnInvalidRequest()
             succeeded.Should().BeFalse();
             webRequest.Should().BeNull();
         }
+
+        [Fact]
+        public void WebRequest_TryParse_HasAcceptablePerformance()
+        {
+            const string RequestString = @"GET /getSomethings?FirstName=test&LastName=test&AddressLine1=555%20110th%20Ave%20NE&City=Bellevue&Country=USA&PostalCode=98004&EmailAddress=test@somedomain.com&BusinessPhone=12345678901&MobilePhone=1234567890&AddressLine2=&AddressLine3 HTTP/1.1
```
lgolding: >3 [](start = 277, length = 1)

Note the lack of an `=` after this parameter.

lgolding: >RequestString [](start = 25, length = 13)

A sanitized version of the actual customer web request that exposed the problem.

src/Test.UnitTests.Sarif/Core/WebRequestTests.cs
```diff
@@ -184,5 +185,27 @@ public void WebRequest_TryParse_ReturnsFalseOnInvalidRequest()
...
+Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
+Referer: https://login.somedomain.com/77777777777777777777/oauth2/authorize?client_id=7777&response_mode=form_post&response_type=code+id_token&scope=openid+profile&state=somestate
+Accept-Encoding: gzip, deflate, br
+Accept-Language: en-US,en;q=0.9
+Cookie: ARRAffinity=somecode; .AspNet.Cookies=somecode
+
+";
+            Action action = () => WebRequest.TryParse(RequestString, out _);
+
+            // On my machine this takes about 7 msec. Leaving a 5x safety factor.
```
lgolding: >factor [](start = 74, length = 6)

It's still too long, but before, it was taking dozens of seconds. This should at least provide acceptable performance for our customer, and we can look at further optimizations latest.

src/Sarif/Core/WebRequest.cs
```diff
@@ -112,30 +112,37 @@ private static Uri SynthesizeAbsoluteUriFrom(Uri uri)
...
-                  (?<name>[^=]+)                # parameter name (everything that's not an = sign),
-                  =                             # an equal sign, and
-                  (
-                    (?<value>[^&]*)             # the value (everything that's not an '&')...
-                    &?                          # and an '&' (except after the last one).
+                  (?>
+                    (?<value>[^&]*)     # the value (everything that's not an '&')...
                   )
-                )*
-                |                               # RFC 3986 does _not_ require a URI's query to
```
lgolding: >| [](start = 16, length = 1)

This alternation was not necessary, because if the query string didn't match the regex, we simply didn't populate the parameters dictionary. Removing it further improves performance (the engine never has to backtrack to the beginning to try the (trivial) alternative).

@mikefan Note that this is a PR comment, not a code comment, because the code it describes is gone.

lgolding:

lgolding:

PR #1615, opened 2019-08-06
Fix #1600: Matched result should use current suppressions.

### LINKED ISSUE #1600: result.suppressions property doesn't prefer most current suppressions info during result matching
Originally reported by Nathan Davidson of MS.

Original text: 

We’ve encountered what seems to be an unexpected behavior of the SARIF SDK result matching. Not sure if it’s unintended behavior though, or a bad assumption on our side. Here’s the test case:

Given:
•	SARIF Log 1 containing Result A, where the suppressions property of Result A is null
•	SARIF Log 2 containing Result A, where the suppressions property of Result A is set to a non-null value (like [{ "kind": "external" }])

When:
•	Result matching is run with Log 1 as the baseline, and Log 2 as the current

Then:
•	(Expected) The suppressions property of Result A in the output from result matching is set to the newer (non-null) value from Log 2
•	(Actual) The suppressions property of Result A in the output from result matching is null

Is the stated expectation here valid (meaning there’s a bug in the SDK) or invalid? Since we assumed this behavior, we’re now seeing some suppressed results being filed to ADO as bugs in the case where we’ve already processed an earlier log (which is when we use result matching).

lgolding:
This was an oversight in the new result matcher. The fix is one line. Everything else is test code and bookkeeping.

### DIFF
--- a/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
+++ b/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
 {
@@ -130,7 +129,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
             // Result exists.
             Result result = CurrentResult.Result.DeepClone();
             result.CorrelationGuid = PreviousResult.Result.CorrelationGuid;
-            result.Suppressions = PreviousResult.Result.Suppressions;
+            result.Suppressions = CurrentResult.Result.Suppressions;
             result.BaselineState = BaselineState.Unchanged;
 
             if (!PreviousResult.Result.TryGetProperty(SarifLogResultMatcher.ResultMatchingResultPropertyName, out originalResultMatchingProperties))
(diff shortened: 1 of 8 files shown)


lgolding:

lgolding:

lgolding:

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -130,7 +129,7 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
             // Result exists.
             Result result = CurrentResult.Result.DeepClone();
             result.CorrelationGuid = PreviousResult.Result.CorrelationGuid;
-            result.Suppressions = PreviousResult.Result.Suppressions;
+            result.Suppressions = CurrentResult.Result.Suppressions;
```
rtaket: >Suppressions [](start = 55, length = 12)

Just curious why the DeepClone on line 130 doesn't copy over the suppressions.

lgolding: It does, good catch. I removed this line. You only need to be explicit about the things that _don't_ come from `CurrentResult`. And frankly I'm curious why it takes `CorrelationGuid` from the previous result. The spec says that `CorrelationGuid` groups results into "equivalence classes"; that is, two results are "logically identical" if they have the same `CorrelationGuid` and "logically distinct" if they have different `CorrelationGuid`s. So the matcher should never associate two results with different `CorrelationGuids`. I filed #1616, "Result matcher should never associate results with different CorrelationGuids". 

---
In reply to: [311116348](https://github.com/microsoft/sarif-sdk/pull/1615#discussion_r311116348) [](ancestors = 311116348)

PR #1618, opened 2019-08-06
Various multitool work item filing fixes

rtaket:


src/Sarif.WorkItemFiling/AzureDevOpsFilingTarget.cs
```diff
@@ -108,18 +108,28 @@ public override async Task<IEnumerable<WorkItemFilingMetadata>> FileWorkItems(IE
...
                 try
                 {
+                    Console.Write($"Creating work item: {metadata.Title}");
                     workItem = await _witClient.CreateWorkItemAsync(patchDocument, project: _projectName, "Bug");
+                    Console.WriteLine($": {workItem.Id}: DONE");
                 }
                 catch (Exception)
                 {
                     // TBD error handling
                     throw;
```
cfaucon: Think we should implement logging here and not throw.

src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -47,7 +47,7 @@ public static WorkItemFilingMetadata CreateWorkItemFilingMetadata(this SarifLog
                             ruleName + ": Exposed credential(s) in " +
                              "build definition: '" + buildDefinitionName + "'";
 
-            metadata.Tags = new List<string>(new string[] { "security" });
+            metadata.Tags = new List<string>(new string[] { "Security" });
```
cfaucon: TODO: This should come from the SARIF or command line arg in the future.

src/Sarif.WorkItemFiling/WorkItemFields.cs
```diff
@@ -24,5 +24,10 @@ public static class WorkItemFields
         /// Title field name.
         /// </summary>
         public const string Title = "System.Title";
+
+        /// <summary>
+        /// Repro Steps field name.
+        /// </summary>
+        public const string ReproSteps = "Microsoft.VSTS.TCM.ReproSteps";
```
cfaucon: This seems specific to ADO, perhaps like AzureDeveOpsFilingTarget this class should be AzureDevOpsWorkItemFields as it seems specific to ADO and not generic across different filing targets.

src/Sarif/Visitors/PerRunPerTargetPerRuleSplittingVisitor.cs
```diff
@@ -10,38 +10,42 @@ public class PerRunPerTargetPerRuleSplittingVisitor : SplittingVisitor
...
 
         public override Result VisitResult(Result node)
         {
             if (!FilteringStrategy(node)) { return node; }
 
-            string ruleId = node.RuleId;
-
-            if (string.IsNullOrEmpty(ruleId) && node.RuleIndex > -1)
+            string ruleId;
+            if (node.RuleIndex > -1 && CurrentRun.Tool.Driver.Rules.Count > node.RuleIndex)
```
cfaucon: Do any of these need ? to protect from null

src/Sarif/Visitors/PerRunPerTargetPerRuleSplittingVisitor.cs
```diff
@@ -10,38 +10,42 @@ public class PerRunPerTargetPerRuleSplittingVisitor : SplittingVisitor
...
             {
                 ruleId = CurrentRun.Tool.Driver.Rules?[node.RuleIndex].Id;
             }
+            else
+            {
+                // Remove the rule pattern index from the rule id. e.g. CSCAN0230/5
+                ruleId = node.RuleId.Substring(0, node.RuleId.LastIndexOf('/'));
```
cfaucon: Protect against indexOf -1

src/Sarif/Visitors/PerRunPerTargetPerRuleSplittingVisitor.cs
```diff
@@ -10,38 +10,42 @@ public class PerRunPerTargetPerRuleSplittingVisitor : SplittingVisitor
...
+            }
 
             ArtifactLocation artifactLocation = s_emptyArtifactLocation;
             if (node.Locations[0].PhysicalLocation?.ArtifactLocation != null)
             {
                 artifactLocation = node.Locations[0].PhysicalLocation?.ArtifactLocation;
             }
 
-            if (!_targetToRuleMap.TryGetValue(artifactLocation, out Dictionary< string, SarifLog > ruleToSarifLogMap))
+            if (!_targetToRuleMap.TryGetValue(artifactLocation.Uri.ToString(), out Dictionary<string, SarifLog> ruleToSarifLogMap))
```
cfaucon: Does the default s_emptyArtifactLocation have a populated Uri, if not protect against null ToString call.

PR #1625, opened 2019-08-07
Fix #1340: Multitool commands do not respect --force option

### LINKED ISSUE #1340: Multitool --force option: advertised default is false, actual default is true.
Repeat invocations of `smt transform input.sarif --output transformed.sarif` will overwrite the output file.

Expected: Since --force flag was not set, transformed.sarif should not be overwritten.

lgolding:
All but two of the Sarif.Multitool verbs ignored the `force` option and always overwrote the output file. (The two good citizens were `convert` and `file-work-items`.)

We fix this by introducing a helper method `DriverUtilities.VerifyOutputFileCanBeCreated` which throws if the output file exists and `force` is not set.

Four of the verbs deserve further consideration:

- `rebaseuri`, `absoluteuri`, and `result-match-set` produce multiple output files. We check for the existence of each file before we overwrite it. The result can be that we successfully create the first few output files, then throw on the next one.
- `page`: This command has `force` set to `true` by default. I don't understand the argument in favor of this (see the comment in PageOptions.cs), and there seems to be no way to provide an explicitly `false` value to a "switch" option in the command line parsing package we use.


src/Sarif.Driver/Sarif.Driver.csproj
```diff
@@ -11,6 +11,22 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
+  <ItemGroup>
+    <EmbeddedResource Update="DriverResources.resx">
```
lgolding: >DriverResources [](start = 30, length = 15)

Still yet another case where, when we converted our project files to .NET Core format, we lost the ability to edit the .resx file and have the .Designer.cs file automatically updated. I thought we had them all...

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -473,7 +473,7 @@ private HashSet<Skimmer<TContext>> CreateSkimmers(TContext context)
 
             try
             {
-                skimmers = DriverUtilities.GetExports<Skimmer<TContext>>(DefaultPlugInAssemblies);
+                skimmers = CompositionUtilities.GetExports<Skimmer<TContext>>(DefaultPlugInAssemblies);
```
lgolding: >CompositionUtilities [](start = 27, length = 20)

Renamed this class because it was in a file named CompositionUtilities.cs. I put my new helper method in a class named `DriverUtilities` in a same-named file.

src/Sarif.Multitool/ValidateCommand.cs
```diff
@@ -45,9 +45,6 @@ protected override SarifValidationContext CreateContext(ValidateOptions options,
             return context;
         }
 
-        private static string s_CurrentSchemaLocation =
```
lgolding: >s_CurrentSchemaLocation [](start = 30, length = 23)

Unused.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -316,6 +316,7 @@ public void IOExceptionRaisedCreatingSarifLog()
                         TargetFileSpecifiers = new string[] { GetThisTestAssemblyFilePath() },
                         OutputFilePath = path,
                         Verbose = true,
+                        Force = true
```
lgolding: >Force [](start = 24, length = 5)

Now that `AnalyzeCommandBase` respects the `force` option, several tests that relied on their ability to blithely overwrite the output file started failing.

lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif.Multitool/PageCommand.cs
```diff
@@ -42,11 +43,7 @@ public void RunWithoutCatch(PageOptions options)
             if (options.Count < 0) { throw new ArgumentOutOfRangeException("count"); }
             if (!_fileSystem.FileExists(options.InputFilePath)) { throw new FileNotFoundException($"Input file \"{options.InputFilePath}\" not found."); }
 
-            if (options.Force == false && _fileSystem.FileExists(options.OutputFilePath))
-            {
-                Console.WriteLine($"Output file \"{options.OutputFilePath}\" already exists. Stopping.");
-                return;
-            }
+            DriverUtilities.VerifyOutputFileCanBeCreated(options.OutputFilePath, options.Force, _fileSystem);
```
rtaket: >VerifyOutputFileCanBeCreated [](start = 28, length = 28)

Since this is in a method named RunWithoutCatch and VerifyOutputFileCanBeCreated throws, does the user still get a nice error message?

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -24,7 +24,9 @@ public int Run(RewriteOptions rewriteOptions)
             try
             {
                 rewriteOptions = ValidateOptions(rewriteOptions);
-                
+
+                DriverUtilities.VerifyOutputFileCanBeCreated(rewriteOptions.OutputFilePath, rewriteOptions.Force, _fileSystem);
```
rtaket: >VerifyOutputFileCanBeCreated [](start = 32, length = 28)

Seems like this should be in ValidateOptions?

src/Sarif.Multitool/AbsoluteUriCommand.cs
```diff
@@ -36,6 +36,12 @@ public int Run(AbsoluteUriOptions absoluteUriOptions)
 
                     // Write output to file.
                     string outputName = sarifLog.GetOutputFileName(absoluteUriOptions);
+
+                    // This isn't optimal. If there are several input files, and the first few output files
+                    // don't exist but the next one does, we will successfully process the first few files
+                    // and then fail. It's better than overwriting the existing file, though.
+                    DriverUtilities.VerifyOutputFileCanBeCreated(outputName, absoluteUriOptions.Force, _fileSystem);
```
rtaket: >DriverUtilities [](start = 20, length = 15)

Can this be done in GetSarifFiles() so we fail earlier?

src/Sarif.Driver/DriverUtilities.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// otherwise false.
+        /// </param>
+        /// <param name="fileSystem">
+        /// An object that provides access to the file system.
+        /// </param>
+        public static void VerifyOutputFileCanBeCreated(string outputFilePath, bool force, IFileSystem fileSystem)
+        {
+            if (fileSystem.FileExists(outputFilePath) && !force)
+            {
+                throw new ArgumentException(
```
rtaket: >ArgumentException [](start = 26, length = 17)

ArgumentException is a bit unexpected. I take it this is the exception we typically throw as part of validation?

src/Sarif.Multitool/AbsoluteUriCommand.cs
lgolding: Yes, but it will take a bit of code rearrangement. `GetSarifFiles` creates a set of `AbsoluteUriFile` objects, each of which contains one of the input file paths. As the code stands, it's not until `GetOutputFileName` tells you the name of one of the corresponding output files that you know what you're looking for.

But there's no reason that an `AbsoluteUriFile` object can't be contain a property `OutputFileName`, and `GetSarifFiles` could certainly call `GetOutputFileName` to populate it for each `AbsoluteUriFile` object it returns.

TL;DR: Good idea, I'll fix it.

---
In reply to: [311713986](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r311713986) [](ancestors = 311713986)

src/Sarif.Multitool/RewriteCommand.cs
lgolding: I'll try to put it there. The problem is that `ValidateOptions` returns a `bool` rather than throwing. I'll try to make `VerifyOutputFileCanBeCreated` fit into that pattern.

---
In reply to: [311712516](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r311712516) [](ancestors = 311712516)

src/Sarif.Driver/DriverUtilities.cs
lgolding: No, apparently `ValidateOptions` returns a `bool` rather than throwing. As I said in another comment, I'll try to fit `VerifyOutputFileCanBeCreated` into that pattern.

---
In reply to: [311714256](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r311714256) [](ancestors = 311714256)

lgolding:
@rtaket, as usual you made me look alot more closely at the code. I already knew that alone among the commands, the `validate` command derives from `AnalyzeCommandBase`. This is the base class for analysis tools that use the Sarif.Driver framework. You might argue that either all the Multitool commands should use that framework or none of them should -- but as the code stands, my logic for reporting "output file already exists" must work both for `AnalyzeCommandBase`-derived commands and for all the other commands. And 

Now, `AnalyzeCommandBase` has an elaborate error reporting mechanism that produces output like this:

```
Analyzing...
error ERR997.OutputFileAlreadyExists : The output file 'c:\dev\sarif\sarif-sdk\src\Test.UnitTests.Sarif\TestData\Baseline\SuppressionTestCurrent-validate.sarif' already exists. Use --force to overwrite.

Analysis did not complete due to one or more unrecoverable execution conditions.
Unexpected fatal runtime condition(s) observed: OutputFileAlreadyExists
```

So much of the churn you will see in the next iteration is needed to fit into this framework.

I'm not going to push the change tonight, but if you see this message before tomorrow, you'll be prepared for the next set of changes you'll see.

src/Sarif.Multitool/RewriteCommand.cs
lgolding: I did this for some commands. For others it was a bit awkward and IMO not worth the time in this PR.

---
In reply to: [311806351](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r311806351) [](ancestors = 311806351,311712516)

src/Sarif.Multitool/PageCommand.cs
lgolding: Yes and no. Yes, because the call to `RunWithoutCatch` is within a `try`/`catch` block, which prints the exception to the console. But no, because the `try/catch` block does print the whole exception, stack trace and all.

I've rearranged this and some other commands to give just the nice friendly error message when the error is due to user error.

---
In reply to: [311711199](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r311711199) [](ancestors = 311711199)

src/Sarif/Errors.cs
```diff
@@ -528,7 +548,7 @@ private static string GetMessageFormatResourceForNotification(string notificatio
             string resourceName = notificationId.Replace('.', '_');
 
             return (string)typeof(SdkResources)
-                            .GetProperty(resourceName, BindingFlags.NonPublic | BindingFlags.Static)
+                            .GetProperty(resourceName, BindingFlags.Public | BindingFlags.Static)
```
lgolding: >Public [](start = 68, length = 6)

`public` because `DriverUtilities` (in the separate Sarif.Driver assembly) needs to see one of the resource strings.

src/Sarif.Converters/ToolFormatConverter.cs
```diff
@@ -44,16 +44,6 @@ public ToolFormatConverter(AssemblyLoadFileDelegate assemblyLoadFileDelegate = n
             if (inputFileName == null) { throw new ArgumentNullException(nameof(inputFileName)); }
             if (outputFileName == null) { throw new ArgumentNullException(nameof(outputFileName)); }
 
-            if (Directory.Exists(outputFileName))
```
lgolding: >Directory [](start = 16, length = 9)

This code moved to `ConvertCommand.Run`.

src/Sarif/Sarif.csproj
```diff
@@ -59,7 +59,7 @@
 
   <ItemGroup>
     <EmbeddedResource Update="SdkResources.resx">
-      <Generator>ResXFileCodeGenerator</Generator>
+      <Generator>PublicResXFileCodeGenerator</Generator>
```
rtaket: >PublicResXFileCodeGenerator [](start = 17, length = 27)

What's a PublicResXFileCodeGenerator? Is that a custom tool of ours?

lgolding: It is not a custom tool of ours. This is part of the MSBuild machinery.

When you double-click a resx file in VS, you'll see a combo box at the top that says "Access Modifier: Internal". You change "Internal" to "Public" if you want MSBuild to generate the C# class as `public` instead of `internal`. That causes the change in the project file that you see here.

Whether MSBuild has an actual tool PublicResXFileCodeGenerator, I don't know.

---
In reply to: [312230107](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312230107) [](ancestors = 312230107)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -226,6 +225,19 @@ private static bool ValidateInvocationPropertiesToLog(TContext context, IEnumera
             return succeeded;
         }
 
+        private bool ValidateOutputFileCanBeCreated(TContext context, string outputFilePath, bool force)
```
rtaket: >force [](start = 98, length = 5)

Kind of an unexpected parameter name. Perhaps forceRequested?

lgolding: I feel you, but the command line option is `--force`, and the property name in the options class(es) is `Force`, so I'd rather leave this one.

---
In reply to: [312250368](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312250368) [](ancestors = 312250368)

src/Sarif.Multitool/AbsoluteUriCommand.cs
```diff
@@ -23,24 +23,34 @@ public int Run(AbsoluteUriOptions absoluteUriOptions)
...
-                    Directory.CreateDirectory(absoluteUriOptions.OutputFolderPath);
+                    outputFilesCanBeCreated &=
+                        DriverUtilities.ReportWhetherOutputFileCanBeCreated(
+                            absoluteUriFile.OutputFilePath,
+                            absoluteUriOptions.Force,
+                            _fileSystem);
                 }
 
-                foreach (var sarifLog in sarifFiles)
+                if (!outputFilesCanBeCreated) { return 1; }
```
rtaket: >1 [](start = 55, length = 1)

Do we use the return codes for anything. i.e. should we return a different error code than the unhandled exception path?

src/Sarif.Multitool/ConvertCommand.cs
```diff
@@ -3,17 +3,41 @@
...
+            if (fileSystem == null) { fileSystem = new FileSystem(); }
+
             try
             {
+                if (string.IsNullOrEmpty(convertOptions.OutputFilePath))
+                {
+                    convertOptions.OutputFilePath = convertOptions.InputFilePath + ".sarif";
+                }
+
+                if (fileSystem.DirectoryExists(convertOptions.OutputFilePath))
```
rtaket: Should this check be in ReportWhetherOutputFileCanBeCreated? Or is it too distinct from the file existence check?

src/Sarif.Multitool/AbsoluteUriCommand.cs
lgolding: We only check 0/non-zero. No need at this time for distinguished non-zero codes.

---
In reply to: [312257820](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312257820) [](ancestors = 312257820)

src/Sarif.Multitool/ConvertCommand.cs
lgolding: *Sigh* Darn, you're good. I think it does belong there. And note that as the code stands, `convert` is the only command that performs this check, yet it's equally applicable to all commands. I found it awkward to group the two checks together -- let me go back now and see if I can manage it.

---
In reply to: [312260058](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312260058) [](ancestors = 312260058)

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -140,6 +135,8 @@ private bool ValidateOptions(FileWorkItemsOptions options)
 
             valid = ValidateOutputFileOptions(options) && valid;
 
+            valid = DriverUtilities.ReportWhetherOutputFileCanBeCreated(options.OutputFilePath, options.Force, fileSystem) && valid;
+
```
rtaket: I don't quite understand the option combinations. If OutputFilePath == null and options.Inline == true and options.Force == default/false, which seems valid, doesn't ReportWhetherOutputFileCanBeCreated fail? That is, it seems like the output/input file swap on line 35 should happen before the ValidateOptions() call.

src/Sarif.Multitool/MergeCommand.cs
```diff
@@ -24,6 +24,14 @@ public  int Run(MergeOptions mergeOptions)
         {
             try
             {
+                string outputDirectory = mergeOptions.OutputFolderPath ?? Environment.CurrentDirectory;
+                string outputFilePath = Path.Combine(outputDirectory, GetOutputFileName(mergeOptions));
+
+                if (!DriverUtilities.ReportWhetherOutputFileCanBeCreated(outputFilePath, mergeOptions.Force, _fileSystem))
+                {
+                    return -1;
```
rtaket: >-1 [](start = 27, length = 2)

It would be nice if we had a set of common return codes for Run().

lgolding: Sorry that should have been `return 1;` Fixed.

---
In reply to: [312280114](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312280114) [](ancestors = 312280114)

src/Sarif.Multitool/QueryCommand.cs
```diff
@@ -21,11 +22,13 @@ public QueryCommand(IFileSystem fileSystem = null)
 
         public int Run(QueryOptions options)
         {
+            bool valid = DriverUtilities.ReportWhetherOutputFileCanBeCreated(options.OutputFilePath, options.Force, _fileSystem);
```
rtaket: >ReportWhetherOutputFileCanBeCreated [](start = 41, length = 35)

Perhaps move this into RunWithoutCatch so it's consistent with PageCommand.

src/Sarif.Multitool/FileWorkItemsCommand.cs
lgolding: Good catch. Yes, in the other commands, I was careful to do all of the "pre-validation" option manipulation before calling `ValidateOptions`. I missed this one.

---
In reply to: [312279620](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312279620) [](ancestors = 312279620)

src/Sarif.Multitool/QueryCommand.cs
lgolding: Good, done.

---
In reply to: [312280805](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312280805) [](ancestors = 312280805)

src/Sarif.Multitool/RebaseUriCommand.cs
```diff
@@ -35,24 +35,34 @@ public int Run(RebaseUriOptions rebaseOptions)
...
                 }
 
-                var sarifFiles = GetSarifFiles(rebaseOptions);
+                var rebaseUriFiles = GetRebaseUriFiles(rebaseOptions);
+
+                bool outputFilesCanBeCreated = true;
+                foreach (RebaseUriFile rebaseUriFile in rebaseUriFiles)
+                {
+                    outputFilesCanBeCreated &=
+                        DriverUtilities.ReportWhetherOutputFileCanBeCreated(
```
rtaket: >ReportWhetherOutputFileCanBeCreated [](start = 40, length = 35)

I'm a little surprised we don't create an overload of ReportWhetherOutputFileCanBeCreated which accepts an IEnumerable<outputfilepath>. Is this a special case? i.e. it's unlikely we'll require the overload anywhere else.

lgolding: This construct (loop over output files) occurs both here (`rebaseuri` command) and in the `absoluteuri` command, and in general it will occur in any command whose options class derives from `MultipleFilesOptionsBase`. So yes, it is a good opportunity to DRY this out.

---
In reply to: [312282499](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312282499) [](ancestors = 312282499)

rtaket:
>                 try

Would the ReportWhetherOutputFileCanBeCreated check be better somewhere up here? It'll fail faster.

---
Refers to: src/Sarif.Multitool/ResultMatchSetCommand.cs:50 in dafd8ff. [](commit_id = dafd8ffe79b94edac4098ab13db6826a99107936, deletion_comment = False)

src/Sarif.Multitool/ResultMatchSetCommand.cs
```diff
@@ -64,7 +65,11 @@ public int Run(ResultMatchSetOptions options)
                         if (mergedLog.Runs[0].Results.Any(r => r.BaselineState != BaselineState.Unchanged))
                         {
                             string outputFilePath = Path.Combine(options.OutputFolderPath, fileName);
-                            WriteSarifFile(_fileSystem, mergedLog, outputFilePath, formatting);
+
+                            if (DriverUtilities.ReportWhetherOutputFileCanBeCreated(outputFilePath, options.Force, _fileSystem))
```
rtaket: >ReportWhetherOutputFileCanBeCreated [](start = 48, length = 35)

Is it OK that the function 'succeeds' even when it can't write one or more files?

src/Sarif.Multitool/RewriteCommand.cs
```diff
@@ -49,14 +50,18 @@ public int Run(RewriteOptions rewriteOptions)
...
         }
 
-        private static RewriteOptions ValidateOptions(RewriteOptions rewriteOptions)
+        private bool ValidateOptions(RewriteOptions rewriteOptions)
         {
+            bool valid = true;
+
             if (rewriteOptions.Inline)
             {
                 rewriteOptions.Force = true;
```
rtaket: >rewriteOptions [](start = 16, length = 14)

Do you also have to set OutputFilePath = InputFilePath when Inline == true? ... or does Inline mean something different here?

src/Sarif.Multitool/ResultMatchSetCommand.cs
lgolding: Not really, no :-) Fixed.

---
In reply to: [312283939](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312283939) [](ancestors = 312283939)

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -117,6 +113,21 @@ public int Run(TransformOptions transformOptions)
...
+        {
+            bool valid = true;
+
+            if (transformOptions.SarifOutputVersion != SarifVersion.OneZeroZero && transformOptions.SarifOutputVersion != SarifVersion.Current)
+            {
+                Console.WriteLine(MultitoolResources.ErrorInvalidTransformTargetVersion);
+                valid = false;
+            }
+
+            valid &= DriverUtilities.ReportWhetherOutputFileCanBeCreated(transformOptions.OutputFilePath, transformOptions.Force, _fileSystem);
```
rtaket: >ReportWhetherOutputFileCanBeCreated [](start = 37, length = 35)

Shouldn't this be checking the value of GetTransformedOutputFileName (on line 34) rather than OutputFilePath?

src/Sarif.Multitool/RewriteCommand.cs
lgolding: No, it means the same thing, and yes, it should be set. Existing bug, of course, now fixed.

---
In reply to: [312284417](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312284417) [](ancestors = 312284417)

lgolding:
>                 try

Yes, but my energy for this kind of optimization is waning. I'm going to leave this.

---
In reply to: [519724080](https://github.com/microsoft/sarif-sdk/pull/1625#issuecomment-519724080) [](ancestors = 519724080)

---
Refers to: src/Sarif.Multitool/ResultMatchSetCommand.cs:50 in dafd8ff. [](commit_id = dafd8ffe79b94edac4098ab13db6826a99107936, deletion_comment = False)

src/Sarif.Multitool/TransformCommand.cs
lgolding: Yes, fixed.

---
In reply to: [312284809](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312284809) [](ancestors = 312284809)

src/Sarif.Multitool/RebaseUriCommand.cs
lgolding: Done.

---
In reply to: [312283038](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312283038) [](ancestors = 312283038,312282499)

src/Sarif.Multitool/ConvertCommand.cs
lgolding: No, I take it back. Ideally you're right but I've reached my limit on overall Multitool command cleanup when the bug being fixed is "Please respect the `--force` option."

I filed [https://github.com/microsoft/sarif-sdk/issues/1631](https://github.com/microsoft/sarif-sdk/issues/1631), "All multitool commands should do the 'directory exists' check like 'convert' does.".

---
In reply to: [312265617](https://github.com/microsoft/sarif-sdk/pull/1625#discussion_r312265617) [](ancestors = 312265617,312260058)

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -140,6 +129,14 @@ private bool ValidateOptions(FileWorkItemsOptions options)
 
             valid = ValidateOutputFileOptions(options) && valid;
 
+            if (options.Inline)
```
lgolding: >options [](start = 16, length = 7)

This has to be done _after_ `ValidateOutputFileOptions`, because that method insists that exactly one of `options.OutputFilePath` and `options.Inline` be set. This is another example of non-uniformity among the command implementations: this is the only command that requires that.

PR #1633, opened 2019-08-09
Fix #1632: UriConverter crashes with NullReferenceException on empty URI

### LINKED ISSUE #1632: UriConverter crashes with NullReferenceException on empty URI
Example: `"helpUri": ""`

Strictly speaking this is invalid SARIF (the value is not a URI) but a prominent SARIF producer creates this output.

Instead of crashing, just don't populate the property (return `null`).

This is the root cause of Microsoft/sarif-visualstudio-extension#85.

lgolding:
The empty string is a valid relative URI although not a valid absolute URI (per RFC 3986).

This bug causes a crash in the Visual Studio Viewer VSIX (Microsoft/sarif-visualstudio-extension#85), to please review promptly so I can publish a NuGet and update the VSIX with it.


lgolding:

lgolding:

lgolding:

PR #1639, opened 2019-08-12
Fix #1594: FxCopConverter produces invalid SARIF.

### LINKED ISSUE #1594: FxCop converter produces invalid SARIF
The XML project converter is producing empty messages.

To repro, convert checked in file \src\Test.FunctionalTests.Sarif\v2\ConverterTestData\FxCop\SarifSdkTest.xml

Or run UpdateBaselines.ps1 to regen the relevant SARIF file. Test works today because checked in pre-release v2 transforms properly to valid SARIF v2

lgolding:
FxCop XML `Message` elements whose `Status` attribute is `"Active"` contain an `Issue` sub-element whose inner text provides the SARIF `result.message.text`. But `Message` elements whose `Status` attribute is `ExcludedInSource` (and possibly other values; I don't know) do not contain an `Issue` sub-element. For those FxCop `Message` elements, the `FxCopConverter` produces a SARIF `result` object whose `message` property is an empty object `{}`. This is invalid SARIF because a `message` object requires a `text` property.

Each FxCop `Message` element contains a `CheckId` attribute whose value points to a `Rule` element, which in turn contains a `Resolution` element from which `result.message.text` might seemingly be constructed. But in general, the FxCop XML file does not contain enough information to reconstruct the message. For example, this `Message`:

```
    <Member Name="#RemoveIf`1(!!0[]&amp;,System.Func`2&lt;!!0,System.Boolean&gt;)"
            Kind="Method" Static="True" Accessibility="Public" ExternallyVisible="True">
      <Messages>
        <Message Id="0#" TypeName="DoNotPassTypesByReference" Category="Microsoft.Design"
                         CheckId="CA1045" Status="ExcludedInSource"
                         Created="2018-08-23 13:07:11Z" FixCategory="Breaking"
                         LastSeen="0001-01-01 00:00:00Z" />
        ...
      </Messages>
    </Member>
```

points to this `Rule`:

```
    <Rule TypeName="DoNotPassTypesByReference" Category="Microsoft.Design" CheckId="CA1045">
      <Name>Do not pass types by reference</Name>
      <Description>...</Description>
      <Resolution Name="Default">Consider a design that does not require that {0} be a reference parameter.</Resolution>
      ...
    </Rule>
```

The sequence `{0}` should be replaced by the name of the first parameter to `RemoveIf`, which happens to be `array`. But "`array`" does not appear in the XML file. And even if all the necessary information were present, there would be no way for the converter to know which piece of information from the XML file would go into which replacement sequence in the `Resolution` string. Finally, some `Rule`s have more than one `Resolution` string, and the converter has no way to know which one to use.

That's all a long way of saying that there's no way to synthesize a valid `result.message.text` string for FxCop `Message`s that don't provide one. And the converter can't just omit `result.message`, because it's required. The best we can do is to provide a canned message: `"FxCop does not provide messages for suppressed results."`

Also:

- Upgrade SarifSdkTest.xml.sarif (which had been generated by a pre-release version of the SDK) to the final version of the SARIF 2.1.0 format. This is necessary because if this file is down-level, the `FxCopConverter`'s call to the `PrereleaseCompatibilityTransformer` produces valid SARIF, masking the bug in the converter.


lgolding:
>   "version": "2.1.0",

This upgrade essentially completely rewrites the file; don't bother trying to diff it. This file was produced by the command line
```
convert --tool FxCop C:\dev\sarif\sarif-sdk\src\Test.FunctionalTests.Sarif\v2\ConverterTestData\FxCop\SarifSdkTest.xml --output C:\dev\sarif\sarif-sdk\src\Test.FunctionalTests.Sarif\v2\ConverterTestData\FxCop\\SarifSdkTest.xml.sarif --pretty-print --force
```

---
Refers to: src/Test.FunctionalTests.Sarif/v2/ConverterTestData/FxCop/SarifSdkTest.xml.sarif:3 in b026c6e. [](commit_id = b026c6e13a92966e5bb3b46b554e6b68d9b2c247, deletion_comment = False)

lgolding:

lgolding:

lgolding:

rtaket:
Does FxCop itself display the wrong/incomplete string in the UI? It seems like we might just borrow the logic from FxCop.
Also, is the static message preferable to something closer to the actual message? Perhaps the rule name? 
That being said, I don't know enough about FxCop or SARIF to know what the best fix is here...

lgolding:
I don't know what FxCop displays. You may recall I sent mail yesterday asking if anyone had the FxCop GUI executable lying around. If I had to guess, I'd guess it displays _nothing_ for a suppressed result.

I don't think it's necessary to display anything beyond the canned message, since we do have `result.ruleId`.

But if you do have FxCop GUI, please send it along and I'll see if it suggests a way for me to do better.

---
In reply to: [520499400](https://github.com/microsoft/sarif-sdk/pull/1639#issuecomment-520499400) [](ancestors = 520499400)

lgolding:
Come to think of it, the output of FxCopCmd.exe, which is still part of Visual Studio (under Team Tools\Static Analysis Tools\FxCop) might provide enough information. I'll try that.

---
In reply to: [520500559](https://github.com/microsoft/sarif-sdk/pull/1639#issuecomment-520500559) [](ancestors = 520500559,520499400)

lgolding:
```
FxCopCmd.exe /f:c:\dev\sarif\sarif-sdk\bld\bin\AnyCPU_Debug\Sarif\net461\Sarif.dll /out:"C:\Users\Larry Golding\Desktop\SarifSdk.xml"
```
doesn't include the suppressed results at all.

---
In reply to: [520502675](https://github.com/microsoft/sarif-sdk/pull/1639#issuecomment-520502675) [](ancestors = 520502675,520500559,520499400)

lgolding:
I'll take that approval! Thanks.

---
In reply to: [520506072](https://github.com/microsoft/sarif-sdk/pull/1639#issuecomment-520506072) [](ancestors = 520506072,520502675,520500559,520499400)

PR #1641, opened 2019-08-13
Fix #1593: FortifyFprConverter produces invalid SARIF

### LINKED ISSUE #1593: FortifyFpr converter produces invalid SARIF
Convert Test.FunctionalTests.Sarif\v2\ConverterTestData\FortifyFpr\FortifyTest.fpr
Validate result

rule ids are missing, GUIDs don't validate, thread flow locations array is non-unique.

You can also run the UpdateBaselines.ps1 script to regen the current baseline to see the problem

lgolding:
There were several bugs in the converter that led to invalid SARIF:

- The trailing 12-digit component of the GUID identifying the `toolComponent` for the CWE taxonomy had only 11 digits.

    This was due to an invalid GUID in `FortifyFprConverter.CweToolComponent.Guid`. Presumably this was a copy/paste error from when the author generated the GUID. I generated a new, valid GUID.

    Also, the invalid GUID had been copy/pasted to another location in the source code, rather than referring to it by its name `FortifyFprConverter.CweToolComponent.Guid`.

- Some `threadFlowLocation.kinds` arrays contained duplicate (non-unique) values.

    This was due to an incorrect initialization of one of the elements of the `ActionTypeToLocationKindMap` (which I also renamed to `ActionTypeToLocationKinds`).

- The array `run.threadFlowLocations` contained duplicate (non-unique) elements.

    This was due to the assumption that the `ThreadFlowLocation` objects we constructed from the Fortify `UnifiedNodePool` were all distinct. Although it is true that all the `Node` sub-elements of the `UnifiedNodePool` element are distinct, we do not use all of the properties of the `Node` element in constructing a `ThreadFlowLocation`. As a result, some of the `ThreadFlowLocation`s were identical.

- `rule.id` was missing. Fortify doesn't have anything other than the GUID to serve as `rule.id` -- which, per the spec, needs to be a "stable, opaque" identifier. So I assigned the GUID to both the `id` and `guid` properties.

Also:

- Upgrade FortifyTest.fpr.sarif (which had been generated by a pre-release version of the SDK) to the final version of the SARIF 2.1.0 format. This is necessary because if this file is down-level, the `FortifyFprConverter`'s call to the `PrereleaseCompatibilityTransformer` produces valid SARIF, masking the bugs in the converter.
- We do a little code cleanup (removing unnecessary parentheses in object initializers, and capitalizing "SARIF" in comments).


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif.Converters/FortifyFprConverter.cs
```diff
@@ -38,7 +38,7 @@ internal class FortifyFprConverter : ToolFileConverterBase
         private readonly ToolComponent CweToolComponent = new ToolComponent
         {
             Name = "CWE",
-            Guid = "2B841697-D0DE-45DD-9F19-1EEE1312429",
+            Guid = "25F72D7E-8A92-459D-AD67-64853F788765",
```
harleenkohli: >25F72D7E-8A92-459D-AD67-64853F788765 [](start = 20, length = 36)

What is the source of this guid and how did we know that it changed? (perhaps we should add a comment in the code well)

harleenkohli: if yes, why?

---
In reply to: [313676882](https://github.com/microsoft/sarif-sdk/pull/1641#discussion_r313676882) [](ancestors = 313676882)

harleenkohli: Oh, looked at the sarif files to understand - it appears this is probably because we introduce it as a key to refer to in places like supportedTaxonomies --> taxa. BUT, doesnt index serve the same purpose?

---
In reply to: [313677253](https://github.com/microsoft/sarif-sdk/pull/1641#discussion_r313677253) [](ancestors = 313677253,313676882)

harleenkohli: ok, talking to myself here, but according to spec - if only index is present in a toolComponentReference, it is assumed to be referencing tool.extensions only. So, we do need a guid.

leaving this whole thing up to validate my understanding :)

---
In reply to: [313677918](https://github.com/microsoft/sarif-sdk/pull/1641#discussion_r313677918) [](ancestors = 313677918,313677253,313676882)

PR #1644, opened 2019-08-14
Fix #1485: SARIF1015 mishandles run.originalUriBaseIds (@kupsch)

### LINKED ISSUE #1485: SARIF1015 mishandles run.originalUriBaseIds (@kupsch)
Original text:
The SARIF validator (https://sarifweb.azurewebsites.net/Validation) gives a "SARIF1015: Certain URIs are required to be absolute" on PACKAGEROOT from the SARIF fragment below.  I believe this should be allowed.  The Sarif.Multitool validator (which I'm guessing the web site is based on) produces the same error.

The check could simply be a uri needs to be absolute if there is not a uriBaseId.

      "originalUriBaseIds": {
         "RESULTSROOT": {
           "uri": "file:///home/kupsch/results/"
         },
         "PACKAGEROOT": {
           "uri": "pkg1/",
           "uriBaseId": "BUILDROOT"
         },
         "BUILDROOT": {
           "uri": "file:///home/kupsch/build/"
         }
       },

Also the web page requires the use of the .sarif extension and does not allow files with the .sarif.json extension.

lgolding:
SARIF1015 complained if a `uri` property in one of the `artifactLocation` objects in `run.originalUriBaseIds` was relative. But that is allowed as long as `artifactLocation.uriBaseId` is also present.

The actual requirement is:

1. If `uriBaseId` is absent, then _either_ `uri` must be absolute or it must be absent.
2. If `uriBaseId` is present, the `uri` must be relative.

But <span>#</span>2 is true for _all_ `artifactLocation` objects, not just those in `run.originalUriBaseIds`. Therefore, in this PR:

- We remove the logic dealing with `run.originalUriBaseIds` from SARIF1015.
- We introduce a new rule SARIF1018 that enforces <span>#</span>1 in `run.originalUriBaseIds`.
- We file issue #1643, "New rule: If artifactLocation.uriBaseId is present, uri must be relative" to enforce <span>#</span>2 for all `artifactLocation` objects. But that is beyond the scope of this PR.

Also:
- DRY out the construction of the names of the validation rule test files.
- Upgrade the test files for SARIF1015 (which are affected by this change) to the final SARIF version, to make it easier to reliably make the required changes in `run.originalUriBaseIds`.

NOTE: #1485 also reported that the web site's validation page did not accept files with the (valid) filename extension .sarif.json. I filed microsoft/sarif-website#104 for that, and labeled it `m156` because _this_ bug is `m156`. I'll do that one next.


lgolding:

lgolding:

lgolding:

lgolding:

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1015.UriMustBeAbsolute_Valid.sarif
```diff
@@ -1,38 +1,39 @@
...
+            "text": "Some testing occurred."
+          },
           "workItemUris": [
             "https://example.com/my-project/issues/42"
           ]
         }
       ],
-      "originalUriBaseIds": {
-        "SRCROOT": "file:///c:/Code/sarif-sdk/src"
-      },
```
lgolding: The important change is that the SARIF1015 test files no longer have `run.originalUriBaseIds`, because that is tested by the new SARIF1018.

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -192,4 +192,10 @@
   <data name="SARIF1017_InvalidIndex" xml:space="preserve">
     <value>If an object contains a property that is used as an array index, then that array must be present and must contain at least "index + 1" elements.</value>
   </data>
+  <data name="SARIF1018_Default" xml:space="preserve">
```
rtaket: >SARIF1018_Default [](start = 14, length = 17)

SARIF1018 seems like a subset of SARIF1015 or SARIF1014 to me. i.e. I'm not clear why a new rule is required. From a naming perspective it's confusing because SARIF1015 is verifying absolute URIs and even has a check for "Certain URIs are required to be absolute." and SARIF1014 is really the corollary check to SARIF1018.
But if we require a new rule (I'm not familiar with the rule definition strategy for SARIF validation,) the _Default string should probably follow the pattern of the other rules. I was expecting something like (I have no idea what the placeholders are for, I'm just copying the other text for the other rules in the resx):
{0}: This artifactLocation object does not contain a "uriBaseId" property, which means that the value of the "uri" property must be absent or an absolute URI.

lgolding: Certainly reasonable people can disagree on the rule partitioning strategy. I think there are three distinct rules:

1. URIs that must always be absolute, no matter what. (SARIF1015).
2. URIs that occur in `artifactLocation` objects where `uriBaseId` is present: there must always be relative. (New rule, not yet written)
3. `artifactLocation` objects where `uriBaseId` is absent _and that occur in run.originalUriBaseIds_: This is the special case covered by SARIF1018. In this case only, `uri` can either be an absolute URI, or it can be `null`.

I think the partitioning I have is reasonable.

---
In reply to: [314373814](https://github.com/microsoft/sarif-sdk/pull/1644#discussion_r314373814) [](ancestors = 314373814)

PR #1647, opened 2019-08-15
 Fix #1532: Multitool transform mishandles dottedQuadFileVersion

### LINKED ISSUE #1532: Multitool transform mishandles dottedQuadFileVersion
The Roslyn compiler emits SARIF v1.0 with `tool.fileVersion` = `"3.100.19.26603 (9d80dea7)"`. The multitool transform command blithely copies it to `tool.driver.dottedQuadFileVersion` -- but the new name makes clear that this is wrong.

What should the transformer do when `fileVersion` doesn't match the expected format of `dottedQuadFileVersion`?

Well, if `fileVersion` does contain a substring of the form n.n.n.n, we should probably populate `dottedQuadFileVersion` with that. But what should we do with the rest?

lgolding:
The SARIF v1 spec does not specify a format for `tool.fileVersion` (although on the Windows platform, the spec anticipated that it would be of the form `n.n.n.n` where each `n` is one or more decimal digits).

The SARIF v2 spec is more prescriptive: it defines a property `toolComponent.dottedQuadFileVersion` with exactly that format. So if the v1 file contained a `tool.fileVersion` in any other format (which we _have_ seen in the wild), the best we can do is put it in the property bag.

Also:
- Remove some unused code.


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif/Visitors/SarifVersionOneToCurrentVisitor.cs
```diff
@@ -1194,23 +1192,48 @@ internal StackFrame CreateStackFrame(StackFrameVersionOne v1StackFrame)
             return stackFrame;
         }
 
+        private const string DottedQuadFileVersionPattern = @"[0-9]+(\.[0-9]+){3}";
```
rtaket: >@"[0-9]+(\.[0-9]+){3}" [](start = 60, length = 22)

I'm guessing we don't care, but should this be restricted to 16 bit integers?

rtaket: >@"[0-9]+(\.[0-9]+){3}" [](start = 60, length = 22)

Are version numbers globalized? i.e. should commas delimiters be accepted?

lgolding: You're right, that exceeds my care factor :-)

---
In reply to: [314363933](https://github.com/microsoft/sarif-sdk/pull/1647#discussion_r314363933) [](ancestors = 314363933)

lgolding: Not globalized, no commas. Think "IP address". Dots are part of the format.

---
In reply to: [314364505](https://github.com/microsoft/sarif-sdk/pull/1647#discussion_r314364505) [](ancestors = 314364505)

lgolding: Come to think of it, this is by design. Although Windows and .NET version numbers are our canonical example of this format, other similarly formatted version numbers might not be so restricted. `dottedQuadFileVersion` by design accepts anything that looks like `n.n.n.n`.

---
In reply to: [314371625](https://github.com/microsoft/sarif-sdk/pull/1647#discussion_r314371625) [](ancestors = 314371625,314363933)

Issue #1584, opened 2019-07-24
Multitool Transform can't upgrade "http://json.schemastore.org/sarif-2.1.0-rtm.1" schema


rscrivens:
Using the latest published multitool, to transform an older sarif filethat has a schema of "http://json.schemastore.org/sarif-2.1.0-rtm.1" the multitool does nothing.

It looks like #1501 updated the schema format to: "https://schemastore.azurewebsites.net/schemas/json/sarif-".

But it didn't account for rtm.1 still being the old schema url format in PrereleaseCompatibilityTransformer.cs:
```
switch (schemaSubVersion)
{
-       case "http://json.schemastore.org/sarif-2.1.0-rtm.1":
+       case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.1.json":
        {
```

PR #1651, opened 2019-08-16
 Fix #1577: PrereleaseCompatibilityTransformer mishandles date/times in property bags.

### LINKED ISSUE #1577: PrereleaseCompatibilityTransformer mishandles date/times in property bag
Given the input file:
[repro-1577.sarif.txt](https://github.com/microsoft/sarif-sdk/files/3397878/repro-1577.sarif.txt)

the command
```
Sarif.Multitool.exe transform --pretty-print --force --output repro-1577-transformed.sarif repro-1577.sarif
```
produces the output file
[repro-1577-transformed.sarif.txt](https://github.com/microsoft/sarif-sdk/files/3397888/repro-1577-transformed.sarif.txt)

repro-1577

Note that `invocation.startTimeUtc` is correctly left alone, but a date/time value in the property bag is transformed from the quoted string `"2019-06-22T19:06:31.209Z"` to the unquoted and reformatted string `6/22/2019 7:06:31 PM`.

lgolding:
The `PrereleaseCompatibilityTransformer` was converting SARIF like this:
```
"properties": {
  "alertDiscoveryDate": "2019-06-22T19:06:31.209Z"
}
```
to this invalid SARIF:
```
"properties": {
  "alertDiscoveryDate": 6/22/2019 7:06:31 PM
}
```
Note that the date format has changed, but worse, it's not enclosed in quotes.

The fix is explained in a comment on the new method `SarifUtilities.DeserializeObject<T>`. Please read that first.

The fix then exposed a subtle bug in our `DateTimeConverter` class that caused several unit tests to fail. The fix for _that_ bug is explained in a comment in `DateTimeConverter.WriteJson`. Please read that next, before proceeding to the rest of the review.


lgolding:

lgolding:

lgolding:

lgolding:

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -22,23 +22,27 @@ public static class PrereleaseCompatibilityTransformer
...
+            {
+                Formatting = formatting,
+                DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate,
+                DateParseHandling = DateParseHandling.None
+            };
 
             if (string.IsNullOrEmpty(prereleaseSarifLog)) { return null; }
 
-            JObject sarifLog = JObject.Parse(prereleaseSarifLog);
+            JObject logObject = SarifUtilities.DeserializeObject<JObject>(prereleaseSarifLog);
```
lgolding: >logObject [](start = 20, length = 9)

I changed this name because this code was using names like `sarifLog` for three different things:

1. A string containing the JSON contents of a SARIF log file.
2. A `SarifLog` object.
3. A `JObject` or a `JToken` resulting from a Newtonsoft parse of a JSON string.

... and I found that confusing.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -22,23 +22,27 @@ public static class PrereleaseCompatibilityTransformer
         {
             bool modifiedLog = false;
             updatedLog = null;
-            var settings = new JsonSerializerSettings { Formatting = formatting, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate };
+            var settings = new JsonSerializerSettings
+            {
+                Formatting = formatting,
+                DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate,
+                DateParseHandling = DateParseHandling.None
```
lgolding: >None [](start = 54, length = 4)

This is the fix. Right here.

src/Test.UnitTests.Sarif.Multitool/MergeCommandTests.cs
```diff
@@ -24,13 +24,16 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
     // requires one or more input files, we'll have to address this. We will probably have to
     // factor the output file writing and comparison infrastructure out of FileDiffingUnitTests.
     // We might be able to push it into a lower base class.
-    public class MergeCommandTests : FileDiffingUnitTests
+    public class MergeCommandTests : FileDiffingUnitTests, IClassFixture<MergeCommandTests.MergeCommandTestsFixture>
```
lgolding: >MergeCommandTests [](start = 17, length = 17)

This change has the effect of removing this class's test output directory before the tests in this class are run, which is useful because otherwise you have files left behind from previous runs. Other Multitool command test classes already had this machinery but I neglected to include it when I created this class yesterday.

src/ReleaseHistory.md
```diff
@@ -448,6 +448,7 @@
 * BUGFIX: Validation rule `SARIF1015` incorrectly required `originalUriBaseIds` to be contain URIs. https://github.com/microsoft/sarif-sdk/issues/1485
 * BUGFIX: Multitool transform mishandled dottedQuadFileVersion. https://github.com/microsoft/sarif-sdk/issues/1532
 * BUGFIX: Restore missing FxCop converter unit test. https://github.com/microsoft/sarif-sdk/issues/1575
+* BUGFIX: Multitool transform mishandled date/time values in property bags. https://github.com/microsoft/sarif-sdk/issues/1577
```
michaelcfanning: >BUGFIX [](start = 2, length = 6)

It is clever to add bug fixes in ascending order of issue id. That's both helpful for locating a problem and also helps with merging this frequently touched file. I've taken a mental note to help maintain this as a standard, thanks for setting the pattern.

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
```diff
@@ -23,6 +23,7 @@
     <None Remove="TestData\PrereleaseCompatibilityTransformer\ExpectedOutputs\ArtifactsWithRoles.sarif" />
     <None Remove="TestData\PrereleaseCompatibilityTransformer\ExpectedOutputs\ComprehensiveToolProperties.01-24.sarif" />
     <None Remove="TestData\PrereleaseCompatibilityTransformer\ExpectedOutputs\ComprehensiveToolProperties.sarif" />
+    <None Remove="TestData\PrereleaseCompatibilityTransformer\ExpectedOutputs\DateTimeInPropertyBag.sarif" />
```
michaelcfanning: >DateTimeInPropertyBag.sarif [](start = 78, length = 27)

A file per bug fix is too granular. Can we please think about collapsing these tests into a smaller set? For example, a file that was dedicated to serializing all types. Or a file that's dedicated to some higher level class.

Generally, for file-based data-driven tests, the goal is to have the smallest set of coherent driver files. Some patterns are mutually exclusive and require multiple files.

src/Test.UnitTests.Sarif/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/DateTimeInPropertyBag.sarif
```diff
@@ -0,0 +1,31 @@
...
+        }
+      ],
+      "results": [
+        {
+          "ruleId": "TST0001",
+          "message": {
+            "text": "Message."
+          },
+          "properties": {
+            "alertDiscoveryDate": "2019-06-22T19:06:31.209Z"
```
michaelcfanning: >"2019-06-22T19:06:31.20 [](start = 34, length = 23)

This bug fix and problem is too specifics and small to warrant its own file unless that file is defined/design to eventually absorb some broader spectrum of testing. Can you put some thought into it?

src/Test.UnitTests.Sarif.Multitool/MergeCommandTests.cs
```diff
@@ -24,13 +24,16 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
...
     {
+        public class MergeCommandTestsFixture : DeletesOutputsDirectoryOnClassInitializationFixture { }
+
         protected override string TestLogResourceNameRoot => "Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool.TestData." + TypeUnderTest;
 
         public MergeCommandTests(ITestOutputHelper outputHelper) : base(outputHelper) { }
 
         [Fact]
+        [Trait(TestTraits.Bug, "1592")]
```
michaelcfanning: Why are we adding this? If you want a reference to a bug, I suggest just adding a code comment with a clickable link. Using the attribute doesn't provide this value.

michaelcfanning:
My suggestions for this review: reconsider the overly granular new file for a small fix and prefer a clickable link to the resolved bug rather than a test trait. If you push back on the latter, please start a broader team conversation as you're implicitly proposing a new standard for the code base that warrants discussion.

src/Test.UnitTests.Sarif.Multitool/MergeCommandTests.cs
lgolding: The benefit of `[Trait]` is that you can use it to select tests to run from the command line. We can have the best of both worlds (clickable link + command line access) like this:

```
[Trait(TestTraits.Bug, "[https://github.com/microsoft/sarif-sdk/1340](https://github.com/microsoft/sarif-sdk/1340)")]
```

... which is what I did for all 15 occurrences in the code base.

---
In reply to: [316738027](https://github.com/microsoft/sarif-sdk/pull/1651#discussion_r316738027) [](ancestors = 316738027)

lgolding: I'll send mail to propose this to the team.

---
In reply to: [316773161](https://github.com/microsoft/sarif-sdk/pull/1651#discussion_r316773161) [](ancestors = 316773161,316738027)

src/Test.UnitTests.Sarif/TestData/PrereleaseCompatibilityTransformer/ExpectedOutputs/DateTimeInPropertyBag.sarif
lgolding: Good idea. I combined the two single-issue test files DateTimeInPropertyBag.sarif for #1577, and Rtm1SchemaLocation.sarif for #1584, into a single file RegressionTests.sarif, and updated the test methods accordingly. See the diff between iterations 8 and 9.

---
In reply to: [316737193](https://github.com/microsoft/sarif-sdk/pull/1651#discussion_r316737193) [](ancestors = 316737193)

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
lgolding: Fixed, see my reply to your other comment on this issue.

---
In reply to: [316736735](https://github.com/microsoft/sarif-sdk/pull/1651#discussion_r316736735) [](ancestors = 316736735)

lgolding:
I addressed both issues. See my replies to your other comments.

---
In reply to: [523951444](https://github.com/microsoft/sarif-sdk/pull/1651#issuecomment-523951444) [](ancestors = 523951444)

PR #1653, opened 2019-08-18
Fix #1642: Don't require --force if --inline is specified

### LINKED ISSUE #1642: Don't require --force if --inline is specified.
Observe this output:
```
C:> Sarif.Multitool transform --inline test.sarif
The output file 'test.sarif' already exists. Use --force to overwrite.
```
Since `--inline` means "overwrite the input file", there's no reason to require force. In my recent work on #1340 I found that some commands do work in the expected way, so this is just another example of non-uniformity among the Multitool command implementations. #1631 is another example.

@rtaket @michaelcfanning FYI

lgolding:
This addresses one of a set of about 10 bugs I recently filed in the course of fixing #1340 ("Multitool --force option: advertised default is false, actual default is true"). I discovered that there were widespread inconsistencies in the behaviors and underlying implementations of the various commands related to their treatment of command line options. This was one of them.

As usual, I took the opportunity to address IDE code warnings, improve documentation, and generally clean up the code I came in contact with while investigating the bug.

The core of the fix is the introduction of the method(s) `ValidateOutputOptions` in DriverExtensionMethods.cs. The code in each Multitool command is rearranged or enhanced to invoke these methods.

Notes:

1. I implemented the options validation methods in Sarif.Driver rather than in Sarif.Multitool because that's where the options classes themselves are defined. It's not obvious that that's where they belong.

2. The validation methods use a helper method `GetOptionDescription` that takes an option property such as `SingleFileOptionsBase.Force` and creates a description for the option like `"-p, --pretty-print"`, which is the format that the CommandLine package uses. This method was originally defined in Sarif.Multitool, but since it is now needed in Sarif.Driver, I moved it there. The corresponding unit test file moved to Test.UnitTests.Sarif.Driver.

3. The CodeFlow diffs suggest that I deleted the unit test file DriverExtensionMethods.cs and created a new file DriverExtensionMethodsTests.cs. Really I just renamed the (incorrectly named) file  and then added some new tests to it. The resulting file was so different from the original that Git doesn't recognize it as the same file.

4. As I've mentioned before, our unit test namespaces are a hodgepodge. In the same assembly, you can find tests in the namespaces `Microsoft.CodeAnalysis.Sarif.Multitool`, `Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool`, and `Test.UnitTests.Sarif.Multitool`. This makes it hard to find  your tests in the Test Explorer. Since `Test.UnitTests.Sarif.Multitool` is the default namespace specified by our project files (that is, that's the namespace you get when you create a new class in VS), I've been opportunistically converting namespaces to that form as I have occasion to edit test files containing the other forms.


lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

src/Test.UnitTests.Sarif.Driver/DriverUtilitiesTests.cs
```diff
@@ -4,12 +4,12 @@
 using System;
 using CommandLine;
 using FluentAssertions;
-using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Microsoft.CodeAnalysis.Sarif.Driver;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+namespace Test.UnitTests.Sarif.Driver
```
michaelcfanning: I don't agree w/this convention. Instead, all unit tests should have the same namespace as the core product types under test. In order to make this more convenient when creating test cases, we should modify every unit test project file to specify the product team default namespace.

To be clear, the namespace of this file s/be the same namespace where the type under test, DriverUtilities, exists. The rationale for this choice is that the test code will have, appropriately, implicit access to the same visible namespace as the product type. This eliminates the need for a using directive.

lgolding: Reverted. After my two current high-pri tasks I'm going to make a single pass and adjust every test file we have, and also modify `build.props` (if I can) to automatically take care of the default namespace for test projects.

---
In reply to: [315956789](https://github.com/microsoft/sarif-sdk/pull/1653#discussion_r315956789) [](ancestors = 315956789)

src/Sarif.Driver/DriverUtilities.cs
```diff
@@ -85,5 +88,49 @@ public static bool ReportWhetherOutputFileCanBeCreated(string outputFilePath, bo
         /// </returns>
         public static bool CanCreateOutputFile(string outputFilePath, bool force, IFileSystem fileSystem)
```
michaelcfanning: >CanCreateOutputFile [](start = 27, length = 19)

This code will throw if there are illegal characters in the output file path. Do we want to handle this situation? This code also does not verify that the system can actually write to the location in question. To address these concerns, we'd need to add logic that attempts to generate a zero-length file at the specified location. This code would need to capture relevant exceptions. All of this feedback points out that these helpers are not properly named, currently. They don't actually verify that the file is creatable, they only ensure that no file exists on disk without having permission to overwrite. Open to your thoughts on filing a bug for tail here, in order to unblock this immediate value.

michaelcfanning:
>                 return 1;

I like constants for the 0 and 1 result codes. See AnalyzeCommandBase.SUCCESS. Should we have a shared constant for this? A bit of a nit comment. :)

---
Refers to: src/Sarif.Multitool/AbsoluteUriCommand.cs:50 in d9b263b. [](commit_id = d9b263b77fd6dee85b3047a8e6da13fa0f3a5a9b, deletion_comment = False)

src/Sarif.Driver/DriverUtilities.cs
lgolding: I love the idea of adding these checks. I filed #1659, "Perform additional checks in DriverUtilities.CanCreateOutputFile."

---
In reply to: [316396582](https://github.com/microsoft/sarif-sdk/pull/1653#discussion_r316396582) [](ancestors = 316396582)

lgolding: I don't want to rename the methods. The name matches the intended purpose of the method. We can certainly add code to make it do its job better.

---
In reply to: [316407280](https://github.com/microsoft/sarif-sdk/pull/1653#discussion_r316407280) [](ancestors = 316407280,316396582)

lgolding:
>                 return 1;

I added `public const int Success = 0;` and `... Failure = 1` to `CommandBase` and adjusted all the command classes to use them.

---
In reply to: [523644673](https://github.com/microsoft/sarif-sdk/pull/1653#issuecomment-523644673) [](ancestors = 523644673)

---
Refers to: src/Sarif.Multitool/AbsoluteUriCommand.cs:50 in d9b263b. [](commit_id = d9b263b77fd6dee85b3047a8e6da13fa0f3a5a9b, deletion_comment = False)

michaelcfanning:
>         [Theory]

[Theory] s/not be used in our code base. Since you touched this file, can I trouble you to open a work item to scrub the tests to eliminate this pattern? Thanks! 

---
Refers to: src/Test.UnitTests.Sarif.Driver/FileSpecifierTests.cs:54 in a168ae3. [](commit_id = a168ae3b9c3e87e578509916d05d900d42f6e493, deletion_comment = False)

src/Sarif.Multitool/TransformCommand.cs
```diff
@@ -25,10 +25,15 @@ public int Run(TransformOptions transformOptions)
...
+                // Only set --output-file if --inline isn't specified. ValidateOptions will check
+                // to make sure that exactly one of those two options is set.
+                if (!transformOptions.Inline)
+                {
+                    transformOptions.OutputFilePath = CommandUtilities.GetTransformedOutputFileName(transformOptions);
+                }
 
                 bool valid = ValidateOptions(transformOptions);
-                if (!valid) { return 1; }
+                if (!valid) { return Failure; }
```
michaelcfanning: >Failure [](start = 37, length = 7)

Our convention in this code base is to use all caps for values that are constants like this. IOW, it is helpful to know that Failure isn't some mutable property somewhere. AnalyzeCommandBase follows this convention. Please use all caps here or start the process of advocating for changing the existing standard (in which case we need to fix other occurrences). In general, we should prefer to keep this code base self-consistent. It's fine to revisit previous style decisions. FWIW, I like SUCCESS and FAILURE because they (to me) clearly indicate they are a readable placeholder for these constants.

src/Sarif.Multitool/QueryCommand.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
     internal class QueryCommand : CommandBase
     {
+        private const int TooManyResults = 2;
```
michaelcfanning: >TooManyResults [](start = 26, length = 14)

same note, s/be all caps.

src/ReleaseHistory.md
```diff
@@ -453,5 +453,7 @@
 * BUGFIX: Multitool merge command produced invalid SARIF if there were 0 input files. https://github.com/microsoft/sarif-sdk/issues/1592
 * BUGFIX: FortifyFpr converter produced invalid SARIF. https://github.com/microsoft/sarif-sdk/issues/1593
 * BUGFIX: FxCop converter produced empty `result.message` objects. https://github.com/microsoft/sarif-sdk/issues/1594
```
michaelcfanning: you introduce an unnecessary newline in this file.

src/Sarif.Multitool/TransformCommand.cs
lgolding: I actually like ALL_CAPS for constants like this, and I'm aware of the usage in `AnalyzeCommandBase`, but it violates the .NET style guidelines. I propose to change `AnalyzeCommandBase`.

I'm not going to hold the merge for this; Iet's discuss today (I'll leave myself a note to bring it up).

---
In reply to: [316743133](https://github.com/microsoft/sarif-sdk/pull/1653#discussion_r316743133) [](ancestors = 316743133)

src/Sarif.Multitool/QueryCommand.cs
lgolding: Opened discussion, per my reply to your other comment.

---
In reply to: [316743793](https://github.com/microsoft/sarif-sdk/pull/1653#discussion_r316743793) [](ancestors = 316743793)

lgolding:
>         [Theory]

[https://github.com/microsoft/sarif-sdk/issues/1660](https://github.com/microsoft/sarif-sdk/issues/1660), "Don't use [Theory]"

---
In reply to: [523953554](https://github.com/microsoft/sarif-sdk/pull/1653#issuecomment-523953554) [](ancestors = 523953554)

---
Refers to: src/Test.UnitTests.Sarif.Driver/FileSpecifierTests.cs:54 in a168ae3. [](commit_id = a168ae3b9c3e87e578509916d05d900d42f6e493, deletion_comment = False)

Issue #1630, opened 2019-08-08
Multitool page command --force defaults to true and can't be changed


lgolding:
The `page` command has `force` set to `true` by default. I don't understand the argument in favor of this (see the comment in PageOptions.cs), and there seems to be no way to provide an explicitly `false` value to a "switch" option in the command line parsing package we use.

PR #1654, opened 2019-08-19
Fix #1629: Page command shouldn't throw on options error.

### LINKED ISSUE #1629: Multitool page command should return option errors without throwing
In the `page` command, if the user specifies invalid values for the `runIndex`, `index`, or `count` options, or the the input file path, it throws, and the user has to dig the useful information out of an exception dump. Just print the message to the console and return non-zero (as we now do for the call to `DriverUtilities.ReportWhetherOutputFileCanBeCreated`).

lgolding:
There's a big comment in PageCommandTests.cs that explains the rationale for the change (the comment is there because it's also the rationale for why I arranged the tests as I did), so I won't reproduce it here. See also the description of Issue #1629.

### DIFF
--- a/src/Sarif.Multitool/CommandBase.cs
+++ b/src/Sarif.Multitool/CommandBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using Microsoft.CodeAnalysis.Sarif.Driver;
 using Newtonsoft.Json;
@@ -12,6 +13,24 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
     public abstract class CommandBase
     {
+        protected static bool ValidateNonNegativeCommandLineOption<T>(int optionValue, string optionName)
+        {
+            bool valid = true;
+
+            if (optionValue < 0)
+            {
+                string optionDescription = CommandUtilities.GetOptionDescription<T>(optionName);
+                Console.Error.WriteLine(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        MultitoolResources.OptionValueMustBeNonNegative,
+                        optionDescription));
+                valid = false;
+            }
+
+            return valid;
+        }
+
         public static T ReadSarifFile<T>(IFileSystem fileSystem, string filePath, IContractResolver contractResolver = null)
         {
             var serializer = new JsonSerializer() { ContractResolver = contractResolver };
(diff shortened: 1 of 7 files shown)


src/Sarif.Multitool/CommandBase.cs
```diff
@@ -12,6 +13,24 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
     public abstract class CommandBase
     {
+        protected static bool ValidateNonNegativeCommandLineOption<T>(int optionValue, string optionName)
+        {
+            bool valid = true;
+
+            if (optionValue < 0)
+            {
+                string optionDescription = CommandUtilities.GetOptionDescription<T>(optionName);
```
lgolding: >CommandUtilities [](start = 43, length = 16)

I have another PR out that moves this method into `DriverUtilities`, so whoever gets in second will have a small merge conflict to resolve.

src/Sarif.Multitool/PageCommand.cs
```diff
@@ -40,22 +41,39 @@ public int Run(PageOptions options)
...
+            if (!ValidateOptions(options, _fileSystem)) { return 1; }
 
             // Load the JsonMap, if previously built and up-to-date, or rebuild it
             JsonMapNode root = LoadOrRebuildMap(options);
 
             // Write the desired page from the Sarif file
             ExtractPage(options, root);
 
-            return 1;
+            return 0;
```
lgolding: >0 [](start = 19, length = 1)

This was just a bug :-)

src/Sarif.Multitool/PageOptions.cs
```diff
@@ -50,7 +50,7 @@ internal class PageOptions
             "force",
             Default = true,
             HelpText = "Force overwrite of output file if it exists.")]
-        public bool Force { get; set; } = true;
```
lgolding: This fixes #1630, "Multitool page command --force defaults to true and can't be changed."

lgolding: There is a bug #1630, "Multitool page command --force defaults to true and can't be changed". This change does not fix that bug!

Rather, this change makes it easier to write my new unit tests in PageCommandTests. Without this change, I had to explicitly set `Force = false` in every one of my test cases except for the single case where `Force` needed to be `true`.

It makes sense to remove this initialization; I don't think there's a single other property in any of the options classes that's initialized like this. It also makes sense to fix #1630 by removing the `Default = true` on line 51, but I don't want to do that in this PR because it's a command line behavior change and I want to validate it with you first.

Note that with `Default = true`, not only do I think it's the wrong default, but there's no way to specify the value "false" on the command line! There's nothing like PowerShell's `-mySwitch:false` syntax to turn it off.

lgolding:

lgolding:

lgolding:

lgolding:

lgolding:

PR #1658, opened 2019-08-21
Fix #1662: ResultCachingLogger must cache tool notifications

### LINKED ISSUE #1662: ResultCachingLogger must cache tool notifications
When a binary analysis tool analyzes the set of binaries in a drop point, a common dependency might occur several times in different locations. The `ResultCachingLogger` associates a set of results with each input file hash, and if it encounters the same file more than once, it can simply replay the results rather than repeating the analysis.

But certain analyses result in tool notifications rather than results (for example, "PDB not found"). To avoid repeating _those_ analyses, the `ResultCachingLogger` must cache the set of notifications associated with each input file hash, alongside the set of results.

michaelcfanning:


michaelcfanning:
This team adds tool configuration notifications to our existing results caching scheme. As a result, any analysis of a file that produces a tool error notification (such as 'this binary PDB cannot be found' will only execute a single time for each copy of a binary in an analysis.

This feature drove a 'clever' use of the MSBuildConverter for functional tests that do not persist results to log files: the MSBuildConverter is used to turn the console output of a test back into a SARIF log file where results can be verified. This is helpful, because we need to ensure that console output always matches log files and we also want to ensure that command-lines which are identical, except for persisting results to a file or not, generate the same results (these code paths differ in some subtle ways in the driver framework). By adding notifications to the mix, we have uncovered a limit in the current converter, there's no mechanism for expressing and round-tripping a tool notification from console output. This is an open TODO to be discussed with @LGolding.

The additional testing for this feature has revealed weakness in the unit-testing, which should be refactored in the way that LGolding has exemplified in the code base, we should be packaging test configuration and validation data as discrete test cases. Also an open TODO to be addressed when we settle the MSBuildConverter issue.

There's some need to factor these work items out due to a need to get this core change flowed into the BinSkim tool (which will be the literal next action taken). After settling the open BinSkim issue, I'll return to the open items above.

src/Sarif/ConsoleLogger.cs
```diff
@@ -320,7 +320,9 @@ private string FormatNotificationMessage(Notification notification)
                 throw new InvalidOperationException("Unknown notification level: " + notification.Level);
             }
 
-            var sb = new StringBuilder(issueType + " ");
+            var sb = new StringBuilder(_toolName + " : ");
```
michaelcfanning: >_toolName [](start = 39, length = 9)

This change is required in order for the MSBuildConverter to transform a notification into a recorded result. Currently, the converter will treat this as a tool result, not a notification. My proposal is that this specific code path, i.e., matching the tool name rather than a file name, should produce a tool configuration notification. That would be a change in the converter itself.

src/Sarif.Driver/DriverResources.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
```
michaelcfanning: >15 [](start = 116, length = 2)

Oops, need to update VS. This file will flicker again the next time the code base is touched.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -575,8 +608,13 @@ public void AnalyzeCommand_CachesResultsWhenComputingTargetHashesAndPersistingTo
             RunTests(persistLogFileToDisk: true);
         }
 
-        private static void RunTests(bool persistLogFileToDisk)
+        private static void RunTests(bool persistLogFileToDisk, bool raiseNotificationPerBinary = false, int expectedReturnCode = 0)
```
michaelcfanning: >RunTests [](start = 28, length = 8)

This test is now officially a mess. Will refactor/clean up in a way that matches other high quality patterns LGolding has produced.

lgolding:
@lgolding is added to the review. #Closed

src/Sarif/ConsoleLogger.cs
```diff
@@ -320,7 +320,9 @@ private string FormatNotificationMessage(Notification notification)
...
+
+            sb.Append(issueType + " ");
```
lgolding: >" " [](start = 34, length = 3)

Don't we want another colon?
```
BinSkim : warning : PDB not found.
```

lgolding:
>             this.currentFileHash = context.Hashes.Sha256;

It took me a while to convince myself this was safe. We set `context.Hashes` if and only if `options.ComputeFileHashes` is set, which is also the condition for when the `ResultCachingLogger` is used.

---
Refers to: src/Sarif/ResultsCachingLogger.cs:36 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

lgolding:
>     public class ResultsCachingLogger : IAnalysisLogger

While we're in here, this class desperately needs a summary comment. Perhaps:

This class ensures that if the same file is analyzed multiple times, the analysis results and tool notifications for that file are reported only once. This can happen if a binary analysis tool analyzes a binary in whose dependency graph the same dependency appears multiple times through different paths (for example A => B, C, B => D, and C => D).

---
Refers to: src/Sarif/ResultsCachingLogger.cs:9 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -601,18 +603,38 @@ private SupportedPlatform GetCurrentRunningOS()
...
+            if (options.ComputeFileHashes)
             {
-                context.Logger.AnalyzingTarget(context);
-                foreach (Tuple<ReportingDescriptor, Result> result in results)
+                _resultsCachingLogger.HashToResultsMap.TryGetValue(context.Hashes.Sha256, out List<Tuple<ReportingDescriptor, Result>> results);
+                _resultsCachingLogger.HashToNotificationsMap.TryGetValue(context.Hashes.Sha256, out List<Notification> notifications);
+
+                bool replayCachedData = (results != null || notifications != null);
+
+                if (replayCachedData)
```
lgolding: >replayCachedData [](start = 20, length = 16)

This logic ensures that if we've already seen this file, we don't repeat the analysis. We just log the existing results for this file. And `ResultCachingLogger` ensures that repeated results appear only once in the results list. So far so good. But why even call the logger if we've already seen this file? Won't that result in getting the same console output repeatedly for the file, even if we don't get duplicate results? IOW, why _do_ we "replay the cached data". Wasn't once enough?

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -601,18 +603,38 @@ private SupportedPlatform GetCurrentRunningOS()
...
                 {
-                    if (result.Item2.Locations?.Count > 0)
+                    context.Logger.AnalyzingTarget(context);
+
+                    if (results != null)
+                    {
+                        foreach (Tuple<ReportingDescriptor, Result> result in results)
```
lgolding: >result [](start = 68, length = 6)

Perhaps a different variable name, like `cachedResult` and `cachedResults` to distinguish these tupes and lists of tuples from `Result` objects.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -601,18 +603,38 @@ private SupportedPlatform GetCurrentRunningOS()
...
+                        {
+                            if (result.Item2.Locations?.Count > 0)
+                            {
+                                result.Item2.Locations[0].PhysicalLocation.ArtifactLocation.Uri = context.TargetUri;
```
lgolding: >TargetUri [](start = 106, length = 9)

Maybe this is the answer to my question? Can the same dependent binary and/or PDB occur in more than one place, if we're analyzing a set of binaries?

lgolding:
>     public class ResultsCachingLogger : IAnalysisLogger

Or is that even right? Is the actual scenario that you are analyzing a _set_ of binaries, and more than one of them has the same dependency? Like
```
BinSkim A.exe B.exe
```
A => C and B => C

---
In reply to: [524078717](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524078717) [](ancestors = 524078717)

---
Refers to: src/Sarif/ResultsCachingLogger.cs:9 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Sarif/ConsoleLogger.cs
michaelcfanning: For this scenario, the colon is added after the descriptor ID. You are pointing out that in the absence of a descriptor or associated rule, we will append the message without a colon. This bug has always existed. In current practice, we are not raising issues that do not have an id). Suggest I open a work item for this clean up (probably not urgent)?

---
In reply to: [316878595](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316878595) [](ancestors = 316878595)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -575,8 +608,13 @@ public void AnalyzeCommand_CachesResultsWhenComputingTargetHashesAndPersistingTo
...
         {
+            // TODO: we require a special case here currently because in cases where we are not persisting to a log file and we
+            // are raising an exception per binary, these notifications are converted to results by the MSBuildConverter. In 
+            // future, we should fix the converter to
```
lgolding: >to  [](start = 51, length = 3)

"... generate tool notifications rather than results for console output lines of the form '<toolName> : ...'"  ?

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -563,6 +563,39 @@ public void AnalyzeCommand_DefaultEndToEndAnalysis()
...
+        public void AnalyzeCommand_CachesNotificationsWhenComputingTargetHashesWithoutPersistingToLogFile()
+        {
+            try
+            {
+                ExceptionRaisingRule.s_exceptionCondition = ExceptionCondition.LoadingPdb;
+
+                // TODO: even though this test raises an exception per binary, we are not persisting
+                // the results to a log file. We therefore validate against the console output. 
+                // The MSBuildConverter we use for this converts all console notifications to results.
+                RunTests(persistLogFileToDisk: false, raiseNotificationPerBinary: false, expectedReturnCode: 1);
```
lgolding: >1 [](start = 109, length = 1)

`AnalyzeCommandBase.FAILURE` (and `SUCCESS` below)?

lgolding:
>                 mockFileSystem.Setup(x => x.OpenRead(Environment.CurrentDirectory + @"\" + files[i])).Returns(new MemoryStream(Encoding.UTF8.GetBytes(logFileContents)));

Existing: `Path.Combine(Environment.CurrentDirectory, files[i])`

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:727 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
michaelcfanning: Haha! Yes, of course. :)

---
In reply to: [316893217](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316893217) [](ancestors = 316893217)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -602,7 +640,7 @@ private static void RunTests(bool persistLogFileToDisk)
 
             IFileSystem mockFileSystem = CreateMockFileSystem(files, specifier, logFileContents, logFileDirectory);
```
lgolding: >mockFileSystem [](start = 24, length = 14)

Ultra-nit naming suggestion: I usually use `mockFileSystem` to refer to the actual mock object `Mock<IFileSystem>`, and just plain `fileSystem` to hold `mockFileSystem.Object`. Note how within the method `CreateMockFileSystem`, `mockFileSystem` refers to the mock object, but here it refers to the "actual" file system.

src/Sarif/ConsoleLogger.cs
lgolding: No need for a work item, but my question and your answer suggest the need for a code comment.

---
In reply to: [316892439](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316892439) [](ancestors = 316892439,316878595)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
lgolding: Yes, see my comment below. I misunderstood the scenario. We are encountering the same file as a dependency of different analysis targets (not multiple times in the course of analyzing a single target), so the URIs will be different.

---
In reply to: [316889393](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316889393) [](ancestors = 316889393)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
lgolding: Yes, that's it, I see now.

---
In reply to: [316890439](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316890439) [](ancestors = 316890439)

lgolding:
>     public class ResultsCachingLogger : IAnalysisLogger

Yes, I understand now, it's the latter. But we do need a comment to avoid people being confused as I was :-)

---
In reply to: [524082882](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524082882) [](ancestors = 524082882,524078717)

---
Refers to: src/Sarif/ResultsCachingLogger.cs:9 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
michaelcfanning: Got it, thanks. As noted, am looking to conform to your general patterns in testing of this kind, i like the conventions you've set.

---
In reply to: [316895841](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316895841) [](ancestors = 316895841)

michaelcfanning:
>                 mockFileSystem.Setup(x => x.OpenRead(Environment.CurrentDirectory + @"\" + files[i])).Returns(new MemoryStream(Encoding.UTF8.GetBytes(logFileContents)));

Nice thanks

---
In reply to: [524086869](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524086869) [](ancestors = 524086869)

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:727 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

lgolding:
At first I didn't know what test pattern of mine you were referring to; now I understand you're talking about my practice of defining a private `TestClass`, creating a list of test case objects, looping over them, and keeping track of the titles of the failing test cases. Yes, that _is_ nice.

---
In reply to: [523998946](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-523998946) [](ancestors = 523998946)

lgolding:
Now after re-reading this I also finally understand (even though I know you've said the words, I didn't fully grok it) that by caching tool notifications, you prevent the re-execution of analyses which produce notifications _rather than results_ -- until now, you had no way of knowing that the analysis had already been performed, so this really is a perf optimization.

---
In reply to: [524097501](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524097501) [](ancestors = 524097501,523998946)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -611,35 +649,44 @@ private static void RunTests(bool persistLogFileToDisk)
...
+            // error', there will still be a notification raised for each scan target (but no per-file result).
+            int baseWarningsCount = raiseNotificationPerBinary ? 0 : files.Count();
 
             // When we run without caching results by hash, we should have a result for 
             // every file that produces a failure level of some kind (error & warning only since not running verbose).
             int expectedResultsCount = GetNonVerboseFailuresCount(files) + baseWarningsCount;
 
-            IList<Result> results = RunTest(mockFileSystem, options, expectedResultsCount);
+            Run run = RunTest(mockFileSystem, options, expectedResultsCount, expectedReturnCode);
+            IList<Result> results = run.Results;
```
lgolding: >results [](start = 26, length = 7)

BUG: We aren't actually verifying that the `results` and `notification` list have the right numbers of elements. We're just extracting them and overwriting them on the next test case.

michaelcfanning:
>     public class ResultsCachingLogger : IAnalysisLogger

Agreed. You know that I don't believe in comprehensive XML doc comments except for actual public reusable API cases (instead, I mostly prefer in-lined comments). I have one exception, though, I believe that nearly every type in the code base should have a strong summary comment. Here's the one I authored for this case:

    /// <summary>
    /// This class caches analysis results for unique files (by hash). Consumers can retrieve and use these cached
    /// results in preference of repeating the analysis. A binary drop point, for example, may contain multiple 
    /// copies of a common dependency that has been copied to the output directory of every component that
    /// references it. During analysis, this logger will capture and retain results produced for a single copy
    /// of the file. A consumer can consult this cache and retrieve the results for a file copy, in preference
    /// of simply repeating the analysis. This can result in significant performance gains, when that analysis
    /// is expensive (such as in the case of a binary analysis that retrieves and crawls binary PDBs).
    /// </summary>


---
In reply to: [524089134](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524089134) [](ancestors = 524089134,524082882,524078717)

---
Refers to: src/Sarif/ResultsCachingLogger.cs:9 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
michaelcfanning: Yep, good eye, I meant to note this detail when I described these tests as a disaster. That lack of validation is an existing test bug. Will address this in the next rewrite.

---
In reply to: [316907814](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316907814) [](ancestors = 316907814)

lgolding:
>     public class ResultsCachingLogger : IAnalysisLogger

LGTM. The only thing I'd change is where you say "a single copy of the file" it should be something like "the first time the analysis encounters each distinct file".

---
In reply to: [524099328](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524099328) [](ancestors = 524099328,524089134,524082882,524078717)

---
Refers to: src/Sarif/ResultsCachingLogger.cs:9 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
michaelcfanning: Nice, yes.

---
In reply to: [316890028](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316890028) [](ancestors = 316890028)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
michaelcfanning: Indeed. Imagine a build out of MyDll, MyOtherDll, and MyExe, all of which have a dependency on Newtonsoft.Json. The build system will copy that dependency to all three locations. The scan automation points at the high level output directly and ends up analyzing all three copies. 

---
In reply to: [316897163](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316897163) [](ancestors = 316897163,316890439)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -682,15 +729,16 @@ private static IFileSystem CreateMockFileSystem(string[] files, string specifier
             return mockFileSystem.Object;
         }
 
-        private static IList<Result> RunTest(IFileSystem fileSystem, TestAnalyzeOptions options, int expectedResultsCount)
+        private static Run RunTest(IFileSystem fileSystem, TestAnalyzeOptions options, int expectedResultsCount, int expectedReturnCode = 0)
```
lgolding: >expectedResultsCount [](start = 91, length = 20)

I assume when you rework these tests there will be an additional argument `expectedNotificationsCount`.

michaelcfanning: That's right. We'll actually package this up the way that you tend to do: define a test case data object that contains configuration variables and validation variables (and a test case title). Create an array. Create a little test case driver the processes the cases. You've implemented this pattern in several places and it's very clean and readable and fairly maintainable,. It's extremely easy for adding/updating tests, for example. The downside is that the failures are very opaque, you know the title of a test case that failed with no other information. In MSTEST, I used to work around this by emitting output messages to a string builder for failures encountered along the way, then output this info if non-empty. This technique breaks down with fluent assertions. A second pattern in our code base is to simply error out (using the fluent assertions) in the middle of test case execution. With this approach, you get a much clearer sense of where the failure occurred at the expense of suspending execution of the remaining test cases. A final approach we could consider would be to author a dedicated test for each test case. There's a bit more code authoring here but it may be the right balance.

---
In reply to: [316909270](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316909270) [](ancestors = 316909270)

src/Sarif/ConsoleLogger.cs
michaelcfanning: Let me just add the colon. :) That's as easy as adding a comment or filing a bug.

---
In reply to: [316896655](https://github.com/microsoft/sarif-sdk/pull/1658#discussion_r316896655) [](ancestors = 316896655,316892439,316878595)

michaelcfanning:
>                 mockFileSystem.Setup(x => x.OpenRead(Environment.CurrentDirectory + @"\" + files[i])).Returns(new MemoryStream(Encoding.UTF8.GetBytes(logFileContents)));

Your suggestion doesn't work. There's something about this initialization path that doesn't allow for this setup. I'd be interested in investigating this problem further with you.

---
In reply to: [524096469](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524096469) [](ancestors = 524096469,524086869)

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:727 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

lgolding:
>                 mockFileSystem.Setup(x => x.OpenRead(Environment.CurrentDirectory + @"\" + files[i])).Returns(new MemoryStream(Encoding.UTF8.GetBytes(logFileContents)));

Yes, I noticed the test break in AppVeyor. When you run the test on the desktop, what is the output? (stack trace, etc.) If you have a moment please send it to me. Thanks.

---
In reply to: [524363021](https://github.com/microsoft/sarif-sdk/pull/1658#issuecomment-524363021) [](ancestors = 524363021,524096469,524086869)

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:727 in 41ab1d2. [](commit_id = 41ab1d281dec71a60bf1f05b53248a77fc043c2a, deletion_comment = False)

PR #1661, opened 2019-08-22
Added a check to see if absolute path is provided then insert the file

SurbhiMS:
Added a check to see if absolute path is provided then insert the text file
if OriginalUriBaseIds is null but fileLocation.Uri is not null that means fileLocation.Uri has absolute path, in that case create a new Artifacts.contents.binary (because the user tried to insert the file contents to the existing sarif file) with contents of file.


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -123,7 +123,7 @@ public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
         public override Artifact VisitArtifact(Artifact node)
         {
             ArtifactLocation fileLocation = node.Location;
-            if (fileLocation != null && _run.OriginalUriBaseIds != null)
+            if (fileLocation != null && (_run.OriginalUriBaseIds != null || fileLocation.Uri != null))
```
lgolding: >(_run.OriginalUriBaseIds != null || fileLocation.Uri != nul [](start = 40, length = 59)

There's no need for the new check on `fileLocation.Uri`. In fact, there's no need for the existing check on `_run.OriginalUriBaseIds`. The check can be just:
```
if (fileLocation != null)
```
That's because the method `ArtifactLocation.TryReconstructAbsoluteUri` (see the call on Line 137 below) does those checks itself. It handles all the cases: whether `originalUriBaseIds` does or doesn't exist, and whether `file.Uri` is `null`, relative or absolute. And its parameter `out Uri uri` always has the right answer.

This means two things:

1. The bug was not that we were missing a check on `fileLocation.Uri` -- it was that the _existing_ check on `originalUriBaseIds` was wrong! The fix is just to remove that check. Because if `fileLocation.Uri` is absolute, it doesn't matter whether `originalUriBaseIds` is there or not. `TryReconstructAbsoluteUri` takes care of it all.

2. The two added lines 154 and 155 below are not necessary. Again, the parameter `out Uri uri`  is always the right answer.

lgolding:
Whenever we change the product code, we need to add one or more unit tests to verify the correctness of the change. The good news is we already have a unit test class `InsertOptionalDataVisitorTests`. So we'll add a new test (and a new test file) to verify the "absolute URI but no originalUriBaseIds" case.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -150,14 +150,17 @@ public override Artifact VisitArtifact(Artifact node)
                         }
 
                         int length = node.Length;
-                        node = Artifact.Create(uri, _dataToInsert, encoding: encoding);
+
+                        if(_run.OriginalUriBaseIds != null) node = Artifact.Create(uri, _dataToInsert, encoding: encoding);
```
michaelcfanning: need a space after if. I think i would have assigned a URI to a variable here using the ternary conditional operator and then called Artifact.Create. Alternately, our coding convention states that if/else bodies must have bracing even in cases where they are expressed on a single line.

// Cleaner IMO
Uri uri = run.OriginalUriBaseIds != null ? uri : fileLocation.Uri;
Artifact.Create(fileLocation.Uri, _dataToInsert, encoding: encoding);

// Your pattern corrected for style. Note that there is little or no other use of this pattern in the code.
// We do use single line if statements but generally not when there's an else clause.
if (_run.OriginalUriBaseIds != null) { node = Artifact.Create(uri, _dataToInsert, encoding: encoding); }
else node = { Artifact.Create(fileLocation.Uri, _dataToInsert, encoding: encoding); }

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
michaelcfanning: Nice fix. Thanks for the explanation.

---
In reply to: [316919117](https://github.com/microsoft/sarif-sdk/pull/1661#discussion_r316919117) [](ancestors = 316919117)

PR #1666, opened 2019-08-29
Work item filing extensions support release defs

cfaucon:
Updating work item filing extensions to support different artifact types such as release definitions, and not be as hard coded to build definitions.

### DIFF
--- a/src/Sarif.WorkItemFiling/Extensions.cs
+++ b/src/Sarif.WorkItemFiling/Extensions.cs
@@ -69,19 +69,7 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
                 ruleName = sarifLog.Runs[0].Tool.Driver.Rules[result.RuleIndex].Name + ":" + ruleName;
             }
 
-            string titleEntity;
-            switch (artifactType)
-            {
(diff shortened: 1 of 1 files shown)


src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -27,13 +27,14 @@ public static WorkItemFilingMetadata CreateWorkItemFilingMetadata(this SarifLog
             ArtifactLocation artifactLocation = sarifLog.Runs[0].Artifacts[0].Location;
 
             string organization = artifactLocation.GetProperty("OrganizationName");
-            string buildDefinitionId = artifactLocation.GetProperty<int>("BuildDefinitionId").ToString();
+            string artifactId = artifactLocation.GetProperty<int>("ArtifactId").ToString();
             string projectName = artifactLocation.GetProperty("ProjectName");
+            string artifactType = artifactLocation.GetProperty("EntityType");
```
rtaket: >EntityType [](start = 64, length = 10)

ArtifactType

cfaucon: Fixed

PR #1669, opened 2019-09-04
Fix #1668: SarifLogger doesn't handle sub-rules.

### LINKED ISSUE #1668: SarifLogger does not handle sub-rules
Given a `Result` object `result` with `result.RuleId == "TEST0001/1"`, and a `ReportingDescriptor` object `rule` with `rule.Id == "TEST0001"`, the call
```
SarifLogger logger = ... ;
...
logger.Log(rule, result);
```
fails with the message:
```
Error: System.ArgumentException: The rule id 'TEST0001/1' specified by the result does not
match the actual id of the rule 'TEST0001'
```

lgolding:


lgolding:

lgolding:

lgolding:

src/Sarif/ExtensionMethods.cs
```diff
@@ -80,6 +80,23 @@ public static string InstanceIdLogicalComponent(this RunAutomationDetails runAut
             return instanceId.Substring(0, instanceId.LastIndexOf('/'));
         }
 
+        public static bool IsEqualToOrHierarchicalDescendantOf(this string child, string parent)
+        {
+            if (child == parent) { return true; }
+
+            string[] childComponents = child.Split(new[] { SarifConstants.HierarchicalComponentSeparator });
```
rtaket: >new [](start = 51, length = 3)

Is the new[] syntax required? I thought the parameter was a params array...

lgolding: Thanks! I really disliked how ugly that call was :-)

---
In reply to: [321820925](https://github.com/microsoft/sarif-sdk/pull/1669#discussion_r321820925) [](ancestors = 321820925)

PR #1672, opened 2019-09-12
SarifTrim and related work

ScottLouvau:
Note: SarifTrim is checked in as a Sample, but will become a Multitool mode after we plan how to fit it in.
Note: 200+ test baselines updated - all files have Run.Artifacts index removed and Run.Results before other properties. Focus on other files, please. =)
Note: Converter tests had to be re-baselined. This was hard, and I made improvements.
Sorry I didn't bring up issues in a discussion first; broader feedback welcome here.

Code Changes:
- SarifConsolidator, a class to remove redundancies from Sarif files, added.
- 'SarifTrim' sample added to allow running consolidator logic.
- SarifToCsv: Add a 'RemoveNewlines' option to make using in Excel easier.
- SarifToCsv: Add FormattedMessage property support.
- JsonMap: Add map size limit; make map closer to requested size.
- SarifLogger: Run.Artifacts shouldn't include Index.
- SarifLogger: ArtifactLocations referencing Run.Artifacts should only have Index.
- Run class reordered to make constructors and properties easier to find in file.

Test Updates:
 - Baselines updated.
 - Actual files now written in a separate folder, so:
   - Folder diff can compare *all* test failures. 
   - One folder copy (sarif-sdk\bld\bin\AnyCPU_Release\Test.FunctionalTests.Sarif\netcoreapp2.0\Actual\v2 -> sarif-sdk\src\Test.FunctionalTests.Sarif\v2) updates all baselines.
 - Removing update baselines script (no longer needed)
 - SemmleQL tests with escaped text in source for expected changed to use files.
 - Fortify and MSBuild converter tests moved to SarifConverterTests.cs.
 - Removed transformation-before-compare so that updating baselines fixes all tests.
 - Fortify file with incorrect GUIDs fixed so test checking them doesn't complain.
 - Test.FunctionalTests.Sarif.csproj cleaned so v1 and v2 outputs always all copied to build output location.
 - FileDiffingUnitTest critical to override property made abstract; helpers used when inheriting not need made public static.


src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
```
michaelcfanning: This file needs a copyright banner.

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+    ///  for handling of a huge log.
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length < 1)
+            {
+                Console.WriteLine(@"Usage: SarifTrim [inputFilePath] [outputFilePath?] [removeParts?]
+    removeParts: Comma-separated sections to remove. Options: UriBaseIds,CodeFlows,RelatedLocations,Graphs,GraphTraversals,Stacks,WebRequests,WebResponses");
```
michaelcfanning: >RelatedLocations [](start = 83, length = 16)

RelatedLocations is a tricky thing to remove as messages have a mechanism for linking to them. i.e., a naive removal of related locations will result in messages with a link that can't be resolved. Might be safer to exclude this, alternately the code could raise an exception on determining that removal might invalidate some messages.

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                return;
+            }
+
+            string inputFilePath = args[0];
+            string outputFilePath = (args.Length > 1 ? args[1] : Path.ChangeExtension(inputFilePath, $"trim{Path.GetExtension(inputFilePath)}"));
```
michaelcfanning: >trim [](start = 103, length = 4)

maybe use 'trimmed' here, so that a file such as MyLog.sarif is renamed to MyLog.trimmed.sarif?

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                consolidator.RemoveCodeFlows = (removeParts.Contains("CodeFlows"));
+                consolidator.RemoveRelatedLocations = (removeParts.Contains("RelatedLocations"));
+                consolidator.RemoveGraphs = (removeParts.Contains("Graphs"));
+                consolidator.RemoveGraphTraversals = (removeParts.Contains("GraphTraversals"));
+                consolidator.RemoveStacks = (removeParts.Contains("Stacks"));
+                consolidator.RemoveWebRequests = (removeParts.Contains("WebRequests"));
+                consolidator.RemoveWebResponses = (removeParts.Contains("WebResponses"));
+
+                // Trim Result CodeFlowLocations, long Messages, and redundant Region properties
+                using (SarifLogger logger = new SarifLogger(outputFilePath, LoggingOptions.OverwriteExistingOutputFile, tool: run.Tool, run: run))
```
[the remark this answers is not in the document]
michaelcfanning: It would be nice if the sample code used CommandlineParser and the built in options base classes, in order to help keep command line interfaces uniform and slow the necessity of authoring new, redundant logic to handle things like file overwrites. I can appreciate that there is another value in sample code, which is to keep dependencies minimal (to ease portability of the code). But if we will be using the samples as a kind of staging ground for things that move into the 'proper' SDK binaries, probably better to consume those existing patterns.

---
In reply to: [323979575](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323979575) [](ancestors = 323979575)

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    foreach (Result result in run.Results)
+                    {
+                        consolidator.Trim(result);
+                        logger.Log(result.GetRule(run), result);
+                    }
+
+                    if (removeParts.Contains("UriBaseIds"))
+                    {
+                        foreach (Artifact a in run.Artifacts)
```
michaelcfanning: >Artifacts [](start = 51, length = 9)

when trimming uri base ids, shouldn't we also trim the uri base id definitions? are we doing this elsewhere?

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                        {
+                            a.Location.UriBaseId = null;
```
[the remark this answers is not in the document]
michaelcfanning: this is true of nearly everything. :)  it's valid for a run to not have any results at all, for example (if the tool has failed to init or some other catastrophic condition, for example, and you don't handle that above. 

---
In reply to: [323957242](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323957242) [](ancestors = 323957242)

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                // Trim Result CodeFlowLocations, long Messages, and redundant Region properties
+                using (SarifLogger logger = new SarifLogger(outputFilePath, LoggingOptions.OverwriteExistingOutputFile, tool: run.Tool, run: run))
+                {
+                    foreach (Result result in run.Results)
+                    {
+                        consolidator.Trim(result);
+                        logger.Log(result.GetRule(run), result);
+                    }
+
+                    if (removeParts.Contains("UriBaseIds"))
```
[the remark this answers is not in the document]
michaelcfanning: Strict removal isn't a consolidation may be what Scott had in mind.

---
In reply to: [323980181](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323980181) [](ancestors = 323980181)

src/Samples/SarifTrim/SarifTrim.csproj
```diff
@@ -0,0 +1,63 @@
...
+    <Compile Include="ConsoleWatch.cs" />
+    <Compile Include="Program.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="App.config" />
+    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\Sarif\Sarif.csproj">
```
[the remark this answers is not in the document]
michaelcfanning: Larry and I do not agree on this point. IMO our samples should compile against the working SDK, so that we have a better sense of how/when we are introducing compatibility and other breaking changes. That is, compat concerns should be considered while the hood is up on an API. Also, Larry's suggestion would require us to remember to update the samples after releasing to NuGet. If on doing so (and only then) we were to observe a problem with how an API change interacted poorly with the sample code, we might need to go back into the code, etc.

To obtain the value Larry is looking for, it would be helpful to create a separate, samples only repository that consumes the current NuGet. There would still be questions here in terms of how to keep that code in sync w/emergent SDK, etc.


---
In reply to: [323955279](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323955279) [](ancestors = 323955279)

src/Sarif/Core/Run.cs
```diff
@@ -92,7 +92,8 @@ public Uri ExpandUrisWithUriBaseId(string key, string currentValue = null)
                         hashData: hashData,
                         encoding: null);
 
-                    fileData.Location = fileLocation;
+                    // Copy ArtifactLocation to ensure changes to Result copy don't affect new Run.Artifacts copy
```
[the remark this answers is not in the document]
michaelcfanning: Can you just use fileLocation.DeepClone() here?

---
In reply to: [323989631](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323989631) [](ancestors = 323989631)

src/Sarif/Core/SarifLog.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <param name="sarifFilePath">File Path to Sarif file to load</param>
+        /// <returns>SarifLog instance for file</returns>
+        public static SarifLog Load(string sarifFilePath)
+        {
+            JsonSerializer serializer = new JsonSerializer();
+
+            using (JsonTextReader jtr = new JsonTextReader(File.OpenText(sarifFilePath)))
+            {
+                return serializer.Deserialize<SarifLog>(jtr);
+            }
```
michaelcfanning: This could be replaced by a simple call to JsonConvert.Deserialize, yes?

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
...
+                long length = stream.Length;
+                if (length <= settings.MinimumSizeForNode) { return null; }
+                settings.AdjustForFileSize(length);
             }
 
             using (JsonPositionedTextReader reader = new JsonPositionedTextReader(streamProvider))
             {
                 if (!reader.Read()) { return null; }
-                return Build(reader, 0, out long unused);
+                return Build(reader, settings, 0, out long unused);
```
michaelcfanning: > 0 [](start = 46, length = 2)

suggest using a named argument literal use here

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static JsonMapSettings DefaultSettings => new JsonMapSettings(0.01, 10 * Megabyte);
+
+        /// <summary>
+        ///  Construct JsonMapSettings for a given target size ratio and size limit.
+        /// </summary>
+        /// <param name="mapSizeRatio">Target Map size (0.01 means map should be 1% of file size or JSON being mapped)</param>
+        /// <param name="mapMaximumSizeBytes">Map size limit, in bytes (10 * 1024 * 1024 means 10 MB), 0 for no limit</param>
+        public JsonMapSettings(double mapSizeRatio, double mapMaximumSizeBytes = 0)
+        {
+            if (mapSizeRatio <= 0 || mapSizeRatio >= 100) { throw new ArgumentOutOfRangeException("maxFileSizePercentage must be between 0 and 100."); }
```
michaelcfanning: > 0 and 100 [](start = 136, length = 10)

Don't know why I'm hanging up on this but your message doesn't make it perfectly clear that 1 - 99 is the range of legal values. i.e., someone could mistakenly believe that 0 or 100 is legal.

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+            MapDesiredSizeRatio = mapSizeRatio;
+            MapMaximumSizeBytes = mapMaximumSizeBytes;
+
+            CurrentSizeRatio = mapSizeRatio;
+            Recalculate();
+        }
+
+        internal void AdjustForFileSize(double fileSizeBytes)
```
michaelcfanning: >AdjustForFileSize [](start = 22, length = 17)

You have argument validation in the ctor but not in this helper. What's to prevent someone from, say, setting MapDesiredSizeRatio to 0 after object construction? If the answer is 'nothing', consider authoring an object state validation helper that you call both on object initialization as well as this helper.

michaelcfanning: >AdjustForFileSize [](start = 22, length = 17)

I will need to go look at who is calling this helper. It's odd that there is an internal only method here. This is a public class, so why is this particular adjustment only relevant for some other internal consumer? Will keep my eye open for the utilization.

src/Sarif/Readers/SerializedPropertyInfoConverter.cs
```diff
@@ -34,7 +34,7 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             }
 
             string serializedValue = ((SerializedPropertyInfo)value).SerializedValue;
-            writer.WriteRawValue(@"""" + serializedValue + @"""");
+            writer.WriteRawValue(serializedValue);
```
michaelcfanning: >serializedValue [](start = 33, length = 15)

it troubles me to see this change. i will talk to Larry about the history of this functionality. properties mgmt. has been a consistent source of parsing/validation errors. it would be very nice to put these behind us. i see you've added some new tests in this area, looking for Larry to validate that he's happy with the comprehensiveness of testing. may warrant a TODO, as this isn't the strict point of your change, obviously.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.Writers
+{
+    /// <summary>
+    ///  SarifConsolidator is used to reduce the size of Sarif objects and the serialized log.
```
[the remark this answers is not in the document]
michaelcfanning: But we won't mention the absence of copyright headers repeatedly (though while i'm here, this file is missing one, too). :)

---
In reply to: [323958666](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323958666) [](ancestors = 323958666)

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
michaelcfanning: I stand corrected. Scott's own class summary for the consolidator notes that removal is one of its functions.

---
In reply to: [324238765](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324238765) [](ancestors = 324238765,323980181)

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        public int TotalLocations { get; private set; }
+        public int UniqueLocations { get; private set; }
+
+        public int? MessageLengthLimitBytes { get; set; }
+        public bool RemoveCodeFlows { get; set; }
+        public bool RemoveRelatedLocations { get; set; }
+        public bool RemoveGraphs { get; set; }
+        public bool RemoveGraphTraversals { get; set; }
+        public bool RemoveStacks { get; set; }
+        public bool RemoveWebRequests { get; set; }
```
michaelcfanning: >RemoveWebRequests [](start = 20, length = 17)

Similar to Larry's point above not sure how flexibility to trim web request data v. response data makes sense (certainly we could add this granularity in future if need be). This all relates to a conversation we had the other day, it would be helpful, given the combinatorics of insert/removal/transform implied by this very complex format to attempt to shape the API around higher-level use cases. 'Zap all graph related things' or 'Zap all web analysis related things' is an example of a small step towards this. We may eventually benefit from an ever higher-level concept such as 'remove all data related to the detailed investigation of an analysis result'. That's really the heart of the problem here, i.e., do we want a log file that neatly summarizes the analysis or do we need one that can drive an end-to-end evaluation/disposition,.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            _run = run;
+            _uniqueThreadFlows = new Dictionary<ThreadFlowLocation, int>(ThreadFlowLocation.ValueComparer);
+
+            if (_run.Artifacts != null)
+            {
+                // Realize Deferred Artifact list
+                _run.Artifacts = _run.Artifacts?.ToList();
+
+                // Remove indices from Run.Artifact array (redundant)
```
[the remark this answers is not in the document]
michaelcfanning: This gets back again to that offline discussion on higher level use cases. There is a log file readability use case in which someone would like to maximize the ability to search around a SARIF log file in an editor. Note that there is a real problem here. A result can refer only to a result location by artifact index (no URI) and the artifact table can omit the index and leave only the URI. With an artifact table in the dozens or hundreds this data is effectively cannot be correlated by someone in a text editor. Arguably, it is preferable to duplicate the index data (as opposed to the result level URI) as the literal # of chars to express is fewer.

---
In reply to: [323983249](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323983249) [](ancestors = 323983249,323969754)

lgolding:
>             string mapPath = Path.ChangeExtension(options.InputFilePath, ".map.json");

While we're in here (or perhaps as a separate PR), IMO we need to be able to specify the location of the map file. The user might not have write permission on the directory containing the log file.

---
Refers to: src/Sarif.Multitool/PageCommand.cs:111 in b2907ed. [](commit_id = b2907ed1615ffe2680311c2ab29f64ed8234e270, deletion_comment = False)

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+        public double CurrentSizeRatio { get; set; }
+        public int MinimumSizeForNode { get; set; }
+
+        public double MapDesiredSizeRatio { get; set; }
+        public double MapMaximumSizeBytes { get; set; }
+
+        /// <summary>
+        ///  Default Settings: Map is 1% of size of file mapped; up to a limit of 10 MB.
+        /// </summary>
+        public static JsonMapSettings DefaultSettings => new JsonMapSettings(0.01, 10 * Megabyte);
```
lgolding: > J [](start = 21, length = 2)

`readonly`

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        ///  Construct JsonMapSettings for a given target size ratio and size limit.
+        /// </summary>
+        /// <param name="mapSizeRatio">Target Map size (0.01 means map should be 1% of file size or JSON being mapped)</param>
```
lgolding: >or [](start = 97, length = 2)

"or" => "of"?

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="mapMaximumSizeBytes">Map size limit, in bytes (10 * 1024 * 1024 means 10 MB), 0 for no limit</param>
```
lgolding: >10 * 1024 * 1024 means 10 MB [](start = 72, length = 28)

Not necessary to say that 10 * 1024 * 1024 = 10 MB

src/Sarif/Map/JsonMapSettings.cs
lgolding: +1

---
In reply to: [324246138](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324246138) [](ancestors = 324246138)

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
     ///  JsonMapBuilder constructs a JsonMap for input Json files. It is passed
     ///  a size percentage target which controls how much detail it includes.
     /// </summary>
-    public class JsonMapBuilder
+    public static class JsonMapBuilder
     {
-        private const long NodeSizeEstimateBytes = 90;         // Bytes for one Node with containing property but not including children
```
lgolding: >private [](start = 8, length = 7)

I suggest a design that splits the difference between the previous design ("the one and only setting (maxFileSizePercentage) is a property of the builder class") and your new proposed design ("the settings are in a separate class, and the new Settings class performs calculations and has variable state").

I think a settings class is fine, but IMO a settings class should contain only the values the caller asks for. If the builder needs to adjust the settings, it's the builder that should perform the calculation, and if there's a need for the caller to find out what changes the builder made, then the adjusted properties should be properties of the builder.

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
     ///  JsonMapBuilder constructs a JsonMap for input Json files. It is passed
     ///  a size percentage target which controls how much detail it includes.
     /// </summary>
-    public class JsonMapBuilder
+    public static class JsonMapBuilder
```
lgolding: >static [](start = 11, length = 6)

By making the builder class `static`, you pay the price of having to pass the settings object as arguments to various private methods. What was your rationale for making the class `static`?

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
...
         ///  Build the JsonMap for a given Json file, given a stream provider.
         ///  Returns null if the source file was too small for any map nodes to fit the size budget.
         /// </summary>
         /// <param name="streamProvider">A function which will open the stream for the desired file</param>
+        /// <param name="settings">JsonMapSettings for map; null for defaults</param>
         /// <returns>JsonMap for file or null if file too small for map</returns>
-        public JsonMapNode Build(Func<Stream> streamProvider)
+        public static JsonMapNode Build(Func<Stream> streamProvider, JsonMapSettings settings = null)
         {
+            if (settings == null) { settings = JsonMapSettings.DefaultSettings; }
```
lgolding: >if [](start = 12, length = 2)

Or:
```
settings = settings ?? JsonMapSettings.Default;
```

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
...
 
         /// <summary>
         ///  Build the JsonMap for a given Json file, given a stream provider.
         ///  Returns null if the source file was too small for any map nodes to fit the size budget.
         /// </summary>
         /// <param name="streamProvider">A function which will open the stream for the desired file</param>
+        /// <param name="settings">JsonMapSettings for map; null for defaults</param>
         /// <returns>JsonMap for file or null if file too small for map</returns>
-        public JsonMapNode Build(Func<Stream> streamProvider)
+        public static JsonMapNode Build(Func<Stream> streamProvider, JsonMapSettings settings = null)
```
lgolding: >Func<Stream> streamProvider [](start = 40, length = 27)

I've never seen this pattern before. Why does `Build` not just accept a `Stream`?

src/Sarif/Map/JsonMapBuilder.cs
```diff
@@ -15,58 +15,46 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
...
+
             using (Stream stream = streamProvider())
             {
-                if (stream.Length <= MinimumSizeForNode) { return null; }
+                long length = stream.Length;
+                if (length <= settings.MinimumSizeForNode) { return null; }
+                settings.AdjustForFileSize(length);
             }
 
             using (JsonPositionedTextReader reader = new JsonPositionedTextReader(streamProvider))
```
lgolding: >streamProvider [](start = 82, length = 14)

... and then I guess I have the same question about `JsonPositionedTextReader`; why doesn't it just take a `Stream`? ...

... oh. Is the problem that if the `JsonMapBuilder` and the `JsonPositionedTextReader` both use the same stream, that one of them (probably the reader) will dispose the stream before the other one (probably the builder) is done with it?

If that's the rationale, it needs a big old comment.

michaelcfanning:
Scott, can you explain why the tool object is now at the end of the log file? We originally designed things so that this data was close to the beginning of the log file, in the spirit of making it easy to identify the tool source for happy path single run, single tool scenarios. What was the quality/functionality driver for repositioning? Maybe some code somewhere will offer a clue, off to go look for it. :)

---
In reply to: [531047570](https://github.com/microsoft/sarif-sdk/pull/1672#issuecomment-531047570) [](ancestors = 531047570)

src/Sarif/Map/JsonMapBuilder.cs
lgolding: >out long unused [](start = 50, length = 15)

C #7 has a lovely idiom for unused ("discarded") `out` parameters:
```
return Build(reader, settings, 0, out _);
```

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -425,11 +399,16 @@ private void CaptureFile(ArtifactLocation fileLocation)
...
-                fileLocation, 
+                fileLocation,
                 addToFilesTableIfNotPresent: true,
                 _dataToInsert,
                 encoding);
+
+            // Remove redundant Uri and UriBaseId once index has been set
+            fileLocation.Uri = null;
+            fileLocation.UriBaseId = null;
         }
```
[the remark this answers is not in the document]
michaelcfanning: +1. we now have completely disconnected the artifacts table from the result. that means for any specific result in a file with dozens or hundreds of artifacts, there's no reasonable way to correlate. i've offered this feedback a bunch of times, but having an option to favor 'readability' over 'compactness' may be the right approach.

also, just want to be clear on this, i do think we are at a point where it is reasonable for 'compact' to be a default', with 'readable' something that humans must explicitly configure or post-process to achieve. that is, we're now ready to favor defaults for machine vs. human consumption. the viewer eco-system provides additional support for this. what i think i'd suggest is extending VS and/or VS Code to not only open a SARIF file but also to post-process it for readability.

---
In reply to: [323985843](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323985843) [](ancestors = 323985843)

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -497,24 +476,24 @@ public bool ShouldLog(FailureLevel level)
...
-                    break;
-                }
 
                 case FailureLevel.Error:
                 case FailureLevel.Warning:
-                {
-                    break;
-                }
+                    {
+                        break;
```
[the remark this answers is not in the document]
michaelcfanning: What's wrong with the braces? I'll offer my usual point when someone suggests a style change that departs from the remainder of the code: we should be open to style changes but should favor retrofitting these in a consistent way over introducing a new pattern that makes the code inconsistent. Everywhere else in the code we use explicit bracing for cases.

---
In reply to: [323985245](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323985245) [](ancestors = 323985245)

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+            result.Stacks = (this.RemoveStacks ? null : result.Stacks);
+            result.WebRequest = (this.RemoveWebRequests ? null : result.WebRequest);
+            result.WebResponse = (this.RemoveWebResponses ? null : result.WebResponse);
+
+            // Truncate long Messages if configured
+            if (this.MessageLengthLimitBytes > 0)
+            {
+                if (result?.Message?.Text?.Length > this.MessageLengthLimitBytes)
+                {
+                    result.Message.Text = result.Message.Text.Substring(0, this.MessageLengthLimitBytes.Value);
```
michaelcfanning: >result [](start = 20, length = 6)

We should add some suffix to the string that indicates text was truncated.  'We found a horrible problem and have gone on and on about it. Here's more detail you likely didn't [remainder of message truncated due to exceeding size limit...].

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+                }
```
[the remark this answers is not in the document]
michaelcfanning: In markdown, we can have a nicely formatted message (with emphasized text perhaps) noting that truncation occurred.

---
In reply to: [323981736](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323981736) [](ancestors = 323981736)

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+
+            // Trim and Consolidate CodeFlows and ThreadFlows
+            ConsolidateCodeFlows(result);
+        }
+
+        public void ConsolidateCodeFlows(Result result)
+        {
+            if (result.CodeFlows != null)
+            {
+                foreach (CodeFlow cFlow in result.CodeFlows)
```
michaelcfanning: >foreach [](start = 16, length = 7)

why isn't there a language feature for conditional enumeration over collections that can handle the case where the collection itself is null?

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            if (locations == null || locations.Count == 0) { return locations; }
+
+            HashSet<Location> uniqueLocations = new HashSet<Location>(Location.ValueComparer);
+            List<Location> newLocations = new List<Location>();
+
+            this.TotalLocations += locations.Count;
+            for (int i = 0; i < locations.Count; ++i)
+            {
+                Location loc = locations[i];
```
michaelcfanning: >loc [](start = 25, length = 3)

Can we use 'location' as a var name here? Generally, we tend to favor verbose (arguably over verbose) names in this code base.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+                {
+                    this.UniqueLocations++;
+                    newLocations.Add(loc);
+                }
+            }
+
+            return newLocations;
+        }
+
+        public void Trim(Location loc)
```
michaelcfanning: >loc [](start = 34, length = 3)

certainly a shortened name shouldn't be used for a public method, let's use 'location'. i suspect VS method extraction created this parameter name for you (based on your local from the originating caller.

src/Sarif/Writers/SarifLogger.cs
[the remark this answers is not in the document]
lgolding: If that's our convention, I withdraw the comment. I disagree with the convention. IMO it unnecessarily consume both vertical and horizontal white space, making it harder to see a block of code at a glance.

---
In reply to: [324275322](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324275322) [](ancestors = 324275322,323985245)

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+            {
+                LogicalLocation loc = locations[i];
+
+                // Trim Locations themselves
+                Trim(loc);
+
+                // Keep only one copy of each unique location
+                if (uniqueLocations.Add(loc))
+                {
+                    this.UniqueLocations++;
```
michaelcfanning: >this.UniqueLocations++; [](start = 20, length = 23)

you could keep incrementing this, or you could simply compute this value a single time before return (set the property to uniqueLocations.Count)

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        public void Trim(LogicalLocation loc)
+        {
+            if (loc != null)
+            {
+                // Leave only index if index is present
+                if (loc.Index != -1)
+                {
+                    loc.DecoratedName = null;
```
michaelcfanning: >loc.DecoratedName [](start = 20, length = 17)

what's our guarantee here and elsewhere that the object pointed to in the run level array contains all this information? ie, if we zap FullyQualifiedName from the result associated logical location, we should be sure that it is expressed in the run.logicalLocations instance pointed to by Index. all this points to an unpleasant inconsistency that could exist in SARIF (for which we don't have a validator), SARIF results could themselves point to a common array resident thing while being inconsistent in its local recapitulation of properties.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            if (r != null && r.StartLine > 0)
+            {
+                // Remove EndLine if the same as StartLine
+                if (r.EndLine == r.StartLine)
+                {
+                    r.EndLine = 0;
+                }
+
+                // Remove Offset/Length if Line/Column available
```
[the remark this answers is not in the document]
michaelcfanning: I actually think that in the spirit of favoring compactness, it would have been more consistent for us to zap the line/column details when offset/length are available. The 'readable' mode would zap offset/length instead.

---
In reply to: [323970078](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r323970078) [](ancestors = 323970078)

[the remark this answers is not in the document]
michaelcfanning: By the way, the start/end line/column stuff has very elaborate conventions for defaulting various values. If start and end line are identical, for example, you don't need to recapitulate the end line. If the endColumn happens to be the last char in the line, you don't need that (which is not a condition that can be computed strictly from the SARIF log file, however). You could consider reviewing these rules and updating Trim(Region) based on them (at least capturing some of the more basic ones).

---
In reply to: [324282310](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324282310) [](ancestors = 324282310,323970078)

michaelcfanning:
>         public void PropertyBagConverter_RoundTripsStringPropertyWithEscapedCharacters()

Is the name of this method to your satisfaction after making the changes you have here? 

---
Refers to: src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs:17 in b2907ed. [](commit_id = b2907ed1615ffe2680311c2ab29f64ed8234e270, deletion_comment = False)

michaelcfanning:
>         public void PropertyBagConverter_RoundTripsStringPropertyWithEscapedCharacters()

If we're putting the hood up, can we go ahead and complete the range of types involved here? For example, consider serializing a double that exceeds Int32.max. Pass a float that is not an integral value. Consider round-tripping a simple custom object. And serialize/deserialize arrays of every supported thing. You may prefer to propose that @larrygolding put the hood up and make all this right (and simply open a TODO issue). I'd like to be sure that we understand and test expected behaviors when setting property values to null as well.

---
Refers to: src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs:17 in b2907ed. [](commit_id = b2907ed1615ffe2680311c2ab29f64ed8234e270, deletion_comment = False)

src/Test.FunctionalTests.Sarif/UpdateBaselines.ps1
```diff
@@ -1,112 +0,0 @@
-﻿Param(
-    [string]$ToolName
```
michaelcfanning: >$ToolName [](start = 12, length = 9)

Great to see this file disappear. Do we have some test output that provides guidance on how to rebaseline? Ideally, this information would appear in test output. I assume that you've gone through the experience of having the converter tests fail and have verified that all the test output is complete and works? (i.e., that the generated file diff command isn't broken, as I assume it wouldn't be).

src/Samples/SarifToCsv/Program.cs
```diff
@@ -39,6 +39,8 @@ public static void Main(string[] args)
                 string sarifFilePath = args[0];
                 string csvFilePath = args[1];
                 IEnumerable<string> columnNames = (args.Length > 2 ? args[2] : ConfigurationManager.AppSettings["ColumnNames"]).Split(',').Select((value) => value.Trim());
+                bool removeNewlines = bool.Parse(ConfigurationManager.AppSettings["RemoveNewlines"] ?? "false");
```
lgolding: >?? [](start = 100, length = 2)

Bug: I just tested this and if a setting is missing, `AppSettings["MissingSetting"]` returns the empty string, not `null`.

src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
```diff
@@ -71,6 +72,9 @@ public static Action<WriteContext> GetWriter(string columnName)
             writers["Tags"] = (c) => { c.Writer.Write(String.Join("; ", ((IEnumerable<string>)c.Result.Tags) ?? Array.Empty<string>())); };
             writers["WorkItemUris"] = (c) => { c.Writer.Write(String.Join("; ", c.Result.WorkItemUris?.Select((uri) => uri.ToString()) ?? Array.Empty<string>())); };
 
+            // Formatted Message
+            writers["FormattedMessage"] = (c) => { c.Writer.Write(c.Result.GetMessageText(c.Result.GetRule(c.Run))); };
```
lgolding: >FormattedMessage [](start = 21, length = 16)

Any reason not to just call this "Message"? That's more understandable to a human user who doesn't know that "formatted message" is a "thing" in the SARIF spec.

src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
```diff
@@ -21,9 +22,17 @@ public void PropertyBagConverter_RoundTripsStringPropertyWithEscapedCharacters()
...
+            string normalStringPropertyValue = "csharp";
+
+            string dateTimePropertyName = "runStart";
+            DateTime dateTimePropertyValue = DateTime.UtcNow.AddHours(-1.43);
+
             var run = new Run();
             run.SetProperty(intPropertyName, 42);
             run.SetProperty(stringPropertyName, stringPropertyValue);
+            run.SetProperty(normalStringPropertyName, normalStringPropertyValue);
+            run.SetProperty<DateTime>(dateTimePropertyName, dateTimePropertyValue);
```
lgolding: >DateTime [](start = 28, length = 8)

The compiler will infer the generic type from the type of the second parameter, but I understand if you want to be explicit in the context of test code (to demonstrate the types you're covering).

michaelcfanning:
Also update build.props so that it properly compiles 2.1.17 packages....

---
In reply to: [531047570](https://github.com/microsoft/sarif-sdk/pull/1672#issuecomment-531047570) [](ancestors = 531047570)

src/Sarif/Writers/SarifConsolidator.cs
lgolding: Yes, the spec addresses that issue. For example, in §3.33.3 (logicalLocation.index property):

    If `index` is present, `thisObject` **SHALL** take all properties present on the cached object. If `thisObject` contains any properties other than `index`, they **SHALL** equal the corresponding properties of the cached object.

I filed [https://github.com/microsoft/sarif-sdk/issues/1674](https://github.com/microsoft/sarif-sdk/issues/1674), "New rule: Ensure that local objects are consistent with cached objects". 

---
In reply to: [324280903](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324280903) [](ancestors = 324280903)

lgolding: That will be a big job, comparable in size to the implementation of `SARIF1017`, "Invalid index".

---
In reply to: [324342614](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324342614) [](ancestors = 324342614,324280903)

src/Samples/SarifTrim/ConsoleWatch.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+
+public class ConsoleWatch : IDisposable
+{
+    private Stopwatch _watch;
```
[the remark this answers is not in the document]
ScottLouvau: Fixed.

src/Samples/SarifTrim/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+
+namespace SarifTrim
+{
+    /// <summary>
+    ///  SarifTrim is a one-off sample removing some redundant and some excessively large parts of a Sarif log
```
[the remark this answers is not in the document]
ScottLouvau: Sure - your wording seems fine.

src/Samples/SarifTrim/SarifTrim.csproj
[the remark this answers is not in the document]
ScottLouvau: After some thought, I want to keep these as Project references. I end up changing the SDK a lot when working on them, as I find bugs and feature needs, and I eventually want them to be part of the main build so that breaking changes won't affect them.

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
ScottLouvau: Yep. Updated comment.

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
ScottLouvau: Yep. I've gotten bitten by a few of these already. I'll look for others.

src/Sarif/Writers/SarifConsolidator.cs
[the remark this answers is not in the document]
ScottLouvau: Fixed comment and copyright comment.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        public void Trim(Location loc)
+        {
+            if (loc != null)
+            {
+                // Remove Ids
+                loc.Id = -1;
+
+                Trim(loc.PhysicalLocation);
+                Trim(loc.LogicalLocation);
```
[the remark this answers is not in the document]
ScottLouvau: Removed.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+            return newLocations;
+        }
+
+        public void Trim(Location loc)
+        {
+            if (loc != null)
+            {
+                // Remove Ids
+                loc.Id = -1;
+
```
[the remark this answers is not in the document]
ScottLouvau: Removed.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+                Trim(loc.PhysicalLocation);
+                Trim(loc.LogicalLocation);
+                Trim(loc.LogicalLocations);
+            }
+        }
+
+        public IList<LogicalLocation> Trim(IList<LogicalLocation> locations)
```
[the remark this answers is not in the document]
ScottLouvau: Fixed.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+            HashSet<LogicalLocation> uniqueLocations = new HashSet<LogicalLocation>(LogicalLocation.ValueComparer);
+            List<LogicalLocation> newLocations = new List<LogicalLocation>();
+
+            this.TotalLocations += locations.Count;
+            for (int i = 0; i < locations.Count; ++i)
+            {
+                LogicalLocation loc = locations[i];
+
+                // Trim Locations themselves
+                Trim(loc);
```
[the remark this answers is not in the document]
ScottLouvau: Fixed.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Writers
+{
+    /// <summary>
+    ///  SarifConsolidator is used to reduce the size of Sarif objects and the serialized log.
+    ///  It removes common redundant properties, uses Run-wide collections to avoid writing duplicate
+    ///  values, and can remove portions of Results entirely.
+    /// </summary>
+    public class SarifConsolidator
+    {
+        private Run _run;
```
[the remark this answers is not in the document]
ScottLouvau: Done.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        private Dictionary<ThreadFlowLocation, int> _uniqueThreadFlows;
+
+        public int TotalThreadFlowLocations { get; private set; }
+        public int UniqueThreadFlowLocations => _uniqueThreadFlows.Count;
+
+        public int TotalLocations { get; private set; }
```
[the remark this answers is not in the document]
ScottLouvau: Yes - this is both LogicalLocations and Locations. Not sure how else to name it. Suggestions?

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        public int UniqueLocations { get; private set; }
+
+        public int? MessageLengthLimitBytes { get; set; }
+        public bool RemoveCodeFlows { get; set; }
+        public bool RemoveRelatedLocations { get; set; }
+        public bool RemoveGraphs { get; set; }
```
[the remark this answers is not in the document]
ScottLouvau: Got it - made just one option which removes both.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+        public bool RemoveGraphTraversals { get; set; }
+        public bool RemoveStacks { get; set; }
+        public bool RemoveWebRequests { get; set; }
+        public bool RemoveWebResponses { get; set; }
+
+        public SarifConsolidator(Run run)
+        {
+            _run = run;
+            _uniqueThreadFlows = new Dictionary<ThreadFlowLocation, int>(ThreadFlowLocation.ValueComparer);
```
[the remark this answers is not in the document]
ScottLouvau: Renamed.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        public int TotalThreadFlowLocations { get; private set; }
+        public int UniqueThreadFlowLocations => _uniqueThreadFlows.Count;
+
+        public int TotalLocations { get; private set; }
+        public int UniqueLocations { get; private set; }
+
+        public int? MessageLengthLimitBytes { get; set; }
+        public bool RemoveCodeFlows { get; set; }
+        public bool RemoveRelatedLocations { get; set; }
```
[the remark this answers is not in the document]
ScottLouvau: Added a class-level remarks comment on SarifConsolidator which notes this.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+            result.CodeFlows = (this.RemoveCodeFlows ? null : result.CodeFlows);
+            result.RelatedLocations = (this.RemoveRelatedLocations ? null : result.RelatedLocations);
+            result.Graphs = (this.RemoveGraphs ? null : result.Graphs);
+            result.GraphTraversals = (this.RemoveGraphTraversals ? null : result.GraphTraversals);
+            result.Stacks = (this.RemoveStacks ? null : result.Stacks);
+            result.WebRequest = (this.RemoveWebRequests ? null : result.WebRequest);
+            result.WebResponse = (this.RemoveWebResponses ? null : result.WebResponse);
+
+            // Truncate long Messages if configured
+            if (this.MessageLengthLimitBytes > 0)
```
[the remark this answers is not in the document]
ScottLouvau: Renamed variable to be consistent.

src/Sarif/Writers/SarifConsolidator.cs
```diff
@@ -0,0 +1,259 @@
...
+                    }
+                }
+            }
+        }
+
+        public IList<Location> Trim(IList<Location> locations)
+        {
+            if (locations == null || locations.Count == 0) { return locations; }
+
+            HashSet<Location> uniqueLocations = new HashSet<Location>(Location.ValueComparer);
```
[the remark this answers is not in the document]
ScottLouvau: I added de-duplication logic for both location List types. I wanted the output to tell me how many ThreadFlowLocations I consolidated (by moving them to Run.ThreadFlowLocations) and how many other locations I consolidated (by removing duplicates within any single result). I didn't think breaking out Locations and LogicalLocations was worthwhile, because it was the same functional change.

I could separate them, but again, it's just to give the user output giving a sense of how much cleanup occurred.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -113,115 +195,8 @@ public class SarifLogger : IDisposable, IAnalysisLogger
...
-
-        public IDictionary<ReportingDescriptor, int> RuleToIndexMap
-        {
-            get
-            {
-                _ruleToIndexMap = _ruleToIndexMap ?? new Dictionary<ReportingDescriptor, int>(ReportingDescriptor.ValueComparer);
-                return _ruleToIndexMap;
-            }
-        }
+        public IDictionary<ReportingDescriptor, int> RuleToIndexMap { get; private set; }
```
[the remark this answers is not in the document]
ScottLouvau: Huh. So this is "readonly" for auto-properties? Weird. Anyway, it works, so I removed the setters.

src/Sarif/Writers/SarifLogger.cs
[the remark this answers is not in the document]
ScottLouvau: Leaving as-is per discussion. The current style here is that single line statements are acceptable when they have braces only. I argued with Chris Meyer about this also. It doesn't seem like there's an alternative that would make us all happier than the status quo. (Braces required, single line ok with braces).

src/Sarif/Writers/SarifLogger.cs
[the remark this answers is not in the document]
lgolding: Yes, that's exactly right.

---
In reply to: [324930138](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r324930138) [](ancestors = 324930138)

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
ScottLouvau: Since Mike wants me to move SarifTrim and SarifToCsv into the Multitool imminently, I'm going to leave these unchanged.

src/Sarif/Core/Run.cs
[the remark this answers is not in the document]
ScottLouvau: Larry, basically if we go to the trouble to move the Artifact to Run.Artifacts, we should enable "slimming down" the Result copy. In Spam logs, ArtifactLocation repeating URIs is a huge proportion of overall log size. 

Mike, is there a preference between DeepClone and the copy constructor? If so, why?

src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
```diff
@@ -21,9 +22,17 @@ public void PropertyBagConverter_RoundTripsStringPropertyWithEscapedCharacters()
             string stringPropertyName = nameof(stringPropertyName);
             string stringPropertyValue = "'\"\\'";
 
+            string normalStringPropertyName = "source.language";
+            string normalStringPropertyValue = "csharp";
+
+            string dateTimePropertyName = "runStart";
+            DateTime dateTimePropertyValue = DateTime.UtcNow.AddHours(-1.43);
```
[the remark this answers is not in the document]
ScottLouvau: Just an arbitrary value to test round tripping. Does it matter?

src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
ScottLouvau: Ran formatter on all SarifToCsv files.

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
ScottLouvau: Fixed.

ScottLouvau: Larry commented on this also. The core changes the consolidator makes should be meaning-preserving. When you ask it to remove sections wholesale, you are losing meaning in exchange for space savings. I want to make this possible in cases like Grant's where the log is just too big and you need a way to grapple with it.

I've added a class comment remark section on SarifConsolidator warning about this.

src/Samples/SarifTrim/Program.cs
ScottLouvau: Moved to Consolidator and added logic to remove the definitions also.

src/Sarif/Core/SarifLog.cs
ScottLouvau: No. JsonConvert doesn't have an overload which takes a TextReader or Stream. Using JsonConvert.Deserialize requires you to have a string.

This increases the memory cost of loading a file by another 2x the file size (for the two-byte-per-char string) and is slower.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Fixed.

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: Fixed.

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: Made setters private. There's no real reason for this class to need to be mutable anyway.

src/Sarif/Readers/SerializedPropertyInfoConverter.cs
ScottLouvau: Larry has logged #1673 for deeper testing here.

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: Yep

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: Changed to be in terms of constant. I still want to propose a default in the comment and propose a representation of it that someone calling this ctor should consider.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: If it isn't static and it uses instance members to record calculated thresholds, it's not thread-safe for no particular reason.

As a caller I just want to call MapBuilder.Build(jsonFilePath), so getting rid of something I have to construct and constraints on the something seemed like a good choice.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Fixed. Was this important?

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Short version: 
- Deferred loading requires a way to build another copy of the stream (to go back and read a DeferredList just-in-time). 
- JsonPositionedTextReader must take this "streamProvider" in order to ferry it from the calling code through to the DeferredList
- JsonPositionedTextReader is required to identify byte offsets in the file for the map.

So, it's not needed for this specific scenario, but JsonPositionedTextReader is needed for map construction and deferred loading, and I don't want you to be able to create one that won't work for your scenario, so I ask for it unconditionally.

src/Samples/SarifTrim/Program.cs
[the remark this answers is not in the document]
ScottLouvau: Removed into Consolidator with option and now removing the Run.OriginalUriBaseIds as well.

src/Sarif/Writers/SarifConsolidator.cs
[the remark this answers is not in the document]
ScottLouvau: Markdown help says '...' is converted to a nice ellipsis. I've appended one in both cases and I now truncate both. I'm not going to parse the markdown to ensure I'm not cutting apart a construct.

src/Sarif/Core/SarifLog.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.IO;
+using Microsoft.CodeAnalysis.Sarif.Readers;
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public partial class SarifLog
+    {
+        /// <summary>
+        ///  Load a Sarif file into a SarifLog object model instance using deferred loading.
```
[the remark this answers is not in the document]
ScottLouvau: Fixed.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: I've consolidated Graphs and GraphTraversals. I expect users who want to preserve log meaning to leave all of these options false. These are for when the user must remove content to make the log small enough to interpret. I thought about combining requests and responses but thought a user might want to keep requests (likely small) to confirm what was called but not responses (likely big) if they found they couldn't afford the file size.

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: Readonly isn't allowed here by the compiler and since I'm using =>, this is a property with no setter, so it's the functional equivalent of public static JsonMapSettings DefaultSettings { get; } with initialization in each constructor.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Good feedback. I struggled on the proper design; I've now separated two classes: JsonMapSettings (user preferences) and internal JsonMapRunSettings (user + calculated). The conversion happens in JsonMapBuilder.Build, when the file size is known.

This design now leaves a simple public static API for someone who wants a map (they don't need to construct a JsonMapBuilder instance for no clear reason), is thread safe, and safely allows reusing JsonMapSettings when simultaneously building multiple files (not that I'd expect that use commonly).

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: The "big old comment" is present as a class comment on JsonPositionedTextReader. I've added more to explain why streamProvider() is needed. I've added a comment in SarifLog.LoadDeferred to explain both uses. I've added a comment in JsonMapBuilder.Build to explain why JsonPositionedTextReader is needed but not streamProvider (as it's not needed for this scenario but demanded by JsonPositionedTextReader for other scenarios). A confused reader who goes to the constructor to see what overloads are available will hopefully see the class comment where it is explained.

src/Sarif/Map/JsonMapSettings.cs
ScottLouvau: I've redone this design based on comments from both of you.
There's now one class with user settings, and one internal class which adds calculated settings. Build() constructs the latter class given the user settings and then proceeds.

This leaves a simple static API for someone who wants a log, passes simple settings through the construction, is thread safe, and safely allows users to re-use the user provided portion of the settings.

src/Sarif/Map/JsonMapBuilder.cs
ScottLouvau: Huh. I actually prefer "unused" because it's self-documenting for people unfamiliar with that feature, but I've changed it as you suggest.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: Added ellipse along with Markdown length check per Larry.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: I don't know. There isn't: 
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/foreach-in
https://stackoverflow.com/questions/6455311/is-ifitems-null-superfluous-before-foreacht-item-in-items

New trick from StackOverflow:
foreach(var item in collection ?? Enumerable<Type>.Empty) 

It's terse, but I don't love it.

ScottLouvau: I'm going to leave as-is to better follow the design of the rest of the class. I generally null check each thing on the "Trim" entry of it, and check collections around both the foreach over elements and any collection-level manipulation together.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: Changed. The victim of repeated refactors in the initial authoring.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: I don't like how wordy and long very long argument names make the method body. I want the reader to focus on what I'm doing with the instance.

I see how since you like named parameters you want the name to be more descriptive (because otherwise the calling code will never show it).

I've changed it. This is a style issue where we disagree and should talk further at some point.

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: I really don't think it should be the SarifConsolidator's job to ensure every property from the ArtifactLocation is copied to the Run.Artifact when an index has been specified already by the producer. The consolidator should omit other properties when an index has been specified, because all of them are redundant, but I think that's a reasonable boundary for what it does.

src/Test.FunctionalTests.Sarif/UpdateBaselines.ps1
ScottLouvau: Yes. There's a base class for the E2E tests. It writes a message, which I updated, which says to copy the folder (and gives the full path) to update the tests.

src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
ScottLouvau: I chose names to try to be consistent with the object model properties or methods, so you know how to get the thing you asked for from code.

I don't mind Message and will rename it.

src/Samples/SarifToCsv/Program.cs
ScottLouvau: image

How did you get that to happen? In the debugger, I see null coming through and defaulting to false if I remove the setting from the app.config completely.

Are you running on the .NET Core runtime and I'm running .NET Framework?

src/Sarif/Writers/SarifConsolidator.cs
ScottLouvau: Good idea, fixed.

src/Sarif/Writers/SarifConsolidator.cs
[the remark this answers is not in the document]
ScottLouvau: I added one more rule - remove StartColumn when it's 1. The other rules can't be evaluated without the file text, which would be really expensive to evaluate in this context.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,8 @@
 # Sarif Driver and SDK Release History
 
+## **v2.1.17** [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.17) | [SDK](https://www.nuget.org/packages/Sarif.Sdk/2.1.17)
```
lgolding: >## [](start = 0, length = 2)

Heads up: I have a PR out that reverses the order of the release history, with newest versions first, as you naturally did here. If my PR goes in first, you'll have a merge to do (taking my version of the file in its entirety and prepending your change).

src/Samples/SarifToCsv/Program.cs
lgolding: I made a little test app. It runs on .NET Framework 4.6.1. I'll zip it up and mail it to you.

---
In reply to: [325840398](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r325840398) [](ancestors = 325840398)

lgolding: But I see your change accounts for both possibilities, so that's fine.

---
In reply to: [325909860](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r325909860) [](ancestors = 325909860,325840398)

lgolding: **UPDATE**: I looked at my test app again and it had a bug. Your original code is correct.

---
In reply to: [325910099](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r325910099) [](ancestors = 325910099,325909860,325840398)

src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
lgolding: Your rationale makes sense (consistency with SARIF OM). Feel free to revert.

---
In reply to: [325837196](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r325837196) [](ancestors = 325837196)

src/Sarif/Map/JsonMapSettings.cs
lgolding: Oh right, this is a get-only auto-property.

---
In reply to: [325825989](https://github.com/microsoft/sarif-sdk/pull/1672#discussion_r325825989) [](ancestors = 325825989)

src/Sarif/Writers/SarifConsolidator.cs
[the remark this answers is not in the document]
ScottLouvau: Overall, we agreed we want an "optimized" versus "optional" path for SarifLogger. We don't know how to expose it to callers yet; I've made it a boolean property (like Verbose) for the moment so I can conditionalize the code for these trimmings based on it. We can wire it through to a LoggingOption or other setting once we decide where to put them.

I think SarifConsolidator is committed to the "optimized" path. SarifLogger should allow configuring it.

src/Sarif/Writers/SarifLogger.cs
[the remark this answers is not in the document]
ScottLouvau: Made conditional on SarifLogger.Optimized, which has no public setter yet while we decide how to expose this control.

src/ReleaseHistory.md
ScottLouvau: Ok. I'll look for merge problems.

PR #1679, opened 2019-09-19
Fix #1678: SARIF1017 does not account for taxonomies

### LINKED ISSUE #1678: SARIF1017 does not account for taxonomies
This SARIF file specifies a `reportingDescriptorRelationship` from an analysis rule to a taxon within a taxonomy:

[Bug1677.sarif.txt](https://github.com/microsoft/sarif-sdk/files/3629027/Bug1677.sarif.txt)

It fails validation:

```
Sarif.Multitool.exe validate Bug1677.sarif
Analyzing...
Bug1677.sarif(21,29): error SARIF1017: runs[0].tool.driver.rules[0].relationships[0].target:
This "reportingDescriptorReference" object contains a property "index" with value 0,
but either "runs[0].tool.extensions[0].rules" is absent, or it has fewer than 1 elements.
Bug1677.sarif(24,38): error SARIF1017: runs[0].tool.driver.rules[0].relationships[0].target.toolComponent:
This "toolComponentReference" object contains a property "index" with value 0,
but either "runs[0].tool.extensions" is absent, or it has fewer than 1 elements.
Analysis completed successfully.
```

... because the validator assumes that the array-index-valued properties refer to `runs[0].tool.extensions`, whereas in fact they refer to `runs[0].tool.driver.taxonomies`.

The validation rule needs to account for this possibility. Note that only taxonomies are at issue here. There is nowhere in the spec where a `reportingDescriptorReference` can validly refer to a translation or a policy.

@dbjorge @michaelcfanning FYI

lgolding:


lgolding:
Decided against this design.

PR #1681, opened 2019-09-22
Remove duplicative 'Analyzing file...' messages.

michaelcfanning:
@lgolding 

Noticed that we've somehow introduced duplicative 'Analyzing blah...' messages (with verbose mode enabled). Now that ConsoleLogger is the only place where this occurs.


lgolding:
I see there were three different places where this message was potentially generated, and you left the one in `ConsoleLogger.AnalyzingTarget` that uses `SdkResources.MSG001_AnalyzingTarget`.

In that case, please remove the unused resources `SdkResources.MSG_Analyzing` and `DriverResources.MSG_Analyzing` (the latter wasn't used even before this PR).

Issue #1575, opened 2019-07-12
FxCop SarifSdkTest is disabled, and fails when enabled.


lgolding:
While fixing #1570, I noticed that after I caused the FxCop converter to emit `run.logicalLocations`, it didn't cause any tests to fail. It seems that there were no FxCop xml files with logical location information. How was that possible?

It turned out that there was one such file, SarifSdkTest.xml. But its corresponding expected output file SarifSdkTest.xml.sarif _did_ contain `run.logicalLocations`, even though the converter wasn't emitting it. Therefore that file must have been created at a time when the converter _did_ emit `run.logicalLocations`, and there was a subsequent regression where that functionality was lost.

Ok, fine, so why was the test passing _before_ I made the fix?

It's because the test wasn't running, and that's because, even though the project file copied the expected output file to the test bin directory, it did _not_ copy the input file.

When I enable the test by adding SarifSdkTest.xml to the project file, it fails. I don't completely understand why.

PR #1686, opened 2019-09-25
Fix #1684:  Multitool match-results-forward does not properly preserve conversion, taxonomies, or rules

### LINKED ISSUE #1684: Multitool match-results-forward does not properly preserve conversion, taxonomies, or rules
## Overview

While trying out the Multitool `match-results-forward` command in an attempt to document how to use it alongside `axe-sarif-converter`, we hit a variety of blocking issues that prevent us from being able to use it with our tool's sarif output.

The SARIF output files in the [microsoft/axe-sarif-converter](https://github.com/microsoft/axe-sarif-converter) are the motivating examples; below, the repro steps use a more minimal repro file that demonstrate all the same issues we encountered.

## Repro steps:

* Download [this repro SARIF file](https://gist.githubusercontent.com/dbjorge/df773200926c55c76fd062bea27059ff/raw/3c00f1779f0ed699114b072a7e657b377f29f24c/multitool-match-results-forward-content-preservation-repro.sarif) as `repro.sarif`
* Run the following commands to produce `output.sarif`:
  ```sh
  dotnet tool install --global sarif.multitool --version 2.1.16
  sarif match-results-forward --previous ./repro.sarif ./repro.sarif --output-file-path ./output.sarif --pretty-print
  ```
  ...or, see [output.sarif](https://gist.githubusercontent.com/dbjorge/df773200926c55c76fd062bea27059ff/raw/0e18a668e80bf74a771a6cd792e1a3e98bb9eb46/output.sarif)
* Observe the following issues with `output.sarif`:
  * The original run's `conversion` section has been omitted
  * The original run's `taxonomies` section has been omitted
  * The original tool driver's rule at index 0 with id `first-rule` has been duplicated at index 2
  * The second result (which originally references the rule at index 1 with id `second-rule`) points to `ruleIndex: 2, ruleId: 'second-rule'`, where ruleIndex 2 is actually the bogus copy of `first-rule`

lgolding:


src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -246,15 +262,27 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
                 indexRemappingVisitor.HistoricalLogicalLocations = resultPair.Run.LogicalLocations;
                 indexRemappingVisitor.VisitResult(result);
 
+                // This logic is incomplete. It doesn't account for the possibility that result.rule
+                // is present (in which case the rule might come from an extension rather than from
+                // the driver).
                 if (result.RuleIndex != -1)
                 {
-                    ReportingDescriptor rule = resultPair.Run.Tool.Driver.Rules[0];
```
lgolding: >0 [](start = 80, length = 1)

The hard-coded "0" was wrong. Find the rule that matches the id.

lgolding:

lgolding:

src/ReleaseHistory.md
```diff
@@ -1,44 +1,45 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
 ## **v2.1.17** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.17) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.17) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.17) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.17)
-* FEATURE: Introduce SarifConsolidator to shrink large log files. https://github.com/microsoft/sarif-sdk/issues/1675
-* BUGFIX: Analysis rule SARIF1017 incorrectly rejected index-valued properties that referred to taxonomies. https://github.com/microsoft/sarif-sdk/issues/1678
 * API NON-BREAKING: emit all core object model members as 'virtual'.
+* FEATURE: Introduce SarifConsolidator to shrink large log files. [#1675](https://github.com/microsoft/sarif-sdk/issues/1675)
+* BUGFIX: Analysis rule SARIF1017 incorrectly rejected index-valued properties that referred to taxonomies. [#1678](https://github.com/microsoft/sarif-sdk/issues/1678)
+* BUGFIX: `match-results-forward-command` dropped log contents and mishandled `rules` array. [#1684](https://github.com/microsoft/sarif-sdk/issues/1684)
```
lgolding: >BUGFIX [](start = 2, length = 6)

This is the new line. The rest of it is just me improving the Markdown by making the links more compact, _e.g._, `[[[https://github.com/microsoft/sarif-sdk/issues/1575`](https://github.com/microsoft/sarif-sdk/issues/1575`)](https://github.com/microsoft/sarif-sdk/issues/1575`)](https://github.com/microsoft/sarif-sdk/issues/1575`) => `[#1575]([[[https://github.com/microsoft/sarif-sdk/issues/1575)`](https://github.com/microsoft/sarif-sdk/issues/1575)`)](https://github.com/microsoft/sarif-sdk/issues/1575)`)](https://github.com/microsoft/sarif-sdk/issues/1575)`)

michaelcfanning: Just curious, why did you stop this scrub where you did? Did you just exhaust your 'opportunistinc improvements' capital? should we continue this clean-up as each of us modifies this file? (a reasonable approach)

---
In reply to: [328351950](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328351950) [](ancestors = 328351950)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -190,37 +190,53 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
...
-                Tool = tool,
-                AutomationDetails = currentRuns.First().AutomationDetails,
+                Tool = tool
             };
 
+            // If there was only one run, we can fill in more information because we don't need to
+            // worry about it being different from run to run.
+            if (currentRuns.Count() == 1)
+            {
+                run.AutomationDetails = firstRun.AutomationDetails;
```
michaelcfanning: >run [](start = 16, length = 3)

help me remember, why in this case don't we simply use 'firstRun' in its entirety?

lgolding: I'm just not 100% sure that everything makes sense to copy. I was confident of these and it was a large superset of what Dan asked for.

---
In reply to: [328779252](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328779252) [](ancestors = 328779252)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -246,15 +262,27 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
...
                 indexRemappingVisitor.VisitResult(result);
 
+                // This logic is incomplete. It doesn't account for the possibility that result.rule
+                // is present (in which case the rule might come from an extension rather than from
+                // the driver).
                 if (result.RuleIndex != -1)
                 {
-                    ReportingDescriptor rule = resultPair.Run.Tool.Driver.Rules[0];
-                    if (!reportingDescriptors.TryGetValue(rule, out int ruleIndex))
+                    ReportingDescriptor rule = resultPair.Run.Tool.Driver.Rules.FirstOrDefault(r => r.Id.Equals(result.RuleId, StringComparison.Ordinal));
```
michaelcfanning: >ReportingDescriptor [](start = 20, length = 19)

Is there an opportunity here to leverage the Result.GetRule(Run) method that is added to our Result object? See Core\Result.cs to assess. Rule look-up isn't easy. Would be great to keep expanding a single utility for this operation.

michaelcfanning:
>             //var ruleData = new Dictionary<string, ReportingDescriptor>();

not your problem, but can we remove this dead code?

---
Refers to: src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs:296 in d99ad1b. [](commit_id = d99ad1b59c83a9fdcdf80eaaa0eaf9a73b97b442, deletion_comment = False)

src/ReleaseHistory.md
lgolding: Exactly that. I stopped at the point where I thought "Michael's going to wonder how much time I spent on this" :-) Answer: a few minutes. Happy to finish it if you want!

---
In reply to: [328778424](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328778424) [](ancestors = 328778424,328351950)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
lgolding: Perfect! I didn't know/remember that one. It also means I can remove the comment above about the "incomplete logic" -- the logic is as complete as the helper is.

---
In reply to: [328779954](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328779954) [](ancestors = 328779954)

lgolding:
>             //var ruleData = new Dictionary<string, ReportingDescriptor>();

Done.

---
In reply to: [535644971](https://github.com/microsoft/sarif-sdk/pull/1686#issuecomment-535644971) [](ancestors = 535644971)

---
Refers to: src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs:296 in d99ad1b. [](commit_id = d99ad1b59c83a9fdcdf80eaaa0eaf9a73b97b442, deletion_comment = False)

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -199,5 +200,151 @@ public void V2ResultMatcher_WhenExistingSuppressedIssueIsNewlyUnsuppressed_DoesN
...
+                        TranslationMetadata = new TranslationMetadata
+                        {
+                            Name = TestConstants.TranslationMetadataName
+                        }
+                    }
+                },
+                Language = TestConstants.LanguageIdentifier,
+                RedactionTokens = new string[]
+                {
+                    TestConstants.RedactionToken
```
michaelcfanning: >TestConstants.RedactionToken [](start = 20, length = 28)

If you have time to search the source for 'REDACTED' and ensure you are happy with how DRY our world is around this constant, would be a nice to have.  I see 6 instance to review.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
lgolding: Done.

---
In reply to: [328781881](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328781881) [](ancestors = 328781881,328779954)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
michaelcfanning: Eventually we should have a precise understanding of this. i.e., it is a behavior that is worth consciously specifying. If this area breaks again, we really should consider doing that. This is a good change that unblocks a colleague. Currently, however this specific behavior (and its corresponding unit test) are on of those 'implementation has a de facto/poorly explained behavior and our test simply ensures there is no change in this unexplained behavior. Absolutely better than no testing, obviously.

---
In reply to: [328779874](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328779874) [](ancestors = 328779874,328779252)

src/ReleaseHistory.md
michaelcfanning: This file changes all the time, I suggest we use the 'waves on the beach' approach and fix it over time. The deserialization issues continue to ring in my head as an alarm bell. :)

---
In reply to: [328780288](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328780288) [](ancestors = 328780288,328778424,328351950)

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
lgolding: I only see three occurrences:

- The existing `SarifConstants.RedactedMarker`, which I forgot we had,
- The new `TestConstants.RedcationToken`, which I now see we don't need and have removed.
- A hard-coded occurrence in a single test in `SarifValidatorTests`, which I've fixed.

So now it's all cleaned up unless I missed something.

---
In reply to: [328782431](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328782431) [](ancestors = 328782431)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
lgolding: Agreed.

Note that the other part of this change (the rules array and the references to it from the results array) _is_ based on a precise understanding. The existing code was just wrong.

---
In reply to: [328783776](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328783776) [](ancestors = 328783776,328779874,328779252)

src/ReleaseHistory.md
lgolding: Agreed.

---
In reply to: [328784413](https://github.com/microsoft/sarif-sdk/pull/1686#discussion_r328784413) [](ancestors = 328784413,328780288,328778424,328351950)

PR #1688, opened 2019-09-26
Fix #1673 (property bag testing); Fix #1687 (NuGet packaging warning)

### LINKED ISSUE #1577: PrereleaseCompatibilityTransformer mishandles date/times in property bag

### LINKED ISSUE #1673: Enhance property bag testing
Per @michaelcfanning: 

Add tests for the range of types that might occur in a property bag. For example, consider serializing a `double` that exceeds `Int32.Max`. Pass a `float` that is not an integral value. Consider round-tripping a simple custom object. And serialize/deserialize arrays of every supported thing. Understand and test expected behaviors when setting property values to `null` as well.

### LINKED ISSUE #1687: Packaging gives warning NU5048
We get this warning for each NuGet package we create:
```
C:\Program Files\dotnet\sdk\3.0.100-preview9-014004\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(198,5):
error NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated.
Consider using the 'PackageIcon'/'icon' element instead.
Learn more at https://aka.ms/deprecateIconUrl
[C:\dev\sarif\sarif-sdk\src\Sarif.Driver\Sarif.Driver.csproj]
```

The error message tells the story and the instructions at https://aka.ms/deprecateIconUrl tell how to fix it. We can't refer to the icon through a URL; we need to store it locally and refer to it with the `<PackageIcon>` element. It turns out we already have a copy of the icon triskelion.png in the repo; we just weren't using it.

See also https://github.com/NuGet/Home/issues/8569

lgolding:
I added a comprehensive end-to-end test of round-tripping property bag properties. I also added a `DateTime` test to the `PropertyBagConverterTests` functional tests.

The problem that @ScottLouvau ran into with the `DateTime` test is related to my fix for #1577. I don't mean that my fix caused his problem; I mean that to avoid the problem, we need to do the same thing I did to fix #1577. For an explanation, see [this comment](https://github.com/microsoft/sarif-sdk/blob/83ecf0fa2d919a0be70a144464fc25d8b898a294/src/Sarif/SarifUtilities.cs#L190-L214) in SarifUtilities.cs.

In short, when Newtonsoft.Json deserializes a string-valued property that looks to it like a `DateTime`, its default behavior is to convert it to a `DateTime` object. That behavior breaks our property round-tripping. To avoid it, we need to call `JsonConvert.DeserializeObject` with a `JsonSerializerSettings` object that specifies `DateParseHandling.None`. That tells Newtonsoft.Json to leave string-valued properties alone and let us handle the `DateTime`-iness. In #1577, I introduced `SarifUtilities.DeserializeObject` to do exactly that.

@michaelcfanning and I have discussed this and we know it's not the end of the story. It's too easy to forget (or not to know about) the need to call the `SarifUtilities` method, which is exactly what happened in that existing unit test. But as far as verifying that our handling of property bag properties is good, I believe we now have sufficient test coverage to be confident of that.

Also:
- Fix a NuGet packaging warning.
- Improve our internal handling of property bag properties with `null` values.

### DIFF
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -61,8 +61,6 @@ namespace Microsoft.CodeAnalysis.Sarif
                         propertyName));
             }
 
-            // Avoid a crash if the string-valued property we are attempting to retrieve has
-            // a null value in the JSON file.
             if (Properties[propertyName] == null) { return null; }
 
             if (!Properties[propertyName].IsString)
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -63,6 +63,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                         propertyName));
             }
 
+            // Avoid a crash if the string-valued property we are attempting to retrieve has
+            // a null value in the JSON file.
             if (Properties[propertyName] == null) { return null; }
 
             if (!Properties[propertyName].IsString)
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -161,6 +161,11 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (value == null)
             {
+                // This is consistent with what the PropertyBagConverter does when it encounters
+                // a null-valued property. Whether we create a property bag dictionary entry
+                // by deserializing a null from the log file or by calling SetProperty("aProp", null),
+                // the internal representation is the same: a null value in the Properties
+                // dictionary.
                 Properties[propertyName] = null;
             }
             else
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -63,6 +63,8 @@ namespace Microsoft.CodeAnalysis.Sarif
                         propertyName));
             }
 
+            if (Properties[propertyName] == null) { return null; }
+
             if (!Properties[propertyName].IsString)
             {
                 throw new InvalidOperationException(SdkResources.CallGenericGetProperty);
@@ -84,7 +86,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 return true;
             }
 
-            value = default(T);
+            value = default;
             return false;
         }
 
@@ -155,10 +157,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         public void SetProperty<T>(string propertyName, T value)
         {
-            if (Properties == null)
-            {
-                Properties = new Dictionary<string, SerializedPropertyInfo>();
-            }
+            Properties = Properties ?? new Dictionary<string, SerializedPropertyInfo>();
 
             bool isString = typeof(T) == typeof(string);
 
@@ -217,10 +216,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         public void RemoveProperty(string propertyName)
         {
-            if (Properties != null)
-            {
-                Properties.Remove(propertyName);
-            }
+                Properties?.Remove(propertyName);
         }
 
         [JsonIgnore]
--- a/src/Sarif/Core/PropertyBagHolder.cs
+++ b/src/Sarif/Core/PropertyBagHolder.cs
@@ -17,8 +17,6 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// </summary>
     public class PropertyBagHolder : IPropertyBagHolder
     {
-        private const string NullValue = "null";
-
         protected PropertyBagHolder()
         {
             Tags = new TagsCollection(this);
@@ -74,10 +72,8 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             string value = Properties[propertyName].SerializedValue;
 
-            // Remove the quotes around the serialized value ("x" => x) -- unless it's null.
-            return value.Equals(NullValue, StringComparison.Ordinal)
-                ? null
-                : value.Substring(1, value.Length - 2);
+            // Remove the quotes around the serialized value ("x" => x).
+            return value.Substring(1, value.Length - 2);
         }
 
         public bool TryGetProperty<T>(string propertyName, out T value)
@@ -163,13 +159,14 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             bool isString = typeof(T) == typeof(string);
 
-            string serializedValue;
             if (value == null)
             {
-                serializedValue = NullValue;
+                Properties[propertyName] = null;
             }
             else
             {
+                string serializedValue;
+
                 if (isString)
                 {
                     serializedValue = JsonConvert.ToString(value);
@@ -190,9 +187,9 @@ namespace Microsoft.CodeAnalysis.Sarif
 
                     serializedValue = JsonConvert.SerializeObject(value, settings);
                 }
-            }
 
-            Properties[propertyName] = new SerializedPropertyInfo(serializedValue, isString);
+                Properties[propertyName] = new SerializedPropertyInfo(serializedValue, isString);
+            }
         }
 
         public void SetPropertiesFrom(IPropertyBagHolder other)
--- a/src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
+++ b/src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
@@ -56,9 +56,9 @@ namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests
             };
 
             var settings = new JsonSerializerSettings { Formatting = Formatting.Indented };
-            string originalLogText = JsonConvert.SerializeObject(originalLog, settings);
+            string originalLogContents = JsonConvert.SerializeObject(originalLog, settings);
 
-            SarifLog deserializedLog = SarifUtilities.DeserializeObject<SarifLog>(originalLogText);
+            SarifLog deserializedLog = SarifUtilities.DeserializeObject<SarifLog>(originalLogContents);
             run = deserializedLog.Runs[0];
 
             int integerProperty = run.GetProperty<int>(intPropertyName);
@@ -75,9 +75,15 @@ namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests
 
             run.GetProperty<DateTime>(dateTimePropertyName).Should().Be(dateTimePropertyValue);
 
-            string reserializedLog = JsonConvert.SerializeObject(deserializedLog, settings);
+            // In addition to checking the individual properties (which is nice for explicitness),
+            // we redundantly check that the entire SarifLog objects match.
+            // THIS ASSERTION IS COMMENTED OUT BECAUSE OF https://github.com/microsoft/sarif-sdk/issues/1689,
+            // "Generated code for property bag comparison is incorrect."
+            //originalLog.ValueEquals(deserializedLog).Should().BeTrue();
 
-            reserializedLog.Should().Be(originalLogText);
+            string reserializedLogContents = JsonConvert.SerializeObject(deserializedLog, settings);
+
+            reserializedLogContents.Should().Be(originalLogContents);
         }
     }
 }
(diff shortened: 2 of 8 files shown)


src/build.props
```diff
@@ -92,14 +92,18 @@
...
+    <PackageIcon>triskelion.png</PackageIcon>
     <!--
     Don't complain about SemVer 2.0.0-compatible version strings.
     See https://github.com/NuGet/Home/issues/4687#issuecomment-393302779.
     -->
     <NoWarn>NU5105</NoWarn>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Include="..\..\triskelion.png" Pack="true" PackagePath="\"/>
```
lgolding: >None [](start = 5, length = 4)

This fixes the warning `NU5048` that we get as we build each package.

src/Test.UnitTests.Sarif/TestData/RoundTripping/RoundTripping.sarif
```diff
@@ -0,0 +1,42 @@
...
+            "date": "2019-09-25T15:07Z",
+            "string": "Hello, string!",
+            "integer": 42,
+            "bigInteger": 5000000000,
+            "true": true,
+            "false": false,
+            "dateArray": [
+  "2019-09-26T15:52Z",
+  "2019-09-26T15:54Z"
+],
```
lgolding: This is how `Formatting.Indented` renders the arrays and sub-objects within `properties`. I don't know why. I wonder if it's related to the actual bug we're trying to ferret out (failure to round-trip certain property bag properties).

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -63,6 +63,8 @@ public string GetProperty(string propertyName)
                         propertyName));
             }
 
+            if (Properties[propertyName] == null) { return null; }
```
lgolding: >Properties [](start = 16, length = 10)

This fixes a crash when you attempted to deserialize a null-valued property (`"aString": null`) into a string with `GetProperty("aString")`.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -63,17 +61,19 @@ public string GetProperty(string propertyName)
...
             if (!Properties[propertyName].IsString)
             {
                 throw new InvalidOperationException(SdkResources.CallGenericGetProperty);
             }
 
             string value = Properties[propertyName].SerializedValue;
 
-            // Remove the quotes around the serialized value ("x" => x) -- unless it's null.
-            return value.Equals(NullValue, StringComparison.Ordinal)
-                ? null
```
lgolding: >null [](start = 18, length = 4)

This null check is no longer needed because we simplified the way that `SetProperty` deals with `null` values (`SetProperty("aProp", null)`). Rather than creating a `SerializedPropertyInfo` object with `SerializedValue` set to `null` and placing that object into `Properties["aProp"]`, we simply set `Properties["aProp"]` to `null`.

This is consistent with what the `PropertyBagConverter` does when it encounters a `null` valued property. Before this change, a `null` valued property read from the log file had a different internal representation from a property you created programatically with `SetProperty("aProp", null)`.

This is one of those explanatory comments that doesn't have a natural place in the source code. I'm explaining why some code _no longer appears_. But, come to think of it, I can add a comment on the `null` path in `SetProperty`, so I'll do that.

src/Test.UnitTests.Sarif/TestData/RoundTripping/RoundTripping.sarif
lgolding: Oh, it's because the `PropertyBagConverter` has taken over the serialization at this point and it doesn't retain the indentation.

---
In reply to: [328869713](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r328869713) [](ancestors = 328869713)

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -217,10 +218,7 @@ public void SetPropertiesFrom(IPropertyBagHolder other)
 
         public void RemoveProperty(string propertyName)
         {
-            if (Properties != null)
-            {
-                Properties.Remove(propertyName);
-            }
+                Properties?.Remove(propertyName);
```
ScottLouvau: It looks like the design is settled here, but I wonder if we should've thrown in this case. Callers could try to clean up properties on the wrong objects in the SARIF hierarchy and not realize nothing is actually happening underneath.

src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
```diff
@@ -28,34 +27,38 @@ public void PropertyBagConverter_RoundTrip()
...
             };
 
             var settings = new JsonSerializerSettings { Formatting = Formatting.Indented };
             string originalLogText = JsonConvert.SerializeObject(originalLog, settings);
 
-            SarifLog deserializedLog = JsonConvert.DeserializeObject<SarifLog>(originalLogText);
+            SarifLog deserializedLog = SarifUtilities.DeserializeObject<SarifLog>(originalLogText);
             run = deserializedLog.Runs[0];
 
             int integerProperty = run.GetProperty<int>(intPropertyName);
```
ScottLouvau: Is there a ValueComparer you can use to verify the entire Run object seems to match the Deserialized one? I think it's good to have the per-property checks for clarity, but an overall compare might be good redundant verification.

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            dateTimeProperty.Should().Be(ParseUtcDateTime("2019-09-25T15:07:01Z"));
+
+            string stringProperty = holder.GetProperty("string");
+            stringProperty.Should().Be("Hello, string!");
+
+            int intProperty = holder.GetProperty<int>("int");
+            intProperty.Should().Be(42);
+
+            long longProperty = holder.GetProperty<long>("long");
+            longProperty.Should().Be(5000000000);
```
ScottLouvau: Comment affirming that this example is over int.MaxValue?

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            bool falseProperty = holder.GetProperty<bool>("false");
+            falseProperty.Should().BeFalse();
+
+            string nullProperty = holder.GetProperty("null");
+            nullProperty.Should().BeNull();
+
+            DateTime[] dateTimeArray = holder.GetProperty<DateTime[]>("dateTimeArray");
+            dateTimeArray.Length.Should().Be(2);
+            dateTimeArray[0].Should().Be(ParseUtcDateTime("2019-09-26T15:52:02Z"));
+            dateTimeArray[1].Should().Be(ParseUtcDateTime("2019-09-26T15:54Z"));
```
ScottLouvau: How much precision do we emit for DateTimes? Does UtcNow roundtrip properly?

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
lgolding: That comment appears on Line 55.

---
In reply to: [329664092](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329664092) [](ancestors = 329664092)

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
lgolding: Does `UtcNow` round-trip properly? Yes.

How much precision do we emit for `DateTime`s?

    - For `DateTime`s in "first-class" properties, we emit three decimal digits. Example: `log.Runs[0].Invocations[0].StartTimeUtc = DateTime.UtcNow;` => `"startTimeUtc": "2019-09-30T17:10:20.185Z"`. 
    - For `DateTime`s in property bag properties, we emit seven decimal digits. Example: `log.Runs[0].SetProperty("now", DateTime.UtcNow)` => `"properties": { "now": "2019-09-30T17:10:20.1851531Z" }` .

The reason for the difference is subtle. First-class `DateTime`-valued properties are decorated with `[JsonConvert(typeof(DateTimeConverter)]`, and `DateTimeConverter` uses `SarifUtilities.SarifDateTimeFormatMillisecondsPrecision`. But `PropertyBagHolder.SetProperty` calls `JsonConvert.SerializeObject` _without a contract resolver_, so the value is serialized with Newtonsoft.Json's default `DateTime` serialization, which (apparently) uses more precision.

`SetProperty` doesn't set a contract resolver because it doesn't know whether it's working with a v1 or a v2 log, so it doesn't know whether to use the v1 or v2 contract resolver. However, I can add code to use the `DateTimeConverter` for `DateTime`-valued properties. Then everything will serialize with millisecond precision.

---
In reply to: [329664303](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329664303) [](ancestors = 329664303)

lgolding: Ugh. Doing that breaks the unit test `PropertyBagHolder_SetProperty_SetsDateTimeProperty`, which has the comment:

> DateTime instances should not be converted to UTC when persisted to a property bag.  The reason is that we can't be  sure that the user has an actual UTC time in hand, or even the ability to convert the property date time to UTC.

Nightmare. I need to think more about this. After lunch.

---
In reply to: [329717572](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329717572) [](ancestors = 329717572,329664303)

lgolding: Ok, I've thought about it. I can live with `DateTime`-valued property bag properties _that are created with `SetProperty`_ having more digits than first-class `DateTime`-valued properties. It's not great but IMO we're at the point of diminishing returns on this.

---
In reply to: [329725369](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329725369) [](ancestors = 329725369,329717572,329664303)

src/Test.FunctionalTests.Sarif/PropertyBagConverterTests.cs
lgolding: Yes, each generated class (`Run`) has an associated generated equality comparer class (`RunEqualityComparer`) with a static `Instance` property. The generated class has a static `ValueComparer` field that is set to that instance, and a `ValueEquals` method that invokes `Equals` on the instance:

```
    public partial class Run : PropertyBagHolder, ISarifNode
    {
        public static IEqualityComparer<Run> ValueComparer => RunEqualityComparer.Instance;

        public bool ValueEquals(Run other) => ValueComparer.Equals(this, other);
```

But the equality comparison fails because of a code generation bug! When comparing the individual `SerializedPropertyInfo` objects in the property bag, the generated comparison code does an object equality comparison instead of invoking `SerializedPropertyInfo.ValueComparer.Equals`.

I filed [#1689](https://github.com/microsoft/sarif-sdk/issues/1689), "Generated code for property bag comparison is incorrect".

---
In reply to: [329663713](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329663713) [](ancestors = 329663713)

src/Sarif/Core/PropertyBagHolder.cs
lgolding: It's a good point. Changing this behavior would of course be breaking (not that that's ever stopped us before ;-)).

---
In reply to: [329663092](https://github.com/microsoft/sarif-sdk/pull/1688#discussion_r329663092) [](ancestors = 329663092)

PR #1690, opened 2019-10-01
Result Matching Improvements

ScottLouvau:
Based on issues found in Compliance Run result matching.

1. Exclude Results already 'Absent' in the Baseline from output. (Only report results absent the first time)
2. Sort Results by RuleID after 'where' criteria so matches from different rules on the same line are compared by Rule, as they should be.
3. Canonicalize any line and column numbers in Messages before comparison to match moved results where the message contains a line or column number.

These reduce mismatched results which should match from 76 to a handful.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -3,6 +3,7 @@
 ## **v2.1.18** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.18) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.18) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.18) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.18)
 * FEATURE: Enhanced property bag serialization unit testing. [#1673](https://github.com/microsoft/sarif-sdk/issues/1673)
 * BUGFIX: Fix packaging warning NU5048 during build. [#1687](https://github.com/microsoft/sarif-sdk/issues/1687)
+* BUGFIX: Improve Result Matching sort and comparison.
 
 ## **v2.1.17** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.17) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.17) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.17) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.17)
 * API NON-BREAKING: emit all core object model members as 'virtual'.
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -3,7 +3,9 @@
 ## **v2.1.18** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.18) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.18) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.18) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.18)
 * FEATURE: Enhanced property bag serialization unit testing. [#1673](https://github.com/microsoft/sarif-sdk/issues/1673)
 * BUGFIX: Fix packaging warning NU5048 during build. [#1687](https://github.com/microsoft/sarif-sdk/issues/1687)
-* BUGFIX: Improve Result Matching sort and comparison.
+* BUGFIX: Result Matching now omits previously Absent results.
+* BUGFIX: Result Matching properly compares results from the same RuleID when multiple Rules match the same source line.
+* BUGFIX: Result Matching works when a result moves and has the line number in the message.
 
 ## **v2.1.17** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.17) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.17) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.17) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.17)
 * API NON-BREAKING: emit all core object model members as 'virtual'.
--- a/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
+++ b/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
@@ -136,9 +136,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
 
         private static void AddUnmatchedResults(List<ExtractedResult> baselineResults, List<ExtractedResult> currentResults, List<MatchedResults> matchedResults)
         {
+            // Add unmatched results from Baseline log which weren't already Absent in previous run
             foreach (ExtractedResult result in baselineResults)
             {
-                matchedResults.Add(new MatchedResults(result, null));
+                if (result.Result.BaselineState != BaselineState.Absent)
+                {
+                    matchedResults.Add(new MatchedResults(result, null));
+                }
             }
 
             foreach (ExtractedResult result in currentResults)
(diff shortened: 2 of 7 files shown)


src/ReleaseHistory.md
```diff
@@ -3,6 +3,7 @@
 ## **v2.1.18** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.18) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.18) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.18) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.18)
 * FEATURE: Enhanced property bag serialization unit testing. [#1673](https://github.com/microsoft/sarif-sdk/issues/1673)
 * BUGFIX: Fix packaging warning NU5048 during build. [#1687](https://github.com/microsoft/sarif-sdk/issues/1687)
+* BUGFIX: Improve Result Matching sort and comparison.
```
lgolding: >. [](start = 53, length = 1)

Consider filing an issue and citing it here (see line above).

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -136,9 +136,13 @@ private SarifLog BaselineSarifLogs(IEnumerable<Run> previous, IEnumerable<Run> c
 
         private static void AddUnmatchedResults(List<ExtractedResult> baselineResults, List<ExtractedResult> currentResults, List<MatchedResults> matchedResults)
         {
+            // Add unmatched results from Baseline log which weren't already Absent in previous run
             foreach (ExtractedResult result in baselineResults)
             {
-                matchedResults.Add(new MatchedResults(result, null));
+                if (result.Result.BaselineState != BaselineState.Absent)
```
lgolding: >Absent [](start = 65, length = 6)

Is this the right test? What if it was absent before but now it's present again?

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -188,5 +189,29 @@ private static string GetFirstSnippet(ExtractedResult result)
...
+        private static string GetCanonicalizedMessage(ExtractedResult result)
+        {
+            string rawMessage = result.Result.GetMessageText(result.Result.GetRule(result.OriginalRun));
+
+            // Canonicalize the message by replacing any line numbers in it with consistent markers
+            Region firstRegion = result.Result?.Locations?.FirstOrDefault()?.PhysicalLocation?.Region;
+            if (firstRegion != null)
+            {
+                rawMessage = rawMessage
+                    .Replace(firstRegion.StartLine.ToString(), "~SL~")
```
lgolding: >( [](start = 59, length = 1)

Stringize with `CultureInfo.InvariantCulture`.

lgolding: >Replace [](start = 21, length = 7)

Should we do a more precise match so if `StartLine` is 2 we don't replace "42" with "42~SL~"?

src/Sarif/Baseline/V2/WhatComparer.cs
```diff
@@ -188,5 +189,29 @@ private static string GetFirstSnippet(ExtractedResult result)
...
+                    .Replace(firstRegion.StartColumn.ToString(), "~SC~")
+                    .Replace(firstRegion.EndLine.ToString(), "~EL~")
+                    .Replace(firstRegion.EndColumn.ToString(), "~EC~");
+            }
+
+            return rawMessage;
+        }
+
+        private static bool IsNonEmptyAndEquals(string left, string right)
```
lgolding: >IsNonEmptyAndEquals [](start = 28, length = 19)

Unused?

lgolding:
Nits:
- Periods at ends of sentences.

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -161,6 +186,47 @@ public void V2ResultMatcher_TwoAdded()
...
+            Run firstRun = SampleRun.DeepClone();
+
+            // Copy the first Result and change the Rule only (they'll have same Message, Fingerprints, Location)
+            firstRun.Results[1] = firstRun.Results[0].DeepClone();
+            firstRun.Results[1].RuleId = "NewRuleId";
+
+            // Make another copy of each result and move them so that the results won't have any per-rule unique traits
+            firstRun.Results.Add(firstRun.Results[0].DeepClone());
+            firstRun.Results.Add(firstRun.Results[1].DeepClone());
+            firstRun.Results[5].Locations[0].PhysicalLocation.Region.StartLine += 1;
```
lgolding: >5 [](start = 29, length = 1)

Consider counting the results before adding the new ones so you can calculate these indices instead of hard-coding them (resilient if the test file ever sprouts new results).

ScottLouvau: Good idea.

src/Sarif/Baseline/V2/WhatComparer.cs
ScottLouvau: Yep - previous experiment.

ScottLouvau: Removed.

src/Sarif/Baseline/V2/WhatComparer.cs
ScottLouvau: Fixed.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
michaelcfanning: Then it will be regarded as new. Anything which appears that did not exist previously is new. What other state would make sense? Updated, Unchanged don't apply. You've put your finger on a core (by design) limitation in our system, we don't retain any historical information to produce a baseline state of 'Reappeared'.

This is a hard problem. On the one hand, there's the scenario of someone fixing a bug and immediately reverting it from a bad merge. In that case, it probably is the same issue (and you'd want to go find the old bug). On the other hand, results that look the same can emerge and there's no good outcome from finding an old bug that addressed the code in a different condition. I think this code is right, for now, but this conversation is likely to continue...


---
In reply to: [330294798](https://github.com/microsoft/sarif-sdk/pull/1690#discussion_r330294798) [](ancestors = 330294798)

src/Sarif/Baseline/V2/WhatComparer.cs
ScottLouvau: I think it's better not to, actually. We'd likely need a Regex to describe the boundaries (non-digits around the number) which would be a lot slower, particularly if the message is huge.

If the Result doesn't move, even when the replacement is cutting up unrelated numbers, it'll do so the same way for the old and new Result, so they'll still match.

So the case where this hinders matching is when the Result moves and the line number before or after happened to be an unrelated number in the message.  If we see that happening too often, I can certainly revisit this.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
ScottLouvau: I think so. Basically we either filter these results before or after matching. This code does it after. If a Result was Absent in the baseline but came back, it'll match again and show up as Existing this time. Only results which remained unmatched will be filtered.

One could argue for either behavior - it might be more clear for an issue matching a previously absent one to be marked "New". However, we also see a need to match issues to ensure we update the same bug as before, which makes considering Absent results when matching a benefit.

I think we'll know more as this code gets more practical use about which way to go.

src/ReleaseHistory.md
ScottLouvau: Hmm. I tried to, but realized it's three different bugs and I don't want to write them all and cite them all. =)

ScottLouvau: I've called out the three changes instead via comments, so at least the Release Notes describe the specific improvements.

PR #1691, opened 2019-10-01
Query, SarifToCsv improvements

ScottLouvau:
**Contentious**: This change required "query" mode to support Uris, which required getting the Run from a Result in order to look up Run.Artifacts when the Result ArtifactLocation has an index. Per e-mail discussion I tried implementing this as a Result.Run property which may or may not be populated and a Run method to fill it out. **Please** comment on the design via the mail thread and just smaller details here.

I added these for my Result Matching debugging workflow:

- Run result matching
- Use SarifToCsv to look at the results
- Find some confusing ones
- Use Sarif.MultiTool query to put them in a log alone
- Run Result Matching in the debugger to investigate

I don't like the design of SetEvaluator, but don't know how else to design it and keep the ability to reuse the other evaluators for primitive matching. I'd like to get it in rather than keep thinking about it, however, so others using query mode can use it for sets of things and not just single properties.

- SarifToCsv: Add support for Run.Tool.Name.
- SarifToCsv: Look up RuleId when Result uses RuleIndex.
- Query: Don't write out Runs which had no remaining Results.
- Query: Add support for string "StartsWith", "Contains", "EndsWith"
- ResultEvaluator: Add support for Uri (meaning: any Location -> ArtifactLocation.Uri matches expression)
- IExpressionEvaluator loosened to require ICollection only (Count needed by And and Or)
- SetEvaluator added to allow matching whole sets


src/Sarif/Core/Run.cs
```diff
@@ -138,7 +140,37 @@ private void InitializeFileToIndexMap()
...
+
+        /// <summary>
+        ///  Set the Run property on each Result to this Run, so that Result methods
+        ///  and properties which may need to look up Run collections can do so.
+        /// </summary>
+        public void SetRunOnResults()
+        {
+            if (this.Results != null)
+            {
+                DeferredList<Result> deferredResults = this.Results as DeferredList<Result>;
```
michaelcfanning: >DeferredList [](start = 16, length = 12)

Oh boy. This pattern makes me wonder whether we need all Lists in the format to be able to distinguish between these conditions? Otherwise we're on a journey to endlessly chase down bugs in various iteration scenarios. Not relevant to this check-in.

ScottLouvau: Yes, this part was pretty lame.

*If* the deserialization of Results could automatically pick up on the Run, the Deferred infrastructure would probably "just work" without this special stuff.

Otherwise, anywhere we have an IList<T> which may be deferred and we have an "outer object reference setter" like this, we need the same pattern.

Maybe we could figure out how to design this so that we can load a comprehensive log with the deferred OM and confirm in one go that all of the upward object references were properly set.

src/Sarif.Multitool/QueryCommand.cs
```diff
@@ -71,11 +75,14 @@ public int RunWithoutCatch(QueryOptions options)
...
                     foreach (Result result in run.Results)
                     {
-                        Console.WriteLine(result.FormatForVisualStudio(result.GetRule(run)));
+                        Console.WriteLine(result.FormatForVisualStudio());
                     }
                 }
             }
 
+            // Remove any Runs with no remaining matches
+            log.Runs = log.Runs.Where(r => (r?.Results?.Count ?? 0) > 0).ToList();
```
michaelcfanning: >log.Runs [](start = 12, length = 8)

I think this is the right decision. In result matching, we wouldn't want to discard a run that was left with no results. That empty run indicates 'run succeeded and hooray, it appears everything was fixed.' here the empty runs resulting from a query simply would require unnecessary post-processing. Just wanted to think through whether there's a useful distinction between something being entirely absent vs. explicitly present and empty here. I don't think there is, interested in your take.

ScottLouvau: Good point. In the query case, it's nice to filter empty runs out to reduce noise when reading the log or when debugging subsequent actions on the log. If a future query scenario makes us want to keep empty runs, we could expose an option at that time.

src/Sarif/Readers/DeferredList.cs
```diff
@@ -109,6 +111,22 @@ private void BuildPositions(JsonPositionedTextReader reader, long currentOffset)
...
+        /// <param name="transformer">Method to change Result instance and return updated Result</param>
+        public void AddTransformer(Func<T, T> transformer)
+        {
+            if (_transformer == null)
+            {
+                _transformer = transformer;
+            }
+            else
+            {
+                _transformer = (item) => transformer(_transformer(item));
```
michaelcfanning: >transformer [](start = 41, length = 11)

This code is non-obvious in terms of what gets evaluated when. Can you clarify in a code comment? May be faster to discuss this in a call.

src/Sarif/Core/SarifLog.cs
```diff
@@ -44,5 +44,31 @@ public static SarifLog Load(string sarifFilePath)
...
             }
         }
+
+        /// <summary>
+        ///  Write a SARIF log to disk as a file.
+        /// </summary>
+        /// <param name="sarifFilePath">File Path to Sarif file to write to</param>
+        public void Save(string sarifFilePath)
+        {
+            using (StreamWriter writer = File.CreateText(sarifFilePath))
```
michaelcfanning: >File.CreateText [](start = 41, length = 15)

We may require all IO in this file to go through our test IFileSystem interface in future, just a heads up, no note for now.

src/Sarif/Core/Run.cs
michaelcfanning: sounds good.

---
In reply to: [330692856](https://github.com/microsoft/sarif-sdk/pull/1691#discussion_r330692856) [](ancestors = 330692856)

src/Sarif/Readers/DeferredList.cs
ScottLouvau: Yes, comment added.

src/Sarif/Core/SarifLog.cs
ScottLouvau: Ok. I will definitely try to add both Stream and file overloads at least; Stream is the most broadly useful and file path is an overload that code consumers want to see there so they don't have to think. Happy to add others to suit design rules if needed.

PR #1694, opened 2019-10-07
Improve handling of DateTime property bag properties

lgolding:
Fix a bug in `PropertyBagConverter` that prevented `DateTime`-valued property bag properties from round-tripping correctly. With this fix in place, it is no longer necessary to work around the problem by circumventing Newtonsoft.Json's handling of `DateTime`s, and so the helper method `SarifUtilities.DeserializeObject` is no longer needed.

Revert calls to the deleted method to their state before the method was introduced. In most cases, the previous call was to `JsonConvert.DeserializeObject`. In one case, the previous call was to `JObject.Parse`.


src/Test.UnitTests.Sarif/TestData/RoundTripping/RoundTripping.sarif
```diff
@@ -26,7 +26,7 @@
             "dateTimeArray": [
   "2019-09-26T15:52:02",
   "2019-09-26T15:54",
-  "2019-09-26T15:55:05.123456789"
+  "2019-09-26T15:55:05.1234567"
```
lgolding: >1234567 [](start = 23, length = 7)

Now that we once again rely on Newtonsoft.Json's serialization and deserialization of `DateTime`s, they only round-trip 7 fractional second digits.

lgolding:

lgolding:

src/Sarif/Readers/PropertyBagConverter.cs
```diff
@@ -49,6 +49,12 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                     serializedValue = JsonConvert.ToString(serializedValue);
                     isString = true;
                 }
+                else if (propertyType == typeof(DateTime))
```
lgolding: >else [](start = 16, length = 4)

The fix. A simple `else` might also work but this is the minimal change that fixes the bug.

src/Test.UnitTests.Sarif/TestData/RoundTripping/RoundTripping.sarif
ScottLouvau: That seems sufficient.

PR #1696, opened 2019-10-08
Fix #1695: SarifLogger.Optimized cannot be set.

### LINKED ISSUE #1695: SarifLogger.Optimized cannot be set.

lgolding:
We introduce an option to set the `Optimize` flag on the `SarifLogger`, which results in a smaller but non-human-readable log file.

There are a few issues to consider:
- I changed the name of the flag from `Optimized` to `Optimize` because the other option names (`pretty-print`, `overwrite-output-file`, etc.) are imperative verbs, whereas `Optimized` was an adjective.
- It wasn't clear whether this flag belongs on `CommonOptionsBase` or `AnalyzeOptionsBase`. Some options that control file contents (`hashes`, `environment`, `invocation-properties`) are on `AnalyzeOptionsBase`, while others (`insert`, `remove`) are on `CommonOptionsBase`. We should make a call and move them to one place or the other.
- Setting `Default` to `true` on a `Boolean` option doesn't work (we've run into this before). Setting it to `false` reverses the behavior we had until now, where it was _hard-coded_ to `true`. We should either decide it's ok to change that behavior, or change the name and the sense of the flag (for example, `human-readable`).

Please comment on these issues.

### DIFF
--- a/src/build.props
+++ b/src/build.props
@@ -12,7 +12,7 @@
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
     
     <!-- VersionPrefix denotes the current Semantic Version for the SDK. Must be updated before every nuget drop. -->
-    <VersionPrefix>2.1.19</VersionPrefix>
+    <VersionPrefix>2.1.20</VersionPrefix>
     
      <!-- SchemaVersionAsPublishedToSchemaStoreOrg identifies the current published version on json schema store at https://schemastore.azurewebsites.net/schemas/json/ -->
     <SchemaVersionAsPublishedToSchemaStoreOrg>2.1.0-rtm.4</SchemaVersionAsPublishedToSchemaStoreOrg>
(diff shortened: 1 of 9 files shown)


src/build.props
```diff
@@ -30,7 +30,7 @@
     place. These properties are actually used by the PowerShell script that
     hides the previous package versions on nuget.org.
     -->
-    <PreviousVersionPrefix>2.1.18</PreviousVersionPrefix>
+    <PreviousVersionPrefix>2.1.17</PreviousVersionPrefix>
```
lgolding: >17 [](start = 31, length = 2)

Because we never successfully published .18 or .19.

lgolding:
>         Verbose = 0x2,

Well, ok, we _did_ have a mix of imperative verbs and adjectives. I don't care what you decide on the naming of the "Optimize/Optimized" flag.

---
Refers to: src/Sarif/Writers/LoggingOptions.cs:17 in af6aa35. [](commit_id = af6aa355cb4e96941e94d26db6d5fc51c4431700, deletion_comment = False)

PR #1697, opened 2019-10-09
Deterministic analysis order

michaelcfanning:


michaelcfanning:
Provide deterministic ordering of driver rules to maximize deterministic ordering of log files. In order to utilized SortedHashSet, was required to update public API to prefer ISet over HashSet types.

@lgolding

PR #1704, opened 2019-10-16
Splitting key value on first ':' only

cfaucon:
Splitting key value on first ':' only as value itself could contain a ':' which should be respected.

### DIFF
--- a/src/Sarif.WorkItemFiling/Extensions.cs
+++ b/src/Sarif.WorkItemFiling/Extensions.cs
@@ -50,7 +50,7 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
 
                 foreach (string fieldKvp in fieldKvps)
                 {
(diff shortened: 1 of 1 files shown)


src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -50,7 +50,7 @@ public static WorkItemFilingMetadata CreateWorkItemFilingMetadata(this SarifLog
 
                 foreach (string fieldKvp in fieldKvps)
                 {
-                    string[] kv = fieldKvp.Split(':');
+                    string[] kv = fieldKvp.Split(new char[] { ':' }, 2);
 
                     if (kv.Length == 2)
```
rtaket: do we need this check anymore?

cfaucon: Yes because it will split at most 2 but could be less.

PR #1705, opened 2019-10-17
Baselining Improvements

ScottLouvau:
- Using Result.Guid as default for Result.CorrelationGuid in baselining (when first assigning BaselineGuid and when copying from an older matched result.
- Null hardening in Result Matching.
- Adding SarifLog.Load/Save overloads which take a Stream.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -6,11 +6,14 @@ Now that an underlying bug in `PropertyBagConverter` has been fixed, there is no
 * FEATURE: Expanding Sarif SDK query mode to support Result.Uri, string StartsWith/EndsWith/Contains.
 * FEATURE: Adding Result.Run and a populating method, so that methods which need the Run context for a given Result have an integrated way to retrieve it.
 * FEATURE: Enhanced property bag serialization unit testing. [#1673](https://github.com/microsoft/sarif-sdk/issues/1673)
+* FEATURE: Added Stream-based SarifLog.Load and Save overloads
 * BUGFIX: Fix packaging warning NU5048 during build. [#1687](https://github.com/microsoft/sarif-sdk/issues/1687)
 * BUGFIX: SarifLogger.Optimized could not be set from the command line. [#1695](https://github.com/microsoft/sarif-sdk/issues/1695)
 * BUGFIX: Result Matching now omits previously Absent results.
 * BUGFIX: Result Matching properly compares results from the same RuleID when multiple Rules match the same source line.
 * BUGFIX: Result Matching works when a result moves and has the line number in the message.
+* BUGFIX: Result Matching uses Result.Guid as default CorrelationGuid in matched Results.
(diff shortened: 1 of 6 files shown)


src/ReleaseHistory.md
```diff
@@ -6,11 +6,14 @@ Now that an underlying bug in `PropertyBagConverter` has been fixed, there is no
 * FEATURE: Expanding Sarif SDK query mode to support Result.Uri, string StartsWith/EndsWith/Contains.
 * FEATURE: Adding Result.Run and a populating method, so that methods which need the Run context for a given Result have an integrated way to retrieve it.
 * FEATURE: Enhanced property bag serialization unit testing. [#1673](https://github.com/microsoft/sarif-sdk/issues/1673)
```
lgolding: >FEATURE [](start = 2, length = 7)

Now I'm confused. Michael successfully published 2.1.19 last week, and the release history in that NuGet package looks like this:
```
## **v2.1.19** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.1.19) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.1.19) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.1.19) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.1.19)
* API BREAKING: Remove helper method `SarifUtilities.DeserializeObject` introduced in 2.1.15 to fix [#1577](https://github.com/microsoft/sarif-sdk/issues/1577).
Now that an underlying bug in `PropertyBagConverter` has been fixed, there is no need to work around it with this helper method. `JsonConvert.DeserializeObject` works fine.
* FEATURE: Expanding Sarif SDK query mode to support Result.Uri, string StartsWith/EndsWith/Contains.
* FEATURE: Adding Result.Run and a populating method, so that methods which need the Run context
for a given Result have an integrated way to retrieve it.
```
That is, it contains the first three lines above (the API BREAKING and the first two FEATUREs). Please rework to release history to include that 2.1.19, and push the rest up to 2.1.20.

ScottLouvau: Will do.

src/Sarif/Baseline/V2/WhatComponent.cs
```diff
@@ -14,10 +14,10 @@ internal class WhatComponent : IEquatable<WhatComponent>
 
         public WhatComponent(string category, string propertySet, string propertyName, string propertyValue)
         {
-            this.Category = category;
-            this.PropertySet = propertySet;
-            this.PropertyName = propertyName;
-            this.PropertyValue = propertyValue;
+            this.Category = category ?? "";
```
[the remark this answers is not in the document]
ScottLouvau: Yes. I prefer "" in the null coalescing expressions because it's drastically shorter and then you can usually read what it is doing with the expression.

I still don't actually know why we recommended string.Empty. I'm pretty sure "" will continue to mean string.Empty forever.

PR #1708, opened 2019-10-23
Result filtering, part 1

lgolding:
In this PR we partially address the second of Scott's four log partitioning scenarios:

```
// Build a filtered log including all results the predicate returns ‘true’ for.
// Should enumerate the Results only once.
SarifLog Filter(SarifLog log, Func<Result, ResultIndex, bool> includePredicate);
```

I chose to do this one first because it seemed an easy one to start with.

This is a partial implementation, demonstrating that I can filter `run.artifacts` along with `run.results`. Still TODO:

- Enhance the test file to demonstrate how the `FilteringVisitor` picks up the other `result`-scoped `artifactLocations` (for example, in `result.relatedLocations`).
- Enhance the test file to demonstrate how the visitor picks up `run`-scoped `artifactLocations` (for example, in `run.attachments[].location`).

In email @michaelcfanning suggested doing the `run`-scoped properties separately, but the single call to `FilteringVisitor.VisitArtifactLocation` works for everything, so no additional product code is needed to handle those properties.

In Scott's proposed API, the predicate took a `ResultIndex` (run index/result index pair) in addition to the `Result` itself. I don't see why that's necessary, and I asked Scott about that in email.

Also differing from Scott's proposal, I defined a delegate type `FilteringVisitor.FilteringPredicate` rather than using a `Func`. I'm not sure I like it, partly because I can't decide whether that delegate type belongs in the `FilteringVisitor` (which means that callers of the wrapper function `SarifPartitioner.Filter` have to know that the `FilteringVisitor` exists, which breaks encapsulation) or in `SarifPartitioner` (which means that `FilteringVisitor` has to know that the wrapper `SarifPartitioner` exists, which seems inside-out).

Finally, I wanted to use our `FileDiffingFunctionalTest` infrastructure, but it didn't meet my needs because I need to run the filtering algorithm over the same file repeatedly, but with different filtering predicates. There was no way to specify a parameter when overriding `FileDiffingUnitTests.ConstructTestOutputFromInputResource`. So I changed the signature to accept a parameter; then I had to modify the other classes derived from `FileDiffingUnitTests` to declare the parameter (even though they don't use it).


src/Test.FunctionalTests.Sarif/TestData/Partitioning/FilterByPredicate/Inputs/FilterByPredicateWithArtifacts.sarif
```diff
@@ -0,0 +1,113 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.4.json",
```
lgolding: >$schema [](start = 3, length = 7)

Despite this file's name (FilterByPredicateWithArtifacts), I will _not_ create new files when I implement filtering of logical locations, code flows, etc. I'll change the name of the file to FilterByPredicateWithRunLevelProperties and use it for all of them.

Actually maybe better is to get rid of the test file that has _not_ run-level properties, and give its name (FilterByPredicate) to this file.

lgolding:

lgolding:

src/Sarif/Visitors/FilteringVisitor.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// <summary>
+    /// A visitor that filters a specified SARIF log to create a new log containing only those
+    /// results for which a specified predicate returns true, and only those elements of run-level
+    /// collections such as Run.Artifacts that are relevant to the filtered results.
+    /// </summary>
+    public class FilteringVisitor : SarifRewritingVisitor
+    {
+        public delegate bool FilteringPredicate(Result result);
+
+        private readonly FilteringPredicate predicate;
```
ScottLouvau: Maybe rename to "IncludeResult" so the desired return value to include a result is clearer?

src/Sarif/Visitors/FilteringVisitor.cs
```diff
@@ -0,0 +1,80 @@
...
+            return base.VisitResult(node);
+        }
+
+        public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
+        {
+            if (node.Index >= 0)
+            {
+                if (!remappedArtifactIndexDictionary.ContainsKey(node.Index))
+                {
+                    remappedArtifactIndexDictionary.Add(node.Index, filteredArtifacts.Count);
```
ScottLouvau: Maybe add a line declaring the newIndex so it's a bit easier to read:
int newArtifactIndex = filteredArtifacts.Count;
remappedArtifactIndexDictionary.Add(node.Index, newArtifactIndex);
...

src/Sarif/Visitors/FilteringVisitor.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            currentRun = node;
+
+            filteredResults = new List<Result>();
+            filteredArtifacts = new List<Artifact>();
+            remappedArtifactIndexDictionary = new Dictionary<int, int>();
+
+            Run visitedRun = base.VisitRun(node);
+
+            visitedRun.Results = filteredResults;
```
ScottLouvau: Do we know that base.VisitRun will consistently return a new Run instance, so we know this won't alter the input instance?

src/Test.FunctionalTests.Sarif/Partitioning/FilterByPredicateTests.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
```
ScottLouvau: Sort + Remove usings

src/Test.FunctionalTests.Sarif/TestData/Partitioning/FilterByPredicate/Inputs/FilterByPredicateWithArtifacts.sarif
```diff
@@ -0,0 +1,113 @@
...
+              "physicalLocation": {
+                "artifactLocation": {
+                  "index": 3
+                }
+              }
+            }
+          ]
+        }
+      ],
+      "columnKind": "utf16CodeUnits"
```
ScottLouvau: Maybe include a second Result which uses the same index, but will not match the predicate.
Include a third Result with the same index which will be included, so you see it doesn't get a second copy of the Artifact added.
Include a fourth Result which will be included and has the Artifact Uri inline (and doesn't use ArtifactLocation.Index).

src/Test.FunctionalTests.Sarif/Partitioning/FilterByPredicateTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using Newtonsoft.Json;
+using Microsoft.CodeAnalysis.Test.Utilities.Sarif;
+using System;
+using Microsoft.CodeAnalysis.Sarif.Visitors;
+
+namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests.Partitioning
+{
+    public class FilterByPredicateTests : FileDiffingFunctionalTests
+    {
+        public FilterByPredicateTests(ITestOutputHelper outputHelper, bool testProducesSarifCurrentVersion = true) :
```
ScottLouvau: This might actually be a good case to not use the file diffing approach.
It seems like the verification criteria would be simple - how many Results left, did they all match the criteria, did only a few Artifacts come through.

A sample file for the inputs (to allow a lot of detail in them) makes sense, but since the code to verify them should be very short, it might be nicer to have that code then to introduce another dozen or two examples which might have to be updated if we change the format or output behavior in some way later.

I'm thinking selfishly about that 200+ file update when I changed SarifLogger. =)

src/Sarif/Visitors/FilteringVisitor.cs
lgolding: Ah, you're saying that somebody might think this predicate should return true if you want to _filter out_ the result. Ok, I'll rename it to `IncludedResultPredicate`.

---
In reply to: [338198865](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338198865) [](ancestors = 338198865)

src/Sarif/Visitors/FilteringVisitor.cs
lgolding: Good idea.

---
In reply to: [338199939](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338199939) [](ancestors = 338199939)

src/Sarif/Visitors/FilteringVisitor.cs
lgolding: To the contrary, the semantics of `SarifRewritingVisitor` are that it alters the input object in place. That's why `SarifPartitioner.Filter` does a `DeepClone` of the input log before turning the visitor loose on it.

---
In reply to: [338200394](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338200394) [](ancestors = 338200394)

src/Test.FunctionalTests.Sarif/Partitioning/FilterByPredicateTests.cs
lgolding: Fixed.

---
In reply to: [338200662](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338200662) [](ancestors = 338200662)

src/Test.FunctionalTests.Sarif/TestData/Partitioning/FilterByPredicate/Inputs/FilterByPredicateWithArtifacts.sarif
lgolding: All good test cases, done.

---
In reply to: [338202341](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338202341) [](ancestors = 338202341)

src/Test.FunctionalTests.Sarif/Partitioning/FilterByPredicateTests.cs
lgolding: I understand your concern (which Michael shares) about having many granular test files. OTOH I like the file diffing approach because it's comprehensive.

Take a look at my comment in the test data file. I do _not_ intend to have (for example) one test file for artifacts, one for logical locations, one for code flows... Instead, I'll have one comprehensive test file. I think that addresses your concern.

---
In reply to: [338203971](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r338203971) [](ancestors = 338203971)

src/Test.FunctionalTests.Sarif/TestData/Partitioning/FilterByPredicate/Inputs/FilterByPredicateWithArtifacts.sarif
lgolding: Done.

---
In reply to: [337815072](https://github.com/microsoft/sarif-sdk/pull/1708#discussion_r337815072) [](ancestors = 337815072)

PR #1709, opened 2019-10-23
Always assign Guids on Baseline

ScottLouvau:
Amend a previous change to ensure a Guid and CorrelationGuid are *always* assigned to all Baselined results.

This fixes the case where baselining has happened with an older SARIF SDK and so an identity wasn't previously assigned. It now consistently assigns a Guid as well, if the producing tool didn't. This ensures a CorrelationGuid will always be available to identify the Results we're mapping to each other over time and also to provide an identity to associate any outside-of-SARIF metadata with the Result.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -8,7 +8,7 @@
 * BUGFIX: Result Matching now omits previously Absent results.
 * BUGFIX: Result Matching properly compares results from the same RuleID when multiple Rules match the same source line.
 * BUGFIX: Result Matching works when a result moves and has the line number in the message.
-* BUGFIX: Result Matching assigned Guids and uses Result.Guid as default CorrelationGuid in matched Results.
+* BUGFIX: Result Matching always assigns Result.CorrelationGuid and Result.Guid.
 * BUGFIX: Null hardening in Result Matching
 * BUGFIX: Console logger now outputs file location, if available, when writing notifications.
 
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -8,7 +8,7 @@
 * BUGFIX: Result Matching now omits previously Absent results.
 * BUGFIX: Result Matching properly compares results from the same RuleID when multiple Rules match the same source line.
 * BUGFIX: Result Matching works when a result moves and has the line number in the message.
-* BUGFIX: Result Matching uses Result.Guid as default CorrelationGuid in matched Results.
+* BUGFIX: Result Matching assigned Guids and uses Result.Guid as default CorrelationGuid in matched Results.
 * BUGFIX: Null hardening in Result Matching
 * BUGFIX: Console logger now outputs file location, if available, when writing notifications.
 
--- a/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
+++ b/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
@@ -74,10 +74,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
             Dictionary<string, object> resultMatchingProperties,
             out Dictionary<string, object> originalResultMatchingProperties)
         {
-            Result result = PreviousResult.Result.DeepClone();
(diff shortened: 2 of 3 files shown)


src/ReleaseHistory.md
```diff
@@ -8,7 +8,7 @@
 * BUGFIX: Result Matching now omits previously Absent results.
 * BUGFIX: Result Matching properly compares results from the same RuleID when multiple Rules match the same source line.
 * BUGFIX: Result Matching works when a result moves and has the line number in the message.
-* BUGFIX: Result Matching uses Result.Guid as default CorrelationGuid in matched Results.
+* BUGFIX: Result Matching assigned Guids and uses Result.Guid as default CorrelationGuid in matched Results.
```
lgolding: >ed [](start = 32, length = 2)

"assigned" => "assigns"?

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -75,6 +75,8 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
             out Dictionary<string, object> originalResultMatchingProperties)
         {
             Result result = PreviousResult.Result.DeepClone();
+            result.Guid = result.Guid ?? Guid.NewGuid().ToString(SarifConstants.GuidFormat);
+            result.CorrelationGuid = result.CorrelationGuid ?? result.Guid;
             result.BaselineState = BaselineState.Absent;
```
michaelcfanning: All of this code could be refactored into a helper, which is passed a result and baseline state, which clones the input results, sets the guids, sets the baseline state and returns the result. This helper could be used for all three places you changed in this file. Just a suggestion.

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -237,6 +237,7 @@ public void V2ResultMatcher_WhenNewIssueIsSuppressed_SupressesTheResultInTheOutp
             Result newSuppressedResult = matchedRun.Results.Single(r => r.Message.Text == "New suppressed result.");
 
             newSuppressedResult.Suppressions.Count.Should().Be(1);
+            AssertBaselinedRunInvariants(matchedRun);
```
lgolding: >AssertBaselinedRunInvariants [](start = 12, length = 28)

Terminology: are you using "baselined run" and "matched run" to mean the same thing? If so, then consider (on the "Ubiquitous Vocabulary" principle) using one or the other.

src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -403,16 +407,32 @@ public void ResultMatchingBaseliner_WhenThereIsOnlyOneCurrentRun_CopiesSelectedR
...
+
+        private void AssertBaselinedRunInvariants(Run baselinedRun)
+        {
+            // Ensure all results always have a BaselineState, CorrelationGuid, and Guid assigned after baselining.
+            // The CorrelationGuid provides a stable identity to identify a Result matching across Runs.
+            // The Guid should be the default CorrelationGuid, so that matched Results have an identity mappable to the first occurrence of the Result (even before it was baselined)
+            // The Guid also needs to be defined to provided an identity to attach other data (like Annotations) to the Result consistently.
+            foreach(Result result in baselinedRun.Results ?? Enumerable.Empty<Result>())
+            {
+                Assert.NotNull(result.Guid);
```
lgolding: >Assert [](start = 16, length = 6)

Use FluentAssertions:
```
result.Guid.Should().NotBeNull();
result.CorrelationGuid().Should().NotBeNull();
result.BaselineState.Should().NotBe(BaselineState.None);
```

src/ReleaseHistory.md
ScottLouvau: Fixed

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
ScottLouvau: Good idea. Fixed.

PR #1712, opened 2019-10-25
Remove unused "Previous" properties from build.props

lgolding:
NOTE: Ignore the title of the single commit below. I had switched from another branch and neglected to update the commit text in VS.

### DIFF
--- a/src/build.props
+++ b/src/build.props
@@ -11,7 +11,7 @@
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF SDK</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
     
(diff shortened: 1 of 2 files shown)


lgolding:

lgolding:

lgolding:

src/build.props
```diff
@@ -11,7 +11,7 @@
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF SDK</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
     
-    <!-- VersionPrefix denotes the current Semantic Version for the SDK. Must be updated before every nuget drop. -->
```
lgolding: >Semantic Version [](start = 43, length = 16)

While I'm in here: We can't claim SemVer compatibility because we make breaking changes in patch-level versions.

PR #1719, opened 2019-11-14
Implement log partitioning

lgolding:
This new partitioning API is intended for use in filing work items.

Since I have logic to avoid excessive cloning (you can see this by looking at where `deepClone` is used), we might not need the `deepClone` flag at all: it might be ok to deep clone always.

Note:
- Testing this feature required an enhancement to `FileDiffingUnitTests` so a single test can produce multiple output files. (A single test can now also consume multiple input files.)


src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+        {
+            SarifLog partitionLog;
+            if (deepClone)
+            {
+                // Save time and space by not cloning the runs unless and until necessary. We will
+                // only need to clone the runs that have results in this partition.
+                IList<Run> originalRuns = originalLog.Runs;
+                originalLog.Runs = null;
+
+                partitionLog = originalLog.DeepClone();
```
ScottLouvau: Is this actually different than new SarifLog(originalLog)? Are there any properties outside the Runs which need DeepCloning?

src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+                    // certain run-level collections such as Run.Artifacts, have been replaced.
+                    Run originalRun = originalLog.Runs[iOriginalRun];
+                    Run partitionRun;
+                    if (deepClone)
+                    {
+                        // Save time and space by only cloning the necessary results and associated
+                        // collection elements. We already cloned the relevant results in VisitResult.
+                        IList <Result> originalResults = originalRun.Results;
+                        IList<Artifact> originalArtifacts = originalRun.Artifacts;
+                        originalRun.Results = null;
```
ScottLouvau: I think a function wrapping the "DeepCloneExceptXAndY" functionality would help. It took me a moment to figure out that's what you're doing (storing, clearing, cloning, and then resetting the references you don't want cloned)

src/Test.FunctionalTests.Sarif/Partitioning/PartitionTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            Partition_ByRuleId_ProducesOneLogFilePerRule(deepClone: true);
+        }
+
+        [Fact]
+        public void Partition_ByRuleId_WithShallowCopy_ProducesOneLogFilePerRule()
+        {
+            Partition_ByRuleId_ProducesOneLogFilePerRule(deepClone: false);
+        }
+
```
ScottLouvau: If you have a sample with more than one run, you should check that Runs are excluded altogether if none of the Run Results is in a given partition. (If not already covered by the above)

src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+        /// result belongs to, or null if the result should be discarded (not placed in any
+        /// partition).
+        /// </summary>
+        /// <param name="result">
+        /// The result to be assigned to a partition.
+        /// </param>
+        /// <typeparam name="T">
+        /// The type of the object returned from the partition function. It must be a reference
+        /// type so that null is a valid value. It must override bool Equals(T other) so that
+        /// two Ts can compare equal even if they are not reference equal.
```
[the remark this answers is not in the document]
lgolding: Interesting point. According to [StackOverflow](https://stackoverflow.com/questions/15889987/when-to-specify-constraint-t-iequatablet-even-though-it-is-not-strictly-re), it doesn't actually add anything because all types are convertible to `System.Object`, which implements `IEquatable<T>`.

**BUT** The poster goes on to argue that it makes for good documentation, telling the user that the objects are going to be compared, so perhaps he should give some thought to how the comparison would work.

On the other other hand, my comment already does that.

TL;DR: I'll add it.

---
In reply to: [357427829](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357427829) [](ancestors = 357427829)

src/Sarif/Visitors/PartitioningVisitor.cs
lgolding: Yes, `sarifLog.inlineExternalProperties`.

---
In reply to: [357431410](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357431410) [](ancestors = 357431410)

src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+
+        public override Result VisitResult(Result node)
+        {
+            inResult = true;
+            currentPartitionValue = partitionFunction(node);
+            if (currentPartitionValue != null)
+            {
+                if (!partitionValues.Contains(currentPartitionValue))
+                {
+                    partitionValues.Add(currentPartitionValue);
```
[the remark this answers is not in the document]
lgolding: Yes, thanks. I did that elsewhere, missed it here.

---
In reply to: [357429173](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357429173) [](ancestors = 357429173)

src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+        /// <summary>
+        /// Returns a mapping from partition value to the partition log for that value.
+        /// </summary>
+        /// <returns></returns>
+        public Dictionary<T, SarifLog> GetPartitionLogs()
+        {
+            if (originalLog == null)
+            {
+                throw new InvalidOperationException(
+                    "You must call " + nameof(VisitSarifLog) +
```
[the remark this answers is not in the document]
lgolding: Done.

---
In reply to: [357428813](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357428813) [](ancestors = 357428813)

src/Sarif/Visitors/PartitioningVisitor.cs
lgolding: It's a nice idea but the function name (which would now be `DeepCloneExceptResultsAndArtifacts`) will become unwieldy once it's `DeepCloneExceptResultArtifactsGraphsLogicalLocationsThreadFlowLocationsWebRequestsAndWebResponses` :-)

---
In reply to: [357431653](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357431653) [](ancestors = 357431653)

src/Test.FunctionalTests.Sarif/Partitioning/PartitionTests.cs
lgolding: The test cases do already cover that. The input file Partition.sarif has two runs. Note that the output file TST0002.sarif contains only one run (corresponding to the second of the two runs in the input file), because only that run has any results with `ruleId` `TST0002`.

---
In reply to: [357432826](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357432826) [](ancestors = 357432826)

src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -0,0 +1,454 @@
...
+        // True if we are currently processing a result, otherwise false.
+        private bool inResult;
+
+        // The set of results in each partition in each original run. That is,
+        //
+        //     partitionedResultDictionaries[runIndex][partitionValue]
+        //
+        // contains the list of results in the run at runIndex for whose partition value is
+        // partitionValue.
+        private List<Dictionary<T, List<Result>>> partitionResultDictionaries;
```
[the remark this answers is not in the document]
lgolding: It's a nice idea, it wouldn't be that hard, and I have a good set of test to backstop me. I'll do it.

---
In reply to: [357428561](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357428561) [](ancestors = 357428561)

src/Sarif/Visitors/PartitioningVisitor.cs
lgolding: Oh, I suppose `DeepCloneExceptRunLevelCollections` would be ok. Let me think about it.

---
In reply to: [357441324](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357441324) [](ancestors = 357441324,357431653)

lgolding: No, the comment already says we're "cloning only the necessary results and associated collection elements." I'm going to leave this.

---
In reply to: [357442600](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357442600) [](ancestors = 357442600,357441324,357431653)

src/Sarif/Visitors/PartitioningVisitor.cs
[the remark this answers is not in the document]
lgolding: Done. Really quite a small change, and a nice one.

---
In reply to: [357442354](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357442354) [](ancestors = 357442354,357428561)

src/Sarif/Visitors/PartitioningVisitor.cs
lgolding: (Even if there weren't any such properties, it's good future-proofing against the day when there might be some.)

---
In reply to: [357440439](https://github.com/microsoft/sarif-sdk/pull/1719#discussion_r357440439) [](ancestors = 357440439,357431410)

PR #1726, opened 2019-12-06
Baselining Fix, SarifToCsv.Properties

ScottLouvau:
- Add a Baselining phase to link the first and last results from the same Artifact Uri
- Add 'Properties' support to SarifToCsv.
- Fix NullRefException for a null serialized property (it end up with a null in the property dictionary, not an instance with a null SerializedValue)
- Fix NullRefException in WhereComparer.ArtifactUri when the ArtifactLocation containing the Uri passed to the function is null.

In Baselining, the new phase ensures that results match properly when the Results have moved (line changes in the file), no results have unique properties, and the Artifacts just before and after the changed Artifact are different.

This occurs when baselining runs between a full baseline and an incremental run, because the full baseline includes many more Artifacts than the incremental run does. We want to ensure baselining gets a "foothold" in files with unchanged Uris even if none of the results within them have unique, stable properties.


src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
```diff
@@ -114,5 +117,20 @@ public static Action<WriteContext> GetWriter(string columnName)
...
+
+        private static void WriteProperties(WriteContext c)
+        {
+            StringBuilder result = new StringBuilder();
+
+            foreach (string propertyName in c?.Result?.PropertyNames ?? Enumerable.Empty<string>())
+            {
+                if (result.Length > 0) { result.Append("; "); }
+                result.Append(propertyName);
+                result.Append(": ");
```
lgolding: >:  [](start = 31, length = 2)

Just curious why you chose `name: value` over `name=value`. Was it a coin toss or is there a reason `:` is better?

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -1,11 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching;
```
lgolding: Nit: I believe the SDK's convention is "System" namespaces first.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -194,6 +231,54 @@ private void LinkIfSimilar(int beforeIndex, int afterIndex)
             }
         }
 
+        private ExtractedResult FirstWithUri(ExtractedResult desiredUri, IList<ExtractedResult> set, ref int fromIndex)
+        {
+            // Find the first Result with a Uri *matching* the desired one, or null if there aren't any
+            for (; fromIndex < set.Count; ++fromIndex)
+            {
+                int whereCmp = WhereComparer.CompareFirstArtifactUri(set[fromIndex], desiredUri);
```
lgolding: >CompareFirstArtifactUri [](start = 45, length = 23)

I kind of "get" the decision to put this method in `WhereComparer` but I don't think it belongs there, based on the Single Responsibility Principle. `WhereComparer`'s sole responsibility is to decide if two `ExtractedResults` have the same location (for a particular definition of "same"). This method doesn't contribute to that responsibility. It _does_ contribute to this class's responsibility of matching two sets of results, so I think it belongs in this class.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -115,6 +115,43 @@ private void LinkResultsWithIdenticalWhere()
...
+                // Get the next After Result (the first for a given Uri)
+                ExtractedResult afterFirstForUri = After[afterIndex];
+
+                // Look for the first Before Result with the same Uri, if any
+                ExtractedResult beforeFirstForUri = FirstWithUri(afterFirstForUri, Before, ref beforeIndex);
+
+                // If there was one...
+                if (beforeFirstForUri != null)
+                {
+                    // ... Try to link the first Results together
```
lgolding: >together [](start = 57, length = 8)

Do we have to worry that the identified pair of before/after results might already be linked because the first phase found that they had identical "where"? Or even if we don't have to "worry" about it, could we improve perf by noticing that they are already linked and avoiding the calls to `LinkIfSimilar`?

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -194,6 +231,54 @@ private void LinkIfSimilar(int beforeIndex, int afterIndex)
             }
         }
 
+        private ExtractedResult FirstWithUri(ExtractedResult desiredUri, IList<ExtractedResult> set, ref int fromIndex)
+        {
+            // Find the first Result with a Uri *matching* the desired one, or null if there aren't any
+            for (; fromIndex < set.Count; ++fromIndex)
```
lgolding: >fromIndex [](start = 44, length = 9)

Do we have some N-squared action going on here? The caller is looping over the list same list. If so, can we avoid it by doing a pre-scan of the two lists, finding the first Uri in each element of each list, and making some dictionaries? Something like `Dictionary<Uri, int> firstBeforeUriIndexDictionary, lastBeforeUriIndexDictionary, firstAfterUriIndexDictionary, lastAfterUriIndexDictionary`?

lgolding:
Approved, with a couple of nits and a couple of design questions.

lgolding:
Nit: Ensure comment sentences end in periods (except that I do like the construct "We do this..." ... "... and then we do that." -- the ellipses are fine with me :-)).

src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
ScottLouvau: I'm reviewing these as a human and find them easier to "parse" visually when there's a space before each distinct property name and value.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: LinkIfSimilar ensures the results are not already linked. Better to have it there to ensure it happens no matter the caller. I suspect the runtime inlines the whole function, since it's quite short, so I don't think there will be any performance impact.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: No, but it's tricky. Added some comments to explain.

Basically, LinkFirstAndLastFromSameArtifact is similar to LinkResultsWithIdenticalWhere in that it walks the two lists together in a shared loop and always moves one or the other forward.

However, it's too complex for one method, so I keep two indices (beforeIndex and afterIndex) and pass them by ref to FirstWithUri and LastWithUri. Neither of them reset the indices, so they are always walking forward in the two sets.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: Fixed. Wish VS would insert them following our formatting guidelines. :/

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: I understand your point, but disagree with this one. I want to encapsulate all of the ExtractedResult where property related extraction and comparison to one place and want to be able to reuse it if some other class wants these comparisons. So, I guess I have a broader definition of that the responsibility of WhereComparer is. I'm thinking, for example, that SarifToCsv might reuse this logic later to get the first artifact uri out for two Results, though then I'd have to resolve the existing conflict between Result and ExtractedResult.

PR #1729, opened 2019-12-14
Fix #1728: Tool.CreateFromAssembly mishandles dottedQuadFileVersion

### LINKED ISSUE #1728: Tool.CreateFromAssembly mishandles dottedQuadFileVersion
If gives something like this:
```
"18.255.65000.0 built by: MY-MACHINE"
```
The spec requires only the four dot-separated integer components.

@michaelcfanning FYI

lgolding:


lgolding:

lgolding:

src/Test.UnitTests.Sarif/Core/ToolTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            public string Input { get; }
+            public string ExpectedOutput { get; }
+        }
+
+        private static ReadOnlyCollection<DottedQuadFileVersionTestCase> s_dottedQuadFileVersionTestCases =
+            new ReadOnlyCollection<DottedQuadFileVersionTestCase>(new DottedQuadFileVersionTestCase[]
+            {
+                new DottedQuadFileVersionTestCase(
```
ScottLouvau: Nit: When I want to test a number of scenarios, I really want them to be one line each so I can read them quickly. You might consider not using named parameters here and formatting each on a single line to read quickly. I also sometimes choose short method names to minimize the "not the test input/output" content on the line.

PR #1734, opened 2020-01-07
Baselining E2E Test Harness

ScottLouvau:
DRAFT PARTS: 
  Extension methods will be ported into Sarif-SDK if signatures and naming look good.
  ResultTextFormatter methods are not planned to be moved, as they're test debug specific.
  Baselining algorithm should support "BaselineFilteringMode": "ToIncludedFiles" internally.

OTHER:
This PR is adding a harness for Sarif Baseline Algorithm end-to-end testing.
It's a console app, because the tests will be far too slow for unit testing and the content is in a separate non-public repo.
The app need only be run when baselining behavior changes.
We can import SARIF logs as-is from different tools and environments to build a comprehensive test set.
The runs output a summary log which shows overall matching statistics, and a detail log which shows what happened to every Result. The detail logs are compared and the test will "pass" if matching behavior was completely identical.
A second set of detail logs can be generated from the first which include other columns from the Result for quick human debugging. This greatly helps in interpreting whether Results should've matched or not when looking at non-matches.
There's another mode to quickly debug matching for a specific Result in a specific Run which skips other work and stops the debugger at the ideal moment.
Each Result is assigned a "RID" (LogIndex + " " + ResultIndex) as the GUID, which gives them all stable canonical identities for the detail log. The RID is also the argument passed to debug mode to debug the result, making it "copy and paste to debug" simple to investigate single results.

Since the Expected folder logs are compared to the Output folder ones, new baselines can be captured simply by renaming the Output folder to Expected. Logs can be investigated by diffing the Output and Expected folders.


src/Sarif.Sdk.sln
```diff
@@ -81,7 +81,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.UnitTests.Sarif.Multit
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sarif.WorkItemFiling", "Sarif.WorkItemFiling\Sarif.WorkItemFiling.csproj", "{5D8A3D57-3593-40E2-A7DF-0B4E182543B6}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Test.UnitTests.Sarif.WorkItemFiling", "Test.UnitTests.Sarif.WorkItemFiling\Test.UnitTests.Sarif.WorkItemFiling.csproj", "{4162EF3A-B3B0-45AD-99DC-C833C73494C9}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Test.UnitTests.Sarif.WorkItemFiling", "Test.UnitTests.Sarif.WorkItemFiling\Test.UnitTests.Sarif.WorkItemFiling.csproj", "{4162EF3A-B3B0-45AD-99DC-C833C73494C9}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Test.EndToEnd.Baselining", "Test.EndToEnd.Baselining\Test.EndToEnd.Baselining.csproj", "{8D674F7B-F6D8-4508-A940-AA7D03D5DA36}"
```
lgolding: >EndToEnd [](start = 58, length = 8)

Note to self: Do we have a convention for distinguishing "functional" tests (_e.g._, `Test.FunctionalTests.Sarif`) from "end to end" tests? It might become clear as I read on.

lgolding: Oh, I see, it's a whole separate executable (`Program.cs`). Now I'm curious how and when we run such tests.

---
In reply to: [366488718](https://github.com/microsoft/sarif-sdk/pull/1734#discussion_r366488718) [](ancestors = 366488718)

lgolding:
> using System;

Nit: Copyright notice on all new files.

---
Refers to: src/Test.EndToEnd.Baselining/Program.cs:1 in 99b3e9b. [](commit_id = 99b3e9b6e0959778100d33ededb3b0d0b19e0b9c, deletion_comment = False)

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Diagnostics;
+using System.IO;
+using CommandLine;
+using Test.EndToEnd.Baselining.Options;
+
+namespace Test.EndToEnd.Baselining
+{
+    /// <summary>
+    ///  Test.EndToEnd.Baselining is a console app for conducting end-to-end testing of the baselining algorithm.
+    ///  It uses sample SARIF logs, some of which contain sensitive data, so the test content is not in this repository.
```
lgolding: >repository [](start = 109, length = 10)

IMO that means this whole project belongs in another, private, repo. Even the "Content Folder Structure" below reveals organizational and project details. We shouldn't put a project that only some people can use in a public repo.

src/Sarif.Sdk.sln
lgolding: ... and now the comment in `Program.cs` answers my question.

---
In reply to: [366489311](https://github.com/microsoft/sarif-sdk/pull/1734#discussion_r366489311) [](ancestors = 366489311,366488718)

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+    ///  ========================
+    ///  BaselineE2E\                      [Test Root Path]
+    ///    Input\                          [SARIF log series to test]
+    ///      CloudMine-Spam\               [Series can be organized in any folder structure; using Tool\Version\ContentType\Series here]
+    ///        v1.2.12\              
+    ///          WorkItem\
+    ///            mseng\                  [A folder with files in it is an individual series]
+    ///              Baseline.sarif        [A starting Baseline before any logs are loaded]
+    ///              20191203_1530.sarif   [Each other SARIF log will be baselined in order by name; use DateTime prefixes to order them]
+    ///              20191204_1135.sarif   [...and use the file name suffix to associate them with their source]
```
lgolding: >source [](start = 105, length = 6)

Just to help me out, what does the suffix `1135` mean in this example? Is it a build number. I understand that in general it might be anything.

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+    ///   
+    ///   Expected_Debug\                  [Each run will create an 'enriched' version of the Expected logs also]
+    ///     
+    /// 
+    ///  Execution Steps
+    ///  ===============
+    ///  Running tests will find each series folder in Input and:
+    ///    - Load the initial baseline
+    ///    - Load each log in order
+    ///    - Replace the GUID with a 'RID' (LogIndex + ' ' + ResultIndex)
```
lgolding: >GUID [](start = 25, length = 4)

I assume you are talking about `result.guid`. There are lots of GUID-valued property, so please specify the `objectName.propertyName` if you decide to keep this. But... strictly speaking, that makes the SARIF file invalid. I suggest that instead you add it to the property bag: `result.properties["RID"] = that-thing`.

src/Test.EndToEnd.Baselining/Options/DebugOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using CommandLine;
+
+namespace Test.EndToEnd.Baselining.Options
+{
+    [Verb("debug", HelpText = "Debug Baselining for a specific Series, Log, and Result")]
+    public class DebugOptions : OptionsBase
+    {
+        [Value(1, Required = false, HelpText = "Debug: Series Path (everything under 'Input/'")]
```
lgolding: >everything [](start = 68, length = 10)

Did you mean to write "default: everything under 'Input/'"?

src/Test.EndToEnd.Baselining/Options/DebugOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+        public string DebugSeriesPath { get; set; }
+
+        [Value(2, Required = false, HelpText = "Debug: Log Index")]
+        public int DebugLogIndex { get; set; }
```
lgolding: Since values 2 and 3 are not required, we need to document their defaults as well.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            foreach (Result result in run.EnumerateResults())
+            {
+                foreach (Uri uri in AllArtifactUris(result))
+                {
+                    yield return uri;
+                }
+            }
+        }
```
lgolding: >} [](start = 8, length = 1)

The problem with this approach to enumeration is that `artifactLocation` objects can occur in many `run` properties other than `result` -- and they can occur in several `result` properties as well. Below you will find a list -- which I don't guarantee to be exhaustive! -- based on searching the spec for the string "artifactLocation object".

IMO you are better off creating a visitor based on `SarifRewritingVisitor`, something like `ArtifactLocationFindingVisitor`, that accumulates all the `artifactLocation` objects it encounters, and then at the end you can run over them:

```
var visitor = new ArtifactLocationFindingVisitor();
visitor.Visit(sarifLog) // or run, or whatever you're interested in.
List<Uri> artifactUris = visitor.ArtifactLocations
    .Where(a => a.Uri != null)
    .Select(a => a.Uri)
    .Distinct()
    .ToList();
```

All of these property paths are relative to a `run` object

- `attachments[].location`
- `conversion.analysisToolLogFiles`
- `conversion.tool.driver.locations`
- `conversion.tool.extensions[].locations`
- `externalPropertyFileReferences[].location`
- `invocations[].responseFiles`
- `invocations[].executableLocation`
- `invocations[].workingDirectory`
- `invocations[].stdin, stdout, stderr, stdoutStderr`
- `originalUriBaseIds[]`
- `policies[].locations`
- `specialLocations[].displayBase`
- `stacks[].frames[].location.physicalLocation.artifactLocation`
- `taxonomies[].locations`
- `threadFlowLocations[].location.physicalLocation.artifactLocation`
- `tool.driver.locations`
- `tool.extensions[].locations`
- `translations[].locations`
- `result.attachments[].location`
- `versionControlProvenance[].mappedTo`

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
```diff
@@ -0,0 +1,92 @@
...
+        public static IEnumerable<Uri> AllArtifactUris(this Result result)
+        {
+            if (result?.Locations != null)
+            {
+                foreach (Location location in result.Locations)
+                {
+                    Uri uri = location?.PhysicalLocation?.ArtifactLocation?.FileUri(result.Run);
+                    if (uri != null) { yield return uri; }
+                }
+            }
```
lgolding: >} [](start = 12, length = 1)

To emphasize my earlier point: `result` also has `codeFlows` and `stacks` properties (probably others) which ultimately contain `artifactLocation` objects.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
lgolding: Of course if the semantics you want is "all result location URIs" then this is fine -- but then you need to name the methods to make clear what they're providing. It's not "all artifact URIs in the result".

---
In reply to: [366541420](https://github.com/microsoft/sarif-sdk/pull/1734#discussion_r366541420) [](ancestors = 366541420)

src/Test.EndToEnd.Baselining/Extensions/IEnumerableStringExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        ///  Filter a set to elements which contain 'requiredPart'. If 'requiredPart' is null or empty,
+        ///  this function returns the full set as-is.
+        /// </summary>
+        /// <param name="set">IEnumerable of string to filter</param>
+        /// <param name="requiredPart">Value which must be within items to return, or null/empty for no filtering</param>
+        /// <returns>Strings in set which contain requiredPart, or full set if no requiredPart provided</returns>
+        public static IEnumerable<string> WhereContainsPart(this IEnumerable<string> set, string requiredPart)
+        {
+            if (set == null) { throw new ArgumentNullException("set"); }
```
lgolding: >"set" [](start = 63, length = 5)

Nit: here and wherever else: `nameof(set)`

src/Test.EndToEnd.Baselining/Extensions/IEnumerableStringExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///  Filter an IEnumerable of string by the file extensions found on the files.
+        /// </summary>
+        /// <param name="set">Set of paths to filter</param>
+        /// <param name="fileExtension">File extension to match</param>
+        /// <returns>Items in the set which have the desired file extension</returns>
+        public static IEnumerable<string> WhereExtensionIs(this IEnumerable<string> set, string fileExtension)
+        {
+            fileExtension = "." + fileExtension.TrimStart('.');
+
+            return set.Where(path => Path.GetExtension(path).Equals(fileExtension, StringComparison.OrdinalIgnoreCase));
```
lgolding: >OrdinalIgnoreCase [](start = 100, length = 17)

This is a problem on Linux. The SDK does not systematically compare paths in a file-system-appropriate way, so I'm sure this problem occurs elsewhere. Let's file a bug on this.

src/Test.EndToEnd.Baselining/Extensions/SarifLogExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    public static class SarifLogExtensions
+    {
+        public const string BaselineFilteringPropertyName = "BaselineFiltering";
+
+        public static BaselineFilteringMode GetBaselineFilteringMode(this SarifLog log)
+        {
+            if (log == null) { throw new ArgumentNullException("log"); }
+
+            BaselineFilteringMode result;
```
lgolding: >result [](start = 34, length = 6)

Nit: call this `mode` because `result` is so heavily used in SARIF to mean something else.

src/Test.EndToEnd.Baselining/Extensions/SarifLogExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+            log.SetProperty<string>(SarifLogExtensions.BaselineFilteringPropertyName, mode.ToString());
+        }
+
+        public static IEnumerable<Run> EnumerateRuns(this SarifLog log)
+        {
+            if (log?.Runs != null)
+            {
+                foreach (Run run in log.Runs)
+                {
+                    yield return run;
```
lgolding: >run [](start = 33, length = 3)

In contrast to the novel I wrote about enumerating `artifactLocations`s, this is fine -- this is the only place you'll find a `run`.

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+            catch (Exception ex) when (!Debugger.IsAttached)
+            {
+                Console.WriteLine(ex.ToString());
+                return 2;
+            }
+        }
+
+        static int Run(RunOptions options)
+        {
+            BaseliningTester tester = new BaseliningTester();
```
lgolding: >BaseliningTester [](start = 12, length = 16)

Nit: `var` ok if type is explicit on RHS.

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+            BaseliningSummary overallSummary = tester.RunAll(options.TestRootPath);
+            Console.WriteLine($"{overallSummary}");
```
lgolding: >$"{overallSummary}" [](start = 30, length = 19)

Nit: Isn't this the same as `Console.WriteLine(overallSummary)`?

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            return 0;
+        }
+
+        static int Debug(DebugOptions options)
+        {
+            BaseliningTester tester = new BaseliningTester();
+            tester.RunSeries(Path.Combine(options.TestRootPath, BaseliningTester.InputFolderName, options.DebugSeriesPath), options.DebugLogIndex, options.DebugResultIndex);
```
lgolding: >Path.Combine(options.TestRootPath, BaseliningTester.InputFolderName, options.DebugSeriesPath [](start = 29, length = 92)

Design nit: IMO the `Program` doesn't deserve to understand what to do with the options -- in this case, combining them with information that actually belongs to the `BaseliningTester` to form a path. Why not just pass the whole options object to the `RunXxx` method (here and in the other two similar places), and let the tester deal with it? It has the additional virtue of not requiring you to modify the `RunXxx` method's signature to change when you add a new option.

src/Test.EndToEnd.Baselining/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            return 0;
+        }
+
+        static int CreateDebugLogs(RebuildDebugLogsOptions options)
+        {
+            BaseliningTester tester = new BaseliningTester();
+            tester.EnrichUnder(Path.Combine(options.TestRootPath, BaseliningTester.InputFolderName));
```
lgolding: >Combine [](start = 36, length = 7)

Just to emphasize the previous comment: here again, `Program` needs to know more about how the tester works than it should.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace SarifBaseline.Extensions
+{
+    public static class ArtifactUriEnumeration
+    {
+        public static IEnumerable<Uri> AllArtifactUris(this SarifLog log)
+        {
+            if (log != null)
+            {
+                foreach (Run run in log.EnumerateRuns())
```
michaelcfanning: >EnumerateRuns [](start = 40, length = 13)

I'm not convinced of the value adding an API to avoid a single null check makes sense. Better would be for the emitted SARIF code to have an accessor that auto-populated log.Runs to a non-null but empty collection for this case. We would want, naturally, deserialization code that elides non-null but empty things (we have this elsewhere).

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    foreach (Uri uri in run.AllArtifactUris())
```
michaelcfanning: >AllArtifactUris [](start = 44, length = 15)

We might consider the GetXXX name convention for these cases. e.g., GetAllArtifactUris.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
michaelcfanning: Larry's suggestion, of course, will collect all these things in memory at once, which might not be ideal. I like the visitor pattern but am not sure it's the right approach (that would depend on consumer use case). I do agree with Larry's core point that, you need to cover other url holding things. It does raise the point, why was this code authored to begin with? i.e., what was the consumer doing with the information? are consumers likely to only want the unique urls or all of them? also, do they want reconstructed URLs if there is a original uri base?


---
In reply to: [366542388](https://github.com/microsoft/sarif-sdk/pull/1734#discussion_r366542388) [](ancestors = 366542388,366541420)

src/Test.EndToEnd.Baselining/Extensions/IEnumerableStringExtensions.cs
michaelcfanning: Ouch. Any setting here needs to be global in the SDK that can flow to all file-system relevant things (which are numerous). i.e., we should expect that the common case here is that any session in which the SDK is invoked is focused on processing files for one kind of file system (case-sensitive) or the other.

---
In reply to: [366544727](https://github.com/microsoft/sarif-sdk/pull/1734#discussion_r366544727) [](ancestors = 366544727)

michaelcfanning:
> using CommandLine;

NIT: you haven't added the copyright header to any files in this change. intentional?

---
Refers to: src/Test.EndToEnd.Baselining/Options/RunOptions.cs:1 in 99b3e9b. [](commit_id = 99b3e9b6e0959778100d33ededb3b0d0b19e0b9c, deletion_comment = False)

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                }
+            }
+        }
+
+        public static IEnumerable<Uri> AllArtifactUris(this Run run)
+        {
+            if (run?.Artifacts != null)
+            {
+                foreach (Artifact artifact in run.Artifacts)
```
michaelcfanning: >Artifacts [](start = 50, length = 9)

here, of course, you haven't provided run.EnumerateArtifacts. Again, I think our auto-generated code could be helpful. i.e., emit property accessors as above. This would have the benefit of making this value consistent through the OM, rather than for the handful of items we've provided extensions for. Note as well that the code emit machinery could itself be used to autogenerated a consistent set of extension methods, just throwing that out in case it provokes interesting thinking.

src/Test.EndToEnd.Baselining/Extensions/IEnumerableStringExtensions.cs
lgolding: I filed #1736, "File path comparisons are not file-system-appropriate".

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
lgolding: We shouldn't auto-populate `log.runs` because `null` is semantically different from empty for this property. See §3.13.4.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: @michaelcfanning, what do you think about this? I made a separate repo for the test content, and could move the harness there. The harness can also be used with non-private content, if it's here, and it's a bit more discoverable to people who might change the Baselining later (so they'd go looking for the private repo)

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: In this case, it's all DateTime (yyyyMMdd_HHmm). In general the file names need to sort in order, and then can have any suffix to help map them to the source. This source is Cosmos crawls, which are keyed by DateTime, so the DateTime to the minute is all we need to find them.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
ScottLouvau: Hmm. My usage needs all the Locations in all the Results, so I'll go with the rename the method option. =)

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
ScottLouvau: Basically, I constantly ended up writing:

if(log?.Runs != null)
{
      foreach(Run run in log.Runs)
      {
            if(run?.Results != null)
            {
                  foreach(Result result in run.Results)
                  {
                         if(result?.Locations != null)
                         {
                                  ...
}}}}}

Basically, each one of these methods merges a null check and foreach loop into one thing. I can then build each extension method on the one above (for EnumerateResults(SarifLog)) and a user can start out anywhere in the SarifLog hierarchy without double statements everywhere.

I agree this method doesn't do a lot, but I think it's useful in making enumeration code simpler and consistent for enumerating from any level to any level in the Sarif hierarchy. That's the overall point.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: I *think* might wanted the harness here and just the content elsewhere. This means anyone can run with public test content (once we have some). I can move it to the content repo later if that's what was wanted.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: Made comment Result.Guid.

I know I'm misusing Result.Guid. I wanted a simple, readable, stable number to make debugging easy. I need to use the Guid property because that's the only one the Baselining algorithm keeps from both the "baseline" and "current" copies of the result, so I can see what matched up.

src/Test.EndToEnd.Baselining/Options/DebugOptions.cs
ScottLouvau: Fixed

src/Test.EndToEnd.Baselining/Options/DebugOptions.cs
ScottLouvau: No; fixed to make required, fixed to explain the arg is the portion of the series path after Input\

src/Test.EndToEnd.Baselining/Options/DebugOptions.cs
ScottLouvau: Fixed.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
ScottLouvau: I won't move these to the SDK. 

Further, I've renamed the ones above Result scope to be "AllResultArtifactUris" and made the Result one include Result.Attachments[].Location, to be complete.

I think technically my scenario doesn't care if it's everything in the Run or just the Result.Locations Uris, so once a variation of this in the SDK is added, I should be able to use it.

Larry, I like the Visitor accumulator idea; that seems like a nice, consistent way to get them all. I'm going to leave my code as-is since it's done and won't be ported to a broader location, however.

src/Test.EndToEnd.Baselining/Extensions/SarifLogExtensions.cs
ScottLouvau: Normally I try to use the convention that a variable which will be returned by a function is called result (to tell a reader which one to pay most attention to) but in this case I agree about the conflict with big-R Result.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: Hmm. My style preference is "no var"; going to leave as-is. Maybe the SDK contributors should all get together to negotiate about var and other style concerns at some point.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: Yes. This is what evolves when BaseliningSummary.ToString didn't always exist. =) Fixed.

src/Test.EndToEnd.Baselining/Program.cs
ScottLouvau: I understand your point. I wanted the classes to take independent inputs (full paths which don't require any adjustment) but I don't want the human to have to type out full paths repeatedly. This makes my Options class and Program the "frontier" between the two goals, so I translate here.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
ScottLouvau: Yep - I'll leave that for the "what will the SDK version look like?" design.

src/Test.EndToEnd.Baselining/Extensions/ArtifactUriEnumeration.cs
ScottLouvau: I think Larry's idea of leveraging the Visitor to accumulate these seems like a good way for the SDK version of this to go. It's very consistent and only requires that the Visitor knows how to walk everything safely. 

To spew a response to another comment here, I also don't think memory use with a Visitor approach will be a problem. One million pointers in a List<Uri> is 8MB. The problem is when each of the things in the list is a new object and they all have 10 pointers in them. Then you're at 92MB.

PR #1738, opened 2020-01-28
Fix #1737: SarifLogBaseliner does not populate firstDetectionTimeUtc

### LINKED ISSUE #1737: SarifLogBaseliner does not populate firstDetectionTimeUtc

lgolding:
Previously, the code just asked whether or not a result from the current log matched a result in the baseline. Now, we actually extract the matching result from the baseline so we can roll its first detection time forward. If the matching result has no first detection time, we fall back to the time that the run was performed. And if we can't find that, we fall back to "now" as the first detection time.

**NOTE** The logic embodies a couple of choices about which reasonable people might disagree:
- I choose "run end time" over "run start time" when both are available.
- I use "now" as the ultimate fallback if there's absolutely no other time information available.

Feel free to disagree.

### DIFF
--- a/src/Sarif/Baseline/SarifLogBaseliner.cs
+++ b/src/Sarif/Baseline/SarifLogBaseliner.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -15,22 +16,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline
             ResultComparator = comparator;
         }
 
-        public Run CreateBaselinedRun(Run baseLine, Run nextLog)
+        public Run CreateBaselinedRun(Run baseline, Run nextLog)
         {
             Run differencedRun = nextLog.DeepClone();
             differencedRun.Results = new List<Result>();
-            
+
+            DateTime now = DateTime.Now;
+            DateTime baselineRunTime = GetRunTime(baseline, now);
+            DateTime newRunTime = GetRunTime(nextLog, now);
+
             foreach (var result in nextLog.Results)
             {
                 Result newResult = result.DeepClone();
 
-                newResult.BaselineState = 
-                    baseLine.Results.Contains(result, ResultComparator) ? BaselineState.Unchanged : BaselineState.New;
+                Result baselineResult = baseline.Results.FirstOrDefault(r => ResultComparator.Equals(r, newResult));
+
+                SetResultBaselineInformation(newResult, baselineResult, newRunTime, baselineRunTime);
 
                 differencedRun.Results.Add(newResult);
             }
 
-            foreach (var result in baseLine.Results)
+            foreach (var result in baseline.Results)
             {
                 if (!nextLog.Results.Contains(result, ResultComparator))
                 {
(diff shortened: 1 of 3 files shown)


lgolding:
>                     newResult.BaselineState = BaselineState.Absent;

I just noticed I'm not setting firstDetectionTime for results marked `Absent` (JC: these are results which have disappeared in the current run). I'll fix that.

---
Refers to: src/Sarif/Baseline/SarifLogBaseliner.cs:44 in f9ab9be. [](commit_id = f9ab9be0a0b5498107cdb7604f5fb2f75804908c, deletion_comment = False)

lgolding:
>                     newResult.BaselineState = BaselineState.Absent;

Of course, in this case, `newResult` is a clone of the baseline result -- so if the baseline result had a first detection time, it is automatically rolled forward -- and if it didn't, perhaps there's no point in populating it now, since the issue is now fixed.

I'll think more about this. It might still make sense to fall back to the baseline run execution time -- but it almost certainly _doesn't_ make sense to fall back to "now" if that is missing.

Thoughts?

---
In reply to: [579430768](https://github.com/microsoft/sarif-sdk/pull/1738#issuecomment-579430768) [](ancestors = 579430768)

---
Refers to: src/Sarif/Baseline/SarifLogBaseliner.cs:44 in f9ab9be. [](commit_id = f9ab9be0a0b5498107cdb7604f5fb2f75804908c, deletion_comment = False)

src/Sarif/Baseline/SarifLogBaseliner.cs
```diff
@@ -15,22 +16,27 @@ public SarifLogBaseliner(IEqualityComparer<Result> comparator)
             ResultComparator = comparator;
         }
 
-        public Run CreateBaselinedRun(Run baseLine, Run nextLog)
+        public Run CreateBaselinedRun(Run baseline, Run nextLog)
```
ScottLouvau: Hmm - a bigger problem. I think this class is no longer in use. I think you want to modify SarifLogResultMatcher, and specifically look at SarifLogResultMatcher.BaselineSarifLogs and MatchedResult.CalculateBaselinedResult.

src/Sarif/Baseline/SarifLogBaseliner.cs
```diff
@@ -42,5 +48,68 @@ public Run CreateBaselinedRun(Run baseLine, Run nextLog)
...
+            if (baselineResult == null)
+            {
+                currentResult.BaselineState = BaselineState.New;
+                SetFirstDetectionTime(currentResult, newRunTime);
+            }
+            else
+            {
+                currentResult.BaselineState = BaselineState.Unchanged;
+
+                DateTime firstDetectionTime = baselineResult.Provenance != null && baselineResult.Provenance.FirstDetectionTimeUtc != DateTime.MinValue
```
ScottLouvau: It might be clearer if you merge all of this into the method - see if a Provenance exists, if there's a previous FirstDetectionTime, if it's less than what you otherwise intend to set, and then set it.

lgolding:
@ScottLouvau points out that this code is in the wrong place. The `SarifLogBaseliner` class is obsolete, replaced by his `SarifLogResultMatcher` class. I'll close this PR, move the logic to the right place, and add an `ObsoleteAttribute` to `SarifLogBaseliner`.

ScottLouvau:
> > ```
> >                 newResult.BaselineState = BaselineState.Absent;
> > ```
> 
> Of course, in this case, `newResult` is a clone of the baseline result -- so if the baseline result had a first detection time, it is automatically rolled forward -- and if it didn't, perhaps there's no point in populating it now, since the issue is now fixed.
> 
> I'll think more about this. It might still make sense to fall back to the baseline run execution time -- but it almost certainly _doesn't_ make sense to fall back to "now" if that is missing.
> 
> Thoughts?
> 
> In reply to: [579430768](https://github.com/microsoft/sarif-sdk/pull/1738#issuecomment-579430768) [](ancestors = 579430768)
> 
> Refers to: src/Sarif/Baseline/SarifLogBaseliner.cs:44 in [f9ab9be](https://github.com/microsoft/sarif-sdk/commit/f9ab9be0a0b5498107cdb7604f5fb2f75804908c). [](commit_id = [f9ab9be](https://github.com/microsoft/sarif-sdk/commit/f9ab9be0a0b5498107cdb7604f5fb2f75804908c), deletion_comment = False)

Ah, right - if the issue is absent, I think ideally the First and Last detection times are left alone, as the values from the prior run. For Unchanged issues, the Last detection time should be from the new run/result, and the FirstDetectionTime should be kept from the old run/result.

PR #1739, opened 2020-01-29
Fix # 1737: SARIF result matcher does not populate firstDetectionTimeUtc

lgolding:
**This replaces the previous PR** and puts the logic in the right place: in Scott's result matcher, which replaced Everett's.

Roll the first detection time forward from the previous result if possible. Otherwise fall back to the time that the run was performed. Finally, fall back to "now" as the first detection time.

**NOTE** The logic embodies a couple of choices about which reasonable people might disagree:
- I choose "run end time" over "run start time" when both are available.
- I use "now" as the ultimate fallback if there's absolutely no other time information available.

Feel free to disagree.

### DIFF
--- a/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
+++ b/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
@@ -30,9 +30,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
         }
 
         /// <summary>
-        /// Creates a new SARIF Result object with contents from the
-        /// most recent result of the matched pair, the appropriate state,
-        /// and some metadata in the property bag about the matching algorithm used.
+        /// Creates a new SARIF Result object with contents from the most recent result of the
+        /// matched pair, the appropriate state and time of first detection, and some metadata
+        /// in the property bag about the matching algorithm used.
         /// </summary>
         /// <returns>The new SARIF result.</returns>
         public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMergeBehavior)
@@ -56,6 +56,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
                 // Baseline result is present, current result is missing => absent.
                 result = ConstructAbsentResult(resultMatchingProperties, out originalResultMatchingProperties);
             }
+
+            SetFirstDetectionTime(result);
 
             resultMatchingProperties = MergeDictionaryPreferFirst(resultMatchingProperties, originalResultMatchingProperties);
 
@@ -125,7 +127,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
         {
             // Result exists.
             Result result = CreateBaselinedResult(BaselineState.Unchanged);
-            
+
             if (!PreviousResult.Result.TryGetProperty(SarifLogResultMatcher.ResultMatchingResultPropertyName, out originalResultMatchingProperties))
             {
                 originalResultMatchingProperties = new Dictionary<string, object>();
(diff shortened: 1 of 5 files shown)


src/Test.UnitTests.Sarif/Baseline2/V2ResultMatcherTests.cs
```diff
@@ -347,12 +349,20 @@ public void ResultMatchingBaseliner_WhenThereIsOnlyOneCurrentRun_CopiesSelectedR
...
+                        }
                     },
                     new Result
                     {
                         RuleId = TestConstants.RuleIds.Rule2,
-                        RuleIndex = 1
+                        RuleIndex = 1,
+                        Provenance = new ResultProvenance
+                        {
+                            FirstDetectionTimeUtc = firstDetectionTime
```
lgolding: >FirstDetectionTimeUtc [](start = 28, length = 21)

Now that the result matcher adds `firstDetectionTimeUtc` to the current result, we must place it on the previous result as well or this test will fail, because it performs a naive property-by-property comparison of the current to the previous result (using `ResultEqualityComparer`).

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -173,6 +175,59 @@ private Result CreateBaselinedResult(BaselineState newBaselineState)
...
+            {
+                firstDetectionTime = DateTime.UtcNow;
+            }
+
+            if (targetResult.Provenance == null)
+            {
+                targetResult.Provenance = new ResultProvenance();
+            }
+
+            targetResult.Provenance.FirstDetectionTimeUtc = firstDetectionTime;
```
ScottLouvau: Should you also set LastDetectionTime to the current Run time for all non-Absent results?

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -173,6 +175,59 @@ private Result CreateBaselinedResult(BaselineState newBaselineState)
...
+            if (firstDetectionTime == default)
+            {
+                firstDetectionTime = DateTime.UtcNow;
+            }
+
+            if (targetResult.Provenance == null)
+            {
+                targetResult.Provenance = new ResultProvenance();
+            }
+
```
ScottLouvau: We should probably add the logic to copy the Run AutomationDetails Guids to the ResultProvenance Guid columns, too, with the same rules.

ScottLouvau: Hmm, sorry, I guess the logic is different. The FirstDetectionRunGuid should only be set for 'New' results (to the current run) and the last detection Guid should be set for all non-Absent results.

src/Test.UnitTests.Sarif/Baseline2/MatchedResultsTests.cs
```diff
@@ -0,0 +1,291 @@
...
+                            new Invocation
+                            {
+                                StartTimeUtc = currentRunStartTime
+                            }
+                        }
+                    },
+                    ExpectedBaselineState = BaselineState.New,
+                    ExpectedFirstDetectionTime = currentRunStartTime
+                },
+                new FirstDetectionTimeTestCase // TODO TEST CASE SHOWING CURRENT RUN WINS FOR NEW AND PREV RUN WINS FOR EXISTING
```
ScottLouvau: Still something to do given comment?

src/Test.UnitTests.Sarif/Baseline2/MatchedResultsTests.cs
```diff
@@ -0,0 +1,291 @@
...
+                        {
+                            new Invocation
+                            {
+                                StartTimeUtc = currentRunStartTime,
+                                EndTimeUtc = currentRunEndTime
+                            }
+                        }
+                    },
+                    ExpectedBaselineState = BaselineState.New,
+                    ExpectedFirstDetectionTime = currentRunEndTime
```
ScottLouvau: If/When you add logic to set LastDetectionTime, FirstDetectionRunGuid, and LastDetectionRunGuid, check for those, too.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -173,6 +175,59 @@ private Result CreateBaselinedResult(BaselineState newBaselineState)
             return result;
         }
 
+        internal void SetFirstDetectionTime(Result targetResult)
```
ScottLouvau: Please sanity test with a multitool match-results-forward case just to make sure this logic runs from end-to-end result matching also. I don't think this needs to be unit test necessarily - the unit tests and a one-off verification that everything is really hooked up is probably fine.

src/Test.UnitTests.Sarif/Baseline2/MatchedResultsTests.cs
lgolding: Good catch, I added the case.

---
In reply to: [372134533](https://github.com/microsoft/sarif-sdk/pull/1739#discussion_r372134533) [](ancestors = 372134533)

src/Test.UnitTests.Sarif/Baseline2/MatchedResultsTests.cs
lgolding: Sure, this will be the natural place to add those. WCP doesn't require any of those.

---
In reply to: [372135318](https://github.com/microsoft/sarif-sdk/pull/1739#discussion_r372135318) [](ancestors = 372135318)

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
lgolding: Thanks for the prodding :-) I made a copy of a test file with two issues. In the copy I added one, removed one of the two originals, and left the third alone. `provenance.firstDetectionTimeUtc` is populated as expected.

---
In reply to: [372135643](https://github.com/microsoft/sarif-sdk/pull/1739#discussion_r372135643) [](ancestors = 372135643)

PR #1741, opened 2020-02-03
NPM Build Enable

ScottLouvau:
Enabling build of NPM package with .NET Core 3.0 single-file-exe packaging of the Sarif.Multitool.

NOTES
- This does not add .NET Core 3.0 as a target by default because that forces VS2019+ and our AzureDevOps build pool does not have VS2019 instances.
- Sarif.Multitool.csproj adds settings conditionally when a runtime has been specified to stay VS2017 compatible; this build has been structured so it runs safely before or after BuildAndTest.ps1.
- Sarif.csproj must be built with the current targets also included, because src/Sarif/obj/project.assets.json is written on every build and breaks NuGet packaging if written without the framework targets that the NuGet package is using.
- Consolidated on "dotnet" for build and test everywhere to unblock .NET Core 3.0 build; should also enable builds where VS is not installed.
- .NET Core 3.0 makes *huge* builds, so outputs are trimmed before publish.
- Published artifact size went from 792MB to 1,564MB (.NET Core 3) to 347MB (trimmed).
- Build times go from ~20-25 minutes to ~10-15 minutes despite extra builds due to the huge publish time savings.

CHANGES
- Adding NPM packaging configuration under /npm (by @jeffersonking).
- Added scripts/BuildMultitoolForNpm.ps1 to build .NET Core 3.0 multitool, single file exes, and completed NPM packages.
- Added scripts/DeleteTestOutputs.ps1 to clean build outputs for publish.
- Added missing test projects to scripts/Projects.psm1.
- Added missing $ScriptPath definitions to avoid error while reporting errors from some scripts.
- Modified output path of published content (NuGet and NPM packages) to bld/Publish to allow AzureDevOps build to easily upload just the needed content; a cleaned bld/bin is also published so we retain signed binaries of everything.
- Added Multitool documentation (separately); linking from NPM READMEs.

### DIFF
--- a/scripts/BuildMultitoolForNpm.ps1
+++ b/scripts/BuildMultitoolForNpm.ps1
@@ -34,7 +34,7 @@ dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcore
 dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.0 /p:TargetFrameworks=netcoreapp3.0 -r osx-x64
 
 Write-Information "Merging binaries [$projectBinDirectory] and NPM configuration [$npmSourceFolder]..."
-New-Item -ItemType Directory -Path $npmBuildFolder\
+New-DirectorySafely $npmBuildFolder\
 Copy-Item -Force -Container -Recurse -Path $npmSourceFolder\* -Destination $npmBuildFolder\
 Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\win-x64\* -Destination $npmBuildFolder\sarif-multitool-win32\
 Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\linux-x64\* -Destination $npmBuildFolder\sarif-multitool-linux\
--- a/scripts/BuildMultitoolForNpm.ps1
+++ b/scripts/BuildMultitoolForNpm.ps1
@@ -33,11 +33,12 @@ dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcore
 dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.0 /p:TargetFrameworks=netcoreapp3.0 -r linux-x64
 dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.0 /p:TargetFrameworks=netcoreapp3.0 -r osx-x64
 
-Write-Information "Merging binaries and NPM configuration..."
-# Copy-Item -Force -Container -Recurse -Path $npmSourceFolder\* -Destination $npmBuildFolder
-Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\win-x64\* -Destination $npmBuildFolder\sarif-multitool-win32
-Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\linux-x64\* -Destination $npmBuildFolder\sarif-multitool-linux
-Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\osx-x64\* -Destination $npmBuildFolder\sarif-multitool-darwin
+Write-Information "Merging binaries [$projectBinDirectory] and NPM configuration [$npmSourceFolder]..."
+New-Item -ItemType Directory -Path $npmBuildFolder\
+Copy-Item -Force -Container -Recurse -Path $npmSourceFolder\* -Destination $npmBuildFolder\
+Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\win-x64\* -Destination $npmBuildFolder\sarif-multitool-win32\
+Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\linux-x64\* -Destination $npmBuildFolder\sarif-multitool-linux\
+Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\osx-x64\* -Destination $npmBuildFolder\sarif-multitool-darwin\
 
 # To Publish:
 # from [sarif-sdk]\bld\bin\AnyCPU_Release\Sarif.Multitool\npm, run:
--- a/scripts/BuildMultitoolForNpm.ps1
+++ b/scripts/BuildMultitoolForNpm.ps1
@@ -51,7 +51,7 @@ if (-not $SkipBuild) {
 
 # TEMP: Use 0.x.y numbering until NPM package shape settled, then follow SDK version exactly by removing the hardcoded set
 $sarifVersion = Get-PackageVersion
-$sarifVersion = "0.1.25"
+$sarifVersion = "0.1.26"
 
 Write-Information "Injecting Sarif SDK version $sarifVersion for NPM Packages..."
 foreach ($package in (Get-ChildItem $npmBuildFolder).FullName) {
--- a/scripts/BuildMultitoolForNpm.ps1
+++ b/scripts/BuildMultitoolForNpm.ps1
@@ -14,6 +14,9 @@ param(
     $Configuration="Release",
 
     [switch]
+    $SkipBuild,
+
+    [switch]
     $NoPostClean
 )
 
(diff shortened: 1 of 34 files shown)


src/build.props
```diff
@@ -42,6 +42,7 @@
     <IntermediateOutputPath>$(MsBuildThisFileDirectory)..\bld\obj\$(OutputSubDir)\$(MSBuildProjectName)\</IntermediateOutputPath>
     <OutputPath>$(MsBuildThisFileDirectory)..\bld\bin\$(OutputSubDir)\$(MSBuildProjectName)\</OutputPath>
     <PublishDir>$(OutputPath)\Publish\$(TargetFramework)\</PublishDir>
+    <PublishDir Condition=" '$(RuntimeIdentifier)' != '' ">$(OutputPath)\Publish\$(TargetFramework)\$(RuntimeIdentifier)\</PublishDir>
```
lgolding: >$(OutputPath)\Publish\$(TargetFramework)\ [](start = 59, length = 41)

You could DRY this out by using `$(PublishDir)` here because the properties are evaluated in order.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -2,11 +2,29 @@
   
   <PropertyGroup>
     <OutputType>Exe</OutputType>
+
+    <!-- To Publish single-file exes, run: 
+      dotnet publish Sarif.Multitool.csproj -c Release -f netcoreapp3.0 -r win-x64
+      dotnet publish Sarif.Multitool.csproj -c Release -f netcoreapp3.0 -r linux-x64
+      dotnet publish Sarif.Multitool.csproj -c Release -f netcoreapp3.0 -r osx-x64
+
+      For settings which we want only for these builds, we use [Condition="'$(RuntimeIdentifier)' != ''"] to identify them
```
lgolding: > F [](start = 5, length = 2)

Suggestion: "We use [Condition="'$(RuntimeIdentifier)' != ''"] to identify settings that are used only for these builds."

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -2,11 +2,29 @@
...
+    -->
+
+    <!-- Keep default targets to allow VS2017 use; our build pool is Hosted VS2017 only. -->
```
lgolding: >our build pool is Hosted VS2017 only [](start = 51, length = 36)

Since this is a public repo, I wouldn't mention "our" build pool. Maybe just "Build for frameworks prior to .NET Core 3.0 to enable building on VS2017."

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -2,11 +2,29 @@
...
     <TargetFrameworks>netcoreapp2.0;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks Condition="'$(RuntimeIdentifier)' != ''">netcoreapp2.0;netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
```
lgolding: >netcoreapp2.0;netcoreapp2.1 [](start = 63, length = 27)

DRY out as `$(TargetFrameworks);netcoreapp3.0.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -2,11 +2,29 @@
...
+
+    <!-- When dotnet publish run, produce trimmed single-file exe -->
```
lgolding: >run [](start = 29, length = 3)

"runs"

scripts/ScriptUtilities.psm1
```diff
@@ -54,12 +54,24 @@ function Write-CommandLine($exeName, $arguments) {
     Write-Verbose "$exeName $($arguments -join ' ')"
 }
 
+function Find-AndReplaceInFile($filePath, $value, $withValue) {
+    $tempFilePath = "$filePath.tmp"
+    if (Test-Path -Path $tempFilePath) { 
+        Remove-Item $tempFilePath 
+    }
+
+    (Get-Content -Path $filePath) -replace $value, $withValue | Add-Content -Path $tempFilePath -Force
```
lgolding: >( [](start = 4, length = 1)

sweet :-)

scripts/BuildMultitoolForNpm.ps1
```diff
@@ -0,0 +1,75 @@
+<#
+.SYNOPSIS
+    Build, and package the Sarif Multitool NPM package.
```
lgolding: >, [](start = 9, length = 1)

no comma

lgolding: >Sarif Multitool [](start = 27, length = 15)

Sarif.Multitool (or "SARIF Multitool" if the idea is to describe it rather than name it).

scripts/BuildMultitoolForNpm.ps1
```diff
@@ -0,0 +1,75 @@
...
+Import-Module -Force $PSScriptRoot\Projects.psm1
+
+$project = "Sarif.Multitool"
+$projectBinDirectory = (Get-ProjectBinDirectory $project $Configuration)
+$npmSourceFolder = "$RepoRoot\npm"
+$npmBuildFolder = "$BuildRoot\Publish\npm"
+
+if (-not $SkipBuild) {
+    Write-Information "Building Sarif.Multitool for Windows, Linux, and MacOS..."
+    dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.0 -r win-x64
```
lgolding: >d [](start = 4, length = 1)

Consider DRYing this out:
```
foreach ($runtime in "win-x64", "linux-x64", "osx-x64") {
    dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.0 -r $runtime
}
```

scripts/BuildMultitoolForNpm.ps1
```diff
@@ -0,0 +1,75 @@
...
+    Copy-Item -Force -Container -Recurse -Path $projectBinDirectory\Publish\netcoreapp3.0\osx-x64\* -Destination $npmBuildFolder\sarif-multitool-darwin\
+}
+
+# TEMP: Use 0.x.y numbering until NPM package shape settled, then follow SDK version exactly by removing the hardcoded set
+$sarifVersion = Get-PackageVersion
+$sarifVersion = "0.1.25"
+
+Write-Information "Injecting Sarif SDK version $sarifVersion for NPM Packages..."
+foreach ($package in (Get-ChildItem $npmBuildFolder).FullName) {
+    Find-AndReplaceInFile "$package\package.json" "0\.1\.0" $sarifVersion
```
lgolding: >0\.1\.0 [](start = 51, length = 7)

This assumes that `"0.1.0"` appears nowhere else in the file (for example as the version number of some other dependency). That's probably safe, but you could choose instead to use some other token, like `$SarifSdkVersion`, in the package.json files.

scripts/DeleteTestOutputs.ps1
```diff
@@ -0,0 +1,32 @@
+<#
+.SYNOPSIS
+    Delete test outputs from bld\bin to reduce published content size.
```
lgolding: >Delete [](start = 4, length = 6)

I don't see anybody calling this. Is it intended as a separate step in a build pipeline definition?

ScottLouvau: Yes. Our new Azure DevOps build calls this near the end. Developers don't need to call this for local builds, so I didn't want to tie it in anywhere by default.

src/build.props
ScottLouvau: Sure

src/Sarif.Multitool/Sarif.Multitool.csproj
ScottLouvau: Good rewording - fixed.

src/Sarif.Multitool/Sarif.Multitool.csproj
ScottLouvau: It turns out the Hosted VS2017 build agent pool is public; I've updated the comment to be clearer.

src/Sarif.Multitool/Sarif.Multitool.csproj
ScottLouvau: Fixed

src/Sarif.Multitool/Sarif.Multitool.csproj
ScottLouvau: Reworded to be more consistent with the next comment.

scripts/BuildMultitoolForNpm.ps1
ScottLouvau: Fixed

ScottLouvau: Fixed

scripts/BuildMultitoolForNpm.ps1
ScottLouvau: Good suggestion. Settled on {version} to not interfere with special symbols in PowerShell ($) or Regex.

PR #1743, opened 2020-02-05
Adding lost documentation from the Wiki and updating some how-to steps

ScottLouvau:


docs/json-map.md
```diff
@@ -0,0 +1,72 @@
+# Overview
+JSON Map is a JSON-based format which describes part of the structure of another JSON document. Maps inform clients of the sizes of parts of the JSON and to allow constant time construction of subsets of the document.
```
lgolding: >and to allow [](start = 150, length = 12)

Either "and allow" or "to allow". I like "to allow" better because that's the _reason_ for doing it, not just another, independent _consequence_ of doing it.

docs/json-map.md
```diff
@@ -0,0 +1,72 @@
...
+
+For example, suppose you have a 20GB JSON file which contains a huge array with one million elements along with some surrounding document structure. You can construct a map with a 0.1% target size (20 MB). The map will identify the path to the huge array, the number of elements it has, where each nth array element begins, and where the array itself starts and ends.
```
lgolding: >huge [](start = 64, length = 4)

Copyediting nit: "huge" isn't necessary. "One million elements" conveys the hugeness :-)

docs/json-map.md
```diff
@@ -0,0 +1,72 @@
...
+
+To write a copy of the file with only the first 1,000 array elements, copy the file content up to the array start, copy the range from element 0 to element 999, omit the last comma, and copy the range from the array end to the end of the file.
+
+
+# JSON Map Structure
+A JSON Map consists of a hierarchy of **nodes** which mirror the hierarchical structure of the underlying JSON document. A map is constructed with a specific **target size ratio** (ex: 1%) which determines a size threshold for including items in the map.
+
+Each **node** contains:
+* A "start" integer, which is the absolute byte position in the file of the start of the object or array (the index of the '[' or '{')
+* An "end" integer, which is the absolute byte position in the file of the end of the object or array (the index of the ']' or '}')
```
lgolding: >end [](start = 6, length = 3)

Just checking -- "end" really points to the closing brace, rather than one past it? In iterators, the "end" iterator conventionally points one past, so that "end" - "start" == length. Not suggesting you change it; just making sure this is really what you meant.

docs/json-map.md
```diff
@@ -0,0 +1,72 @@
+# Overview
+JSON Map is a JSON-based format which describes part of the structure of another JSON document. Maps inform clients of the sizes of parts of the JSON and to allow constant time construction of subsets of the document.
+
+For example, suppose you have a 20GB JSON file which contains a huge array with one million elements along with some surrounding document structure. You can construct a map with a 0.1% target size (20 MB). The map will identify the path to the huge array, the number of elements it has, where each nth array element begins, and where the array itself starts and ends. 
+
+To write a copy of the file with only the first 1,000 array elements, copy the file content up to the array start, copy the range from element 0 to element 999, omit the last comma, and copy the range from the array end to the end of the file.
+
+
+# JSON Map Structure
+A JSON Map consists of a hierarchy of **nodes** which mirror the hierarchical structure of the underlying JSON document. A map is constructed with a specific **target size ratio** (ex: 1%) which determines a size threshold for including items in the map.
```
lgolding: >nodes [](start = 40, length = 5)

It would really help to have an example showing the skeleton of a SARIF file together with the corresponding map file.

docs/json-map.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Uses
+JSON Maps can be used to:
+
+### Construct subsets of JSON documents
+A map can be used to find any large object or array in a JSON document. One can then parse that subset of the JSON individually, make a new JSON document containing only that item, or make a JSON document like the original which excludes that item.
+
+
+### Inform clients of JSON size and structure
+A map can be sent to a client (for example, a web viewer) to allow it to determine whether to request the whole file or only subsets. A viewer can request a subset of the file with large portions excluded and then, if needed, can request those subsets only to reconstruct the full document. A viewer can request "pages" of elements from large arrays, allowing it to show only the first 100 elements, for example, and then retrieve additional pages as a user scrolls. The exact count is available in the map, so the viewer can display the total number of items available and size scroll bars and similar interface elements appropriately.
```
lgolding: >those subsets only [](start = 238, length = 18)

"only those subsets"

docs/query-mode.md
```diff
@@ -0,0 +1,51 @@
+The Sarif.Multitool has a 'query' mode. Query mode takes a SarifLog and query expression and determines how many Results match the expression. It can write the count to the console, return the count as an exit code, signal if the count exceeds a threshold, or write an output log with the filtered results only.
```
lgolding: General comment: consider using backticks around code elements and commands, for example,

`Sarif.Multitool query In.sarif -e "RuleId = SM00251"`

`=`, `==`

lgolding: >mode [](start = 34, length = 4)

FWIW, the CommandLine package refers to these as verbs, and it matters a little because the terminology leaks into the help messages, for example:
```
C:\dev\sarif\WebScout\Source>sarif /help
SARIF Multitool 2.1.24
c Microsoft Corporation. All rights reserved.

ERROR(S):
  Verb '/help' is not recognized.
```

docs/query-mode.md
```diff
@@ -0,0 +1,51 @@
...
+| Command | Meaning |
+| ------- | ------- |
+| Sarif.Multitool query In.sarif -e "RuleId = SM00251"    | Output Result count with RuleId SM00251 |
+| Sarif.Multitool query In.sarif -e "RuleId = SM00251" -o Out.sarif | ...and write filtered log to 'Out.sarif' |
+| ... -e "Uri : 'Test Framework'" | Find Results where any URI contains 'Test Framework' |
+| ... -e "!(Uri >&#124;> .cs)"   | Find Results where no URI EndWith '.cs' |
+| ... -e "OccurrenceCount > 10 AND RuleId = SM00251" | Look for Results which match both criteria |
+
+
+### Comparison Operators
```
lgolding: >Operators [](start = 15, length = 9)

Is there a precedence hierarchy? If so, this table should be in hierarchical order (and should say so). If there's no hierarchy, we should _really_ say so :-)

docs/json-map.md
ScottLouvau: fixed

docs/json-map.md
ScottLouvau: fixed

docs/json-map.md
ScottLouvau: Yes, that's the intent. It's easier to get the position of a token ('{') from the parser/reader than the position after it. The position after it also may or may not be whitespace, depending on whether the JSON is indented, causing other weirdness when concatenating subsets of JSON from these offsets.

docs/json-map.md
ScottLouvau: fixed

docs/json-map.md
ScottLouvau: Added.

docs/query-mode.md
ScottLouvau: Now "has a 'query' verb"

ScottLouvau: Good suggestion - fixed.

docs/query-mode.md
ScottLouvau: Very good suggestion. They are: NOT, AND, OR. I've ordered them and added an example.

docs/query-mode.md
```diff
@@ -0,0 +1,56 @@
...
+Query mode is not as sophisticated as the SARIF viewers, but can be used to separate logs which are too large into meaningful subsets or to trigger on simple conditions as part of an automated workflow. For arbitrarily complex scenarios, it's easiest to reference the Sarif SDK and write the logic as code.
+
+### Examples
+
+| Command | Meaning |
+| ------- | ------- |
+| `Sarif.Multitool query In.sarif -e "RuleId = SM00251"`    | Output Result count with RuleId SM00251 |
+| `Sarif.Multitool query In.sarif -e "RuleId = SM00251" -o Out.sarif` | ...and write filtered log to 'Out.sarif' |
+| `... -e "Uri : 'Test Framework'"` | Find Results where any URI contains 'Test Framework' |
+| `... -e "!(Uri >&#124;> .cs)"`   | Find Results where no URI EndWith '.cs' |
```
lgolding: >&#124; [](start = 18, length = 6)

Crud. Now the entity `&#124` will render literally on the wiki page.

docs/query-mode.md
lgolding: And the other table, with `==`, `<`, etc -- does that also have a hierarchy? I think we need a single table with the entire hierarchy.

---
In reply to: [375554889](https://github.com/microsoft/sarif-sdk/pull/1743#discussion_r375554889) [](ancestors = 375554889)

PR #1744, opened 2020-02-05
Work item cleanup

michaelcfanning:
@nickmarston @chinarosesz 

Draft request. I'll comment thoroughly on this tomorrow am.


src/Test.FunctionalTests.Sarif/v2/ConverterTestData/FxCop/FxCopExceptions.xml.sarif
```diff
@@ -28,6 +23,11 @@
           }
         }
       ],
+      "tool": {
```
michaelcfanning: >tool [](start = 7, length = 4)

This data move happens now when we refresh our test files via automated scripts, so I've gone ahead and checked the change in.

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -76,18 +77,19 @@ public int Run(FileWorkItemsOptions options, IFileSystem fileSystem)
...
+                        WorkItemModel workItemModel = splitLog.CreateWorkItemModel(projectName);
+                        workItemModels.Add(workItemModel);
                     }
 
                     try
                     {
-                        IEnumerable<WorkItemFilingMetadata> filedWorkItems = filer.FileWorkItems(options.ProjectUri, workItemMetadata, options.PersonalAccessToken).Result;
-
+                        string securityToken = Environment.GetEnvironmentVariable("SarifWorkItemFilingSecurityToken");
+                        IEnumerable<WorkItemModel> filedWorkItems = filer.FileWorkItems(options.ProjectUri, workItemModels, securityToken).Result;
```
chinarosesz: >Result [](start = 139, length = 6)

Minor: Is it possible if we don't block here?

src/Sarif.WorkItemFiling/AzureDevOpsClient.cs
```diff
@@ -67,37 +67,40 @@ public override async Task<IEnumerable<WorkItemFilingMetadata>> FileWorkItems(IE
...
+                        Path = $"/fields/{AzureDevOpsFieldNames.Tags}",
+                        Value = string.Join(",", metadata.Tags)
                     }
                 };
 
-                foreach (var customField in metadata.CustomFields)
+                if (metadata.CustomFields != null)
                 {
-                    patchDocument.Add(new JsonPatchOperation
+                    foreach (var customField in metadata.CustomFields)
```
chinarosesz: >var [](start = 29, length = 3)

Not in this PR but would be nice if this is replaced with type since it's not obvious what metadata.CustomFields is from the right side.

src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -218,7 +45,7 @@ public static string GetProjectName(this Uri projectUri)
             return WebUtility.UrlDecode(projectName);
         }
 
-        public static string GetAccountUriString(this Uri projectUri)
+        public static string GetAccountName(this Uri projectUri)
```
chinarosesz: >GetAccountName [](start = 29, length = 14)

Just an idea and not for this PR: I think it might be easier to take in the name of the project and the organization for Azure DevOps, then we wouldn't need to do any URL parsing.

src/Sarif.Multitool/FileWorkItemsCommand.cs
michaelcfanning: Can you clarify your comment? :)

---
In reply to: [375953433](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375953433) [](ancestors = 375953433)

src/Sarif.WorkItemFiling/AzureDevOpsClient.cs
michaelcfanning: agreed, fixed.

---
In reply to: [375955418](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375955418) [](ancestors = 375955418)

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,31 +9,22 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
-        // For an explanation of the difference between ProjectUriString and ProjectUri, see FileWorkItemsCommand.ValidateOptions.
+        // ProjectUri is the URI representation of ProjectUriString above. We convert the command-line argument to the URI
+        // form as part of argument validation at runtime (in order to provide better error messages than those that 
+        // the CommandlineLibrary component provides). See FileWorkItemsCommand.ValidateOptions for more details.
         public Uri ProjectUri { get; internal set; }
 
-        [Option(
-            "personal-access-token",
-            HelpText = "TEMPORARY: Specifes the personal access used to access the project")]
-        public string PersonalAccessToken { get; internal set; }
```
chinarosesz: >PersonalAccessToken [](start = 22, length = 19)

We should probably somehow ensure users are aware of setting environment variable now that this is removed. Maybe document this the verb <code>file-work-items</file>?

src/Sarif.Multitool/FileWorkItemsCommand.cs
chinarosesz: This is probably outside of the scope of this PR. Since this is an async method, ideally it should await but then we would have to update Run() to return Task<int> instead of int. I wouldn't worry too much about this right but just something for future thoughts. 

---
In reply to: [375957765](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375957765) [](ancestors = 375957765,375953433)

src/Sarif.Multitool/FileWorkItemsOptions.cs
```diff
@@ -9,31 +9,22 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
...
-
         [Option(
             "group",
             HelpText = "Apply a grouping strategy to factor input log file to multiple bugs. Must be one of All, PerRunPerRule or PerRunPerTargetPerRule.",
             Default = GroupingStrategy.PerRun)]
```
nickmarston: Not technically in this change, but the default grouping strategy is not one of the ones listed in the help text (or if it is, it's ambiguous between PerRunPerRule and PerRunPerTargerPerRule

src/Sarif.WorkItemFiling/FilingClientFactory.cs
```diff
@@ -7,25 +7,36 @@
...
+            {
+                return new AzureDevOpsClient();
             }
             else if (s_gitHubUriRegex.IsMatch(projectUriString))
             {
-                return new GitHubFilingTarget();
+                return new GitHubClient();
             }
             else
             {
```
nickmarston: I suspect your most commonly in this state because the user left off the project portion of the URI.  If we don't take Lila's suggestion above of having the user specify them explicitly, I would definitely re-iterate the format expect in this exception, even if it's already in the help text.

src/Sarif.WorkItemFiling/Extensions.cs
michaelcfanning: That is true. However, if this information is already expressed in the URL then it is redundant to require that it be passed again. Also creates new negative cases where the alternate identifier might not match the URL. For ADO URLs, the account and project are definitely codified in a specific location, so I think this parsing is a fair approach. Thoughts?

---
In reply to: [375957323](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375957323) [](ancestors = 375957323)

src/Sarif.Multitool/FileWorkItemsCommand.cs
michaelcfanning: Got it. The way this thing works currently is that we must generate a new SARIF file that contains the URLs to the constructed work items. There's no option currently to skip this (i.e., to simply fire the work item creation request and not wait around for it). That option could be important in the future, I agree. 

---
In reply to: [375968167](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375968167) [](ancestors = 375968167,375957765,375953433)

src/Sarif.WorkItemFiling/FilingClientFactory.cs
michaelcfanning: Agreed, will put the hood up here in the next check-in and do what you suggest.

---
In reply to: [375988292](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375988292) [](ancestors = 375988292)

src/Sarif.WorkItemFiling/Extensions.cs
chinarosesz: Sorry I forgot to mention that either we require the URL approach or org&projectname approach. I wouldn't do both. If we are to take in the org name and project name, we can construct the URL based on the input params from org/project, this approach avoids us from having to parse the URL all together. Given that said, I think it's fine to keep what we have as it's working, just wanted to see if my approach might be easier to manage.

---
In reply to: [375990369](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375990369) [](ancestors = 375990369,375957323)

michaelcfanning: Makes sense! This might require us to have a host identifier (GH vs. ADO) or to infer this based on having different command-line arguments. Let's defer this conversation as you suggest, but definitely let's consider in a future change. 

---
In reply to: [376006810](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r376006810) [](ancestors = 376006810,375990369,375957323)

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: TODO for next PR. Let's discuss offline.

---
In reply to: [375966103](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375966103) [](ancestors = 375966103)

src/Sarif.Multitool/FileWorkItemsOptions.cs
michaelcfanning: Good catch, fixed.

---
In reply to: [375983193](https://github.com/microsoft/sarif-sdk/pull/1744#discussion_r375983193) [](ancestors = 375983193)

PR #1746, opened 2020-02-07
More code clean-up. Add simple GitHub integration via Octokit.

michaelcfanning:
1. Added a reference to a custom-compiled version of Octokit (which is signed, a requirement for our scenario).
2. Update filing client factory to retrieve project/account details from host uri and populate client object.
3. Work item filing works for GitHub now, as well as using legacy or current ADO URIs.

@chinarosesz @nickmarston @harleenkohli


src/Sarif/SarifUtilities.cs
```diff
@@ -184,7 +184,7 @@ public static int GetByteLength(string s, Encoding encoding)
         public static Regex RegexFromPattern(string pattern) =>
             new Regex(
                 pattern,
-                RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.ExplicitCapture | RegexOptions.IgnorePatternWhitespace);
+                RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.IgnorePatternWhitespace);
```
michaelcfanning: >IgnorePatternWhitespace [](start = 85, length = 23)

Explicit capture is a more performant option for intensive regex use. I deleted it for expedience sake, but would be glad to update the regexes to use explicitly named groups. Will do this as a follow-on.

src/Sarif.WorkItemFiling/FilingClientFactory.cs
```diff
@@ -3,49 +3,65 @@
...
+                    filingClient.AccountOrOrganization = match.Groups[1].Value;
+                    filingClient.ProjectOrRepository = match.Groups[2].Value;
+                    break;
+                }
             }
-            else
+
+            if (filingClient == null)
             {
                 throw new ArgumentException(
```
michaelcfanning: >ArgumentException [](start = 26, length = 17)

Improved reporting per previous feedback. We now provide example URIs.

src/Sarif.WorkItemFiling/FilingClient.cs
```diff
@@ -13,22 +13,33 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItemFiling
     /// </summary>
     public abstract class FilingClient
     {
+        /// <summary>
+        /// The URI to the work item filing host.
+        /// </summary>
+        public Uri HostUri { get; internal set; }
```
michaelcfanning: >HostUri [](start = 19, length = 7)

New properties are populated by the factory.

michaelcfanning:
1. Added a reference to a custom-compiled version of Octokit (which is signed, a requirement for our scenario).
2. Update filing client factory to retrieve project/account details from host uri and populate client object.
3. Work item filing works for GitHub now, as well as using legacy or current ADO URIs.

src/Sarif.WorkItemFiling/AzureDevOpsClientWrapper.cs
```diff
@@ -123,7 +119,7 @@ public override async Task<IEnumerable<WorkItemModel>> FileWorkItems(IEnumerable
                 try
                 {
                     Console.Write($"Creating work item: {metadata.Title}");
-                    workItem = await _witClient.CreateWorkItemAsync(patchDocument, project: _projectName, "Bug");
+                    workItem = await _witClient.CreateWorkItemAsync(patchDocument, project: this.ProjectOrRepository, "Bug");
```
chinarosesz: >Bug [](start = 119, length = 3)

Should we change this default now or later? Currently AzureDevOps defaults work item type as "Issue".

src/Sarif.WorkItemFiling/GitHubClientWrapper.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            var credentials = new Credentials(this.AccountOrOrganization, personalAccessToken);
+            var credentialsStore = new InMemoryCredentialStore(credentials);
+
+            await Task.Run(() =>
+            {
+                this.gitHubClient = new GitHubClient(new ProductHeaderValue(this.AccountOrOrganization), credentialsStore);
+            });
+        }
+        public override async Task<IEnumerable<WorkItemModel>> FileWorkItems(IEnumerable<WorkItemModel> workItemFilingMetadata)
```
chinarosesz: Line break after brace

src/Sarif.WorkItemFiling/GitHubClientWrapper.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            foreach (WorkItemModel workItemModel in workItemFilingMetadata)
+            {
+                var newIssue = new NewIssue(workItemModel.Title)
+                {
+                    Body = workItemModel.Body,
+                };
+                Issue issue = await this.gitHubClient.Issue.Create("michaelcfanning", "bug-dummy", newIssue);
```
chinarosesz: >"michaelcfanning", "bug-dummy" [](start = 67, length = 30)

Just for now I assume?

src/Sarif.WorkItemFiling/Sarif.WorkItemFiling.csproj
```diff
@@ -4,9 +4,16 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard2.0|AnyCPU'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsAsErrors />
+    <NoWarn>NU5105kk</NoWarn>
```
chinarosesz: >kk [](start = 18, length = 2)

I'm curious what this rule means, never seen one with kk at the end before.

src/Sarif.WorkItemFiling/Sarif.WorkItemFiling.csproj
```diff
@@ -22,4 +29,10 @@
     <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.143.2" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="Octokit">
+      <HintPath>..\..\refs\Octokit.dll</HintPath>
```
chinarosesz: >Octokit [](start = 27, length = 7)

Can we reference this from nuget.org as a package or is it because this particular one is signed and can only work for us?

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.CodeAnalysis.Sarif.WorkItemFiling;
 using Microsoft.Json.Schema;
 using Microsoft.Json.Schema.Validation;
+using Microsoft.TeamFoundation.Build.WebApi;
```
chinarosesz: >using Microsoft.TeamFoundation.Build.WebApi [](start = 0, length = 43)

Pretty sure this can be removed. I remember this is only used for build specific API's and doesn't look like we are doing anything around Azure DevOps build related stuff yet.

src/Sarif.WorkItemFiling/AzureDevOpsClientWrapper.cs
nickmarston: I like using the default for now, though this should be one of our first customization points.

michaelcfanning: An issue is a task in the ADO parlance. I think tasks are likely not to be the default for an automated filing system (i.e., bugs would be). Interested in others thoughts on this. Agree with Nick, let's defer a change for now.

---
In reply to: [376555448](https://github.com/microsoft/sarif-sdk/pull/1746#discussion_r376555448) [](ancestors = 376555448)

src/Sarif.WorkItemFiling/Sarif.WorkItemFiling.csproj
michaelcfanning: That's a mistake, this code shouldn't have been added. the kk is a typo. Thanks for catching this, not sure how I missed it on my preliminary review...

---
In reply to: [376502450](https://github.com/microsoft/sarif-sdk/pull/1746#discussion_r376502450) [](ancestors = 376502450)

src/Sarif.Multitool/FileWorkItemsCommand.cs
michaelcfanning: Good catch, removed.

---
In reply to: [376508933](https://github.com/microsoft/sarif-sdk/pull/1746#discussion_r376508933) [](ancestors = 376508933)

src/Sarif.WorkItemFiling/Sarif.WorkItemFiling.csproj
michaelcfanning: Right, we cannot reference from NuGet as this version is unsigned. I recompiled and signed this version using our checked in key.

---
In reply to: [376503279](https://github.com/microsoft/sarif-sdk/pull/1746#discussion_r376503279) [](ancestors = 376503279)

src/Sarif.WorkItemFiling/GitHubClientWrapper.cs
michaelcfanning: No! this s.be picked up from the FileClient properties. :P Fixed.

---
In reply to: [376501091](https://github.com/microsoft/sarif-sdk/pull/1746#discussion_r376501091) [](ancestors = 376501091)

PR #1748, opened 2020-02-10
Attempt to fix up LGTM Semmle build.

michaelcfanning:
@chinarosesz @nickmarston @harleenkohli  FYI

We should see the checks phase succeed if this fix works.


src/lgtm.yml
```diff
@@ -0,0 +1,6 @@
+extraction:
+  csharp:
+     index:
+        all_solutions: true
+        build_command:
+        - BuildAndTest.cmd
```
chinarosesz: >BuildAndTest [](start = 10, length = 12)

Looks like PR succeeded after this is moved

PR #1749, opened 2020-02-10
Refactor work item filing code to isolate non-SARIF aware pieces.

michaelcfanning:


src/WorkItemFiling/WorkItemFilingUtilities.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.WorkItemFiling
+{
+    public static class WorkItemFilingUtilities
+    {
+        internal const string PROJECT = "account";
+        internal const string ACCOUNT = "project";
+
+        private static readonly string GitHubUriPattern = $"^https://github\\.com/(?<{ACCOUNT}>[^/]+)/(?<{PROJECT}>[^/]+)$";
```
michaelcfanning: >PROJECT [](start = 106, length = 7)

Note: as per previous PR I've added named groups so that we can use the RegexOptions.ExplicitCapture option (which improves performance).

michaelcfanning:
This change moves all low-level host-agnostic work-item related stuff into its own assembly (which the SARIF work item filer will take a dependency on). When these helpers/abstractions stabilize, they should be moved to their own repository/nuget package. Note that as part of this change, I've disabled the code that populates the work item back into a rewritten SARIF log file. Will add this back in the next PR (or this one, depending on everyone's speed of review).

@chinarosesz @nickmarston @harleenkohli

src/WorkItemFiling/AzureDevOpsClientWrapper.cs
```diff
@@ -64,31 +64,31 @@ public override async Task<IEnumerable<WorkItemModel>> FileWorkItems(IEnumerable
...
-                        Value = string.Join(",", metadata.Tags)
+                        Value = string.Join(",", workItemModel.Tags)
                     }
                 };
 
-                if (metadata.CustomFields != null)
+                if (workItemModel.CustomFields != null)
                 {
-                    foreach (var customField in metadata.CustomFields)
+                    foreach (var customField in workItemModel.CustomFields)
```
chinarosesz: >var [](start = 29, length = 3)

Replace with type here

src/WorkItemFiling/WorkItemFiling.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <AssemblyTitle>Universal Work Item Filing Library</AssemblyTitle>
+    <Description>Classes for filing work items in either GitHub or AzureDevOps.</Description>
+  </PropertyGroup>
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Json.Schema" Version="1.1.1" />
+    <PackageReference Include="Microsoft.Json.Schema.Validation" Version="1.1.1" />
+    <PackageReference Include="Microsoft.VisualStudio.Services.Client" Version="16.143.2" />
```
chinarosesz: >16.143.2 [](start = 80, length = 8)

Not sure if we want to do an update now, but the latest version 16.153.2 was released several months ago.

src/Sarif.WorkItemFiling/Extensions.cs
```diff
@@ -16,7 +18,7 @@ public static WorkItemModel CreateWorkItemModel(this SarifLog sarifLog, string p
             {
                 AdditionalData = sarifLog,
                 Area = projectName,
-                Attachment = new WorkItemFiling.Attachment
+                Attachment = new MicrosoftWorkItemFiling.Attachment
```
chinarosesz: >MicrosoftWorkItemFiling [](start = 33, length = 23)

Nit: We could replace this with Microsoft.WorkItemFiling and then there is no need for the extra using on line 7

michaelcfanning: Yes, of course! I tried using a more minimal qualifier, WorkItemFiling.Attachment but this won't work because 'WorkItemFiling' is a suffix on the current namespace. Your suggestion works thought. :)

---
In reply to: [377356810](https://github.com/microsoft/sarif-sdk/pull/1749#discussion_r377356810) [](ancestors = 377356810)

src/WorkItemFiling/WorkItemFiling.csproj
michaelcfanning: good point! will do this next.

---
In reply to: [377353664](https://github.com/microsoft/sarif-sdk/pull/1749#discussion_r377353664) [](ancestors = 377353664)

src/WorkItemFiling/AzureDevOpsClientWrapper.cs
michaelcfanning: Yes

---
In reply to: [377348570](https://github.com/microsoft/sarif-sdk/pull/1749#discussion_r377348570) [](ancestors = 377348570)

PR #1786, opened 2020-02-20
Multitool rewrite mode: Region performance, Guids

ScottLouvau:
- Add 'Guid' to OptionallyEmittedData and add support to add Result.Guids.
  - Fix rewrite add regions to not throw and fail on invalid regions.
  - Fix FileRegionsCache to keep previously read files in cache, up to a limit.
  - Added a simple LRU cache for FileRegionsCache to use (no .NET builtin without dependencies added)
   - Add tests for Cache, FileRegionsCache for invalid regions, Guid insertion, caching happening
   - Fix FileDiffingUnitTest output path to avoid path length limit problems (VS2019 uses an extremely long starting path for test binaries, and the Path.Combine was going over the ~240 char limit).


src/Sarif/Cache.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// <typeparam name="TKey">Type of Keys in collection</typeparam>
+    /// <typeparam name="TValue">Type of Values being cached in collection</typeparam>
+    public class Cache<TKey, TValue> where TKey : IComparable<TKey>
+    {
+        private readonly Func<TKey, TValue> _builder;
+        private readonly Dictionary<TKey, TValue> _cache;
+        private readonly LinkedList<TKey> _keysInUseOrder;
+
+        public const int DefaultCapacity = 100;
+        public int Capacity { get; private set; }
```
lgolding: >private set; [](start = 35, length = 12)

Since you set this only in the ctor, you don't need `private set`. (C# 6 feature: you can set a readonly property in the ctor.)

src/Sarif/Cache.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        ///  Returns the number of items currently cached
+        /// </summary>
+        public int Count => _cache.Count;
+
+        /// <summary>
+        ///  Returns the keys of items currently cached
+        /// </summary>
+        public IEnumerable<TKey> CachedKeys => _cache.Keys;
```
lgolding: IMO just `Keys` is a fine name.

src/Sarif/Cache.cs
```diff
@@ -0,0 +1,97 @@
...
+                return value;
+            }
+        }
+
+        /// <summary>
+        ///  Return whether the cache currently includes the value for the given key.
+        /// </summary>
+        /// <param name="key">Key to check cache for</param>
+        /// <returns>True if value in cache, False otherwise</returns>
+        public bool Contains(TKey key)
```
lgolding: >Contains [](start = 20, length = 8)

In keeping with the idea of using names from `Dictionary`, I suggest `ContainsKey`.

src/Sarif/FileRegionsCache.cs
```diff
@@ -15,20 +13,20 @@ namespace Microsoft.CodeAnalysis.Sarif
...
-        internal IFileSystem _fileSystem;
-
+        public const int DefaultCacheCapacity = 100;
         private readonly Run _run;
-        private readonly Dictionary<string, NewLineIndex> _filePathToNewLineIndexMap;
+        private readonly IFileSystem _fileSystem;
+        private readonly Cache<string, Tuple<string, NewLineIndex>> _cache;
 
-        public FileRegionsCache(Run run)
+        public FileRegionsCache(Run run, int capacity = DefaultCacheCapacity, IFileSystem fileSystem = null)
```
lgolding: >fileSystem [](start = 90, length = 10)

All fans of mocking thank you for this!

src/Sarif/FileRegionsCache.cs
```diff
@@ -274,23 +275,48 @@ private void PopulateCharLength(NewLineIndex lineIndex, Region region)
...
                 region.CharLength = charLength;
             }
-            SarifUtilities.DebugAssert(region.CharLength == charLength);
+            Assert(region.CharLength == charLength);
         }
 
         private NewLineIndex GetNewLineIndex(Uri uri, out string fileText)
         {
-            fileText = null;
+            Tuple<string, NewLineIndex> entry = _cache[uri.LocalPath];
```
lgolding: >_cache [](start = 48, length = 6)

This is so cool, it makes the entry if it wasn't there before! ... is this how caches like this usually work? It's so automagical I wonder if it's confusing... I only understood it because I'd _just_ read Cache.cs. Never mind, I'm probably overthinking this.

src/Sarif/FileRegionsCache.cs
```diff
@@ -274,23 +275,48 @@ private void PopulateCharLength(NewLineIndex lineIndex, Region region)
...
+         
+            fileText = entry.Item1;
+            return entry.Item2;
+        }
+
+        private Tuple<string, NewLineIndex> BuildIndexForFile(string localPath)
+        {
```
lgolding: It might be worth a function-level comment that this is the builder function for the cache.

This is so elegant, I love it.

src/Sarif/FileRegionsCache.cs
```diff
@@ -274,23 +275,48 @@ private void PopulateCharLength(NewLineIndex lineIndex, Region region)
...
+
+            return new Tuple<string, NewLineIndex>(fileText, index);
+        }
+
+        private static void Assert(bool condition)
+        {
+            // Placeholder to report issues in a situationally appropriate way.
+            //  We don't want Multitool rewrite to blow up.
+            //  We don't want unit tests for invalid Regions to block on asserts; we want to verify the code leaves those results alone.
+            //  We may want console or output log output when invalid Regions are detected during Multitool rewrite use.
```
lgolding: >use [](start = 116, length = 3)

Handling asserts in a situationally appropriate way... nice idea, I wonder that I've never seen it before. Perhaps you can inject an assert-handler function into this class:

```
private readonly Action<bool> _assertHandler;

FileRegionCache(... , Action<bool> assertHandler)
{
    ... 
    _assertHandler = assertHandler ?? ((bool condition) => Debug.Assert(condition));
}

...

private static void Assert(bool condition)
{
    _assertHandler(condition);
}
```

src/Test.UnitTests.Sarif/CacheTests.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.UnitTests
+{
+    public class CacheTests
+    {
+        [Fact]
+        public void CacheBasics()
+        {
+            int buildCount = 0;
+            Cache<int, int> cache = new Cache<int, int>(
+                (key) => { buildCount++; return key; },
```
lgolding: >key [](start = 48, length = 3)

Perhaps your builder could return, say, `100 * key` so that you can distinguish keys from values -- that is, you can be sure your code isn't using a key in a place where it meant to use a value.

src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
```diff
@@ -311,15 +312,49 @@ private static void ExecuteTests(ReadOnlyCollection<TestCaseData>  testCases, Fi
...
+
+            StringBuilder fileContents = new StringBuilder();
+            for (int i = 0; i < 1000; ++i)
+            {
+                fileContents.AppendLine("0123456789");
+            }
+
             var run = new Run();
-            var fileRegionsCache = new FileRegionsCache(run);
+            var mockFileSystem = MockFactory.MakeMockFileSystem(uri.LocalPath, fileContents.ToString());
```
chinarosesz: >var [](start = 12, length = 3)

How about replace with type here?

src/Sarif/Cache.cs
ScottLouvau: Huh. Ok, cool. I tried marking it "readonly" and it told me the language version wasn't late enough. I didn't think about trying to remove the setter, but that works.

src/Sarif/Cache.cs
ScottLouvau: Yes, making it more consistent with Dictionary is a good idea.

src/Sarif/Cache.cs
ScottLouvau: Also fixed.

src/Sarif/FileRegionsCache.cs
ScottLouvau: =) np

src/Sarif/FileRegionsCache.cs
ScottLouvau: Yes, this is the normal way, not my own idea. If you write a cache without this, the caller has to call TryGet and then load the value and then remember to stuff it in with a setter - the "self-loading" design avoids the risk of exactly the bug this class had before (building a cache and never putting anything in). =)

src/Sarif/FileRegionsCache.cs
ScottLouvau: Good idea. Added a comment here and where it's referenced in the ctor.

src/Sarif/FileRegionsCache.cs
ScottLouvau: Yes. I'm not sure where we want to land, so I just added the stub and redirected all of the Asserts so we didn't lose that verification logic.

src/Test.UnitTests.Sarif/CacheTests.cs
ScottLouvau: Good idea.

src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
ScottLouvau: Sure. I had moved this code up and left it alone, but happy to provide the resolved types.

src/Sarif/Cache.cs
lgolding: No, sorry, a `reaonly` property is one with only a getter. You can remove the private setter because it's ok to set such a property from the ctor even tho it has no setter.

---
In reply to: [382333251](https://github.com/microsoft/sarif-sdk/pull/1786#discussion_r382333251) [](ancestors = 382333251)

PR #1788, opened 2020-02-21
Fix #1785: Misspelled property in schema

### LINKED ISSUE #1785: Fix misspelled property name in schema
Per the spec, `suppression.state` should be `suppression.status`.

lgolding:
Per the spec, `suppression.state` should have been `suppression.status`.

Also:
- Mention Scott's new `Guid` value for inserting optional data in the `HelpText` for the `insert` option.
- Mention Scott's work in the release history.


michaelcfanning:
Please see my offline comments. I have a version of this change as well that also eliminates the obsolete schema reference from many test files. I suggest I try to merge our versions.

lgolding:
Agreed.

---
In reply to: [589459760](https://github.com/microsoft/sarif-sdk/pull/1788#issuecomment-589459760) [](ancestors = 589459760)

PR #1789, opened 2020-02-21
Suppression state to status

### LINKED ISSUE #1785: Fix misspelled property name in schema

michaelcfanning:
Resolves #1785. Rename `suppression.state` to `suppression.status`. Additional unit test 'ExercisesRtm5Changes.sarif' in prerelease transformer tests covers the new case. [FYI code flow isn't showing all changed files on my machine, including this one.]

@lgolding 
@nickmarston @chinarosesz FYI in case you'd like to see what it looks like to make a schema change (though honestly this could very well be the last one for 2.1)

### DIFF
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -41,9 +41,8 @@
     <!--
     Marking these items as content causes them to be included in the NuGet package.
     -->
-    <Content Include="Schemata\sarif-2.1.0-rtm.5.json" />
-    <Content Include="Schemata\sarif-external-property-file-2.1.0-rtm.5.json" />
-
+    <Content Include="Schemata\sarif-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
+    <Content Include="Schemata\sarif-external-property-file-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
     <!-- This item will be added to the NuGet package, however. -->
     <Content Include="..\ReleaseHistory.md" />
   </ItemGroup>
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -41,8 +41,8 @@
     <!--
     Marking these items as content causes them to be included in the NuGet package.
     -->
-    <Content Include="Schemata\sarif-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
-    <Content Include="Schemata\sarif-external-property-file-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
+    <Content Include="Schemata\sarif-2.1.0-rtm.5.json" />
+    <Content Include="Schemata\sarif-external-property-file-2.1.0-rtm.5.json" />
 
     <!-- This item will be added to the NuGet package, however. -->
     <Content Include="..\ReleaseHistory.md" />
--- a/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
+++ b/src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
@@ -245,17 +245,29 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
 
         private static bool ConvertSuppressionStateToSuppressionStatus(JObject sarifLog)
         {
-            string suppressionsPathToUpdate = "runs[].results[].suppressions[]";
+            bool modifiedLog = false;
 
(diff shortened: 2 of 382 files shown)


src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -79,9 +79,17 @@ public static class PrereleaseCompatibilityTransformer
                         case "http://json.schemastore.org/sarif-2.1.0-rtm.1":
                         case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.1.json":
                         {
-                            modifiedLog |= ApplyRtm2and3Changes(logObject);
```
michaelcfanning: >ApplyRtm2and3Changes [](start = 43, length = 20)

There were no transformer changes for 2 and 3, this helper just updated the schema. I've made this an explicitly named operation.

lgolding:
@michaelcfanning 

If CodeFlow is not showing your new file, is it possible you didn't `git add` it?

src/Sarif/Sarif.csproj
```diff
@@ -41,8 +41,8 @@
     <!--
     Marking these items as content causes them to be included in the NuGet package.
     -->
-    <Content Include="Schemata\sarif-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
-    <Content Include="Schemata\sarif-external-property-file-$(SchemaVersionAsPublishedToSchemaStoreOrg).json" />
+    <Content Include="Schemata\sarif-2.1.0-rtm.5.json" />
+    <Content Include="Schemata\sarif-external-property-file-2.1.0-rtm.5.json" />
```
lgolding: >sarif-external-property-file-2.1.0-rtm.5.json [](start = 31, length = 45)

I don't know why the changes to these lines happened -- they happened for me as well in my PR -- but please revert these two lines to continue using the symbol `$(SchemaVersionAsPublishedToSchemaStoreOrg)`.

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -219,9 +233,41 @@ public static class PrereleaseCompatibilityTransformer
             return transformer.SarifLog;
         }
 
-        private static bool ApplyRtm2and3Changes(JObject sarifLog)
+        private static bool ApplyRtm5Changes(JObject sarifLog)
+        {
+            bool modifiedLog = false;
+
+            // https://github.com/oasis-tcs/sarif-spec/issues/449
+            modifiedLog |= ConvertSuppressionStateToSuppressionStatus(sarifLog);
```
lgolding: Good factoring.

lgolding:
I don't see the file with your new unit test -- is that the file we're both talking about? Anyway, approved.

---
In reply to: [589856951](https://github.com/microsoft/sarif-sdk/pull/1789#issuecomment-589856951) [](ancestors = 589856951)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -219,9 +233,41 @@ public static class PrereleaseCompatibilityTransformer
             return transformer.SarifLog;
         }
 
-        private static bool ApplyRtm2and3Changes(JObject sarifLog)
+        private static bool ApplyRtm5Changes(JObject sarifLog)
```
chinarosesz: >JObject [](start = 45, length = 7)

Just a thought for future consideration: Loading JObject takes up more memory than System.Text.Document.JDocument. I did some perf profiling comparing the two in the past. For example, a 26MB json file, JObject allocates 393MB compared to only 54MB from JDocument.

src/Sarif/Sarif.csproj
michaelcfanning: File-based search and replace is what did it to me.

---
In reply to: [382834618](https://github.com/microsoft/sarif-sdk/pull/1789#discussion_r382834618) [](ancestors = 382834618)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
michaelcfanning: +1

---
In reply to: [382840623](https://github.com/microsoft/sarif-sdk/pull/1789#discussion_r382840623) [](ancestors = 382840623)

PR #1792, opened 2020-02-25
Set up CI with Azure Pipelines

chinarosesz:


chinarosesz:
Disregard. Just testing.

PR #1793, opened 2020-02-28
Fix #1791: ESLint formatter returns invalid SARIF in some cases

### LINKED ISSUE #1791: ESLint formatter returns invalid SARIF in some cases
Specifically, we do not always provide the rule ID, e.g.,

~~~
{
  "level": "error",
  "message": {
    "text": "Parsing error: ImportDeclaration should appear when the mode is ES6 and in the module context."
  },
  "locations": [
    {
      "physicalLocation": {
        "artifactLocation": {
          "uri": "file:///C:/Dev/SecureApp/SecureApp/ts/node_modules/esrecurse/gulpfile.babel.js",
          "index": 905
        }
      }
    }
  ]
},
~~~

lgolding:
At times, ESLint emits a message with no `ruleId`, indicating an internal error in ESLint. Represent this in SARIF as a `toolExecutionNotification` with `level: "error"`, rather than as a `result` with no `ruleId`, which is invalid.

This affects some unit tests, which now must always provide a `ruleId`.

Also:
- Per the spec, emit `run.results` even if the array is empty, as a positive indicator that the tool found no results. This affects the unit test which previously expected `run.results` to be `undefined` in this case.

### DIFF
--- a/src/ESLint.Formatter/sarif.js
+++ b/src/ESLint.Formatter/sarif.js
@@ -54,12 +54,19 @@ module.exports = function (results, data) {
 
     const sarifFiles = {};
     const sarifArtifactIndices = {};
-    var nextArtifactIndex = 0;
+    let nextArtifactIndex = 0;
     const sarifRules = {};
     const sarifRuleIndices = {};
-    var nextRuleIndex = 0;
+    let nextRuleIndex = 0;
     const sarifResults = [];
     const embedFileContents = process.env.SARIF_ESLINT_EMBED === "true";
+
+    // Emit a tool execution notification with this id if ESLint emits a message with
+    // no ruleId (which indicates an internal error in ESLint).
+    const internalErrorId = "ESL9999"
+
+    const toolExecutionNotifications = [];
+    let executionSuccessful = true;
 
     results.forEach(result => {
 
@@ -103,7 +110,7 @@ module.exports = function (results, data) {
 
                 result.messages.forEach(message => {
 
-                    const sarifResult = {
+                    const sarifRepresentation = {
                         level: getResultLevel(message),
                         message: {
                             text: message.message
@@ -121,7 +128,7 @@ module.exports = function (results, data) {
                     };
 
                     if (message.ruleId) {
-                        sarifResult.ruleId = message.ruleId;
+                        sarifRepresentation.ruleId = message.ruleId;
 
                         if (rulesMeta && typeof sarifRules[message.ruleId] === "undefined") {
                             const meta = rulesMeta[message.ruleId];
@@ -145,12 +152,29 @@ module.exports = function (results, data) {
                         }
 
                         if (sarifRuleIndices[message.ruleId] !== "undefined") {
-                            sarifResult.ruleIndex = sarifRuleIndices[message.ruleId];
+                            sarifRepresentation.ruleIndex = sarifRuleIndices[message.ruleId];
+                        }
+                    } else {
+                        // ESLint produces a message with no ruleId when it encounters an internal
+                        // error. SARIF represents this as a tool execution notification rather
+                        // than as a result, and a notification has a descriptor.id property rather
+                        // than a ruleId property.
+                        sarifRepresentation.descriptor = {
+                            id: internalErrorId
+                        };
+
+                        // As far as we know, whenever ESLint produces a message with no rule id,
+                        // it has severity: 2 which corresponds to a SARIF error. But check here
+                        // anyway.
+                        if (sarifRepresentation.level === "error") {
+                            // An error-level notification means that the tool failed to complete
+                            // its task.
+                            executionSuccessful = false
                         }
                     }
 
                     if (message.line > 0 || message.column > 0) {
-                        sarifResult.locations[0].physicalLocation.region = {
+                        sarifRepresentation.locations[0].physicalLocation.region = {
                             startLine: message.line,
                             startColumn: message.column
                         };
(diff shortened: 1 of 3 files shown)


lgolding:

lgolding:
@boAndron is added to the review. #Closed

src/ESLint.Formatter/sarif.js
```diff
@@ -54,13 +54,20 @@ module.exports = function (results, data) {
 
     const sarifFiles = {};
     const sarifArtifactIndices = {};
-    var nextArtifactIndex = 0;
+    let nextArtifactIndex = 0;
```
lgolding: >let [](start = 4, length = 3)

`let` is preferable because its block scoping behavior is more understandable to most people than `var`'s function body scoping.

src/ESLint.Formatter/sarif.js
```diff
@@ -145,12 +152,29 @@ module.exports = function (results, data) {
...
-                            sarifResult.ruleIndex = sarifRuleIndices[message.ruleId];
+                            sarifRepresentation.ruleIndex = sarifRuleIndices[message.ruleId];
+                        }
+                    } else {
+                        // ESLint produces a message with no ruleId when it encounters an internal
+                        // error. SARIF represents this as a tool execution notification rather
+                        // than as a result, and a notification has a descriptor.id property rather
+                        // than a ruleId property.
+                        sarifRepresentation.descriptor = {
+                            id: internalErrorId
```
michaelcfanning: >internalErrorId [](start = 32, length = 15)

What kind of internal errors are we talking about? If the internal error is something like: you passed us code to analyze that isn't syntactically correct, that would be a user configuration error, not an internal error. Some tools seem to mix these concepts together. A tool notification error is intended to convey 'we should have been able to scan for you, but an issue in the engine itself prevented us from succeeding'. A configuration error is 'we received inputs or configuration by the user which prevented us from completing a scan'.

src/ESLint.Formatter/sarif.js
```diff
@@ -178,8 +207,17 @@ module.exports = function (results, data) {
         });
     }
 
-    if (sarifResults.length > 0) {
-        sarifLog.runs[0].results = sarifResults;
+    // Per the SARIF spec §3.14.23, run.results must be present even if there are no results.
+    // This provides a positive indication that the run completed and no results were found.
+    sarifLog.runs[0].results = sarifResults;
```
michaelcfanning: it is acceptable for run.results to be null in cases where a catastrophic engine error prevented analysis from occurring. are you permitting this case?

e.g., if a user passes a single syntactically invalid file to analyze, run.results may be null (in cases where a scanner cannot function against syntactically incorrect code).

src/ESLint.Formatter/sarif.js
```diff
@@ -54,13 +54,20 @@ module.exports = function (results, data) {
...
     const sarifRules = {};
     const sarifRuleIndices = {};
-    var nextRuleIndex = 0;
+    let nextRuleIndex = 0;
     const sarifResults = [];
     const embedFileContents = process.env.SARIF_ESLINT_EMBED === "true";
 
+    // Emit a tool execution notification with this id if ESLint emits a message with
+    // no ruleId (which indicates an internal error in ESLint).
+    const internalErrorId = "ESL9999"
```
michaelcfanning: >ESL9999 [](start = 29, length = 7)

Suggest using ESL0999 for this. A poorly adopted convention with MS compilers is for the 1000 - 1999 range to be preserved for core compiler/language errors and everything beyond is for additional analysis. IDs below 1000 are reserved for internal engine issues.

src/ESLint.Formatter/sarif.js
michaelcfanning: "If the tool started but failed to begin its analysis (for example, because its command line was invalid), then again results MAY be present, and if present SHALL be null."

---
In reply to: [385957192](https://github.com/microsoft/sarif-sdk/pull/1793#discussion_r385957192) [](ancestors = 385957192)

lgolding: As we discussed offline: the engine did not fail to begin its analysis; it can produce real results before (and after) hitting this problem.

---
In reply to: [385958991](https://github.com/microsoft/sarif-sdk/pull/1793#discussion_r385958991) [](ancestors = 385958991,385957192)

src/ESLint.Formatter/sarif.js
lgolding: Fixed.

---
In reply to: [385957944](https://github.com/microsoft/sarif-sdk/pull/1793#discussion_r385957944) [](ancestors = 385957944)

src/ESLint.Formatter/sarif.js
lgolding: I'm still researching the root cause, but it appears to be ESLint's parsing engine (specifically, a component called the "scope analyzer") choking on a construct like this:
```
import Example from 'example';
```
This is valid starting in ES 2015. You could argue that this is a configuration error because the ESLint user didn't set that option. But

1. I'm still investigating, and
2. Even if that turned out to be the case for this particular error, it's not clear how the SARIF generator could look at an ESLint message-without-a-ruleId and determine in general whether it's a tool failure or a configuration error.

I mean, maybe I can actually look at the message and, if I recognize the string as a known configuration issue, I can put it in the tool configuration notifications, otherwise in the tool execution notifications.

But it's still early in my research. Don't take any of what I just said too seriously yet.

This issue was reported in November 2015 in [[https://github.com/eslint/eslint/issues/4344](https://github.com/eslint/eslint/issues/4344)](https://github.com/eslint/eslint/issues/4344).


---
In reply to: [385956543](https://github.com/microsoft/sarif-sdk/pull/1793#discussion_r385956543) [](ancestors = 385956543)

lgolding: Per our discussion: we will record these as tool configuration notifications for now, and record the question of how to distinguish tool configuration notifications from tool execution notifications in [https://github.com/Microsoft/sarif-sdk/issues/1798](https://github.com/Microsoft/sarif-sdk/issues/1798). @mikefan@microsoft.com FYI.

---
In reply to: [385975793](https://github.com/microsoft/sarif-sdk/pull/1793#discussion_r385975793) [](ancestors = 385975793,385956543)

PR #1794, opened 2020-03-02
Adding ExtractAllLocationUri's visitor

nickmarston:
Adds a visitor to create a HashSet of every ArtifactLocation that is present in a SARIF file and resolves those URIs to their index location.

This will be needed for bug filing, where the URIs in a SARIF location will be used as the key to indexing services.  I will next be looking at the best place to transform this artifact location list into a list of unique URIs, normalized, which I can then pass to the needed service.

### DIFF
--- a/src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
+++ b/src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT        
+// license. See LICENSE file in the project root for full license information. 
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+
+    public class ExtractAllArtifactLocationsVisitor : SarifRewritingVisitor
+    {
+        private Run _currentRun;
+        public HashSet<ArtifactLocation> allArtifactLocations { get; private set; }
+
+        public ExtractAllArtifactLocationsVisitor()
+        {
+            allArtifactLocations = new HashSet<ArtifactLocation>(ArtifactLocation.ValueComparer);
+        }
+
+        public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
+        {
+            if (node.Uri == null)
+            {
+                return node;
+            }
+            allArtifactLocations.Add(node.Resolve(_currentRun));
+            return node;
+        }
+
+        public override Run VisitRun(Run node)
+        {
+            _currentRun = base.VisitRun(node);
+            return node;
+        }
+    }
+}
--- a/src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
+++ b/src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
@@ -11,12 +11,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
     /// This will extract every location present in the log file, which may then need to be filtered by a
     /// downstream consumer.
     /// 
-    /// May optionally be based a run at contsruction, which allows the visitor to later visit only specific parts of
-    /// the sarif log where location data aggregation is desired.
+    /// User may optionally set a specific run before invoking, in order to only fetch artifact locations
+    /// from a subset on an entire SARIF file.
     /// </summary>
     public class ExtractAllArtifactLocationsVisitor : SarifRewritingVisitor
     {
-        private Run _currentRun;
+        private Run CurrentRun { get; set; }
         public HashSet<ArtifactLocation> AllArtifactLocations { get; private set; }
 
         public ExtractAllArtifactLocationsVisitor()
@@ -24,25 +24,16 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             AllArtifactLocations = new HashSet<ArtifactLocation>(ArtifactLocation.ValueComparer);
         }
 
-        public ExtractAllArtifactLocationsVisitor(Run node)
-        {
-            AllArtifactLocations = new HashSet<ArtifactLocation>(ArtifactLocation.ValueComparer);
-            _currentRun = node;
-        }
-
         public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
         {
-            if (node.Uri == null)
-            {
-                return node;
-            }
-            AllArtifactLocations.Add(node.Resolve(_currentRun));
+            AllArtifactLocations.Add(node.Resolve(CurrentRun));
             return node;
         }
 
         public override Run VisitRun(Run node)
         {
-            _currentRun = base.VisitRun(node);
+            CurrentRun = node;
+            CurrentRun = base.VisitRun(node);
             return node;
         }
     }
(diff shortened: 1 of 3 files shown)


src/Test.Utilities.Sarif/TestConstants.cs
```diff
@@ -28,6 +28,14 @@ public static class RuleIds
             public const string Rule10 = "TST0010";
         }
 
+        public static class FileLocations
+        {
+            public const string Location1 = "C:\\Test\\Data\\File1.sarif";
```
lgolding: >" [](start = 44, length = 1)

Nit: consider using verbatim constants `@"C:\Test\Data\File.sarif"`.

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    public class ExtractAllArtifactLocationsVisitor : SarifRewritingVisitor
+    {
+        private Run _currentRun;
+        public HashSet<ArtifactLocation> allArtifactLocations { get; private set; }
+
+        public ExtractAllArtifactLocationsVisitor()
+        {
+            allArtifactLocations = new HashSet<ArtifactLocation>(ArtifactLocation.ValueComparer);
```
lgolding: >allArtifactLocations [](start = 12, length = 20)

It's not clear to me whether you want to reset this at the start of each run (at the start of `VisitRun`) or accumulate across runs, as the code stands. You'll need to consider the expected use case, and whatever the answer, you'll need to document it in the class summary comment.

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+
+    public class ExtractAllArtifactLocationsVisitor : SarifRewritingVisitor
+    {
+        private Run _currentRun;
+        public HashSet<ArtifactLocation> allArtifactLocations { get; private set; }
```
lgolding: >a [](start = 41, length = 1)

Convention: All our `public` property names are CamelCase.

src/Test.UnitTests.Sarif/Visitors/ExtractAllArtifactionLocationsVisitorTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            };
+
+            var visitor = new ExtractAllArtifactLocationsVisitor();
+            visitor.VisitSarifLog(sarifLog);
+
+            foreach (Location location in sarifLog.Runs[0].Results[0].Locations)
+            {
+                visitor.allArtifactLocations.Contains(location.PhysicalLocation.ArtifactLocation).Should().BeTrue();
+            }
+        }
```
lgolding: >} [](start = 8, length = 1)

Do you want to accumulate URIs that occur in other places, such as `Result.RelatedLocations`? That's what the code does today. If you don't want that, the visitor will need to maintain state ("I'm in `Result.Locations`!"). Again, it depends on your use case, and again, either way, you'll need to document the decision. Keep in mind that this is a general purpose SDK; what's right your your use case might not be appropriate in general.

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
nickmarston: Currently, I expect to get all locations in a given set of runs.  I'll add the documentation confirming this.  I'd be curious if you see the reset version being the more common use case, outside of the bug filer.

---
In reply to: [386741455](https://github.com/microsoft/sarif-sdk/pull/1794#discussion_r386741455) [](ancestors = 386741455)

src/Test.UnitTests.Sarif/Visitors/ExtractAllArtifactionLocationsVisitorTests.cs
nickmarston: I tend to think the more useful SDK piece is to extract everything.  Downstream, the filer is going to need to take these extracted locations and filter on them in some way.  Given, however, that ArtifactLocation itself doesn't seem to contain any of the data necessary to make that decision as to where the node came from, it sounds like I may need to store that data.

I need to think on this a bit.  The filer definitely only cares about the locations extracted from the Results.Locations collection, as those are what we want to resolve into area paths for bugs.

I obviously can change the visitor to only hit the nodes I care about (probably at the Results.Location level), but I'm not sure that fits as well.

---
In reply to: [386742798](https://github.com/microsoft/sarif-sdk/pull/1794#discussion_r386742798) [](ancestors = 386742798)

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    /// <summary>
+    /// A visitor that examines a specified SARIF log to extract all artifact locations for later processing.
+    /// This will extract every location present in the log file, which may then need to be filtered by a
+    /// downstream consumer.
+    /// 
+    /// May optionally be based a run at contsruction, which allows the visitor to later visit only specific parts of
```
michaelcfanning: 'be configured by'

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// the sarif log where location data aggregation is desired.
+    /// </summary>
+    public class ExtractAllArtifactLocationsVisitor : SarifRewritingVisitor
+    {
+        private Run _currentRun;
```
michaelcfanning: >_currentRun [](start = 20, length = 11)

Suggestion: convert this to a writable public property, CurrentRun. Remove the ctor. Set this property on VisitRun. Now users can set this and visit interior items. But the class always stays consistent. Don't forget to update the class summary if you accept this suggestion.

src/Sarif/Visitors/ExtractAllArtifactLocationsVisitor.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public ExtractAllArtifactLocationsVisitor(Run node)
+        {
+            AllArtifactLocations = new HashSet<ArtifactLocation>(ArtifactLocation.ValueComparer);
+            _currentRun = node;
+        }
+
+        public override ArtifactLocation VisitArtifactLocation(ArtifactLocation node)
+        {
+            if (node.Uri == null)
```
michaelcfanning: >node [](start = 16, length = 4)

Ditch this conditional. An artifact location may only include an index, which points to the artifact table where the non-null URI is. Add a unit test for this case!

src/Test.UnitTests.Sarif/Visitors/ExtractAllArtifactionLocationsVisitorTests.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        [Fact]
+        public void ExtractAllArtifactLocationsVisitor_ExtractsMultipleLocations()
+        {
+            var sarifLog = new SarifLog
+            {
+                Runs = new[]
+                {
+                    new Run
```
michaelcfanning: >Run [](start = 24, length = 3)

Add an artifacts table with one item at least. Add a result location that refers to this only by index. Add a second case that has both an index and URI.

PR #1797, opened 2020-03-04
First pass on applying editorconfig

chinarosesz:
Some core changes to the formatting after using editorconfig are:
* Replace var with explicit type when type is not apparent from the right side of the assignment
* Place System.* directives first
* Remove unused locals
* Use "this." to reference members and fields 

Next: Remove underscores in private fields if "this." usage is a comfortable adoption.


src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+
+# Parameter preferences
+dotnet_code_quality_unused_parameters = all:error
+
+#### C# Coding Conventions ####
+
+# var preferences
+csharp_style_var_elsewhere = false:error
+csharp_style_var_for_built_in_types = false:error
+csharp_style_var_when_type_is_apparent = true:suggestion
```
lgolding: >apparent [](start = 30, length = 8)

What does this cover? I assume it's:
- Casts: `var x = (T)expr;`
- Constructors: `var x = new T();`
- `as` operator: `var x = x as T;`

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+
+# Expression-bodied members
+csharp_style_expression_bodied_accessors = true:suggestion
+csharp_style_expression_bodied_constructors = false:suggestion
+csharp_style_expression_bodied_indexers = true:suggestion
+csharp_style_expression_bodied_lambdas = true:suggestion
+csharp_style_expression_bodied_local_functions = false:suggestion
+csharp_style_expression_bodied_methods = false:suggestion
```
lgolding: >false:suggestion [](start = 41, length = 16)

I really like expression-bodied methods, _e.g._
```
private static string MakeRuleIdFromRuleNumber(int ruleNumber)
    => $"WS{ruleNumber:0000}`;
```

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+csharp_style_expression_bodied_operators = false:suggestion
+csharp_style_expression_bodied_properties = true:suggestion
+
+# Pattern matching preferences
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
+csharp_style_pattern_matching_over_is_with_cast_check = true:error
+csharp_style_prefer_switch_expression = true:suggestion
```
lgolding: >prefer_switch_expression [](start = 13, length = 24)

I assume this means "prefer switch to if-else if-else when the comparands are constants". I assume a simple "if-else" is ok, right?

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+
+# Null-checking preferences
+csharp_style_conditional_delegate_call = true:suggestion
```
lgolding: >conditional_delegate_call [](start = 13, length = 25)

Huh, I have no idea what this means.

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+
+# Modifier preferences
+csharp_prefer_static_local_function = true:suggestion
```
lgolding: >prefer [](start = 7, length = 6)

Prefer "static local function" to what?

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async
+
+# Code-block preferences
+csharp_prefer_braces = true:warning
+csharp_prefer_simple_using_statement = true:silent
+
+# Expression-level preferences
+csharp_prefer_simple_default_expression = true:suggestion
```
lgolding: >csharp_prefer_simple_default_expression [](start = 0, length = 39)

I assume this means "Prefer `T x = default` to `T x = default(T)`.

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+
+#### Naming styles ####
+
+# Naming rules
+
+dotnet_naming_rule.interface_should_be_begins_with_i.severity = suggestion
+dotnet_naming_rule.interface_should_be_begins_with_i.symbols = interface
+dotnet_naming_rule.interface_should_be_begins_with_i.style = begins_with_i
+
+dotnet_naming_rule.types_should_be_pascal_case.severity = suggestion
```
lgolding: >suggestion [](start = 58, length = 10)

Why only a suggestions? When is this ok?

src/.editorconfig
```diff
@@ -0,0 +1,197 @@
...
+# Code-block preferences
+csharp_prefer_braces = true:warning
+csharp_prefer_simple_using_statement = true:silent
+
+# Expression-level preferences
+csharp_prefer_simple_default_expression = true:suggestion
+csharp_style_deconstructed_variable_declaration = true:silent
+csharp_style_inlined_variable_declaration = true:silent
+csharp_style_pattern_local_over_anonymous_function = true:suggestion
+csharp_style_prefer_index_operator = true:silent
```
lgolding: >csharp_style_prefer_index_operator [](start = 0, length = 34)

In general I don't know what all of these "prefers" mean (prefer index operator to what? prefer range operator to what?)

src/Sarif/SarifUtilities.cs
```diff
@@ -168,15 +168,15 @@ public static int GetByteLength(char[] chars, Encoding encoding)
...
             return GetByteLength(s, encoding);
         }
 
         public static int GetByteLength(string s, Encoding encoding)
         {
             s = s ?? throw new ArgumentNullException(nameof(s));
             encoding = encoding ?? throw new ArgumentNullException(nameof(encoding));
-            
+
             byte[] bytes = encoding.GetBytes(s);
```
lgolding: >b [](start = 12, length = 1)

Do you know which rule inserted the blank line above? I don't object to it.

src/Sarif/Baseline/SarifLogBaseliner.cs
```diff
@@ -8,31 +8,31 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline
...
-                newResult.BaselineState = 
-                    baseLine.Results.Contains(result, ResultComparator) ? BaselineState.Unchanged : BaselineState.New;
+                newResult.BaselineState =
+                    baseLine.Results.Contains(result, this.ResultComparator) ? BaselineState.Unchanged : BaselineState.New;
 
                 differencedRun.Results.Add(newResult);
             }
 
-            foreach (var result in baseLine.Results)
+            foreach (Result result in baseLine.Results)
```
lgolding: >Result [](start = 21, length = 6)

yay

src/Sarif/VersionOne/Readers/AnnotatedCodeLocationIdConverter.cs
```diff
@@ -29,7 +29,7 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
             // when the input is a string that parses as a valid integer, such as "1". In
             // that case, objectType is Int32, but the object returned by reader.Value is
             // a string!
-            var stringValue = value as string;
+            string stringValue = value as string;
```
lgolding: >string [](start = 12, length = 6)

Huh, that answers I question I had in the editorconfig file. The `as` operator does _not_ make the type "apparent", and it wants an explicit type here (where I'd've been happy with `var`). No matter, it's fine.

src/.editorconfig
chinarosesz: Correct because the type is apparent on the right side for all three of these cases.

---
In reply to: [388463418](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388463418) [](ancestors = 388463418)

src/Sarif/VersionOne/Readers/AnnotatedCodeLocationIdConverter.cs
chinarosesz: Sorry I was testing to see if the suggestion to use var shows up by replacing it with explicit type on the left side. I'll revert this.

---
In reply to: [388500810](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388500810) [](ancestors = 388500810)

chinarosesz: Looks like as operator is not treated as obvious. 

I was testing var usage with the as operator and at some point, I thought the IDE didn't complain and had made the assumption that var usage is ok with the as operator, but looks like that's not the case.

---
In reply to: [388500810](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388500810) [](ancestors = 388500810)

src/Sarif/SarifUtilities.cs
chinarosesz: I'm not sure, looks like it inserted a line and deleted a line, resulting in no change from what I see, so I think we are fine.

---
In reply to: [388483612](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388483612) [](ancestors = 388483612)

src/.editorconfig
chinarosesz: This rule is around the new switch expression syntax, so it won't have any effect if you use a simple if/else statement. 

Since this is the default generated from VS, I'll keep as is. We can always silent it if no one likes the new switch expression syntax. I personally think it's a little too difficult to get used to. 

More about this syntax can be found in the below links, introduced in language version 8.0+. 

[https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-language-conventions?view=vs-2019#csharp_style_prefer_switch_expression](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-language-conventions?view=vs-2019#csharp_style_prefer_switch_expression)

[https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions)

---
In reply to: [388466187](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388466187) [](ancestors = 388466187)

src/.editorconfig
chinarosesz: Another syntax shortening for null check. Default is set to true for suggestion in VS, we can turn it off as needed.

// csharp_style_conditional_delegate_call = true
func?.Invoke(args);

// csharp_style_conditional_delegate_call = false
if (func != null) { func(args); }

---
In reply to: [388466336](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388466336) [](ancestors = 388466336)

src/.editorconfig
chinarosesz: This is another default for editorconfig. I agree that these names are difficult to understand and sometimes I have to look them up. 

An example:
// csharp_prefer_static_local_function = true
void M()
{
    Hello();
    static void Hello()
    {
        Console.WriteLine("Hello");
    }
}

// csharp_prefer_static_local_function = false
void M()
{
    Hello();
    void Hello()
    {
        Console.WriteLine("Hello");
    }
}


---
In reply to: [388477979](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388477979) [](ancestors = 388477979)

src/.editorconfig
chinarosesz: Yes

---
In reply to: [388478558](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388478558) [](ancestors = 388478558)

src/.editorconfig
chinarosesz: The index operator can be used when calculating an index from the end of the collection. It's a new language feature in C# 8.0. Currently set to silent for now. These rules are generated when creating an editor config to ensure the repository picks them up from here rather than from someone's default VS configuration. 

To be honest, I can't even keep up with these new syntax shortening tricks that are introduced, I think it's harder to read.

---
In reply to: [388482218](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388482218) [](ancestors = 388482218)

src/.editorconfig
chinarosesz: Agreed that type should be PascalCase always, this is currently a VS default. Keeping as is and we can iterate more on this config in the future.

---
In reply to: [388480705](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388480705) [](ancestors = 388480705)

src/.editorconfig
chinarosesz: Makes sense, if not one objects here I'll put it back to its default which is false:silient that way if you have the expression-bodied methods, it won't suggest you to convert to the longer version.

---
In reply to: [388465079](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r388465079) [](ancestors = 388465079)

src/.editorconfig
lgolding: Oh gosh, _casting_ isn't apparent? That's lame. No matter, let's leave this alone.

---
In reply to: [389053631](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r389053631) [](ancestors = 389053631,388463418)

chinarosesz:
>     [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.62.0.0")]

Should I tell editorconfig to leave these auto generated files alone?

---
Refers to: src/Sarif/Autogenerated/Address.cs:17 in ec123a2. [](commit_id = ec123a225a7674b55822e17cc9a51898a5cf7a8b, deletion_comment = False)

michaelcfanning:
>     [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "0.62.0.0")]

Yes, or we need to fix jschema not to introduce this unncessary namespace. @lgolding, do you happen to recall the details of how this unnecessary thing is getting introduced? My guess is that we scrape the namespace from somewhere in a codegenhints reference and it's being overapplied.



---
In reply to: [596013157](https://github.com/microsoft/sarif-sdk/pull/1797#issuecomment-596013157) [](ancestors = 596013157)

---
Refers to: src/Sarif/Autogenerated/Address.cs:17 in ec123a2. [](commit_id = ec123a225a7674b55822e17cc9a51898a5cf7a8b, deletion_comment = False)

src/Sarif/Core/Artifact.cs
```diff
@@ -40,13 +40,13 @@ public partial class Artifact : ISarifNode
             // occurred. Something to discuss moving forward.
             try
             {
-                bool workTodo = dataToInsert.HasFlag(OptionallyEmittedData.Hashes)    ||
```
michaelcfanning: >     [](start = 82, length = 4)

As you can see, I use whitespace often to line up boolean operators for better readability. Wonder if we can dial this one back. My care factor is low, though. :)

src/Sarif/VersionOne/Readers/AnnotatedCodeLocationIdConverter.cs
michaelcfanning: No worries. We can zap most 'as' usage anyway with something like.

if (value is string stringValue)

---
In reply to: [389063699](https://github.com/microsoft/sarif-sdk/pull/1797#discussion_r389063699) [](ancestors = 389063699,388500810)

PR #1804, opened 2020-03-07
Ado yaml build

michaelcfanning:
This change enables an AzureDevOps public + test validation pipeline. In order to make this work, I was required to clean up some inconsistent application of the TargetFrameworks variable. This also reduces references to a more manageable number (so we can update to .net core 3.1, which comes soon).

The ADO build/test environment runs in a directory perilously close to MAX_PATH in length. As a result, I was required to shorten names of four test files in the validation tests that happened to spill over MAX_PATH (due to use of the Driver FileSpecifier class).


michaelcfanning:
@lgolding  @ScottLouvau  FYI. Doing some offline work to get all our broken checks working. 

The microsoft.sarif-sdk ADO pipeline is building and testing for net461 now. Looks like it will need some tinkering to automatically trigger during PR, this isn't working quite yet.

Next stop is to restore the Semmle builds, however.

PR #1805, opened 2020-03-07
Restore LGTM Code QL analysis

michaelcfanning:
Helpful colleagues offline have provided advice on getting LGTM to function. Applying it here.


michaelcfanning:
Build failures  master LGTM yml needs to be right. Merging and will test master condition afterwards.

PR #1807, opened 2020-03-09
Baselining New Entities Fixes

ScottLouvau:
Baselining Refinements based on analysis of New Entities data and "well mannered SARIF" review.

Overall, this work fixes:
  - Results were overmatched on a partialFingerprint with a constant value.
    [Larry] IOW, if a partial fingerprint has the same value for every result, it shouldn't participate in the majority voting logic that decides whether two results match, because it has no discriminatory power.
  - Results were undermatched with unique partialFingerprints, but only unique per file.
    [Larry] IOW, if you have a fingerprint that's unique per file, but the same for every result in a file, you don't want to ignore it -- because then you'll fail to detect two results that are really the same, but the file they're in was renamed.
  - Results undermatched with minor reorderings within a file.

Changes:
  - Match Category on DeprecatedRuleIds, if found.
  - Match Regions on omitted property inferred values correctly.
    [Larry] For example, `"region": { "startLine" 42 } ` matches `"region": { "startLine": 42, "startColumn": 1, "endLine": 42 }`
  - Match unique 'what' properties by URI, using first common URI across logs.
  - Add 'TrustMap' and compute trust score per partialFingerprint to ensure we trust fingerprints selectively based on how unique the values are and how often they match in the two logs.
  - Expand 'Adjacent' match phase to 'Nearby' results with threshold to match well for minor reorderings.
  - Add code to remove Guids from results in RemoveOptionalDataVisitor. (gap in add GUIDs work, earlier)
  - BaselineE2E: Delete old output folder to ensure Output is just the latest run.
  - BaselineE2E: Fix performance problem with generating detail log by finding Results by Guid via a Dictionary. 
  - BaselineE2E: Write URI suffix (not just file name) in detail logs, so multi-part URIs (prThread/<PRID>/<ThreadID>) are understandable.
  - Unit Tests for new behaviors.


src/build.props
```diff
@@ -24,8 +24,8 @@
       NOTE: Version in scripts/BuildMultitoolForNpm.ps1 also has to be updated for each release
       until it adopts the same versions as the SDK itself.
     -->
-    <VersionPrefix>2.2.0</VersionPrefix>
-    <PreviousVersionPrefix>2.1.25</PreviousVersionPrefix>
+    <VersionPrefix>2.2.1</VersionPrefix>
```
lgolding: >2 [](start = 21, length = 1)

IMO we should start being more [SemVer]([https://semver.org/)-friendly](https://semver.org/)-friendly). This is a backward-compatible feature improvement: a minor version change to `2.3.0`. It's not just a backward-compatible bug fix.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v2.2.1** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.1) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.1)
+* FEATURE: Multitool `remove` option now supports `Guids` value to remove `Result.Guid`.
+* FEATURE: Significant Baselining algorithm improvements: dynamic partialFingerprint trust, location-specific unique what property matching, 'nearby' matching, correct omitted Region property handling, correct ReportingDescriptor.DeprecatedIds handling.
```
lgolding: >partialFingerprint [](start = 66, length = 18)

Ultra-nit: Markdown backticks around code elements, as you did on the previous line.

src/Sarif/Baseline/V2/TrustMap.cs
```diff
@@ -0,0 +1,177 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Baseline
+{
+    /// <summary>
+    ///  TrustMap determines how much to trust different Result attributes for matching.
+    ///  
+    ///  Trust is the product of how often values match and how unique the values were.
+    ///  An attribute which always had the same value will have very low trust.
```
lgolding: >had [](start = 35, length = 3)

had => has on this line, OR matches => matched on the next line. (The present tense sounds better to me.)

src/Sarif/Baseline/V2/TrustMap.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        /// <summary>
+        ///  Add a Result attribute to the map.
+        /// </summary>
+        /// <param name="component">WhatComponent for a Result attribute used in baselining</param>
+        public void Add(WhatComponent component)
+        {
+            TrustKey key = new TrustKey(component);
+
+            TrustValue value = null;
```
lgolding: > = null [](start = 28, length = 7)

Nit: `= null` unnecessary because `out` parameter is guaranteed to be assigned. But even better, you can now declare the `out` parameter inline:
```
if (!Map.TryGetValue(key, out TrustValue value))
```

src/Test.EndToEnd.Baselining/BaseliningTester.cs
```diff
@@ -316,5 +328,11 @@ private static void SortForBaselining(Run run)
             List<Result> results = (List<Result>)run.Results;
             results.Sort(DirectResultMatchingComparer.Instance);
         }
+
+        private static void DeleteAndCreateDirectory(string directoryPath)
+        {
+            if (Directory.Exists(directoryPath)) { Directory.Delete(directoryPath, true); }
```
lgolding: >true [](start = 83, length = 4)

Ultra-nit: Since `bool` parameters aren't self-documenting, I like to use named parameters for them:
```
Directory.Delete(directoryPath, recursive: true);
```

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
```diff
@@ -62,7 +69,8 @@ public void Write(SarifLog newBaselineLog, SarifLog baselineLog, SarifLog curren
                     case BaselineState.Unchanged:
                     case BaselineState.Updated:
                         // Find and write old Result from previous Baseline (to get pre-merged properties)
-                        Result previousResult = baselineLog.FindByGuid(result.CorrelationGuid);
+                        Result previousResult = null;
+                        baselineResultsByGuid.TryGetValue(result.CorrelationGuid, out previousResult);
```
lgolding: >TryGetValue [](start = 46, length = 11)

Ultra-nit: Now that you're using `TryGetValue`, it's more idiomatic to test on its return value:
```
if (!baselineResultsByGuid.TryGetValue(result.CorrelationGuid, out Result previousResult)
```

src/Test.UnitTests.Sarif/Visitors/RemoveOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            Runs = new List<Run>()
+            {
+                new Run()
+                {
+                    Results = new List<Result>()
+                    {
+                        new Result()
+                        {
+                            Guid = Guid.NewGuid().ToString(SarifConstants.GuidFormat)
```
lgolding: >GuidFormat [](start = 74, length = 10)

Yay!! Not everyone knows about this constant :-)

src/Test.UnitTests.Sarif/Visitors/RemoveOptionalDataVisitorTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        };
+
+        [Fact]
+        public void RemoveGuids()
+        {
+            SarifLog log = SampleLog.DeepClone();
+
+            RemoveOptionalDataVisitor v = new RemoveOptionalDataVisitor(OptionallyEmittedData.None);
+            v.Visit(log);
+            Assert.NotNull(log.Runs[0].Results[0].Guid);
```
lgolding: >Assert [](start = 12, length = 6)

Code base convention: Use FluentAssertions:
```
log.Runs[0].Results[0].Guid.Should().NotBeNull();
```

lgolding: Oh -- I see that all of these baselining tests use the XUnit assertion library. Never mind (but do consider addressing this on a rainy day.)

---
In reply to: [390021991](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390021991) [](ancestors = 390021991)

src/build.props
ScottLouvau: Sure - fixed.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -53,6 +81,19 @@ public bool MatchesAnyWhat(ExtractedResult other)
...
 
+        /// <summary>
+        ///  Match enough of the 'What' properties of two ExtractedResults (Guid, Fingerprints, PartialFingerprints, Snippets, Message, Properties).
+        ///  A match in high-confidence identity properties is a match (Guid, Fingerprint, >= 50% of PartialFingerprint).
+        ///  A non-match in high-confidence identity properties is a non-match (Fingerprint, 0% of PartialFingerprints, Properties).
+        ///  Otherwise, Results match if Message and first Snippet match.
+        /// </summary>
+        /// <param name="other">ExtractedResult to match</param>
+        /// <returns>True if *any* 'What' property matches, False otherwise</returns>
+        internal bool MatchesAnyWhat(ExtractedResult other, TrustMap trustMap)
```
lgolding: >trustMap [](start = 69, length = 8)

Add to XML comment.

src/Sarif/Baseline/V2/TrustMap.cs
ScottLouvau: Fixed

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
lgolding: >trustMap [](start = 69, length = 8)

How about defaulting `trustMap` to `null` and not having to add an overload?

src/Sarif/Baseline/V2/TrustMap.cs
ScottLouvau: Yes. I sometimes have compound conditions and then the method doesn't always fire and the compiler requires the default assignment. I like to leave it there just to be clear to the reader what value it will have if for some reason the call doesn't happen.

The sneakiest are conditions like this:
TrustValue value = null;
if (!Map?.TryGetValue(key, out value)) {
 ...
}

Here, thanks to ?., if Map is null, the method will not get called at all.

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
ScottLouvau: Left alone - the reason I wrote this as a statement is that I want to be clear the dictionary lookup is a step, and not a side-effect free conditional. I personally find it confusing when conditional expressions have side effects which the later code requires, so re-ordering the conditional blocks (for example) will cause bugs.

When it's a single condition if statement (if (!Map.TryGetValue(key, out value)) { value = new().. }) I tend to leave them inline, but when there are multiple conditions I don't like them. That's the rationale, anyway.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
ScottLouvau: Fixed.

ScottLouvau: I had to have separate overloads because TrustMap, WhatMap, and WhatComponent are internal types, and I didn't really want to make them all public. If I make MatchesAnyWhat internal, however, it'll be a breaking change.

src/Test.UnitTests.Sarif/Baseline2/TrustMapTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            // Match the two runs to allow trust scoring based on matches
+            Assert.False(beforeMap.WasMatched);
+            Assert.False(afterMap.WasMatched);
+            afterMap.CountMatchesWith(beforeMap);
+            Assert.True(beforeMap.WasMatched);
+            Assert.True(afterMap.WasMatched);
+
+            // After comparing runs, trust is based on uniqueness and number of matches
+            // Note: Constants have no trust, unique and no match have low (default) trust.
```
lgolding: >unique [](start = 46, length = 6)

"non-unique"?

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
lgolding: Oh, I see -- you can't have a parameter of non-public type on a public API. I see now you've made the new overloads internal.

---
In reply to: [390454469](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390454469) [](ancestors = 390454469)

src/Test.UnitTests.Sarif/Baseline2/TrustMapTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            Assert.Equal(1.0f, beforeMap.Trust(Set, NameGreat));
+            Assert.Equal((3.0f / 20.0f), beforeMap.Trust(Set, NameMedium));
+            Assert.Equal(TrustMap.DefaultTrust, beforeMap.Trust(Set, NameNoMatches));
+            Assert.Equal(0.0f, beforeMap.Trust(Set, NameConstant));
+
+            // Same scores from both maps
+            Assert.Equal(1.0f, afterMap.Trust(Set, NameGreat));
```
lgolding: >; [](start = 62, length = 1)

Consider writing these four tests as
```
Assert.Equal(beforeMap.Trust(Set, NameGreat), afterMap.Trust(Set, NameGreat));
```
for two reasons:
1. It better expresses the intent ("the two maps have the same scores")
2. It is DRYer; if you changed the values, you wouldn't have to remember to change them here too.

src/build.props
michaelcfanning: I continue to be concerned about this. We are versioning many packages in sync. Diligent SemVer versioning across all these packages will greatly accelerate increase in versioning. Also, we're flat out lying about the need to update components. i.e., we are communicating that there is a breaking change everywhere, when that is not the case.

I would prefer to stick with our current model (which isn't strict SemVer, definitely admit that) unless/until we have that more complete conversation (i.e., agree that we have consensus we will uniformly apply the standard to all our various distinct packages).


---
In reply to: [390015584](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390015584) [](ancestors = 390015584)

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
```diff
@@ -18,13 +18,25 @@ public class ExtractedResult
         public Run OriginalRun { get; set; }
         public string RuleId { get; set; }
 
+        private ReportingDescriptor Rule { get; set; }
```
michaelcfanning: >ReportingDescriptor [](start = 16, length = 19)

I always find it odd when a private member is UpperCased. Because it makes me wonder if the developer intended for it to be public. :) Did you?

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
michaelcfanning: I don't think you need to be concerned about breaking changes. We can project non-trivial churn in this area, we've only just started to apply it ourselves, mostly to a single tool. 

---
In reply to: [390461605](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390461605) [](ancestors = 390461605,390454469)

src/Sarif/Baseline/V2/V2ResultMatcher.cs
```diff
@@ -29,16 +29,19 @@ public IList<MatchedResults> Match(IList<ExtractedResult> before, IList<Extracte
     /// </remarks>
     internal class StatefulResultMatcher
     {
+        // Threshold for how many 'nearby' Results are considered after other match phases
+        private const int NearnessThreshold = 3;
+
         // Results in the first set.
         private List<ExtractedResult> Before { get; }
+        private TrustMap BeforeTrustMap { get; }
```
michaelcfanning: >BeforeTrustMap [](start = 25, length = 14)

Why are these properties and not fields?

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
```diff
@@ -46,6 +47,12 @@ private static string RelativeSeriesPath(string seriesPath)
 
         public void Write(SarifLog newBaselineLog, SarifLog baselineLog, SarifLog currentLog, BaseliningSummary summary)
         {
+            Dictionary<string, Result> baselineResultsByGuid = new Dictionary<string, Result>();
```
michaelcfanning: >Dictionary<string, Result> [](start = 12, length = 26)

When we rerun our style guidelines, this explicit type declaration will be converted to 'var'. I guess as a rule we no longer need to raise this sort of comment in a PR, though, right? i.e., we can just rerun that standard periodically and take the automated clean-up.

src/Test.UnitTests.Sarif/Baseline2/TrustMapTests.cs
ScottLouvau: The second clause means "values which are all unique but which have no matches across runs should have low trust". I'll reword to make it clearer.

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
```diff
@@ -89,7 +97,12 @@ private void Write(char marker, string guid)
 
         public void Dispose()
         {
-            if (Writer != null)
+            Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
michaelcfanning: >protected [](start = 8, length = 9)

You can add the DIspose(true) pattern if you want but if you don't have native interop, it's not strictly necessary. This pattern exists for finalized objects that weren't disposed by callers but have an urgent need to make sure native handles are disconnected properly.

src/Sarif/Baseline/ResultMatching/DataStructures/ExtractedResult.cs
ScottLouvau: The coding guidelines I'm familiar with specified that all methods and properties are PascalCased, regardless of access modifier. It's just fields that vary.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: I use fields only when the performance of the getter method might be an issue. I tend to prefer properties when possible.

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
ScottLouvau: I thought it was removing var for ambiguous cases, and not converting to var in any case. Either way, if the formatter changes it, that's fine.

I ran the formatting command earlier and it didn't seem to change anything, and I believe it was after this line was here.

src/Test.EndToEnd.Baselining/BaseliningDetailLogger.cs
ScottLouvau: Not necessary but the IDE now squiggles all classes that don't have both overloads. I want all the squiggles I see to really be errors worth fixing, so I've been converting to this form to "fix" these warnings from VS.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
michaelcfanning: This is very inconsistent with the rest of the code base. I think it alters the readability of the code. How strongly do you feel about retaining this for your repo contributions? If it's important, we can be consistent in this pattern for the Baseline API. I'd need to be convinced it's useful before blessing it as a universal standard. It compromises code readability from my perspective. 

---
In reply to: [390520014](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390520014) [](ancestors = 390520014)

src/Test.UnitTests.Sarif/Baseline2/TrustMapTests.cs
ScottLouvau: Good idea.

src/Test.UnitTests.Sarif/Visitors/RemoveOptionalDataVisitorTests.cs
ScottLouvau: I've fixed this file and others I added.

src/build.props
lgolding: Just a detail: 2.2.1 => 2.3.0 does not communicate that there is a breaking change. It says that there is enhanced, backward-compatible functionality. But perhaps you weren't referring to the particular version I was suggesting for this PR.

---
In reply to: [390504146](https://github.com/microsoft/sarif-sdk/pull/1807#discussion_r390504146) [](ancestors = 390504146,390015584)

ScottLouvau: Given that our team approach to versioning isn't settled yet, I'll go back to 2.2.1 for this release per Mike and am happy to participate in a thread or in person discussion about how we want to handle versioning.

src/Sarif/Baseline/V2/V2ResultMatcher.cs
ScottLouvau: I don't disagree with this one. The code is mostly consistent. I've changed all of my private property usage to fields with underscore prefix.

I think other "style above naming" would be better discussed in threads, settled, and then fixed up across the codebase.

PR #1809, opened 2020-03-10
Move to .NET Core 3.0 build

ScottLouvau:
Per Mike, this change updates SARIF framework targets to netcoreapp3.0;net461. This is needed to unblock our NPM build, which must be .NET Core 3.0 and used a trick to not disturb the normal project files which no longer works.

Targetting .NET Core 3.0 also improves our build and release time, as the same target will be in our NuGet packages and our NPM release.

### DIFF
--- a/BuildPackagesFromLayoutDirectory.cmd
+++ b/BuildPackagesFromLayoutDirectory.cmd
@@ -41,7 +41,7 @@ goto :Exit
 xcopy /Y %LayoutForSigningDirectory%\net452\%1 %BinaryOutputDirectory%\net452\
 
 :: For .NET core, .exes are renamed to .dlls due to packaging conventions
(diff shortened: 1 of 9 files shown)


michaelcfanning:
Looks right. Let me pull your branch, build, then inspect the package. Will ensure the x64 multitool is callable.

michaelcfanning:
FYI, build break on my machine due to failure of providing a RuntimeIdentifier. I will push a fix into your branch momentarily.

---
In reply to: [597360354](https://github.com/microsoft/sarif-sdk/pull/1809#issuecomment-597360354) [](ancestors = 597360354)

ScottLouvau:
Abandoning; after more investigation, I found this broke with the work to enable the Semmle build. If we undo some of those changes, we can fix the build without forcing consumers to .NET Core 3.1.

PR #1812, opened 2020-03-12
Restore npm build

michaelcfanning:


src/Sarif/Sarif.csproj
```diff
@@ -18,14 +18,18 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
ScottLouvau: You want net45 here as well for the SDK library itself, right?

src/Test.Plugins/Test.Plugins.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFrameworks>net461</TargetFrameworks>
```
ScottLouvau: Should the test projects all be netcoreapp2.1, for consistency? The test execution logic was a bit weird, and it seems like tests in VS sometimes run on all framework targets and sometimes just one. If we make all test projects just one target, hopefully this means all tests would always run and would only run once.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -7,6 +7,10 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
+  <PropertyGroup>
```
ScottLouvau: It seems like about half the time you're letting build.props set the targets, and the other half you aren't - is it worth having build.props try to specify them at this point? It seems like it's not helping a lot, and then you add the gotcha that many projects need to ensure they set TargetFrameworks only after the import.

src/Test.Plugins/Test.Plugins.csproj
michaelcfanning: This isn't a test binary, it is a test plug-in. That is, a plug-in we use for testing. Its build is separate from the rest of the system.

---
In reply to: [391736031](https://github.com/microsoft/sarif-sdk/pull/1812#discussion_r391736031) [](ancestors = 391736031)

src/Sarif.Converters/Sarif.Converters.csproj
michaelcfanning: Currently all product code requires it be sent after the import. For consistency, we should make the test assemblies work the same way.

---
In reply to: [391737075](https://github.com/microsoft/sarif-sdk/pull/1812#discussion_r391737075) [](ancestors = 391737075)

src/Sarif/Sarif.csproj
michaelcfanning: Yes, added this back.

---
In reply to: [391735210](https://github.com/microsoft/sarif-sdk/pull/1812#discussion_r391735210) [](ancestors = 391735210)

src/Test.Plugins/Test.Plugins.csproj
michaelcfanning: all test projects now verified to be netcoreapp2.1

---
In reply to: [391844821](https://github.com/microsoft/sarif-sdk/pull/1812#discussion_r391844821) [](ancestors = 391844821,391736031)

PR #1813, opened 2020-03-13
Add SARIF-derived bug titles to SARIF work item filer

michaelcfanning:
This change adds a simple extension method off the SARIF `run` object that produces a descriptive bug title from the first result in the run. The basic format of the title consists of the tool name, the failure level, the opaque id, the friendly id (if one exists) and the logical (preferred) or physical location if these exist.

We do not alter the title (yet) in cases where the bug contains multiple issues (e.g. by appending something like `and X more results.`

The title generation is wired into the core filer. I will create a mocked test for this in a follow-on change (and push the core helper coverage from 86% to 95%+). Would like a review of current work in order to unblock providing real world samples to PM to evaluate.

Here are some sample titles produced by our checked in test content:
~~~
[AndroidStudio:Warning] unused declaration (in 'app\microsoft.androidstudioformatexample\microsoft.androidstudioformatexample.ApplicationTest ApplicationTest()')

[Clang Analyzer:Warning] Dead assignment (in jcparam.c)

[Contrast Security:Error] sql-injection: SQL Injection (in /webgoat/Content/ForgotPassword.aspx)

[CppCheck:Warning] uselessAssignmentArg (in addons/test/cert-test.c)

[Fortify:Warning] Insecure Randomness (in bannedAPIs.java)

[FxCop:Warning] CA1704: IdentifiersShouldBeSpelledCorrectly (in 'Microsoft.CodeAnalysis.Sarif')

[MSBuild:Error] CS1026 (in VisualStudioBuildLogConverterTests.cs)

[PREfast:Warning] 6011 (in 'IrrelevantBranch_6011')
~~~

@alisonmaclellan @cfaucon @nickmarston @chinarosesz


src/Sarif/Core/Result.cs
```diff
@@ -128,7 +128,7 @@ public override string ToString()
                 {
                     sb.Append(argument + ",");
                 }
-                sb.Length = sb.Length - 1;
+                sb.Length -= 1;
```
michaelcfanning: >Length [](start = 19, length = 6)

Just a random style change VS asked for.

src/Sarif.WorkItems/SarifWorkItemModel.cs
```diff
@@ -43,11 +43,9 @@ public SarifWorkItemModel(SarifLog sarifLog, SarifWorkItemContext context = null
...
             };
 
-            // TODO: Provide a useful SARIF-derived issue title
-            //
-            //       https://github.com/microsoft/sarif-sdk/issues/1755
-            //
-            this.Title = $"Default {nameof(this.Title)}";
+            string title = sarifLog.Runs?[0]?.CreateWorkItemTitle();
+
+            this.Title = title ??  "[ERROR GENERATING TITLE]";
```
michaelcfanning: >[ERROR GENERATING TITLE] [](start = 36, length = 24)

1. When Alison is done with the logging, we will record any substantive problems trying to produce the work item. Alternately, we could raise an exception and refuse to complete the work. Alternately, we could improve on this error and keep processing. Thoughts?

src/Test.FunctionalTests.Sarif/SarifValidatorTests.cs
```diff
@@ -192,7 +192,7 @@ public void ValidatesAllTestFiles()
             @"v2\ConverterTestData",
             @"v2\SpecExamples",
             @"v2\ObsoleteFormats",
-            @"..\..\Test.UnitTests.Sarif\net461\TestData"
+            @"..\..\Test.UnitTests.Sarif\netcoreapp2.1\TestData"
```
michaelcfanning: >netcoreapp2 [](start = 41, length = 11)

we no longer build test binaries for net461, only for netcoreapp2.1. eliminates some unnecessary redundancy in testing.

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
```diff
@@ -8,7 +8,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
```
michaelcfanning: >net461 [](start = 36, length = 6)

Failure to remove this target was an oversight in a previous build clean-up change. Noticed it only in the VS Test Explorer UI.

src/Test.UnitTests.Sarif.Driver/Sdk/FormatForVisualStudioTests.cs
```diff
@@ -13,22 +14,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     // also exercise Region.FormatForVisualStudio and ResultKind.FormatForVisualStudio.
     public class FormatForVisualStudioTests
     {
-        private const string TestRuleId = "TST0001";
-        private const string TestMessageStringId = "testMessageStringId";
```
michaelcfanning: Moved this data to a dedicated TestData class, which we should continue building out to hold all the test things. There are lots of important SARIF snippets to author (such as one in this PR in which we construct SARIF that explores the complete range of how to provide rule ids). There are analogs with logical locations, physical locations, etc. Need to make progress organizing these assets (and authoring comprehensive shared patterns). A WIP.

src/Sarif.Sdk.sln
```diff
@@ -96,6 +96,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		.editorconfig = .editorconfig
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ConsoleApp1", "ConsoleApp1\ConsoleApp1.csproj", "{416A443A-B0C0-44A8-AA4A-B1221B79F6C2}"
```
chinarosesz: >ConsoleApp1 [](start = 80, length = 11)

Is this a new project?

src/Sarif.WorkItems/SarifWorkItemModel.cs
chinarosesz: >  [](start = 34, length = 1)

Extra space?

chinarosesz: In a scenario where there are thousands of bugs to create and there is an error in generating the title, would the users be ok having a tons of incorrect/modified titles? I'm trying to see it from a customers' perspective.

---
In reply to: [392391848](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392391848) [](ancestors = 392391848)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            // We encapsulate logical names in apostrophes to help indicate they are a symbol
+            string locationName = "'" + firstResult.Locations?[0].LogicalLocation?.FullyQualifiedName + "'";
+
+            if (locationName.Equals("''"))
+            {
+                // We don't bother to wrap a file path or URL in apostrophes as these are self-evident
+                //
+                // Lines of code like this that inspire strong feelings in SARIF consumers as far as its design.
+                locationName = firstResult.Locations?[0].PhysicalLocation?.ArtifactLocation?.Resolve(run)?.Uri.OriginalString;
```
cfaucon: missing Uri? since it could be null as well?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+            //
+            // Informational: a strictly informational message was emitted.
+        }
+
+        public static string CreateWorkItemTitle(this Run run)
+        {
+            Result firstResult = null;
+            var ruleIds = new HashSet<string>();
+
+            foreach (Result result in run?.Results)
```
chinarosesz: >run? [](start = 38, length = 4)

If run is null, foreach will throw a NullReferenceException

src/Sarif.Sdk.sln
michaelcfanning: Test project I used to produce a title from every checked in SARIF result. :) Updated the SLN to remove this reference.

---
In reply to: [392436877](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392436877) [](ancestors = 392436877)

src/Sarif.WorkItems/SarifWorkItemModel.cs
michaelcfanning: Nope, this would be a bad customer experience. Do you have a suggestion?

---
In reply to: [392450784](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392450784) [](ancestors = 392450784,392391848)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
michaelcfanning: Yes, nice. Note to self: missing unit test case, will close that hole.

---
In reply to: [392451392](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392451392) [](ancestors = 392451392)

src/Sarif.WorkItems/SarifWorkItemModel.cs
chinarosesz: I guess I would fail fast and provide lots of logging for developers to debug and/or customers to fix up their input if that's the case. 

---
In reply to: [392490164](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392490164) [](ancestors = 392490164,392450784,392391848)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+            // Returns strings like:
+            //
+            // [Tool:Warning] RULE3067 (in 'Namespace.Type.MyMethod()')
+            // [Tool:Error] RULE2001: Null derefernece (in c:\src\build.cpp)
+
+            return "[" + run.Tool.Driver.Name + ":" +
+                    firstResult.Level.ToString() + "]: " +
+                    fullRuleId +
+                    (locationName == null ? "" : " (in " + locationName + ")");
+        }
```
alisonmaclellan: I was looking at the TSA bug title generation code yesterday. Thought I would share in case we want to incorporate any of it.  I think your code is taking the title in a slightly different direction. There are folks though who already have post processing hooks looking for this pattern. 

GlobalConfig.Environment is the stamp, it is empty in V2.  

string valueTitle = string.Format(
                "{0}[ {1} ][ {2} ] - {3} ({4})",
                Formatter.FormatIfNotEmpty("( {0} )", GlobalConfig.Environment),
                tsaWorkItem.CodeBase.Name,
                tsaWorkItem.ScanTool.DisplayName,
                tsaWorkItem.IssueGroup,
                Formatter.ToFriendlyQuantity(issueCount, "issue", "issues"));

src/Sarif.WorkItems/SarifWorkItemModel.cs
michaelcfanning: I tend to agree. The implications of this approach are that a corner case issue with title construction could complete torpedo a bug filing session for a specific tool. By failing entirely, we put pressure (which could be useful) on acquiring a tool fix as soon as possible. Tool updates themselves come with complications (as in many cases, a spot fix for a detail issue like this might not be applied to the current version. Instead, the upgrade path may imply receiving a minor/major update with analysis delta). All of this makes a point I repeat all the time: our systems need to be designed for minimizing friction for tool servicing changes. It's interesting to see how this local decision in behavior needs to connect to a pretty high-level design philosophy. :) In any case, I will make this fail-fast for now (in the next clean-up change).

---
In reply to: [392501325](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392501325) [](ancestors = 392501325,392490164,392450784,392391848)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
michaelcfanning: Well, yes. And if run.Results is null that will throw as well. :P Fixing, nice catch. I'm surprised LGTM doesn't catch this, could be worth a problem report. @hakohli

---
In reply to: [392462133](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392462133) [](ancestors = 392462133)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
michaelcfanning: I'm definitely aware of this code. :) Trying to thoughtfully rebuild titles from first principles. I think you will find that the TSA bug bodies argue for something similar. I'm glad you raised this point, though, because it makes me realize we could move some of the old TSA title content into a simplified message body. Wil follow up with you offline. 

---
In reply to: [392506645](https://github.com/microsoft/sarif-sdk/pull/1813#discussion_r392506645) [](ancestors = 392506645)

PR #1817, opened 2020-03-18
 Initial body description for work items

nickmarston:
This is the initial draft for a basic body for work items in Sarif.  I'll add a couple of notes in the review, but the main thing to call out here is that it is currently plain text and generic for ADO\github (technically the bullets are in markdown, but that looks fine in text as well).

As the Sarif workitem is agnostic to platform, we may eventually need to store relevant fields and let the filers create the appropriate text.  This means they would have minimal knowledge of some Sarif like fields, but wouldn't need to know Sarif concepts directly.


src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -75,6 +77,35 @@ public static string CreateWorkItemTitle(this Run run)
...
 
+        public static Dictionary<string, int> FetchToolResultCounts(this SarifLog log)
+        {
+            if (log == null) { throw new NullReferenceException(); }
+            if (log.Runs == null) { throw new ArgumentNullException(nameof(log.Runs)); }
+
+            Dictionary<string, int> resultCountsByTool = new Dictionary<string, int>();
+            int appropriateResultCount = 0;
+
+            foreach (Run run in log?.Runs)
```
nickmarston: I feel like they're might be some Sarif trick here to make this faster, as it has to iterate through all the results to make sure we only count the one's we plan on filing.  I don't know it, though.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -85,5 +116,28 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
             }
             return fullRuleIdentifier;
         }
+
+        public static string CreateWorkItemDescription(this SarifLog log)
```
nickmarston: Taking any and all suggestions formatting here.  I do think starting with plain text is fine to get something available.  I'll be looking at possibly just storing some very basic data and having the specific filers customize the description.

src/Sarif.WorkItems/SarifWorkItemModel.cs
```diff
@@ -56,23 +56,11 @@ public SarifWorkItemModel(SarifLog sarifLog, SarifWorkItemContext context = null
...
-            //       (markdown vs. HTML). We only have a single tracking issue below - 
-            //       in the event, this work is likely to be broken aparts into 
-            //       multiple tasks, each of which may entail breaking changes to 
-            //       models/API.
-            //
-            //       https://github.com/microsoft/sarif-sdk/issues/1757
-            //
-            this.BodyOrDescription = $"Default {nameof(this.BodyOrDescription)}";
+            this.BodyOrDescription = sarifLog.CreateWorkItemDescription();
+            if (string.IsNullOrEmpty(this.BodyOrDescription))
```
michaelcfanning: >IsNullOrEmpty [](start = 23, length = 13)

I think we can live with an empty body or description if your helper bellies up. So suggest deleting this setting of default entirely in that condition.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -75,6 +77,35 @@ public static string CreateWorkItemTitle(this Run run)
                     (locationName == null ? "" : " (in " + locationName + ")");
         }
 
+        public static Dictionary<string, int> FetchToolResultCounts(this SarifLog log)
+        {
+            if (log == null) { throw new NullReferenceException(); }
```
michaelcfanning: >NullReferenceException [](start = 41, length = 22)

Throw ArgumentNullException here

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -75,6 +77,35 @@ public static string CreateWorkItemTitle(this Run run)
...
+            if (log.Runs == null) { throw new ArgumentNullException(nameof(log.Runs)); }
+
+            Dictionary<string, int> resultCountsByTool = new Dictionary<string, int>();
```
michaelcfanning: > Dictionary<string, int> [](start = 11, length = 24)

NIT, our automated code style will convert this explicit type declaration to 'var'.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -75,6 +77,35 @@ public static string CreateWorkItemTitle(this Run run)
                     (locationName == null ? "" : " (in " + locationName + ")");
         }
 
+        public static Dictionary<string, int> FetchToolResultCounts(this SarifLog log)
```
michaelcfanning: >FetchToolResultCounts [](start = 46, length = 21)

Maybe name this "ComputePerToolResultsCount"? @lgolding may also have an idea.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -75,6 +77,35 @@ public static string CreateWorkItemTitle(this Run run)
...
+
+            foreach (Run run in log?.Runs)
+            {
+                if (run != null)
+                {
+                    if (run?.Tool?.Driver?.Name != null && run?.Results != null)
+                    {
+                        foreach (Result result in run?.Results)
+                        {
+                            if (result.AppropriateForFiling())
```
michaelcfanning: >AppropriateForFiling [](start = 39, length = 20)

Suggest removing this filter. The general helper should just provide the actual data. That means any caller needs to have filtered out what data they aren't interested. Our filer will have completed this work before your helper gets called.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -85,5 +116,28 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
             }
             return fullRuleIdentifier;
         }
+
+        public static string CreateWorkItemDescription(this SarifLog log)
+        {
+            Dictionary<string, int> resultCountsByTool = FetchToolResultCounts(log);
+            StringBuilder templateText = new StringBuilder(@"This bug has been filed by the Sarif Work Item Automatic Filer.  It contains results for the following tools and issues:");
```
michaelcfanning: >This bug has been filed by the Sarif Work Item Automatic Filer.  It contains results for the following tools and issues: [](start = 61, length = 120)

Larry and I can work with you offline on the text. I can tell you that I don't want to burn the  name of our filer in this message. The focus is not on us, it is on the bug content. :) 

Just a note. We need to think about how the system will dynamically produce some data while allow other pieces to be configurable. We can define things on the command-line and in context for things like a support alias.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -85,5 +116,28 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
+            {
+                templateText.Append(string.Format("*Tool: {0}", toolResult.Key));
+                templateText.AppendLine();
+                templateText.Append(string.Format("     Result count: {0}", toolResult.Value));
+                templateText.AppendLine();
+                templateText.AppendLine();
+            }
+
+            templateText.AppendLine();
+            templateText.Append(@"To see result details, please visit the Scans tab of this bug, or the attached Sarif log.");
```
michaelcfanning: >, please visit the Scans tab of this bug, [](start = 55, length = 41)

Here's an example of why we need some sort of configurable footer. The 'scans' tab is specific to the MS ADO instance, doesn't apply to GH or other ADO instances.

A very simple idea for the MVP, produce your summary using the text below, then inject an optional static string which has been provided to us in the context file.


"This work item was filed for the following issues:\r\n{your summary here}\r\n{inject a configurable footer here if available}"

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilingContextTests.cs
```diff
@@ -50,6 +50,17 @@ public void WorkItemFilingContext_NullSarifLogRaisesArgumentNullException()
             workItemModel.Area.Should().BeNull();
         }
 
+        [Fact]
+        public void WorkItemFilingContext_PopulateDescription()
```
michaelcfanning: >WorkItemFilingContext_PopulateDescription [](start = 20, length = 41)

The name of this test s/be 'WorkItemFilingContext_PopulatesDescription'. Per Larry, the test name communicates what is accomplishes.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilingContextTests.cs
```diff
@@ -50,6 +50,17 @@ public void WorkItemFilingContext_NullSarifLogRaisesArgumentNullException()
...
+        {
+            var context = new SarifWorkItemContext();
+            SarifLog sarifLog = TestData.SarifLogs.OneIdThreeLocations;
+
+            var workItemModel = new SarifWorkItemModel(sarifLog, context);
```
michaelcfanning: >SarifWorkItemModel [](start = 36, length = 18)

Why is this test in the SarifWorkItemFilingContext test file? The context doesn't currently influence the description creation (though I've suggested otherwise in another PR comment. :) You should move this to WorkItemModel tests, I think, or test at a lower level in WorkItemExtesionsTests.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilingContextTests.cs
michaelcfanning: >WorkItemFilingContext_PopulateDescription [](start = 20, length = 41)

You should consider unit-testing the FetchToolResultCounts helper as well. s/be straightforward.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: @mikefan@microsoft.com your name sounds good. @nickmarston I suggest only using Fetch when retrieving an item from a repository such as a database. Even then I prefer Get.

---
In reply to: [394696836](https://github.com/microsoft/sarif-sdk/pull/1817#discussion_r394696836) [](ancestors = 394696836)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: "This bug was filed automatically. ..."

---
In reply to: [394698234](https://github.com/microsoft/sarif-sdk/pull/1817#discussion_r394698234) [](ancestors = 394698234)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: Changed the identical check for title as well.

---
In reply to: [394695607](https://github.com/microsoft/sarif-sdk/pull/1817#discussion_r394695607) [](ancestors = 394695607)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: Removing the filter makes this a simple grab of the count.

---
In reply to: [394674496](https://github.com/microsoft/sarif-sdk/pull/1817#discussion_r394674496) [](ancestors = 394674496)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: Discussed offline with multiple parties.

---
In reply to: [394675544](https://github.com/microsoft/sarif-sdk/pull/1817#discussion_r394675544) [](ancestors = 394675544)

PR #1820, opened 2020-03-20
Initial logging framework for WorkItems

rtaket:
This logger is only used by the WorkItems verb. It provides the ability to log to the console and to Application Insights. To enable the logger, an appsettings.json file must be provided.


src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -38,6 +42,8 @@ public SarifWorkItemFiler(Uri filingUri = null, SarifWorkItemContext filingConte
             }
 
             this.FilingClient = FilingClientFactory.Create(this.FilingContext.HostUri);
+
+            this.Logger = ServiceProviderFactory.ServiceProvider.GetService<ILogger<FilingClient>>();
```
rtaket: >this [](start = 12, length = 4)

This is how you retrieve a logger.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -79,65 +87,82 @@ public virtual void FileWorkItems(string sarifLogFileContents)
...
-                SarifLog splitLog = logsToProcess[splitFileIndex];
-                FileWorkItemsHelper(splitLog, this.FilingContext, this.FilingClient);
+                var logsToProcessMetrics = new Dictionary<string, object>
+                {
+                    { "splittingStrategy", splittingStrategy },
+                    { "logsToProcessCount", logsToProcess.Count },
+                    { "splittingDurationInMilliseconds", splittingStopwatch.ElapsedMilliseconds },
+                };
+
+                this.Logger.LogMetrics(EventIds.LogsToProcessMetrics, logsToProcessMetrics);
```
rtaket: >this [](start = 16, length = 4)

@alisonmaclellan, this is how you would publish metrics to AI.

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -79,4 +79,13 @@
     <EmbeddedResource Include="..\Sarif\Schemata\sarif-2.1.0-rtm.5.json" Link="sarif-2.1.0.json" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="appsettings.json">
+      <CopyToOutputDirectory Condition="'$(Configuration)' == 'Debug'">PreserveNewest</CopyToOutputDirectory>
```
rtaket: >Condition="'$(Configuration)' == 'Debug'" [](start = 29, length = 41)

Although I added an appsettings.json file to the Multitool project, it doesn't get packaged in the Release build. i.e. I added it here for devs (and as an example.)

src/Sarif.Multitool/appsettings.local.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
rtaket: If you want to override the appsettings, use this file. Anything in here will override the values in appsettings.json. I also added this file to .gitignore. So you shouldn't have to worry about checking in secrets.

src/Sarif.Multitool/Program.cs
```diff
@@ -12,33 +15,42 @@ internal static class Program
...
+                    (QueryOptions queryOptions) => new QueryCommand().Run(queryOptions),
+                    (ResultMatchingOptions baselineOptions) => new ResultMatchingCommand().Run(baselineOptions),
+                    (ResultMatchSetOptions options) => new ResultMatchSetCommand().Run(options),
+                    (FileWorkItemsOptions fileWorkItemsOptions) => new FileWorkItemsCommand().Run(fileWorkItemsOptions),
+                    errs => CommandBase.FAILURE);
+            }
+            finally 
+            {
+                ITelemetryChannel channel = ServiceProviderFactory.ServiceProvider.GetService<ITelemetryChannel>();
+                channel?.Flush();
```
rtaket: >Flush [](start = 25, length = 5)

@michaelcfanning, we have to flush the telemetry stream before exiting. Are you OK doing it here? Or do you want to move this to a try/finally in the SarifWorkItemFiler?

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -79,65 +87,82 @@ public virtual void FileWorkItems(string sarifLogFileContents)
...
+                    return;
                 }
-                default:
+
+                IList<SarifLog> logsToProcess = new List<SarifLog>(new SarifLog[] { sarifLog });
+
+                PartitionFunction<string> partitionFunction = null;
+
+                Stopwatch splittingStopwatch = Stopwatch.StartNew();
+
```
alisonmaclellan: Did you add this just to include in the logging below?

src/Sarif.WorkItems/SarifWorkItemFiler.cs
alisonmaclellan: @ So if I want to add more event ids, I'll add it to the EventIds.cs. 

---
In reply to: [395894423](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r395894423) [](ancestors = 395894423)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -79,65 +87,82 @@ public virtual void FileWorkItems(string sarifLogFileContents)
...
 
-            logsToProcess = new List<SarifLog>(partitioningVisitor.GetPartitionLogs().Values);
+                logsToProcess = new List<SarifLog>(partitioningVisitor.GetPartitionLogs().Values);
 
-            for (int splitFileIndex = 0; splitFileIndex < logsToProcess.Count; splitFileIndex++)
-            {
-                SarifLog splitLog = logsToProcess[splitFileIndex];
-                FileWorkItemsHelper(splitLog, this.FilingContext, this.FilingClient);
+                var logsToProcessMetrics = new Dictionary<string, object>
+                {
```
alisonmaclellan: Thoughts on if I should look at building a class that creates the dictionary rather than in-lining?

src/Sarif.WorkItems/SarifWorkItemFiler.cs
rtaket: Yes, I think that would be the plan. I don't think it makes sense to have a different EventIds.cs file per project since the logger is only used for work items. i.e. we're treating the WorkItems project as a 'core' project that can hold infrastructure object.

---
In reply to: [395897554](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r395897554) [](ancestors = 395897554,395894423)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
rtaket: Yes. It's really just an example. I doubt we really care about the perf of splitting the SARIF files. I suspect we'll remove this when we add the real logging messages.

---
In reply to: [395896508](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r395896508) [](ancestors = 395896508)

alisonmaclellan: Great. Thanks. I was confused by it at first until I saw you use it and wanted to confirm it was just an example.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
rtaket: I suspect we'll want a method that accepts an EventId and a list of environment variables and then creates the dictionary and sends the telemetry. Mostly because I suspect we'll want to capture environment variables from multiple environments. i.e. not only ADO build agents.

---
In reply to: [395897876](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r395897876) [](ancestors = 395897876)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -79,65 +87,82 @@ public virtual void FileWorkItems(string sarifLogFileContents)
...
-            OptionallyEmittedData optionallyEmittedData = this.FilingContext.DataToRemove;
-            if (optionallyEmittedData != OptionallyEmittedData.None)
-            {
-                var dataRemovingVisitor = new RemoveOptionalDataVisitor(optionallyEmittedData);
-                dataRemovingVisitor.Visit(sarifLog);
-            }
-
-            optionallyEmittedData = this.FilingContext.DataToInsert;
-            if (optionallyEmittedData != OptionallyEmittedData.None)
+            using (IDisposable loggingScope = Logger.BeginScope("FileWorkItems"))
```
rtaket: >BeginScope [](start = 53, length = 10)

This is also optional. It's just an example, if anyone wants to use logging scopes.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
alisonmaclellan: Ok, but will I want that method in a new class or should I put it in one you already created. I was trying to decide if it fit in anything you already added.

src/Sarif.Multitool/Program.cs
nickmarston: It feels like the SarifWorkItemFiler would be the better place, simply because we know we often won't be used from this specific command line entry.  I'd prefer to keep everything crucial within the workitemfiler, even if in some cases it might not be necessary (when we run as a service, for example).

src/Sarif.Multitool/appsettings.local.json
nickmarston: Does it make sense to also have setting in the SarifWorkItemContext?  This is where we allow our configuration file to setup and override other keys, though this may be a problem the context is not initialized when we first need the logger.

rtaket: I wouldn't have expected these settings to be settable from the SarifWorkItemContext. But per the standup conversation, I'd prefer to defer this decision until we have a better understanding of how and where we handle configuration.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
rtaket: I suspect a new class. I think we'll need a TelemetryHelpers class (or similarly named) to hold all the common telemetry functions. 

---
In reply to: [395908403](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r395908403) [](ancestors = 395908403)

src/Sarif.Multitool/appsettings.local.json
rtaket: I wouldn't have expected these settings to be settable from the SarifWorkItemContext. But per the standup conversation, I'd prefer to defer this decision until we have a better understanding of how and where we handle configuration.

---
In reply to: [396154828](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396154828) [](ancestors = 396154828)

src/Sarif.Multitool/Program.cs
rtaket: Fixed. I moved the flush to the Dispose method of SarifWorkItemFiler.

---
In reply to: [396154484](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396154484) [](ancestors = 396154484)

src/Sarif.Multitool/Program.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using CommandLine;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.WorkItems;
```
michaelcfanning: >WorkItems [](start = 16, length = 9)

Why have these been added here? I don't see relevant changes that would require it. In any case, if truly unnecessary, we can let the automated code clean-up take care of it.

src/WorkItems/ServiceProviderFactory.cs
```diff
@@ -0,0 +1,88 @@
+﻿using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
```
michaelcfanning: >Extensions [](start = 16, length = 10)

System directives come first. (Change your VS settings to make this a default). Add copyright header to this file.

src/Sarif.Multitool/Program.cs
rtaket: When I removed the telemetry flush from this file, I forgot to remove the usings. I've removed them now.

---
In reply to: [396822720](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396822720) [](ancestors = 396822720)

src/WorkItems/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Resources;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Test.UnitTests.WorkItems, PublicKey=0024000004800000940000000602000000240000525341310004000001000100433fbf156abe9718142bdbd48a440e779a1b708fd21486ee0ae536f4c548edf8a7185c1e3ac89ceef76c15b8cc2497906798779a59402f9b9e27281fb15e7111566cdc9a9f8326301d45320623c5222089cf4d0013f365ae729fb0a9c9d15138042825cd511a0f3d4887a7b92f4c2749f81b410813d297b73244cf64995effb1")]
```
michaelcfanning: >0024000004800000940000000602000000240000525341310004000001000100433fbf156abe9718142bdbd48a440e779a1b708fd21486ee0ae536f4c548edf8a7185c1e3ac89ceef76c15b8cc2497906798779a59402f9b9e27281fb15e7111566cdc9a9f8326301d45320623c5222089cf4d0013f365ae729fb0a9c9d15138042825cd511a0f3d4887a7b92f4c2749f81b410813d297b73244cf64995effb1 [](start = 67, length = 320)

FYI, I'm moving this into \src\Shared\CommonAssemblyInfo.cs

src/WorkItems/ServiceProviderFactory.cs
rtaket: Added copyright and sorted using directives for the files.

---
In reply to: [396823390](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396823390) [](ancestors = 396823390)

src/WorkItems/Logging/MetricsLogValues.cs
```diff
@@ -0,0 +1,78 @@
...
+        private readonly string _originalMessage;
+        private readonly EventId _eventId;
+
+        public KeyValuePair<string, object> this[int index]
+        {
+            get
+            {
+                if (index < 0 || index >= Count)
+                {
+                    throw new IndexOutOfRangeException("index");
```
michaelcfanning: >IndexOutOfRangeException [](start = 30, length = 24)

Aren't you just recreating existing index behavior here?

rtaket: Yeah, it's from the original implementation. When I removed some of the logic, it was reduced to this. I'll remove the if block.

---
In reply to: [396828009](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396828009) [](ancestors = 396828009)

src/WorkItems/Logging/ApplicationInsightsTelemetryInitializer.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry?.Context?.Operation == null)
+            {
+                throw new ArgumentNullException(nameof(telemetry));
+            }
+
+            if (string.IsNullOrEmpty(telemetry.Context.Operation.Id))
```
michaelcfanning: >if (string.IsNullOrEmpty(telemetry.Context.Operation.Id)) [](start = 12, length = 57)

Why is this right? i.e., that we preserve the id of the incoming thing when we call initialize. Why don't we, for example, reset m_operationId in this case? i.e., how is it legal/useful for the ids to diverge.

src/WorkItems/Logging/EventIds.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.WorkItems.Logging
+{
+    public static class EventIds
+    {
+        public static EventId LogsToProcessMetrics => new EventId(9001, nameof(LogsToProcessMetrics));
```
michaelcfanning: >9001 [](start = 66, length = 4)

Why this value? Will we have a convention for assigning ids?

src/WorkItems/Logging/LoggerExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.WorkItems.Logging
+{
+    public static class LoggerExtensions
+    {
+        public static void LogMetrics(this ILogger logger, EventId eventId, IDictionary<string, object> customDimensions)
```
michaelcfanning: >LogMetrics [](start = 27, length = 10)

Why don't we have a single extension method here that puts the optional message arg at the end? Are you trying to match existing overload patterns on ILogger?

src/WorkItems/Logging/MetricsLogValues.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.WorkItems.Logging
+{
+    /// <summary>
+    /// MetricsLogValues is variation of FormattedLogValues that accepts a dictionary to store customDimensions Application Insights.
```
michaelcfanning: >customDimensions Application Insights [](start = 95, length = 37)

Comment needs a little grammar/other attention.

src/WorkItems/Logging/ApplicationInsightsTelemetryInitializer.cs
rtaket: I think the risk of mixing the operation ids is that there's something else logging telemetry (perhaps capturing perf counters) and they are setting their own operation id (there's a chain of initializers, so they could insert themselves in the chain.) In this case, the operation ids are supposed to be different.

---
In reply to: [396832352](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396832352) [](ancestors = 396832352)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -38,6 +42,8 @@ public SarifWorkItemFiler(Uri filingUri = null, SarifWorkItemContext filingConte
             }
 
             this.FilingClient = FilingClientFactory.Create(this.FilingContext.HostUri);
```
michaelcfanning: >FilingClientFactory.Create [](start = 32, length = 26)

In future, we could have an IFilingClientFactory service. This might be useful in helping to generalize a mocking story. Probably not required now, just would introduce another interface type...

src/WorkItems/Logging/EventIds.cs
rtaket: We should come up with a convention. This was just an example. I don't expect it to survive when we add the real log messages. i.e. I don't think the LogToProcess metrics are interesting.

---
In reply to: [396832563](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396832563) [](ancestors = 396832563)

src/WorkItems/Logging/LoggerExtensions.cs
rtaket: Yes. it loosely follows the existing pattern. If we add the message at the end, I think it'll be a bit confusing because it'll be out-of-order compared to the existing ILogger extensions.

---
In reply to: [396834358](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396834358) [](ancestors = 396834358)

src/WorkItems/Logging/MetricsLogValues.cs
rtaket: fixed

---
In reply to: [396834511](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396834511) [](ancestors = 396834511)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
rtaket: That's a good idea. I've noticed a few places where services could be injected. 

---
In reply to: [396835947](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396835947) [](ancestors = 396835947)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -79,65 +88,82 @@ public virtual void FileWorkItems(string sarifLogFileContents)
...
 
-            SplittingStrategy splittingStrategy = this.FilingContext.SplittingStrategy;
-            if (splittingStrategy == SplittingStrategy.None)
-            {
-                FileWorkItemsHelper(sarifLog, this.FilingContext, this.FilingClient);
-                return;
-            }
-
-            IList<SarifLog> logsToProcess = new List<SarifLog>(new SarifLog[] { sarifLog });
+                Logger.LogInformation("Connecting to filing client: {accountOrOrganization}", this.FilingClient.AccountOrOrganization);
```
michaelcfanning: >Logger.LogInformatio [](start = 16, length = 20)

This logging doesn't have an event id. Is that considered a poor practice?

rtaket: I think we're using the logging for two different purposes. One is for capturing specific events. e.g. the call to create the work item took 150 ms. The other purpose is for tracing. I think we need events ids for the former, but not for the latter. And I think this example is the latter, tracing. Perhaps this should have been a LogDebug or LogTrace instead. But I didn't want to set the default trace level to Debug or Trace.

---
In reply to: [396837768](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r396837768) [](ancestors = 396837768)

src/WorkItems/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,2 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
rtaket: >Licensed [](start = 3, length = 8)

You can probably delete this file.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
michaelcfanning: ok! Yes, maybe rename or refactor these methods slightly. Later.

---
In reply to: [397177520](https://github.com/microsoft/sarif-sdk/pull/1820#discussion_r397177520) [](ancestors = 397177520,396837768)

PR #1822, opened 2020-03-23
Loosen timing restrictions on regions caching test.

michaelcfanning:
@ScottLouvau.  This is failing non-deterministically in some agent environments. I've loosened the timing window and we now emit the actual timing observed in log file.

@lgolding @rtaket @nickmarston @cfaucon 

Could use a review on this. The test that's modified is causing random pipeline failures, so good to resolve asap.

### DIFF
--- a/src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
+++ b/src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
@@ -480,7 +480,7 @@ namespace Microsoft.CodeAnalysis.Sarif.UnitTests
 
             // Runtime should be way under 100ms if caching, and much longer otherwise
             w.Stop();
-            Assert.True(w.ElapsedMilliseconds < 100);
+            Assert.True(w.ElapsedMilliseconds < 150, $"Time observed was {w.ElapsedMilliseconds}. Expected < 150 ms.");
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
```diff
@@ -478,9 +478,11 @@ public void FileRegionsCache_ProperlyCaches()
                 Region populated = fileRegionsCache.PopulateTextRegionProperties(copy, uri, populateSnippet: true);
             }
 
+            int timeThreshold = 150;
```
lgolding: `const`?

michaelcfanning: :v:

ScottLouvau: Got it. Sorry about that. I need to find a new model for performance tests which are probably relative to measured performance of something else just run in the same environment.

michaelcfanning: Side note: in the agent 'wild', this code path is taking closer to 200 ms. You have a fine idea, though, we could update the test to run a baseline operation (no caching), then repeat with caching. We should see a substantive difference that's roughly consistent in relative terms. Will look at that.

---
In reply to: [397266682](https://github.com/microsoft/sarif-sdk/pull/1822#discussion_r397266682) [](ancestors = 397266682)

PR #1823, opened 2020-03-23
Modify description body to new template.

nickmarston:


src/Test.Utilities.Sarif/TestData.cs
```diff
@@ -54,6 +54,13 @@ public static SarifLog CreateOneIdThreeLocations()
                     new Run
                     {
                         Tool = CreateSimpleLog().Runs[0].Tool,
+                        VersionControlProvenance = new []
+                        {
+                            new VersionControlDetails
+                            {
+                                RepositoryUri = new Uri(@"https://bugfiler.visualstudio.com/")
```
michaelcfanning: >https://bugfiler.visualstudio.com/ [](start = 58, length = 34)

use [https://bugfiler.example.com](https://bugfiler.example.com) for this URL. example.com is a reserved domain for illustrative purposes in docs (and code).

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,19 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
 
-            foreach (KeyValuePair<string, int> toolResult in resultCountsByTool)
+            foreach (KeyValuePair<Run, int> run in resultCountsByRun)
             {
-                templateText.Append(string.Format("*Tool: {0}", toolResult.Key));
-                templateText.AppendLine();
-                templateText.Append(string.Format("     Result count: {0}", toolResult.Value));
-                templateText.AppendLine();
+                Uri detectionLocation = !string.IsNullOrEmpty(run.Key?.VersionControlProvenance?.FirstOrDefault().RepositoryUri?.OriginalString) ? run.Key?.VersionControlProvenance?.FirstOrDefault().RepositoryUri : locationUri;
+                templateText = new StringBuilder(string.Format(@"This work item contains {0} {1} issue(s) detected in {2}.  ", run.Value, run.Key?.Tool?.Driver?.Name, detectionLocation.OriginalString));
```
michaelcfanning: rewrite these lines so that they are more readable?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,19 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
         {
-            Dictionary<string, int> resultCountsByTool = ComputeToolResultCounts(log);
-            StringBuilder templateText = new StringBuilder(@"This bug contains results for the below tool(s) and issue(s).  It was filed automatically.");
-            templateText.AppendLine();
-            templateText.AppendLine();
+            Dictionary<Run, int> resultCountsByRun = ComputeRunResultCounts(log);
+            StringBuilder templateText = new StringBuilder();
 
-            foreach (KeyValuePair<string, int> toolResult in resultCountsByTool)
+            foreach (KeyValuePair<Run, int> run in resultCountsByRun)
```
michaelcfanning: >resultCountsByRun [](start = 51, length = 17)

This helper looks like it creates an undesirable run-on sentence when used with multiple tools. 

This work items contains X Tool1 issues detected in 'Blah'. Click the 'Scans' tab to review results. This work items contains X Tool2 issues detected in 'Baz'. Click the 'Scans' tab to review results. This work items contains X Tool3 issues detected in 'Bar'. Click the 'Scans' tab to review results. 

Instead, consider aggregating the result count and gathering all the tool names into a single sentence. 

This work item contains X 'FxCop', 'BinSkim' and 'Fortify' issue(s) detected in 'blah.cs' (and other locations). Click the 'Scans' tab to review results."

Write a little helper that will concatenate your tool names. Be sure to enquote these with apostrophes. Be sure to enquote the location with apostrophes as well. The text 'and other locations' would only be appended in the multi-run case.

src/Test.Utilities.Sarif/TestData.cs
lgolding: I don't see why Nick can't use it. See [https://www.iana.org/domains/reserved](https://www.iana.org/domains/reserved). "Reserved" just means "not available for registration or transfer", _i.e._, nobody can own it.

---
In reply to: [397236588](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r397236588) [](ancestors = 397236588)

lgolding:
>         public SarifWorkItemModel(SarifLog sarifLog, SarifWorkItemContext context = null)

It is well worth a comment -- I had to puzzle over this for a couple of minutes -- that this `SarifLog` contains exactly a set of results that are intended to be filed as a single work item (and this log will be attached to the work item). I couldn't understand on Line 30 what was special about "the first artifact location that has a URI".

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:18 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

lgolding:
>                     break;

You need to be careful here. `run.artifacts` can contain artifacts other than analysis targets, for example, configuration files, debug logs, or external property files. See §3.24.6, `artifact.roles` property.

If you know that the creator of the `SarifLog` only included analysis targets in `run.artifacts`, or if the creator explicitly populated `artifact.roles` so you could look at only those with the `"analysisTarget"` role, you'd be golden. I recommend (1) a comment to that effect, and (2) that we fix the upstream code (the code that creates the incoming per-bug logs) to populate `artifact.roles` for those artifacts it knows to be analysis targets.

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:31 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

lgolding:
I'm having to refresh my memory of this code, so making some comments unrelated to your current changes -- which I think are valid, but which in the context of this PR you can take or leave. You might want to file bugs for some of them.

src/Sarif.WorkItems/SarifWorkItemModel.cs
```diff
@@ -83,9 +83,14 @@ private string CreateBugFooter()
             if(this.Context.CurrentProvider == FilingClient.SourceControlProvider.AzureDevOps)
             {
                 StringBuilder azureDevOpsFooter = new StringBuilder();
-                azureDevOpsFooter.Append(@"To see result details, please visit the Scans tab of this bug, or the attached SARIF log.");
+                azureDevOpsFooter.Append(@"<br><br>Other viewing options:<br>");
```
lgolding: ><br> [](start = 43, length = 4)

I take it ADO's work item text fields accept HTML rather than Markdown. I seem to remember us talking about that a few days ago.

lgolding: ><br><br> [](start = 43, length = 8)

Sometimes you choose `<br>` and sometimes an extra call to `AppendLine`. Is there a principled reason to choose one over the other on a case-by-case basis? If so it's worth a comment.

lgolding:
>                 Text = JsonConvert.SerializeObject(sarifLog),

This will produce unformatted JSON. If you're doing that to save space, that's fine (but worth a comment). If you want users to be able to easily open the attachment in their IDE without having to reformat it.

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:45 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemModel.cs
```diff
@@ -83,9 +83,14 @@ private string CreateBugFooter()
...
                 azureDevOpsFooter.AppendLine();
-                azureDevOpsFooter.Append(@"If the scans tab is missing or unavailable, please install the SARIF viewer from https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer");
+                azureDevOpsFooter.AppendLine();
+                azureDevOpsFooter.Append(@"<li>Examine the complete <a href=""https://dev.azure.com/office/Office/_componentGovernance/Office?_a=alerts&typeId=1731351&alerts-view-option=active"">CG Scan</a> for this repository.</li>");
+                azureDevOpsFooter.AppendLine();
```
lgolding: So not only does this mention CG, but it mentions a _specific scan_ :) Is there actually a URL you can usefully put here?

src/Sarif.WorkItems/SarifWorkItemModel.cs
```diff
@@ -83,9 +83,14 @@ private string CreateBugFooter()
...
+                azureDevOpsFooter.Append(@"<li>Load the attached log file in the <a href=""https://marketplace.visualstudio.com/_apis/public/gallery/publishers/WDGIS/vsextensions/MicrosoftSarifViewer/2.1.7/vspackage"">CG Scan</a> Visual Studio SARIF add-in.</li>");
```
lgolding: >Visual Studio SARIF add-i [](start = 230, length = 25)

This is what wants to be in the `<a>...</a>` element, instead of "CG Scan"; likewise on the next line.

src/Sarif.WorkItems/SarifWorkItemModel.cs
nickmarston: This is strongly related to the above.  As this is the ADO footer, and it does support HTML, I'll remove the appendLines here.  I do use them elsewhere in more genereal (non-HTML enabled text) where I think they make sense.

---
In reply to: [398187126](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398187126) [](ancestors = 398187126)

src/Sarif.WorkItems/SarifWorkItemModel.cs
nickmarston: I'm...not sure.  This is in the template example, but I would need the Sarif log to contain this URI.  A cursory search by me did not find it, but if it exists then that should be what goes here.

---
In reply to: [398187964](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398187964) [](ancestors = 398187964)

src/Sarif.WorkItems/SarifWorkItemModel.cs
nickmarston: yeah, ADO is HTML, github is markdown.  I think the long term solution is to do all markdown and then pass it through a markdown->html converter, which I've seen a couple of.  We'd just want to evaluate which one we would use.

---
In reply to: [398186438](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398186438) [](ancestors = 398186438)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -74,22 +75,22 @@ public static string CreateWorkItemTitle(this Run run)
...
         {
             if (log == null) { throw new ArgumentNullException(nameof(log)); }
             if (log.Runs == null) { throw new ArgumentNullException(nameof(log.Runs)); }
 
-            var resultCountsByTool = new Dictionary<string, int>();
+            var resultCountsByRun = new Dictionary<Run, int>();
 
             foreach (Run run in log?.Runs)
             {
                 if (run != null && run.Results != null)
```
lgolding: >run != null && run.Results != null [](start = 20, length = 34)

With null coalescing operator: `if (run?.Results != null)`

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
 
-        public static string CreateWorkItemDescription(this SarifLog log)
+        public static string CreateWorkItemDescription(this SarifLog log, Uri locationUri)
         {
-            Dictionary<string, int> resultCountsByTool = ComputeToolResultCounts(log);
-            StringBuilder templateText = new StringBuilder(@"This bug contains results for the below tool(s) and issue(s).  It was filed automatically.");
-            templateText.AppendLine();
-            templateText.AppendLine();
+            Dictionary<Run, int> resultCountsByRun = ComputeRunResultCounts(log);
+            StringBuilder templateText = new StringBuilder();
```
lgolding: >StringBuilder [](start = 12, length = 13)

NIT: `var`

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
+            int runningResults = 0;
+            string toolNames = string.Empty;
+            string multipleToolsFooter = string.Empty;
+
+            Uri runRepositoryUri = resultCountsByRun.FirstOrDefault().Key?.VersionControlProvenance?.FirstOrDefault().RepositoryUri;
```
lgolding: >. [](start = 69, length = 1)

`?.` here as well, since the whole point of `FirstOrDefault` is that it returns `null` if the collection is empty.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
+            string detectionLocation = !string.IsNullOrEmpty(runRepositoryUri?.OriginalString) ? runRepositoryUri?.OriginalString : locationUri?.OriginalString;
```
lgolding: >detectionLocation [](start = 19, length = 17)

BUG? What has the repo URI (_e.g_, `[https://github.com/microsoft/sarif-sdk`](https://github.com/microsoft/sarif-sdk`)) to do with the detection location(_e.g_, `src\Sarif\Cache.cs`)?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
+
+            toolNames = string.Format("'{0}'", resultCountsByRun.FirstOrDefault().Key?.Tool?.Driver?.Name);
```
lgolding: >toolNames [](start = 12, length = 9)

Since the expression `resultCountsByRun.FirstOrDefault().Key?.Tool?.Driver?.Name` can end up `null` -- or at least, your code thinks it needs to anticipate that -- you can't safely pass it to `string.Format`.

`resultCountsByRun.FirstOrDefault().Key?.Tool?.Driver?.Name ?? string.Empty`

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: That's directly from Mike's template, it's the general pointer to what was being scanned.  If it's null, then we use the filename from the location.  Since the scan is likely to cover multiple issues over multiple files, I agree that this makes a good general description assuming it's present.

---
In reply to: [398214479](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398214479) [](ancestors = 398214479)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: Good point.  Definitely several more cases of this or similar things with string.Join below.

---
In reply to: [398216527](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398216527) [](ancestors = 398216527)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
 
-            foreach (KeyValuePair<string, int> toolResult in resultCountsByTool)
+            if (resultCountsByRun.Count > 1)
             {
-                templateText.Append(string.Format("*Tool: {0}", toolResult.Key));
-                templateText.AppendLine();
-                templateText.Append(string.Format("     Result count: {0}", toolResult.Value));
-                templateText.AppendLine();
+                Run lastrun = resultCountsByRun.Last().Key;
+                multipleToolsFooter = " (and other locations)";
```
lgolding: >" (and other locations)" [](start = 38, length = 24)

All of these user-facing strings need to be in a resource file.

But be aware that constructing a sentence by pasting together phrases is in general not localizable because of differences among language grammars. In this case it looks like you might get away with it, but just keep it in mind.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
+                foreach (KeyValuePair<Run, int> run in resultCountsByRun)
+                {
+                    if (run.Key == resultCountsByRun.First().Key)
+                    {
+                        continue;
+                    }
+                    else if (run.Key == lastrun)
+                    {
+                        toolNames = string.Join(" and ", toolNames, string.Format("'{0}'", run.Key?.Tool?.Driver?.Name));
+                        runningResults += run.Value;
```
lgolding: Oh, I see, here's the Oxford comma issue. ;-) I'm not going to fight it.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -103,20 +104,47 @@ private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingD
...
-            foreach (KeyValuePair<string, int> toolResult in resultCountsByTool)
+            if (resultCountsByRun.Count > 1)
             {
-                templateText.Append(string.Format("*Tool: {0}", toolResult.Key));
-                templateText.AppendLine();
-                templateText.Append(string.Format("     Result count: {0}", toolResult.Value));
-                templateText.AppendLine();
+                Run lastrun = resultCountsByRun.Last().Key;
+                multipleToolsFooter = " (and other locations)";
+                foreach (KeyValuePair<Run, int> run in resultCountsByRun)
```
lgolding: >run [](start = 48, length = 3)

Let's use a different variable name since `run` would usually be, well, a `Run`. I suggest `pair`, or I sometimes use `entry` (as in "dictionary entry").

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemModelTests.cs
```diff
@@ -36,7 +37,23 @@ public void SarifWorkItemModel_PopulatesADODescription()
             var workItemModel = new SarifWorkItemModel(sarifLog, context);
             workItemModel.BodyOrDescription.Should().NotBeNullOrEmpty();
             workItemModel.BodyOrDescription.Should().Contain(nameof(TestData.TestToolName));
-            workItemModel.BodyOrDescription.Should().Contain("To see result details, please visit the Scans tab of this bug, or the attached SARIF log.");
+            workItemModel.BodyOrDescription.Should().Contain(sarifLog.Runs[0].VersionControlProvenance[0].RepositoryUri.OriginalString);
+            workItemModel.BodyOrDescription.Should().Contain("Visual Studio SARIF add-in.");
```
lgolding: >"Visual Studio SARIF add-in." [](start = 61, length = 29)

Another reason for this to be in a resource file: If you change it, your test doesn't start failing.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemModelTests.cs
```diff
@@ -36,7 +37,23 @@ public void SarifWorkItemModel_PopulatesADODescription()
...
+        }
+
+        [Fact]
+        public void SarifWorkItemModel_MultipleToolsADODescription()
```
lgolding: >ADO [](start = 52, length = 3)

.NET style guidelines: acronyms have only the first letter capitalized, so `Ado`. (Same on previous test.)

As for the test name, a style that I advocate (and I've persuaded Michael that it's a good one) is that the test name describes the behavior: `Car_CanStart`. Sometimes there is a condition in the name `Car_WhenOutOfGas_CannotStart`. This general naming convention comes from Behavior-Driven Development (BDD), FWIW.

So here, perhaps `SarifWorkItemModel_IncorporatesMultipleToolNamesIntoAdoDescription`.

nickmarston: Michael did mention this, I haven't internalized it yet, but agree with the convention.
Stealing your name for this one.

---
In reply to: [398221599](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398221599) [](ancestors = 398221599)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: Thanks for the explanation. I now understand the rationale for including the repo URI. But as the code stands, if repo URI is present, you'll use it _even if there's only one location_. You can't tell at this point in the code if there's more than one location, and you need to know to decide whether to mention the repo URL.

---
In reply to: [398216755](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398216755) [](ancestors = 398216755,398214479)

lgolding:
>                     break;

The problem here is that the visitor extracts _all_ artifact locations, and we only want those that occur as values of a `result.locations[].physicalLocation.artifactLocation` property.

---
In reply to: [604099383](https://github.com/microsoft/sarif-sdk/pull/1823#issuecomment-604099383) [](ancestors = 604099383)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:31 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

src/Test.Utilities.Sarif/TestData.cs
nickmarston: Since this was an example, I had no strong preference.  Nice to know there's a domain dedicated to this.  Earlier versions used a visualstudio.com domain which might well be picked up.

---
In reply to: [398170991](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398170991) [](ancestors = 398170991,397236588)

nickmarston:
>                 Text = JsonConvert.SerializeObject(sarifLog),

I just did this, in general the snippets are not that large and I think it makes sense.

---
In reply to: [604102450](https://github.com/microsoft/sarif-sdk/pull/1823#issuecomment-604102450) [](ancestors = 604102450)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:45 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

nickmarston:
>                     break;

I've talked about this one with Mike previously, it needs further discussion but the previous discussion have all indicated those are more exception circumstances we will solve later after we have some of the basics in place.

---
In reply to: [604134008](https://github.com/microsoft/sarif-sdk/pull/1823#issuecomment-604134008) [](ancestors = 604134008,604099383)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemModel.cs:31 in 06fdbd9. [](commit_id = 06fdbd92046ad9bae486d5c16c8be8385fc9e3d0, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: If there's a bug here, it's very much as specc'ed.  The design is to default to the repo URL, and only mention other locations if that is empty, even in the single location case.  I believe the intent of the design is that the Sarif attachement will contain the specifics needed for the location.

---
In reply to: [398223529](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398223529) [](ancestors = 398223529,398216755,398214479)

michaelcfanning: First, of course, the specification may have rough edges to thinking. :) The rationale here is that for a single location issue, the title already contains the issue. And so it is more valuable in the description to add information about what repo/branch/commit was scanned. Note that this information isn't surfaced in any of the viewers, you have to look at the raw SARIF to get it. So worth raising visibility on it if available. If it isn't, we recap the first location (and specifically call out if there are other locations to examine, without listing them in detail).

---
In reply to: [398271699](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398271699) [](ancestors = 398271699,398223529,398216755,398214479)

src/Sarif.WorkItems/WorkItemsResources.resx
```diff
@@ -117,7 +117,28 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="AdoSARIFAddInMessage" xml:space="preserve">
```
lgolding: >SARIF [](start = 17, length = 5)

`Sarif` in code per .NET coding guidelines; "SARIF" (all caps) in English text.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: I'm sorry, I pointed my comment at the wrong place. It's the output of `FirstOrDefault()` that might be `null`, so you have to say `...FirstOrDefault()?.Key?.VersionControlProvenance?...`.

I see from looking at  `ComputeRunResultsCount` that `resultCountsByRun` _cannot_ be `null`, so you don't have to write `resultCountsByRun?.`.

Finally, if a dictionary entry does exist, then `Key` cannot be `null`.

So in summary: `resultCountsByRun.FirstOrDefault()?.Key.VersionControlProvenance?.FirstOrDefault()?.RepositoryUri`

Note the `?.` after _both_ calls to `FirstOrDefault()` in this expression.

---
In reply to: [398195637](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398195637) [](ancestors = 398195637)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: Marking "Won't Fix" since we will surely give more thought to this question, especially as customers comment on the quality of the bugs we file.

---
In reply to: [398679769](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398679769) [](ancestors = 398679769,398271699,398223529,398216755,398214479)

src/Sarif.WorkItems/SarifWorkItemModel.cs
lgolding: This is now moot with your new resource strings.

---
In reply to: [398191142](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398191142) [](ancestors = 398191142,398187964)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: I was wrong. This is correct as is. `KeyValuePair` is a value type, so can't be `null` -- but the `Key` _can_ be `null`, so your code is exactly right.

---
In reply to: [398745325](https://github.com/microsoft/sarif-sdk/pull/1823#discussion_r398745325) [](ancestors = 398745325,398195637)

PR #1826, opened 2020-03-24
Do not include absent results when baselining

rtaket:
When you call the match-results-forward verb on the multitool, and the current SARIF file has absent results, the baseline file identifies these as new. This is because when we gather the results from the current file, we assume that all the results are present. To fix this, I excluded the absent results from the list of current results.


src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -46,7 +46,7 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
                 // Baseline result and current result have been matched => existing.
                 result = ConstructExistingResult(resultMatchingProperties, out originalResultMatchingProperties);
             }
-            else if (PreviousResult == null && CurrentResult != null)
+            else if (PreviousResult == null && CurrentResult != null && CurrentResult.Result.BaselineState != BaselineState.Absent)
```
ScottLouvau: Do you need this after you added the code to only construct ExtractedResults for the ones which aren't absent?

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -175,8 +175,8 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns,
                 {
                     foreach (Result result in run.Results)
                     {
-                        // Include all Results except Absent results in the baseline
-                        if (!(isBaselineRun && result.BaselineState == BaselineState.Absent))
+                        // Include all Results except Absent results
+                        if (result.BaselineState != BaselineState.Absent)
```
ScottLouvau: This is a sensible way to go about things. 
A user complained about us ignoring Absent results in the baseline; he keeps them in his log and wants the baseline to have a rolling history (which I don't in SarifBaseliner).

If you're using Absent results in the current log to mean something, and want them to persist through to the output log, maybe you want them to come through but to be emitted in the state Absent as well.

This would mean that if you have a clear check that a condition is gone, you can emit an Absent result. Rather than hoping baselining will not match the old copy of the result to anything (and will therefore report the copy of the result from the Baseline as Absent) it would match the result with the older one, but see that the new one is Absent and write it out that way.

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
rtaket: Sorry, this was where I was going to make the fix originally. I'll remove this change.

---
In reply to: [397397692](https://github.com/microsoft/sarif-sdk/pull/1826#discussion_r397397692) [](ancestors = 397397692)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
michaelcfanning: The baseline is not an appropriate place for persisting complete history for many reasons you both understand. OTOH, it's quite reasonable for a current baseline to contain absent messages (though that have gone absent due to direct comparison to the prior baseline). What we should do is actively flush all absent messages from the baseline before beginning the diff.

---
In reply to: [397400299](https://github.com/microsoft/sarif-sdk/pull/1826#discussion_r397400299) [](ancestors = 397400299)

PR #1829, opened 2020-03-25
Fix #1774. Updates SARIF log file with filed work item details. Incid…

### LINKED ISSUE #1774: Persist filed work item details to input SARIF log.
This functionality s/be added to the FileWorkItemsCommand. That means the SarifWorkItemFiler needs to be updated to either return all the final work item models that it processed from a relevant method or to persist this data as a new property on the class.

michaelcfanning:
…entally updates language support to 8.0.

@rtaket @nickmarston @alisonmaclellan


src/Sarif/Visitors/PartitioningVisitor.cs
```diff
@@ -174,7 +174,7 @@ public override Result VisitResult(Result node)
 
                 Result partitionedResult = deepClone
                     ? node.DeepClone()
-                    : new Result(node);
```
michaelcfanning: >new Result(node); [](start = 22, length = 17)

This code created a new Result reference but then shared properties from the original result (which could potentially be mutable). Instead, let's just reuse the result. This allows maximal reuse of the object. In our case, we update the shared result with the work item filing details and this therefore flows directly back to the original log file (even though we're filing against a partition of it).

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -66,20 +67,21 @@ public int Run(FileWorkItemsOptions options, IFileSystem fileSystem)
                 filingContext.DataToInsert = options.DataToInsert.ToFlags();
             }
 
-            using (var filer = new SarifWorkItemFiler(filingUri: filingContext.HostUri, filingContext: filingContext))
+            var filer = new SarifWorkItemFiler(filingContext.HostUri, filingContext);
```
rtaket: >filer [](start = 16, length = 5)

You should probably still dispose of the filer.

rtaket:
>         public virtual void FileWorkItems(Uri sarifLogFileLocation)

Shouldn't this overload have the 'out SarifLog' param too?

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:60 in d7b30e1. [](commit_id = d7b30e1ac77172771e1e51d3b6857357f518963d, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -194,13 +197,23 @@ private void FileWorkItemsHelper(SarifLog sarifLog, SarifWorkItemContext filingC
...
-                //
-                //       https://github.com/microsoft/sarif-sdk/issues/1774
+                foreach (Run run in sarifLog.Runs)
+                {
+                    if (run.Results == null) { continue; }
+
+                    foreach (Result result in run.Results)
+                    {
+                        result.WorkItemUris ??= new List<Uri>();
+                        result.WorkItemUris.Add(workItemModel.HtmlUri);
```
rtaket: >. [](start = 30, length = 1)

It's non-obvious that this is mutating the original sarif file (at least I presume it is.) You might add a comment.

michaelcfanning:
>         public virtual void FileWorkItems(Uri sarifLogFileLocation)

I made all these overloads just return the updated log. This will allow us to do things like pass a boolean in future to update input log in place or to rewrite from a deep clone. All these methods returned void, so it's a natural change.

---
In reply to: [604061506](https://github.com/microsoft/sarif-sdk/pull/1829#issuecomment-604061506) [](ancestors = 604061506)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:60 in d7b30e1. [](commit_id = d7b30e1ac77172771e1e51d3b6857357f518963d, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -78,11 +78,12 @@ public virtual void FileWorkItems(Uri sarifLogFileLocation)
             throw new ArgumentException($"Specified URI was not an absolute file URI: {sarifLogFileLocation}");
         }
 
-        public virtual void FileWorkItems(string sarifLogFileContents)
+        public virtual void FileWorkItems(string sarifLogFileContents, out SarifLog sarifLog)
```
nickmarston: I might be missing something here, but why an out parameter vs. making it the method return?  We overwrite the parameter without checking it almost immediately, so to me this is confusing as it gives the impression we may be using it in some way but, in fact, do not.  In this case I'd prefer to make it a returned value.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
michaelcfanning: +1 done.

---
In reply to: [398143633](https://github.com/microsoft/sarif-sdk/pull/1829#discussion_r398143633) [](ancestors = 398143633)

rtaket:
>         public virtual void FileWorkItems(Uri sarifLogFileLocation)

Hmm, that's weird. I think it's confusing. I don't expect a method named FileWorkItems() to return a log file. ... Is this essentially an internal method? If so, it probably doesn't matter. If not, I'd find a different pattern. Maybe don't provide an overload that accepts a Uri or string contents.

---
In reply to: [604065709](https://github.com/microsoft/sarif-sdk/pull/1829#issuecomment-604065709) [](ancestors = 604065709,604061506)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:60 in d7b30e1. [](commit_id = d7b30e1ac77172771e1e51d3b6857357f518963d, deletion_comment = False)

PR #1830, opened 2020-03-26
Add repro steps field to work item patch

rtaket:


src/WorkItems/FilingClientFactory.cs
```diff
@@ -26,7 +26,7 @@ public static FilingClient Create(Uri filingHostUri)
                 if (match.Success)
                 {
                     filingClient = isGitHub ? (FilingClient)new GitHubFilingClient() : new AzureDevOpsFilingClient();
-                    filingClient.CurrentProvider = isGitHub ? FilingClient.SourceControlProvider.Github : FilingClient.SourceControlProvider.Github;
+                    filingClient.CurrentProvider = isGitHub ? FilingClient.SourceControlProvider.Github : FilingClient.SourceControlProvider.AzureDevOps;
```
michaelcfanning: >CurrentProvider [](start = 33, length = 15)

Do me a favor, change this property name to 'Provider' since I don't think we allow swapping providers in an out w/a common instance. And so 'Current' is a bit of a misqualification.

michaelcfanning:
>                 var patchDocument = new JsonPatchDocument

This patch document isn't populating the standard Description field. 'ReprosSteps' is an optional field that happens to be the default for the accounts we're working with. Probably, what we should be doing is trying to query the server to get a sense of which field we should be populating.

---
Refers to: src/WorkItems/AzureDevOpsFilingClient.cs:86 in b1a9a56. [](commit_id = b1a9a5659a3eb67dd1b74d583901e62056f94caf, deletion_comment = False)

src/WorkItems/AzureDevOpsFilingClient.cs
```diff
@@ -102,6 +102,12 @@ public override async Task<IEnumerable<WorkItemModel>> FileWorkItems(IEnumerable
                         Operation = Operation.Add,
                         Path = $"/fields/{AzureDevOpsFieldNames.Tags}",
                         Value = string.Join(",", workItemModel.LabelsOrTags)
+                    },
+                    new JsonPatchOperation
+                    {
+                        Operation = Operation.Add,
+                        Path = $"/fields/{AzureDevOpsFieldNames.ReproSteps}",
```
michaelcfanning: >ReproSteps [](start = 64, length = 10)

What will happen here if Repro steps isn't enabled in the target account?

michaelcfanning:
There no update to the mock unit test in this change. The issue you found also points to a gap in the filing client factory unit tests (we didn't add a test to ensure the CurrentProvider property was set. Maybe @nickmarston would be interested in closing these gaps.

src/WorkItems/FilingClientFactory.cs
rtaket: Renamed to Provider.

---
In reply to: [398645608](https://github.com/microsoft/sarif-sdk/pull/1830#discussion_r398645608) [](ancestors = 398645608)

src/WorkItems/AzureDevOpsFilingClient.cs
rtaket: I believe it's still added to the work item. i.e. AFAIK, it will still save successfully.

---
In reply to: [398651812](https://github.com/microsoft/sarif-sdk/pull/1830#discussion_r398651812) [](ancestors = 398651812)

rtaket:
>                 var patchDocument = new JsonPatchDocument

Yes, that would be good. I only added ReproSteps because it looked like we are trying to log Bugs. If that's not a requirement, I believe Task + System.Description is a safer choice because I think it's part of all the out-of-box process templates.

---
In reply to: [604484098](https://github.com/microsoft/sarif-sdk/pull/1830#issuecomment-604484098) [](ancestors = 604484098)

---
Refers to: src/WorkItems/AzureDevOpsFilingClient.cs:86 in b1a9a56. [](commit_id = b1a9a5659a3eb67dd1b74d583901e62056f94caf, deletion_comment = False)

PR #1831, opened 2020-03-26
Refactor body description method into multiple parts.

nickmarston:
Refactors the template body description for work items into several helper methods, and enhances the testing associated with them.

### DIFF
--- a/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
+++ b/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
@@ -83,7 +83,7 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItems
                     (locationName == null ? "" : " (in " + locationName + ")");
         }
 
-        public static List<string> GetRunToolNames(this SarifLog log)
+        public static List<string> GetToolNames(this SarifLog log)
         {
             return log?.Runs?.Where(r => r?.Results?.Count > 0).Select(r => r.Tool?.Driver?.Name).Distinct().ToList();
         }
@@ -93,44 +93,23 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItems
             return log?.Runs?.Select(r => r?.Results?.Count).Sum() ?? 0;
         }
 
-        public static string PhraseToolNames(List<string> toolNames)
-        {
-            toolNames = toolNames.Where(r => !string.IsNullOrWhiteSpace(r)).ToList();
-            if (toolNames.Count == 0)
-            {
-                return string.Empty;
-            }
-            else if (toolNames.Count == 1)
-            {
-                return string.Format("'{0}'", toolNames[0]);
-            }
-            else
-            {
-                string phrasedToolNames = string.Format("'{0}'", toolNames[0]);
-                for (int j = 1; j < toolNames.Count - 1; j++)
-                {
-                    phrasedToolNames = string.Join(", ", phrasedToolNames, string.Format("'{0}'", toolNames[j]));
-                }
-                phrasedToolNames = string.Join(" and ", phrasedToolNames, string.Format("'{0}'", toolNames[toolNames.Count - 1]));
-                return phrasedToolNames;
-            }
-        }
+        
 
         public static string CreateWorkItemDescription(this SarifLog log, Uri locationUri)
         {
-            List<string> toolNames = log.GetRunToolNames();
             int totalResults = log.GetAggregateResultCount();
-            string phrasedToolNames = PhraseToolNames(toolNames);
+            List<string> toolNames = log.GetToolNames();
+            string phrasedToolNames = toolNames.ToAndPhrase();
             string multipleToolsFooter = toolNames.Count > 1 ? WorkItemsResources.MultipleToolsFooter : string.Empty;
 
             Uri runRepositoryUri = log?.Runs.FirstOrDefault()?.VersionControlProvenance?.FirstOrDefault().RepositoryUri;
             string detectionLocation = !string.IsNullOrEmpty(runRepositoryUri?.OriginalString) ? runRepositoryUri?.OriginalString : locationUri?.OriginalString;
           
-            var descriptionText = new StringBuilder(string.Format(WorkItemsResources.WorkItemBodyTemplateText, totalResults, phrasedToolNames, detectionLocation, multipleToolsFooter));
-            descriptionText.Append(WorkItemsResources.ViewScansTabResults);
-            descriptionText.AppendLine();
+            var descriptionBuilder = new StringBuilder(string.Format(WorkItemsResources.WorkItemBodyTemplateText, totalResults, phrasedToolNames, detectionLocation, multipleToolsFooter));
+            descriptionBuilder.Append(WorkItemsResources.ViewScansTabResults);
+            descriptionBuilder.AppendLine();
 
-            return descriptionText.ToString();
+            return descriptionBuilder.ToString();
         }
 
         private static string ConstructFullRuleIdentifier(ReportingDescriptor reportingDescriptor)
--- a/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
+++ b/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
@@ -83,22 +83,54 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItems
                     (locationName == null ? "" : " (in " + locationName + ")");
         }
 
-        public static Dictionary<Run, int> ComputeRunResultCounts(this SarifLog log)
+        public static List<string> GetRunToolNames(this SarifLog log)
         {
-            if (log == null) { throw new ArgumentNullException(nameof(log)); }
-            if (log.Runs == null) { throw new ArgumentNullException(nameof(log.Runs)); }
+            return log?.Runs?.Where(r => r?.Results?.Count > 0).Select(r => r.Tool?.Driver?.Name).Distinct().ToList();
+        }
 
-            var resultCountsByRun = new Dictionary<Run, int>();
+        public static int GetAggregateResultCount(this SarifLog log)
+        {
+            return log?.Runs?.Select(r => r?.Results?.Count).Sum() ?? 0;
+        }
 
-            foreach (Run run in log?.Runs)
+        public static string PhraseToolNames(List<string> toolNames)
+        {
+            toolNames = toolNames.Where(r => !string.IsNullOrWhiteSpace(r)).ToList();
+            if (toolNames.Count == 0)
             {
-                if (run?.Results != null)
-                {
-                    resultCountsByRun.Add(run, run.Results.Count);
-                }
+                return string.Empty;
             }
+            else if (toolNames.Count == 1)
+            {
+                return string.Format("'{0}'", toolNames[0]);
+            }
+            else
+            {
+                string phrasedToolNames = string.Format("'{0}'", toolNames[0]);
+                for (int j = 1; j < toolNames.Count - 1; j++)
+                {
+                    phrasedToolNames = string.Join(", ", phrasedToolNames, string.Format("'{0}'", toolNames[j]));
+                }
+                phrasedToolNames = string.Join(" and ", phrasedToolNames, string.Format("'{0}'", toolNames[toolNames.Count - 1]));
+                return phrasedToolNames;
+            }
+        }
 
-            return resultCountsByRun;
+        public static string CreateWorkItemDescription(this SarifLog log, Uri locationUri)
+        {
+            List<string> toolNames = log.GetRunToolNames();
+            int totalResults = log.GetAggregateResultCount();
+            string phrasedToolNames = PhraseToolNames(toolNames);
+            string multipleToolsFooter = toolNames.Count > 1 ? WorkItemsResources.MultipleToolsFooter : string.Empty;
+
+            Uri runRepositoryUri = log?.Runs.FirstOrDefault()?.VersionControlProvenance?.FirstOrDefault().RepositoryUri;
+            string detectionLocation = !string.IsNullOrEmpty(runRepositoryUri?.OriginalString) ? runRepositoryUri?.OriginalString : locationUri?.OriginalString;
+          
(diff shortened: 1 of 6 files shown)


src/Test.Utilities.Sarif/TestData.cs
```diff
@@ -243,5 +243,19 @@ public static SarifLog CreateSimpleLog()
...
+        public static SarifLog CreateEmptyRun()
+        {
+            return new SarifLog
+            {
+                Runs = new Run[]
+                {
+                    new Run
+                    {
+
+                    }
```
lgolding: >} [](start = 20, length = 1)

Ultra-nit: remove extra blank line 255.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -52,26 +53,109 @@ public void SarifWorkItemExtensions_CreateWorkItemTitle_HandlesSingleResultWithR
...
+            public List<string> Input { get; }
+            public string ExpectedOutput { get; }
+        }
+
+        private static readonly ReadOnlyCollection<PhraseToolNamesTestCase> s_phraseToolNamesTestCases =
+            new ReadOnlyCollection<PhraseToolNamesTestCase>(new PhraseToolNamesTestCase[]
+            {
+                new PhraseToolNamesTestCase(
+                    input: new List<string>() {""},
+                    expectedOutput: ""),
```
lgolding: >"" [](start = 36, length = 2)

I suggest your helper function emits `''` here instead of nothing, to make it clearer to the reader that something is missing.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -52,26 +53,109 @@ public void SarifWorkItemExtensions_CreateWorkItemTitle_HandlesSingleResultWithR
...
+                    expectedOutput: "'SEMMLE' and 'Binskim'"),
+
+                new PhraseToolNamesTestCase(
+                    input: new List<string>() {"", "CredentialScanner", "SEMMLE", "Binskim"},
+                    expectedOutput: "'CredentialScanner', 'SEMMLE' and 'Binskim'"),
+
+                new PhraseToolNamesTestCase(
+                    input: new List<string>() {"", "CredentialScanner"},
+                    expectedOutput: "'CredentialScanner'"),
+            });
```
lgolding: >} [](start = 12, length = 1)

Nice set of test cases. I suggest you handle the `null` case as well as the empty string case. SARIF requires `tool.driver.name` to be present, but if it's absent and the file hasn't been through the validator, it can slip through.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -52,26 +53,109 @@ public void SarifWorkItemExtensions_CreateWorkItemTitle_HandlesSingleResultWithR
...
+
         [Fact]
-        public void SarifWorkItemExtensions_ComputeToolResultCounts_CountsSingleResult()
+        public void SarifWorkItemExtensions_PhraseToolNames_ConstructPhraseCorrectly()
         {
+            StringBuilder sb = new StringBuilder();
+
+            foreach (PhraseToolNamesTestCase testCase in s_phraseToolNamesTestCases)
+            {
+                string actualOutput = SarifWorkItemsExtensions.PhraseToolNames(testCase.Input);
```
lgolding: >SarifWorkItemsExtensions.PhraseToolNames(testCase.Input) [](start = 38, length = 56)

The point of an extension is to be called as an extension :) so:
```
string actualOutput = testCase.Input.PhraseToolNames();
```

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -83,22 +83,54 @@ public static string CreateWorkItemTitle(this Run run)
                     (locationName == null ? "" : " (in " + locationName + ")");
         }
 
-        public static Dictionary<Run, int> ComputeRunResultCounts(this SarifLog log)
+        public static List<string> GetRunToolNames(this SarifLog log)
```
lgolding: >GetRunToolNames [](start = 35, length = 15)

IMO `GetToolNames` suffices.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -83,22 +83,54 @@ public static string CreateWorkItemTitle(this Run run)
...
+                phrasedToolNames = string.Join(" and ", phrasedToolNames, string.Format("'{0}'", toolNames[toolNames.Count - 1]));
+                return phrasedToolNames;
             }
+        }
+
+        public static string CreateWorkItemDescription(this SarifLog log, Uri locationUri)
+        {
+            List<string> toolNames = log.GetRunToolNames();
+            int totalResults = log.GetAggregateResultCount();
+            string phrasedToolNames = PhraseToolNames(toolNames);
```
lgolding: >string [](start = 12, length = 6)

Here is a nit so small you'll need an electron microscope to see it, yet IMO it's the sort of tiny readability improvement that, when practiced uniformly in the code base, makes a real difference.

On line 121 you assign to `toolNames`, on line 122 you do something unrelated, and on line 123 you _use_ `toolNames`. There's a readability principle that says to limit variable scope as much as possible, because that _minimizes how much stuff you have to remember/keep in mind_.

So if you you reverse the order of lines 121 and 122, you don't introduce `toolName` until you actually need it, and you don't have to remember that it was initialized somewhere above. Sure, here it's just one line of separation, but sometimes it's more. And if you always wait to define something until you need it, you'll reduce the "cognitive load" on your readers.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -83,22 +83,54 @@ public static string CreateWorkItemTitle(this Run run)
...
+            string multipleToolsFooter = toolNames.Count > 1 ? WorkItemsResources.MultipleToolsFooter : string.Empty;
+
+            Uri runRepositoryUri = log?.Runs.FirstOrDefault()?.VersionControlProvenance?.FirstOrDefault().RepositoryUri;
+            string detectionLocation = !string.IsNullOrEmpty(runRepositoryUri?.OriginalString) ? runRepositoryUri?.OriginalString : locationUri?.OriginalString;
+          
+            var descriptionText = new StringBuilder(string.Format(WorkItemsResources.WorkItemBodyTemplateText, totalResults, phrasedToolNames, detectionLocation, multipleToolsFooter));
```
lgolding: >descriptionText [](start = 16, length = 15)

Naming nit: I prefer "builder" in the name when using a string builder. otherwise it looks like you're trying to append to a string variable. So: `descriptionBuilder`.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -83,22 +83,54 @@ public static string CreateWorkItemTitle(this Run run)
...
+        }
 
-            var resultCountsByRun = new Dictionary<Run, int>();
+        public static int GetAggregateResultCount(this SarifLog log)
+        {
+            return log?.Runs?.Select(r => r?.Results?.Count).Sum() ?? 0;
+        }
 
-            foreach (Run run in log?.Runs)
+        public static string PhraseToolNames(List<string> toolNames)
```
lgolding: >PhraseToolNames [](start = 29, length = 15)

This is a general purpose helper, not related to work item filing at all. So:

- Its name should be (for example) `MakePhraseWithAnd`. ("`MakeAndPhrase`" unfortunately sounds like it's doing two things, "making" and "phrasing".)
- Its parameter should be named (for example) `item` or `words`.
- Its local variable/return value should be named (for example) `phraseWithAnd`.
- It should be an extension method on `IEnumerable<string>`.
- It should live in src\Sarif\ExtensionMethods.cs.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -52,26 +53,109 @@ public void SarifWorkItemExtensions_CreateWorkItemTitle_HandlesSingleResultWithR
...
+
+                bool succeeded = (testCase.ExpectedOutput == null && actualOutput == null)
+                    || (actualOutput?.Equals(testCase.ExpectedOutput, StringComparison.Ordinal) == true);
+
+                if (!succeeded)
+                {
+                    sb.AppendLine($"    Input: {testCase.Input} Expected: {testCase.ExpectedOutput} Actual: {actualOutput}");
+                }
+            }
+            sb.Length.Should().Be(0,
```
lgolding: >s [](start = 12, length = 1)

I know you copied the example I showed you. Be aware that the style guidelines will want you to leave a blank line after a closing brace.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
lgolding: On second thought, how about `ToAndPhrase` as the name of the extension method. `words.ToAndPhrase()` seems like it reads nicely.

---
In reply to: [398969489](https://github.com/microsoft/sarif-sdk/pull/1831#discussion_r398969489) [](ancestors = 398969489)

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
nickmarston: This feels a little odd to me, in that I wouldn't expect anything returned for a blank item, but I don't have a strong objection.

---
In reply to: [398963914](https://github.com/microsoft/sarif-sdk/pull/1831#discussion_r398963914) [](ancestors = 398963914)

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
nickmarston: I was wondering where the appropriate place to put this was.  As you commented on elsewhere, the fact that the testcase was not naturally calling it as an extension method seemed to imply something was off.

I definitely don't like MakeAndPhrase simply because I did at a glance read it as two different things.  ToAndPhrase seems fine to me.

---
In reply to: [398994168](https://github.com/microsoft/sarif-sdk/pull/1831#discussion_r398994168) [](ancestors = 398994168,398969489)

src/Test.UnitTests.Sarif/Core/ToolTests.cs
```diff
@@ -69,15 +69,12 @@ public void Tool_ParseFileVersion_ExtractsDottedQuadFileVersion()
...
-                    sb.AppendLine($"    Input: {SafeFormat(testCase.Input)} Expected: {SafeFormat(testCase.ExpectedOutput)} Actual: {SafeFormat(actualOutput)}");
+                    sb.AppendLine($"    Input: {Utilities.SafeFormat(testCase.Input)} Expected: {Utilities.SafeFormat(testCase.ExpectedOutput)} Actual: {Utilities.SafeFormat(actualOutput)}");
                 }
             }
 
             sb.Length.Should().Be(0,
                 $"all test cases should pass, but the following test cases failed:\n{sb.ToString()}");
         }
-
-        private static string SafeFormat(string s)
```
nickmarston: As I'm adding a null check, I have need of this.  It felt like it would be generally useful, so I moved it to the existing utilities class.  If there's a better place, or if you would prefer I just duplicate it, given it's size, just yell and I can undo\move.

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -52,26 +53,114 @@ public void SarifWorkItemExtensions_CreateWorkItemTitle_HandlesSingleResultWithR
...
+                    expectedOutput: "'SEMMLE' and 'Binskim'"),
+
+                new PhraseToolNamesTestCase(
+                    input: new List<string>() {"", "CredentialScanner", "SEMMLE", "Binskim"},
+                    expectedOutput: "'CredentialScanner', 'SEMMLE' and 'Binskim'"),
+
+                new PhraseToolNamesTestCase(
+                    input: new List<string>() {"", "CredentialScanner"},
+                    expectedOutput: "'CredentialScanner'"),
+            });
```
lgolding: >} [](start = 12, length = 1)

Sorry I missed this the first time through: You should have test cases for a `null` list and for an empty list. `null` should probably throw (remember we discussed adding ShouldThrow to the test case class for this situation; your ctor can default it to `false`). Empty list should ... I dunno, up to you.

src/Sarif/ExtensionMethods.cs
```diff
@@ -102,6 +102,29 @@ public static Message ToMessage(this string text)
...
+                return string.Format("'{0}'", words[0]);
+            }
+            else
+            {
+                string phrasedWithAnd = string.Format("'{0}'", words[0]);
+                for (int j = 1; j < words.Count - 1; j++)
+                {
+                    phrasedWithAnd = string.Join(", ", phrasedWithAnd, string.Format("'{0}'", words[j]));
+                }
+                phrasedWithAnd = string.Join(" and ", phrasedWithAnd, string.Format("'{0}'", words[words.Count - 1]));
```
lgolding: > and  [](start = 46, length = 5)

Resource string.

src/Test.UnitTests.Sarif/Core/ToolTests.cs
lgolding: LGTM.

---
In reply to: [399076307](https://github.com/microsoft/sarif-sdk/pull/1831#discussion_r399076307) [](ancestors = 399076307)

src/Sarif/ExtensionMethods.cs
```diff
@@ -102,6 +102,29 @@ public static Message ToMessage(this string text)
             return new Message { Text = text };
         }
 
+        public static string ToAndPhrase(this List<string> words)
```
michaelcfanning: >ToAndPhrase [](start = 29, length = 11)

In future, a bool to control whether discrete terms are quoted with apostrophes could be useful. Just something to keep in mind.

michaelcfanning:
Nice work, thanks to both of you. :v:

PR #1841, opened 2020-04-01
WebScout fixes

ScottLouvau:
- Result.GetRule now looks up by RuleId (after RuleIndex or Guid)
- ToolComponent builds lookup cache by RuleId and Guid to prevent performance problems looking up every Result's Rule.
- Baselining now persists Run.Tool.Driver.Rules referenced by Results via RuleId properly.
- SarifToCsv: Match Sarif SDK Newtonsoft version.
- SarifToCsv: Make use of DeferredOM configurable, as workaround if needed.
- Fix Deferred OM to detect and handle files with BOMs.
- Fix broken unit tests (unrelated to these changes)

### DIFF
--- a/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
+++ b/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
@@ -246,7 +246,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
                 properties = currentRuns.Last().Properties;
             }
 
-            var reportingDescriptors = new Dictionary<ReportingDescriptor, int>(ReportingDescriptor.ValueComparer);
+            var reportingDescriptors = new Dictionary<string, int>();
 
             var indexRemappingVisitor = new RemapIndicesVisitor(currentArtifacts: null, currentLogicalLocations: null);
 
@@ -268,23 +268,23 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
                 indexRemappingVisitor.HistoricalLogicalLocations = resultPair.Run.LogicalLocations;
                 indexRemappingVisitor.VisitResult(result);
 
-                if (result.RuleIndex != -1)
+                string ruleId = result.ResolvedRuleId(resultPair.Run);
+                if (!string.IsNullOrEmpty(ruleId))
                 {
-                    ReportingDescriptor rule = result.GetRule(resultPair.Run);
-                    if (rule != null)
+                    if (reportingDescriptors.TryGetValue(ruleId, out int ruleIndex))
                     {
-                        if (reportingDescriptors.TryGetValue(rule, out int ruleIndex))
-                        {
-                            result.RuleIndex = ruleIndex;
-                        }
-                        else
-                        {
-                            result.RuleIndex = reportingDescriptors.Count;
-                            reportingDescriptors[rule] = reportingDescriptors.Count;
+                        result.RuleIndex = ruleIndex;
+                        result.RuleId = ruleId;
+                    }
+                    else
+                    {
+                        ReportingDescriptor rule = result.GetRule(resultPair.Run);
 
-                            run.Tool.Driver.Rules = run.Tool.Driver.Rules ?? new List<ReportingDescriptor>();
-                            run.Tool.Driver.Rules.Add(rule);
-                        }
+                        result.RuleIndex = reportingDescriptors.Count;
+                        reportingDescriptors[ruleId] = reportingDescriptors.Count;
+
+                        run.Tool.Driver.Rules = run.Tool.Driver.Rules ?? new List<ReportingDescriptor>();
+                        run.Tool.Driver.Rules.Add(rule);
                     }
                 }
 
--- a/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
+++ b/src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
@@ -250,7 +250,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
 
             var indexRemappingVisitor = new RemapIndicesVisitor(currentArtifacts: null, currentLogicalLocations: null);
 
-            properties = properties ?? new Dictionary<string, SerializedPropertyInfo>();
+            properties ??= new Dictionary<string, SerializedPropertyInfo>();
 
             List<Result> newRunResults = new List<Result>();
             foreach (MatchedResults resultPair in results)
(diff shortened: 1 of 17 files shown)


src/Sarif/Core/Result.cs
```diff
@@ -73,18 +73,21 @@ public ReportingDescriptor GetRule(Run run = null)
                     return GetRuleByIndex(rules, this.Rule.Index);
                 }
 
-                // Look up by this.RuleDescriptor, Guid, if present
-                if (!string.IsNullOrEmpty(this.Rule?.Guid))
+                // Look up by this.RuleDescriptor.Guid, if present
```
lgolding, line 76: >RuleDescriptor [](start = 35, length = 14)

".RuleDescriptor" => ".Rule" (but per Michael, and I agree, we don't need to look up by GUID)

src/Sarif/Core/Result.cs
```diff
@@ -73,18 +73,21 @@ public ReportingDescriptor GetRule(Run run = null)
...
-                        foreach (ReportingDescriptor rule in rules)
-                        {
-                            if (rule.Guid == this.Rule.Guid) { return rule; }
-                        }
-
-                        throw new ArgumentException($"ReportingDescriptorReference referred to Guid {this.Rule.Guid}, which was not found in toolComponent.Rules.");
-                    }
+                    ReportingDescriptor rule = component.GetRuleByGuid(this.Rule.Guid);
+                    if (rule != null) { return rule; }
+                    throw new ArgumentException($"ReportingDescriptorReference referred to Guid {this.Rule.Guid}, which was not found in toolComponent.Rules.");
```
lgolding, line 81: >throw new ArgumentException($"ReportingDescriptorReference referred to Guid {this.Rule.Guid}, which was not found in toolComponent.Rules.") [](start = 20, length = 139)

I don't think we should throw because the SARIF is invalid. Just don't return a `ReportingDescriptor` if you can't find one (or, in this case, fall through to the next block). Comment applies to the ruleId lookup as well.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -268,23 +268,23 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
...
-                        }
-                        else
-                        {
-                            result.RuleIndex = reportingDescriptors.Count;
-                            reportingDescriptors[rule] = reportingDescriptors.Count;
+                        result.RuleIndex = ruleIndex;
+                        result.RuleId = ruleId;
+                    }
+                    else
+                    {
```
lgolding, line 280: >{ [](start = 20, length = 1)

It looks like you don't set `result.RuleId` in the `else` branch. If so, a test hole as well.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -268,23 +268,23 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
...
+                        ReportingDescriptor rule = result.GetRule(resultPair.Run);
 
-                            run.Tool.Driver.Rules = run.Tool.Driver.Rules ?? new List<ReportingDescriptor>();
-                            run.Tool.Driver.Rules.Add(rule);
-                        }
+                        result.RuleIndex = reportingDescriptors.Count;
+                        reportingDescriptors[ruleId] = reportingDescriptors.Count;
+
+                        run.Tool.Driver.Rules = run.Tool.Driver.Rules ?? new List<ReportingDescriptor>();
```
lgolding, line 286: >= [](start = 46, length = 1)

Are we building SDK with C# 8 now? If so can change this to, `run.Tool.Driver.Rules ??= new List<ReportingDescriptor>();`

src/Sarif/Core/ToolComponent.cs
```diff
@@ -1,13 +1,60 @@
...
+                if (r.Guid != null) { _cachedRulesByGuid[r.Guid] = r; }
+            }
+        }
+
+        public ReportingDescriptor GetRuleById(string ruleId)
+        {
+            ReportingDescriptor rule = null;
+
+            // Build lookup if not built or possibly out-of-date
+            if (_cachedRulesById == null || !_cachedRulesById.TryGetValue(ruleId, out rule))
```
lgolding, line 35: >) [](start = 91, length = 1)

Nits:

1. you don't have to predeclare `rule`: `TryGetValue(ruleId, out ReportingDescriptor rule)`
2. you could write `if (_cachedRulesById?.TryGetValue(ruleId, out ReportingDescriptorRule) != true)`

src/Sarif/Readers/LineMappingStreamReader.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Readers
     ///  in the part of the file currently read. This allows it to map a line and char back to a byte offset
     ///  to allow consumers to seek back to the position later.
     /// </summary>
-    internal class LineMappingStreamReader : StreamReader
```
lgolding, line 14: >internal [](start = 4, length = 8)

Why this change?

src/Sarif/Readers/LineMappingStreamReader.cs
```diff
@@ -41,15 +41,55 @@ internal class LineMappingStreamReader : StreamReader
...
+            _bytesReadPreviously = bomWidth;
+
             // (1, 1) is the 0th byte, so the line number before the read is 1 and there is one character (the newline before the first line) before the first read.
             _firstLineNumber = 1;
             _firstLineCharsBeforeBuffer = 1;
 
             _overflowCorrector = new OverflowCorrector();
         }
 
+        private static Stream FindBomWidth(Stream stream, out int bomWidth)
```
lgolding, line 56: >private [](start = 8, length = 7)

Should have unit tests (using our `testCase` pattern).

src/Sarif/Core/Result.cs
ScottLouvau: Fixed. I'm going to leave it in. Feel free to remove GUID based matching based on how you and Mike resolve the conversation.

src/Sarif/Core/Result.cs
ScottLouvau: Fixed

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
ScottLouvau: Changed.

src/Sarif/Core/ToolComponent.cs
ScottLouvau: I agree with #2 but I don't like inline declaration when the variable will be used in a broader context than the statement. Since this one isn't just for the if true case, I think it's weird to declare it inline, technically in an expression that won't execute (if _cachedRulesById was null) and then use it outside the whole block. I know the compiler handles it fine, but it looks weird to me.

ScottLouvau: Though actually I don't love #2 either. You have to think carefully about how the null boolean evaluates to the condition, whereas a X != null && .. condition is very trivial to understand.

src/Sarif/Readers/LineMappingStreamReader.cs
ScottLouvau: Left from running testing in another project because I couldn't get the SDK unit tests to execute. Undone.

src/Sarif/Readers/LineMappingStreamReader.cs
ScottLouvau: Added. I wasn't able to get tests to run, so I wasn't excited about adding them at that point.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
ScottLouvau: Yep - good catch, and added a test to cover Rule lookup by index and id and setting index and id on the result logs.

PR #1845, opened 2020-04-02
Pipeline modifications and improved telemetry.

michaelcfanning:
@rtaket


src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -32,7 +32,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.0;netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
```
rtaket, line 35: >netcoreapp3.1 [](start = 22, length = 13)

This will change the output directory, so just a reminder that the Transformers project will also require this change too.

michaelcfanning: I will revert this as the check-in doesn't contain other required clean-up as well.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -238,21 +253,28 @@ private static void UpdateLogWithWorkItemDetails(SarifLog sarifLog, Uri htmlUri,
...
+                    { "workItemGuid", workItemGuid },
                     { "area", sarifWorkItemModel.Area },
                     { "bodyOrDescription", sarifWorkItemModel.BodyOrDescription },
+                    { "filingResult", filingResult.ToString() },
                     { "commentOrDiscussion", sarifWorkItemModel.CommentOrDiscussion },
                     { "htmlUri", sarifWorkItemModel.HtmlUri },
                     { "iteration", sarifWorkItemModel.Iteration },
                     { "labelsOrTags", tags },
-                    { "locationUri", sarifWorkItemModel.LocationUri },
+                    { "locationUri", sarifWorkItemModel.LocationUris },
```
rtaket, line 277: >LocationUris [](start = 56, length = 12)

I'm not sure this will serialize as you expect. You might have to join or json serialize them.

michaelcfanning: absolutely right. thanks.

michaelcfanning: we should think about how to raise an exception in the telemetry code for this common mistake.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -238,21 +255,29 @@ private static void UpdateLogWithWorkItemDetails(SarifLog sarifLog, Uri htmlUri,
             }
         }
 
-        private void LogMetricsForFiledWorkItem(SarifLog sarifLog, SarifWorkItemModel sarifWorkItemModel)
+        private void LogMetricsForFiledWorkItem(SarifLog sarifLog, SarifWorkItemModel sarifWorkItemModel, FilingResult filingResult)
```
rtaket, line 258: >LogMetricsForFiledWorkItem [](start = 21, length = 26)

Rename to LogMetricsForWorkItem?

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -238,21 +255,29 @@ private static void UpdateLogWithWorkItemDetails(SarifLog sarifLog, Uri htmlUri,
...
         {
-            string guid = Guid.NewGuid().ToString();
+            this.FilingResult = filingResult;
+
+            string workItemGuid = Guid.NewGuid().ToString();
```
rtaket, line 262: >workItemGuid [](start = 19, length = 12)

If the logGuid is generated in FileWorkItemsHelper() and one work item is filed per split sarif log, what's the difference between workItemGuid and logGuid? It seems like there will always be a 1:1 correlation between them.

src/Sarif.Sdk.runsettings
```diff
@@ -0,0 +1,47 @@
+<RunSettings>
```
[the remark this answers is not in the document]
michaelcfanning: Yes. What the heck. I went through and removed things we are very unlikely to use.

PR #1850, opened 2020-04-08
Introduce SarifConstants.SarifFileExtension

lgolding:
I can't believe we never did this. I finally did it after defining the same constant in the n + 1st tool's source code.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -4,6 +4,7 @@
 * BUGFIX: Result.GetRule will look up by RuleId if RuleIndex not present.
 * BUGFIX: Baselining will properly persist Run.Tool.Driver.Rules if Results reference by RuleId.
 * BUGFIX: DeferredOM will properly load files with a BOM. (LineMappingStreamReader fix)
+* FEATURE: Introduce `SarifConstants.SarifFileExtension` with value `".sarif"`.
 
 ## **v2.2.2** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.2) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.2) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.2) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.2)
 * BUGFIX: `dotnet tool install` command for Multitool now produces a working installation rather than reporting missing `Sarif.Converters` binary.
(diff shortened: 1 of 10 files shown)


src/Test.FunctionalTests.Sarif/v2/ConverterTestData/FortifyFpr/ScanWithFailureLevelMatrices.fvdl
```diff
@@ -43576,8 +43576,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         private string RunConverter(StringBuilder sb, string toolFormat, string inputFileName)
         {
-            string expectedFileName = inputFileName + ".sarif";
-            string generatedFileName = inputFileName + ".actual.sarif";
+            string expectedFileName = inputFileName + SarifConstants.SarifFileExtension;
```
harleenkohli, line 43579: >SarifConstants.SarifFileExtension; [](start = 54, length = 34)

i am probably splitting hair, but this irks me! The fvdl files added here extracted from the corresponding fpr file (With the same name). These fvdl files are here for reference only (as far as i recall). I think the functional test uses the fpr file as an input. So, let us not change this file so that it is an exact match with what is extracted out of the FPR.

Alternately, we should re-generate the FPR file as well.

harleenkohli: aah, my memory did not betray me (this time!), see this comment:

[https://github.com/microsoft/sarif-sdk/blob/ffff4b3c9265a441fb00043ebb8e079f6690f870/src/Sarif.Converters/FortifyFprConverter.cs#L127](https://github.com/microsoft/sarif-sdk/blob/ffff4b3c9265a441fb00043ebb8e079f6690f870/src/Sarif.Converters/FortifyFprConverter.cs#L127)

---
In reply to: [406390399](https://github.com/microsoft/sarif-sdk/pull/1850#discussion_r406390399) [](ancestors = 406390399)

lgolding: Oh, good catch! I saw that this code was grayed out and I thought it was an #if block. I will revert this change.

---
In reply to: [406391355](https://github.com/microsoft/sarif-sdk/pull/1850#discussion_r406391355) [](ancestors = 406391355,406390399)

PR #1851, opened 2020-04-14
Remove Sarif.Converters CsvHelper dependency.

ScottLouvau:
- Add CsvReader/TsvReader (ported Elfie variants using .NET string)
  - Port tests (100% code coverage)
  - Convert SemmleQaConverter to CsvReader, performance improvements.
  - Add SemmleQlConverter tests (in Multitool unit tests)


src/Sarif.Converters/TextFormats/CsvReader.cs
```diff
@@ -0,0 +1,148 @@
...
+
+                // If the last delimiter was not a cell delimiter, we finished reading the row
+                if (!(cellEnd < _bufferFilledLength && _buffer[cellEnd] == ','))
+                {
+                    break;
+                }
+            }
+
+            // Skip newline for \r\n to set _nextIndex to the beginning of the next line
+            if (_nextIndex <= _bufferFilledLength && _buffer[_nextIndex - 1] == '\r') { _nextIndex++; }
```
rtaket, line 83: >'\r' [](start = 80, length = 4)

Shouldn't we be looking for "\r\n" or "\n" here, in case the file was created with unix line endings?

ScottLouvau: It handles both styles. The logic earlier to look for separators looks for ',', '\r', and '\n'. This line skips over the \n also if we got \r as the separator, since the next value won't start until after both parts of Windows style newlines.

Unit tests cover both newline styles (even mixed in one file)

PR #1855, opened 2020-04-15
Fix #1854: Add validation rule: originalUriBaseId URIs must end with a slash.

### LINKED ISSUE #1854: New rule: originalUriBaseIds must end with a slash
See microsoft/sarif-visualstudio-extension#127. The spec requires the trailing slash. Some consumers might compensate for its absence, but the producer can't rely on that.

@michaelcfanning

lgolding:


src/build.props
```diff
@@ -20,12 +20,8 @@
     having all the relevant values in one place. This property is actually used by the PowerShell
     script that hides ("delists") the previous package versions on nuget.org.
     -->
-    <!-- 
-      NOTE: Version in scripts/BuildMultitoolForNpm.ps1 also has to be updated for each release
-      until it adopts the same versions as the SDK itself.
```
lgolding, line 25: >itself [](start = 51, length = 6)

The npm script now gets the version number from the SDK.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Valid.sarif
```diff
@@ -28,7 +28,7 @@
         }
       },
       "originalUriBaseIds": {
-        "SRCROOT": "file:///c:/Code/sarif-sdk/src"
+        "SRCROOT": "file:///c:/Code/sarif-sdk/src/"
```
lgolding, line 31: >/ [](start = 49, length = 1)

This file is supposed to be valid against SARIF1003 ("URIs must be valid"). Don't let it be invalid because of the new SARIF1019.

lgolding:

lgolding:

src/Sarif.Multitool/Rules/SARIF1019.OriginalUriBaseIdMustEndWithSlash.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Json.Pointer;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class OriginalUriBaseIdMustEndWithSlash : SarifValidationSkimmerBase
```
michaelcfanning, line 10: >OriginalUriBaseIdMustEndWithSlash [](start = 17, length = 33)

A better rule would be 'DirectoryUrisMustEndWIthSlash' and to add all examples.  Consider `invocation.workingDirectory`, another time-bomb waiting to go off. Go look at all artifacts of role == directory. I can't think of anything else off the top of my head but you might want to scan for more. In any case, a bucket rule with these three examples would provide a place for more if we discover we overlooked them.

src/Sarif.Multitool/Rules/SARIF1019.OriginalUriBaseIdMustEndWithSlash.cs
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }
+
+        private void AnalyzeOriginalUriBaseIdsEntry(string originalUriBaseId, ArtifactLocation artifactLocation, string pointer)
+        {
+            string uriString = artifactLocation?.Uri?.OriginalString;
+
+            // If it's not a well-formed URI, don't bother triggering this rule. Rule SARIF1003,
+            // UrisMustBeValid, will catch it.
```
michaelcfanning, line 49: Yes, absolutely true

src/Sarif.Multitool/Rules/SARIF1019.OriginalUriBaseIdMustEndWithSlash.cs
lgolding: I think that's a good idea.

`originalUriBaseIds` is particularly important because it's the only URI in SARIF (that I know of) whose _purpose_ in life is to be combined with another URI. But sure, I'll scan for other examples in the spec.

---
In reply to: [409168496](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r409168496) [](ancestors = 409168496)

michaelcfanning: Are we missing the matching analysis, which is to ensure that any relative path which is matched to a uriBaseId does not have a leading slash?

Or is it legal for those URIs? I can't recall the convention now for pointing at, say, the root within a ZIP archive. Would have to go refresh my memory.

---
In reply to: [409169955](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r409169955) [](ancestors = 409169955,409168496)

lgolding: Good catch. There are really two cases here.

- The normal case of a "top-level" (non-nested) file. You are right, **the leading slash should not be allowed here**, because it can produce a result that is _not underneath_ the base URI. Here's what the C# immediate window says:

```
> var uri = new Uri("[file:///C:/code/src/](file:///C:/code/src/)");
> var relative = new Uri("/io/kb.c", UriKind.Relative);
> var combined = new Uri(uri, relative);
> combined
[[file:///C:/io/kb.c](file:///C:/io/kb.c)]
> 
```

- The case of a nested file. I think you are worried about the case where the uriBaseId _is_ a zip file, and then you have to worry about how to represent the files contained within the zip file. But that's not allowed:

```
"originalUriBaseIds": {
  "ROOT": {
    "uri": "[file:///C:/data/myFiles.zip/](file:///C:/data/myFiles.zip/)"
  }
}
```

It's not allowed because nested files can't refer to their parent by means of a `urisBaseId`. §3.4.4 says:

> If this `artifactLocation` object describes a nested artifact, `uriBaseId` **SHALL** be absent.

Nested artifacts refer to their parents by means of their `parentIndex` property.


---
In reply to: [409171723](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r409171723) [](ancestors = 409171723,409169955,409168496)

lgolding: I looked at the spec and the only things that are directories are -- exactly as you said -- the `invocation.workingDirectory` property, and any `artifact` whose `roles` property includes `"directory"`. I will add those to this PR.

I'd like to handle the opposite case (no leading slash on relative paths which have a `uriBaseId`) in a separate PR, possibly a separate rule.

---
In reply to: [409199116](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r409199116) [](ancestors = 409199116,409171723,409169955,409168496)

lgolding: Also, the `originalUriBaseIds` part of this rule should be an error (it directly violates the spec), while IMO the other two are warnings (the spec doesn't actually require them to end in a slash). And all three need different messages.

---
In reply to: [410303424](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r410303424) [](ancestors = 410303424,409199116,409171723,409169955,409168496)

michaelcfanning: Disagree on two rules. Let's talk offline.

---
In reply to: [410304168](https://github.com/microsoft/sarif-sdk/pull/1855#discussion_r410304168) [](ancestors = 410304168,410303424,409199116,409171723,409169955,409168496)

lgolding:
Per discussion with @michaelcfanning, we will implement this "trailing slash" check in the existing rule SARIF1008, `InvalidUriInOriginalUriBaseIds`, with which it coheres very nicely.

There remains the question of which rule or rules should report these additional URI-related concerns:

- `invocation.workingDirectory` should end with a slash.
- The `uri` property of any `artifactLocation` whose roles include `"directory"` should end in a slash.
- The `uri` property of any `artifactLocation` with a non-null `uriBaseId` property should _not begin_ with a slash.

We have opened #1857 to track this work.

PR #1858, opened 2020-04-17
In SARIF1018, also require a trailing slash.

### LINKED ISSUE #1860: SARIF1008: Don't require Markdown messages to end with a period
For example, they might end with a bullet list item on which a period doesn't make sense.

lgolding:
Also:
- Fix #1860: Don't require Markdown messages to end with a period.

### DIFF
--- a/src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1018.InvalidUriInOriginalUriBaseIds_Invalid.sarif
+++ b/src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1018.InvalidUriInOriginalUriBaseIds_Invalid.sarif
@@ -14,11 +14,14 @@
                 "text": "In the artifactLocation objects contained in run."
               },
               "fullDescription": {
-                "text": "In the artifactLocation objects contained in run.originalUriBaseIds, if uriBaseId is absent, then uri must either be an absolute URI or it must be absent."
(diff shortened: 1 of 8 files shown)


src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1018.InvalidUriInOriginalUriBaseIds_Invalid.sarif
```diff
@@ -14,11 +14,14 @@
                 "text": "In the artifactLocation objects contained in run."
               },
               "fullDescription": {
-                "text": "In the artifactLocation objects contained in run.originalUriBaseIds, if uriBaseId is absent, then uri must either be an absolute URI or it must be absent."
+                "text": "In the artifactLocation objects contained in run.originalUriBaseIds, if uriBaseId is absent, then uri must either be an absolute URI or it must be absent. Also, uri must end with a slash, so that it can safely be combined with the relative URIs in artifactLocation objects elsewhere in the log file."
               },
               "messageStrings": {
-                "Default": {
-                  "text": "In this artifactLocation object contained in run.originalUriBaseIds, uriBaseId is absent, but uri is '{0}', which is a relative URI."
+                "NotAbsolute": {
```
michaelcfanning, line 20: >NotAbsolute [](start = 17, length = 11)

Remember our standard that the locating/uniquely qualifying dynamic data should be as far left in the text as possible. The reason is that this maximizes the readability of the message in something like the error list, where the column width could be constrained.

I can see that this is tricky in this case, but perhaps you could take a crack at it.

lgolding:
@michaelcfanning While I was in here I took a few minutes to fix #1860: Don't require Markdown messages to end with a period.

lgolding:
@michaelcfanning FYI: As @eddynaka points out, this fixes #1854, "New rule: originalUriBaseIds must end with a slash". We didn't create a new rule, we enhanced an existing rule. I closed #1854.

PR #1859, opened 2020-04-17
Test Multi-Repository Pipeline

nickmarston:


michaelcfanning:
Maybe put this in the \scripts\ directory? or some other root? thoughts?

michaelcfanning:
Approving since we can make this adjustment easily later.

---
In reply to: [616205466](https://github.com/microsoft/sarif-sdk/pull/1859#issuecomment-616205466) [](ancestors = 616205466)

PR #1861, opened 2020-04-24
Add test knob to debug builds to limit # of partitioned files that ar…

michaelcfanning:
…e sent for filing.

@rtaket @nickmarston @alisonmaclellan


michaelcfanning:
@lgolding

PR #1863, opened 2020-04-24
Fix #1862: "Trailing slash check" is not done when uriBaseId is present.

### LINKED ISSUE #1862: "Trailing slash check" is not done when uriBaseId is present
Rule SARIF1018 validates two things about the `uri` properties under `originalUriBaseIds`:

1. If `uriBaseId` is absent, `uri` must be absolute.
2. `uri` must end with a slash.

<span>#</span>2 was just added recently. But because of the `uriBaseId` condition in the existing logic for <span>#</span>1, I was exiting early if `uriBaseId` was present. That meant I wasn't checking for a trailing slash when `uriBaseId` was present.

@michaelcfanning FYI

lgolding:


lgolding:

PR #1866, opened 2020-04-28
NPM Non-Trimmed, DeferredList fix

ScottLouvau:
An NPM build configuration fix to try to unblock Gabriel for a demo, and a small fix to DeferredList to make it work with SarifRewritingVisitor. (Feature request: Generate a parallel visitor for non-mutating traversal of the log. VisitX would return void and the generated calling methods would not try to set instances to any new values).

### DIFF
--- a/src/Sarif.Multitool/Sarif.Multitool.csproj
+++ b/src/Sarif.Multitool/Sarif.Multitool.csproj
@@ -11,10 +11,11 @@
       We use [Condition="'$(RuntimeIdentifier)' != ''"] to identify settings used only for these builds.      
     -->
 
-    <!-- Publish trimmed single-file exe -->
+    <!-- Publish single-file exe -->
+    <PublishSingleFile Condition="'$(RuntimeIdentifier)' != ''">true</PublishSingleFile>
+
     <!-- Seeing assembly load failures with trimmed exe, but not without. On some machines, for some commands. :/ -->
     <!-- <PublishTrimmed Condition="'$(RuntimeIdentifier)' != ''">true</PublishTrimmed> -->
-    <PublishSingleFile Condition="'$(RuntimeIdentifier)' != ''">true</PublishSingleFile>
 
     <!-- Publish 'ready-to-run' on Windows (Linux/Mac not yet supported) -->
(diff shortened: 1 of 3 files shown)


src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -12,7 +12,8 @@
     -->
 
     <!-- Publish trimmed single-file exe -->
```
lgolding, line 14: >trimmed [](start = 17, length = 7)

So is this comment now wrong?

src/Sarif/Readers/DeferredList.cs
```diff
@@ -40,12 +40,16 @@ public class DeferredList<T> : IList<T>, IDisposable
 
         private Func<T, T> _transformer;
 
+        private T _lastItem;
```
lgolding, line 43: >_last [](start = 18, length = 5)

My only comment is about naming: Whenever I see "last" I immediately wonder "does this mean 'last in order' or 'most recent'?". In this case, it turned out to mean "most recent", but I had to carry that question in my mind from the time I read this line until I understood the answer from the setter below.

Consider `_mostRecentItem` or `_lastAccessedItem`.

ScottLouvau: Reasonable. Fixed.

src/Sarif.Multitool/Sarif.Multitool.csproj
ScottLouvau: I've fixed. Were you really confused? If so, fine. 

I feel like the first comment explains the original intent, and the second comment (and commented setting) shows the change and documents why we're not using that setting. I left the setting in and commented to explain that we did consider PublishTrimmed but found problems with it. If I'd removed the PublishTrimmed line, we would lose the record that we tried it and had problems.

Overall, I feel like this combination of comments captures the initial intend and current state, and it's easy to understand which came first and which is current.

If you did understand the order, I feel like this feedback was overdoing it a little on comment pickiness. I don't want comments to be confusing or conflicting, but if people can understand the order of what happened and the comments collectively tell the right story, I feel like that's sufficient.

PR #1871, opened 2020-05-12
UriConverter fix to avoid doubling Uris with certain escaped characters.

ScottLouvau:
- Now serializing OriginalString as-is to ensure no doubling can occur.
  - This means we won't try to sanitize or canonicalize user-provided Uris anymore.
  - Consolidated test data into one of the two test classes.
  - Changed verification to check that the Uri reloads equal to the user-created original, not that the written SARIF has a canonicalized value.
  - Update an invalid Uri test expected file which expected somewhat-sanitized Uris.

  See https://github.com/dotnet/runtime/issues/36288 for a .NET Framework bug I've reported on this issue.


src/Test.UnitTests.Sarif/UriConverterTests.cs
```diff
@@ -76,5 +83,72 @@ public void WriteJson_ConvertsUriObjectsToStrings()
 
             JsonConvert.SerializeObject(testObject, settings).Should().Be(s_testFileText);
         }
+
+        [Theory]
```
lgolding, line 87: >Theory [](start = 9, length = 6)

Rather than `Theory`, which in the past (at least) has been slow, we have an established pattern of looping over test cases and accumulating the failures to report at the end. For an example of the pattern, see src\Test.FunctionalTests.Sarif\SarifValidatorTests.cs.

If you have more recent experience of `Theory`, please let us know.

src/Test.UnitTests.Sarif/UriConverterTests.cs
```diff
@@ -76,5 +83,72 @@ public void WriteJson_ConvertsUriObjectsToStrings()
...
+            // Put in a class with a Uri using the Sarif 'UriConverter'
+            SingleUri sample = new SingleUri();
+            sample.Uri = value;
+
+            // Serialize and Deserialize
+            string json = JsonConvert.SerializeObject(sample);
+            SingleUri roundTripped = JsonConvert.DeserializeObject<SingleUri>(json);
+
+            roundTripped.Uri.Should().Be(value);
+            return roundTripped.Uri;
```
lgolding, line 140: >return roundTripped.Uri; [](start = 12, length = 24)

Return value not used, needed?

ScottLouvau: It's not critical. I follow this pattern so that if the test wants to assert other details of the output, it can easily do so.

src/Test.UnitTests.Sarif/UriConverterTests.cs
ScottLouvau: I didn't use Theory initially, but really liked it for telling me which cases were broken (instead of stopping the test on the first failed case).

Runtime is 47ms (Debug) versus 17ms as subsequent calls in the method.

I've changed it to accumulate non-matches. I wish the SDK unit tests didn't take 60 seconds to run, so might as well do my part to help...

PR #1872, opened 2020-05-14
Splitting strategy per partial fingerprint.

cfaucon:


src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -41,6 +41,11 @@ public static bool ShouldBeFiled(this Result result)
             // Informational: a strictly informational message was emitted.
         }
 
+        public static string GetFingerprintSplittingStrategyId(this Result result)
```
rtaket, line 44: >GetFingerprintSplittingStrategyId [](start = 29, length = 33)

Perhaps a more specific name. Mostly because we could have a splitting strategy only based on the fingerprint.

PR #1873, opened 2020-05-14
Adding switch to allow for filing work items for results including ba…

cfaucon:
…selinestate unchanged.

### DIFF
--- a/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
+++ b/src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
@@ -10,12 +10,16 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItems
 {
     public static class SarifWorkItemsExtensions
     {
-        public static bool ShouldBeFiled(this Result result)
+        public static bool ShouldBeFiled(this Result result, bool shouldFileUnchanged = false)
         {
             if (result.BaselineState != BaselineState.None &&
-                result.BaselineState != BaselineState.New) 
-            { 
-                return false; 
+                result.BaselineState != BaselineState.New)
+            {
+                if (result.BaselineState != BaselineState.Unchanged ||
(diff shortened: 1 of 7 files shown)


src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -46,7 +50,7 @@ public static string GetFingerprintSplittingStrategyId(this Result result)
             return $"{result.GetProperty("OrganizationName")}:{result.GetProperty("EtlEntity")}:{result.PartialFingerprints["SecretHash/v1"]}";
         }
 
-        public static string CreateWorkItemTitle(this Run run)
+        public static string CreateWorkItemTitle(this Run run, bool shouldFileUnchanged = false)
```
rtaket, line 53: >false [](start = 90, length = 5)

Why default the value here?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -121,17 +125,17 @@ public static List<string> GetToolNames(this SarifLog log)
                 .ToList();
         }
 
-        public static int GetAggregateFilableResultsCount(this SarifLog log)
+        public static int GetAggregateFilableResultsCount(this SarifLog log, bool shouldFileUnchanged = false)
```
rtaket, line 128: >false [](start = 104, length = 5)

why default?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
cfaucon: removed default

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
cfaucon: removed default

Issue #1740, opened 2020-01-29
Serialization omits required properties with null values


lgolding:

eddynaka:
@lgolding , I would like to know if this is still a problem.

lgolding:
@eddynaka, thank you for your contributions! I just ran the program above on the latest SARIF SDK (2.2.5), and yes, it is still a problem.

eddynaka:
Hi @lgolding , i have some questions:
- in your example, when we do the first `DeserializeObject`, the `FullDescription` should be available with the `Text`? Right now, it's a null object.
- in your example, when we do the first `SerializeObject`, the `FullDescription` should be available with the `Text`? Right now, it's just showing as empty.

lgolding:
Here is the "minimal repro file" from the issue description at the top:
```
{
  "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.4",
  "version": "2.1.0",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "Bug1740",
          "rules": [
            {
              "id": "TST0001",
              "name": "Test rule",
              "fullDescription": {
                "text": null
              }
            }
          ]
        }
      },
      "results": [],
      "columnKind": "utf16CodeUnits"
    }
  ]
}
```
You see that `fullDescription.text` is present, but it is `null`. So when you do the first deserialize operation, the `FullDescription.Text` property is `null`, as you see.

Now, when you re-serialize, it becomes this:
```
{
  "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.4",
  "version": "2.1.0",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "Bug1740",
          "rules": [
            {
              "id": "TST0001",
              "name": "Test rule",
              "fullDescription": {
              }
            }
          ]
        }
      },
      "results": [],
      "columnKind": "utf16CodeUnits"
    }
  ]
}
```

You see that now, `fullDescription.text` is completely absent, even though the schema says it is a required property. That is the bug.

Finally, when you do the second deserialization operation, it fails, because the required property `fullDescription.text` is absent from the JSON.

Does that help?

eddynaka:
@lgolding , using your sample, I could achieve what you are saying doing this:
For the following code, I had to remove/always return true;
[...]

PR #1883, opened 2020-05-22
updating shouldSerializeText return

eddynaka:
Refs #1740

- updating ShouldSerializeText so we could be able to Deserialize and Serialize sarifLog without problem if FullDescription.text is null


eddynaka:
closing since i have opened a internal one. #1885

PR #1884, opened 2020-05-29
Update release history and version number.

lgolding:
Update release history for change made by @eddynaka, and bump the version number.

Also:
- Remove an obsolete comment in build props. The script BuildMultitoolForNpm.ps1 now [automatically tracks](https://github.com/microsoft/sarif-sdk/blob/75592fc8346809f36e093f79fc66bb84f4bb28cd/scripts/BuildMultitoolForNpm.ps1#L52) the SDK version number.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,5 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v2.2.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.6)
+* BUGFIX: `ResultLogJsonWriter` now creates an empty `results` array if there are no results, rather than leaving `results` as `null`. [#1821](https://github.com/microsoft/sarif-sdk/issues/1821)
+
 ## **v2.2.5** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.5) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.5) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.5) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.5)
 * BUGFIX: Fix SDK doubling Uris with certain escaped characters (ex: '-' and '_') on every Load/Save cycle (cause: https://github.com/dotnet/runtime/issues/36288)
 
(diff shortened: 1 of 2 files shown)


src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v2.2.6** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.6) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.6) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.6) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.6)
+* BUGFIX: `ResultLogJsonWriter` now creates an empty `results` array if there are no results, rather than leaving `results` as `null`. [#1821](https://github.com/microsoft/sarif-sdk/issues/1821)
+* BUGFIX: In validation rules, `shortDescription` is now calculated by `GetFirstSentence` method, fixing a bug in sentence breaking. [#1887](https://github.com/microsoft/sarif-sdk/issues/1887)
+* FEATURE: Add validation rule `SARIF1019`, which requires every result to have at least one of `result.ruleId` and `result.rule.id`. If both are present, they must be equal. [#1880](https://github.com/microsoft/sarif-sdk/issues/1880)
```
michaelcfanning, line 6: >FEATURE [](start = 2, length = 7)

This is a good rule. btw - I think the next important rule for us to author is to write a detection for rules that return static strings. This would be a heuristic driven analysis. i.e., we'd need to see more than a single message in order to determine non-uniqueness. We'd also want to ensure that there was some differentiation in logical/physical locations. There's not really a precise way to handle this that I can think of. But roughly, there are >1 messages that span multiple files/logical locations and all messages are identical, this could be a good start. We should consider filing bugs against the many code ql checks that are in this condition.

lgolding: Another heuristic is to look for results whose messages don't have any enquoted strings (does not match `'.+'` -- or to look for rules whose message strings don't have any substitution sequences `{0}`. (Of course those heuristics don't tell you that it's a location being substituted, while yours at leasts suggests that that's a possibility.)

michaelcfanning: Another one of our annoying standards is to enquote dynamic information. Might as well storm right in there. It would be helpful for supporting better discoverability for this analysis. If people properly enquoted dynamic details, the absence of quotes could be a clue re: a static rule result even in cases where we have a single instance to examine. 

Certainly the existence of message args is a good sign of dynamism, so we'd assume the standard is enforced in that case. The use of message arguments is another (lower priority) standard we should be enforcing. This standard helps improve result matching results. That fact s/be noted in the rule output.

So, seems like at least three ideas here. But we probably should be enforcing a single rule: AuthorMessageTextCorrectly or some such, with judicious use of sub-rules. I'm sure we'll have an interesting chat about it.


---
In reply to: [432984183](https://github.com/microsoft/sarif-sdk/pull/1884#discussion_r432984183) [](ancestors = 432984183)

PR #1885, opened 2020-05-29
updating shouldSerializeText return

eddynaka:
Refs #1740

- updating ShouldSerializeText so we could be able to Deserialize and Serialize sarifLog without problem if FullDescription.text is null

@lgolding , i thought in create a new test getting a default sariflog that we have in the repository, setting the FullDescription.Text to null -> serialize -> desserialize -> serialize again.
The expect would be to work. The question is: where should I create that file?

### DIFF
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -40,6 +40,11 @@ namespace Microsoft.CodeAnalysis.Sarif
             return collection != null && collection.Any((m) => m != null);
         }
 
+        public static bool HasAtLeastOneNonNullEmptyValue(this MultiformatMessageString multiformatMessageString)
+        {
+            return !string.IsNullOrEmpty(multiformatMessageString?.Text);
+        }
+
         public static bool HasAtLeastOneNonDefaultValue<T>(this IEnumerable<T> collection, IEqualityComparer<T> comparer) where T : new()
         {
             var defaultInstance = new T();
--- a/src/Sarif/Core/ReportingDescriptor.cs
+++ b/src/Sarif/Core/ReportingDescriptor.cs
@@ -29,6 +29,11 @@ namespace Microsoft.CodeAnalysis.Sarif
             return this.DeprecatedNames.HasAtLeastOneNonNullValue();
         }
 
+        public bool ShouldSerializeFullDescription()
+        {
+            return this.FullDescription.HasAtLeastOneNonNullEmptyValue();
+        }
+
         public bool ShouldSerializeRelationships()
         {
             return this.Relationships.HasAtLeastOneNonDefaultValue(ReportingDescriptorRelationship.ValueComparer);
--- a/src/Test.FunctionalTests.Sarif/SarifConverterTests.cs
+++ b/src/Test.FunctionalTests.Sarif/SarifConverterTests.cs
@@ -2,6 +2,7 @@
 // license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Text;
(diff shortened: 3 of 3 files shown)


src/Sarif/Core/MultiformatMessageString.cs
```diff
@@ -7,6 +7,6 @@ public partial class MultiformatMessageString
     {
         public bool ShouldSerializeMarkdown() { return !string.IsNullOrWhiteSpace(this.Markdown); }
 
-        public bool ShouldSerializeText() { return !string.IsNullOrWhiteSpace(this.Text); }
+        public bool ShouldSerializeText() { return true; }
```
michaelcfanning, line 10: >true [](start = 51, length = 4)

True is a default for all serialization, and so if this change were correct, we should be able to remove it entirely. I think we added this special logic, though, to handle the case where users populated text with a non-null but empty string.

eddynaka:
The problem in removing, if I'm not mistaken is that that would change the public api. It's why, we preferred to maintain.

src/Sarif/Core/MultiformatMessageString.cs
michaelcfanning: See my comment to Larry on #1740, I'm not convinced we have the right fix here.

---
In reply to: [432989835](https://github.com/microsoft/sarif-sdk/pull/1885#discussion_r432989835) [](ancestors = 432989835)

michaelcfanning:
Well, that makes sense. It does leave a slight performance cost, since JSON.NET will need to dynamically acquire/invoke the (now unnecessary) method. Probably too small a perf issue to worry about. We could consider marking the method as [Obsolete] to prep for future removal. In any case, I'm proposing a different change to you and Larry, let's discuss over on #1740.

---
In reply to: [636538361](https://github.com/microsoft/sarif-sdk/pull/1885#issuecomment-636538361) [](ancestors = 636538361)

src/Sarif/ExtensionMethods.cs
```diff
@@ -40,6 +40,11 @@ public static bool HasAtLeastOneNonNullValue<T>(this IEnumerable<T> collection)
             return collection != null && collection.Any((m) => m != null);
         }
 
+        public static bool HasAtLeastOneNonNullEmptyValue(this MultiformatMessageString multiformatMessageString)
```
michaelcfanning, line 43: >HasAtLeastOneNonNullEmptyValue [](start = 27, length = 30)

This isn't quite right, I think. Your serialization logic would fail to emit a property value with a multiformat message string that had a markdown property but no text value. This isn't valid SARIF but the point here is not to ensure valid SARIF by dropping explicitly provided data, make sense? Instead, it is here to drop verifiably empty objects. MutiformatMessageString is a lightweight object, so you can simply add checks to see whether the Markdown property is available, and/or the properties bag is non-null with a count of items that exceeds zero. I suggest adding tests first that would have caught these removals (i.e., try to serialize a multiformat message string with only a markdown value and author a test that verifies it survives. same for the property bag only scenario).

eddynaka: Changing locally as you suggested:

the section below is the rules inside a sarif to test if we can serialize/deserialize successfully
```csharp
Rules = new List<ReportingDescriptor>
                                {
                                    new ReportingDescriptor
                                    {
                                        FullDescription = new MultiformatMessageString
                                        {
                                            Markdown = "markdown text"
                                        }
                                    }
                                }
```

it generates after the first serialization the following json:
```json
{
                            "fullDescription": {
                                "markdown": "markdown text",
                            }
                        }
```

when i try to deserialize to sariflog, it throws an exception:
`  Newtonsoft.Json.JsonSerializationException : Required property 'text' not found in JSON. Path 'runs[0].tool.driver.rules[0].fullDescription', line 1, position 209.`

if we just add the code that i implemented, the fullDescription wouldn't be serialized, so it would be wrong as well. But, if we serialize, we have to add somehow the text to be serializable again....

What do you think?


---
In reply to: [440881150](https://github.com/microsoft/sarif-sdk/pull/1885#discussion_r440881150) [](ancestors = 440881150)

michaelcfanning: I see. So we have dueling values here. On the one hand, we don't want people to be able to serialize SARIF which subsequently doesn't deserialize. On the other hand, we don't want to unexpectedly remove any data in order to ensure what's persisted to disk is valid. @lgolding do you have any thoughts on this?

---
In reply to: [444911014](https://github.com/microsoft/sarif-sdk/pull/1885#discussion_r444911014) [](ancestors = 444911014,440881150)

lgolding:
Here's my opinion after all the thought we've put into this:

The problem occurs because when Newtonsoft's deserializer sees
```
"fullDescription": {
    "text": null
}
```
... it says to itself "Great! This is valid. The `text` property is there.". But it's not really valid. The required value of the `text` property is `string` -- and `null` isn't a `string` in JSON; they're distinct data types (see the [JSON spec](https://www.json.org/json-en.html)).

So when we serialize, and `text` disappears altogether:
```
"fullDescription": {}
```
... we're really no worse off than before. The SARIF was invalid before, and it's still invalid. The bug is _not_ that we fail to deserialize the empty object; it's that we _succeeded_ in deserializing the object with a `null` value for `text`.

And that bug is in Newtonsoft. It sees the declaration of the `Text` property in the `MultiformatMessageString` class as a .NET `string`, and so it accepts a `null` value because in .NET `null` can stand in for a string. But it shouldn't.

One option is to introduce a "string converter" that intercepts deserialization of all string properties, and if the property is marked `Required` but it's `null`, we throw.

Or we could "won't fix" #1740, and file a bug on Newtonsoft.

Issue #1880, opened 2020-05-21
New rule: Validate presence and consistency of result.ruleId/result.rule.id


lgolding:
At least one of `result.ruleId` or `result.rule.id` must be present, and if both are present, they must be equal.

eddynaka:
Hi @lgolding ,

I have some questions about this rule:
1. **Naming**: for the other rules that we have in `Sarif.Multitool.Rules`, every rule starts with a code (for example: SARIF1001.DoNotUseFriendlyNameAsRuleId). Are we going to maintain that pattern? If yes, what code should I use?
2. **Resources**: looking at the file `RuleResources.resx`, we have at least two rows for each rule: `Default` and `Problem`. Based on the other rules, the `_default` is used as full description and, when we find an issue, we assign a new value, right? For this case, I was thinking in creating 3 rows in the resources file: `_default`, `_ruleIdMustbepresent`, `_ruleIdMustBeEqual`.

Thank you for the help

lgolding:
1. Yes, we will maintain that pattern. Choose the next available number in src\Sarif.Multitool\Rules\RuleId.cs, namely `SARIF1019`.

2. To understand the naming, you need to know that the definition of a SARIF rule includes two kinds of user-facing strings: its _description_ and its _result messages_. A SARIF rule can actually have two descriptions, a "short description" and a "full description", but the validation rules don't bother with the short description. You can recognize the "full description" resource string because it has the same name as the class. For example, in `SARIF1014.UriBaseIdRequiresRelativeUri`, you see that the rule's full description comes from the resource string `SARIF1014_UriBaseIdRequiresRelativeUri`.

    Now for the result messages. Most rules produce the same result message for all results, and for those, the resource name is (for example) `SARIF1014_Default`. But occasionally it's helpful to produce different messages in different circumstances. According to the resources file, we currently do this for `SARIF1009` and `SARIF1018`. You can recognize these variant messages in two ways:

- They aren't named "Default", and they don't have the same name as the class.
- They include formatting replacement sequences like `{0}`.

Now, having said all that: for this new rule `SARIF1019`, in my opinion we should have two messages:

- When neither `result.ruleId` nor `result.rule.id` is present, the message should be `SARIF1019_MissingResultRuleId`: "{0}: The result contains neither result.ruleId nor result.rule.id."
[...]

PR #1886, opened 2020-05-31
Adding rule SARIF1019

eddynaka:
Refs #1880

- created rule sarif1019
- created unit test to sarif1019

### DIFF
--- a/src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
+++ b/src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
@@ -10,7 +10,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
     {
         private readonly MultiformatMessageString _fullDescription = new MultiformatMessageString
         {
-            Text = RuleResources.SARIF1019_MissingResultRuleId
+            Text = RuleResources.SARIF1019_RuleIdMustBePresentAndConsistent
         };
 
         public override MultiformatMessageString FullDescription => _fullDescription;
--- a/src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
+++ b/src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class RuleIdMustBePresentAndConsistent : SarifValidationSkimmerBase
+    {
+        private readonly MultiformatMessageString _fullDescription = new MultiformatMessageString
+        {
+            Text = RuleResources.SARIF1019_MissingResultRuleId
+        };
+
+        public override MultiformatMessageString FullDescription => _fullDescription;
+
+        public override FailureLevel DefaultLevel => FailureLevel.Error;
+
+        /// <summary>
+        /// SARIF1019
+        /// </summary>
+        public override string Id => RuleId.RuleIdMustBePresentAndConsistent;
+
+        protected override IEnumerable<string> MessageResourceNames => new string[]
+        {
+            nameof(RuleResources.SARIF1019_InconsistentResultRuleId),
+            nameof(RuleResources.SARIF1019_MissingResultRuleId)
+        };
+
+        protected override void Analyze(Result result, string resultPointer)
+        {
+            AnalyzeRuleId(result, resultPointer);
+        }
+
+        private void AnalyzeRuleId(Result result, string pointer)
+        {
+            // At least one of result.ruleId or result.rule.id must be present
+            if (string.IsNullOrWhiteSpace(result.RuleId) && string.IsNullOrWhiteSpace(result.Rule?.Id))
+            {
+                LogResult(pointer, nameof(RuleResources.SARIF1019_InconsistentResultRuleId), result.RuleId, result.Rule?.Id);
+            }
+            // if both are present, they must be equal.
+            else if (!string.IsNullOrWhiteSpace(result.RuleId)
+                && !string.IsNullOrWhiteSpace(result.Rule?.Id)
+                && !result.RuleId.Equals(result.Rule?.Id, StringComparison.OrdinalIgnoreCase))
+            {
+                LogResult(pointer, nameof(RuleResources.SARIF1019_MissingResultRuleId), result.RuleId, result.Rule?.Id);
+            }
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class RuleIdMustBePresentAndConsistent : SarifValidationSkimmerBase
+    {
+        private readonly MultiformatMessageString _fullDescription = new MultiformatMessageString
+        {
+            Text = RuleResources.SARIF1019_MissingResultRuleId
```
lgolding, line 13: >SARIF1019_MissingResultRuleId [](start = 33, length = 29)

The "full description" string that describes the _rule_ is distinct from any of the message strings that describe individual _results_ from the rule. There needs to be a separate resource string, whose name matches the rule (`SARIF1019_RuleIdMustBePresentAndConsistent`). Let's give it this text:

In every result, at least one of the properties result.ruleId and result.rule.id must be present. If both are present, they must be equal.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1019.RuleIdMustBePresentAndConsistent_Invalid.sarif
```diff
@@ -0,0 +1,102 @@
...
+          "rules": [
+            {
+              "id": "SARIF1019",
+              "name": "RuleIdMustBePresentAndConsistent",
+              "shortDescription": {
+                "text": "{0}: The result contains neither result."
+              },
+              "fullDescription": {
+                "text": "{0}: The result contains neither result.ruleId nor result.rule.id."
+              },
```
lgolding, line 18: Changing the rule description message will affect this portion of the output file, but you won't have to edit it by hand. Just re-run the test and pick up the new "actual" output.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1019.RuleIdMustBePresentAndConsistent_Valid.sarif
```diff
@@ -0,0 +1,79 @@
...
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "Sarif Functional Testing"
+        }
+      },
+      "logicalLocations": [
```
lgolding, line 11: >logicalLocations [](start = 7, length = 16)

To demonstrate your fix, you don't need logical locations. In fact, you don't need any locations at all. (A `result` object is not required to have a `locations` property.) Just remove this, and remove the `locations` property from all the results.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1019.RuleIdMustBePresentAndConsistent_Valid.sarif
```diff
@@ -0,0 +1,79 @@
...
+        },
+        {
+          "name": "B"
+        }
+      ],
+      "results": [
+        {
+          "ruleId": "TST0001",
+          "message": {
+            "text": "Message."
```
lgolding, line 23: >Message [](start = 21, length = 7)

You can use this property to make each test case self-describing. For example: "This result contains only a ruleId property.", "This result contains only a rule.id property", "This result contains both a ruleId and a rule.id property, and they are the same".

Similarly, in the Invalid file, you can describe your result: "This result contains both a ruleId and a rule.id property, but they are different."

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1019.RuleIdMustBePresentAndConsistent_Valid.sarif
```diff
@@ -0,0 +1,79 @@
...
+            {
+              "logicalLocations": [
+                {
+                  "index": 1
+                }
+              ]
+            }
+          ],
+          "rule": {
+            "index": 1
```
lgolding, line 35: >"index": 1 [](start = 12, length = 10)

The presence of the `index` property makes this an invalid SARIF file. The index is supposed to point into the array `run.tool.driver.rules`, but that array does not exist in this sample. IMO the simplest fix is to remove the `index` property from each of the three `result.rule` objects. This means that the first result will have a `result.rule` object with no properties, so just remove it.

I'm surprised this test passes, because I thought we had a validation rule that requires all "index" properties to be consistent with the arrays they refer to.

src/Sarif.Multitool/Rules/SARIF1019.RuleIdMustBePresentAndConsistent.cs
eddynaka: ok!

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1019.RuleIdMustBePresentAndConsistent_Invalid.sarif
eddynaka: ok!

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1019.RuleIdMustBePresentAndConsistent_Valid.sarif
eddynaka: ok!

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1019.RuleIdMustBePresentAndConsistent_Invalid.sarif
```diff
@@ -0,0 +1,30 @@
...
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "Sarif Functional Testing"
+        }
+      },
+      "results": [
+        {
```
lgolding, line 12: >{ [](start = 8, length = 1)

Our convention is that the _Valid file contains only valid results, and the _Invalid file contains only invalid results. This result is valid, so remove it from the _Invalid file.

PR #1898, opened 2020-06-02
Upgrade ValidateCommand test inputs to v2.1.0 final.

lgolding:
Upgrade all test files to SARIF v2.1.0. This makes it much easier to write new validation rules, because we don't have to worry about how to express the valid and invalid conditions in earlier versions of the SARIF schema. And it's less important than it used to be to have these files server as additional exercises of our "upgrade" functionality, because (a) SARIF 2.1.0 is final, and (b) the upgrade functionality is now very well tested.

Also:
- Remove the `expectedResults` property bag property from all tests files. This was a remnant of an earlier test design, before we had the Inputs and ExpectedOutputs directories.


src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Invalid.sarif
```diff
@@ -1,38 +1,55 @@
...
-            {
-              "id": "TST0001",
-              "helpUri": "ht%tp://www.example.com/rules/tst0001.html"
-            }
-          ],
           "notifications": [
             {
               "id": "MSG0001",
-              "helpUri": "ht%tp://www.example.com/rules/msg0001.html"
+              "helpUri": "ht%25tp://www.example.com/rules/msg0001.html"
```
harleenkohli, line 13: >%25 [](start = 28, length = 3)

seems like a strange change to introduce, i am sure you had a reason to switch this to %25. just curious?

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Invalid.sarif
```diff
@@ -1,38 +1,55 @@
 {
-  "$schema": "ht%tp://json.schemastore.org/sarif-2.0.0",
```
harleenkohli, line 2: >ht%tp [](start = 14, length = 5)

This seems like a good test, i see that we have over-written it. was there a reason for this?

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1015.UriMustBeAbsolute_Invalid.sarif
```diff
@@ -42,21 +42,7 @@
           ]
         }
       ],
-      "columnKind": "utf16CodeUnits",
-      "properties": {
-        "expectedResults": {
```
harleenkohli, line 47: > "expectedResults": { [](start = 7, length = 21)

thanks for explanation offline, i remember using these in tests a long whiel ago. Glad to hear we no longer have a need for these.

src/Test.FunctionalTests.Sarif/TestData/Partitioning/Partition/ExpectedOutputs/TrivialPartitionFunction.sarif
```diff
@@ -51,6 +51,9 @@
       "results": [
         {
           "ruleId": "TST0001",
+          "message": {
+            "text": "Message 1.1."
+          },
```
harleenkohli, line 56: thanks for taking care of these. when i rebaselined locally, a bunch of these files were popping up changed. :)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Invalid.sarif
harleenkohli: it does appear from the expected output file that we never cuaght it to begin with. maybe thats why u removed the invalidity here.

---
In reply to: [434245686](https://github.com/microsoft/sarif-sdk/pull/1898#discussion_r434245686) [](ancestors = 434245686)

lgolding: An unintended consequence of `sarif transform`. Will revert.

---
In reply to: [434250204](https://github.com/microsoft/sarif-sdk/pull/1898#discussion_r434250204) [](ancestors = 434250204,434245686)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Invalid.sarif
lgolding: Again, a result of `sarif transform`. I don't know whether it's `System.Uri` or Newtonsoft that does it, but during reserialization, characters that are invalid in a URI are percent-escaped. But that renders the URI valid, which is exactly what we don't want for this test.

Reverting. Good catch.

---
In reply to: [434245424](https://github.com/microsoft/sarif-sdk/pull/1898#discussion_r434245424) [](ancestors = 434245424)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1003.UrisMustBeValid_Invalid.sarif
lgolding: Ok, I now remember the reason for this one. We want each test to exhibit exactly one class of error. In this test, we want the output to include _only_ `SARIF1003.UrisMustBeValid`. With the introduction of `SARIF1020.ReferToFinalSchema`, this line also caused a `SARIF1020` error.

HOWEVER: On further reflection, the value of keeping this line as a test case outweighs the value of uniform test output -- in this one case. So I am reverting this as you asked.

---
In reply to: [434587307](https://github.com/microsoft/sarif-sdk/pull/1898#discussion_r434587307) [](ancestors = 434587307,434250204,434245686)

PR #1899, opened 2020-06-02
fix #1889 validation rule: missing tool version

### LINKED ISSUE #1889: New rule: Tool version should always be provided.
run.tool.driver.version should be provided.

harleenkohli:


harleenkohli:
@lgolding @eddynaka i still need to merge the changes from Larry's PR (after its in master), but other than that it is ready for your review.

Please take a look.

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        protected override IEnumerable<string> MessageResourceNames => new string[]
+        {
+            nameof(RuleResources.SARIF1021_Default)
+        };
+
+        protected override void Analyze(Tool tool, string pointer)
+        {
+            // 'Driver' is a required property, hence we do not need a null check for it. 
+            if (tool.Driver.Version == null)
```
eddynaka, line 32: Here, you are just checking for a null value. The other cases that I can see:
- version is white space or empty: is that an error as well?
- version is not a version. For example, `tool.Driver.Version = "version"`: is that an error as well?

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                LogResult(pointer, nameof(RuleResources.SARIF1021_MissingToolVersion));
+                return;
+            }
+
```
eddynaka, line 37: I would remove the extra space and adjust the identation

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -140,6 +140,14 @@ public void SARIF1020_SchemaMustBePresentAndConsistent_Valid()
         public void SARIF1020_SchemaMustBePresentAndConsistent_Invalid()
             => RunTest(MakeInvalidTestFileName(RuleId.ReferToFinalSchema, nameof(RuleId.ReferToFinalSchema)));
 
+        [Fact]
+        public void SARIF1021_MissingToolVersion_Valid()
+    => RunTest(MakeValidTestFileName(RuleId.MissingToolVersion, nameof(RuleId.MissingToolVersion)));
```
eddynaka, line 145: I would adjust the identation

harleenkohli:
oops! i forgot. thank u!

---
In reply to: [423448617](https://github.com/microsoft/sarif-sdk/pull/1899#issuecomment-423448617) [](ancestors = 423448617)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
harleenkohli: great idea, i should add a check for whitespaces. reg the version scheme, yes that could be a potential test but i am worried if it will be too constraining. each tool is free to choose how they version themselves. 

---
In reply to: [434486864](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434486864) [](ancestors = 434486864)

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -219,4 +219,10 @@
   <data name="SARIF1020_SchemaReferenceMissing" xml:space="preserve">
     <value>{0}: The SARIF log file does not contain a $schema property. Add a $schema property that refers to the final version of the SARIF 2.1.0 schema. This enables IDEs to provide Intellisense for SARIF log files.</value>
   </data>
+  <data name="SARIF1021_Default" xml:space="preserve">
+    <value>{0}: The tool driver version is missing.</value>
```
lgolding, line 223: >. [](start = 50, length = 1)

We can include dynamic data:
`{0}: The description of the tool '{1}' does not include a 'version' property. The 'version' property is required in workflows that must verify the version of the tool being used."

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -219,4 +219,10 @@
...
+  </data>
+  <data name="SARIF1021_MissingToolVersion" xml:space="preserve">
+    <value>The 'version' property for tool driver should be present.</value>
```
lgolding, line 226: >T [](start = 11, length = 1)

And then, very similar to the result message:
"The property 'tool.driver.version' should be present. This property is important in workflows that must verify the version of the tool being used."

src/Sarif.Multitool/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -387,6 +387,11 @@ private void Visit(Conversion conversion, string conversionPointer)
                     Visit(conversion.AnalysisToolLogFiles[i], analysisToolLogFilesPointer.AtIndex(i));
                 }
             }
+
+            if (conversion.Tool != null)
+            {
+                Visit(conversion.Tool, conversionPointer.AtProperty(SarifPropertyName.Tool));
+            }
```
lgolding, line 394: >} [](start = 12, length = 1)

Yup.

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
lgolding: Yes, each tool can choose its versioning scheme. It is only `semanticVersion` that is constrained.

I see you expanded the check to include empty or white space.

---
In reply to: [434737873](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434737873) [](ancestors = 434737873,434486864)

src/Sarif.Multitool/Rules/RuleResources.resx
harleenkohli: nice! thanks!

---
In reply to: [434751150](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434751150) [](ancestors = 434751150)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class MissingToolVersion : SarifValidationSkimmerBase
+    {
+        private readonly MultiformatMessageString _fullDescription = new MultiformatMessageString
+        {
+            Text = RuleResources.SARIF1021_MissingToolVersion
+        };
+
+        public override MultiformatMessageString FullDescription => _fullDescription;
+
+        public override FailureLevel DefaultLevel => FailureLevel.Error;
```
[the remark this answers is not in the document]
harleenkohli: got it. i like the idea of reserving errors for spec violations. i will also check with Mike.

---
In reply to: [434754853](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434754853) [](ancestors = 434754853)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1001.SyntaxError.sarif
```diff
@@ -5,7 +5,8 @@
   "runs": [
     {
       "tool": {
-        "name": "CodeScanner"
+        "name": "CodeScanner",
+        "version": "1.0.0"
```
lgolding, line 9: No need for this. This is file is invalid JSON (see the comma on line 3).

harleenkohli: oh! nice :) Thanks


---
In reply to: [434767894](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434767894) [](ancestors = 434767894)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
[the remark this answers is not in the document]
harleenkohli: If Mike disagrees, we can revert this in future PR.

---
In reply to: [434762932](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434762932) [](ancestors = 434762932,434754853)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            nameof(RuleResources.SARIF1021_Default)
+        };
+
+        protected override void Analyze(Tool tool, string pointer)
+        {
+            // 'Driver' is a required property, hence we do not need a null check for it. 
+            if (string.IsNullOrWhiteSpace(tool.Driver.Version))
+            {
+                LogResult(pointer, nameof(RuleResources.SARIF1021_MissingToolVersion), tool.Driver.Name);
```
lgolding, line 34: >pointer [](start = 26, length = 7)

This should be `pointer.AtProperty(SarifPropertyNames.Driver)`. The place to report the error is (for example) `/runs/0/tool/driver` -- that's the object that we're complaining doesn't have the `version` property.

harleenkohli: OOH, thats wat it does. this explains why i was seeing the line numbers point to parent object in the output. fixing!

---
In reply to: [434787246](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434787246) [](ancestors = 434787246)

michaelcfanning:
>         public const string EndLineMustNotBeLessThanStartLine = "SARIF1012";

We should discuss/evaluate factoring of all rules together as we start this new initiative. These two checks, EndLineMustNotBeLessThanStartLine and EndColumnMustNotBeLessThanStartColumn are too granular. We should have a single rule that notes that Regions should be well-formed. One way to think about this is to ask the question, is it helpful/necessary to have two distinct help topics for these checks? The answer is, clearly not, it's the same sort of issue with the same trivial guidance to resolve. A second question is to ask whether a developer might want to exclude one class of defect over another. Clearly, these rules fail this test as well, there's no good reason to permit a problem with EndColumn but fire only on EndLine issues.

I see some other opportunities to collapse/refine but will stop here. I will open an issue on this.

---
Refers to: src/Sarif.Multitool/Rules/RuleId.cs:14 in 5a2a5ea. [](commit_id = 5a2a5eaa8eed8e827933906e0a5896b487680b8d, deletion_comment = False)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
+{
+    public class MissingToolVersion : SarifValidationSkimmerBase
+    {
+        private readonly MultiformatMessageString _fullDescription = new MultiformatMessageString
+        {
+            Text = RuleResources.SARIF1021_MissingToolVersion
+        };
+
+        public override MultiformatMessageString FullDescription => _fullDescription;
```
michaelcfanning, line 16: >_fullDescription [](start = 68, length = 16)

Can you please eliminate the _fullDescription field and simply assign the object construction directly here?

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
michaelcfanning: Reactivating. Eddy has a good idea for expanding this rule, which would be both to require the presence of a version and that it seems some sort of standard in terms of its contents. Note that this may require the rule to return multiple severities. It also raises the question of whether sub-analyses are covered by subrules. So probably a bit more discussion to do.

---
In reply to: [434753356](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434753356) [](ancestors = 434753356,434737873,434486864)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,40 @@
...
+        public override FailureLevel DefaultLevel => FailureLevel.Warning;
+
+        /// <summary>
+        /// SARIF1021
+        /// </summary>
+        public override string Id => RuleId.MissingToolVersion;
+
+        protected override IEnumerable<string> MessageResourceNames => new string[]
+        {
+            nameof(RuleResources.SARIF1021_Default)
```
michaelcfanning, line 27: >SARIF1021_Default [](start = 33, length = 17)

This isn't correct. You don't actually return this string from the rule, do you? Also, this data doesn't declare the actual user-facing string that you *are* returning.

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
michaelcfanning, line 16: >FullDescription [](start = 49, length = 15)

You should recapitulate the full description as a doc comment on this method, because it's helpful for rules readability. See the following BinSkim rule for thoughts on a standard in terms of commenting, organization, etc. 

[https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2010.DoNotMarkImportsSectionAsExecutable.cs](https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2010.DoNotMarkImportsSectionAsExecutable.cs)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
```diff
@@ -0,0 +1,40 @@
...
+        };
+
+        protected override void Analyze(Tool tool, string pointer)
+        {
+            // 'Driver' is a required property, hence we do not need a null check for it. 
+            if (string.IsNullOrWhiteSpace(tool.Driver.Version))
+            {
+                LogResult(pointer.AtProperty(SarifPropertyName.Driver), nameof(RuleResources.SARIF1021_MissingToolVersion), tool.Driver.Name);
```
michaelcfanning, line 35: >LogResult [](start = 16, length = 9)

It is very helpful to recapitulate the user-facing string as a code comment, because it makes it quite clear what's happening in this return condition. Also, I think LogResult should *require* an argument that indicates the failure severity. That information is also important for rules maintainability. See this example rule from BinSkim that demonstrates the standard: [https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2010.DoNotMarkImportsSectionAsExecutable.cs](https://github.com/microsoft/binskim/blob/master/src/BinSkim.Rules/PERules/BA2010.DoNotMarkImportsSectionAsExecutable.cs)

src/Sarif.Multitool/Rules/SARIF1021.MissingToolVersion.cs
lgolding: This is actually correct. This is the name of a resource string into which `LogResult` (below) formats the message arguments.

---
In reply to: [434826465](https://github.com/microsoft/sarif-sdk/pull/1899#discussion_r434826465) [](ancestors = 434826465)

harleenkohli:
closing this PR. Eddy already took care of this rule in rules validation feature branch (rule2005)

PR #1913, opened 2020-06-12
Incremental baselining fix

ScottLouvau:
With Jeremiah, added Run.MergeResultsFrom(Run), exposing the logic the baseliner uses to merge results. This will allow us to implement incremental run baselining in the SarifBaseliner service.

We've redesigned and renamed the RemapIndicesVisitor class to do this work. The class doesn't have any consumers in the Sarif SDK solution other than baselining, but it is a breaking change. The new name is RunMergingVisitor, which more clearly indicates the purpose, and it now takes a whole Run as the "historical / current" context and has a "Populate" method to add the merged collections to the output run. This will allow it to merge additional collections in the future without further design changes.

NOTE: We altered the baselining property merging behavior in this change. When baselining in "KeepOldest" property merging mode, the old logic would actually keep the old Artifact entirely. It makes sense to use old Result.Properties when merging, but using an old Artifact whole seems wrong. Rather than implementing Artifact property bag merging, we're assuming that behavior is not actually wanted. Please comment if that's not the case.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -6,6 +6,8 @@
 * BUGFIX: `WorkItemFiler` now logs correctly the details for `LogMetricsForProcessedModel` method [#1896](https://github.com/microsoft/sarif-sdk/issues/1896)
 * FEATURE: Add validation rule `SARIF1019`, which requires every result to have at least one of `result.ruleId` and `result.rule.id`. If both are present, they must be equal. [#1880](https://github.com/microsoft/sarif-sdk/issues/1880)
 * FEATURE: Add validation rule `SARIF1020`, which requires that the $schema property should be present, and must refer to the final version of the SARIF 2.1.0 schema.  [#1890](https://github.com/microsoft/sarif-sdk/issues/1890)
+* FEATURE: Expose Run.MergeResultsFrom(Run) to merge Results from multiple Runs using code from result matching algorithm.
+* BREAKING: Rename 'RemapIndicesVisitor' to 'RunMergingVisitor' and redesign to control how much merging occurs internally.
 
 ## **v2.2.5** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.2.5) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.2.5) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.2.5) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.2.5)
 * BUGFIX: Fix SDK doubling Uris with certain escaped characters (ex: '-' and '_') on every Load/Save cycle (cause: https://github.com/dotnet/runtime/issues/36288)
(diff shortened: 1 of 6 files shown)


src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -1,82 +1,113 @@
...
+    /// </remarks>
+    public class RunMergingVisitor : SarifRewritingVisitor
     {
-        public RemapIndicesVisitor(
-            IList<Artifact> currentArtifacts,
-            IList<LogicalLocation> currentLogicalLocations)
-        {
-            CurrentArtifacts = new List<Artifact>();
-            CurrentLogicalLocations = new List<LogicalLocation>();
+        private List<Result> Results { get; set; }
```
lgolding, line 28: >Results [](start = 29, length = 7)

It looks like these first four properties can be `{ get; }`-only.

src/ReleaseHistory.md
```diff
@@ -6,6 +6,8 @@
 * BUGFIX: `WorkItemFiler` now logs correctly the details for `LogMetricsForProcessedModel` method [#1896](https://github.com/microsoft/sarif-sdk/issues/1896)
 * FEATURE: Add validation rule `SARIF1019`, which requires every result to have at least one of `result.ruleId` and `result.rule.id`. If both are present, they must be equal. [#1880](https://github.com/microsoft/sarif-sdk/issues/1880)
 * FEATURE: Add validation rule `SARIF1020`, which requires that the $schema property should be present, and must refer to the final version of the SARIF 2.1.0 schema.  [#1890](https://github.com/microsoft/sarif-sdk/issues/1890)
+* FEATURE: Expose Run.MergeResultsFrom(Run) to merge Results from multiple Runs using code from result matching algorithm.
+* BREAKING: Rename 'RemapIndicesVisitor' to 'RunMergingVisitor' and redesign to control how much merging occurs internally.
```
michaelcfanning, line 10: >BREAKING [](start = 2, length = 8)

Just a consistency nit, we use markdown back-ticks to enquote things like `RemapIndicesVisitor`. A method name like `Run.MergeResultsFrom(Run)` would also typically be enquoted. This is the sort of thing that Larry and I typically scrub for before we update to NuGet (which we should do shortly), so we can clean it up for you later if you are UTG. @lgolding,I do suggest we publish 2.2.6 soon, particularly to pick up the empty `results` array. Before doing that, let's talk offline about the automatic version control provenance generation I have in mind.

src/Test.UnitTests.Sarif/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
```diff
@@ -359,42 +359,41 @@ public void SarifLogResultMatcher_MultipleLogsDuplicateData_WorksAsExpected()
         [Fact]
         public void SarifLogResultMatcher_PreservesPropertiesProperly()
         {
+            // Verify that SarifLog matching keeps old or latest property values on matched Results
+            // depending on how it was configured.
+
+            // This test used to verify that old Artifact properties are also preserved, but this behavior
```
michaelcfanning, line 365: >This [](start = 15, length = 4)

I'd have to think about this more to understand why we put this behavior in originally, i.e., whether it was driven by a concrete scenario or just some fuzzy thinking. On surface, if you are allowing the oldest thing to have precedence, you might prefer to retain a sense of the artifact at the time of the oldest run. looking over the properties on the artifact object, though, I tend to agree this really doesn't make much sense. perhaps there'd be a use for knowing the last modified time at some earlier date. Possibly the 'roles' in play at the earliest date might be relevant. In any case, these remarks are entirely consistent with your observation that merging will be a careful, property-specific operation, rather than a blanket retention of one version over the other

src/Sarif/Visitors/RunMergingVisitor.cs
ScottLouvau: Fixed.

src/ReleaseHistory.md
ScottLouvau: Fixed.

src/Test.UnitTests.Sarif/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
ScottLouvau: Yeah - it's wrong the way it was. By removing the logic and releasing, we'll find out if anyone really wanted the behavior. Since it'll be complex to fix properly (merging the latest Artifact with the corresponding older properties) it seemed better to release and find out if this matters.

PR #1924, opened 2020-06-22
Serialization Consistency fixes

ScottLouvau:
Serialization improvements to make SDK serialization more consistent.
 - Add CodeGenHints and regenerate OM to ensure all DateTime properties use DateTimeConverter.
 - Fix properties serialization to keep DateTime format when roundtripped.
 - Consistently load and emit properties without indentation (saves memory when loaded, consistent output, and 20% faster overall log load when Results have properties).

### DIFF
--- a/src/Sarif/CodeGenHints.json
+++ b/src/Sarif/CodeGenHints.json
@@ -731,6 +731,26 @@
       }
     }
   ],
+  "ResultProvenance.FirstDetectionTimeUtc": [
+    {
+      "kind": "AttributeHint",
+      "arguments": {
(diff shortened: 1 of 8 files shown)


src/Sarif/CodeGenHints.json
```diff
@@ -731,6 +731,26 @@
       }
     }
   ],
+  "ResultProvenance.FirstDetectionTimeUtc": [
```
michaelcfanning, line 734: >ResultProvenance [](start = 3, length = 16)

I see. So we just had inconsistently applied hints pointing to this converter for all UTC types in the schema.

PR #1926, opened 2020-06-23
Adding rule SARIF2005

eddynaka:


src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Json.Pointer;
+using System.Collections.Generic;
```
lgolding, line 5: >Generic [](start = 25, length = 7)

Our convention is to put the `System` usings first.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+            nameof(RuleResources.SARIF2005_ProvideHelpfulToolInformation_Warning_ProvideConciseToolName_Text),
+            nameof(RuleResources.SARIF2005_ProvideHelpfulToolInformation_Warning_UseNumericToolVersions_Text)
+        };
+
+        protected override void Analyze(Tool tool, string toolPointer)
+        {
+            if (tool == null)
+            {
+                return;
+            }
```
lgolding, line 33: >} [](start = 12, length = 1)

You don't need this check because the "visitor" code will not call this method if `tool` is `null`.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            if (tool.Driver != null)
+            {
+                AnalyzeToolDriver(tool.Driver, toolPointer);
```
lgolding, line 37: >toolPointer [](start = 47, length = 11)

Pass `toolPointer.AtProperty(SarifPropertyName.Driver)` as the second argument.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        private void AnalyzeToolDriver(ToolComponent toolComponent, string toolPointer)
```
lgolding, line 41: >toolPointer [](start = 75, length = 11)

Rename: `toolDriverPointer`

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            // ProvideConciseToolName: Ensure that tool.driver.name isn't more than 3 words long
+            if (!string.IsNullOrEmpty(toolComponent.Name))
+            {
+                const int maxWordCount = 3;
+                int wordCount = toolComponent.Name.Split(' ').Length;
```
lgolding, line 47: >Split [](start = 51, length = 5)

Use `StringSplitOptions.RemoveEmptyEntries` in case there is more than one space between words.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+                AnalyzeToolDriver(tool.Driver, toolPointer);
+            }
+        }
+
+        private void AnalyzeToolDriver(ToolComponent toolComponent, string toolPointer)
+        {
+            // ProvideConciseToolName: Ensure that tool.driver.name isn't more than 3 words long
+            if (!string.IsNullOrEmpty(toolComponent.Name))
+            {
+                const int maxWordCount = 3;
```
lgolding, line 46: >maxWordCount [](start = 26, length = 12)

Our convention is to use PascalCase for constants: `MaxWords`.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+                int wordCount = toolComponent.Name.Split(' ').Length;
+                if (wordCount > maxWordCount)
+                {
+                    string driverPointer = toolPointer.AtProperty("driver").AtProperty("name");
```
lgolding: >driverPointer [](start = 27, length = 13)

`driverNamePointer`

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -204,4 +204,16 @@
   <data name="SARIF1007_RegionPropertiesMustBeConsistent_FullDescription_Text" xml:space="preserve">
     <value>Placeholder_SARIF1007_RegionPropertiesMustBeConsistent_FullDescription_Text</value>
   </data>
+  <data name="SARIF2005_ProvideHelpfulToolInformation_FullDescription_Text" xml:space="preserve">
+    <value>Placeholder</value>
+  </data>
+  <data name="SARIF2005_ProvideHelpfulToolInformation_Warning_ProvideConciseToolName_Text" xml:space="preserve">
+    <value>{0}: Placeholder {1} {2}</value>
```
lgolding, line 211: >{1} {2} [](start = 28, length = 7)

Put single quotes around the replacement sequences (except for the `{0}`. So: `{0}: Placeholder '{1}' '{2}'`

lgolding: The reason is that the replacement values might be strings with spaces in them, or even empty strings, so we want to see clearly where each replacement value is. In the real messages, I won't put quotes around integer values.

---
In reply to: [444574154](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444574154) [](ancestors = 444574154)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+                        wordCount.ToString(),
+                        maxWordCount.ToString());
+                }
+            }
+
+            // ProvideToolVersion: Either tool.driver.version or tool.driver.semanticVersion should be there.
+            if (string.IsNullOrWhiteSpace(toolComponent.Version) && string.IsNullOrWhiteSpace(toolComponent.SemanticVersion))
+            {
+                LogResult(
+                        toolPointer,
```
lgolding: >toolPointer [](start = 24, length = 11)

`toolDriverPointer`. The idea is that the bad object is the driver itself. You want to say "There's something wrong with tool/driver. What's wrong? It doesn't have a version."

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+                else
+                {
+                    AnalyzeVersion(toolComponent.SemanticVersion, toolPointer.AtProperty("driver").AtProperty("semanticVersion"));
+                }
+            }
+        }
+
+        private void AnalyzeVersion(string version, string pointer)
+        {
+            Regex regex = new Regex(@"^\d+\.\d+.*");
```
lgolding, line 78: >Regex regex = new Regex(@"^\d+\.\d+.*"); [](start = 12, length = 40)

We only want to construct this regex once. So, outside of this method, put:
```
private static readonly s_versionRegexe = new Regex(@"^\d+\.\d+.*", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
and use that here.

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+            Match match = regex.Match(version);
+            if (!match.Success)
```
lgolding, line 80: You can just say
```
if (!regex.IsMatch(version))
{
```

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
eddynaka: for net461, we need to do a cast to (char).


---
In reply to: [444573658](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444573658) [](ancestors = 444573658)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+            else
+            {
+                // UseNumericToolVersions
+                if (!string.IsNullOrWhiteSpace(toolComponent.Version))
+                {
+                    AnalyzeVersion(toolComponent.Version, toolPointer.AtProperty("driver").AtProperty("version"));
+                }
+                else
+                {
+                    AnalyzeVersion(toolComponent.SemanticVersion, toolPointer.AtProperty("driver").AtProperty("semanticVersion"));
```
lgolding: >AnalyzeVersion(toolComponent.SemanticVersion, toolPointer.AtProperty("driver").AtProperty("semanticVersion")); [](start = 20, length = 110)

The `semanticVersion` doesn't match the simple regex we are using for `version`. It actually matches the complicated expression you showed me earlier. Hang on while I find the official source for this...

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
eddynaka: should i point to version? 

---
In reply to: [444574759](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444574759) [](ancestors = 444574759)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
eddynaka: Can i change to RegexVersion? or something like that?

---
In reply to: [444575685](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444575685) [](ancestors = 444575685)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: See the regexes at the end of SemVer.org.

---
In reply to: [444576559](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444576559) [](ancestors = 444576559)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: Yikes, really??? That's can't be right, let me look. I think you might be using the wrong overload.

---
In reply to: [444575924](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444575924) [](ancestors = 444575924,444573658)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: Not on this one. The reason is that this `driver` object doesn't _have_ a `version` property. So the pointer would point nowhere. And again, the object that is _faulty_ is `tool.driver`. It's faulty because it doesn't have a version! So the message should point to the faulty object, `tool.driver`.

---
In reply to: [444576865](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444576865) [](ancestors = 444576865,444574759)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
eddynaka: at least, net461 didnt like it...

---
In reply to: [444577532](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444577532) [](ancestors = 444577532,444575924,444573658)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: 1) I made a typo. It should be `s_versionRegex`. I added an extra `e` at the end, sorry.

---
In reply to: [444577200](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444577200) [](ancestors = 444577200,444575685)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
eddynaka: ok. just created another static regex and passed as variable in the method

---
In reply to: [444577239](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444577239) [](ancestors = 444577239,444576559)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: Yes, this is wrong. It should be `Split(new char[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)`. Take a look at the list of overloads of `String.Split` to see why.

---
In reply to: [444577532](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444577532) [](ancestors = 444577532,444575924,444573658)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
lgolding: The problem with that is that the message will be wrong. For `semanticVersion` you'd want to say something like "semanticVersion doesn't match the SemVer.org specification". For `version` you'd want to say "Version isn't numeric".

---
In reply to: [444578598](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444578598) [](ancestors = 444578598,444577239,444576559)

lgolding: For now, _do not verify_ `semanticVersion`. Just remove this branch of the logic. There needs to be a separate rule for `semanticVersion`.

---
In reply to: [444579925](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444579925) [](ancestors = 444579925,444578598,444577239,444576559)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2005.ProvideHelpfulToolInformation_Valid.sarif
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "SARIF Functional Testing",
+          "version": "1.2.3"
```
lgolding, line 9: >3 [](start = 26, length = 1)

Let's make it `"1.2.3s"` to show that we accept a non-numeric suffix.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF2005.ProvideHelpfulToolInformation_Invalid.sarif
```diff
@@ -0,0 +1,102 @@
...
+            "uriBaseId": "TEST_DIR"
+          }
+        }
+      ],
+      "results": [
+        {
+          "ruleId": "SARIF2005",
+          "ruleIndex": 0,
+          "message": {
+            "id": "Warning_ProvideConciseToolName",
```
lgolding, line 53: >Warning_ProvideConciseToolName [](start = 19, length = 30)

This should not fire because the name "SARIF functional testing" above is three words. This will go away once you fix your `Split` call.

BUT! Once you have verified that, please change the name to "SARIF functional testing long" to make sure that it _does_ fire when it's more than three words.

eddynaka: If you look at the test, i'm adding the "SARIF functional testing long". But, when we get the actual, its changing the sarif to a valid one.

---
In reply to: [444580896](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444580896) [](ancestors = 444580896)

lgolding: I'm not sure what you mean. The input file has "SARIF Functional Testing long". The expected output file (this file) correctly says that it found 4 words instead of the maximum of 3 (lines 56 and 57 below). If the "actual" file matches the "expected output" file, the test would pass. Are you saying that the test fails? If so, can you send me the "actual" file?

---
In reply to: [444582777](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444582777) [](ancestors = 444582777,444580896)

src/Sarif.Multitool/Rules/SARIF2005.ProvideHelpfulToolInformation.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (!string.IsNullOrEmpty(toolComponent.Name))
+            {
+                const int MaxWords = 3;
+                int wordCount = toolComponent.Name.Split(new char[] { ' ' }, StringSplitOptions.RemoveEmptyEntries).Length;
+                if (wordCount > MaxWords)
+                {
+                    string driverNamePointer = toolDriverPointer.AtProperty(SarifPropertyName.Name);
+                    LogResult(
+                        driverNamePointer,
+                        nameof(RuleResources.SARIF2005_ProvideHelpfulToolInformation_Warning_ProvideConciseToolName_Text),
```
lgolding, line 51: >SARIF2005_ProvideHelpfulToolInformation_Warning_ProvideConciseToolName_Text [](start = 45, length = 75)

Let's do one more thing. Let's add an argument which is the actual tool driver name. Let's make that argument 1, argument 2 is the word count, and argument 3 is the actual word count. Don't forget to update the placeholder resource string.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF2005.ProvideHelpfulToolInformation_Invalid.sarif
eddynaka: what i'm saying is that when the actual contains "SARIF Functional Testing Long", the output tool.driver.name is changed to "SARIF Functional Testing". So, probably, the multitool is solving the issue as well, creating a valid sarif file

---
In reply to: [444604279](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444604279) [](ancestors = 444604279,444582777,444580896)

lgolding: The test populates `tool.driver`. It's the _results_ that tell us whether the analysis was correct.

---
In reply to: [444743290](https://github.com/microsoft/sarif-sdk/pull/1926#discussion_r444743290) [](ancestors = 444743290,444604279,444582777,444580896)

PR #1929, opened 2020-06-24
Adding rule SARIF2001

eddynaka:


src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
-                };
+                AnalyzeToolDriver(tool.Driver, toolPointer.AtProperty(SarifPropertyName.Driver));
             }
         }
 
-        protected override void Analyze(ReportingDescriptor reportingDescriptor, string reportingDescriptorPointer)
+        private void AnalyzeToolDriver(ToolComponent toolComponent, string toolDriverPointer)
         {
-            AnalyzeMessageStrings(reportingDescriptor.MessageStrings, reportingDescriptorPointer);
+            if (toolComponent.Rules != null && toolComponent.Rules.Count > 0)
```
lgolding, line 45: > && toolComponent.Rules.Count > 0 [](start = 43, length = 33)

You don't need this check because if Rules is empty, then the `foreach` loop will simply execute 0 times.

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
+            {
+                foreach (ReportingDescriptor rule in toolComponent.Rules)
+                {
+                    AnalyzeReportingDescriptor(rule, toolDriverPointer.AtProperty(SarifPropertyName.Rules));
```
lgolding, line 49: >toolDriverPointer.AtProperty(SarifPropertyName.Rules) [](start = 53, length = 53)

You actually want to point to the specific rule, say, the rule at index 0. To do that:

1. Before the `foreach` loop, define `string rulesPointer = toolDriverPointer.AtProperty(SarifPropertyName.Rules;`
2. Inside the `foreach` loop, keep a counter (or just use a simple `for` loop. Suppose the counter is `index`.
3. Then when you call `AnalyzeReportingDescriptor`, the pointer argument should be `rulesPointer.AtIndex(index)`.

lgolding: The result will be `/runs/0/tool/driver/rules/0` (for the rule at index 0) and so on.

---
In reply to: [445100702](https://github.com/microsoft/sarif-sdk/pull/1929#discussion_r445100702) [](ancestors = 445100702)

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
+            }
         }
 
-        private void AnalyzeMessageStrings(
-            IDictionary<string, MultiformatMessageString> messageStrings,
-            string reportingDescriptorPointer)
+        private void AnalyzeReportingDescriptor(ReportingDescriptor rule, string reportingDescriptorPointer)
         {
-            if (messageStrings != null)
+            if (rule.MessageStrings != null && rule.MessageStrings.Count > 0)
```
lgolding, line 56: > && rule.MessageStrings.Count > 0 [](start = 43, length = 33)

This check isn't needed.

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
-
-                foreach (string key in messageStrings.Keys)
+                foreach (KeyValuePair<string, MultiformatMessageString> message in rule.MessageStrings)
                 {
-                    MultiformatMessageString messageString = messageStrings[key];
-
-                    string messageStringPointer = messageStringsPointer.AtProperty(key);
-
-                    AnalyzeMessageString(messageString.Text, messageStringPointer, SarifPropertyName.Text);
+                    AnalyzeMessageString(message.Value.Text, messageStringsPointer.AtProperty(message.Key));
```
lgolding, line 61: >message.Value.Text [](start = 41, length = 18)

You'll also need to pass the message.Key to `AnalyzeMessageString` because it will surely be part of the message string. For example, "The message string 'x' in rule 'y' has a problem..."

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
-        }
+            string textPointer = messagePointer.AtProperty(SarifPropertyName.Text);
 
-        private void AnalyzeMessageString(
-            string messageString,
-            string messagePointer,
-            string propertyName)
-        {
-            if (!string.IsNullOrEmpty(messageString) && DoesNotEndWithPeriod(messageString))
+            // IncludeDynamicContent: check if messageString has dynamic content
```
lgolding, line 75: >t [](start = 79, length = 1)

The comment should be a complete sentence, so end in a period.

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
+            // EnquoteDynamicContent: check if messageString has enquoted dynamic content
+            if (!s_enquoteDynamicContentRegex.IsMatch(messageString))
             {
-                string textPointer = messagePointer.AtProperty(propertyName);
+                // {0}: Placeholder '{1}'
+                LogResult(
+                    textPointer,
+                    nameof(RuleResources.SARIF2001_AuthorHighQualityMessages_Warning_EnquoteDynamicContent_Text),
+                    messageString
+                    );
```
lgolding, line 94: >); [](start = 20, length = 2)

Put closing paren on same line as the last argument. You did it correctly on Line 104 below.

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
eddynaka: ok. just changed to a `for`, so i dont need to work with the int variable.

---
In reply to: [445101027](https://github.com/microsoft/sarif-sdk/pull/1929#discussion_r445101027) [](ancestors = 445101027,445100702)

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
+        protected override IEnumerable<string> MessageResourceNames => new string[] {
+            nameof(RuleResources.SARIF2001_AuthorHighQualityMessages_Warning_EnquoteDynamicContent_Text),
+            nameof(RuleResources.SARIF2001_AuthorHighQualityMessages_Warning_IncludeDynamicContent_Text),
+            nameof(RuleResources.SARIF2001_AuthorHighQualityMessages_Warning_TerminateWithPeriod_Text)
         };
 
         public override FailureLevel DefaultLevel => FailureLevel.Warning;
 
-        public override string Id => RuleId.AuthorHighQualityMessages;
+        private static readonly Regex s_dynamicContentRegex = new Regex(@"{[0-9]+}", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
lgolding, line 32: >{ [](start = 74, length = 1)

Curly brace '{' is a regular expression metacharacter, so you must escape it: @"\{[0-9]+\}"

src/Sarif.Multitool/Rules/SARIF2001.AuthorHighQualityMessages.cs
```diff
@@ -3,85 +3,106 @@
...
+        private static readonly Regex s_enquoteDynamicContentRegex = new Regex(@"'{[0-9]+}'", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
lgolding, line 33: >s_enquoteDynamicContentRegex [](start = 38, length = 28)

How about this:

```
private static readonly Regex s_nonEnquotedDynamicContextRegex = new Regex(@"(^|[^'])\{[0-9]+\}", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```

If the start of the line (`^`) _or_ anything that is _not_ a single quote (`[^']`) is followed by a replacement sequence (`\{[0-9]+\}`), that's a bad thing. So you test:

```
if (s_nonEnquotedDynamicContentRegex.IsMatch(text))
{
    // Report error.
}
```

PR #1932, opened 2020-06-24
Fix #1915: Allow result message to be truncated

### LINKED ISSUE #1915: Allow result message to be truncated
The extension method `Result.GetMessageText` takes an optional parameter `bool concise` (default: `false`) which truncates the message to the first sentence. Provide an additional optional parameter `int maxLength` (default: `120`?) to truncate the message before the end of the first sentence if the sentence is too long.

lgolding:
I chose a default maxLength of 200.


src/Sarif/ExtensionMethods.cs
```diff
@@ -333,6 +333,10 @@ public static string GetMessageText(this Result result, ReportingDescriptor rule
             if (concise)
             {
                 text = GetFirstSentence(text);
+                if (text.Length > maxLength)
+                {
+                    text = text.Substring(0, maxLength) + "...";
```
michaelcfanning, line 338: >... [](start = 59, length = 3)

Why aren't you using the ellipsis char?

src/Sarif/ExtensionMethods.cs
```diff
@@ -288,7 +288,7 @@ public static string GetMessageText(this Result result, ReportingDescriptor rule
             return GetMessageText(result, rule, concise: false);
         }
 
-        public static string GetMessageText(this Result result, ReportingDescriptor rule, bool concise = false)
+        public static string GetMessageText(this Result result, ReportingDescriptor rule, bool concise = false, int maxLength = 200)
```
michaelcfanning, line 291: >200 [](start = 128, length = 3)

is 120 or 128 a better default?

src/Test.UnitTests.Sarif/SarifExtensionsTests.cs
```diff
@@ -113,5 +113,29 @@ public void SarifExtensions_Result_GetMessageText_Concise()
...
+            var rule = new ReportingDescriptor
+            {
+                MessageStrings = new Dictionary<string, MultiformatMessageString>
+                {
+                    ["ruleStr1"] = new MultiformatMessageString { Text = "First sentence is very long. Second sentence." }
+                }
+            };
+
+            const string Expected = "First sentence is ve...";
+            string actual = result.GetMessageText(rule, concise: true, maxLength: 20);
```
michaelcfanning, line 137: compute this from your expected value (length - length of your trailing dots or ellipses.

src/Sarif/ExtensionMethods.cs
lgolding: When I filed the issue I said 120; when I coded it, I worried it was too short. But OK, anybody who calls the API and decides it's not enough can always increase it.

---
In reply to: [445546182](https://github.com/microsoft/sarif-sdk/pull/1932#discussion_r445546182) [](ancestors = 445546182)

michaelcfanning: The key question is 'too short for what'? i.e., under what circumstances would someone be likely to request a concise string and expect something around 200 characters. If you have a scenario in mind, let me know. I've mostly been thinking about flat list views in IDEs as a driver for this API, or work item titles. In these contexts, 200 chars seems more than what's required. Open to thinking about other UX scenarios if you have one.

---
In reply to: [445558418](https://github.com/microsoft/sarif-sdk/pull/1932#discussion_r445558418) [](ancestors = 445558418,445546182)

src/Sarif/ExtensionMethods.cs
```diff
@@ -333,6 +333,10 @@ public static string GetMessageText(this Result result, ReportingDescriptor rule
             if (concise)
             {
                 text = GetFirstSentence(text);
+                if (text.Length > maxLength)
+                {
+                    text = text.Substring(0, maxLength) + "\u2026";
```
michaelcfanning, line 338: >\u2026 [](start = 59, length = 6)

You should either directly embed the ellipsis or at least comment this so that readers know what the character is.

PR #1934, opened 2020-06-24
Provide messages for SARIF1005.ProvideHelpfulToolInformation

lgolding:
Also:
- Loosen up the regex to enable a single integer followed by arbitrary characters, for example `1x`.
- Add the tool name as an argument to the "missing version" message.

### DIFF
--- a/src/Sarif.Multitool/Rules/RuleResources.Designer.cs
+++ b/src/Sarif.Multitool/Rules/RuleResources.Designer.cs
@@ -344,7 +344,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Placeholder.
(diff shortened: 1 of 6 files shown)


src/Sarif.Multitool/Rules/RuleResources.Designer.cs
```diff
@@ -344,7 +344,11 @@ internal class RuleResources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Placeholder.
+        ///   Looks up a localized string similar to Provide information that makes it easy to identify the name and version of your tool.
+        ///
+        ///The tool&apos;s &apos;name&apos; property should be no more than three words long. This makes it easy to remember and allows it to fit into a narrow column when displaying a list of results. If you need to provide more information about your tool, use the &apos;fullName&apos; property.
```
harleenkohli, line 349: >///The tool&apos;s &apos;name&apos; property should be no more than three words long. This m [](start = 8, length = 92)

split into multiple lines for readability? (other places in the file as well).

PR #1935, opened 2020-06-24
sarif1007 subrule: RegionStartPropertyMustBePresent

harleenkohli:


src/Sarif.Multitool/Rules/SARIF1007.RegionPropertiesMustBeConsistent.cs
```diff
@@ -2,38 +2,54 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+
 using Microsoft.Json.Pointer;
+
 using Newtonsoft.Json.Linq;
```
lgolding, line 8: >u [](start = 0, length = 1)

Yes, this is why I don't like the "separate using groups" option :-) But I won't argue it!

src/Sarif.Multitool/Rules/SARIF1007.RegionPropertiesMustBeConsistent.cs
```diff
@@ -2,38 +2,54 @@
...
+        public override FailureLevel DefaultLevel => FailureLevel.Error;
 
         protected override void Analyze(Region region, string regionPointer)
         {
             var jsonPointer = new JsonPointer(regionPointer);
             JToken regionToken = jsonPointer.Evaluate(Context.InputLogToken);
 
+            if (!region.IsBinaryRegion &&
+                !region.IsLineColumnBasedTextRegion &&
+                !region.IsOffsetBasedTextRegion)
```
lgolding, line 39: Nice!!

src/Sarif.Multitool/Rules/SARIF1007.RegionPropertiesMustBeConsistent.cs
harleenkohli: "don't like" ??? i'd say - i hate it with the fire of a thousand suns

---
In reply to: [445228637](https://github.com/microsoft/sarif-sdk/pull/1935#discussion_r445228637) [](ancestors = 445228637)

PR #1939, opened 2020-06-25
sarif2004 eliminate location only artifacts

harleenkohli:


src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public override FailureLevel DefaultLevel => FailureLevel.Warning;
+
+        protected override void Analyze(Run run, string runPointer)
+        {
+            // We only verify first item in the results and artifacts array,
+            // since tools will typically generate similar nodes.
+            // This approach may cause false negatives.
+            ArtifactLocation firstLocationInArtifactsArray = run.Artifacts[0].Location;
+            ArtifactLocation firstlocationInResults = run.Results[0].Locations[0].PhysicalLocation.ArtifactLocation;
```
harleenkohli, line 34: i am unsure if i have gotten it right. this is what i understood from our convo, so i could be wrong!

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!ArrayBasedLocationHasAdditionalInfo(firstLocationInArtifactsArray, firstlocationInResults))
+            {
+                // {0}: Placeholder_SARIF2004_OptimizeFileSize_Warning_EliminateLocationOnlyArtifacts_Text
+                LogResult(
+                    runPointer.AtProperty(SarifPropertyName.Artifacts),
+                    nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_EliminateLocationOnlyArtifacts_Text));
+            }
+        }
+
+        private bool ArrayBasedLocationHasAdditionalInfo(ArtifactLocation arrayBasedLocation, ArtifactLocation resultBasedLocation)
```
harleenkohli, line 45: >ArrayBasedLocationHasAdditionalInfo [](start = 21, length = 35)

if the logic is indeed what we want to verify, then the name of this sub-rule needs to be changed. we are trying to verify if the array based location has any additonal info vs what result has.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -0,0 +1,65 @@
...
+                    nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_EliminateLocationOnlyArtifacts_Text)
+                };
+
+        public override FailureLevel DefaultLevel => FailureLevel.Warning;
+
+        protected override void Analyze(Run run, string runPointer)
+        {
+            // We only verify first item in the results and artifacts array,
+            // since tools will typically generate similar nodes.
+            // This approach may cause false negatives.
```
lgolding, line 32: >negatives [](start = 45, length = 9)

Yes, that's what we discussed.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
lgolding: Yes, that is what we discussed. Careful about possible nulls: `run?.Artifacts?[0]?.Location` and `run?.Results?[0].Locations?[0]?.PhysicalLocation?.ArtifactLocation`.

---
In reply to: [445259975](https://github.com/microsoft/sarif-sdk/pull/1939#discussion_r445259975) [](ancestors = 445259975)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
lgolding: Not quite: Let's discuss the logic offline, and I'll craft a pseudo-SARIF sample to illustrate.

---
In reply to: [445260360](https://github.com/microsoft/sarif-sdk/pull/1939#discussion_r445260360) [](ancestors = 445260360)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        private bool HasLocationOnlyArtifacts(string artifactPointer)
+        {
+            JToken artifactToken = artifactPointer.ToJToken(Context.InputLogToken);
+            return
+                artifactToken.HasProperty(SarifPropertyName.Location) &&
+                artifactToken.Children().Count() == 1;
+        }
+    }
```
lgolding, line 77: >} [](start = 4, length = 1)

All very nice and clean.

PR #1942, opened 2020-06-27
Adding Rule SARIF2006

eddynaka:
- Adding rule SARIF2006
- Adding static HttpClient to be reusable
- Adding static Dictionary to prevent useless calls during the analysis


src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2006.UrisShouldBeReachable_Valid.sarif
```diff
@@ -0,0 +1,40 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "CodeScanner",
+          "version": "1.0",
+          "downloadUri": "https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html",
```
lgolding, line 10: >https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html [](start = 26, length = 64)

Let's make this [https://github.com/microsoft/sarif-sdk](https://github.com/microsoft/sarif-sdk). I don't want to ping OASIS.

src/Sarif.Multitool/Rules/SARIF2006.UrisShouldBeReachable.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        private void AnalyzeUri(string uriString, string pointer)
+        {
+            // If it's not a well-formed URI of _any_ kind, then don't bother triggering this rule.
+            // Rule SARIF1003, UrisMustBeValid, will catch it.
+            // Check for well-formedness first, before attempting to create a Uri object, to
+            // avoid having to do a try/catch. Unfortunately Uri.TryCreate will return true
+            // even for a malformed URI string.
+            if (uriString != null && Uri.IsWellFormedUriString(uriString, UriKind.RelativeOrAbsolute))
```
lgolding, line 80: You can simplify this logic because if it's not absolute, SARIF1005.UriMustBeAbsolute will catch it. So:

```
        private void AnalyzeUri(string uriString, string pointer)
        {
            // If it's not a well-formed URI, or if it's not absolute, then don't bother triggering this rule.
            // If it's not well-formed, SARIF1003.UrisMustBeValid will catch it, and if it's not absolute,
            // SARIF1005.UriMustBeAbsolute will catch it.
            //
            // Check for well-formedness first, before attempting to create a Uri object, to
            // avoid having to do a try/catch. Unfortunately Uri.TryCreate will return true
            // even for a malformed URI string.
            if (uriString != null && Uri.IsWellFormedUriString(uriString, UriKind.Absolute))
            {
                // Ok, it's a well-formed absolute URI. If it's not reachable, _now_ we can report it.
                Uri uri = new Uri(uriString, UriKind.Absolute);
                if (!IsUriReachable(uri.AbsoluteUri))
                {
                    // Placeholder
                    LogResult(pointer, nameof(RuleResources.SARIF1005_UriMustBeAbsolute_Error_Default_Text), uriString);
                }
            }
        }
```

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2006.UrisShouldBeReachable_Valid.sarif
lgolding: I do understand that the validator always uses the OASIS copy of the spec for the `helpUri` so there's nothing you can do about that.

---
In reply to: [446546315](https://github.com/microsoft/sarif-sdk/pull/1942#discussion_r446546315) [](ancestors = 446546315)

src/Sarif.Multitool/Rules/SARIF2006.UrisShouldBeReachable.cs
eddynaka: If i do that and the uri is not absolute, we will have a null exception. It's why I added. Unless I use the uriString

---
In reply to: [446547297](https://github.com/microsoft/sarif-sdk/pull/1942#discussion_r446547297) [](ancestors = 446547297)

lgolding:
I need to think about the whole testing issue. We really don't want to run this rule on any test files _except_ for the test files that belong to this rule.

Please revert the changes to the other test files while I think about how to disable rules within tests.

src/Sarif.Multitool/Rules/SARIF2006.UrisShouldBeReachable.cs
lgolding: As we discussed: I now use `UriKind.Absolute` in the call to `IsWellFormedUriString` so it should work.

---
In reply to: [446547509](https://github.com/microsoft/sarif-sdk/pull/1942#discussion_r446547509) [](ancestors = 446547509,446547297)

lgolding:
This all looks good now. I just have to think about the testing issue.

---
In reply to: [650588810](https://github.com/microsoft/sarif-sdk/pull/1942#issuecomment-650588810) [](ancestors = 650588810)

lgolding:
Ok, I know how to do this. I will pull your branch and apply the change. Please DO NOT PUSH ANY FURTHER CHANGES to the branch.

---
In reply to: [650589956](https://github.com/microsoft/sarif-sdk/pull/1942#issuecomment-650589956) [](ancestors = 650589956,650588810)

PR #1943, opened 2020-06-29
adding placeholders for all resource strings and rule ids.

harleenkohli:


src/Sarif.Multitool/Rules/RuleId.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Runtime.CompilerServices;
```
harleenkohli, line 4: this came in by mistake, will remove!

PR #1944, opened 2020-06-30
Adding rule SARIF1012

eddynaka:


src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,92 @@
...
+            nameof(RuleResources.SARIF1012_MessageArgumentsMustBeConsistentWithRule_Error_SupplyEnoughMessageArguments_Text)
+        };
+
+        public override FailureLevel DefaultLevel => FailureLevel.Error;
+
+        private IList<ReportingDescriptor> currentRules;
+
+        protected override void Analyze(Run run, string runPointer)
+        {
+            this.currentRules = run?.Tool?.Driver?.Rules;
```
lgolding, line 33: >?.Tool? [](start = 35, length = 7)

Actually:
- `Analyze` won't be called if `run` is `null`.
- `Tool` is a required property of `Run`, so if it's missing, the schema validation (which takes place before any validation rules are run) will catch it, and the MultiTool will stop.
- Similarly, `Driver` is a required property of `Tool`.
So really all you need is `run.Tool.Driver?.Rules`.

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        protected override void Analyze(Result result, string resultPointer)
+        {
+            // if message.id is not null, we have to check if that exists in rules
+            if (!string.IsNullOrEmpty(result.Message.Id))
+            {
+                ReportingDescriptor rule = result.RuleIndex >= 0
```
lgolding, line 41: >rule [](start = 36, length = 4)

The SARIF SDK has a great helper for this: `result.GetRule(run)`. So:

1. In `Analyze(Run run...)` above, save the `run` in `this.run`.
2. Here, just say

```
ReportingDescriptor rule = result.GetRule(this.run);
```

eddynaka: Ok. With that in mind, we will probably change the following condition to this:
```csharp
if (this.currentRules == null
                    || rule.MessageStrings == null
                    || !rule.MessageStrings.ContainsKey(result.Message.Id))
                {
```

Because we dont initialize MessageStrings in the last scenario. What do you think?

---
In reply to: [447332733](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447332733) [](ancestors = 447332733)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
eddynaka: Will push soon!

---
In reply to: [447331417](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447331417) [](ancestors = 447331417)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,92 @@
...
+                    || !rule.MessageStrings.ContainsKey(result.Message.Id))
+                {
+                    // {0}: Placeholder {1} {2}
+                    LogResult(
+                        resultPointer,
+                        nameof(RuleResources.SARIF1012_MessageArgumentsMustBeConsistentWithRule_Error_MessageIdMustExist_Text),
+                        result.Message.Id,
+                        (result.RuleIndex >= 0
+                        ? result.RuleIndex.ToString()
+                        : (result.RuleId ?? result.Rule?.Id ?? "null")));
```
lgolding, line 56: Here you want to display the rule id, and the SARIF SDK has a helper for that too: `result.ResolveRuleId(run)`. It might return `null`, so you _do_ have to keep the `null` check logic.

eddynaka: will push soon

---
In reply to: [447337251](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447337251) [](ancestors = 447337251)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
lgolding: This looks good. You could simplify to
```
if (this.currentRules == null || rule.MessageStrings?.ContainsKey(result.Message.Id) == true)
```
You need the explicit comparison to `true` because the result of a null coalescing operator is a `Nullable<bool>`, not a `bool`, so you can't do an implicit logical test on it.

---
In reply to: [447336153](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447336153) [](ancestors = 447336153,447332733)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,88 @@
...
+                        resultPointer,
+                        nameof(RuleResources.SARIF1012_MessageArgumentsMustBeConsistentWithRule_Error_MessageIdMustExist_Text),
+                        result.Message.Id,
+                        result.ResolvedRuleId(run) ?? "null");
+                    return;
+                }
+
+                // we have rules, we find it but, the count of parameters are invalid
+                string messageText = rule.MessageStrings[result.Message.Id].Text;
+                int countOfPlaceholders = CountOfPlaceholders(messageText);
```
lgolding, line 60: >CountOfPlaceholders [](start = 42, length = 19)

It's not the _number_ of placeholders that matters. It's the _highest numbered_ placeholder. If the highest numbered placeholder is `{4}` then there must be at least 5 arguments.

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
eddynaka: will push soon

---
In reply to: [447339939](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447339939) [](ancestors = 447339939,447336153,447332733)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1012.MessageArgumentsMustBeConsistentWithRule_Valid.sarif
```diff
@@ -0,0 +1,103 @@
...
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "SARIF Functional Testing",
+          "rules": [
+            {
+              "id": "SARIF2001",
```
lgolding, line 11: >SARIF2001 [](start = 21, length = 9)

Construct the "valid" input file so it gives no errors.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1012.MessageArgumentsMustBeConsistentWithRule_Valid.sarif
```diff
@@ -0,0 +1,101 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "SARIF Functional Testing",
+          "version": "1.2.3",
+          "rules": [
```
lgolding, line 10: >rules [](start = 11, length = 5)

Use simple test rules instead of real ones (`"TST1001"`). And construct the messages so the do not cause "high quality message" errors. For example: `"This is a test with an enquoted placeholder '{0}'."`

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
eddynaka: soon will push the change

---
In reply to: [447340538](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447340538) [](ancestors = 447340538)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,93 @@
...
+                // we have rules, we find it but, the count of parameters are invalid
+                string messageText = rule.MessageStrings[result.Message.Id].Text;
+                int placeholderMaxPosition = PlaceholderMaxPosition(messageText);
+                if (placeholderMaxPosition > (result.Message.Arguments?.Count ?? 0))
+                {
+                    // {0}: Placeholder {1} {2} {3} {4} {5}
+                    LogResult(
+                        resultPointer,
+                        nameof(RuleResources.SARIF1012_MessageArgumentsMustBeConsistentWithRule_Error_SupplyEnoughMessageArguments_Text),
+                        result.Message.Arguments.Count.ToString(),
```
lgolding, line 66: >.ToString() [](start = 54, length = 11)

`ToString()` not necessary.

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        protected override void Analyze(Run run, string runPointer)
+        {
+            this.run = run;
+            this.currentRules = run.Tool.Driver?.Rules;
+        }
+
+        protected override void Analyze(Result result, string resultPointer)
+        {
+            // if message.id is not null, we have to check if that exists in rules
```
lgolding, line 40: >rules [](start = 77, length = 5)

Use full sentences for comments (capitalize first letter, end with period). Also I suggest this phrasing:

`If message.id is present, check that a message with that id exists in the rule.`

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,93 @@
...
+                    // {0}: Placeholder {1} {2}
+                    LogResult(
+                        resultPointer,
+                        nameof(RuleResources.SARIF1012_MessageArgumentsMustBeConsistentWithRule_Error_MessageIdMustExist_Text),
+                        result.Message.Id,
+                        result.ResolvedRuleId(run) ?? "null");
+                    return;
+                }
+
+                // we have rules, we find it but, the count of parameters are invalid
```
lgolding, line 57: >w [](start = 19, length = 1)

"A message with the specified key is present in the rule. Check if the result supplied enough arguments."

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
eddynaka: LogResult only accepts strings.

---
In reply to: [447716571](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447716571) [](ancestors = 447716571)

lgolding: Oh wow. That's not good, ok, thanks.

---
In reply to: [447719968](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447719968) [](ancestors = 447719968,447716571)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,93 @@
...
+                        result.ResolvedRuleId(run) ?? "null",
+                        placeholderMaxPosition.ToString(),
+                        messageText);
+                }
+            }
+        }
+
+        private int PlaceholderMaxPosition(string text)
+        {
+            MatchCollection matchCollection = Regex.Matches(text, "{\\d+}");
```
lgolding, line 77: >"{\\d+}" [](start = 66, length = 8)

You can use a regex "capture" to get just the integer part: `@"\{(<index>\d+)\}"`

lgolding: Also, `'{'` and `}` are regex metacharacters, so you must escape them (as I did above).

---
In reply to: [447722378](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447722378) [](ancestors = 447722378)

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
```diff
@@ -0,0 +1,93 @@
...
+            int max = -1;
+            foreach (Match match in matchCollection)
+            {
+                if (int.TryParse(match.Value.Replace("{", string.Empty).Replace("}", string.Empty), out int temp))
```
lgolding, line 81: >int [](start = 20, length = 3)

Now you know that the captured group is an integer, so:
```
int index = int.Parse(match.Groups["index"].Value);
max = Math.Max(max, index);
```

src/Sarif.Multitool/Rules/SARIF1012.MessageArgumentsMustBeConsistentWithRule.cs
lgolding: Finally, note that "verbatim strings" (`@"..."`) are very useful for regexes because you need lots of backslashes, and with a verbatim string you don't have to double the backslashes.

---
In reply to: [447723010](https://github.com/microsoft/sarif-sdk/pull/1944#discussion_r447723010) [](ancestors = 447723010,447722378)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1012.MessageArgumentsMustBeConsistentWithRule_Valid.sarif
```diff
@@ -0,0 +1,42 @@
...
+            {
+              "id": "TEST1001",
+              "name": "Test 1001",
```
lgolding, line 13: Even `name` is optional :-)

PR #1948, opened 2020-06-30
Adding Rule SARIF2011

eddynaka:


src/Sarif.Multitool/Rules/SARIF2011.ProvideContextRegion.cs
```diff
@@ -0,0 +1,55 @@
...
+                    AnalyzeLocation(result.Locations[i], locationsPointer.AtIndex(i));
+                }
+            }
+        }
+
+        private void AnalyzeLocation(Location location, string locationPointer)
+        {
+            if (location.PhysicalLocation?.Region != null)
+            {
+                string physicalLocation = locationPointer.AtProperty(SarifPropertyName.PhysicalLocation);
```
lgolding, line 44: >string physicalLocation [](start = 16, length = 23)

1. You can put this inside the `if` statement since you don't need it unless the test fails.
2. Call it `physicalLocationPointer` since it's not actually a physical location object.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2011.ProvideContextRegion_Valid.sarif
```diff
@@ -0,0 +1,45 @@
...
+        "driver": {
+          "name": "CodeScanner",
+          "version": "1.0"
+        }
+      },
+      "versionControlProvenance": [
+        {
+          "repositoryUri": "https://github.com/microsoft/sarif-sdk"
+        }
+      ],
```
lgolding, line 16: I see, this necessary because `SARIF2003.ProvideVersionControlProvenance` is"note"-level and so is this rule.

PR #1956, opened 2020-07-01
user msgs verified for 1001 to 1005

harleenkohli:


src/Sarif.Multitool/Rules/SARIF1004.ExpressUriBaseIdsCorrectly.cs
```diff
@@ -33,11 +37,13 @@ public class ExpressUriBaseIdsCorrectly : SarifValidationSkimmerBase
 
         protected override void Analyze(ArtifactLocation fileLocation, string fileLocationPointer)
         {
-            // UriBaseIdRequiresRelativeUri: The 'uri' property of 'fileLocation' must be a relative uri, since 'uriBaseId' is present.
             if (fileLocation.UriBaseId != null && fileLocation.Uri.IsAbsoluteUri)
             {
-                // {0}: This fileLocation object contains a "uriBaseId" property, which means that the value
-                // of the "uri" property must be a relative URI reference, but "{1}" is an absolute URI reference.
+                // TODO: hakohlli - fix parameters.
```
lgolding, line 42: > TODO: hakohlli - fix parameters. [](start = 18, length = 33)

As the code stands, there are not enough arguments in the call to `LogResult` to fill up all the replacement sequences. So why doesn't this crash in the functional test?

harleenkohli: as discussed offline, we are going to change functionality here so that this test is performed only for originaluribaseids.

---
In reply to: [448531084](https://github.com/microsoft/sarif-sdk/pull/1956#discussion_r448531084) [](ancestors = 448531084)

harleenkohli:
>               "message": {

in theory we can remove all these additional nodes since we are not testing for all locations anymore. But these are really good test cases and i really think we should add back the sub-rule some place else. if we plan to do it quickly, then we can re-purpose these, so i left them be.

if we are not planning to add the other rule quickly, then we should get rid of these.

---
Refers to: src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1004.ExpressUriBaseIdsCorrectly_Invalid.sarif:72 in f23fe08. [](commit_id = f23fe0895b77ddee7875df2c5e92d222e050dfe8, deletion_comment = False)

PR #1985, opened 2020-07-06
Remove extra version header from release history.

lgolding:
In preparation for publishing 2.3.1

Also:
- Increase the timeout on a test that's failing on one of our VMs.


michaelcfanning:
>         public void WebRequest_TryParse_HasAcceptablePerformance()

This test is not appropriate for unit testing and should be removed or ignored for now. This sort of validation is fine for a functional test in a controlled environment, not a good idea for a core unit test which runs in arbitrary environments (such as the server where it's currently failing).

---
Refers to: src/Test.UnitTests.Sarif/Core/WebRequestTests.cs:190 in e548393. [](commit_id = e5483939c036d68675c80eb266d075774a69bf7e, deletion_comment = False)

PR #1986, opened 2020-07-09
Guarantee "execute" permissions for NPM package.

### LINKED ISSUE #1963: npx @microsoft/sarif-multitool <args> fails on Linux
Instructions at https://www.npmjs.com/package/@microsoft/sarif-multitool state the following usage instructions:

```
npm i -g @microsoft/sarif-multitool
npx @microsoft/sarif-multitool <args>
```

However on Linux this fails with an error like the following, because the module tries to execute a native executable that doesn't have the executable flag:

```
/bin/sh: 1: /usr/local/lib/node_modules/@microsoft/sarif-multitool/node_modules/@microsoft/sarif-multitool-linux/Sarif.Multitool: Permission denied
```

This behavior was observed on Ubuntu for Windows, and on a GitHub-hosted Ubuntu runner.

A simple `chmod +x ...` fixes this error, however this step is not documented, and preferably the module should take care of this by itself, without requiring users to manually change permissions in some global node modules directory.

jeffersonking:
Closes #1963.

Summary: After installation, we now edit the permissions to the binary to allow "execute". Previously it seemed to depend on how the package was packed.

### DIFF
--- a/npm/sarif-multitool-darwin/package.json
+++ b/npm/sarif-multitool-darwin/package.json
@@ -2,6 +2,9 @@
 	"name": "@microsoft/sarif-multitool-darwin",
 	"description": "SARIF Multitool for MacOS (Darwin)",
 	"version": "{version}",
+	"scripts": {
+		"postinstall": "chmod u+x Sarif.Multitool"
+	}
 	"author": "Microsoft",
 	"license": "MIT",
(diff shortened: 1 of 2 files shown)


lgolding:
Das ist doch ganz klar.

PR #1987, opened 2020-07-10
marking webrequest perf test to be ignored due to its flakiness.

harleenkohli:


src/Test.UnitTests.Sarif/Core/WebRequestTests.cs
```diff
@@ -186,7 +186,7 @@ public void WebRequest_TryParse_ReturnsFalseOnInvalidRequest()
             webRequest.Should().BeNull();
         }
 
-        [Fact]
+        [Fact(Skip="This test is flaky due to the environment specific limitations.")]
```
michaelcfanning, line 189: >="This test is flaky due to the environment specific limitations." [](start = 18, length = 66)

Can you please label this as 'Disabling this test due to inconsistency in timing across various execution environments." I think it captures the reason a bit more precisely.

PR #1988, opened 2020-07-10
Make all Multitool command and options classes public.

lgolding:


src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
     /// configured options are valid, the command will instantiate an instance of 
     /// SarifWorkItemFiler in order to complete the work.
     /// </summary>
-    internal class FileWorkItemsCommand : CommandBase
+    public class FileWorkItemsCommand : CommandBase
```
lgolding, line 28: >public [](start = 4, length = 6)

Are you comfortable exposing this one, or do you think it still needs more private baking?

(Note: its Options class was already public. That's why you don't see it in the PR.)

PR #1995, opened 2020-07-14
Updating SARIF2004

### LINKED ISSUE #1994: [RULE REQUEST] Additional file size optimizations
### Rule metadata
- Id: `SARIF2004` -- The proposal is to add additional checks to this existing rule.
- Name: `OptimizeFileSize`
- Level: `warning`
- Description: _Add the following paragraph to the existing rule description:_

    In addition to the avoiding unnecessary arrays, there are other ways to optimize the size of SARIF log files.

    Prefer the result object properties 'ruleId' and 'ruleIndex' to the nested object-valued property 'result.rule', unless the rule comes from a tool component other than the driver (in which case only 'result.rule' can accurately point to the metadata for the rule). The 'ruleId' and 'ruleIndex' properties are shorter and just as clear.

    Do not specify the result object's 'analysisTarget' property unless it differs from the result location. The canonical scenario for using 'result.analysisTarget' is a C/C++ language analyzer that is instructed to analyze example.c, and detects a result in the included file example.h. In this case, 'analysisTarget' is example.c, and the result location is in example.h.

- User-facing strings:
  
    - `AvoidDuplicativeAnalysisTarget`: {0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.

    - `AvoidDuplicativeResultRuleInformation`: {0}: This result specifies both 'result.ruleId' and 'result.rule'. Prefer 'result.ruleId' because it is shorter and just as clear.

    - `PreferRuleId`: {0}: This result uses the 'rule' property to specify the rule metadata, but the 'ruleId' property suffices because the rule is defined by 'tool.driver'. Prefer 'result.ruleId' because it is shorter and just as clear.

---

### How to resolve
If the relevant rule is defined by tool.driver, then use `result.ruleId` and `result.ruleIndex` instead of the nested object `result.rule`, because the top-level properties make the log file smaller and are just as clear.

### Links/Additional Information
See the SARIF spec [§3.27.13](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317651), "analysisTarget property" for an explanation of when `analysisTarget` is needed.

See the SARIF spec [§3.27.7](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317645), "rule property" for guidance on the use of `ruleId`, `ruleIndex`, and `rule`.

---

eddynaka:
Fixes #1994

### DIFF
--- a/docs/Producing effective SARIF.md
+++ b/docs/Producing effective SARIF.md
@@ -393,13 +393,25 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
 
 #### Messages
 
+##### `AvoidDuplicativeAnalysisTarget`: warning
+
+The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.
+
+#### `AvoidDuplicativeResultRuleInformation`: warning
+
+The result at '{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and 'ruleIndex' properties.
+
 ##### `EliminateLocationOnlyArtifacts`: warning
 
-{0}: The 'artifacts' array contains no information beyond the locations of the artifacts. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.
+The 'artifacts' array at '{0}' contains no information beyond the locations of the artifacts. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.
 
 ##### `EliminateIdOnlyRules`: warning
 
-{0}: The 'rules' array contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.
+The 'rules' array at '{0}' contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.
+
+#### `PreferRuleId`: warning
+
+The result at '{0}' uses the 'rule' property to specify the violated rule, but this is not necessary because the rule is defined by 'tool.driver'. Use the 'ruleId' and 'ruleIndex' instead, because they are shorter and just as clear.
 
 ---
 
--- a/docs/Producing effective SARIF.md
+++ b/docs/Producing effective SARIF.md
@@ -395,11 +395,11 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
 
 ##### `AvoidDuplicativeAnalysisTarget`: warning
 
-The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.
+The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header). 
 
 #### `AvoidDuplicativeResultRuleInformation`: warning
 
-The result at '{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and 'ruleIndex' properties.
+'{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. This unnecessarily increases log file size. Remove the 'ruleId' and 'ruleIndex' properties.
 
 ##### `EliminateLocationOnlyArtifacts`: warning
 
--- a/docs/Producing effective SARIF.md
+++ b/docs/Producing effective SARIF.md
@@ -395,7 +395,7 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
 
 ##### `AvoidDuplicativeAnalysisTarget`: warning
 
-The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header). 
+The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a tool detects a result in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header). 
 
 #### `AvoidDuplicativeResultRuleInformation`: warning
 
--- a/src/Sarif.Multitool/Extensions.cs
+++ b/src/Sarif.Multitool/Extensions.cs
@@ -1,0 +1,24 @@
+﻿using System;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public static class Extensions
+    {
+        public static bool RefersToDriver(this ToolComponentReference toolComponent, string driverGuid)
+        {
+            if (toolComponent.Index == -1)
+            {
+                if (string.IsNullOrWhiteSpace(toolComponent.Guid))
+                {
+                    return true;
+                }
+                else
+                {
+                    return toolComponent.Guid.Equals(driverGuid, StringComparison.OrdinalIgnoreCase);
+                }
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/Sarif.Multitool/Extensions.cs
+++ b/src/Sarif.Multitool/Extensions.cs
@@ -8,7 +8,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
         {
             if (toolComponent.Index == -1)
             {
-                if (string.IsNullOrWhiteSpace(toolComponent.Guid))
+                if (toolComponent.Guid == null)
                 {
                     return true;
                 }
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -281,7 +281,7 @@ Prefer the result object properties 'ruleId' and 'ruleIndex' to the nested objec
 Do not specify the result object's 'analysisTarget' property unless it differs from the result location. The canonical scenario for using 'result.analysisTarget' is a C/C++ language analyzer that is instructed to analyze example.c, and detects a result in the included file example.h. In this case, 'analysisTarget' is example.c, and the result location is in example.h.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_EliminateLocationOnlyArtifacts_Text" xml:space="preserve">
-    <value>{0}: The 'artifacts' array contains no information beyond the locations of the artifacts. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.</value>
+    <value>The 'artifacts' array at '{0}' contains no information beyond the locations of the artifacts. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.</value>
   </data>
   <data name="SARIF2002_ProvideMessageArguments_FullDescription_Text" xml:space="preserve">
     <value>In result messages, use the 'message.id' and 'message.arguments' properties rather than 'message.text'. This has several advantages. If 'text' is lengthy, using 'id' and 'arguments' makes the SARIF file smaller. If the rule metadata is stored externally to the SARIF log file, the message text can be improved (for example, by adding more text, clarifying the phrasing, or fixing typos), and the result messages will pick up the improvements the next time it is displayed. Finally, SARIF supports localizing messages into different languages, which is possible if the SARIF file contains 'message.id' and 'message.arguments', but not if it contains 'message.text' directly.</value>
@@ -296,7 +296,7 @@ Do not specify the result object's 'analysisTarget' property unless it differs f
     <value>{0}: This run does not provide 'versionControlProvenance'. As a result, it is not possible to determine which version of code was analyzed, nor to map relative paths to their locations within the repository.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_EliminateIdOnlyRules_Text" xml:space="preserve">
-    <value>{0}: The 'rules' array contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.</value>
+    <value>The 'rules' array at '{0}' contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.</value>
   </data>
   <data name="SARIF2006_UrisShouldBeReachable_FullDescription_Text" xml:space="preserve">
     <value>URIs that refer to locations such as rule help pages and result-related work items should be reachable via an HTTP GET request.</value>
@@ -367,12 +367,12 @@ This is part of a set of authoring practices that make your rule messages more r
     <value>{0}: This result location does not provide any of the 'uriBaseId' values that specify repository locations: '{1}'. As a result, it will not be possible to determine the location of the file containing this result relative to the root of the repository that contains it.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text" xml:space="preserve">
-    <value>{0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.</value>
+    <value>The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeResultRuleInformation_Text" xml:space="preserve">
-    <value>{0}: This result specifies both 'result.ruleId' and 'result.rule'. Prefer 'result.ruleId' because it is shorter and just as clear.</value>
+    <value>The result at '{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and 'ruleIndex' properties.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_PreferRuleId_Text" xml:space="preserve">
-    <value>{0}: This result uses the 'rule' property to specify the rule metadata, but the 'ruleId' property suffices because the rule is defined by 'tool.driver'. Prefer 'result.ruleId' because it is shorter and just as clear.</value>
+    <value>The result at '{0}' uses the 'rule' property to specify the violated rule, but this is not necessary because the rule is defined by 'tool.driver'. Use the 'ruleId' and 'ruleIndex' instead, because they are shorter and just as clear.</value>
   </data>
 </root>
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -367,10 +367,10 @@ This is part of a set of authoring practices that make your rule messages more r
     <value>{0}: This result location does not provide any of the 'uriBaseId' values that specify repository locations: '{1}'. As a result, it will not be possible to determine the location of the file containing this result relative to the root of the repository that contains it.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text" xml:space="preserve">
-    <value>The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.</value>
+    <value>The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header).</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeResultRuleInformation_Text" xml:space="preserve">
-    <value>The result at '{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and 'ruleIndex' properties.</value>
+    <value>'{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. This unnecessarily increases log file size. Remove the 'ruleId' and 'ruleIndex' properties.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_PreferRuleId_Text" xml:space="preserve">
     <value>The result at '{0}' uses the 'rule' property to specify the violated rule, but this is not necessary because the rule is defined by 'tool.driver'. Use the 'ruleId' and 'ruleIndex' instead, because they are shorter and just as clear.</value>
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -272,7 +272,13 @@ Many tools follow a conventional format for the 'reportingDescriptor.id' propert
 
 In several parts of a SARIF log file, a subset of information about an object appears in one place, and the full information describing all such objects appears in an array elsewhere in the log file. For example, each 'result' object has a 'ruleId' property that identifies the rule that was violated. Elsewhere in the log file, the array 'run.tool.driver.rules' contains additional information about the rules. But if the elements of the 'rules' array contained no information about the rules beyond their ids, then there might be no reason to include the 'rules' array at all, and the log file could be made smaller simply by omitting it. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.
 
-Similarly, most 'result' objects contain at least one 'artifactLocation' object. Elsewhere in the log file, the array 'run.artifacts' contains additional information about the artifacts that were analyzed. But if the elements of the 'artifacts' array contained not information about the artifacts beyond their locations, then there might be no reason to include the 'artifacts' array at all, and again the log file could be made smaller by omitting it. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.</value>
+Similarly, most 'result' objects contain at least one 'artifactLocation' object. Elsewhere in the log file, the array 'run.artifacts' contains additional information about the artifacts that were analyzed. But if the elements of the 'artifacts' array contained not information about the artifacts beyond their locations, then there might be no reason to include the 'artifacts' array at all, and again the log file could be made smaller by omitting it. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.
+
+In addition to the avoiding unnecessary arrays, there are other ways to optimize the size of SARIF log files.
+
+Prefer the result object properties 'ruleId' and 'ruleIndex' to the nested object-valued property 'result.rule', unless the rule comes from a tool component other than the driver (in which case only 'result.rule' can accurately point to the metadata for the rule). The 'ruleId' and 'ruleIndex' properties are shorter and just as clear.
+
+Do not specify the result object's 'analysisTarget' property unless it differs from the result location. The canonical scenario for using 'result.analysisTarget' is a C/C++ language analyzer that is instructed to analyze example.c, and detects a result in the included file example.h. In this case, 'analysisTarget' is example.c, and the result location is in example.h.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_EliminateLocationOnlyArtifacts_Text" xml:space="preserve">
     <value>{0}: The 'artifacts' array contains no information beyond the locations of the artifacts. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'artifacts' array might be used to record the full set of artifacts that were analyzed. In such a scenario, the 'artifacts' array should be retained even if it contains only location information.</value>
@@ -359,5 +365,14 @@ This is part of a set of authoring practices that make your rule messages more r
   </data>
   <data name="SARIF2007_ExpressPathsRelativeToRepoRoot_Warning_ExpressResultLocationsRelativeToMappedTo_Text" xml:space="preserve">
     <value>{0}: This result location does not provide any of the 'uriBaseId' values that specify repository locations: '{1}'. As a result, it will not be possible to determine the location of the file containing this result relative to the root of the repository that contains it.</value>
+  </data>
+  <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text" xml:space="preserve">
+    <value>{0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.</value>
+  </data>
+  <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeResultRuleInformation_Text" xml:space="preserve">
+    <value>{0}: This result specifies both 'result.ruleId' and 'result.rule'. Prefer 'result.ruleId' because it is shorter and just as clear.</value>
+  </data>
+  <data name="SARIF2004_OptimizeFileSize_Warning_PreferRuleId_Text" xml:space="preserve">
+    <value>{0}: This result uses the 'rule' property to specify the rule metadata, but the 'ruleId' property suffices because the rule is defined by 'tool.driver'. Prefer 'result.ruleId' because it is shorter and just as clear.</value>
   </data>
 </root>
--- a/src/Sarif.Multitool/Rules/RuleResources.resx
+++ b/src/Sarif.Multitool/Rules/RuleResources.resx
@@ -367,7 +367,7 @@ This is part of a set of authoring practices that make your rule messages more r
     <value>{0}: This result location does not provide any of the 'uriBaseId' values that specify repository locations: '{1}'. As a result, it will not be possible to determine the location of the file containing this result relative to the root of the repository that contains it.</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text" xml:space="preserve">
-    <value>The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header).</value>
+    <value>The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. The 'analysisTarget' property is used to distinguish cases when a tool detects a result in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header).</value>
   </data>
   <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeResultRuleInformation_Text" xml:space="preserve">
     <value>'{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. This unnecessarily increases log file size. Remove the 'ruleId' and 'ruleIndex' properties.</value>
--- a/src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
+++ b/src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
@@ -67,10 +67,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
 
         public override FailureLevel DefaultLevel => FailureLevel.Warning;
 
+        private string driverGuid;
+
         protected override void Analyze(Run run, string runPointer)
         {
             AnalyzeLocationOnlyArtifacts(run, runPointer);
             AnalyzeIdOnlyRules(run, runPointer);
+
+            this.driverGuid = run.Tool.Driver?.Guid;
         }
 
         protected override void Analyze(Result result, string resultPointer)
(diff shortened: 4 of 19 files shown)


src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -55,6 +71,66 @@ protected override void Analyze(Run run, string runPointer)
...
+                foreach (Location location in result.Locations)
+                {
+                    if (location?.PhysicalLocation?.ArtifactLocation?.Uri == result.AnalysisTarget.Uri)
+                    {
+                        // {0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as
+                        // the result location. Remove the 'analysisTarget' property.
+                        LogResult(
+                            resultPointer.AtProperty(SarifPropertyName.AnalysisTarget),
+                            nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text),
+                            result.AnalysisTarget.Uri.ToString());
```
lgolding, line 130: >ToString() [](start = 54, length = 10)

I'm not sure whether `ToString()` or `OriginalString` is better.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -142,7 +218,6 @@ private void AnalyzeIdOnlyRules(Run run, string runPointer)
                     nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_EliminateIdOnlyRules_Text));
             }
         }
-
         private bool HasIdOnlyRules(string rulePointer)
```
lgolding, line 221: >p [](start = 8, length = 1)

Restore the blank line. (Our coding standard is "blank line after closing brace".)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Valid.sarif
```diff
@@ -42,6 +42,25 @@
               }
             }
           ]
+        },
+        {
+          "ruleIndex": -1,
+          "rule": {
```
lgolding, line 48: Don't specify `ruleIndex` here. The default is -1, so if it it missing, it will be set to -1 when we deserialize the log file.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Valid.sarif
```diff
@@ -42,6 +42,25 @@
...
+        {
+          "ruleId": "TEST1001",
+          "ruleIndex": 0,
+          "locations": [
+            {
+              "physicalLocation": { "artifactLocation": { "uri": "local/test/code/src/file1.h" } }
+            }
+          ],
+          "analysisTarget": { "uri": "local/test/code/src/file1.c" },
+          "message": { "text": "Good result: artifacts array provides additional useful information." }
```
lgolding, line 63: >artifacts array provides additional useful information [](start = 45, length = 54)

The `artifacts` array doesn't contain additional information. All it has is a `location`. (I understand that Line 62 doesn't have `uriBaseId`, and the `artifacts` array does, but still, the intent of the rule is that the `artifacts` array should provide something besides location information. In any case, why are we adding this test? Don't we already have tests for non-useful artifacts arrays?

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1010.RuleIdMustBeConsistent_Valid.sarif
```diff
@@ -47,7 +77,54 @@
           }
         }
       ],
-      "results": [],
+      "results": [
```
lgolding, line 80: >results [](start = 7, length = 7)

The "expected output" from a "valid" test shouldn't have any results. Something must be wrong in the logic.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
eddynaka: Changed to OriginalString

---
In reply to: [454705359](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r454705359) [](ancestors = 454705359)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
eddynaka: Oops

---
In reply to: [454705570](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r454705570) [](ancestors = 454705570)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Valid.sarif
eddynaka: Besides the point that i added a non-absoluteuri, this testing is a good one saying that we have a location.physicallocation.artifactlocation.uri != analysisTarget. The message is wrong and i will solve it

---
In reply to: [454706843](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r454706843) [](ancestors = 454706843)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1010.RuleIdMustBeConsistent_Valid.sarif
eddynaka: The problem here is that the rule 1010 and 2004 overlap in one scenario. Should i remove that scenario (when result.ruleId exists and result.rule.Id exists)

---
In reply to: [454707521](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r454707521) [](ancestors = 454707521)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -55,6 +71,66 @@ protected override void Analyze(Run run, string runPointer)
         {
             AnalyzeLocationOnlyArtifacts(run, runPointer);
             AnalyzeIdOnlyRules(run, runPointer);
+
+            if (run.Results != null)
+            {
+                AnalyzeResults(run.Results, runPointer.AtProperty(SarifPropertyName.Results));
```
lgolding, line 77: >AnalyzeResults [](start = 16, length = 14)

You don't need to write `AnalyzeResults` and call it from here. Just override `Analyze(Result result, string resultPointer)`.

This works because you really want to analyze every result. This is unlike the cases where you only want to analyze _certain_ physical locations, or _certain_ URIs -- and so you have to navigate down from the parent to make sure that you only analyze the ones you care about.

Make sense?

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -55,6 +71,66 @@ protected override void Analyze(Run run, string runPointer)
...
+                foreach (Location location in result.Locations)
+                {
+                    if (location?.PhysicalLocation?.ArtifactLocation?.Uri == result.AnalysisTarget.Uri)
+                    {
+                        // {0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as
+                        // the result location. Remove the 'analysisTarget' property.
+                        LogResult(
+                            resultPointer.AtProperty(SarifPropertyName.AnalysisTarget),
+                            nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text),
+                            result.AnalysisTarget.Uri.OriginalString);
```
lgolding, line 130: >result [](start = 28, length = 6)

I would `break` after logging the first one. If you discover that "analysisTarget isn't necessary because it's the same as location 0", there's no need to also report "analysisTarget isn't necessary because it's the same as location 1".

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Valid.sarif
```diff
@@ -42,6 +42,24 @@
               }
             }
           ]
+        },
+        {
+          "rule": {
+            "id": "TEST1001",
+            "toolComponent": { "name": "Plugin01" }
+          },
+          "message": { "text": "Good result: artifacts array provides additional useful information." }
```
lgolding, line 51: >Good result: artifacts array provides additional useful information [](start = 32, length = 67)

This comment isn't right. The reason this is valid is that `rule.toolComponent` is present.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
```diff
@@ -43,6 +43,36 @@
               }
             }
           ]
+        },
+        {
+          "ruleId": "TEST1001",
+          "ruleIndex": 0,
+          "message": { "text": "Bad result: artifacts array does not provide any additional useful information." },
```
lgolding, line 50: >artifacts array does not provide any additional useful information [](start = 44, length = 66)

Comment is wrong. This result is bad because `analysisTarget` is the same as one of the `locations`.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
```diff
@@ -43,6 +43,36 @@
...
+                "artifactLocation": {
+                  "uri": "local/test/code/src/file1.c"
+                }
+              }
+            }
+          ]
+        },
+        {
+          "ruleIndex": -1,
+          "rule": {
```
lgolding, line 64: Do not include `"ruleIndex": -1`

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
```diff
@@ -43,6 +43,36 @@
...
+            "id": "TEST1001"
+          },
+          "message": { "text": "Bad result: artifacts array does not provide any additional useful information." }
```
lgolding, line 67: >artifacts array does not provide any additional useful information [](start = 44, length = 66)

Comment is wrong: This is invalid because `ruleIndex` and `rule` are both present.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
```diff
@@ -43,6 +43,36 @@
...
+        },
+        {
+          "ruleId": "TEST1001",
+          "ruleIndex": 0,
+          "rule": {
+            "id": "TEST1001"
+          },
+          "message": { "text": "Bad result: artifacts array does not provide any additional useful information." }
```
lgolding, line 75: >Bad [](start = 32, length = 3)

Same here.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1010.RuleIdMustBeConsistent_Valid.sarif
lgolding: No, just disable 2004 for this test.

---
In reply to: [454713672](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r454713672) [](ancestors = 454713672,454707521)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1010.RuleIdMustBeConsistent_Invalid.sarif
```diff
@@ -118,6 +146,29 @@
               }
             }
           ]
+        },
+        {
+          "ruleId": "SARIF2004",
```
lgolding, line 151: >SARIF2004 [](start = 21, length = 9)

Disable 2004 for this test, too.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1009.IndexPropertiesMustBeConsistentWithArrays_Invalid.sarif
```diff
@@ -69,35 +69,6 @@
         }
       ],
       "results": [
-        {
-          "ruleId": "SARIF1009",
```
lgolding, line 73: >SARIF1009 [](start = 21, length = 9)

Why did this result disappear?

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
eddynaka: Yes!

---
In reply to: [455195243](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455195243) [](ancestors = 455195243)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1009.IndexPropertiesMustBeConsistentWithArrays_Invalid.sarif
eddynaka: i rollback my changes and added to ignore the rule 2004 as well.

---
In reply to: [455262015](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455262015) [](ancestors = 455262015)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
lgolding: 1. It is legal to "not have enough information to locate the rule metadata". Rule metadata is entirely optional in SARIF.
2. But in any case, "not enough information to locate the rule metadata" is not one of the cases we are testing for with this new rule. So I don't understand why this test case is here.

---
In reply to: [455257253](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455257253) [](ancestors = 455257253)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -57,6 +73,58 @@ protected override void Analyze(Run run, string runPointer)
...
+                else if (result.Rule.ToolComponent == null)
+                {
+                    // {0}: This result uses the 'rule' property to specify the rule
+                    // metadata, but the 'ruleId' property suffices because the rule
+                    // is defined by 'tool.driver'. Prefer 'result.ruleId' because it
+                    // is shorter and just as clear.
+                    LogResult(
+                        resultPointer,
+                        nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_PreferRuleId_Text));
+                    return;
```
lgolding, line 104: >return [](start = 20, length = 6)

There is no need for either of these return statements. There is no loop here. No matter which branch the code takes, control will pass out of the end of the `if` statement and out of this function. I think you were trying to respond to my comment below...

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
lgolding: This was marked Resolved but it wasn't actually fixed. I think you tried to fix it with the `return` statements above, but that's not what I meant.

---
In reply to: [455252796](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455252796) [](ancestors = 455252796)

eddynaka: i see! i though it was in another place.

---
In reply to: [455344867](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455344867) [](ancestors = 455344867,455252796)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
eddynaka: for this case, we are analyzing the rule where we don't have the ruleIndex AND we don't have the toolComponent. So thats a bad case. With that, i changed to not enough information. Isn't that right?

---
In reply to: [455342702](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455342702) [](ancestors = 455342702,455257253)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
eddynaka: updated the wrong place. removing that

---
In reply to: [455344319](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455344319) [](ancestors = 455344319)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
lgolding: It is legal not to have rule index. I see now that this test case is "You should use `result.ruleId` instead of `result.rule.id`"

---
In reply to: [455382915](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455382915) [](ancestors = 455382915,455342702,455257253)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,94 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+            ReportRuleDuplication(result, resultPointer);
+        }
+
+        private void ReportRuleDuplication(Result result, string resultPointer)
+        {
+            if (result.Rule != null)
+            {
+                if (result.Rule.ToolComponent != null)
+                {
+                    if (result.Rule.ToolComponent.RefersToDriver(this.driverGuid))
```
lgolding, line 92: You can shorten this:
```
if (result.Rule.ToolComponent?.RefersToDriver(this.driverGuid) == true)
```

Note that you need the explicit comparison `== true` because the null-coalescing operator `?.` causes the left-hand side to have the type `Nullable<bool>` -- and you can't do an implicit logical test on a `Nullable<bool>`.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,94 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+                    {
+                        // {0}: This result specifies both 'result.ruleId' and 'result.rule'.
+                        // Prefer 'result.ruleId' because it is shorter and just as clear.
+                        LogResult(
```
lgolding, line 96: >LogResult [](start = 24, length = 9)

This logic is backwards. Take a look at the notes I attached to the chat. The idea is:

If toolComponent is present, and if it does _not_ refer to the driver -- which means that it _has_ to be present! -- then `result.ruleId` and `result.ruleIndex` should not _also_ be present.

Here's what the notes said:
```
  if (result.Rule.ToolComponent != null && !result.Rule.ToolComponent.RefersToDriver(this.currentDriverGuid))
  {
    // We HAD to use result.Rule (because otherwise we couldn't have specified the extension component).
    if either result.ruleId or result.ruleIndex is present
        // Complain, because we should have put them in one place: under result.Rule along with the toolComponent.
        "This result provides the 'rule' property, but it also provides either or both of 'ruleId' and 'ruleIndex'. If 'rule' is present, then use 'rule.id' instead of 'ruleId', and use 'rule.index' instead of 'ruleIndex'."
  }
```

src/Test.UnitTests.Sarif.Multitool/ExtensionsTests.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public class ExtensionsTests
+    {
+        [Theory]
+        [InlineData(-1, "", "", true)]
```
lgolding, line 11: >InlineData [](start = 9, length = 10)

Need a test where `guid` is `null`. That is actually the more likely case -- the SARIF file doesn't even _mention_ `guid`. It is less likely that `guid` is present and its value is the empty string. Keep the "empty string" test, though, by all means.

src/Test.UnitTests.Sarif.Multitool/ExtensionsTests.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public class ExtensionsTests
+    {
+        [Theory]
```
lgolding, line 10: >Theory [](start = 9, length = 6)

We don't use `Theory` in this code base because we had some bad perf experience with it some years ago. It might be much better now. We use another technique for data-driven tests. I'll go find an example and add it to this comment. Stand by...

lgolding: For an example of our pattern, see src\Test.UnitTests.Sarif.WorkItems\SarifWorkItemExtensionsTests.cs, Lines 118-186.

---
In reply to: [455431411](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455431411) [](ancestors = 455431411)

lgolding: The set of test cases is fine. Please use the pattern in src\Test.UnitTests.Sarif.WorkItems\SarifWorkItemExtensionsTests.cs, Lines 118-186. This pattern accumulates failures across all test cases, and then displays them at the end. Also, use FluentAssertions.

---
In reply to: [455441780](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455441780) [](ancestors = 455441780,455431411)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,107 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+                            resultPointer,
+                            nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_PreferRuleId_Text));
+                    }
+                    else
+                    {
+                        if (!string.IsNullOrWhiteSpace(result.RuleId) || result.RuleIndex >= 0)
+                        {
+                            // {0}: This result specifies both 'result.ruleId' and 'result.rule'.
+                            // Prefer 'result.ruleId' because it is shorter and just as clear.
```
lgolding, line 107: I will provide new strings.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
```diff
@@ -43,6 +44,50 @@
...
+        },
+        {
+          "rule": {
+            "id": "TEST1001",
+            "toolComponent": {
+              "name": "toolComponent",
+              "guid": "686D3462-EC4F-499F-8738-9D6ED96F7A27"
+            }
+          },
+          "message": { "text": "Bad result: rule could be simplified, because it's pointing to default tool." }
```
lgolding, line 71: >default tool [](start = 95, length = 12)

"default tool" => "tool.driver"

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2004.OptimizeFileSize_Invalid.sarif
lgolding: This comment still doesn't describe this test case. In this test case, `toolComponent` does _not_ refer to the driver, so `result.rule` is necessary. But since `result.rule` is present, `result.ruleId` should _not_ be present. So the message might be "Bad result: ruleId duplicates information in rule."

---
In reply to: [455389470](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455389470) [](ancestors = 455389470,455382915,455342702,455257253)

src/Sarif.Multitool/Extensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public static class Extensions
+    {
+        public static bool RefersToDriver(this ToolComponentReference toolComponent, string driverGuid)
+        {
+            if (toolComponent.Index == -1)
+            {
+                if (string.IsNullOrWhiteSpace(toolComponent.Guid))
```
lgolding, line 11: >IsNullOrWhiteSpace [](start = 27, length = 18)

Just `null`. We are checking if `toolComponent.guid` is _absent_ from the SARIF file. If it's absent from the SARIF file, `toolComponent.guid` will be `null`.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,107 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+            if (result.Rule != null)
+            {
+                if (result.Rule.ToolComponent != null)
+                {
+                    if (result.Rule.ToolComponent.RefersToDriver(this.driverGuid))
+                    {
+                        // {0}: This result uses the 'rule' property to specify the rule
+                        // metadata, but the 'ruleId' property suffices because the rule
+                        // is defined by 'tool.driver'. Prefer 'result.ruleId' because it
+                        // is shorter and just as clear.
```
lgolding, line 97: New message:

The result at `{0}` uses the 'rule' property to specify the rule metadata, but the 'ruleId' property suffices because the rule is defined by 'tool.driver'. Prefer 'result.ruleId' because it is shorter and just as clear.

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
lgolding: New message:

The result at '{0}' uses the 'rule' property to specify the metadata for a rule defined by a tool extension, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and/or 'ruleIndex' properties.

---
In reply to: [455809589](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455809589) [](ancestors = 455809589)

src/Sarif.Multitool/Extensions.cs
eddynaka: ok. i will remove the test cases where we test for "", because that won't exist

---
In reply to: [455816703](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455816703) [](ancestors = 455816703)

src/Test.UnitTests.Sarif.Multitool/ExtensionsTests.cs
lgolding: Ok, we are now defining the test data according to the pattern in src\Test.UnitTests.Sarif.WorkItems\SarifWorkItemExtensionsTests.cs, Lines 118-186. We still need to do the following:

- Accumulate the failing test cases. See how `SarifWorkItemExtensions_PhraseToolNames_ConstructPhraseCorrectly` accumulates a list of the failed test cases in a `StringBuilder`.
- Use FluentAssertions. See how that test method uses `sb.Length.Should().Be(0, "... message listing the failed test cases ...")`

---
In reply to: [455806704](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455806704) [](ancestors = 455806704,455441780,455431411)

lgolding: The reason for accumulating the test cases is that you can report all the failures at once, instead of stopping at the first failure.

---
In reply to: [455835254](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455835254) [](ancestors = 455835254,455806704,455441780,455431411)

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,107 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+                                nameof(RuleResources.SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeResultRuleInformation_Text));
+                        }
+                    }                    
+                }
+                else
+                {
+                    // {0}: This result uses the 'rule' property to specify the rule
+                    // metadata, but the 'ruleId' property suffices because the rule
+                    // is defined by 'tool.driver'. Prefer 'result.ruleId' because it
+                    // is shorter and just as clear.
```
lgolding, line 119: Same message as on Line 94 above ("The result at '{0}' uses the 'rule' property to specify the violated rule, but this is not necessary because ...").

src/Sarif.Multitool/Rules/RuleResources.resx
```diff
@@ -360,4 +366,13 @@ This is part of a set of authoring practices that make your rule messages more r
   <data name="SARIF2007_ExpressPathsRelativeToRepoRoot_Warning_ExpressResultLocationsRelativeToMappedTo_Text" xml:space="preserve">
     <value>{0}: This result location does not provide any of the 'uriBaseId' values that specify repository locations: '{1}'. As a result, it will not be possible to determine the location of the file containing this result relative to the root of the repository that contains it.</value>
   </data>
+  <data name="SARIF2004_OptimizeFileSize_Warning_AvoidDuplicativeAnalysisTarget_Text" xml:space="preserve">
+    <value>{0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.</value>
```
lgolding, line 370: >{0}:  [](start = 11, length = 5)

"In the result at '{0}', the 'analysisTarget' property '{1}' is unnecessary..."

src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs
```diff
@@ -41,20 +41,107 @@ public class OptimizeFileSize : SarifValidationSkimmerBase
...
+        private void ReportUnnecessaryAnalysisTarget(Result result, string resultPointer)
+        {
+            if (result.Locations != null && result.AnalysisTarget != null)
+            {
+                foreach (Location location in result.Locations)
+                {
+                    if (location?.PhysicalLocation?.ArtifactLocation?.Uri == result.AnalysisTarget.Uri)
+                    {
+                        // {0}: The 'analysisTarget' property '{1}' is unnecessary because it is the same as
+                        // the result location. Remove the 'analysisTarget' property.
```
lgolding, line 136: >the [](start = 27, length = 3)

The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.

lgolding:
>                     firstArtifactPointer,

I want the new message to point to the `artifacts` array, not to the first element of the array. So define `artifactPointer = runPointer.AtProperty(SarifPropertyName.Artifacts)` and use it here. I'll provide the new message shortly.

---
Refers to: src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs:178 in 84bde22. [](commit_id = 84bde2234d6438347eca8f9f3aaff6d0cd89b855, deletion_comment = False)

lgolding:
>                     firstArtifactPointer,

Here you go:

The 'artifacts' array at '{0}' contains no information..."

---
In reply to: [659480012](https://github.com/microsoft/sarif-sdk/pull/1995#issuecomment-659480012) [](ancestors = 659480012)

---
Refers to: src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs:178 in 84bde22. [](commit_id = 84bde2234d6438347eca8f9f3aaff6d0cd89b855, deletion_comment = False)

lgolding:
>             if (HasIdOnlyRules(firstRulePointer))

Similarly here: define `rulesPointer` to point to the array (and calculate `firstRulePointer` from `rulesPointer`, and use `rulesPointer` in the new message:

"The 'rules' array at '{0}' contains no information..."

---
Refers to: src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs:219 in 84bde22. [](commit_id = 84bde2234d6438347eca8f9f3aaff6d0cd89b855, deletion_comment = False)

lgolding:
>             // null check in not required.

"in" =>"is"

---
Refers to: src/Sarif.Multitool/Rules/SARIF2004.OptimizeFileSize.cs:205 in 84bde22. [](commit_id = 84bde2234d6438347eca8f9f3aaff6d0cd89b855, deletion_comment = False)

src/Test.UnitTests.Sarif.Multitool/ExtensionsTests.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
```
lgolding, line 4: >FluentAssertions [](start = 6, length = 16)

`System` usings come first.

docs/Producing effective SARIF.md
```diff
@@ -393,13 +393,25 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
 
 #### Messages
 
+##### `AvoidDuplicativeAnalysisTarget`: warning
```
michaelcfanning, line 396: We should have code that auto-emits this content, rather than maintaining it in three places (resource strings, as comments in rule source, and here). Action item for later.

docs/Producing effective SARIF.md
```diff
@@ -393,13 +393,25 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
...
+
+The 'analysisTarget' property '{1}' at '{0}' is unnecessary because it is the same as the result location. Remove the 'analysisTarget' property.
```
michaelcfanning, line 398: >unnecessary [](start = 48, length = 11)

Please add something like: 'The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header). This C++ example example is the canonical scenario that led to this design pattern. 

Also, though it is implied, the reason we recommend removing it is to optimize file size.

The 'analysisTarget' property '{1}' at '{0}' can be removed because it is the same as the result location. This unnecessarily increases log file size. 'The 'analysisTarget' property is used to distinguish cases when a result fires in a file (such as an included header) that is different than the file that was scanned (such as a .cpp file that included the header).

docs/Producing effective SARIF.md
```diff
@@ -393,13 +393,25 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
...
+
+#### `AvoidDuplicativeResultRuleInformation`: warning
+
+The result at '{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. Remove the 'ruleId' and 'ruleIndex' properties.
```
michaelcfanning, line 402: You have a tendency to introduce descriptive prefixes in rules that I find distracting. The '{0}' value tends to be sufficiently descriptive to illuminate what's under discussion. This is most easily seen when actually reviewing output of the tool (which I encourage you to do), particularly real world quantities of output, not just a small # of test messages. In this second example. we see that we're trying to evolve to a common output format string for duplicative information (and we should converge on a standard).

'{0}' uses the 'rule' property to specify the violated rule, so it is not necessary also to specify 'ruleId' or 'ruleIndex'. This unnecessarily increases log file size. Remove the 'ruleId' and 'ruleIndex' properties.

docs/Producing effective SARIF.md
lgolding: @eddynaka, +1 on Michael's rephrasing (the very last paragraph in his comment above).

---
In reply to: [455904829](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455904829) [](ancestors = 455904829)

lgolding: ... except that instead of "when a result fires" I suggest "when a tool detects a result".

---
In reply to: [455907085](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455907085) [](ancestors = 455907085,455904829)

docs/Producing effective SARIF.md
```diff
@@ -393,13 +393,25 @@ Similarly, most 'result' objects contain at least one 'artifactLocation' object.
...
 
 ##### `EliminateIdOnlyRules`: warning
 
-{0}: The 'rules' array contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.
+The 'rules' array at '{0}' contains no information beyond the ids of the rules. Removing this array might reduce the log file size without losing information. In some scenarios (for example, when assessing compliance with policy), the 'rules' array might be used to record the full set of rules that were evaluated. In such a scenario, the 'rules' array should be retained even if it contains only id information.
```
michaelcfanning, line 410: >The 'rules' array [](start = 0, length = 17)

I'd like to see actual instances of this rule output before approving. Can you provide some from your testing?

docs/Producing effective SARIF.md
michaelcfanning: I didn't intend to take over reauthoring the words, just point you in the right direction. :) feel free to send Eddy updates after it goes through your wordsmithing process. Every string needs a pass in my estimation. I need the team to send me the logs you generated in testing so that I can see what your new output looks like in bulk. If we aren't yet doing that testing, it needs to be included in the formalized process.

---
In reply to: [455907934](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455907934) [](ancestors = 455907934,455907085,455904829)

docs/Producing effective SARIF.md
lgolding: I've just sent you the output from running the validator on a SARIF file produced from scanning a significant repo.

---
In reply to: [455908031](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455908031) [](ancestors = 455908031)

docs/Producing effective SARIF.md
lgolding: I'm open to this but I ask you to consider the potential value of the descriptive prefixes. The problem is that the JSON paths can be very long, and without the prefix, the first thing that meets your eye is this long string. Consider an example where it's _really_ long:

```
'/runs/0/results/0/locations/0/physicalLocation/artifactLocation/uri' contains an invalid URI 'fi:le:///C:/dev/'
```

I think this provides the reader with an easier entry to the message:

```
The URI 'fi:le:///C:/dev/' at '/runs/0/results/0/locations/0/physicalLocation/artifactLocation/uri' is invalid.
```

... because it says "I'm about to tell you something about a URI". Without the prefix, you have to scan to the end of the path string to learn that it's about a URI.

---
In reply to: [455907068](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455907068) [](ancestors = 455907068)

michaelcfanning: Your example doesn't correspond to my feedback, which notes that describing the SARIF JSON type is generally not required because the type information tends to be encoded in the JSON path. The term 'URI' isn't typically encoded in the URI and so it is useful to identify the string literal as one.

---
In reply to: [455936290](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455936290) [](ancestors = 455936290,455907068)

lgolding: Ok, I understand the distinction.

---
In reply to: [455941928](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455941928) [](ancestors = 455941928,455936290,455907068)

eddynaka: Just updated

---
In reply to: [455982124](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455982124) [](ancestors = 455982124,455941928,455936290,455907068)

docs/Producing effective SARIF.md
eddynaka: just updated this as well!

---
In reply to: [455913218](https://github.com/microsoft/sarif-sdk/pull/1995#discussion_r455913218) [](ancestors = 455913218,455907934,455907085,455904829)

PR #1996, opened 2020-07-15
Adding rule SARIF2016

### LINKED ISSUE #1993: [RULE REQUEST] Avoid absolute file URIs in artifact locations
### Rule metadata
- Id: `SARIF2016`
- Name: `FileUrisShouldBeRelative`
- Level: `note`
- Description:

    When an artifact location refers to a file on the local file system, specify a relative reference for the uri property and provide a uriBaseId property, rather than specifying an absolute URI.

    There are several advantages to this approach:

    Portability: A log file that contains relative references together with uriBaseI properties can be interpreted on a machine where the files are located at a different absolute location.

    Determinism: A log file that uses uriBaseId properties has a better chance of being “deterministic”; that is, of being identical from run to run if none of its inputs have changed, even if those runs occur on machines where the files are located at different absolute locations.

    Security: The use of uriBaseId properties avoids the persistence of absolute path names in the log file. Absolute path names can reveal information that might be sensitive.

    Semantics: Assuming the reader of the log file (an end user or another tool) has the necessary context, they can understand the meaning of the location specified by the uri property, for example, “this is a source file”.

- User-facing strings:

    - `Default`: {0}: The file location '{1}' is specified with absolute URI. Prefer a relative reference together with a uriBaseId property.

---

### Links/Additional Information
See the SARIF Specification, [§3.4.4](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317431): uriBaseId property

### Implementation Notes
The analyzer should examine all URI-valued properties except for those the spec requires to be absolute. The arguments in favor of relative references and uriBaseIds apply equally well to result locations, related locations, code flow locations, stack frame locations, and so on.

### How to resolve
The simplest approach is for the tool to assign a uriBaseId symbol (we recommend 'SCAN_ROOT') to the directory the tool was instructed to scan, and express all file URIs relative to SCAN_ROOT.

---

eddynaka:
Fixes #1993

## Changes:
- adding rule sarif2016
- updated docs with the new rule

### DIFF
--- a/src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
+++ b/src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
@@ -205,6 +205,7 @@ namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests.Multitool
         public void SARIF2007_ExpressPathsRelativeToRepoRoot_Valid()
             => RunTest(MakeValidTestFileName(RuleId.ExpressPathsRelativeToRepoRoot, nameof(RuleId.ExpressPathsRelativeToRepoRoot)),
                 parameter: new TestParameters(configFileName: "enable2007.configuration.xml"));
+
         [Fact]
         public void SARIF2007_ExpressPathsRelativeToRepoRoot_WithoutVersionControlProvenance_Valid()
             => RunTest("SARIF2007.ExpressPathsRelativeToRepoRoot_WithoutVersionControlProvenance_Valid.sarif",
@@ -302,6 +303,16 @@ namespace Microsoft.CodeAnalysis.Sarif.FunctionalTests.Multitool
         public void SARIF2015_EnquoteDynamicMessageContent_Invalid()
             => RunTest(MakeInvalidTestFileName(RuleId.EnquoteDynamicMessageContent, nameof(RuleId.EnquoteDynamicMessageContent)),
                 parameter: new TestParameters(verbose: true));
+
+        [Fact]
+        public void SARIF2016_FileUrisShouldBeRelative_Valid()
+            => RunTest(MakeValidTestFileName(RuleId.FileUrisShouldBeRelative, nameof(RuleId.FileUrisShouldBeRelative)),
+                parameter: new TestParameters(verbose: true, configFileName: "enable2016.configuration.xml"));
+
+        [Fact]
+        public void SARIF2016_FileUrisShouldBeRelative_Invalid()
+            => RunTest(MakeInvalidTestFileName(RuleId.FileUrisShouldBeRelative, nameof(RuleId.FileUrisShouldBeRelative)),
+                parameter: new TestParameters(verbose: true, configFileName: "enable2016.configuration.xml"));
 
         private const string ValidTestFileNameSuffix = "_Valid.sarif";
         private const string InvalidTestFileNameSuffix = "_Invalid.sarif";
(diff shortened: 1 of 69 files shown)


michaelcfanning:
>   <Properties Key="SARIF2002.ProvideMessageArguments.Options">

I don't like these checked in files that appear to drive specific tests. Instead, we should have some helper that autogenerates the configuration based on a set of desired enabled/disabled rules. 

---
Refers to: src/Test.FunctionalTests.Sarif/default.configuration.xml:6 in 9dd2f98. [](commit_id = 9dd2f9826dcc6fddb711a7171cc4a3a17b9b9d9b, deletion_comment = False)

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -306,6 +307,16 @@ public void SARIF2015_EnquoteDynamicMessageContent_Invalid()
             => RunTest(MakeInvalidTestFileName(RuleId.EnquoteDynamicMessageContent, nameof(RuleId.EnquoteDynamicMessageContent)),
                 parameter: new TestParameters(verbose: true));
 
+        [Fact]
+        public void SARIF2016_FileUrisShouldBeRelative_Valid()
+            => RunTest(MakeValidTestFileName(RuleId.FileUrisShouldBeRelative, nameof(RuleId.FileUrisShouldBeRelative)),
+                parameter: new TestParameters(verbose: true, configFileName: "enable2016.configuration.xml"));
```
michaelcfanning, line 313: >enable2016 [](start = 78, length = 10)

If the point is simply to enable a specific rule, we should have a helper that constructs the appropriate config using a rule id. And dispense with maintaining a checked in file.

src/Test.FunctionalTests.Sarif/default.configuration.xml
```diff
@@ -12,4 +12,7 @@ This file contains the default rule configurations for the functional tests of t
   <Properties Key="SARIF2007.ExpressPathsRelativeToRepoRoot.Options">
     <Property Key="RuleEnabled" Value="Disabled" />
   </Properties>
+  <Properties Key="SARIF2016.FileUrisShouldBeRelative.Options">
+    <Property Key="RuleEnabled" Value="Disabled" />
```
michaelcfanning, line 16: >Disabled [](start = 39, length = 8)

Why is this rule disabled by default? I am beginning to think we've made a mistake: these rules should be notes, enabled by default. The validation tool should not show them unless appropriate verbosity is set.

lgolding:
Eddy, I will show you how to do some "functional-level" testing on real SARIF (in addition to the existing functional tests).

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1011.ReferenceFinalSchema_Invalid.sarif
```diff
@@ -52,6 +52,14 @@
               "descriptor": {
                 "id": "WRN999.RuleExplicitlyDisabled"
               }
+            },
+            {
+              "message": {
+                "text": "Rule 'SARIF2016' was explicitly disabled by the user. As result, this tool run cannot be used for compliance or other auditing processes that require a comprehensive analysis."
```
michaelcfanning, line 58: >SARIF2016 [](start = 31, length = 9)

Here's a clue we're off track: our current approach requires us to emit an explicit change to a large set of test files to accommodate a new rule. Changing these rules to 'note' and requiring an explicit opt-in argument to see 'note' messages will eliminate this requirement.

eddynaka:
> Eddy, I will show you how to do some "functional-level" testing on real SARIF (in addition to the existing functional tests).

ok. thank you

PR #2013, opened 2020-07-28
Fix #2009: Don't require per-test config files in validator functional tests.

### LINKED ISSUE #2009: Address PR tail in implementation of analysis rule SARIF2016.

lgolding:
This PR is in preparation for adding new validation rules related to the GitHub Developer Security Portal (DSP). It addresses tail work identified in PR #1996: "Adding rule SARIF2016".

In the validator functional tests, we run the validator, which runs all the validation rules, over every functional test file. As we add new rules, we potentially introduce new violations into existing test files, breaking the comparison between the expected and actual outputs of the validation.

Until now we've addressed this problem either by updating the existing test files to accommodate the new rules or by introducing per-rule configuration files that disable those rules that introduce new violations. Neither of these approaches is scalable.

In the comments to #1996 @michaelcfanning suggested a helper method to automatically create the configuration file for each test. In this PR we take a different approach: we post-process the log file that the validator produces, removing all results that don't come from the rule under test.

Also:
- Remove the suppressions of rules `SARIF2002` and `SARIF2006` from the default configuration file that ships with the validator.
- Demote rules `SARIF2002` and `SARIF2006` from `"warning"` to `"note"`.


src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -389,16 +340,26 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
...
+                ?.Where(r => IsRelevant(r.Id, ruleUnderTest))
+                ?.ToList();
+
+            // Since there's only one rule in the metadata, the ruleIndex for all remaining results
+            // must be 0.
+            foreach (Result result in run.Results)
             {
-                if (result.RuleId == ruleUnderTest)
-                {
-                    newResults.Add(result);
```
lgolding, line 400: >newResults [](start = 20, length = 10)

There already _was_ logic to remove results from other rules -- but it was never hooked up (the `newResults` list wasn't assigned back to the log file). In this PR, we hook it up, and we make additional necessary adjustments to the results and the rules.

lgolding:

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -53,278 +45,242 @@ public void JSON1002_DeserializationError()
...
-                MakeValidTestFileName(RuleId.ConsiderConventionalIdentifierValues, nameof(RuleId.ConsiderConventionalIdentifierValues)),
-                parameter: new TestParameters(verbose: true));
+            => RunValidTestForRule(RuleId.ConsiderConventionalIdentifierValues);
 
         [Fact]
         public void SARIF2009_ConsiderConventionalIdentifierValues_Invalid()
-            => RunTest(
-                MakeInvalidTestFileName(RuleId.ConsiderConventionalIdentifierValues, nameof(RuleId.ConsiderConventionalIdentifierValues)),
-                parameter: new TestParameters(verbose: true));
+            => RunInvalidTestForRule(RuleId.ConsiderConventionalIdentifierValues);
```
michaelcfanning, line 208: >RunInvalidTestForRule [](start = 15, length = 21)

Simplified! yay. :)

PR #2019, opened 2020-07-29
Fix #2003: Don't report missing snippet if file content is present.

### LINKED ISSUE #2003: Refinement to SARIF2010.ProvideCodeSnippets: Don't report if embedded content is present
@michaelcfanning What's your opinion on this? One could argue that the snippets should be present anyway, so in case the user strips out the embedded content, the snippets will remain behind.

eddynaka:
Fixes #2003


src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -215,6 +215,10 @@ public void SARIF2010_ProvideCodeSnippets_Valid()
         public void SARIF2010_ProvideCodeSnippets_Invalid()
             => RunInvalidTestForRule(RuleId.ProvideCodeSnippets);
 
+        [Fact]
+        public void SARIF2010_ProvideCodeSnippets_WithArtifacts_Valid()
+            => RunTest("SARIF2010.ProvideCodeSnippets_WithArtifacts_Valid.sarif");
```
lgolding, line 220: >WithArtifacts [](start = 54, length = 13)

Let's move this just below the _other_ "valid" test for this rule. Also, it's not the "with artifacts" that's important here, it's that the artifacts have embedded file content. So perhaps `"SARIF2010.ProvideCodeSnippets_WithEmbeddedContent.sarif"`.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2010.ProvideCodeSnippets_WithArtifacts_Valid.sarif
```diff
@@ -0,0 +1,78 @@
...
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "CodeScanner",
+          "version": "1.0"
+        }
+      },
+      "versionControlProvenance": [
```
lgolding, line 12: >versionControlProvenance [](start = 7, length = 24)

FYI: with the changes I merged this morning, it is no longer necessary for every test file to obey every analysis rule -- because the test framework now only looks at results that come from the rule you are testing. So, for example, you don't need to supply `versionControlProvenance`.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2010.ProvideCodeSnippets_WithArtifacts_Valid.sarif
```diff
@@ -0,0 +1,78 @@
...
+          "message": {
+            "text": "Some testing occurred."
+          },
+          "locations": [
+            {
+              "physicalLocation": {
+                "artifactLocation": {
+                  "uri": "src/test.c"
+                },
+                "contextRegion": {
```
lgolding, line 54: >contextRegion [](start = 17, length = 13)

Similarly, you don't need `contextRegion` any more. That's the nice thing about this morning's changes: not only do we not have to keep the test files up to date with new rules, but also, the test files can be smaller and focused on exactly the rule they are testing.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,46 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+
+        private bool AnalyzeArtifactLocation(ArtifactLocation artifactLocation)
+        {
+            // no artifactLocation / no artifacts, so we should look for the snippet
+            if (artifactLocation == null || this.artifacts == null)
+            {
+                return true;
+            }
+
+            artifactLocation.TryReconstructAbsoluteUri(this.originalUriBaseIds, out Uri resolvedUri);
```
lgolding, line 90: >TryReconstructAbsoluteUri [](start = 29, length = 25)

`TryReconstructAbsoluteUri` can return `false`, in which case `resolvedUri` will be null, and you'll need to use `artifactLocation.Uri` instead. So, something like this:
```
Uri resolvedUri;
if (!artifactLocation.TryReconstructAbsoluteUri(this.originalUriBaseIds, out resolvedUri))
{
    resolvedUri = artifactLocation.Uri;
}
```

Please look at `TryReconstructAbsoluteUri` to see how it can fail; then provide a test case that exercises this logic.

lgolding: Come to think of it:

```
if (!artifactLocation.TryReconstructAbsoluteUri(this.originalUriBaseIds, out Uri resolvedUri))
{
    return true;
}
```

If a SARIF viewer can't resolve the URI, it won't be able to find the artifact, so it should insist on the snippet being present.

We still need a test case for this, thought.

---
In reply to: [462536376](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462536376) [](ancestors = 462536376)

lgolding, line 90: >TryReconstructAbsoluteUri [](start = 29, length = 25)

If a SARIF viewer can't resolve the URI, it won't be able to find the embedded content, and so it will need the snippet to be present. So:
```
if (!artifactLocation.TryReconstructAbsoluteUri(this.originalUriBaseIds, out Uri resolvedUri))
{
    return false;
}
```

We need a test case for this, so please take a look at `TryReconstructAbsoluteUri` to see how you can make it fail. Probably using an unknown `uriBaseId` will do it.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,46 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+
+            foreach (Artifact artifact in this.artifacts)
+            {
+                // content/text doesn't exist, continue to next
+                if (string.IsNullOrEmpty(artifact.Contents?.Text))
+                {
+                    continue;
+                }
+
+                artifact.Location.TryReconstructAbsoluteUri(this.originalUriBaseIds, out Uri artifactUri);
```
lgolding, line 100: >TryReconstructAbsoluteUri [](start = 34, length = 25)

```
if (!artifact.Location.TryReconstructAbsoluteUri(this.originalUriBaseIds, out Uri artifactUri)
{
    continue;
}
```
Need a test case for this too (an artifact that _does_ have content but that _cannot_ be resolved.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,46 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+
+                if (resolvedUri != null && artifactUri != null)
```
lgolding, line 102: >if (resolvedUri != null && artifactUri != null) [](start = 16, length = 47)

Since you will have previously checked the return values from the two calls to `TryReconstructAbsoluteUri`, you won't need this test.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,46 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+                {
+                    if (artifactUri.AbsolutePath.Equals(resolvedUri.AbsolutePath, StringComparison.OrdinalIgnoreCase))
```
lgolding, line 104: >OrdinalIgnoreCase [](start = 99, length = 17)

You can't ignore case in some file systems. On the other hand, you have to ignore case in Windows. We are aware of this general problem in the SARIF SDK.

Let's remove this for now.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,46 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    // couldn't generate the absoluteUri's, so let's compare everything
```
lgolding, line 111: >c [](start = 23, length = 1)

A small thing: Please begin each comment sentence with a capital letter, and end with a period. Thanks.

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
eddynaka: yes and no...if we couldn't resolve doesn't mean that we need to look for snippet. Imagine the situation where we don't have the uriBaseId but we have the file. That file can be presented in the artifacts list, right?

---
In reply to: [462538548](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462538548) [](ancestors = 462538548)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
eddynaka: Same thing as before, what if we don't have the uribaseid but the path is right, so we have to continue validating. Right?

---
In reply to: [462543051](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462543051) [](ancestors = 462543051)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
lgolding: You're right. Please add a comment, because people reading the code will be surprised (as I was) that you call a `Try` API without checking for success.

---
In reply to: [462546597](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462546597) [](ancestors = 462546597,462538548)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
lgolding: By the argument you made above, we _do_ need these tests.

---
In reply to: [462543518](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462543518) [](ancestors = 462543518)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
lgolding: You are right. Please do add the test case.

---
In reply to: [462547813](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462547813) [](ancestors = 462547813,462543051)

eddynaka: I added some comments in the sarif, so we can understand what i'm validating. Let me know what do you think

---
In reply to: [462576825](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462576825) [](ancestors = 462576825,462547813,462543051)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
eddynaka: Added some comments, let me know what do you think

---
In reply to: [462575560](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462575560) [](ancestors = 462575560,462546597,462538548)

lgolding: Good. Wording: "... since neither originalUriBaseIds nor artifactLocation.UriBaseId is required."

---
In reply to: [462582648](https://github.com/microsoft/sarif-sdk/pull/2019#discussion_r462582648) [](ancestors = 462582648,462575560,462546597,462538548)

src/Sarif.Multitool/Rules/SARIF2010.ProvideCodeSnippets.cs
```diff
@@ -65,5 +78,52 @@ private void AnalyzeRegion(Region region, string regionPointer)
...
+                if (resolvedUri != null && artifactUri != null)
+                {
+                    if (artifactUri.AbsolutePath.Equals(resolvedUri.AbsolutePath))
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    // Couldn't generate the absoluteUri's, so let's compare everything.
```
lgolding, line 117: >' [](start = 56, length = 1)

No apostrophe.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2010.ProvideCodeSnippets_WithEmbeddedContent.sarif
```diff
@@ -0,0 +1,69 @@
...
+          "level": "error",
+          "message": {
+            "text": "Some testing occurred."
+          },
+          "locations": [
+            {
+              "physicalLocation": {
+                "artifactLocation": {
+                  // this test will NOT generate a valid AbsoluteUri, but it will find in artifacts
+                  // So, it don't need a snippet
```
lgolding, line 48: Comments not valid in a SARIF file. You can put a comment in the property bag, for example:
```
"physicalLocation": {
  "artifactLocation": {
    "uri": "src/test.c"
  },
  "properties": {
    "comment": "This test will NOT generate a valid AbsoluteUri, but it will find find the file in the artifacts array. So it doesn't need a snippet."
  }
}
```

PR #2021, opened 2020-07-30
Introduce GitHub DSP analysis rules

lgolding:
We implement:

- `SARIF2017.LocationsMustProvideRequiredProperties`
- `SARIF2018.InlineThreadFlowLocations`
- `SARIF2019.RegionsMustProvideRequiredProperties`
- `SARIF2020.ReviewArraysThatExceedConfigurableDefaults`
- `SARIF2021.LocationsMustBeRelativeUrisOrFilePaths`
- `SARIF2022.ProvideCheckoutPath`
- `SARIF2023.RelatedLocationsMustProvideRequiredProperties`

These rules are all error-level and are all disabled by default.

Since this is the first time we've had to support rules that are disabled by default, some changes to the Driver framework and functional test infrastructure were required.

There are a couple of questions:

- Should we reserve a range for "special-purpose" checks like this, for example `SARIF3000`-`3999`? Or even a sub-range for each "class" of special-purpose checks, like `SARIF3000`-`3099` for DSP checks?

- Remind me, do we need to apply `SARIF2017` or `SARIF2019` to `relatedLocations` as well? or any other locations?

    I would expect the answer to be "no" since the idea is for DSP to reject results whose actual result location is unknown.


src/Sarif.Multitool/Rules/SARIF2017.LocationsMustHaveRequiredProperties.cs
```diff
@@ -0,0 +1,99 @@
...
+                    locationsPointer,
+                    nameof(RuleResources.SARIF2017_LocationsMustHaveRequired_Properties_Error_EmptyLocationsArray_Text));
+                return;
+            }
+
+            for (int i = 0; i < result.Locations.Count; i++)
+            {
+                Location location = result.Locations[i];
+                string missingProperty = null;
+                string jsonPointer = null;
```
eddynaka, line 66: >string jsonPointer = null [](start = 16, length = 25)

I would start jsonPointer as LocationPointer.AtIndex(i), since every place you have that. Then, update the other places below to specific cases when necessary.

lgolding: I chose to do it this way because otherwise, we unnecessarily calculate `.AtIndex(i)` for every location in every result in the log file. This way, we only calculate it in the rare case where the rule fires.

---
In reply to: [463585630](https://github.com/microsoft/sarif-sdk/pull/2021#discussion_r463585630) [](ancestors = 463585630)

PR #2032, opened 2020-08-12
Improve package creation

lgolding:
- Create both a `DotnetTool` package and a `Dependency` package for `Sarif.Multitool` (the `Dependency` package is called `Sarif.Multitool.Library`.
- Add `netcoreapp3.1` as a supported target for `Sarif.Multitool`.
- DRY out the specification of target frameworks across all projects.
- Side effect: `WorkItems` now must create a package, because `Sarif.Multitool.Library` requires it.

### DIFF
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -20,7 +20,7 @@
 
   <PropertyGroup>
     <!--
-    For compatibility with older clients, we still build Sarid.Sdk (but no other packages)
+    For compatibility with older clients, we still build Sarif.Sdk (but no other packages)
     for .NET Framework 4.5.
     -->
     <TargetFrameworks>$(TargetFrameworks);net45</TargetFrameworks>
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -19,7 +19,11 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;net45</TargetFrameworks>
+    <!--
+    For compatibility with older clients, we still build Sarid.Sdk (but no other packages)
+    for .NET Framework 4.5.
+    -->
+    <TargetFrameworks>$(TargetFrameworks);net45</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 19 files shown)


src/build.props
```diff
@@ -1,8 +1,27 @@
...
+  <PropertyGroup Condition="'$(IsTestProject)' == ''">
+    <IsTestProject>False</IsTestProject>
+    <IsTestProject Condition="$(MSBuildProjectName.StartsWith('Test.'))">True</IsTestProject>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <IsExeProject>False</IsExeProject>
+    <IsExeProject Condition="'$(OutputType)' == 'Exe'">True</IsExeProject>
+
+    <TargetFrameworks Condition="$(IsExeProject)">netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
```
michaelcfanning, line 19: >netcoreapp2 [](start = 64, length = 11)

Why do we have both netcoreapp2.1 and netcoreapp3.1? can we simply make do with netcoreapp3.1?

src/Sarif/Sarif.csproj
```diff
@@ -19,7 +19,11 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;net45</TargetFrameworks>
+    <!--
+    For compatibility with older clients, we still build Sarid.Sdk (but no other packages)
```
michaelcfanning, line 23: >Sarid [](start = 57, length = 5)

typo s/be 'Sarif'

src/Sarif/Sarif.csproj
```diff
@@ -19,7 +19,11 @@
...
+    for .NET Framework 4.5.
+    -->
+    <TargetFrameworks>$(TargetFrameworks);net45</TargetFrameworks>
```
michaelcfanning, line 26: >net45 [](start = 42, length = 5)

Probably fine to zap this compatibility build moving forward. A good item to throw on the technical debt pile.

src/build.props
lgolding: Agreed. Also agree with your offline comment that we can defer.

---
In reply to: [476587808](https://github.com/microsoft/sarif-sdk/pull/2032#discussion_r476587808) [](ancestors = 476587808)

src/Sarif/Sarif.csproj
lgolding: Will file an infa-debt work item.

---
In reply to: [476588289](https://github.com/microsoft/sarif-sdk/pull/2032#discussion_r476588289) [](ancestors = 476588289)

src/build.props
lgolding: Will add this to the infra-debt work item for removing `net45` support from `Sarif.Sdk` package.

---
In reply to: [476590516](https://github.com/microsoft/sarif-sdk/pull/2032#discussion_r476590516) [](ancestors = 476590516,476587808)

PR #2035, opened 2020-08-18
Add GitInformation helper.

lgolding:
`GitInformation` and helper classes are from @michaelcfanning, tweaked and tests added by @lgolding. 

Integrated with `InsertOptionalDataVisitor`; introduce `--insert VersionControlInformation` option.

### DIFF
--- a/src/Sarif/Core/ArtifactLocation.cs
+++ b/src/Sarif/Core/ArtifactLocation.cs
@@ -106,5 +106,30 @@ namespace Microsoft.CodeAnalysis.Sarif
                 ? run.Artifacts[Index].Location
                 : this;
         }
+
+        /// <summary>
+        /// Creates a <see cref="Location"/> instance from data representing a single-line region.
+        /// </summary>
+        /// <param name="lineNumber">The line number within the file (1 based).</param>
+        /// <param name="column">The starting column number within the line (1-based).</param>
+        /// <param name="length">The length of the region, measured in characters.</param>
+        /// <param name="offset">The offset of the region start from the beginning of the file, measured in characters.</param>
+        /// <returns>An instance of a Location class.</returns>
+        public Location ToLocation(int lineNumber, int column, int length, int offset) =>
+            new Location
+            {
+                PhysicalLocation = new PhysicalLocation
+                {
+                    ArtifactLocation = this,
+                    Region = new Region
+                    {
+                        StartLine = lineNumber,
+                        StartColumn = column,
+                        EndColumn = column + length,
+                        CharOffset = offset,
+                        CharLength = length,
+                    },
+                },
+            };
     }
 }
--- a/src/Sarif/ExternalProcess.cs
+++ b/src/Sarif/ExternalProcess.cs
@@ -40,7 +40,6 @@ namespace Microsoft.CodeAnalysis.Sarif
             psi.WindowStyle = ProcessWindowStyle.Hidden;
 
             using (Process process = Process.Start(psi))
-            using (ManualResetEvent mreOut = new ManualResetEvent(false), mreErr = new ManualResetEvent(false))
             {
                 StdErr = new ConsoleStreamCapture();
                 StdOut = stdOut ?? new ConsoleStreamCapture();
--- a/src/Sarif/ExternalProcess.cs
+++ b/src/Sarif/ExternalProcess.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class ExternalProcess
+    {
+        public IConsoleCapture StdOut { get; private set; }
+
+        public IConsoleCapture StdErr { get; private set; }
+
+        public ExternalProcess(
+            string workingDirectory,
+            string fileName,
+            string arguments,
+            IConsoleCapture stdOut = null,
+            int[] acceptableReturnCodes = null)
+        {
+            acceptableReturnCodes = acceptableReturnCodes ?? new int[] { 0 };
+
+            ProcessStartInfo psi = new ProcessStartInfo();
+
+            psi.FileName = fileName;
+            psi.Arguments = arguments;
+            psi.WorkingDirectory = workingDirectory;
+
+            psi.ErrorDialog = false;
+            psi.CreateNoWindow = true;
+            psi.UseShellExecute = false;
+            psi.RedirectStandardInput = true;
+            psi.RedirectStandardError = true;
+            psi.RedirectStandardOutput = true;
+            psi.StandardOutputEncoding = System.Text.Encoding.UTF8;
+            psi.WindowStyle = ProcessWindowStyle.Hidden;
+
+            using (Process process = Process.Start(psi))
+            using (ManualResetEvent mreOut = new ManualResetEvent(false), mreErr = new ManualResetEvent(false))
+            {
+                StdErr = new ConsoleStreamCapture();
+                StdOut = stdOut ?? new ConsoleStreamCapture();
+
+                var tasks = new Task<string>[2];
+
+                tasks[0] = StdOut.Capture(process.StandardOutput, CancellationToken.None);
+                tasks[1] = StdErr.Capture(process.StandardError, CancellationToken.None);
+
+                process.WaitForExit();
+
+                Task.WaitAll(tasks);
+
+                if (!acceptableReturnCodes.Contains(process.ExitCode))
+                {
+                    Console.WriteLine("Command execution FAILED.");
+                    Console.WriteLine();
+                    Console.WriteLine($"Command-line     : {psi.FileName} {psi.Arguments}");
+                    Console.WriteLine($"Working directory: {Environment.CurrentDirectory}");
+                    Console.WriteLine();
+                    Console.WriteLine(StdErr.Text);
+                    Console.WriteLine();
+                    throw new InvalidOperationException();
+                }
+            }
+        }
+    }
+}
--- a/src/Sarif/GitInformation.cs
+++ b/src/Sarif/GitInformation.cs
@@ -1,0 +1,180 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal class GitInformation : IDisposable
+    {
+        private readonly IFileSystem fileSystem;
+        private readonly ReaderWriterLockSlim cacheLock = new ReaderWriterLockSlim();
+        private readonly Dictionary<string, VersionControlDetails> repoRoots = new Dictionary<string, VersionControlDetails>(StringComparer.OrdinalIgnoreCase);
+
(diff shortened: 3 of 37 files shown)


src/Sarif/Core/ArtifactLocation.cs
```diff
@@ -106,5 +106,30 @@ public ArtifactLocation Resolve(Run run)
                 ? run.Artifacts[Index].Location
                 : this;
         }
+
+        /// <summary>
```
lgolding, line 110: >/// [](start = 8, length = 3)

This was an extension method, but it doesn't need to be because we have partial clases.

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
```diff
@@ -4,10 +4,9 @@
 using System.Diagnostics;
 using System.IO;
 using FluentAssertions;
-using Microsoft.CodeAnalysis.Sarif;
 using Xunit;
 
-namespace Microsoft.CodeAnalysis.Test.FunctionalTests.Sarif
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
```
lgolding, line 9: >Multitool [](start = 39, length = 9)

While I'm in here, clean up a few namespaces to conform to code base standard that test classes live in the same namespace as corresponding product code.

I did not make the same change in Test.UnitTests.Sarif, even though I added a test file there as well, because there would have been too many changes for this PR.

src/Sarif/GitInformation.cs
```diff
@@ -0,0 +1,180 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+
+namespace Microsoft.CodeAnalysis.Sarif
```
lgolding, line 9: >Sarif [](start = 33, length = 5)

I did not put this in a `Git` subdirectory/namespace, since the other three files (IConsoleCapture.cs, ConsoleStreamCapture.cs, and ExternalProcess) are non Git-specific, so this would have been the only type in the namespace.

src/Sarif/ExternalProcess.cs
```diff
@@ -0,0 +1,71 @@
...
+            psi.CreateNoWindow = true;
+            psi.UseShellExecute = false;
+            psi.RedirectStandardInput = true;
+            psi.RedirectStandardError = true;
+            psi.RedirectStandardOutput = true;
+            psi.StandardOutputEncoding = System.Text.Encoding.UTF8;
+            psi.WindowStyle = ProcessWindowStyle.Hidden;
+
+            using (Process process = Process.Start(psi))
+            using (ManualResetEvent mreOut = new ManualResetEvent(false), mreErr = new ManualResetEvent(false))
```
lgolding, line 43: >mreErr [](start = 74, length = 6)

LGTM complains that these two assignments are useless because neither `mreOut` nor `mreErr` is used. Is there a reason to created these events?

src/Sarif/Core/ArtifactLocation.cs
michaelcfanning: Right but does this functionality conceptually belong as part of ArtifactLocation? I'm not sure there's a good case for it.

---
In reply to: [472406181](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r472406181) [](ancestors = 472406181)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -62,6 +68,28 @@ public override Run VisitRun(Run node)
             return visited;
         }
 
+        private void InsertVersionControlInformation()
+        {
+            var gitInformation = new GitInformation(s_fileSystem);
+            string currentDirectory = s_executionEnvironment.CurrentDirectory;
```
michaelcfanning, line 74: >CurrentDirectory [](start = 61, length = 16)

This will work some of the time but seems haphazard. What we need to happen is for every physical location that is included in the log file be examined for relevance to source control.

src/Test.UnitTests.Sarif/TestData/InsertOptionalDataVisitor/ExpectedOutputs/CoreTests-Absolute_TextFiles.sarif
```diff
@@ -65,7 +65,7 @@
       "artifacts": [
         {
           "location": {
-            "uri": "file:///REPLACED_AT_TEST_RUNTIME/src/Test.UnitTests.Sarif/TestData/InsertOptionalDataVisitor/InsertOptionalDataVisitor.txt"
+            "uri": "file:///ENLISTMENT_ROOT/src/Test.UnitTests.Sarif/TestData/InsertOptionalDataVisitor/InsertOptionalDataVisitor.txt"
```
michaelcfanning, line 68: >ENLISTMENT_ROOT [](start = 28, length = 15)

Why did you change this name? It was named previously as a clue that this value is rewritten. We've lost that information now.

Or did you discover that the renaming no longer occurs?

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -26,22 +27,32 @@ public InsertOptionalDataVisitorTests(ITestOutputHelper outputHelper, DeletesOut
...
                 GetResourceText(inputResourceName),
                 formatting: Formatting.Indented,
-                out string transformedLog);
+                updatedLog: out _);
+
+            // Some of the tests operate on SARIF files that mention the absolute path of the file
+            // that was "analyzed" (InsertOptionalDataVisitor.txt). That path depends on the repo
+            // root, and so can vary depending on the machine where the tests are run. To avoid
+            // this problem, both the input files and the expected output files contain a fixed
+            // string "ENLISTMENT_ROOT" in place of the directory portion of the path. But some
```
michaelcfanning, line 39: >ENLISTMENT_ROOT [](start = 23, length = 15)

Can you please think of a name that is an improvement on the previous (that meets the standard that prompted you to change it) but which still indicates that the value is rewritten by the test infrastructure? This is not obvious after the change you've made.

src/Sarif/ExternalProcess.cs
michaelcfanning: This is just dead code from an earlier draft, you can remove it.

---
In reply to: [472434440](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r472434440) [](ancestors = 472434440)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
lgolding: `ENVIRONMENT_DEPENDENT_ENLISTMENT_ROOT`?

---
In reply to: [475661249](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475661249) [](ancestors = 475661249)

src/Test.UnitTests.Sarif/TestData/InsertOptionalDataVisitor/ExpectedOutputs/CoreTests-Absolute_TextFiles.sarif
lgolding: `ENVIRONMENT_DEPENDENT_ENLISTMENT_ROOT`?

---
In reply to: [475662994](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475662994) [](ancestors = 475662994)

michaelcfanning: That doesn't indicate to readers that the test runtime replaces this value. Can you please just restore it to its previous value?

---
In reply to: [475669236](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475669236) [](ancestors = 475669236,475662994)

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
lgolding: This is the directory from which `GitInformation.GetRepositoryRoot` starts its upward search for the directory containing `.git`. The assumption was that this operation works only if you're running from somewhere under the repo root.

**BUT**: Now I see in `VisitPhysicalLocation` below, where it is enriching the regions and filling in the snippets, that it relies on `originalUriBaseIds` to resolve URIs -- and it if can't resolve a URI it silently fails to fill in the information.

So how about this: We try the operation starting with each entry in `originalUriBaseIds`. If all of them fail (that is, if there is no entry that points at or into a directory containing a `.git`), only then do we throw. (And we _do_ throw; I don't think it's ok to silently do nothing here.) We have to be careful in the presence of submodules to pick the topmost directory with a `.git`.

---
In reply to: [475661249](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475661249) [](ancestors = 475661249)

lgolding: (Having said that, I don't understand what you mean by examining every physical location. Do you mean "See if any physical location mentioned anywhere in the log file is an absolute URI from which the repo root can be determined?")

---
In reply to: [475691019](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475691019) [](ancestors = 475691019,475661249)

michaelcfanning: Yes. The goal of this feature is that any physical location referenced in the log file is expressed as a relative path with corresponding enlistment root, if that PLC is in fact a source-controlled item. The code as authored is not accomplishing this. 

---
In reply to: [475692378](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475692378) [](ancestors = 475692378,475691019,475661249)

lgolding: There are two distinct issues here:

1. How do we determine the repo root (and thus populate `run.versionControlProvenance`)? My implementation relies on the current directory; my alternative proposal above would rely on `originalUriBaseIds`, similar to what `VisitPhysicalLocation` does below.

2. Having determined the repo root, how do we express all URIs with respect to it? We're 100% agreed that we need to do that. **BUT**: as the Multitool is currently structured, that's not the job of the `rewrite` command/`InsertOptionalDataVisitor`. Rather, it's the job of the `rebaseuri` command/`RebaseUriVisitor`.

So IMO the code is structured correctly, and what the normalizer should do is invoke both the `rewrite` command and the `rebaseuri` command -- **EXCEPT** it turns out that the `rebaseuri` command supports the options for, and invokes, the `InsertOptionalDataVisitor`, and so it's actually a superset of the `rewrite` command, so a single invocation of `rebaseuri` will suffice.

This gets into the whole multitool redesign discussion. We can choose to do it now, or after the normalizer is done. But either way, the URI rebasing doesn't belong in the `InsertOptionalDataVisitor`.

---
In reply to: [475696046](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475696046) [](ancestors = 475696046,475692378,475691019,475661249)

michaelcfanning: I suggest we get this code in now. As we discussed offline, commands like rebaseuri are redudant and more granular than we need. 

I will put some thought into the larger design.

lgolding: Ok, I will merge what I have.

---
In reply to: [475714891](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475714891) [](ancestors = 475714891)

lgolding: I mean, "Ok, please merge what I have."

---
In reply to: [475769205](https://github.com/microsoft/sarif-sdk/pull/2035#discussion_r475769205) [](ancestors = 475769205,475714891)

PR #2038, opened 2020-08-20
Defer loading of the filing client until it's used.

rtaket:
Currently, the filing client is created and connected early in the process. This pattern breaks down when we obtain the organization/project information late in the process. i.e. from an external service. Deferred loading of the filing client fixes this issue because it allows the organization/project to be changed until the model is locked and the work item is going to be created.

### DIFF
--- a/src/Sarif.WorkItems/SarifWorkItemFiler.cs
+++ b/src/Sarif.WorkItems/SarifWorkItemFiler.cs
@@ -21,6 +21,9 @@ namespace Microsoft.CodeAnalysis.Sarif.WorkItems
 {
     public class SarifWorkItemFiler : IDisposable
     {
+        private readonly object m_syncRoot = new object();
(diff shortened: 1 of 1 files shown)


src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -45,15 +48,34 @@ public SarifWorkItemFiler(Uri filingUri = null, SarifWorkItemContext filingConte
...
+                            this.FilingClient = FilingClientFactory.Create(this.FilingContext.HostUri);
+                        }
+                    }
+                }
+
+                return m_filingClient;
+            }
+            set
+            {
+                m_filingClient = value;
```
lgolding, line 76: >m_filingClient = value; [](start = 16, length = 23)

Do we need to lock around this?

PR #2046, opened 2020-08-28
Make organization optional

nickmarston:
Makes the organization and project URI's optional, and able to be set by plugins if not provided on the command line.

### DIFF
--- a/src/Sarif.Multitool/FileWorkItemsCommand.cs
+++ b/src/Sarif.Multitool/FileWorkItemsCommand.cs
@@ -11,6 +11,7 @@ using Microsoft.CodeAnalysis.Sarif.Driver;
 using Microsoft.CodeAnalysis.Sarif.WorkItems;
 using Microsoft.Json.Schema;
 using Microsoft.Json.Schema.Validation;
+using Microsoft.TeamFoundation.Common;
 using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool
@@ -83,7 +84,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                 }
 
                 SarifLog sarifLog = null;
-                using (var filer = new SarifWorkItemFiler(filingContext.HostUri, filingContext))
+                using (var filer = new SarifWorkItemFiler(filingContext.WorkItemFilerUri, filingContext))
                 {
                     sarifLog = filer.FileWorkItems(logFileContents);
                 }
@@ -129,7 +130,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
             // If it succeeds, we can assign the result to a Uri-valued property (persisted in the
             // context object); if it fails, we can produce a helpful error message.
 
-            valid &= ValidateHostUri(options.HostUri, sarifWorkItemContext);
+            valid &= ValidateHostUri(options.workItemFilingUri, sarifWorkItemContext);
 
             valid &= options.ValidateOutputOptions();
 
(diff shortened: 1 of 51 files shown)


rtaket:
>         public int Run(FileWorkItemsOptions options, IFileSystem fileSystem)

Doesn't FileWorkItemsOptions need to be updated to make HostUri optional?

---
Refers to: src/Sarif.Multitool/FileWorkItemsCommand.cs:37 in 9a38bcf. [](commit_id = 9a38bcffd7bec4665b842e245243ce1131a18160, deletion_comment = False)

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -143,31 +144,26 @@ private bool ValidateOptions(FileWorkItemsOptions options, SarifWorkItemContext
...
-                // No project URI was provided via the --project-uri argument or as
-                // part of an input file specified via --configuration.
-                Console.Error.WriteLine(MultitoolResources.WorkItemFiling_NoHostUriSpecified);
-                return false;
-            }
 
-            Uri hostUri = null;
-            if (!string.IsNullOrEmpty(hostUriString) && !Uri.TryCreate(hostUriString, UriKind.RelativeOrAbsolute, out hostUri))
+            Uri workItemUri = null;
+            if (!string.IsNullOrEmpty(workItemUriString) && !Uri.TryCreate(workItemUriString, UriKind.RelativeOrAbsolute, out workItemUri))
```
rtaket, line 154: >RelativeOrAbsolute [](start = 102, length = 18)

I'm note sure how a relative URI makes sense. Shouldn't this just be absolute, since we're trying to extract the domain information?

src/Sarif.Multitool/FileWorkItemsCommand.cs
```diff
@@ -143,31 +144,26 @@ private bool ValidateOptions(FileWorkItemsOptions options, SarifWorkItemContext
             return valid;
         }
 
-        private static bool ValidateHostUri(string hostUriString, SarifWorkItemContext workItemFilingConfiguration)
+        // Validates that the URI coming from either the commandline or configuration is valid for filing work items.  A null or empty 
+        // value is valid if and only if a plug-in will later provide the value.  This cannot be determined until plugins are loaded later,
```
rtaket, line 148: >plug-in will later provide the value [](start = 43, length = 36)

Have you verified that using the org/project from the ownership service works? i.e. is it OK to retrieve the values that late in the process. Because if it doesn't work, we could just require the values in the command line or config.
Alternatively, we could create an IOwnershipTransformer interface and add the existence of ownership transformer to the check.

rtaket:
>         public SarifWorkItemModel FileWorkItemInternal(SarifLog sarifLog, SarifWorkItemContext filingContext, FilingClient filingClient)

Can we remove the filingClient parameter? Why did it exist if there's a FilingClient property?

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:267 in 9a38bcf. [](commit_id = 9a38bcffd7bec4665b842e245243ce1131a18160, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -272,26 +270,26 @@ public SarifWorkItemModel FileWorkItemInternal(SarifLog sarifLog, SarifWorkItemC
...
-                    sarifWorkItemModel.OwnerOrAccount = filingClient.AccountOrOrganization;
-                    sarifWorkItemModel.RepositoryOrProject = filingClient.ProjectOrRepository;
-
-                    if (filingContext.SyncWorkItemMetadata)
+                    if(!filingContext.HostUri.OriginalString.IsNullOrEmpty())
                     {
-                        Task<WorkItemModel> getMetadataTask = filingClient.GetWorkItemMetadata(sarifWorkItemModel);
-                        getMetadataTask.Wait();
-                        sarifWorkItemModel = (SarifWorkItemModel)getMetadataTask.Result;
+                        // The helper below will initialize the sarif work item model with a copy
```
rtaket, line 279: >// The helper [](start = 24, length = 13)

Does this comment block belong above the model creation on line 273?

rtaket:
>         public SarifWorkItemModel FileWorkItemInternal(SarifLog sarifLog, SarifWorkItemContext filingContext, FilingClient filingClient)

I see, I forgot to scrub the file after i added the FilingClient property? If that's the case, we should just remove the parameter, since it's unused.

---
In reply to: [683160735](https://github.com/microsoft/sarif-sdk/pull/2046#issuecomment-683160735) [](ancestors = 683160735)

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:267 in 9a38bcf. [](commit_id = 9a38bcffd7bec4665b842e245243ce1131a18160, deletion_comment = False)

PR #2050, opened 2020-08-31
Generate dependency package Sarif.Multitool.Library

lgolding:
We need to create both a DotnetTool package and a Dependency package for the Sarif.Multitool project. Users need a DotnetTool package so they can install the Multitool and run it from the command line. They need a Dependency package so they can add a NuGet reference to their project and invoke the package's public APIs. This is how we will expose the "normalization" functionality.

We can't expose the normalization functionality from the DotnetTool package we currently build because if you try to add a NuGet reference to a DotnetTool package you get the error message "Package '<packageId>' has a package type 'DotnetTool' that is not supported by project '<project>'."
 
It is possible to create a single package that declares multiple package types, but when you try to add a NuGet reference to such a package, you get the error message "Package '<packageId>' has multiple package types, which is not supported."

We create the Dependency package from the project file. We can't create the DotnetTool package from the project file for two reasons:

1. The tool package is standalone and requires dependent binaries that are not specified in the project file.
2. It's not possible (AFAIK) to create multiple packages from the same project file.

So we continue to use a .nuspec file to create the DotnetTool package.

### DIFF
--- a/scripts/NuGetUtilities.psm1
+++ b/scripts/NuGetUtilities.psm1
@@ -43,7 +43,7 @@ function New-NuGetPackageFromProjectFile($configuration, $project, $version) {
         "--include-source", "--include-symbols",
         "-p:Platform=$Platform",
         "--output", (Get-PackageDirectoryName $configuration),
-		"-p:Version=$version"
+        "-p:Version=$version"
 
     Write-CommandLine dotnet $arguments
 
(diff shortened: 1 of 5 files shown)


lgolding:
We're going to take a different approach to this: creating a separate library _project_, so we don't have to try to fool the build system into creating different packages from the same project. That should be much cleaner.

PR #2051, opened 2020-08-31
Mention VersionControlInformation in Multitool help

lgolding:
As of cef88f8, the `rewrite` command's `--insert` option accepts the value `VersionControlInformation`. When this value is specified, the command populates `run.versionControlProvenance`. In addition, any absolute URI that points into a Git repo is rewritten as a relative reference with respect to the nearest enclosing repo root.

We neglected to mention the new value in the command line help; do that now.

Also, we note that the description of the `rewrite` command did not at all say what the command actually does. Fix that.


lgolding:

lgolding:

lgolding:

PR #2054, opened 2020-09-01
Extract Sarif.Multitool.Library from Sarif.Multitool

lgolding:
We extract almost all the functionality from the `Sarif.Multitool` exe project into a separate library project `Sarif.Multitool.Library`. We continue to build a "DotnetTool package" from `Sarif.Multitool`, but now we also build a "Dependency" package from `Sarif.Multitool.Library`. This allows other projects to consume the library and access its public API.

The motivation for all this is to expose the "normalization API" (which is actually just an invocation of the `RewriteCommand` with certain options) so that it can be consumed.

NOTE: We ended up having to create a package from the `WorkItems` project, which we hadn't previously done, to get the `Sarif.Multitool.Library` package to build.


eddynaka (approved):

rtaket:
>     <None Update="default.configuration.xml">

Does this need to move too?

---
Refers to: src/Sarif.Multitool/Sarif.Multitool.csproj:62 in 8f631e6. [](commit_id = 8f631e607576da9ede325c007b62a5da966e61da, deletion_comment = False)

rtaket (approved):
:shipit:

scripts/Projects.psm1
```diff
@@ -26,7 +26,9 @@ $Projects.Libraries = @(
     "Sarif",
     "Sarif.Converters",
     "Sarif.Driver",
-    "Sarif.WorkItems"
+    "Sarif.Multitool.Library"
```
rtaket, line 29: >Sarif.Multitool.Library [](start = 5, length = 23)

Will the multitool library be shipped in the Sarif.Sdk nuget package?

lgolding: No, it's a separate NuGet package `Sarif.Multitool.Library`.

---
In reply to: [483762094](https://github.com/microsoft/sarif-sdk/pull/2054#discussion_r483762094) [](ancestors = 483762094)

lgolding:
>     <None Update="default.configuration.xml">

Great catch! Thanks.

---
In reply to: [687284296](https://github.com/microsoft/sarif-sdk/pull/2054#issuecomment-687284296) [](ancestors = 687284296)

---
Refers to: src/Sarif.Multitool/Sarif.Multitool.csproj:62 in 8f631e6. [](commit_id = 8f631e607576da9ede325c007b62a5da966e61da, deletion_comment = False)

Issue #2060, opened 2020-09-03
GitHubDspIngestionVisitor removes codeFlows entirely rather than inlining threadFlowLocations


lgolding:
@michaelcfanning FYI

lgolding:
Fixed in #2061.

Issue #2058, opened 2020-09-02
GitHubDspIngestionVisitor limits number of results only if some of them are non-errors


lgolding:
The visitor is supposed to truncate the results list to 500, but the logic is such that it only does that if some of the results are non-errors (_e.g._, warnings or notes).

@michaelcfanning FYI

lgolding:
Fixed in #2061

PR #2061, opened 2020-09-03
Fix GitHubDspIngestionVisitor

lgolding:
Add file-based unit tests for the `GitHubDspIngestionVisitor` and fix the bugs that they reveal:

- #2058: GitHubDspIngestionVisitor limits number of results only if some of them are non-errors
- #2060: GitHubDspIngestionVisitor removes codeFlows entirely rather than inlining threadFlowLocations

When inlining `threadFlowLocations`, we have to take care to distinguish those properties that are shared among all usages of a given location (for example, the `location` and `module` properties) from those properties that vary across usages (for example, `executionTimeUtc` and `state`). The `location.message` requires special handling because we do want to take `location` from the shared object, but we don't necessarily want to take `location.message`. Finally, we merge the property bags, preferring the values in the per-usage property bag if there are any overlaps.

We also have to merge property bags when inlining artifact locations.

### DIFF
--- a/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
+++ b/src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
@@ -59,7 +59,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
 
             SetFirstDetectionTime(result);
 
-            resultMatchingProperties = MergeDictionaryPreferFirst(resultMatchingProperties, originalResultMatchingProperties);
+            resultMatchingProperties = resultMatchingProperties.MergePreferFirst(originalResultMatchingProperties) as Dictionary<string, object>;
 
             if (PreviousResult != null &&
                 propertyBagMergeBehavior == DictionaryMergeBehavior.InitializeFromOldest)
@@ -226,22 +226,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching
             }
 
             targetResult.Provenance.FirstDetectionTimeUtc = firstDetectionTime;
-        }
-
-        private Dictionary<string, object> MergeDictionaryPreferFirst(
-            Dictionary<string, object> firstPropertyBag,
-            Dictionary<string, object> secondPropertyBag)
-        {
-            Dictionary<string, object> mergedPropertyBag = firstPropertyBag;
-
-            foreach (string key in secondPropertyBag.Keys)
-            {
-                if (!mergedPropertyBag.ContainsKey(key))
-                {
-                    mergedPropertyBag[key] = secondPropertyBag[key];
-                }
-            }
-            return mergedPropertyBag;
         }
 
         public override string ToString()
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -441,6 +441,35 @@ namespace Microsoft.CodeAnalysis.Sarif
             return null;
         }
 
+        /// <summary>
+        /// Merge this property bag with another one, preferring the properties in this property
+        /// bag if there are any duplicates.
+        /// </summary>
+        /// <param name="propertyBag">
+        /// The property bag into which <paramref name="otherPropertyBag"/> is to be merged.
+        /// </param>
+        /// <param name="otherPropertyBag">
+        /// A property bag containing properties to merge into <paramref name="propertyBag"/>.
+        /// </param>
+        /// <returns>
+        /// The original <paramref name="propertyBag"/> object, into which the properties
+        /// from <paramref name="otherPropertyBag"/> have now been merged.
+        /// </returns>
+        public static IDictionary<string, T> MergePreferFirst<T>(
+            this IDictionary<string, T> propertyBag,
+            IDictionary<string, T> otherPropertyBag)
+        {
+            foreach (string key in otherPropertyBag.Keys)
+            {
+                if (!propertyBag.ContainsKey(key))
+                {
+                    propertyBag[key] = otherPropertyBag[key];
+                }
+            }
+
+            return propertyBag;
+        }
+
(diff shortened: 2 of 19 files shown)


src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -228,22 +228,6 @@ internal void SetFirstDetectionTime(Result targetResult)
             targetResult.Provenance.FirstDetectionTimeUtc = firstDetectionTime;
         }
 
-        private Dictionary<string, object> MergeDictionaryPreferFirst(
```
lgolding, line 231: >MergeDictionaryPreferFirst [](start = 43, length = 26)

Moved to ExtensionMethods.cs because the `GitHubDspIngestionVisitor` now needs it, too.

src/Sarif/ExtensionMethods.cs
```diff
@@ -441,6 +441,39 @@ internal static string PropertyValue(this Dictionary<string, string> properties,
...
+        /// </param>
+        /// <typeparam name="T">
+        /// The type of object in the property bag. For SARIF property bags, the type will
+        /// be <see cref="SerializedPropertyInfo"/>.
+        /// </typeparam>
+        /// <returns>
+        /// The original <paramref name="propertyBag"/> object, into which the properties
+        /// from <paramref name="otherPropertyBag"/> have now been merged.
+        /// </returns>
+        public static IDictionary<string, T> MergePreferFirst<T>(
```
lgolding, line 462: >T [](start = 62, length = 1)

Type parameter needed because the `GitHubDspIngestionVisitor` uses this method with SARIF property bags -- so `T` is `SerializedPropertyInfo` -- whereas the result matching code where I found this was using it with a dictionary whose values were of type `object`.

src/Sarif/Visitors/GitHubDspIngestionVisitor.cs
```diff
@@ -43,32 +55,103 @@ public override Run VisitRun(Run node)
                             errors.Add(result);
                         }
 
-                        // GitHub DSP reportedly has an ingestion limit of 500 issues
-                        if (errors.Count > 500) { break; }
+                        if (errors.Count > s_MaxResults) { break; }
```
rtaket, line 58: >> [](start = 41, length = 1)

== instead of > ?

lgolding: Right :)

---
In reply to: [483744382](https://github.com/microsoft/sarif-sdk/pull/2061#discussion_r483744382) [](ancestors = 483744382)

rtaket:
>                     foreach (Result result in node.Results)

Is there a reason we don't merge the two foreach loops?

---
Refers to: src/Sarif/Visitors/GitHubDspIngestionVisitor.cs:51 in 7570d52. [](commit_id = 7570d52dd1d28a49158edad0b681ffc42cd84f9d, deletion_comment = False)

lgolding:
>                     foreach (Result result in node.Results)

No principled reason. This was the way the code came to me and I just didn't consider that rearrangement. I'll take a look.

---
In reply to: [687266645](https://github.com/microsoft/sarif-sdk/pull/2061#issuecomment-687266645) [](ancestors = 687266645)

---
Refers to: src/Sarif/Visitors/GitHubDspIngestionVisitor.cs:51 in 7570d52. [](commit_id = 7570d52dd1d28a49158edad0b681ffc42cd84f9d, deletion_comment = False)

lgolding:
>                     foreach (Result result in node.Results)

Oh wait, right, I did look. The reason is that we don't want to create a new list until we have to.

---
In reply to: [687267202](https://github.com/microsoft/sarif-sdk/pull/2061#issuecomment-687267202) [](ancestors = 687267202,687266645)

---
Refers to: src/Sarif/Visitors/GitHubDspIngestionVisitor.cs:51 in 7570d52. [](commit_id = 7570d52dd1d28a49158edad0b681ffc42cd84f9d, deletion_comment = False)

src/Sarif/Visitors/GitHubDspIngestionVisitor.cs
```diff
@@ -78,15 +161,11 @@ public override Result VisitResult(Result node)
...
-                    // DSP requires that every related location have a message. It's
-                    // not clear why this requirement exists, as this data is mostly 
-                    // used to build embedded links from results (where the link
-                    // anchor text actually resides).
                     if (string.IsNullOrEmpty(relatedLocation.Message?.Text))
                     {
                         relatedLocation.Message = new Message
                         {
-                            Text = "[No message provided.]" 
+                            Text = PlaceholderRelatedLocationMessage
```
rtaket, line 168: >Text [](start = 28, length = 4)

I thought that Messages could have either a Text or Id property to define the string. Here, if the message contains an Id, we're adding a text string. Is that OK?

rtaket: ... I should have added, I realize the previous code did this too, but I thought I'd ask anyway.

---
In reply to: [483750179](https://github.com/microsoft/sarif-sdk/pull/2061#discussion_r483750179) [](ancestors = 483750179)

lgolding: It's a good question. We're compensating here for a deficiency in the DSP -- the DSP looks for the `Text` property.

---
In reply to: [483750179](https://github.com/microsoft/sarif-sdk/pull/2061#discussion_r483750179) [](ancestors = 483750179)

lgolding: ... when in fact it shouldn't care about the presence of related location messages at all.

---
In reply to: [483751157](https://github.com/microsoft/sarif-sdk/pull/2061#discussion_r483751157) [](ancestors = 483751157,483750179)

rtaket (approved):
:shipit:

PR #2065, opened 2020-09-05
Support queries against properties in the result's and the rule's property bags

lgolding:
For example:

 ```
sarif query "properties.confidence >= 0.95 AND rule.properties.category == security" fortify.sarif
```

- We can query against properties in both the result's and the rule's property bags. Identify result properties with `properties.<propertyName>`, and properties belonging to the associated rule with `rule.properties.<propertyName>`, as in the example.
- We can query against string, integer, and float properties.

Behavioral details open for discussion:

1. If the comparison operator is string-specific (namely, `:` (contains), `|>` (starts with), or `>|` (ends with)), then we always treat the property value as a string. For all other comparison operators (for example, `==`, `>`, `<=`, _etc._), we decide whether to use a string comparison or a numeric comparison by attempting to parse the value on the right-hand side as a number. If it can be parsed, we use a numeric comparison; otherwise we use a string comparison. This could cause problems if you have string-valued properties that look like numbers.

2. Property name lookups are case-insensitive. If a property bag contains two properties whose names differ only by case, the query will evaluate whichever one it sees first. Should we use case-sensitive property name lookup? or throw if there's more than one property name that differs only by case?

3. String value comparisons are case-insensitive, and there is no syntax for specifying a case-sensitive comparison. Should there be a way to specify? Should the default behavior be case-sensitive comparison?

4. If a numeric comparison is specified (because the operator is compatible with numeric values, and the right-hand side parses successfully as an integer), and if the property value _cannot_ be parsed, then the property is treated as having value numeric 0. Should we throw instead?

### CODE REVIEW

File: src/Sarif/Query/Evaluators/SarifEvaluators.cs [hunk 1]
```diff
@@ -43,7 +45,19 @@ public static IExpressionEvaluator<Result> ResultEvaluator(TermExpression term)
...
 
                 default:
-                    throw new QueryParseException($"Property Name {term.PropertyName} unrecognized. Known Names: baselineState, correlationGuid, guid, hostedViewerUri, kind, level, message.text, occurrenceCount, rank, ruleId");
+                    if (propertyNameLower.StartsWith(PropertyBagPropertyEvaluator.ResultPropertyPrefix) ||
+                        propertyNameLower.StartsWith(PropertyBagPropertyEvaluator.RulePropertyPrefix))
+                    {
+                        return new PropertyBagPropertyEvaluator(term);
+                    }
+                    else
+                    {
```
michaelcfanning, line 54: you don't need this else block, simply fall through to the throw.

src/Test.UnitTests.Sarif.Multitool/QueryCommandPropertyBagTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        [Fact]
+        public void QueryCommand_CanAccessResultAndRulePropertyBags()
+        {
+            RunAndVerifyCount(2, "properties.name == 'Terisa'");
+            RunAndVerifyCount(2, "rule.properties.Category == 'security'");
+        }
+
+        [Fact]
+        public void QueryCommand_ComparesPropertyBagPropertyNamesCaseInsensitively()
```
michaelcfanning, line 35: >QueryCommand_ComparesPropertyBagPropertyNamesCaseInsensitively [](start = 20, length = 62)

curious, why do we permit this? are our property bag keys case-insensitive?

src/Sarif/Query/Evaluators/PropertyBagPropertyEvaluator.cs
```diff
@@ -0,0 +1,121 @@
...
+            List<string> propertyNames = holder.PropertyNames
+                .Where(key => key.Equals(_propertyName, StringComparison.OrdinalIgnoreCase))
+                .ToList();
+
+            if (propertyNames.Any())
+            {
+                try
+                {
+                    // If the property bag contains two properties whose names differ only
+                    // by case, we'll take the first one.
```
michaelcfanning, line 103: >we'll take the first on [](start = 32, length = 23)

why have we implemented a case-insensitive property name search? it seems like this is simply adding more complexity for us (and introducing surprising results such as returning an arbitrary match).

michaelcfanning (changes requested):
:clock1:

michaelcfanning:
If you have a moment to resolve conflicts with checked in code, that would be good.

src/Test.UnitTests.Sarif.Multitool/QueryCommandPropertyBagTests.cs
lgolding: Our property bag keys are case-sensitive. Property "TeRisA" is indeed separate from "Terisa". All I can say is, "It seemed like a good idea at the time." I'll fix it.

---
In reply to: [496994367](https://github.com/microsoft/sarif-sdk/pull/2065#discussion_r496994367) [](ancestors = 496994367)

lgolding:
Done.

---
In reply to: [700939098](https://github.com/microsoft/sarif-sdk/pull/2065#issuecomment-700939098) [](ancestors = 700939098)

src/Test.UnitTests.Sarif.Multitool/QueryCommandPropertyBagTests.cs
lgolding: Actually, I do remember -- I wanted to make it easier for a user to type the query without having to worry about the case of the property name. But still, I'll fix it.

---
In reply to: [497042412](https://github.com/microsoft/sarif-sdk/pull/2065#discussion_r497042412) [](ancestors = 497042412,496994367)

src/Sarif/Query/Evaluators/PropertyBagPropertyEvaluator.cs
lgolding: Agreed, fixing.

---
In reply to: [496995760](https://github.com/microsoft/sarif-sdk/pull/2065#discussion_r496995760) [](ancestors = 496995760)

src/Sarif/Query/Evaluators/PropertyBagPropertyEvaluator.cs
```diff
@@ -0,0 +1,108 @@
...
+        private PropertyBagHolder GetPropertyBagHolder(Result result) =>
+            _propertyBelongsToRule
+            ? result.GetRule() as PropertyBagHolder
+            : result;
+
+        private T GetPropertyFromHolder<T>(PropertyBagHolder holder)
+        {
+            try
+            {
+                return holder.TryGetProperty(_propertyName, out T value) ? value : default;
```
michaelcfanning, line 92: >TryGetProperty [](start = 30, length = 14)

this looks weird. Why is this named TryGetProperty if it can raise an exception? seems wrong to me. instead, TryProperty should handle the exception and return 'false'.

i won't block on this but i don't this it's right.

PR #2068, opened 2020-09-09
Fixing a couple of edge-cases that cause the ESLint formatter to create invalid SARIF files

cbraynor:
This fixes two issues we've seen in the wild in the ESLint formatter:

- When a rule doesn't have a description the resulting SARIF output would have a `shortDescription` but contain no `text` field - which is required. This removes the outer object `shortDescription` in that case.
- When a result returned the invalid column number of 0 (or negative) the resulting SARIF output would contain the invalid column number. This treats invalid column numbers the same as undefined column numbers

### DIFF
--- a/src/ESLint.Formatter/sarif.js
+++ b/src/ESLint.Formatter/sarif.js
@@ -181,9 +181,12 @@ module.exports = function (results, data) {
                     }
 
                     if (message.line > 0 || message.column > 0) {
-                        sarifRepresentation.locations[0].physicalLocation.region = {
-                            startLine: message.line,
-                            startColumn: message.column
+                        sarifRepresentation.locations[0].physicalLocation.region = {};
+                        if (message.line > 0) {
+                            sarifRepresentation.locations[0].physicalLocation.region.startLine = message.line;
+                        }
+                        if (message.column > 0) {
+                            sarifRepresentation.locations[0].physicalLocation.region.startColumn = message.column;
                         };
                     }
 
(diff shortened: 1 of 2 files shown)


cbraynor:
@lgolding looks like you are familiar with this area of the codebase, would you mind taking a look. Thanks!

lgolding:
Hi Chris, I'm off work this week. I'll be happy to take a look when I return on Monday.

cbraynor:
Much appreciated, thanks!

src/ESLint.Formatter/sarif.js
```diff
@@ -179,9 +181,12 @@ module.exports = function (results, data) {
                     }
 
                     if (message.line > 0 || message.column > 0) {
-                        sarifRepresentation.locations[0].physicalLocation.region = {
-                            startLine: message.line,
-                            startColumn: message.column
+                        sarifRepresentation.locations[0].physicalLocation.region = {};
+                        if (message.line > 0) {
+                            sarifRepresentation.locations[0].physicalLocation.region.startLine = message.line;
+                        }
```
lgolding, line 187: >} [](start = 24, length = 1)

This change is fine. Or, to be precise, it's an improvement over the existing code. Be aware that if a SARIF region is being represented with line/column properties (SARIF can represent regions in other ways, such as offset/length), then the `startLine` property is required. So if you did have a message with a valid column but without a valid line, then the resulting SARIF file would be invalid. Having said that -- again, you're no worse off than before.

lgolding (approved):
:shipit:

lgolding:
Chris, this change looks good. Thanks for adding the test. I will merge it. @michaelcfanning FYI

src/ESLint.Formatter/sarif.js
cbraynor: That makes sense - though it's probably less common that a tool outputs a column without a line. Thanks

Issue #2070, opened 2020-09-15
ESLint.Formatter - Add eslint version to run.tool.driver


tosmolka:
We would like to add ESLint version into `tool.driver.version`. According to sarif spec, this property is optional. In our case, this will help us with reporting later on, when processing large number of sarif files from multiple projects. 

Also, this will avoid following validation error from Sarif.Multitool:

`warning SARIF2005: runs[0].tool.driver: The tool 'ESLint' provides neither a 'version' property nor a 'semanticVersion' property. Providing a version enables the log file consumer to determine whether the file was produced by an up to date version, and to avoid accidentally comparing log files produced by different tool versions.`

I have linked a small PR as a proposal, let me know what you think. Thanks.

lgolding:
Fixed in a22af18.

PR #2071, opened 2020-09-15
ESLint.Formatter - Add eslint version to run.tool.driver #2070

tosmolka:
Proposal for #2070

### DIFF
--- a/src/ESLint.Formatter/package.json
+++ b/src/ESLint.Formatter/package.json
@@ -34,5 +34,8 @@
     "lodash": "^4.17.14",
     "mocha": "^7.1.2",
     "utf8": "^3.0.0"
+  },
+  "devDependencies": {
+    "rewire": "^5.0.0"
   }
 }
--- a/src/ESLint.Formatter/package.json
+++ b/src/ESLint.Formatter/package.json
@@ -29,13 +29,13 @@
   },
   "homepage": "https://github.com/microsoft/sarif-sdk/tree/master/src/ESLint.Formatter#readme",
   "dependencies": {
-    "chai": "^4.2.0",
     "jschardet": "latest",
     "lodash": "^4.17.14",
-    "mocha": "^7.1.2",
     "utf8": "^3.0.0"
   },
   "devDependencies": {
+    "chai": "^4.2.0",
+    "mocha": "^7.1.2",
     "rewire": "^5.0.0"
   }
 }
--- a/src/ESLint.Formatter/test/sarif.js
+++ b/src/ESLint.Formatter/test/sarif.js
@@ -67,6 +67,7 @@ describe("formatter:sarif", () => {
 
             assert.strictEqual(log.runs[0].tool.driver.name, "ESLint");
             assert.strictEqual(log.runs[0].tool.driver.informationUri, "https://eslint.org");
+            assert.strictEqual(log.runs[0].tool.driver.version, undefined);
         })
     });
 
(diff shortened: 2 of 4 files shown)


lgolding:
>             assert.strictEqual(log['$schema'], 'http://json.schemastore.org/sarif-2.1.0-rtm.4');

While you're in here -- I just noticed this. The final version of the schema -- the one that corresponds to the final, approved public specification from OASIS -- is `-rtm.5`. The only difference between `-rtm.4` and `-rtm.5` is to correct a misspelling of a property name which I see you do not use.

To ensure that the resulting SARIF is good and conforms to `-rtm.5`, you can browse to the [SARIF Validator](http://sarifweb.azurewebsites.net/Validation) and drop your SARIF file onto the drop target. If you turn on note-level rules, you'll get lots of advice on how to improve the SARIF.

---
Refers to: src/ESLint.Formatter/test/sarif.js:65 in 0335691. [](commit_id = 03356913632fda150234596145ee5cb90b920951, deletion_comment = False)

src/ESLint.Formatter/test/sarif.js
```diff
@@ -67,6 +67,7 @@ describe("formatter:sarif", () => {
 
             assert.strictEqual(log.runs[0].tool.driver.name, "ESLint");
             assert.strictEqual(log.runs[0].tool.driver.informationUri, "https://eslint.org");
+            assert.strictEqual(log.runs[0].tool.driver.version, undefined);
```
lgolding, line 70: >undefined [](start = 64, length = 9)

Kind of a pity that there can't be a test that demonstrates the "real" value. Is there a way to "mock" the source of the version property, so you can see if flows through to the SARIF properly?

lgolding:
Thank you for your contribution!

@michaelcfanning FYI

lgolding (changes requested):
:clock1:

src/ESLint.Formatter/test/sarif.js
tosmolka: Thanks. Added new test case, I had to use `rewire` to mock the value, can you please take a look?

src/ESLint.Formatter/package.json
```diff
@@ -34,5 +34,8 @@
     "lodash": "^4.17.14",
     "mocha": "^7.1.2",
```
tosmolka, line 35: Would you be OK with moving `chai` and `mocha` to `devDependencies` as well?

lgolding: Yes, thank you.

---
In reply to: [490201370](https://github.com/microsoft/sarif-sdk/pull/2071#discussion_r490201370) [](ancestors = 490201370)

src/ESLint.Formatter/test/sarif.js
```diff
@@ -59,15 +60,28 @@ describe("formatter:sarif", () => {
     describe("when run", () => {
         const code = [];
 
-        it ("should return a log with correct version and tool metadata", () => {
+        it ("should return a log with correct sarif version and tool metadata", () => {
```
lgolding, line 63: >sarif [](start = 46, length = 5)

The ultimate nitpick: "SARIF", capitalized.

src/ESLint.Formatter/test/sarif.js
lgolding: Nice! Thanks.

---
In reply to: [490200329](https://github.com/microsoft/sarif-sdk/pull/2071#discussion_r490200329) [](ancestors = 490200329)

lgolding:
@tosmolka, Thanks again. I approved the PR. Would you please just capitalize "SARIF" in the place I mentioned in the comment. Then we'll merge this.

lgolding (approved):
:shipit:

tosmolka:
@lgolding , updated.

PR #2072, opened 2020-09-16
Adding Delete in IFileSystem

### LINKED ISSUE #2033: Add IFileSystem.DeleteFile

eddynaka:
Fix #2033


src/Sarif/IFileSystem.cs
```diff
@@ -217,5 +217,11 @@ public interface IFileSystem
         /// The name of the empty directory to remove. This directory must be writable and empty.
         /// </param>
         void DeleteDirectory(string path, bool recursive = false);
+
+        /// <summary>
+        /// Deletes a file from a specified path.
```
lgolding, line 222: >Deletes [](start = 12, length = 7)

I have been copying this documentation from Microsoft online docs, here: [https://docs.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=netcore-3.1)

So, summary: "Deletes the specified file.", param: "The name of the file to be deleted. Wildcard characters are not supported."

lgolding (approved):
:shipit:

src/Sarif/FileSystem.cs
```diff
@@ -266,5 +266,16 @@ public void DeleteDirectory(string path, bool recursive = false)
...
             Directory.Delete(path, recursive);
         }
+
+        /// <summary>
+        /// Deletes the specified file.
+        /// </summary>
+        /// <param name="path">
+        /// The name of the file to be deleted. Wildcard characters are not supported.
+        /// </param>
+        public void DeleteFile(string path)
```
michaelcfanning, line 276: >DeleteFile [](start = 20, length = 10)

Generally, we want these names to match the corresponding system API. In this case, we have a problem: Delete is shared as a name between multiple types. I suggest that we prefix Delete with that enclosing type name. So this method s/be named 'FileDelete' and the one preceding it s/be named 'DirectoryDelete'

michaelcfanning:
>         public void DeleteDirectory(string path, bool recursive = false)

Suggest renaming this as DirectoryDelete, as rationalized in separate comment.

---
Refers to: src/Sarif/FileSystem.cs:265 in d077367. [](commit_id = d077367a2b163c8c3dc3b417fcd34fa84dedb28e, deletion_comment = False)

michaelcfanning (changes requested):
:clock1:

src/Sarif/FileSystem.cs
eddynaka: Should we rename the method CreateDirectory as well?

---
In reply to: [496997632](https://github.com/microsoft/sarif-sdk/pull/2072#discussion_r496997632) [](ancestors = 496997632)

michaelcfanning: Well, yes, to be consistent,this looks right. I think that also means we need to rename 'Create' to 'FileCreate', since it maps to File.Create?

---
In reply to: [497004102](https://github.com/microsoft/sarif-sdk/pull/2072#discussion_r497004102) [](ancestors = 497004102,496997632)

eddynaka: just updated both

---
In reply to: [497005275](https://github.com/microsoft/sarif-sdk/pull/2072#discussion_r497005275) [](ancestors = 497005275,497004102,496997632)

michaelcfanning (approved):
:shipit:

PR #2075, opened 2020-09-17
Refactor the SARIF Multitool unit tests

lgolding:
Since we have factored a library out of `Sarif.Multitool`, we should factor the tests as well.

At first I thought this was a pure rename, since the only file left in `Sarif.Multitool` was `Program.cs`, and (as far as I knew) there were no tests for `Program.cs` (which does nothing but dispatch the verb to the appropriate command class).

But it turns out that there is exactly one test which does call `Program.cs` directly (whether it _should_ is another question), so I have retained the project `Test.UnitTests.Sarif.Multitool` to hold that one test, and moved all the others into `Test.UnitTests.Sarif.Multitool.Library`.

Only two test classes required changes due to the move:
- `GenericCommandTests.cs`
- `MergeCommandTests.cs`


eddynaka (approved):

src/Test.UnitTests.Sarif.Multitool/Test.UnitTests.Sarif.Multitool.csproj
```diff
@@ -13,48 +13,16 @@
...
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="coverlet.collector" Version="1.2.0" />
```
michaelcfanning, line 21: >coverlet.collector [](start = 31, length = 18)

why was this introduced?

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif.Multitool/Test.UnitTests.Sarif.Multitool.csproj
lgolding: The sequence was:

1. Rename `Test.UnitTests.Sarif.Multitool` to `Test.UnitTests.Sarif.Multitool.Library`. That way I made sure all the test data files and so on got carried over.
2. Create a _new_ `Test.UnitTests.Sarif.Multitool` project to hold the one test that actually uses the `Program` class.

So apparently, newly created .NET Core xUnit Test projects come equipped with `coverlet.collection` (a code coverage library that VS uses when use select **Test**, **Analyze Code Coverage**.

---
In reply to: [490579714](https://github.com/microsoft/sarif-sdk/pull/2075#discussion_r490579714) [](ancestors = 490579714)

PR #2082, opened 2020-09-22
Generalize and harden docs exporter

### LINKED ISSUE #2080: Generalize and harden the export-docs command
The method `ExportRulesDocumentationCommandBase` makes certain assumptions which were valid in the context of the SARIF validator – because we controlled all the analysis rule classes – but which might not be valid for all tools that implement the `export-docs` command:

1.	The description is taken from `rule.FullDescription`. But `FullDescription` might be absent and only `ShortDescription` present – or no description at all.
2.	It assumes a particular format for the result message string identifiers (underscore-separated identifier, with the last segment being the “friendly name”).
3.	It always displays the `Text` of the description and the rule messages, but if `Markdown` is available it should be preferred.

eddynaka:
Fix #2080


src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
```diff
@@ -5,12 +5,14 @@
...
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
     public class ExportRulesDocumentationCommandBase<TContext> : PlugInDriverCommand<ExportRulesDocumentationOptions>
     {
         private readonly IFileSystem _fileSystem;
+        private static readonly Regex s_friendlyNameRegex = new Regex(@"(?<level>Error|Warning|Note|None)_(?<friendlyName>[^_]+)$");
```
lgolding, line 15: >) [](start = 130, length = 1)

`RegexOptions.Compiled | RegexOptions.IgnoreCase`

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
```diff
@@ -31,7 +33,7 @@ public override int Run(ExportRulesDocumentationOptions options)
                     BuildRule(rule, sb);
                 }
 
-                _fileSystem.WriteAllText(options.OutputFilePath, sb.ToString());
+                _fileSystem.WriteAllText(options.OutputFilePath ?? "Rules.md", sb.ToString());
```
lgolding, line 36: >"Rules.md" [](start = 67, length = 10)

`private const string DefaultOutputFileName';`

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
```diff
@@ -46,13 +48,20 @@ internal void BuildRule(Skimmer<TContext> rule, StringBuilder sb)
         {
             sb.AppendLine($"## Rule `{rule.Moniker}`{Environment.NewLine}");
             sb.AppendLine($"### Description{Environment.NewLine}");
-            sb.AppendLine($"{rule.FullDescription.Text}{Environment.NewLine}");
+            sb.AppendLine($"{rule.FullDescription.Text ?? rule.ShortDescription.Text ?? "No description available."}{Environment.NewLine}");
```
lgolding, line 51: >"No description available." [](start = 88, length = 27)

Resource string `NoRuleDescription`. Your message text is fine.

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
```diff
@@ -46,13 +48,20 @@ internal void BuildRule(Skimmer<TContext> rule, StringBuilder sb)
...
 
             foreach (KeyValuePair<string, MultiformatMessageString> message in rule.MessageStrings)
             {
-                sb.AppendLine($"#### `{message.Key.Split('_').Last()}`: {rule.DefaultLevel}{Environment.NewLine}");
-                sb.AppendLine($"{message.Value.Text}{Environment.NewLine}");
+                string ruleName = message.Key;
+                Match match = s_friendlyNameRegex.Match(message.Key);
+                if (match.Success)
+                {
+                    ruleName = match.Groups[2].Value;
```
lgolding, line 60: >2 [](start = 44, length = 1)

For code readability and also for robustness (because if you change the regex, this index might change): `match.Groups["friendlyName"].Value`

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
```diff
@@ -46,13 +48,20 @@ internal void BuildRule(Skimmer<TContext> rule, StringBuilder sb)
...
+                }
+
+                sb.AppendLine($"#### `{ruleName}`: {rule.DefaultLevel}{Environment.NewLine}");
+                sb.AppendLine($"{(string.IsNullOrEmpty(message.Value.Markdown) ? message.Value.Text : message.Value.Markdown)}{Environment.NewLine}");
```
lgolding, line 64: >IsNullOrEmpty [](start = 41, length = 13)

If you're going to do this test, I'd use `IsNullOrWhiteSpace`. Otherwise just `message.Value.Markdown ?? message.Value.Text`.

src/Test.UnitTests.Sarif.Multitool.Library/ExportValidationRulesDocumentationCommandTests.cs
```diff
@@ -36,5 +45,31 @@ private class TestRule : SarifValidationSkimmerBase
                 { "TEST0001_TEST_Note_Default", new MultiformatMessageString{ Text="default text"} }
             };
         }
+
+        private class TestRule2 : SarifValidationSkimmerBase
+        {
+            public override string Id => "TEST0002";
+            public override string Name => "TEST2";
+            public override FailureLevel DefaultLevel => FailureLevel.Note;
+            public override MultiformatMessageString FullDescription => new MultiformatMessageString { Text = null };
```
lgolding, line 54: >ext = null [](start = 104, length = 10)

`null` isn't a valid value for `Text` (although the SDK incorrectly accepts it). Just use `FullDescription => null;`,

lgolding: (similarly both the `ShortDescription` and the `FullDescription` in `TestRule3`.)

---
In reply to: [493032568](https://github.com/microsoft/sarif-sdk/pull/2082#discussion_r493032568) [](ancestors = 493032568)

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
lgolding, line 51: >Text [](start = 50, length = 4)

Prefer Markdown description to text if present. So IMO it should be
```
rule.FullDescription.Markdown ?? rule.FullDescription.Text ?? rule.ShortDescription.Markdown ?? rule.ShortDescription.Text
```

lgolding: And need test cases for this.

---
In reply to: [493035978](https://github.com/microsoft/sarif-sdk/pull/2082#discussion_r493035978) [](ancestors = 493035978)

src/Test.UnitTests.Sarif.Multitool.Library/ExportValidationRulesDocumentationCommandTests.cs
```diff
@@ -14,18 +14,27 @@ public class ExportValidationRulesDocumentationCommandTests
         [Fact]
         public void BuildRule_GeneratesExpectedMarkdown()
         {
+            var tests = new Dictionary<string, SarifValidationSkimmerBase>
+            {
+                { "Test1.md", new TestRule1() },
+                { "Test2.md", new TestRule2() },
+                { "Test3.md", new TestRule3() }
```
lgolding, line 21: >Test3 [](start = 19, length = 5)

Use descriptive test file names like "FullMarkdownRuleDescription.md", "NoDescription.md", "StandardMessageStringKey.md", "NonStandardMessageStringKey.md", and so on.

Also: change only one thing in each test. For Test2 is different from Test1 in _two_ ways: it changes from full description to short description, and from a standard message string key to a non-standard key. Let's have separate test cases for each "difference".

lgolding (changes requested):
:clock1:

src/Sarif.Driver/Sdk/ExportRulesDocumentationCommandBase.cs
lgolding: Should be a resource string with that resource identifier (not a local const string).

---
In reply to: [493028111](https://github.com/microsoft/sarif-sdk/pull/2082#discussion_r493028111) [](ancestors = 493028111)

lgolding (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #2093, opened 2020-09-29
Add explicit reference to newtonsoft. Update to 11.0.2. Update to minimist 1.2.3

michaelcfanning:
@cfaucon @jeffersonking


cfaucon (approved):
If removing FluentAssertions reference is expected in addition to the newtonsoft update, lgtm.

PR #2095, opened 2020-09-30
Fix #2090: Validator should warn of leading / in relative artifact location URIs

### LINKED ISSUE #2090: Validator should warn of leading / in relative artifact location URIs
The rationale for requiring `artifactLocation.uri` to be a relative reference when `uriBaseId` is present is that we should be able to _combine_ `uri` with any `file`-based URI to form an absolute path. `/dir/file.c` is technically a relative URI reference, but it is an absolute _file path_.  That means you _cannot_ combine it with a base value such as `file:///C:/dev/myProject/`.

The `validate` command should warn of  `/`s on `artifactLocation.uri`. It doesn't even matter whether or not `artifactLocation.uriBaseId` is present: if `artifactLocation.uri` is _not absolute_, the validator should warn of leading `/`.

The spec should have prohibit this. I filed oasis-tcs/sarif-spec#480, "Don't allow leading / in relative artifact location URIs."

This should probably go in existing rule `SARIF1004.UseUriBaseIds`. We don't have any rule that's a better home, and it's too granular to put it in a rule of its own.

@michaelcfanning FYI

lgolding:


src/Sarif.Multitool.Library/Rules/SARIF1004.ExpressUriBaseIdsCorrectly.cs
```diff
@@ -55,6 +62,16 @@ protected override void Analyze(ArtifactLocation artifactLocation, string artifa
                     artifactLocation.UriBaseId,
                     artifactLocation.Uri.OriginalString);
             }
+
+            if (!artifactLocationUri.IsAbsoluteUri && artifactLocationUri.OriginalString.StartsWith("/") == true)
```
eddynaka, line 66: I personally don't like to compare to == true if the method returns a boolean. Not sure if we have this pattern in this repo

lgolding: Good catch. That is left over from when I was using a nullable bool. Fixed.

---
In reply to: [497694414](https://github.com/microsoft/sarif-sdk/pull/2095#discussion_r497694414) [](ancestors = 497694414)

eddynaka (approved):

PR #2096, opened 2020-09-30
Fix #2089: GitHub policy should not turn off any note level rules

### LINKED ISSUE #2089: GitHub policy should not turn off any note level rules
Fix the GitHub configuration files (XML and .sarif) so they only turn _on_ the GitHub rules; they doesn't turn _off_ any other rules.

@michaelcfanning @jeffersonking FYI

lgolding:
The policy:
- Enables the `GH` rules.
- Elevates two `SARIF` rules that are more important in the context of GitHub ingestion
- Does disable one `SARIF` rule because one of the `GH` rules is actually stricter, and if both were enabled the same condition would be reported twice.

@michaelcfanning We are still going to make a "quick fix" to the web site. Approving this PR and publishing an SDK that contains this change is _not_ a prerequisite for the quick fix. But since I already had this PR started I decided to raise it.


michaelcfanning (approved):
:shipit:

PR #2097, opened 2020-10-04
m.me/hostmeverification

JuriBurakov:


michaelcfanning:
@JuriBurakov, how are you? This is an interesting PR you've proposed. Can you provide some more background/context on the change?

On surface, this looks like a bundle of actions that run a variety of tools, the output of which is intended to be published to the GH Advanced Security code scanning service. I was not aware that this full suite of tools had sufficient SARIF support to enable this, if so.

PR #2102, opened 2020-10-08
Fixing index out of range in baseliner

### LINKED ISSUE #2101: Baselining crash: index out of range on rules array
The problem is that the "context run" for an unchanged or updated result is the original run, which isn't guaranteed to have the same number or order of rules as the new run. So you can't use the ruleIndex from the result, which comes from the more recent run (right? confirm) to index into the rules array. If you're lucky, there are fewer rules in the previous run than in the current run -- because the previous run has already been baselined, and so has only a subset of the full rule set -- and the result has a rule index that's off the end -- and you'll crash with an index out of range exception. If you're unlucky, the index will happen to be in range of the smaller rules array from the previous run, and you won't notice the problem.

eddynaka:
Fixes #2101

I saw two options based on the current code:
1. update the line:
`Run contextRun = (result.BaselineState == BaselineState.Unchanged || result.BaselineState == BaselineState.Updated) ? resultPair.PreviousResult.OriginalRun : resultPair.Run;`
2. fill result.Run and reuse the variable from class.

Going for two, we could see when the Run was filled and what Run was always using the previous vs current.


michaelcfanning:
Hm. I think this might work. The best next thing to do, though, would be to author a new test that first breaks because of the bug. Then you can validate that your change fixes it. I need to look at the run merge visitor a bit more to comment completely but can't do that right now (will tomorrow).

eddynaka:
> Hm. I think this might work. The best next thing to do, though, would be to author a new test that first breaks because of the bug. Then you can validate that your change fixes it. I need to look at the run merge visitor a bit more to comment completely but can't do that right now (will tomorrow).

I see. I did using the other repository. I will check how are we testing it and try to create one specific for that.

eddynaka:
> > Hm. I think this might work. The best next thing to do, though, would be to author a new test that first breaks because of the bug. Then you can validate that your change fixes it. I need to look at the run merge visitor a bit more to comment completely but can't do that right now (will tomorrow).
> 
> I see. I did using the other repository. I will check how are we testing it and try to create one specific for that.

I created a simple test. Let me know what do you think.

src/Test.UnitTests.Sarif.Multitool/ResultMatchingCommandTests.cs
```diff
@@ -0,0 +1,172 @@
...
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Baseline.ResultMatching;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Multitool
+{
+    public class ResultMatchingCommandTests
```
michaelcfanning, line 11: >ResultMatchingCommandTests [](start = 17, length = 26)

This is a good start. Some ideas: first, make sure and use (and build out if required) our standard test helpers for creating testable log files, see src\Test.Utilities.Sarif\TestData.cs. e.g., CreateSimpleLog(). Next, build out these tests to cover all possible states, absent, new, unchanged, etc. Minimal tests are fine. Finally, be sure to create a test that explicitly covers the bug you've fixed. Don't use any real world MS repro content for this, instead, check in constructed files that provoke the issue. 

btw - are these tests in the right place? looking at the content, this test files s/be testing the high level match commands, but instead, they are testing the lower level functionality. Did you look at the content in src\sarif\src\Test.UnitTests.Sarif\Baseline\ResultMatching? These tests likely already have basics of new, unchanged, etc. So maybe we can drop this file and go add new cases to that other content?

michaelcfanning:
> using FluentAssertions;

this file is missing our standard namespace banner. also, you need to change you VS settings to put System directives first (which is our standard). Tools Menu, Options –> Text Editor –-> C# –> Advanced. In the Advance options select the “Place 'System' directives first when sorting usings” check box. T

---
Refers to: src/Test.UnitTests.Sarif.Multitool/ResultMatchingCommandTests.cs:1 in 012ae40. [](commit_id = 012ae40477f3539b601711b748ab188ee283dca9, deletion_comment = False)

src/Test.UnitTests.Sarif.Multitool/ResultMatchingCommandTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        [Fact]
+        public void ResultMatchingCommand_TestingSameResult()
+        {
+            SarifLog baselineSarif, currentSarif;
+
+            #region .: Arrange :.
```
michaelcfanning, line 18: >region [](start = 13, length = 6)

We tend to avoid #region use such as this in the code.

src/Test.UnitTests.Sarif.Multitool/ResultMatchingCommandTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            baselineSarif = CreateBaseline();
+            currentSarif = CreateCurrent();
+
+            #endregion
+
+            ISarifLogMatcher matcher = ResultMatchingBaselinerFactory.GetDefaultResultMatchingBaseliner();
+            SarifLog output = matcher.Match(new SarifLog[] { baselineSarif }, new SarifLog[] { currentSarif }).First();
+
+            output.Runs[0].Results[0].BaselineState.Should().Be(BaselineState.Unchanged);
```
michaelcfanning, line 46: >Unchanged [](start = 78, length = 9)

This test doesn't actually appear to be verifying the 'new' state. But! As mentioned, I think the other test content does, just noting this for completeness. :)

eddynaka: just added more tests. Now we have absent/unchanged/new

---
In reply to: [502494403](https://github.com/microsoft/sarif-sdk/pull/2102#discussion_r502494403) [](ancestors = 502494403)

src/Test.UnitTests.Sarif.Multitool/ResultMatchingCommandTests.cs
eddynaka: Just added more sarifs in TestData and moved the test itself to the same file we were using for baselining.

---
In reply to: [502492851](https://github.com/microsoft/sarif-sdk/pull/2102#discussion_r502492851) [](ancestors = 502492851)

eddynaka:
sure. so, I'm adding a sample test here and the real one there, just like the others.

---
In reply to: [705863548](https://github.com/microsoft/sarif-sdk/pull/2102#issuecomment-705863548) [](ancestors = 705863548)

michaelcfanning (approved):
:shipit:

PR #2107, opened 2020-10-12
Checking PATH environment variable

### LINKED ISSUE #2106: GitHelper: Search for git.exe on PATH if not present in default location

eddynaka:
Fixes #2106

### DIFF
--- a/src/Sarif/FileSearcherHelper.cs
+++ b/src/Sarif/FileSearcherHelper.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal static class FileSearcherHelper
+    {
+        /// <summary>
+        /// This method will search in the environment variable for a specific file name.
+        /// It will return the first file found.
+        /// </summary>
+        /// <param name="environmentVariable">Environment variable that we will look for</param>
+        /// <param name="fileName">Name of the file that we will look for in the environment variable</param>
+        /// <returns>Path to the file name or empty string.</returns>
+        public static string SearchForFileInEnvironmentVariable(string environmentVariable, string fileName)
+        {
+            string variable = Environment.GetEnvironmentVariable(environmentVariable);
+            if (string.IsNullOrEmpty(variable))
+            {
+                return string.Empty;
+            }
+
+            string[] paths = variable.Split(';');
+            foreach (string path in paths)
+            {
+                string returnedPath = SearchForFileNameInPath(path, fileName);
+                if (!string.IsNullOrEmpty(returnedPath))
+                {
+                    return returnedPath;
+                }
+            }
+
+            return string.Empty;
+        }
+
+        /// <summary>
+        /// This method will search for a file name in a specific path.
+        /// </summary>
+        /// <param name="path">Path where it will search.</param>
+        /// <param name="fileName">Name of the file that it will search</param>
+        /// <returns>Path to the file name or empty string.</returns>
+        public static string SearchForFileNameInPath(string path, string fileName)
+        {
+            string filePath = $@"{path}\{fileName}";
+            return File.Exists(filePath) ? filePath : string.Empty;
+        }
+    }
+}
--- a/src/Sarif/FileSearcherHelper.cs
+++ b/src/Sarif/FileSearcherHelper.cs
@@ -20,7 +20,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             string variable = Environment.GetEnvironmentVariable(environmentVariable);
             if (string.IsNullOrEmpty(variable))
             {
-                return string.Empty;
+                return null;
             }
 
             string[] paths = variable.Split(';');
@@ -33,7 +33,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 }
             }
 
-            return string.Empty;
+            return null;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Sarif/GitHelper.cs
```diff
@@ -104,6 +112,40 @@ private string GetSimpleGitCommandOutput(string repositoryPath, string args)
                 .Replace("\r", string.Empty)
                 .Replace("\n", string.Empty);
 
+        private string GetGitFromPath()
```
michaelcfanning, line 115: >GetGitFromPath [](start = 23, length = 14)

Suggestions for you:

1. Factor out a general helper that retrieves the full path for a binary (by consulting the %%PATH% variable. Make this helper static and mark it as internal so that you can unit test it.
2. Create unit tests for this helper in GitHelperTests. The unit tests can check for a file path the we expect to be available in any Windows test environment, such as  ntdll.dll.
3. Create another test for GetGitFromPath() and use a mocked file system to verify the path is retrieved.
4. I suggest removing all the GIT-specific optimizations in your code, for example, looking for Git in the path or for other evidence you are in the git.exe install directory. Git.exe could be copied anywhere.
5. For completeness, I will note that the system might find multiple locations of the same file name, after processing all the directories in the %PATH% var. In our case, though, it's fine to break on the first occurrence and to use it. You might want to comment the helper's behavior in this regard, though.

src/Sarif/FileSearcherHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        /// <param name="environmentVariable">Environment variable that we will look for</param>
+        /// <param name="fileName">Name of the file that we will look for in the environment variable</param>
+        /// <returns>Path to the file name or empty string.</returns>
+        public static string SearchForFileInEnvironmentVariable(string environmentVariable, string fileName)
+        {
+            string variable = Environment.GetEnvironmentVariable(environmentVariable);
+            if (string.IsNullOrEmpty(variable))
+            {
+                return string.Empty;
```
michaelcfanning, line 23: >string.Empty [](start = 23, length = 12)

let's return null. it's a value that's more likely to result in a crash if someone is mishandling the negative condition.

src/Sarif/FileSearcherHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+            foreach (string path in paths)
+            {
+                string returnedPath = SearchForFileNameInPath(path, fileName);
+                if (!string.IsNullOrEmpty(returnedPath))
+                {
+                    return returnedPath;
+                }
+            }
+
+            return string.Empty;
```
michaelcfanning, line 36: >Empty [](start = 26, length = 5)

here, too.

src/Sarif/GitHelper.cs
```diff
@@ -73,8 +73,16 @@ public void Checkout(string repoPath, string commitSha)
...
-            => this.fileSystem.FileExists(s_expectedGitExePath) ? s_expectedGitExePath : null;
+        internal string GetGitExePath()
+        {
+            if (this.fileSystem.FileExists(s_expectedGitExePath))
+            {
+                return s_expectedGitExePath;
+            }
+
+            string path = FileSearcherHelper.SearchForFileInEnvironmentVariable("PATH", "git.exe");
+            return !string.IsNullOrEmpty(path) ? path : null;
```
michaelcfanning, line 84: >!string.IsNullOrEmpty(path) ? path : null [](start = 19, length = 41)

return path;

src/Test.UnitTests.Sarif/GitHelperTests.cs
```diff
@@ -117,5 +117,48 @@ public void GetRepositoryRoot_WhenCalledOnTheDefaultInstanceWithCachingDisabled_
...
+        [Fact]
+        public void SearchForFileInEnvironmentVariable_WhenVariableDoesNotExist()
+        {
+            FileSearcherHelper.SearchForFileInEnvironmentVariable("PATH_THAT_DOES_NOT_EXIST", "filename.exe").Should().BeEmpty();
+        }
+
+        [Fact]
+        public void SearchForFileInEnvironmentVariable_WhenVariableExistsButFileDoesnt()
+        {
+            // proposital error in ntdll name
```
michaelcfanning, line 154: >proposital [](start = 15, length = 10)

// The error in the ntdll name here is intentional.

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif/GitHelperTests.cs
```diff
@@ -118,6 +118,49 @@ public void GetRepositoryRoot_WhenCalledOnTheDefaultInstanceWithCachingDisabled_
...
+        public void SearchForFileInEnvironmentVariable_WhenVariableExistsButFileDoesnt()
+        {
+            // The error in the ntdll name here is intentional.
+            FileSearcherHelper.SearchForFileInEnvironmentVariable("PATH", "ntdll.dlll").Should().BeNull();
+        }
+
+        [Fact]
+        public void SearchForFileInEnvironmentVariable_WhenVariableAndFileExists()
+        {
+            FileSearcherHelper.SearchForFileInEnvironmentVariable("PATH", "ntdll.dll").Should().NotBeNull();
```
lgolding, line 161: >ntdll.dll [](start = 75, length = 9)

This test will fail on non-Windows platforms. @michaelcfaning @eddynaka

lgolding: I filed [https://github.com/microsoft/sarif-sdk/issues/2113](https://github.com/microsoft/sarif-sdk/issues/2113): Unit test fails on non-Windows platforms.

---
In reply to: [505655086](https://github.com/microsoft/sarif-sdk/pull/2107#discussion_r505655086) [](ancestors = 505655086)

PR #2109, opened 2020-10-13
Compute and Apply policies

### LINKED ISSUE #2094: Provide an API to apply policy to a results list
The API would both filter the list for rules that are disabled, and adjust level.

@michaelcfanning FYI

eddynaka:
Fixes #2094


eddynaka:
@michaelcfanning @cfaucon , created a draft pr so we can discuss this approach.

The idea:
- compute: that will generate the dictionary that would apply the policies later
- apply: based on the computed policies, it would apply

What is still missing:
- enum to check what will happen if we have two policies that contains the same rule
- tests to validate that this would work.

What do you think?

src/Sarif/Core/Run.cs
```diff
@@ -217,5 +217,50 @@ public bool ShouldSerializeColumnKind()
...
+            }
+
+            PoliciesCache.Clear();
+            foreach (ToolComponent policy in this.Policies)
+            {
+                foreach (ReportingDescriptor rule in policy.Rules)
+                {
+                    if (PoliciesCache.ContainsKey(rule.Id))
+                    {
+                        PoliciesCache[rule.Id] = rule.DefaultConfiguration.Level;
```
michaelcfanning, line 238: >PoliciesCache[rule.Id] = rule.DefaultConfiguration.Level; [](start = 24, length = 57)

you only need this line of, remove other conditionals

src/Sarif/Core/Run.cs
```diff
@@ -217,5 +217,50 @@ public bool ShouldSerializeColumnKind()
         public bool ShouldSerializeLogicalLocations() { return this.LogicalLocations.HasAtLeastOneNonDefaultValue(LogicalLocation.ValueComparer); }
 
         public bool ShouldSerializeNewlineSequences() { return this.NewlineSequences.HasAtLeastOneNonNullValue(); }
+
+        private static readonly Dictionary<string, FailureLevel> PoliciesCache = new Dictionary<string, FailureLevel>();
+
+        public void ComputePolicies(/*enum?*/)
```
michaelcfanning, line 223: >public [](start = 8, length = 6)

make this an internal static for testing, make it accept an enumerable policies collection

src/Sarif/Core/Run.cs
```diff
@@ -217,5 +219,41 @@ public bool ShouldSerializeColumnKind()
         public bool ShouldSerializeLogicalLocations() { return this.LogicalLocations.HasAtLeastOneNonDefaultValue(LogicalLocation.ValueComparer); }
 
         public bool ShouldSerializeNewlineSequences() { return this.NewlineSequences.HasAtLeastOneNonNullValue(); }
+
+        internal static void ComputePolicies(IEnumerable<ToolComponent> policies)
```
michaelcfanning, line 223: >ComputePolicies [](start = 29, length = 15)

Make this API return a dictionary (the cache), then the caller assigns it to PoliciesCache

src/Sarif/Core/Run.cs
```diff
@@ -217,5 +219,41 @@ public bool ShouldSerializeColumnKind()
...
+            foreach (ToolComponent policy in policies)
+            {
+                foreach (ReportingDescriptor rule in policy.Rules)
+                {
+                    PoliciesCache[rule.Id] = rule.DefaultConfiguration.Level;
+                }
+            }
+        }
+
+        public void ApplyPolicies()
```
michaelcfanning, line 241: >ApplyPolicies [](start = 20, length = 13)

Let's add a doc comment to this. 'Applies the policies contained in this run, if any, to remap result failure levels. When multiple policies remap the same rule, the last policy in the policies collection has precedence.'

michaelcfanning:
let's add a SarifLog helper, ApplyPolicies that iterates over all embedded runs and calls Run.ApplyPolicies on each.

PR #2110, opened 2020-10-13
Adding setter to GitExePath

### LINKED ISSUE #2105: GitHelper: Allow client to explicitly set git.exe location

eddynaka:
Fixes #2105

### DIFF
--- a/src/Sarif/GitHelper.cs
+++ b/src/Sarif/GitHelper.cs
@@ -46,7 +46,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             GitExePath = GetGitExePath();
         }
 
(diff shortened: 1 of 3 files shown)


src/Sarif/GitHelper.cs
```diff
@@ -46,7 +46,7 @@ public GitHelper(IFileSystem fileSystem = null, ProcessRunner processRunner = nu
             GitExePath = GetGitExePath();
         }
 
-        public string GitExePath { get; }
+        public string GitExePath { get; set; }
```
michaelcfanning, line 49: >set [](start = 40, length = 3)

Add a test that validates this.

PR #2111, opened 2020-10-14
Simplifying tests and adding fix in changelog

eddynaka:
Refs #2102


src/Test.Utilities.Sarif/TestData.cs
```diff
@@ -258,183 +258,52 @@ public static SarifLog CreateEmptyRun()
             };
         }
 
-        public static SarifLog CreateBaseline()
+        public static SarifLog CreateSimpleLogWithRules(int startIndex, int count)
```
michaelcfanning, line 261: >startIndex [](start = 60, length = 10)

startIndex -> ruleIdStartIndex
count -> resultCount

PR #2115, opened 2020-10-15
Creating FileSystem singleton

### LINKED ISSUE #2114: Convert all API 'new FileSystem()' utilization to consume a static instance
We have patterns in code where API consult a parameter (which may be a test mock) and otherwise instantiate an instance of FileSystem() to perform work.

Rather than instantiating this class for each method call, we should have a singleton property (named `Instance`) that we use instead. The default file system capability is effectively a static helper class.

After this work, patterns like so:

fileSystem ??= new FileSystem();

will look like this:

fileSystem ??= FileSystem.Instance;

@eddynaka

eddynaka:
Fixes #2114

### DIFF
--- a/src/Sarif.Multitool.Library/ConvertCommand.cs
+++ b/src/Sarif.Multitool.Library/ConvertCommand.cs
@@ -19,7 +19,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
     {
         public int Run(ConvertOptions convertOptions, IFileSystem fileSystem = null)
         {
-            if (fileSystem == null) { fileSystem = FileSystem.Instance; }
+            fileSystem ??= FileSystem.Instance;
 
             try
             {
--- a/src/Sarif.Multitool.Library/ConvertCommand.cs
+++ b/src/Sarif.Multitool.Library/ConvertCommand.cs
@@ -19,7 +19,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
     {
         public int Run(ConvertOptions convertOptions, IFileSystem fileSystem = null)
         {
-            if (fileSystem == null) { fileSystem = new FileSystem(); }
+            if (fileSystem == null) { fileSystem = FileSystem.Instance; }
 
             try
             {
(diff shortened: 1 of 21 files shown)


src/Sarif/HashUtilities.cs
```diff
@@ -20,7 +20,7 @@ internal static IFileSystem FileSystem
         {
             get
             {
-                _fileSystem = _fileSystem ?? new FileSystem();
+                _fileSystem ??= new FileSystem();
```
lgolding, line 23: >new FileSystem() [](start = 32, length = 16)

`FileSystem.Instance`

src/Sarif.Multitool.Library/ConvertCommand.cs
```diff
@@ -19,7 +19,7 @@ public class ConvertCommand : CommandBase
     {
         public int Run(ConvertOptions convertOptions, IFileSystem fileSystem = null)
         {
-            if (fileSystem == null) { fileSystem = new FileSystem(); }
+            if (fileSystem == null) { fileSystem = FileSystem.Instance; }
```
lgolding, line 22: >if  [](start = 12, length = 3)

`fileSystem ??= FileSystem.Instance;`

src/Sarif/HashUtilities.cs
eddynaka: this is complicated...since this class has FileSystem property, i couldn't replace with FileSystem.Instance. Should  I rename the property?

---
In reply to: [505677459](https://github.com/microsoft/sarif-sdk/pull/2115#discussion_r505677459) [](ancestors = 505677459)

lgolding: You can namespace it: `_fileSystem ??= Microsoft.CodeAnalysis.FileSystem.Instance;`

---
In reply to: [505680343](https://github.com/microsoft/sarif-sdk/pull/2115#discussion_r505680343) [](ancestors = 505680343,505677459)

lgolding (approved):
:shipit:

src/Sarif/HashUtilities.cs
michaelcfanning: I would have removed this property and just marked the underlying private static as 'internal'. :)

---
In reply to: [505681908](https://github.com/microsoft/sarif-sdk/pull/2115#discussion_r505681908) [](ancestors = 505681908,505680343,505677459)

michaelcfanning (approved):
:shipit:

PR #2116, opened 2020-10-15
Enforcing file normalization

eddynaka:
This PR enforces the following rules for all files except autogenerated folder:
- insert_final_newline = true
- dotnet_separate_import_directive_groups = true
- dotnet_sort_system_directives_first = true
- file_header_template = `Copyright (c) Microsoft. All rights reserved.\nLicensed under the MIT license. See LICENSE file in the project root for full license information.`

Using dotnet-format tool to update all files.


michaelcfanning:
Eddy, nice change but you need to revert all the code in src\Sarif\Autogenerated. This code is, erm, auto-generated and will be overwritten at next build. To fix issues in this code, you have to go fix the Jschema schema -> C# OM code to correct the formatting issue. [https://github.com/microsoft/jschema](https://github.com/microsoft/jschema). Also @lgolding to give you a leg up in making the fix and we'll take it. :)

michaelcfanning (changes requested):
:clock1:

eddynaka:
> Eddy, nice change but you need to revert all the code in src\Sarif\Autogenerated. This code is, erm, auto-generated and will be overwritten at next build. To fix issues in this code, you have to go fix the Jschema schema -> C# OM code to correct the formatting issue. https://github.com/microsoft/jschema. Also @lgolding to give you a leg up in making the fix and we'll take it. :)

Besides that, I was planning to do one other thing: the header normalization. I saw that we have two types one saying that is MIT and one saying that is Microsoft reserved.

Should I include here or create a separate pr? It's near 2** files

eddynaka:
> > Eddy, nice change but you need to revert all the code in src\Sarif\Autogenerated. This code is, erm, auto-generated and will be overwritten at next build. To fix issues in this code, you have to go fix the Jschema schema -> C# OM code to correct the formatting issue. https://github.com/microsoft/jschema. Also @lgolding to give you a leg up in making the fix and we'll take it. :)
> 
> Besides that, I was planning to do one other thing: the header normalization. I saw that we have two types one saying that is MIT and one saying that is Microsoft reserved.
> 
> Should I include here or create a separate pr? It's near 2** files

Just updated to include...

they are almost the same files.

michaelcfanning:
MIT template is correct. Let's use a separate PR.

---
In reply to: [709686739](https://github.com/microsoft/sarif-sdk/pull/2116#issuecomment-709686739) [](ancestors = 709686739)

michaelcfanning (approved):
:shipit:

PR #2119, opened 2020-10-16
Fixing xml when that doesn't contain location

### LINKED ISSUE #2103: Sarif.Multitool can't read CppCheck XML version 2
I'm trying to use Sarif.Multitool to convert CppCheck output to SARIF. Unfortunately the expected XML structure doesn't match what current CppCheck versions produce, presumably because they now exclusively produce the CppCheck XML format version 2.

Example code:

```c
#include <stdio.h>
#include <sys/socket.h>

int main()
{
	int i;
	printf("%d\n", i);
}
```

Analyzing this code with the command `cppcheck --enable=all --xml` leads to the following output:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<results version="2">
    <cppcheck version="2.1"/>
    <errors>
        <error id="uninitvar" severity="error" msg="Uninitialized variable: i" verbose="Uninitialized variable: i" cwe="457">
            <location file="/test/test.c" line="7" column="17"/>
            <symbol>i</symbol>
        </error>
        <error id="missingIncludeSystem" severity="information" msg="Cppcheck cannot find all the include files (use --check-config for details)" verbose="Cppcheck cannot find all the include files. Cppcheck can check the code without the include files found. But the results will probably be more accurate if all the include files are found. Please check your project&apos;s include directories and add all of them as include directories for Cppcheck. To see what files Cppcheck cannot find use --check-config."/>
    </errors>
</results>
```

Both the `<symbol>` element and the `<error>` element without child elements (the `missingIncludeSystem` one) cause exceptions in Sarif.Multitool when trying to convert the file:
```
System.Xml.XmlException: Could not parse CppCheck location node; element name was not 'location'. Line 7, position 14.
```
And after removing the `<symbol>` element (causing the line number to shift):
```
System.Xml.XmlException: A CppCheck entry must have a nonzero number of locations. Line 9, position 7.
```
After removing the second `<error>` element as well the XML can be converted.

I'd very much appreciate if you could update the CppCheck XML parser. :slightly_smiling_face:

eddynaka:
Fixes #2103


eddynaka:
@michaelcfanning @lgolding , I was looking at the issue that we have a problem converting the following xml from cppCheck:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<results version="2">
    <cppcheck version="2.1"/>
    <errors>
        <error id="uninitvar" severity="error" msg="Uninitialized variable: i" verbose="Uninitialized variable: i" cwe="457">
            <location file="/test/test.c" line="7" column="17"/>
            <symbol>i</symbol>
        </error>
        <error id="missingIncludeSystem" severity="information" msg="Cppcheck cannot find all the include files (use --check-config for details)" verbose="Cppcheck cannot find all the include files. Cppcheck can check the code without the include files found. But the results will probably be more accurate if all the include files are found. Please check your project&apos;s include directories and add all of them as include directories for Cppcheck. To see what files Cppcheck cannot find use --check-config."/>
    </errors>
</results>
```

When debugging I saw some problems:
1. the actual code is not prepared if it finds a tag that is different than `location`. So, it would throw an exception because we have a `symbol` tag
2. the actual code is not prepared if an error does not contain locations.

With this changes (besides the tests), it already solves the current issue.

Questions:
- Do you know why do we have those specific checks?
- Can I remove those specific conditions and make it work?

Thank you!

src/Sarif.Converters/CppCheckError.cs
```diff
@@ -31,11 +31,6 @@ internal class CppCheckError
         /// <param name="locations">The locations to which the error generated by CppCheck refer.</param>
         public CppCheckError(string id, string message, string verboseMessage, string severity, ImmutableArray<CppCheckLocation> locations)
         {
-            if (locations.IsDefaultOrEmpty)
-            {
-                throw new ArgumentException(ConverterResources.CppCheckMissingLocation);
```
lgolding, line 36: >CppCheckMissingLocation [](start = 63, length = 23)

If this resource is now unused, remove it.

lgolding:
Need unit tests for the changed functionality.

lgolding:
I don't remember why those checks are there. If there is real CppCheck output out there that violates these checks, we just need to remove them.

---
In reply to: [710041033](https://github.com/microsoft/sarif-sdk/pull/2119#issuecomment-710041033) [](ancestors = 710041033)

eddynaka:
> Need unit tests for the changed functionality.

yes, i will have to take a look, since they are failing. The overall idea was to check the behavior before changing everything.

thanks for reviewing!

src/Sarif.Converters/CppCheckError.cs
```diff
@@ -111,11 +106,6 @@ public static CppCheckError Parse(XmlReader reader, CppCheckStrings strings)
         /// <returns>This instance as an <see cref="Result"/>.</returns>
         public Result ToSarifIssue()
         {
-            if (this.Locations.Length == 0)
-            {
-                throw new InvalidOperationException("At least one location must be present in a SARIF result.");
```
lgolding, line 116: >location [](start = 66, length = 8)

This check was definitely wrong. It is absolutely not required for a SARIF `result` to have a `locations` property.

lgolding:
>             else

Do we need the special-case "length == 1" branch? It looks like the `else` branch works fine for that case.

---
Refers to: src/Sarif.Converters/CppCheckError.cs:134 in 1e15093. [](commit_id = 1e15093be20d0d963dec62065dddb833e778ebc7, deletion_comment = False)

src/Sarif.Converters/CppCheckError.cs
```diff
@@ -133,7 +123,11 @@ public Result ToSarifIssue()
             }
 
             PhysicalLocation lastLocationConverted;
-            if (this.Locations.Length == 1)
+            if (this.Locations.Length == 0)
+            {
+                return result;
```
lgolding, line 128: >result [](start = 23, length = 6)

This "early out" is a little hard to spot. I suggest restructuring to:
```
if (this.Locations.Length > 0)
{
    // Do all the location-related stuff.
}
```

Then for bonus points, because this is a long method, I would extract a helper method
```
AddLocationsToResult(IEnumerable<CppCheckLocation> cppCheckLocations, Result result)
```

Which would contain everything between the braces.

lgolding (changes requested):
:clock1:

src/Sarif.Converters/CppCheckError.cs
```diff
@@ -132,49 +123,49 @@ public Result ToSarifIssue()
...
+                    Importance = ThreadFlowLocationImportance.Essential
+                });
             }
 
+            result.CodeFlows = new List<CodeFlow>()
+            {
+                SarifUtilities.CreateSingleThreadedCodeFlow(locations)
+            };
+
+            // In the N != 1 case, set the overall location's location to
```
lgolding, line 159: >In the N != 1 case, s [](start = 15, length = 21)

Let's fix this existing comment:

- You do this in call cases now, not just N != 1.
- It's the result's location not the "overall location's location".
- The SARIF term is "code flow", not "execution flow".

So all together, change the comment to just:

// Set the result's location to the last location in the code flow.

src/Sarif.Converters/CppCheckError.cs
lgolding: Why is this "By Design"? Is the resource used elsewhere?

---
In reply to: [506538800](https://github.com/microsoft/sarif-sdk/pull/2119#discussion_r506538800) [](ancestors = 506538800)

eddynaka: updating!

---
In reply to: [506609059](https://github.com/microsoft/sarif-sdk/pull/2119#discussion_r506609059) [](ancestors = 506609059,506538800)

lgolding: I see you removed an unused resource CppCheckCppCheckElementMissing, which is fine. But the resource here is CppCheckMissingLocation, which you can (also) remove -- if indeed it is not used elsewhere.

---
In reply to: [506618367](https://github.com/microsoft/sarif-sdk/pull/2119#discussion_r506618367) [](ancestors = 506618367,506609059,506538800)

src/Test.UnitTests.Sarif.Converters/CppCheckErrorTests.cs
```diff
@@ -15,6 +15,10 @@ public class CppCheckErrorTests
         private readonly ImmutableArray<CppCheckLocation> _dummyLocations = ImmutableArray.Create(new CppCheckLocation("file.cpp", 42));
         private const string ExampleFileName = "example.cpp";
         private const string ExampleFileName2 = "example2.cpp";
+        private const string exampleErrorXmlBase = "<error id=\"id\" msg=\"message\" verbose=\"verbose\" severity=\"style\"";
```
lgolding, line 18: >e [](start = 29, length = 1)

Our code base convention is for `const` to be `PascalCase`.

src/Test.UnitTests.Sarif.Converters/CppCheckErrorTests.cs
```diff
@@ -187,18 +207,16 @@ public void CppCheckError_InvalidParse_BadRootNodeDetected()
         }
 
         [Fact]
-        public void CppCheckError_InvalidParse_BadChildrenNodeDetected()
+        public void CppCheckError_InvalidParse_BadChildrenNodeDetectedShouldNotThrow()
```
lgolding, line 210: >ChildrenNode [](start = 50, length = 12)

"ChildrenNode" => "ChildNode"

lgolding (approved):
:shipit:

src/Sarif.Converters/CppCheckError.cs
eddynaka: Looking again, I saw that this is used by other places. so I'm reverting it.

---
In reply to: [506620738](https://github.com/microsoft/sarif-sdk/pull/2119#discussion_r506620738) [](ancestors = 506620738,506618367,506609059,506538800)

eddynaka:
@cfaucon @michaelcfanning , if u can, can you re-run the pipeline? its showing an error, but it was the test host. so just to be sure.
Thank you!

src/Sarif.Converters/CppCheckError.cs
```diff
@@ -132,49 +123,52 @@ public Result ToSarifIssue()
...
                     SarifUtilities.CreateSingleThreadedCodeFlow(locations)
                 };
-
-                // In the N != 1 case, set the overall location's location to
-                // the last entry in the execution flow.
-                lastLocationConverted = locations[locations.Count - 1].Location.PhysicalLocation;
             }
 
+            // Set the result's location to the last location in the code flow.
+            lastLocationConverted = locations[locations.Count - 1].Location.PhysicalLocation;
```
cfaucon, line 164: The caller of AddLocationToResult already checks locations length > 0 which is why I am assuming there is no safeguards here. Should probably put in safeguards in the function itself to avoid invalid index exception for empty locations if that scenario is possible here.

cfaucon (changes requested):
One comment around potential invalid index scenario.

src/Sarif.Converters/CppCheckError.cs
eddynaka: your idea is to check if we have at least one element before in the first line of this function? So, if someone removes accidentally that condition, the code wouldn't break. Is that it?

cfaucon: Yes, my comment is that we shouldn't rely on the caller of a method to have made certain checks to make the called method safe.  In this example, we shouldn't rely on the caller of AddLocationToResult to have checked for empty before calling.  the AddLocationToResult method should be self contained in its ability to ensure it does not cause scenarios like invalid index exceptions, etc.

eddynaka: ok! let me change! :) Thanks for the review

eddynaka: just updated!

cfaucon (approved):

michaelcfanning (approved):
:shipit:

PR #2122, opened 2020-10-19
Fixing FileSpecifier for Linux/Windows environment

### LINKED ISSUE #2055: sarif-multitool validate fails with ERR997.NoValidAnalysisTargets
I installed the multitool on Linux via npm:

    npm i @microsoft/sarif-multitool

If I run `validate` on a local file, `report.sarif`:

    npx @microsoft/sarif-multitool validate report.sarif

The tool fails with an error:

```
Analyzing...
SARIF.MULTITOOL : error ERR997.NoValidAnalysisTargets : No valid analysis targets were specified.

Analysis did not complete due to one or more unrecoverable execution conditions.
Unexpected fatal runtime condition(s) observed: NoValidAnalysisTargets
```

The error message is identical if we omit `report.sarif` from the command line, which is very confusing.

Luckily, a colleague had identified the issue and was able to tell me that a `./` must be added to the filepath:

    npx @microsoft/sarif-multitool validate ./report.sarif

In this case, the tool works. But this behavior is impossible to guess, since it is contrary to command-line conventions. The `validate --help` message does *not* mention it at all, and its only description of the non-option argument is:

    value pos. 0                        One or more specifiers to a file, directory, or filter pattern that resolves to one or more binaries to analyze.

`value pos. 0` is a strange way of specifying arguments; the "specifier to a file, ..." does not mention the need to have a `./` in the path; plus the fact that "resolves to one or more binaries" is incorrect: the sarif file is a text file, not a binary. All of this leads to confusion and provides no clue to the actual issue.

In hopes that Google will help future users understand the error, or that a less surprising behavior will be implemented, I'm creating this issue.

eddynaka:
Fixes #2055


src/Test.UnitTests.Sarif.Converters/FortifyUtilitiesTests.cs
```diff
@@ -11,7 +12,7 @@ public class FortifyUtilitiesTests
         public void FortifyUtilities_ParseFormattedContentText_Correct()
         {
             string content = @"<Content><Paragraph><b>General relativity</b> is the <pre>geometric theory of gravitation</pre> published by <i>Albert Einstein</i> in 1915 and the current description of <IfDef var=""ConditionalDescriptions""><ConditionalText attr=""value"">gravitation in modern physics</ConditionalText></IfDef>.<AltParagraph>General relativity generalizes special relativity and <Replace key=""Scientist.lastName""/>'s law of universal gravitation, providing a unified description of gravity as a geometric property of space and time, or spacetime. In particular, the curvature of spacetime is directly related to the energy and momentum of whatever matter and radiation are present.</AltParagraph>The relation is specified by the <code>Einstein field equations</code>, a system of <code>partial differential equations</code>.";
-            content += "\r\nAnd here's a garbage\n\r\n \n  \n\ndouble line break!</Paragraph></Content>";
+            content += $"{Environment.NewLine}And here's a garbage\n\r\n \n  \n\ndouble line break!</Paragraph></Content>";
```
cfaucon, line 15: Do these also need to be replaced as well?

cfaucon (approved):
One comment for consideration, otherwise looks good.

src/Test.UnitTests.Sarif.Converters/FortifyUtilitiesTests.cs
eddynaka: When testing in WSL, the only part needed was the first one. Let me change and test again!

eddynaka:
> One comment for consideration, otherwise looks good.

Thanks. This PR is solving the issue and bringing our current solution one step closer to be developed in linux as well. (this is why i changed the csproj)

src/Sarif/Sarif.csproj
```diff
@@ -19,7 +19,8 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;net45</TargetFrameworks>
```
lgolding, line 22: >net461 [](start = 37, length = 6)

Removing `net45` is fine. I know @michaelcfanning has talked before about removing .NET Framework support altogether. Is that what's happening here? That he said "Go ahead and remove it"?

src/Sarif.Driver/FileSpecifier.cs
```diff
@@ -59,7 +59,7 @@ private void BuildFilesAndDirectoriesList()
 
             if (directory.Length == 0)
             {
-                directory = @".\";
+                directory = $".{Path.DirectorySeparatorChar}";
```
lgolding, line 62: >DirectorySeparatorChar [](start = 37, length = 22)

Yup.

src/Sarif/Sarif.csproj
lgolding: Oh no, sorry, I see what's going on. Never mind.

---
In reply to: [507934421](https://github.com/microsoft/sarif-sdk/pull/2122#discussion_r507934421) [](ancestors = 507934421)

src/Sarif/Sarif.csproj
```diff
@@ -19,7 +19,8 @@
...
+    <TargetFrameworks>netstandard2.0;</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net461;net45</TargetFrameworks>
```
lgolding, line 23: >OS [](start = 35, length = 2)

But wait. That means "If I'm _building_ on a non-Windows platform, then don't _build_ the .NET Framework version. Shouldn't we always build for the same set of targets, no matter _where_ we're building?

lgolding (approved):
:shipit:

src/Sarif/Sarif.csproj
eddynaka: when i was solving this issue, I couldn't find a way to install netframework on linux without using wine or other tools that simulate executable files from windows.

---
In reply to: [507937299](https://github.com/microsoft/sarif-sdk/pull/2122#discussion_r507937299) [](ancestors = 507937299)

src/Test.UnitTests.Sarif.Converters/FortifyUtilitiesTests.cs
eddynaka: just did one more changing replacing the other \r\n for the environment.newline

michaelcfanning (approved):
:shipit:

PR #2125, opened 2020-10-22
Reversion to 2.3.8.

michaelcfanning:
@eddynaka 

If this looks right, will you approve and merge?


src/Sarif/Autogenerated/Address.cs
```diff
@@ -6,7 +6,9 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Runtime.Serialization;
+
```
eddynaka, line 9: Should we add the new grouping for autogenerated files?

src/Sarif/VersionConstants.cs
```diff
@@ -8,4 +8,4 @@ public static class VersionConstants
         public const string SchemaVersionAsPublishedToSchemaStoreOrg = "2.1.0-rtm.5";
         public const string StableSarifVersion = "2.1.0";
     }
-}
+}
```
eddynaka, line 11: we should have a new line in the end. With dotnet-format, we are checking that as well
```bash
Run dotnet-format --folder --check --exclude .\src\Sarif\Autogenerated\
  Formatting code files in workspace 'D:\a\sarif-sdk\sarif-sdk'.
  sarif-sdk\src\Sarif\VersionConstants.cs(11,2): Fix final newline.
  Formatted code file 'D:\a\sarif-sdk\sarif-sdk\src\Sarif\VersionConstants.cs'.
  Format complete in 5983ms.
```

src/Sarif/Autogenerated/Run.cs
```diff
@@ -6,7 +6,9 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Runtime.Serialization;
```
michaelcfanning, line 8: >Serialization [](start = 21, length = 13)

why are we seeing these changes? isn't our dotnet-format exclusion getting honored? we can certainly update the autogenerator code to make this fix.

src/Sarif/VersionConstants.cs
michaelcfanning: It should be the following, I think. This will still work in the pipeline (where the bld and packages dirs don't exist)

Run dotnet-format --folder --check --exclude .\src\Sarif\Autogenerated\ --exclude .\src\packages\ --exclude .\bld\

---
In reply to: [510433793](https://github.com/microsoft/sarif-sdk/pull/2125#discussion_r510433793) [](ancestors = 510433793)

src/Sarif/Autogenerated/Run.cs
eddynaka: probably when you opened / saved, it separated in groups.
About the real fix, sure that would be a good idea! 

---
In reply to: [510438497](https://github.com/microsoft/sarif-sdk/pull/2125#discussion_r510438497) [](ancestors = 510438497)

src/Sarif/VersionConstants.cs
eddynaka: no problem, i can create a new pr adding those exclusions. The point here is that we need to add a new line in the end for this file :)

---
In reply to: [510439559](https://github.com/microsoft/sarif-sdk/pull/2125#discussion_r510439559) [](ancestors = 510439559,510433793)

src/Sarif/Autogenerated/Address.cs
eddynaka: already fixed

---
In reply to: [510432670](https://github.com/microsoft/sarif-sdk/pull/2125#discussion_r510432670) [](ancestors = 510432670)

src/Sarif/Autogenerated/Run.cs
eddynaka: already fixed

---
In reply to: [510440931](https://github.com/microsoft/sarif-sdk/pull/2125#discussion_r510440931) [](ancestors = 510440931,510438497)

eddynaka (approved):
:shipit:

PR #2127, opened 2020-10-23
Add check in SARIF1010 that rule is in rules setion

mveytsman:
This PR adds a check to SARIF1010 that that if `ruleId` or `rule.id` is given, then the rule MUST exist in the `rules` list.

This is required for GitHub's code scanning features to work properly, and I believe is consistent with the SARIF spec.

I was unable to get the `RuleResources.resx` file to compile properly, so some new strings need to be added in order to describe this behavior and the assocated error message. For now I used the same error message as `Error_ResultMustSpecifyRuleId` which is not exactly correct.

cc @marcogario

### DIFF
--- a/src/Sarif.Multitool.Library/Rules/SARIF1010.RuleIdMustBeConsistent.cs
+++ b/src/Sarif.Multitool.Library/Rules/SARIF1010.RuleIdMustBeConsistent.cs
@@ -26,6 +26,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
         };
         public override FailureLevel DefaultLevel => FailureLevel.Error;
 
+        private List<string> ruleIds;
+        protected override void Analyze(Run run, string resultPointer)
+        {
+            this.ruleIds = GetRuleIds(run);
+        }
+
         protected override void Analyze(Result result, string resultPointer)
         {
             AnalyzeRuleId(result, resultPointer);
@@ -58,6 +64,28 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
                     result.RuleId,
                     result.Rule?.Id);
             }
+
+            // rules in file must contain the rule id
+           else if ((!string.IsNullOrWhiteSpace(result.RuleId) && !ruleIds.Contains(result.RuleId)) || (!string.IsNullOrWhiteSpace(result.Rule?.Id) && !ruleIds.Contains(result.Rule?.Id))) {
+                // TODO: update message text 
+                LogResult(
+                    pointer,
+                    nameof(RuleResources.SARIF1010_RuleIdMustBeConsistent_Error_ResultMustSpecifyRuleId_Text));
+            }
+        }
+
+        private List<string> GetRuleIds(Run run)
+        {
+            var ruleIds = new List<string>();
+
+            if (run.Tool?.Driver?.Rules != null)
+            {
+                foreach (ReportingDescriptor rule in run.Tool.Driver.Rules)
+                {
+                    ruleIds.Add(rule.Id);
+                }
+            }
+            return ruleIds;
         }
     }
 }
--- a/src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1010.RuleIdMustBeConsistent_Invalid.sarif
+++ b/src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1010.RuleIdMustBeConsistent_Invalid.sarif
@@ -6,7 +6,15 @@
       "tool": {
         "driver": {
           "name": "Sarif Functional Testing",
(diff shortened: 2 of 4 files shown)


eddynaka:
> This PR adds a check to SARIF1010 that that if `ruleId` or `rule.id` is given, then the rule MUST exist in the `rules` list.
> 
> This is required for GitHub's code scanning features to work properly, and I believe is consistent with the SARIF spec.
> 
> I was unable to get the `RuleResources.resx` file to compile properly, so some new strings need to be added in order to describe this behavior and the assocated error message. For now I used the same error message as `Error_ResultMustSpecifyRuleId` which is not exactly correct.
> 
> cc @marcogario

Hi @mveytsman , thanks for the contribution!

I was double checking here and we might have some similar to this: `SARIF1009`, which is checking for indexes (if they are right or wrong). Can you take a look, please?

src/Sarif.Multitool.Library/Rules/SARIF1010.RuleIdMustBeConsistent.cs
```diff
@@ -26,6 +26,12 @@ public class RuleIdMustBeConsistent : SarifValidationSkimmerBase
         };
         public override FailureLevel DefaultLevel => FailureLevel.Error;
 
+        private List<string> ruleIds;
```
eddynaka, line 29: I can see some issues:
1. Imagine that the sarif you are validating has two runs. In each run you have rules and results. For the first run, we have rule01 and result01 (fine). for the second run, we can have rule02' and result01', but from a different tool, so by coincidence you are saying it's right but it isn't because the result01' should be looking for rule01'.
2. If you have multiple equal rules registered, your list will be big. Probably you should change it to a Dict<toolname,hashset>, with that, you would prevent the first issue and it would minimize the search/duplication

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1010.RuleIdMustBeConsistent_Invalid.sarif
```diff
@@ -23,6 +31,12 @@
           "message": {
             "text": "This result contains both a ruleId and a rule.id property, but they are different."
           }
+        },
+        {
+          "ruleId": "TST0003",
```
eddynaka, line 36: we normally try to create one test for each scenario.

michaelcfanning:
@mveytsman, sorry again for delay in getting to this. I think you're asking for a new validation standard here, which is: please provide rules metadata. 2010 is really only about ensuring SARIF consistency (i.e., if you provide duplicative properties, they all must match),

SARIF2012.ProvideRuleProperties would be a good place for this. We could add a new 'note' level analysis which ensures that every SARIF result lines up with a reporting descriptor. 

@eddynaka , let's talk about picking up this analysis. An obvious first question is whether we want a generalized analysis that accounts for notifications (I tend to think that we should).

@mveytsman, can you help me understand the importance of this new analysis to the GH ingestion scenario? Our default here would be to provide this analysis as a low priority message (a SARIF 'note'). For GHAS code scanning ingestion, should this be elevated to a warning (always visible to developers but non-blocking) or as an error (indicating that developers really need to fix it).

michaelcfanning:
@mveytsman, do you have time to weigh in, as per my previous comment?

mveytsman:
Hi @michaelcfanning thanks for the ping! - I no longer work at GitHub / on the code scanning team. My intention was to start this work and pass it on as I left but I did not communicate that clearly enough -- I'm ccing @marcogario and @anaarmas to continue this conversation.

marcogario:
@michaelcfanning sorry, I completely lost track of this and https://github.com/microsoft/sarif-sdk/pull/2128. Let me review this in the next few days and get back to you.

marcogario:
There are 2 issues that motivated this change.

1. A few 3rd parties do not provide rule information in their results. While they do not implement the spec, we do handle this case in Code Scanning by mapping the result to a generic "unknown-rule" rule.
2. To provide things like Tags, Help, Severity etc, Code Scanning was working under the expectation that rule information would be available for all results. More recently, however, we decided to accept this case and treat it as a "unknown rule" again (I backlinked the relevant issue). Therefore, your proposal makes a lot of sense to me:

> SARIF2012.ProvideRuleProperties would be a good place for this. We could add a new 'note' level analysis which ensures that every SARIF result lines up with a reporting descriptor.

Having this as a Note would encourage SARIF producers to actually include this data, and better leverage Code Scanning.

My proposal here would be to revert/move the changes of this PR to a check within SARIF2012.ProvideRuleProperties.

michaelcfanning:
@marcogario , got it and agree, we'll move this analysis. @eddynaka, want to port those changes over?

eddynaka:
@michaelcfanning @marcogario , since I already created a PR. I will close this one.

PR #2128, opened 2020-10-23
Relax check in GH1005

mveytsman:
This PR relaxes the check in GH1005 : If the uri is not a file, we can still determine the file from the working direction in the invocations list. Now we make sure that the if the uri isn't a file, we have a working directory.

I was unable to compile RuleResouces.resx, so the relevant strings & docs still need to be updated

cc @marcogario

### DIFF
--- a/src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
+++ b/src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
@@ -4,8 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-
+using System.Linq.Expressions;
 using Microsoft.Json.Pointer;
+using Microsoft.Win32.SafeHandles;
 
 namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
 {
@@ -28,9 +29,15 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
 
         public override bool EnabledByDefault => false;
 
+        private Uri workingDirectoryUri;
+        protected override void Analyze(Run run, string runPointer)
+        {
+            this.workingDirectoryUri = GetWorkingDirectoryUri(run);
+        }
+
         protected override void Analyze(Result result, string resultPointer)
         {
-            if (!(result.Locations?.Any() == true))
+            if (!(result.Locations?.Any() == true)) 
             {
                 // Rule GH1001.ProvideRequiredLocationProperties will catch this, so don't
                 // report it here.
@@ -63,7 +70,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
                 return;
             }
 
-            if (uri.IsAbsoluteUri && uri.Scheme != "file")
+            if (uri.IsAbsoluteUri && uri.Scheme != "file" && this.workingDirectoryUri == null)
             {
                 string uriPointer = locationPointer
                     .AtProperty(SarifPropertyName.PhysicalLocation)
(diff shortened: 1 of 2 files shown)


src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
```diff
@@ -4,8 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-
+using System.Linq.Expressions;
 using Microsoft.Json.Pointer;
+using Microsoft.Win32.SafeHandles;
```
eddynaka, line 9: do we need this using?

src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
```diff
@@ -4,8 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-
+using System.Linq.Expressions;
 using Microsoft.Json.Pointer;
```
eddynaka, line 8: we have a style configuration to enforce the using by groups. If you can check the dotnet-format check you will find some issues. Can you resolve those?

michaelcfanning:
@mveytsman, sorry for the delay in getting to this. Can you explain a bit more what pattern you want to be able to accept into GH without provoking this validation warning? I understand that the presence of a working directory uri is useful to you, but don't see how this allows for loosening the requirement that any absolute URI have a "file" scheme.

Can you clarify? I guess what I'd expect in this scenario is that, in the presence of a non-null working directory, SARIF producers would either emit an absolute file path (and the working directory could be used in comparison to figure out that non-deterministic root) or SARIF producers would simply emit a relative path.

I suspect I'm missing some specific case you have in mind based on actual GHAS code scanning ingestion. :) Can you clarify?

michaelcfanning:
@mveytsman, do you have time to weigh in, as per my previous comment?

mveytsman:
@michaelcfanning thanks again per the ping. As per https://github.com/microsoft/sarif-sdk/pull/2127#issuecomment-728999997 it's best to pass this to  @marcogario and @anaarmas

michaelcfanning:
Thanks! I will reach out to them if I don't see further comments here. :)

marcogario:
I think the change in this PR needs refining, as it might be weaker than what we wanted.
We will probably tackle any repair in 2 weeks, but would be good to agree on what changes we want to do.

> Can you explain a bit more what pattern you want to be able to accept into GH without provoking this validation warning? [...]
> I guess what I'd expect in this scenario is that, in the presence of a non-null working directory, SARIF producers would either emit an absolute file path (and the working directory could be used in comparison to figure out that non-deterministic root) or SARIF producers would simply emit a relative path.

We want to accept a result with the following `artifactLocation`:
```
   "artifactLocation": {
                    "uri": "https://github.com/github/example/blob/0000000000000000000000000000000000000000/src/promiseUtils.js",
                    "index": 1
                  },
```

Only if we are given a working directory in the form:

```
"invocations": [
          {
            "workingDirectory": {
                 "uri": https://github.com/github/example/blob/0000000000000000000000000000000000000000"
             }
          }
        ]
```

As long as Code Scanning can extract a relative path, then we are good.
I think we ran into this with a 3rd party tool, but if you still have reservations, I can try to find out which one.

marcogario:
:wave: Wondering if you had any opinion on the above. We might try to tackle it in 2021, just want to make sure the direction makes sense. Thanks! :bow:

eddynaka:
> 👋 Wondering if you had any opinion on the above. We might try to tackle it in 2021, just want to make sure the direction makes sense. Thanks! 🙇

Hi @marcogario , let's create an issue and try to draw a design?
We are having some trouble understanding the following:
- what is happening with this current rule?
- which case is failing or passing?
- what are you expecting to fail or pass?

thank you

michaelcfanning:
Just to repeat, we're ready to engage, just need more information on the pattern we want. Maybe get on a call together?

marcogario:
I created https://github.com/microsoft/sarif-sdk/issues/2197 to put down the requirements more clearly. Happy to have a call either this week or in January if there are any doubts. Thank you very much!

eddynaka:
Closing since we merged this: https://github.com/microsoft/sarif-sdk/pull/2248

PR #2130, opened 2020-10-25
Exposing public methods from FileRegionsCache as interface

### LINKED ISSUE #2129: Improve FileRegionsCache testability
I'm writing some unit tests for the VS Extension that require the use of a `FileRegionsCache` object, a concrete class that interacts with the file system. I can create, and use in my tests, a `FileRegionsCache` object with a mock `IFileSystem`. But that requires me to understand the internal workings of `FileRegionsCache` so I can mock the correct set of file system calls, and I also have to provide fake file system content to the mock.

What I want/need instead is an `IFileRegionsCache` interface that I can mock, so that when the VS Extension product code that I'm unit testing calls `fileRegionsCache.PopulateTextRegionProperties`, I can directly return the value I want.

@michaelcfanning @eddynaka @GabeDeBacker FYI

eddynaka:
Fixes #2129

### DIFF
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -32,16 +32,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             _cache = new Cache<string, Tuple<string, NewLineIndex>>(BuildIndexForFile);
         }
 
-        /// <summary>
-        /// Accepts a physical location and returns a Region object, based on the input
-        /// physicalLocation.region property, that has all its properties populated. If an
-        /// input text region, for example, only specifies the startLine property, the returned
-        /// Region instance will have computed and populated other properties, such as charOffset,
-        /// charLength, etc.
-        /// </summary>
-        /// <param name="physicalLocation">The physical location containing the region which should be populated.</param>
-        /// <param name="populateSnippet">Specifies whether the physicalLocation.region.snippet property should be populated.</param>
-        /// <returns></returns>
+        /// <inheritdoc/>
         public Region PopulateTextRegionProperties(Region inputRegion, Uri uri, bool populateSnippet)
         {
             if (inputRegion == null || inputRegion.IsBinaryRegion)
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.IO;
 
@@ -11,7 +12,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// associated with a SARIF log, and to construct text
     /// snippets associated with region instances.
     /// </summary>
-    public class FileRegionsCache
+    public class FileRegionsCache : IFileRegionsCache
     {
         public const int DefaultCacheCapacity = 100;
         private readonly Run _run;
@@ -33,10 +34,10 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         /// <summary>
         /// Accepts a physical location and returns a Region object, based on the input
-        /// physicalLocation.region property, that has all its properties populated. If an 
+        /// physicalLocation.region property, that has all its properties populated. If an
         /// input text region, for example, only specifies the startLine property, the returned
         /// Region instance will have computed and populated other properties, such as charOffset,
-        /// charLength, etc. 
+        /// charLength, etc.
         /// </summary>
         /// <param name="physicalLocation">The physical location containing the region which should be populated.</param>
         /// <param name="populateSnippet">Specifies whether the physicalLocation.region.snippet property should be populated.</param>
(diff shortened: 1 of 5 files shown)


src/Sarif/FileRegionsCache.cs
```diff
@@ -33,10 +34,10 @@ public FileRegionsCache(Run run, int capacity = DefaultCacheCapacity, IFileSyste
 
         /// <summary>
```
lgolding, line 35: >/// [](start = 8, length = 3)

Suggestion: Place the XML doc comments on the interface methods, and mark the implementations with `/// <inheritdoc/>`.

lgolding (approved):
:shipit:

lgolding:
Don't forget to bump version to `2.3.8` and update release history. Please publish at your earliest convenience. I might ask for additional publishes this week as I continue to add tests to previously untested portions of the VS viewer.

src/Sarif/FileRegionsCache.cs
eddynaka: just updated!

---
In reply to: [511616376](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511616376) [](ancestors = 511616376)

eddynaka:
just updated the release notes!

---
In reply to: [716172085](https://github.com/microsoft/sarif-sdk/pull/2130#issuecomment-716172085) [](ancestors = 716172085)

src/Sarif/FileRegionsCache.cs
lgolding: Let's provide XML doc comments for both `public` methods, while we're in here. Don't forget to fill in descriptions of each parameter, and also the return value.

---
In reply to: [511620145](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511620145) [](ancestors = 511620145,511616376)

eddynaka: Just updated! Can you check if that is good enough?

---
In reply to: [511622336](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511622336) [](ancestors = 511622336,511620145,511616376)

src/Sarif/IFileRegionsCache.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// charLength, etc.
+        /// </summary>
+        /// <param name="inputRegion">Input region to be based on.</param>
+        /// <param name="uri">Uri that will be used to get NewLineIndex from cache.</param>
+        /// <param name="populateSnippet">Boolean that indicates if the snipper will be populated.</param>
+        /// <returns>An instance of a Region class.</returns>
+        Region PopulateTextRegionProperties(Region inputRegion, Uri uri, bool populateSnippet);
+
+        /// <summary>
+        /// Construct new region based on input region and uri.
```
lgolding, line 24: >new region [](start = 22, length = 10)

What this method actually does is to create a _context region_ around the input region, and populate its code snippet. It probably should have been named `ConstructContextRegionWithSnippet`, but don't change that for now because it's a breaking change. Saying that it creates a "new region" doesn't tell the reader quite enough. So let's have the comment say "Create a context region around the specified region, including a code snippet" and the return value be "A context region around <param cref="inputRegion">."

lgolding: Other than that, it's fine.

---
In reply to: [511624864](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511624864) [](ancestors = 511624864)

lgolding: Oh darn it, wait. The interface itself needs a summary comment. In this case, the implementation class can't just `<inheritdoc/>` on the interface. I suggest that, just as the class summary comment starts with "A class that can be used to...", the interface summary comment can start with "An interface that can be used to..." and then the rest can be the same.

---
In reply to: [511624890](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511624890) [](ancestors = 511624890,511624864)

eddynaka: Can you check again?

---
In reply to: [511625163](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511625163) [](ancestors = 511625163,511624890,511624864)

lgolding: Eddy, I'm sorry I wasn't clear about what comments to put where. Let me just pull your branch and tweak it.

---
In reply to: [511629640](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511629640) [](ancestors = 511629640,511625163,511624890,511624864)

lgolding: There you go.

---
In reply to: [511638795](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511638795) [](ancestors = 511638795,511629640,511625163,511624890,511624864)

eddynaka: thank you! 

---
In reply to: [511641601](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511641601) [](ancestors = 511641601,511638795,511629640,511625163,511624890,511624864)

src/Sarif/FileRegionsCache.cs
```diff
@@ -1,23 +1,35 @@
...
+        /// <param name="run">
+        /// The <see cref="Run"/> object whose data is to be cached.
+        /// </param>
+        /// <param name="capacity">
+        /// The initial capacity of the cache.
+        /// </param>
+        /// <param name="fileSystem">
+        /// An object that provides access to file system services.
+        /// </param>
         public FileRegionsCache(Run run, int capacity = DefaultCacheCapacity, IFileSystem fileSystem = null)
```
lgolding, line 33: >capacity [](start = 45, length = 8)

By the way, I noticed that the parameter `capacity` is unused. It should have been passed as the second argument to the `Cache` ctor just below.

eddynaka: just pushed a change fixing it.

---
In reply to: [511647723](https://github.com/microsoft/sarif-sdk/pull/2130#discussion_r511647723) [](ancestors = 511647723)

michaelcfanning:
Please see my comments in #2129, I'd prefer not to take all this code to accomplish the goal of testability. My suggestion is to simply mark relevant methods as virtual. The VS extension can simply extend this class and come up with its own test injection mechanism (if required) to test.

lgolding:
In that case please take the improved doc comments and the fix to the `capacity` bug along with marking the methods virtual.

For discussion's sake: one benefit of exposing the interface is that test code doesn't have to take a dependency on the types used in the class's implementation, but only on those defined in its interface.

---
In reply to: [716213314](https://github.com/microsoft/sarif-sdk/pull/2130#issuecomment-716213314) [](ancestors = 716213314)

PR #2131, opened 2020-10-25
Improve FileRegionsCache testability

### LINKED ISSUE #2129: Improve FileRegionsCache testability

eddynaka:
Fixes #2129

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,8 +1,5 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
-* FEATURE: Improve `FileRegionsCache` testability [#2131](https://github.com/microsoft/sarif-sdk/pull/2131)
-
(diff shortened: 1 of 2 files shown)


src/ReleaseHistory.md
```diff
@@ -1,5 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+* FEATURE: Improve `FileRegionsCache` testability [#2131](https://github.com/microsoft/sarif-sdk/pull/2131)
```
michaelcfanning, line 3: I'm not sure I see this change as significant as to require a readme note (particularly since marking a non-virtual method as virtual is a non-breaking change. If we start to simply document each merge, maybe we should be handling this via descriptive commit messages (and letting the GH release process provide our docs. Let's discuss with Chris at Monday's stand-up.

eddynaka: Yeah, you are right. Let me revert! Thanks for reviewing!

michaelcfanning (approved):
:shipit:

PR #2134, opened 2020-10-27
Fixing artifact content for non-binary types

### LINKED ISSUE #1848: Reconsider Artifact.GetEncodedFileContents behavior
**Repro**

- Create an InsertOptionalDataVisitor
- Specify TextFiles as an insert option
- Provide an originalUriBaseIds definition that maps to an actual enlistment root
- Pass the visitor a SARIF file with a relative path within that enlistment root to a file suffixed with .JSON

**Result**
Contents are encoded as binary, not text.

The reason has to do with a check in `Artifact.GetEncodedFileContents`, in which a null `inputFileEncoding` argument is an over-riding clue to emit everything as binary data, not text. The only way I found (admittedly didn't look long) to overcome this default behavior was to specify a default encoding at the run level. We need more of a 'pit of success' behavior in this API, though.
~~~
            if (SarifWriters.MimeType.IsBinaryMimeType(mimeType) || inputFileEncoding == null)
            {
                fileContent.Binary = Convert.ToBase64String(fileContents);
            }
~~~

@lgolding

eddynaka:
Fixes #1848


src/Sarif/Core/Artifact.cs
```diff
@@ -88,13 +88,21 @@ private static ArtifactContent GetEncodedFileContents(IFileSystem fileSystem, st
...
             }
             else
             {
-                fileContent.Text = inputFileEncoding.GetString(fileContents);
+                inputFileEncoding ??= new UTF8Encoding();
+                try
+                {
+                    fileContent.Text = inputFileEncoding.GetString(fileContents);
+                }
+                catch (Exception)
```
michaelcfanning, line 102: >Exception [](start = 23, length = 9)

I'd removed this exception handler. First, I don't think it's likely that Encoding.GetString will fail against C# string values. Do you know of an expected negative condition? If there *is* one we need to handle, we should catch a more precise exception (swallowing all Exceptions is not a good idea).

eddynaka: We can receive a few: [https://docs.microsoft.com/en-us/dotnet/api/system.text.utf8encoding.getstring?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.text.utf8encoding.getstring?view=netcore-3.1), but normally we wouldn't face that. I will remove that for now. Thanks for reviewing!

---
In reply to: [513607954](https://github.com/microsoft/sarif-sdk/pull/2134#discussion_r513607954) [](ancestors = 513607954)

michaelcfanning (approved):
:shipit:

PR #2135, opened 2020-10-27
Upgrading .NET Framework from .NET45 to .NET452

### LINKED ISSUE #417: Can we drop the SDK to .NET 4.5
Encountered yet another project compiled for 4.5 that needed to be updated to 4.51 to use SARIF.

eddynaka:
Fixes #417

## Changes:
- upgraded all projects below .NET Framework 4.5.2 to .NET Framework 4.5.2
- added missing projects to sample solution

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,6 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
-* FEATURE: SDK BREAKING CHANGE: Upgrade from .NET Framework 4.5 to .NET Framework 4.5.2 [#2135](https://github.com/microsoft/sarif-sdk/pull/2135)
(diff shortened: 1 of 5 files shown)


src/ReleaseHistory.md
```diff
@@ -1,5 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+* FEATURE: SDK BREAKING CHANGE: Upgrade from .NET Framework 4.5 to .NET Framework 4.5.2 [#2135](https://github.com/microsoft/sarif-sdk/pull/2135)
```
michaelcfanning, line 3: >FEATURE: SDK BREAKING CHANGE [](start = 2, length = 28)

We usually call this kind of change 'PACKAGE BREAKING'. i.e., it is a compatibility break at the package level. It's not a great name. :) But that's the pattern.

michaelcfanning (approved):
:shipit:

src/ReleaseHistory.md
eddynaka: just updated!

---
In reply to: [513101256](https://github.com/microsoft/sarif-sdk/pull/2135#discussion_r513101256) [](ancestors = 513101256)

PR #2137, opened 2020-10-28
Removing unused files

### LINKED ISSUE #2136: Remove appveyor stuff; fold BeforeBuild.ps1 into BuildAndTest.ps1
At one time, AppVeyor could not run `BuildAndTest.ps1` because AppVeyor did not allow you to specify a build script. It only allowed you to specify (1) the project to build, and (2) a script to run before the build step. Now, AppVeyor does allow you to specify a build script, and in fact `appveyor.yml` does run `BuildAndTest.ps1`. So there's no reason for `BeforeBuild.ps1` to exist as a separate script. Fold its contents into `BuildAndTest.ps1`.

Taking it a step further, we don't build with AppVeyor at all any more; we have an ADO build pipeline. So just remove `appveyor.yml`. Also check if anybody needs the scripts in the `appveyor/` directory, and if not, remove them.

I discovered this when I started wondering why the `sarif-visualstudio-extension` hard-codes the version number in so many places, where the SDK does not [*]. It turns out that that repo's build scripts don't even _call_ `BeforeBuild.ps1` -- even though a copy of it does exist in that repo! -- so `New-VersionConstantsFile.ps1` is never called, so there is no such thing as a `VersionConstants.cs` file (which is what `New-VersionConstantsFile.ps1` generates), so you have to hard-code the version everywhere.

I'm fixing that in the extension project.

---

[*] And I noticed _that_ because I discovered that the extension's version as displayed in the VS **Help**, **About** dialog was "2.0 beta"! And I tried to specify it as `VersionConstants.FileVersion`, and found that there was no such thing.

@michaelcfanning @eddynaka

eddynaka:
Fixes #2136


michaelcfanning (approved):
:shipit:

GabeDeBacker (approved):
:shipit:

PR #2144, opened 2020-11-03
Apply policies accept external policies

### LINKED ISSUE #2139: Update apply-policy command to accept/apply an external policy file
@eddynaka

eddynaka:
Fixes #2139

### DIFF
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -128,7 +128,6 @@ namespace Microsoft.CodeAnalysis.Sarif
             SarifLog sarifLog = Load(sarifLogPath);
             if (sarifLog == null || sarifLog.Runs == null)
             {
-                ApplyPolicies();
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


src/Sarif/Core/SarifLog.cs
```diff
@@ -115,7 +116,46 @@ public void ApplyPolicies()
...
+            }
+        }
+
+        /// <summary>
+        /// Applies the policies contained from a sarif file
+        /// </summary>
+        public void ApplyPolicies(string sarifLogPath)
+        {
+            SarifLog sarifLog = Load(sarifLogPath);
+            if (sarifLog == null || sarifLog.Runs == null)
```
michaelcfanning, line 129: So this code appears to fall back to the internal policies on this log, if any, if the external file can't be loaded. Is that right? If someone specifies an external policy file that doesn't have anything, should we stop processing policy altogether?

That strikes me as cleaner: if you specify any external policy, it replaces use of any internal policies (on the log being processed).

eddynaka: ok. Let me make that change!

---
In reply to: [517495418](https://github.com/microsoft/sarif-sdk/pull/2144#discussion_r517495418) [](ancestors = 517495418)

michaelcfanning (approved):
:shipit:

PR #2146, opened 2020-11-05
Implement multithreaded driver analyze command w/deterministic output

michaelcfanning:
This change introduces a new MultithreadedAnalyzeCommand that parallelizes analysis while also guaranteeing that all results are output in a deterministic order run-over-run. The mechanism is like so:

1) A single-threaded enumerator/channel writer enqueues an alphabetized list of directories (based on command-line arguments), then dequeues each directory, enumerates files within each, alphabetizes them and creates an analysis context object for each that is stored in a list. The index of each created context item is written to the channel.
2) A multithreaded channel reader receives an index for a file to scan, retrieves its corresponding context object from the global array, and analyzes it. Any results are written to the per-scan target context object. On completing analysis, the analysis thread writes the current context index to the results logging channel.
3) A single-threaded reader receives the index of a context object which represents a scanned target. If the reader has already processed all scan targets leading up to that index, it retrieves the results from the current context and writes them again to the global, actual persisted log (as well as the console). The writer also 'runs ahead' on persisting results if it observes that other threads have completed analysis work which hasn't yet resulted in a channel callback. The writer nulls each context object in the array as work is completed.

To test the above, I've added an 'analyze-test' command to the multitool, which can be a temporary addition until we harden the new code and dropped the single-threaded analysis entirely (this analysis command can eventually be replaced by the multithread analysis configured to run on a single thread). Because we've hit our limit for verbs (as supported by CommandLineLibrary), I had to drop a result matching verb previously used mostly for testing.

Other work and open items: 

1. Most unit tests go through both the single and multi-threaded code paths but not all.
2. This change includes an optimization to the SarifLogger to stop emitting non-useful artifacts table data when hashing is not enabled.
3. I discovered an inefficiency in the previous analysis where file hashing is duplicated (i.e., it is performed once when processing scan targets and again the first time any file actually produces a result).

@eddynaka, this is still a draft.


src/Sarif/Sarif.csproj
```diff
@@ -31,12 +31,12 @@
...
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Condition="'$(TargetFramework)' == 'net452' Or '$(TargetFramework)' == 'net461' " Include="System.Web" />
-    <Reference Condition="'$(TargetFramework)' == 'net452' " Include="System.Runtime" />
+    <Reference Condition="'$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net461' " Include="System.Web" />
+    <Reference Condition="'$(TargetFramework)' == 'net45' " Include="System.Runtime" />
```
eddynaka, line 39: Let's remove this, since we don't target net45

eddynaka: And, for line 38, we should update and add net452 and probably net472.

src/Sarif/HashUtilities.cs
```diff
@@ -48,7 +48,12 @@ internal static IFileSystem FileSystem
                     {
                         if (queue.TryDequeue(out string filePath))
                         {
-                            fileToHashDataMap[filePath] = HashUtilities.ComputeHashes(filePath);
+                            HashData hashData = HashUtilities.ComputeHashes(filePath);
+
+                            lock (fileToHashDataMap)
```
eddynaka, line 53: Since we are using a ConcurrentDictionary, probably we won't need the lock.

src/Sarif.Multitool/AnalyzeTestSkimmer.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using System.Collections.Generic;
+using System.Resources;
+using System.Text;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public class AnalyzeTestSkimmer : Skimmer<AnalyzeTestContext>
```
eddynaka, line 13: Since this in a test, should we mark as internal or add a #if debug condition?

src/Sarif.Multitool/AnalyzeTestOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text;
+
+using CommandLine;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    [Verb("analyze-test", HelpText = "Test the analysis driver framework.")]
+    public class AnalyzeTestOptions : MultithreadedAnalyzeOptionsBase
```
eddynaka, line 15: Since this in a test, should we mark as internal or add a #if debug condition?

src/Sarif.Multitool/AnalyzeTestContext.cs
```diff
@@ -1,16 +1,23 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
 {
-    public class AnalysisContext : IAnalysisContext
+    public class AnalyzeTestContext : IAnalysisContext
     {
         public Exception TargetLoadException { get; set; }
 
-        public bool IsValidAnalysisTarget { get; set; }
+        public bool IsValidAnalysisTarget
```
eddynaka, line 16: This could be replaced for this: `public bool IsValidAnalysisTarget => true;`

src/Sarif.Multitool/AnalyzeTestContext.cs
```diff
@@ -1,16 +1,23 @@
-﻿
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
```
eddynaka, line 5: should we remove those using's that we are not using?

src/Sarif.Multitool/AnalyzeTestCommand.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    public class AnalyzeTestCommand : MultithreadedAnalyzeCommandBase<AnalyzeTestContext, AnalyzeTestOptions>
```
eddynaka, line 13: should we mark as internal or add a #if debug since its a test?

eddynaka (approved):

PR #2156, opened 2020-11-17
Updating export-docs

### LINKED ISSUE #2155: export-validation-docs should accept another binary
Today, the command `export-validation-docs` does not accept an external assembly containing rules.

## Suggestion

Create a new property that can point to an assembly which contains rules, for example, binskim.rules.dll. With that, we would be able to generate any rules from any provider who implement the `Skimmer` class. If the property was not supplied, so it will warn or use the self assembly (just the way we have now).

@michaelcfanning

eddynaka:
Fixes #2155

## Changes:
- updating multitool parameters in alphabetical order
- enabling export-docs to accept list of assemblies that implement `ReportingDescriptor`


src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## **Unreleased**
 * FEATURE: Multitool SARIF rewrite accepts `remove` parameter[#2160](https://github.com/microsoft/sarif-sdk/pull/2160)
+* BREAKING: Remove command `export-validation-docs` and add `export-validation-rules` command can export markdown file [#2156](https://github.com/microsoft/sarif-sdk/pull/2156)
```
michaelcfanning, line 5: >Remove command `export-validation-docs` and add `export-validation-rules` command can export markdown file [#2156](https://github.com/microsoft/sarif-sdk/pull/2156) [](start = 12, length = 164)

Remove command `export-validation-docs` and extend `export-validation-rules` command to export markdown file [#2156](https://github.com/microsoft/sarif-sdk/pull/2156)

michaelcfanning (approved):
:shipit:

PR #2159, opened 2020-11-21
Migrating extensions from sarif-visualstudio-extension

### LINKED ISSUE #2157: Migrate suppressed and absent issue helpers from VS extension to SDK
Migrate the following helpers from the sarif-visualstudio-extension repo to the SDK repo:

- `Run.HasAbsentResults`
- `Run.HasSuppressedResults`
- `Result.IsSuppressed`

@michaelcfanning @eddynaka FYI

eddynaka:
Fixes #2157


src/Sarif/Extensions/ResultExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.Extensions
+{
+    public static class ResultExtensions
```
lgolding, line 10: >ResultExtensions [](start = 24, length = 16)

One we are in the SARIF SDK, these no longer need extension methods. The generated classes in `Sarif\Autogenerated` are `partial` classes, meaning that we can extend them with first-class methods. See the classes in `Sarif\Core`. That's where these methods now belong.

lgolding (changes requested):
:clock1:

lgolding:
Bring along the unit tests as well.

src/Sarif/Extensions/ResultExtensions.cs
eddynaka: fixed!

---
In reply to: [528762018](https://github.com/microsoft/sarif-sdk/pull/2159#discussion_r528762018) [](ancestors = 528762018)

eddynaka:
done!

---
In reply to: [732213999](https://github.com/microsoft/sarif-sdk/pull/2159#issuecomment-732213999) [](ancestors = 732213999)

src/Sarif/Sarif.csproj
```diff
@@ -32,7 +32,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
```
lgolding, line 35: >GitHub [](start = 52, length = 6)

What is this for?

lgolding (approved):
:shipit:

src/Sarif/Core/Result.cs
```diff
@@ -93,6 +93,23 @@ public ReportingDescriptor GetRule(Run run = null)
...
 
+        public bool IsSuppressed()
+        {
+            if (this == null)
+            {
+                throw new ArgumentNullException(nameof(Result));
+            }
+
+            IList<Suppression> suppressions = this.Suppressions;
+            if (suppressions == null || suppressions.Count == 0)
```
michaelcfanning, line 104: >suppressions == null [](start = 16, length = 20)

if suppressions is null, we don't have *any* suppressions related data in this log and so we can't answer the question. Two ideas, either throw an exception for this case of use bool TryIsSuppressed...

michaelcfanning (changes requested):
:clock1:

src/Sarif/Core/Result.cs
eddynaka: updated. One thing that i changed: we won't throw exception. What will happen is if the Result isn't available, then we will return false.

---
In reply to: [529072719](https://github.com/microsoft/sarif-sdk/pull/2159#discussion_r529072719) [](ancestors = 529072719)

michaelcfanning: that's right! in the TryXXX pattern, this is exactly the behavior you want. 

---
In reply to: [529483550](https://github.com/microsoft/sarif-sdk/pull/2159#discussion_r529483550) [](ancestors = 529483550,529072719)

michaelcfanning (approved):
:shipit:

PR #2160, opened 2020-11-24
Enable RemoveOptionalData for RewriteCommand

eddynaka:
## Changes:
- Enable `RemoveOptionalData` for `RewriteCommand`
- Enable `CodeFlows` as `OptionallyEmittedData`

With this we would be able to use the following command:
```
rewrite oldfile.sarif --output-path newfile.sarif --remove "CodeFlows" --minify true --force
```

Adding some data:
- based on a sarif with size of 70MB with CodeFlows, after the command, the sarif size dropped to 7MB
- based on a sarif with size of 400MB with CodeFlows, after the command, the sarif size dropped to 8MB

### DIFF
--- a/docs/multitool-usage.md
+++ b/docs/multitool-usage.md
@@ -34,6 +34,9 @@ Sarif.Multitool convert Current.fpr -tool FortifyFpr -output Current.sarif
 : Add file contents from analyzed files and snippets from result regions to SARIF
 Sarif.Multitool rewrite Current.sarif --insert "TextFiles,RegionSnippets" --inline
 
+: Remove codeFlows from results regions to SARIF
+Sarif.Multitool rewrite Current.sarif --remove "CodeFlows" --inline
+
 : Rebase URIs from local paths to a token (for later resolution against remote URLs)
 Sarif.Multitool rebaseuri Current.sarif --base-path-value "C:\Local\Path\To\Repo" --base-path-token REPO
 
@@ -72,4 +75,5 @@ Run ```Sarif.Multitool convert --help``` for the current list.
 | Name | Purpose |
 | ---- | ------- |
 | pretty-print | Produce pretty-printed JSON output rather than compact form. |
+| minify | Produce compact JSON output (all white space removed) rather than pretty-printed output. |
 | force | Force overwrite of output file if it exists. |
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+* FEATURE: Multitool SARIF rewrite accepts `remove` parameter[#2160](https://github.com/microsoft/sarif-sdk/pull/2160)
+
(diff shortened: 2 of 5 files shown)


docs/multitool-usage.md
```diff
@@ -72,4 +75,5 @@ Run ```Sarif.Multitool convert --help``` for the current list.
 | Name | Purpose |
 | ---- | ------- |
 | pretty-print | Produce pretty-printed JSON output rather than compact form. |
```
michaelcfanning, line 77: >form [](start = 72, length = 4)

We should mention this is the default! :) Probably, this argument should be deprecated over time (i.e., we'd only have minify which would reverse our default pretty print behavior).

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+* FEATURE: Multitool SARIF rewrite accepts `remove` parameter[#2160](https://github.com/microsoft/sarif-sdk/pull/2160)
+
+## **v2.3.8** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.3.8) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.3.8) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.3.8) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.3.8) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/2.3.8)
+
 * FEATURE: PACKAGE BREAKING: Upgrade from .NET Framework 4.5 to .NET Framework 4.5.2 [#2135](https://github.com/microsoft/sarif-sdk/pull/2135)
```
michaelcfanning, line 7: >FEATURE [](start = 2, length = 7)

Eliminate whitespace on line 6? Go ahead and add the v2.3.9 banner at the top of the file?

michaelcfanning (approved):
:shipit:

michaelcfanning:
Nice work! Left a couple of small suggestions to act on, just update as you think best and merge.

src/ReleaseHistory.md
eddynaka: I was thinking about this and it may cause some confusion because it will show that we have already the version 2.3.9, which is not true. What i will do is change and create a ## Unreleased.
What do you think?

---
In reply to: [529763647](https://github.com/microsoft/sarif-sdk/pull/2160#discussion_r529763647) [](ancestors = 529763647)

michaelcfanning: Sure, that's a fine idea. We just need to be disciplined about updating it when we actually release. 

---
In reply to: [529771234](https://github.com/microsoft/sarif-sdk/pull/2160#discussion_r529771234) [](ancestors = 529771234,529763647)

docs/multitool-usage.md
eddynaka: I took a look at our base class and pretty-print isn't default. What we could do is: in a separate PR remove the minify since minify is the opposite of pretty-print. What do you think?

---
In reply to: [529762644](https://github.com/microsoft/sarif-sdk/pull/2160#discussion_r529762644) [](ancestors = 529762644)

eddynaka:
Ok, since the pretty-print isn't default , i won't change for now. I'm waiting the pipeline to merge.


---
In reply to: [733135964](https://github.com/microsoft/sarif-sdk/pull/2160#issuecomment-733135964) [](ancestors = 733135964)

PR #2167, opened 2020-11-25
Validate slash file relative locations

eddynaka:


src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs
```diff
@@ -49,6 +49,40 @@ protected override void Analyze(Run run, string runPointer)
...
+            {
+                return;
+            }
+
+            string uri = artifactLocation.Uri.ToString();
+            if (artifactLocation.Uri.IsAbsoluteUri)
+            {
+                if (artifactLocation.Uri.IsFile)
+                {
+                    if (uri.EndsWith("/"))
```
michaelcfanning, line 64: >EndsWith [](start = 28, length = 8)

Don't worry about the ends with cases

src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs
```diff
@@ -49,6 +49,40 @@ protected override void Analyze(Run run, string runPointer)
...
+                    {
+                        // TODO: we have a problem, since folders should end with slash
+                    }
+                }
+            }
+            else
+            {
+                if (uri.StartsWith("/"))
+                {
+                    // TODO: we have a problem, since relative paths should not start with slash
```
michaelcfanning, line 81: > // TODO: we have a problem, since relative paths should not start with slash [](start = 19, length = 77)

Yes, for this case, emit an error message (you have a relative url that is prefixed with a slash.

michaelcfanning:
>                     }

this is where to add teh extended analysis that produces an error on observing a leading slash in a relative URL

---
Refers to: src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs:117 in e6e52dd. [](commit_id = e6e52dde72f7d8d434a68f1a873dbead5a8581be, deletion_comment = False)

src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs
```diff
@@ -49,6 +49,40 @@ protected override void Analyze(Run run, string runPointer)
...
+        {
+            if (artifactLocation.Uri == null)
+            {
+                return;
+            }
+
+            string uri = artifactLocation.Uri.ToString();
+            if (artifactLocation.Uri.IsAbsoluteUri)
+            {
+                if (artifactLocation.Uri.IsFile)
```
michaelcfanning, line 62: >artifactLocation.Uri.IsFile [](start = 20, length = 27)

Fire a note for this case, we observed an absolute file url, which should be rendered instead as a relative URL.

eddynaka:
>                     }

just added the new logic. Can you validate the message and see if that is OK?
Thank you :)

---
In reply to: [736061872](https://github.com/microsoft/sarif-sdk/pull/2167#issuecomment-736061872) [](ancestors = 736061872)

---
Refers to: src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs:117 in e6e52dd. [](commit_id = e6e52dde72f7d8d434a68f1a873dbead5a8581be, deletion_comment = False)

src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs
```diff
@@ -72,14 +73,32 @@ private void AnalyzeResultLocations(IList<Location> locations, string locationsP
...
+                                nameof(RuleResources.SARIF2016_FileUrisShouldBeRelative_Note_Default_Text),
+                                location.PhysicalLocation.ArtifactLocation.Uri.OriginalString);
+                        }
+                    }
+                    else
+                    {
+                        if (location.PhysicalLocation.ArtifactLocation.Uri.OriginalString.StartsWith("/"))
+                        {
+                            // {0}: The file location '{1}' start with slash. Prefer a relative reference
+                            // together with a uriBaseId property.
```
michaelcfanning, line 93: let's use this text:

{0}: The relative file URL '{1}' is prefixed with a leading slash, which can lead to unintended behavior when concatenating with absolute URLs. Remove the leading slash.

src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs
```diff
@@ -72,14 +73,32 @@ private void AnalyzeResultLocations(IList<Location> locations, string locationsP
...
+                            // reference together with a uriBaseId property.
+                            LogResult(
+                                locationPointer.AtProperty(SarifPropertyName.PhysicalLocation).AtProperty(SarifPropertyName.ArtifactLocation).AtProperty(SarifPropertyName.Uri),
+                                nameof(RuleResources.SARIF2016_FileUrisShouldBeRelative_Note_Default_Text),
+                                location.PhysicalLocation.ArtifactLocation.Uri.OriginalString);
+                        }
+                    }
+                    else
+                    {
+                        if (location.PhysicalLocation.ArtifactLocation.Uri.OriginalString.StartsWith("/"))
```
michaelcfanning, line 90: >location.PhysicalLocation.ArtifactLocation.Uri.OriginalString [](start = 28, length = 61)

Are we sure this is right? Try creating a relative URL like so, new Url(@"\myTest\myFile.cpp", UriKind.Relative). I think URL.ToString() will convert to forward slashes but OriginalString will have back slashes.

eddynaka: I did some testing:
```csharp
            var uri = new Uri(@"\myTest\myFile.cpp", UriKind.Relative);
            var uri2 = new Uri(@"/myTest/myFile.cpp", UriKind.Relative);

            Console.WriteLine(uri.OriginalString);
            Console.WriteLine(uri.ToString());

            Console.WriteLine(uri2.OriginalString);
            Console.WriteLine(uri2.ToString());
```

the value printed:
```
\myTest\myFile.cpp
\myTest\myFile.cpp
/myTest/myFile.cpp
/myTest/myFile.cpp
```

Based on this, I updated the logic to check for `\` and for `/`.

What do you think?

---
In reply to: [533633936](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533633936) [](ancestors = 533633936)

michaelcfanning: Well, I think this raises the possibility of updating this analysis further! Confusion around forward or back slashes is really prevalent for file URLs, particularly when considering Windows. We should probably encourage the use of forward slashes only (and update the validator to check for this). Try this test: new Uri(@"c:\test\myfile.cpp", UriKind.Absolute). Does uri.ToString() convert back slashes to forward in this case?

---
In reply to: [533669651](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533669651) [](ancestors = 533669651,533633936)

eddynaka: Below the results:
```
c:\test\myfile.cpp (OriginalString)
[file:///c:/test/myfile.cpp](file:///c:/test/myfile.cpp) (tostring())
```

---
In reply to: [533692935](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533692935) [](ancestors = 533692935,533669651,533633936)

michaelcfanning: So, I ran a test and is a relative URI contains back slashes, they are preserved in the final constructed URI. This is not good, as this value could be compared to other (say, locally constructed) URIs. So let's do update this analysis to prevent back slashes in any relative URI. You can open a new issue and do it as follow on work, if you like.

[file:///c:/mydir\anotherDir\myFile](file:///c:/mydir\anotherDir\myFile)

---
In reply to: [533694852](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533694852) [](ancestors = 533694852,533692935,533669651,533633936)

eddynaka: ok, so, we would have:
1. if relative and start with "/" -> that is a problem
2. if relative and contains "\" -> that is a problem

Is that right?

---
In reply to: [533731429](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533731429) [](ancestors = 533731429,533694852,533692935,533669651,533633936)

eddynaka: Just updated the logic. Can you validate and check the new message that i just created?
Thank you!

---
In reply to: [533735403](https://github.com/microsoft/sarif-sdk/pull/2167#discussion_r533735403) [](ancestors = 533735403,533731429,533694852,533692935,533669651,533633936)

michaelcfanning:
>                     }



{0}: The relative file URL '{1}' contains one or more backslashes, which will be preserved when concatenating to an absolute URL. This can result in inconsistent representations, compared to URLs created from an absolute file path, which may be regarded as not equivalent. Replace all backslashes with forward slashes.

---
In reply to: [736115881](https://github.com/microsoft/sarif-sdk/pull/2167#issuecomment-736115881) [](ancestors = 736115881,736061872)

---
Refers to: src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs:117 in e6e52dd. [](commit_id = e6e52dde72f7d8d434a68f1a873dbead5a8581be, deletion_comment = False)

michaelcfanning (approved):
:shipit:

michaelcfanning:
>                     }

After you fix up the messages, go ahead and merge.

---
In reply to: [736115881](https://github.com/microsoft/sarif-sdk/pull/2167#issuecomment-736115881) [](ancestors = 736115881,736061872)

---
Refers to: src/Sarif.Multitool.Library/Rules/SARIF2016.FileUrisShouldBeRelative.cs:117 in e6e52dd. [](commit_id = e6e52dde72f7d8d434a68f1a873dbead5a8581be, deletion_comment = False)

PR #2168, opened 2020-11-25
Various driver framework improvements.

michaelcfanning:
@eddynaka


src/Sarif/NotYetAutoGenerated/ReportingDescriptorEqualityComparer.cs
```diff
@@ -0,0 +1,338 @@
...
+                if (obj.MessageStrings != null)
+                {
+                    // Use xor for dictionaries to be order-independent.
+                    int xor_0 = 0;
+                    foreach (var value_7 in obj.MessageStrings)
+                    {
+                        xor_0 ^= value_7.Key.GetHashCode();
+                        if (value_7.Value != null)
+                        {
+                            xor_0 ^= MultiformatMessageStringEqualityComparer
```
michaelcfanning, line 282: >MultiformatMessageStringEqualityComparer [](start = 37, length = 40)

By chance, I found a serious problem in our equality comparer, which doesn't itself call the equality comparer for any dictionaries that contain values which are SARIF reference types (like this case, where we have a dictionary whose values are MultiformatMessageString instances. We probably have other exposure to this bug in the format, I will open an issue on it. For now, I've made a manual update to the autogenerated code to handle this case.

src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -231,7 +231,7 @@ private static void ValidateKindAndLevel(ResultKind kind, FailureLevel level)
 
             string messageText =
                    (path != null ? (path + location) : toolName) + ": " +
-                   issueType + (!string.IsNullOrEmpty(ruleId) ? " " : "") +
+                   issueType + " " +
```
michaelcfanning, line 234: >issueType [](start = 19, length = 9)

Without this fix, in the absence of a rule id, the 'level' would be immediately followed by the message, e.g., 'errorHereis the message'.

src/Sarif.Driver/DriverExtensionMethods.cs
```diff
@@ -12,12 +12,13 @@ public static class DriverExtensionMethods
     {
         public static LoggingOptions ConvertToLoggingOptions(this AnalyzeOptionsBase analyzeOptions)
         {
-            LoggingOptions loggingOptions = LoggingOptions.None;
+            LoggingOptions loggingOptions = LoggingOptions.PrettyPrint;
```
michaelcfanning, line 15: >PrettyPrint [](start = 59, length = 11)

Here's our behavior now, if neither minify or pretty print is specified, we pretty print. If only minify is specified, we minify, if both minify and pretty print are specified, we pretty print. We could consider raising an exception instead for the case where both settings exist.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -153,9 +158,9 @@ private void Analyze(TOptions analyzeOptions, AggregatingLogger logger)
             IEnumerable<Skimmer<TContext>> skimmers,
             ISet<string> disabledSkimmers)
         {
-            options.ThreadCount = options.ThreadCount > 0
-                ? options.ThreadCount
-                : Environment.ProcessorCount;
+            options.ThreadCount = options.ThreadCount > 0 ?
```
michaelcfanning, line 156: >ThreadCount [](start = 20, length = 11)

with this change, if no thread count is explicitly specified, we drop the thread count to one if the debugger is attached. generally, if debugging, you don't want multithreading enabled. you can still force this when debugging by providing an explicit thread count > 1.

src/Sarif.Driver/DriverExtensionMethods.cs
eddynaka: In this case, i think we are missing one validation in the ValidateOptions method.
If we check, we have another implementation `ValidateOutputFormatOptions` that checks exactly for this behavior

---
In reply to: [530638428](https://github.com/microsoft/sarif-sdk/pull/2168#discussion_r530638428) [](ancestors = 530638428)

eddynaka: Taking a second look, we wouldn't be able to use just the way it is because it uses another class as option. In the future, we could refactor this.

---
In reply to: [530655298](https://github.com/microsoft/sarif-sdk/pull/2168#discussion_r530655298) [](ancestors = 530655298,530638428)

eddynaka (approved):

src/Sarif.Driver/DriverExtensionMethods.cs
michaelcfanning: You're right, we should make this a general helper in the driver.

---
In reply to: [530675864](https://github.com/microsoft/sarif-sdk/pull/2168#discussion_r530675864) [](ancestors = 530675864,530655298,530638428)

PR #2169, opened 2020-11-29
Driver om changes

michaelcfanning:
@eddynaka


src/Sarif/DefaultTraces.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    [Flags]
+    public enum DefaultTraces
```
michaelcfanning, line 9: >DefaultTraces [](start = 16, length = 13)

Creating a new pattern here allowing for specific kinds of traces to be enabled on the command-line. Currently, the --verbose flag enables all this output, which is very copious.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -87,5 +87,15 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
...
+
+        [Option(
+            "trace",
+            Separator = ';',
+            Default = new string[] { },
+            HelpText = "Execution traces, expressed as a semicolon-delimited list, that " +
+                       "should be emitted to the console and log file (if appropriate). " +
+                       "Valid values: ScanTime.")]
+
+        public virtual IEnumerable<string> Traces { get; set; }
```
michaelcfanning, line 99: >string [](start = 35, length = 6)

In order to provide a common argument, this must be of 'string' type. This will allow driver apps to replace DefaultTraces with a more expansive enum.

eddynaka, line 99: remove extra space

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -695,7 +704,7 @@ public void InvokeCatchingRelevantIOExceptions(Action action, Action<Exception>
             }
         }
 
-        protected virtual ISet<Skimmer<TContext>> CreateSkimmers(TContext context)
+        protected virtual ISet<Skimmer<TContext>> CreateSkimmers(TOptions options, TContext context)
```
eddynaka, line 707: >TOptions options [](start = 65, length = 16)

do we need this parameter?

src/Sarif.Driver/Sdk/Skimmer.cs
```diff
@@ -51,7 +49,7 @@ public Skimmer()
 
         public override MultiformatMessageString FullDescription => throw new InvalidOperationException($"The {nameof(FullDescription)} property must be overridden in the SkimmerBase-derived class.");
 
-        public override MultiformatMessageString ShortDescription => new MultiformatMessageString { Text = ExtensionMethods.GetFirstSentence(FullDescription.Text) };
+        public override MultiformatMessageString ShortDescription => null;
```
eddynaka, line 52: >null [](start = 69, length = 4)

is this right? if we see the tests, we removed all short descriptions

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning: Good question. We do need this in order to allow for the command-line options to drive skimmer creation. We do this in the SPAM checker. In the future, for example, we might have an argument in our options that points to an assembly that holds skimmers (and the tool will load them dynamically).

---
In reply to: [532125229](https://github.com/microsoft/sarif-sdk/pull/2169#discussion_r532125229) [](ancestors = 532125229)

src/Sarif.Driver/Sdk/Skimmer.cs
michaelcfanning: Yes, this is right. In SARIF, this null return value will indicate that the short description isn't populated (in the default implementation) and it will be omitted from the SARIF log file entirely (which is correct). In SARIF, the very first sentence of the full description serves as the concise short description, if it exists. Rule authors who follow this pattern don't need to provide a short description (and in fact this is the recommended practice).

---
In reply to: [532184727](https://github.com/microsoft/sarif-sdk/pull/2169#discussion_r532184727) [](ancestors = 532184727)

eddynaka (approved):
:shipit:

PR #2170, opened 2020-11-30
Include Multitool.library to nuspec

### LINKED ISSUE #2064: Sarif.Multitool package does not include sources
The nuspec file `src\NuGet\Sarif.Multitool.nuget` mentions
```xml
<file src="src\Sarif.Multitool\**\*.cs" target="src" />
```
... but there is no `src` directory in the package.

This is not a regression (that is, it is not due to my having separated out the `Sarif.Multitool.Library` into a separate package). I just noticed it now. It was broken in 2.3.4, not just in the new 2.3.5.

This matters less than it used to because the only source file in `Sarif.Multitool` is now `Program.cs`. The source files _do_ appear correctly in the new `Sarif.Multitool.Library` package.

@michaelcfanning FYI

eddynaka:
Fixes #2064

### DIFF
--- a/src/Nuget/Sarif.Multitool.nuspec
+++ b/src/Nuget/Sarif.Multitool.nuspec
@@ -47,6 +47,6 @@
 
     <file src="triskelion.png" />
     <file src="src\ReleaseHistory.md" />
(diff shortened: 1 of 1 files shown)


michaelcfanning (approved):
:shipit:

PR #2172, opened 2020-12-02
Update all JSON.NET use to 11.0 (to support .NET standard 2.0)

michaelcfanning:
Oops. We *said* we updated to 11.0 previously but we neglected to update the core SARIF dll.

@eddynaka, take a look. Please merge if you agree with the change.


eddynaka:
@michaelcfanning , we have to take a look at our `.netstandard2.1` reference.
Why:
```

1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(1662,5): warning NU1702: ProjectReference 'C:\Microsoft\sarif-pattern-matcher\Src\sarif-sdk\src\Sarif.Driver\Sarif.Driver.csproj' was resolved using '.NETFramework,Version=v4.7.2' instead of the project target framework '.NETStandard,Version=v2.0'. This project may not be fully compatible with your project.
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(1662,5): warning NU1702: ProjectReference 'C:\Microsoft\sarif-pattern-matcher\Src\sarif-sdk\src\Sarif\Sarif.csproj' was resolved using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETStandard,Version=v2.0'. This project may not be fully compatible with your project.
```

This came from the new spam project...
i think our library should target `.netstandard2.0` and our tool can keep using `netcoreapp3.1`.

eddynaka (approved):

PR #2177, opened 2020-12-07
Fixing message issues

eddynaka:


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -571,12 +569,9 @@ internal string GetConfigurationFileName(TOptions options, bool unitTestFileExis
...
             if (string.IsNullOrEmpty(options.ConfigurationFilePath))
             {
-                if (!FileSystem.FileExists(DefaultConfigurationPath) && !unitTestFileExists)
-                {
-                    return null;
-                }
-
-                return DefaultConfigurationPath;
+                return !this.FileSystem.FileExists(this.DefaultConfigurationPath)
+                    ? null
```
michaelcfanning, line 573: >? null [](start = 20, length = 6)

we need to decide whether we put the '?' and ':' operators on the dropped down line or leave them at the end of the previous. Our code is inconsistent in this regard.

michaelcfanning (approved):
:shipit:

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: I see. you are right about the consistency. Added that because the line would be very long. I will check if we can validate that.

PR #2178, opened 2020-12-09
Provide text to file region cache

michaelcfanning:
SDK fix required to populate snippets from pre-retrieved text buffer.

@eddynaka , @cfaucon

### DIFF
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -75,7 +75,11 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             NewLineIndex newLineIndex = GetNewLineIndex(uri, fileText);
 
-            return PopulateTextRegionProperties(newLineIndex, inputRegion, fileText, populateSnippet);
+            return PopulateTextRegionProperties(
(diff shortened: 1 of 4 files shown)


src/Sarif/FileRegionsCache.cs
```diff
@@ -65,8 +73,13 @@ public virtual Region PopulateTextRegionProperties(Region inputRegion, Uri uri,
...
             }
 
-            NewLineIndex newLineIndex = GetNewLineIndex(uri, out string fileText);
-            return PopulateTextRegionProperties(newLineIndex, inputRegion, fileText, populateSnippet);
+            NewLineIndex newLineIndex = GetNewLineIndex(uri, fileText);
+
+            return PopulateTextRegionProperties(
+                newLineIndex,
+                inputRegion,
+                newLineIndex?.Text,
```
eddynaka, line 81: >newLineIndex?.Text [](start = 16, length = 18)

this could generate a null reference. Looked at the method and the other calls, we are not protecting it against null. Should we default empty?

eddynaka (approved):

PR #2189, opened 2020-12-16
Add tests for GetTopLevel change allows path to a file

### LINKED ISSUE #2188: Add tests for GitHelper GetTopLevel change allows path to a file in the repo
related change in PR https://github.com/microsoft/sarif-sdk/pull/2186

yongyan-gh:
fixes #2188 
added test cases which path is full path to a file & a directory.

### DIFF
--- a/src/Test.UnitTests.Sarif/GitHelperTests.cs
+++ b/src/Test.UnitTests.Sarif/GitHelperTests.cs
@@ -171,5 +171,56 @@ namespace Microsoft.CodeAnalysis.Sarif
             gitHelper.GitExePath = @"C:\dev";
             gitHelper.GitExePath.Should().Be(@"C:\dev");
         }
+
+        [Fact]
+        public void GetTopLevel_WhenRepoPathIsToAFile()
+        {
+            string pathToFile = typeof(GitHelperTests).Assembly.Location;
+            string repoRootPath = GitHelper.Default.GetTopLevel(pathToFile);
+            repoRootPath.Should().NotBeNull();
+            pathToFile.StartsWith(repoRootPath).Should().BeTrue();
+        }
+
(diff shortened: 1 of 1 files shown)


src/Test.UnitTests.Sarif/GitHelperTests.cs
```diff
@@ -171,5 +171,56 @@ public void GitExePath_WhenPathDoesNotExist_SettingManuallyShouldWork()
...
+            string pathToFile = typeof(GitHelperTests).Assembly.Location;
+            string repoRootPath = GitHelper.Default.GetTopLevel(pathToFile);
+            repoRootPath.Should().NotBeNull();
+            pathToFile.StartsWith(repoRootPath).Should().BeTrue();
+        }
+
+        [Fact]
+        public void GetTopLevel_WhenRepoPathIsToADictory()
+        {
+            string pathToDirectory = Environment.CurrentDirectory;
```
michaelcfanning, line 187: >Environment.CurrentDirecto [](start = 37, length = 26)

You should generate this path from typeof(GitHelperTests).Assembly.Location as you did in the previous test. Use Path.GetDirectoryName, of course, to get that data. The reason for this is that it will ensure greater determinism in test execution. Currently, various tests reset the current working directory! And so we can be subject to strange problems depending on the order of test execution and/or test execution environment. The paths generated to hold executing test binaries are stable for all tests.

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif/GitHelperTests.cs
yongyan-gh: good to learn about the issue in test execution, updated the test case.

PR #2200, opened 2020-12-23
Add HttpClient wrapper

### LINKED ISSUE #2193: Provide a mocking facility for SARIF2006.UrisShouldBeReachable
Currently, our unit tests invoke an actual web client that verifies reachability of test files that we pass to the validator. This approach is causing non-deterministic test failures (as a URI somewhere in our test code happened to be having some problems with availability).

We should use a mock object during testing instead. Because there's no good way to parameterize this rule, my suggestion is to:

1. Create an interface, IHttpClient::Get(uri) which returns a bool.
2. Create a simple wrapper object for the product code which wraps an HttpClient instance. Note that this wrapped instance could itself handle the caching that is currently driven by s_checkedUris.
3. Add an internal static field to the rule that in most cases holds an instance of this wrapper. In test code, we can inject a mock object instead. The property probably should be declared as thread-static.

Open to other approaches, this is the first thing that occurs to me. One thing: let's keep the mockable interface very simply, i.e., don't try to burn the async pattern. We don't need to return a complete HttpResponseMessage either, since we only consume its IsSuccessStatusCode property currently.

@eddynaka, another good opportunity to @yongyan-gh to spin up?

~~~
            HttpResponseMessage httpResponseMessage = s_httpClient.GetAsync(uriString).GetAwaiter().GetResult();
            s_checkedUris.Add(uriString, httpResponseMessage.IsSuccessStatusCode);
            return httpResponseMessage.IsSuccessStatusCode;
~~~

yongyan-gh:
Fixes #2193 
HttpClientWrapper can be used as mocking facility, below is code example:

```
var response = new HttpResponseMessage { StatusCode = HttpStatusCode.BadRequest, Content = new StringContent("Too bad request") };

var httpMock = new Mock<IHttpProvider>();

httpMock.Setup(x => x.GetAsync(It.IsAny<string>())).ReturnsAsync(response);

UrisShouldBeReachable rule = new UrisShouldBeReachable(httpMock.Object)
```

HttpClientWrapper constructor also accept a HttpClient object which can be a mock http client with mocked HttpMessageHandler for easier mocking. (All methods of HttpClient: GetAsync, PostAsync, PatchAsync, PutAsync, DeleteAsync, and SendAsync use the SendAsync method in the HttpMessageHandler internally) E.g.

```
            var handlerMock = new Mock<HttpMessageHandler>();
            var response = new HttpResponseMessage
            {
                StatusCode = HttpStatusCode.OK,
                Content = new StringContent(@"[{ ""id"": 1, ""title"": ""Cool post!""}, { ""id"": 100, ""title"": ""Some title""}]"),
            };

            handlerMock
               .Protected()
               .Setup<Task<HttpResponseMessage>>(
                  "SendAsync",
                  ItExpr.IsAny<HttpRequestMessage>(),
                  ItExpr.IsAny<CancellationToken>())
               .ReturnsAsync(response);
            var httpClient = new HttpClient(handlerMock.Object);

```

@michaelcfanning should I use static cache s_checkedUris here?


src/Sarif.Multitool.Library/Rules/SARIF2006.UrisShouldBeReachable.cs
```diff
@@ -32,9 +35,6 @@ public UrisShouldBeReachable()
             nameof(RuleResources.SARIF2006_UrisShouldBeReachable_Note_Default_Text)
         };
 
-        private static readonly HttpClient s_httpClient = new HttpClient();
-        private static readonly Dictionary<string, bool> s_checkedUris = new Dictionary<string, bool>();
```
eddynaka, line 36: >s_checkedUris [](start = 57, length = 13)

Keep this here, since this logic is for this rule only. The wrapper just need to do the calls

src/Sarif.Multitool.Library/Rules/SARIF2006.UrisShouldBeReachable.cs
```diff
@@ -100,13 +100,8 @@ private void AnalyzeUri(string uriString, string pointer)
...
         private bool IsUriReachable(string uriString)
         {
-            if (s_checkedUris.ContainsKey(uriString))
-            {
-                return s_checkedUris[uriString];
-            }
-
-            HttpResponseMessage httpResponseMessage = s_httpClient.GetAsync(uriString).GetAwaiter().GetResult();
-            s_checkedUris.Add(uriString, httpResponseMessage.IsSuccessStatusCode);
+            // http wrapper handles the cache all Uris has been accessed
```
eddynaka, line 103: >// http wrapper handles the cache all Uris has been accessed [](start = 12, length = 60)

Keep the cache logic in this file (just like my previous comment)

src/Sarif/HttpClientWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Send a GET request to the specified Uri as an asynchronous operation
+        /// if cache doesn't have this Uri entry, otherwize return it from cache
+        /// </summary>
+        /// <param name="requestUri">The Uri the request is sent to.</param>
+        /// <returns>The task object representing the asynchronous operation.</returns>
+        public Task<HttpResponseMessage> GetAsync(string requestUri)
+        {
+            // pls note the cache is case sensitive
+            if (s_checkedUris.ContainsKey(requestUri))
```
michaelcfanning, line 46: >s_checkedUris.ContainsKey(requestUri)) [](start = 16, length = 38)

I agree with Eddy, the caching should remain in the rule itself. It would be surprising for users if an HTTP client wrapper repeatedly returned a cached result for a request.

michaelcfanning (changes requested):
:clock1:

michaelcfanning:
This is a very nice start! Let's act on Eddy's feedback and then we can merge.

src/Sarif.Multitool.Library/Rules/SARIF2006.UrisShouldBeReachable.cs
yongyan-gh: done

src/Sarif.Multitool.Library/Rules/SARIF2006.UrisShouldBeReachable.cs
yongyan-gh: done

src/Sarif/HttpClientWrapper.cs
yongyan-gh: have done the change

eddynaka (approved):

PR #2203, opened 2020-12-29
Cache notifications in the caching logger.

michaelcfanning:
@eddynaka 

Need to add tests that comprehensively fill out the caching logger (we don't handle basic messages still, for example).


eddynaka (approved):

eddynaka:
> @eddynaka
> 
> Need to add tests that comprehensively fill out the caching logger (we don't handle basic messages still, for example).

Hi @michaelcfanning , updated your pull request and added a few tests and some null checks.

PR #2207, opened 2020-12-31
Fixing concurrency issues during analysis

eddynaka:


src/Sarif/RuleUtilities.cs
```diff
@@ -30,10 +29,10 @@ public static Result BuildResult(FailureLevel level, IAnalysisContext context, R
                 ? ResultKind.Fail
                 : ResultKind.None;
 
-            return BuildResult(level, kind, context, region, ruleMessageId, arguments);
+            return BuildResult(level, kind, context, region, ruleMessageId, reportingDescriptor: null, arguments);
         }
 
-        public static Result BuildResult(FailureLevel level, ResultKind kind, IAnalysisContext context, Region region, string ruleMessageId, params string[] arguments)
+        public static Result BuildResult(FailureLevel level, ResultKind kind, IAnalysisContext context, Region region, string ruleMessageId, ReportingDescriptor reportingDescriptor = null, params string[] arguments)
```
michaelcfanning, line 35: >ReportingDescriptor [](start = 141, length = 19)

I'm not sure this is right. The context object already has a reporting descriptor, why should we provide another one? I think it's probably better to create copies of context objects than to depend on the context except for this additional reporting descriptor object (which in your example below can override the context instance).

src/Sarif/Notes.cs
```diff
@@ -21,11 +21,15 @@ public static void LogNotApplicableToSpecifiedTarget(IAnalysisContext context, s
...
             // '{0}' was not evaluated for check '{1}' because the analysis
             // is not relevant for the following reason: {2}.
-            context.Logger.Log(context.Rule,
+
+            var rule = context.Rule.DeepClone();
+
+            context.Logger.Log(rule,
                 RuleUtilities.BuildResult(ResultKind.NotApplicable, context, null,
                     nameof(SdkResources.NotApplicable_InvalidMetadata),
+                    reportingDescriptor: rule,
```
michaelcfanning, line 30: >reportingDescriptor: rule [](start = 20, length = 25)

I don't understand why you'd need to provide this object here after cloning. ?

eddynaka: I did this because the rule and the context are changing. So, when I cloned the rule I had to use it... I'm not sure as well if that is the best way...

michaelcfanning: Eddy, looks like you're not using the GH CodeFlow chrome extension, is that so?

---
In reply to: [550514500](https://github.com/microsoft/sarif-sdk/pull/2207#discussion_r550514500) [](ancestors = 550514500)

src/Sarif/RuleUtilities.cs
eddynaka: reverted. thanks for reviewing!

---
In reply to: [550509927](https://github.com/microsoft/sarif-sdk/pull/2207#discussion_r550509927) [](ancestors = 550509927)

src/Sarif/Notes.cs
eddynaka: reverted!

---
In reply to: [550511491](https://github.com/microsoft/sarif-sdk/pull/2207#discussion_r550511491) [](ancestors = 550511491)

eddynaka: I was in the phone, so used the github interface. but yeah, i know the codeflow extension :)
thanks!

---
In reply to: [550532276](https://github.com/microsoft/sarif-sdk/pull/2207#discussion_r550532276) [](ancestors = 550532276,550514500)

michaelcfanning (approved):
:shipit:

PR #2215, opened 2021-01-04
Adding lock to prevent concurrency issue

eddynaka:
## Description
When we have many threads working with our cache class, we can face a concurrency issue since we are using LinkedList, which is not thread safe. With our latest pull request in the cache area, we changed from dictionary to concurrencyDictionary, that minimized the issue, but the issue was still happening.

Added a new test so we can verify if new changes to the fileRegionsCache keeps working.


michaelcfanning (approved):
:shipit:

PR #2217, opened 2021-01-04
ConsoleLogger will expose ExceptionData

### LINKED ISSUE #2216: Console logger should emit exception.ToString() for notifications
.. if they are available. 

@eddynaka

eddynaka:
Fixes #2216

### DIFF
--- a/src/Sarif/Writers/ConsoleLogger.cs
+++ b/src/Sarif/Writers/ConsoleLogger.cs
@@ -6,8 +6,6 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 
-using Newtonsoft.Json;
-
 namespace Microsoft.CodeAnalysis.Sarif.Writers
 {
     public class ConsoleLogger : IAnalysisLogger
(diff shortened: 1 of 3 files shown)


src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -322,6 +324,12 @@ static internal string FormatNotificationMessage(Notification notification, stri
 
             sb.Append(notification.Message.Text);
 
+            if (notification.Exception != null)
+            {
+                sb.AppendLine();
+                sb.AppendLine(JsonConvert.SerializeObject(notification.Exception));
```
eddynaka, line 330: >JsonConvert.SerializeObject(notification.Exception) [](start = 30, length = 51)

Since Exception isn't an exception class, its an ExceptionData class from sarif, it does not expose a ToString which generate all the fields. With that in mind, I switched to serializing it.

michaelcfanning: Let's write an ExceptionData.ToString() implementation here that mirrors System.Exception.ToString() behavior

---
In reply to: [551626348](https://github.com/microsoft/sarif-sdk/pull/2217#discussion_r551626348) [](ancestors = 551626348)

eddynaka: done! thanks for guiding me!

---
In reply to: [551649906](https://github.com/microsoft/sarif-sdk/pull/2217#discussion_r551649906) [](ancestors = 551649906,551626348)

michaelcfanning (approved):
:shipit:

PR #2221, opened 2021-01-05
Bump 2.3.14

eddynaka:


src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v2.3.14** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.3.14) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.3.14) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.3.14) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.3.14) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/2.3.14)
+* BUGFIX: Fix concurrency issue in when using `Cache` [#2215](https://github.com/microsoft/sarif-sdk/pull/2215)
+* FEATURE: `ConsoleLogger` will print exception if that exists [#2217](https://github.com/microsoft/sarif-sdk/pull/2217)
+* BUGFIX: Fix WebRequest parameters parse [#2219](https://github.com/microsoft/sarif-sdk/pull/2219)
```
michaelcfanning, line 6: >WebRequest [](start = 14, length = 10)

it's good to mention the symptoms of the break here, as that is what people may search the readme for. In this case, something like 'Fix `WebRequest` parameters parse that resulted in regex hang.'

In other cases, you'd mention the specific exception type (if an unhandled exception was what users saw).

michaelcfanning (approved):
:shipit:

PR #2224, opened 2021-01-08
Fix hashing messages not respecting --quiet

### LINKED ISSUE #2213: hashing message doesn't respect the --quiet flag.
in Sarif\HashUtilities, we emit messages like this when computing file hashes.

>Computing file hashes...
>Hash computation complete.

These messages should not appear when the --quiet flag is passed.

It'd be nice to add a unit test for this, why not? :) 

@eddynaka, another very good issue for onboarding someone new.

jameswinkler:
Addressing issue #2213 (https://github.com/microsoft/sarif-sdk/issues/2213), this simply adds an optional parameter to the HashUtility method, defaulting to false, which allows callers to suppress console messages.

Fixes #2213


eddynaka (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #2231, opened 2021-01-12
Move code from BinSkim, refactor, add unit tests

### LINKED ISSUE #2211: AnalyzeCommand base class should handle obsolete `ComputeFileHashes` arg
This argument is obsolete in the driver framework, replaced by `--insert Hashes`. We've marked it as obsolete and it warns on compilation dependency. But I just updated BinSkim to the latest SDK and the driver framework isn't providing backwards compatibility for use of the obsolete arg.

I've added the following code to BinSkim but really it should be in `AnalyzeCommand.Run`:

~~~
#pragma warning disable CS0618 // Type or member is obsolete
            if (analyzeOptions.ComputeFileHashes)
#pragma warning restore CS0618
            {
                OptionallyEmittedData dataToInsert = analyzeOptions.DataToInsert.ToFlags();
                dataToInsert |= OptionallyEmittedData.Hashes;

                analyzeOptions.DataToInsert  = 
                    dataToInsert.ToString().Split('|')
                        .Select(d => { return Enum.Parse<OptionallyEmittedData>(d); })
                        .ToList();
~~~

jameswinkler:
Fixes #2211 

Move code from BinSkim per issue and add all necessary infrastructure to raise warning.  Some light refactoring.

### DIFF
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -77,7 +77,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 _rootContext = CreateContext(options, logger, RuntimeErrors);
                 if (options.ComputeFileHashes)
                 {
-                    Warnings.LogObsoleteOption(_rootContext, options.ComputeFileHashes.ToString(), SdkResources.ComputeFileHashes_ReplaceInsertHashes);
+                    Warnings.LogObsoleteOption(_rootContext, "--hashes", SdkResources.ComputeFileHashes_ReplaceInsertHashes);
 #pragma warning restore CS0618
                 }
 
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -57,23 +57,25 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         public override int Run(TOptions options)
         {
-#pragma warning disable CS0618 // Type or member is obsolete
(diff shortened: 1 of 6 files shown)


src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -57,10 +57,30 @@ public string DefaultConfigurationPath
...
+                // 0. Initialize an common logger that drives all outputs. This
+                //    object drives logging for console, statistics, etc.
+                using (AggregatingLogger logger = InitializeLogger(options))
             {
+                //  Once the logger has been correctly initialized, we can raise a warning
+                _rootContext = CreateContext(options, logger, RuntimeErrors);
+                if (options.ComputeFileHashes)
+                {
+                    Warnings.LogObsoleteOption(_rootContext, "--hashes", SdkResources.ComputeFileHashes_ReplaceInsertHashes);
+#pragma warning restore CS0618
```
eddynaka, line 81: i think we should do the pragma just for the if option instead of the entire block of code.

jameswinkler: Were you thinking two separate pragma's?  We refer to the ComputeFileHashes property twice so we need it to cover both.

eddynaka: exactly. just an style issue. otherwise, its good :)

---
In reply to: [555959576](https://github.com/microsoft/sarif-sdk/pull/2231#discussion_r555959576) [](ancestors = 555959576)

jameswinkler: Done

eddynaka (approved):

PR #2235, opened 2021-01-13
Accepting relative URLs

eddynaka:


src/Sarif/RuleUtilities.cs
```diff
@@ -60,16 +60,15 @@ public static Result BuildResult(FailureLevel level, ResultKind kind, IAnalysisC
...
+            if (context.TargetUri != null)
             {
                 result.Locations = new List<Location> {
                     new Sarif.Location {
                         PhysicalLocation = new PhysicalLocation
                         {
                             ArtifactLocation = new ArtifactLocation
                             {
-                                Uri = new Uri(targetPath)
+                                Uri = context.TargetUri.IsAbsoluteUri ?  new Uri(context.TargetUri.LocalPath) : context.TargetUri
```
michaelcfanning, line 71: >Uri [](start = 77, length = 3)

can we just reuse context.TargetYri here?

src/Sarif/ExtensionMethods.cs
```diff
@@ -173,7 +175,8 @@ public static string GetFileName(this Uri uri)
         {
             if (!uri.IsAbsoluteUri)
             {
-                throw new InvalidOperationException();
+                var newAbsoluteUri = new Uri(BaseUri, uri.OriginalString);
+                return Path.GetFileName(newAbsoluteUri.LocalPath);
```
michaelcfanning, line 179: >GetFileName [](start = 28, length = 11)

make sure this works with a path like \blah.txt and blah.txt (no prefixed slash).

src/Sarif/ExtensionMethods.cs
```diff
@@ -16,6 +16,8 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public static class ExtensionMethods
     {
+        private static readonly Uri BaseUri = new Uri("https://github.com/microsoft/sarif-sdk/");
+
```
michaelcfanning, line 20: use [https://example.com](https://example.com) here (and provide a comment). actually! how about declaring this as an inlined string const in your helper?

michaelcfanning (approved):
:shipit:

PR #2242, opened 2021-01-14
Multifile processing

michaelcfanning:
* Push inline into common base class.
* Create a base class for performing multifile log file processing and test abstraction in the absolute URI command.

@eddynaka 
@jameswinkler 
@yongyan-gh


src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -71,7 +71,6 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
         public IEnumerable<string> PluginFilePaths { get; set; }
 
         [Option(
-            'i',
```
michaelcfanning, line 74: >', [](start = 14, length = 2)

This short-cut interferes with 'inline'. This is a breaking change and needs to be mentioned in the release notes, will add.

src/Sarif.Driver/Sdk/MultifileCommandBase.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.Driver
+{
+    public abstract class MultifileCommandBase : CommandBase
```
michaelcfanning, line 11: >MultifileCommandBase [](start = 26, length = 20)

all this code was moved/generalized from AbsoluteUriCommand

eddynaka (approved):

PR #2248, opened 2021-01-19
Relaxing GH1005

### LINKED ISSUE #2197: Relax GH1005
GH1005 requires file paths (`location.PhysicalLocation.ArtifactLocation.Uri`) to be either relative URIs or start with the `file://` schema.

In Code Scanning, we ran into tools (e.g., bandit via sast-scan) that use a full URI as file path such as:

```
"artifactLocation": {
                  "uri": "https://github.com/mona/electron-test-action/blob/52712737fce704582b8c229c14dded478730eb81/script/release/uploaders/upload-index-json.py"
                },
```

Such a URI can be interpreted as the "GitHub path". Code Scanning is able to convert such a URI into relative URI, only if the invocation section includes a `workingDirectory` that can be used as a base for the URI (i.e., it is a prefix of the path uri), e.g.,:

```
"invocations": [
          {
            "executionSuccessful": true,
            "workingDirectory": {
              "uri": "https://github.com/mona/electron-test-action/blob/52712737fce704582b8c229c14dded478730eb81/"
            }
          }
        ],
```

### Cases
File path (`location.PhysicalLocation.ArtifactLocation.Uri`)
1. is a relative URI :heavy_check_mark:
2. is an absolute URI with `file:` schema :heavy_check_mark: 
3. is an absolute URI with a schema that is not `file`:
    1. Has a `invocations.workingDirectory.uri` that is a prefix of the file path :heavy_check_mark: 
    2. All other cases :x: 
  
### Example
[Here is an example (good.txt)](https://github.com/microsoft/sarif-sdk/files/5728925/good.txt) that we consider valid but the Validator does not (using the "GitHub ingestion rules" extension).

[Here is an example (bad.txt)](https://github.com/microsoft/sarif-sdk/files/5728948/bad.txt) that we do not accept but is valid SARIF, and should continue to be rejected if the user selects the "GitHub extension rules". This is the same example as `good.txt` with the `invocation` section removed.

### Technical Notes

I believe https://github.com/microsoft/sarif-sdk/pull/2128 was on the right track, but the check [here]( https://github.com/microsoft/sarif-sdk/blob/f0dbf19d49e9beb14cfef789a5eae63a779bd67c/src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs#L73) is too weak. We want `workingDirectoryUri` to be not null AND not a prefix of `uri.OriginalString`.

eddynaka:
Fixes #2197

### DIFF
--- a/src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
+++ b/src/Sarif.Multitool.Library/Rules/GH1005.LocationsMustBeRelativeUrisOrFilePaths.cs
@@ -31,12 +31,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool.Rules
 
         public override bool EnabledByDefault => false;
 
-        private List<Uri> workingDirectoriesUri;
+        private List<Uri> workingDirectoryUris;
 
         protected override void Analyze(Run run, string runPointer)
         {
-            this.workingDirectoriesUri = new List<Uri>();
-            GetWorkingDirectoryUri(run);
+            this.workingDirectoryUris = new List<Uri>();
+            GetWorkingDirectoryUris(run);
         }
 
         protected override void Analyze(Result result, string resultPointer)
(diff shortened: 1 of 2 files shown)


marcogario (approved):
Looks good! Thanks!

michaelcfanning (approved):
:shipit:

PR #2253, opened 2021-01-21
Enable NETAnalyzers

eddynaka:
## Changes
- Enable NETAnalyzers
- Fix a few error / warning / messages

## Why
- From our security rules, we should use Roslyn analyzers in IDE/build

### DIFF
--- a/src/Sarif.Multitool.Library/MergeCommand.cs
+++ b/src/Sarif.Multitool.Library/MergeCommand.cs
@@ -235,7 +235,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                 new FixupVisitor().VisitSarifLog(sarifLog);
             }
 
-            _mergeLogsChannel.Writer.WriteAsync(sarifLog);
+            _mergeLogsChannel.Writer.TryWrite(sarifLog);
             Interlocked.Decrement(ref _filesToProcessCount);
         }
 
--- a/src/Sarif/Map/JsonMapSettings.cs
+++ b/src/Sarif/Map/JsonMapSettings.cs
@@ -30,7 +30,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
         /// <param name="mapMaximumSizeBytes">Map size limit, in bytes (ex: 10 * JsonMapSettings.Megabyte), 0 for no limit</param>
         public JsonMapSettings(double mapSizeRatio, double mapMaximumSizeBytes = 0)
         {
-            if (mapSizeRatio <= 0 || mapSizeRatio > 100) { throw new ArgumentOutOfRangeException(nameof(mapSizeRatio), "mapSizeRatio must be > 0 and <= 100."); }
+            if (mapSizeRatio <= 0 || mapSizeRatio > 100) { throw new ArgumentOutOfRangeException(nameof(mapSizeRatio), $"{nameof(mapSizeRatio)} must be > 0 and <= 100."); }
             MapDesiredSizeRatio = mapSizeRatio;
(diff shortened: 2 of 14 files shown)


src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -235,7 +235,7 @@ private void ProcessInputSarifLog(string filePath)
                 new FixupVisitor().VisitSarifLog(sarifLog);
             }
 
-            _mergeLogsChannel.Writer.WriteAsync(sarifLog);
+            _mergeLogsChannel.Writer.TryWrite(sarifLog);
```
jameswinkler, line 238: Does it matter that we've gone from an asynchronous command to a synchronous one?  Also, we don't do anything with the boolean indicating whether or not the write was successful.

src/Sarif/Map/JsonMapSettings.cs
```diff
@@ -30,7 +30,7 @@ public class JsonMapSettings
         /// <param name="mapMaximumSizeBytes">Map size limit, in bytes (ex: 10 * JsonMapSettings.Megabyte), 0 for no limit</param>
         public JsonMapSettings(double mapSizeRatio, double mapMaximumSizeBytes = 0)
         {
-            if (mapSizeRatio <= 0 || mapSizeRatio > 100) { throw new ArgumentOutOfRangeException("maxFileSizePercentage must be > 0 and <= 100."); }
+            if (mapSizeRatio <= 0 || mapSizeRatio > 100) { throw new ArgumentOutOfRangeException(nameof(mapSizeRatio), "mapSizeRatio must be > 0 and <= 100."); }
```
jameswinkler, line 33: In one case, we use nameof to get the variable name, but in the other it is hard-coded.  Perhaps store nameof(mapSizeRatio) in a variable and use string.format to add it to the front of the string?

src/Sarif.Multitool.Library/MergeCommand.cs
eddynaka: using async method without await or task could lead to some issues: "CA2012 ValueTask instances returned from method calls should always be used, typically awaited. Not doing so often represents a functional bug, but even if it doesn't, it can result in degraded performance if the target method pools objects for use with ValueTasks.".

---
In reply to: [562075944](https://github.com/microsoft/sarif-sdk/pull/2253#discussion_r562075944) [](ancestors = 562075944)

src/Sarif/Map/JsonMapSettings.cs
eddynaka: thanks for the suggestion. changed to interpolation :)

---
In reply to: [562078645](https://github.com/microsoft/sarif-sdk/pull/2253#discussion_r562078645) [](ancestors = 562078645)

src/Sarif.Multitool.Library/MergeCommand.cs
jameswinkler: That makes sense.  Are you concerned that the write might fail and we don't tell anyone about it?

---
In reply to: [562084169](https://github.com/microsoft/sarif-sdk/pull/2253#discussion_r562084169) [](ancestors = 562084169,562075944)

eddynaka: for now, no. if that happens, we will have to think carefully, because we would need to create a retry logic. or something related

---
In reply to: [562087447](https://github.com/microsoft/sarif-sdk/pull/2253#discussion_r562087447) [](ancestors = 562087447,562084169,562075944)

jameswinkler (approved):
:shipit:

PR #2254, opened 2021-01-21
Fixing null reference in SARIF1012 validation

eddynaka:
## Changes
- When we have a result without message arguments and the rule contains arguments, we were causing a null reference, since we were trying to get the actual count of arguments that the result had. With that, if arguments are null, the count is zero.


michaelcfanning (approved):
:shipit:

PR #2262, opened 2021-01-23
Change validation in BaseLogger and correct unit tests and code 

jameswinkler:
"kinds" must not be empty, and many of our unit tests left it that way.  I put some default values in to most of them.  This probably could have been abstracted a little better, but I wanted to handle that separately.

### DIFF
--- a/src/Sarif/Writers/BaseLogger.cs
+++ b/src/Sarif/Writers/BaseLogger.cs
@@ -23,12 +23,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
 
         private void ValidateParameters()
         {
-            if (_resultKinds.Count==0 || (_resultKinds.Count==1 && _resultKinds[0] == ResultKind.None))
+            if (_resultKinds.Count == 0 || (_resultKinds.Count == 1 && _resultKinds[0] == ResultKind.None))
             {
                 throw new ArgumentException("At least one kind is required");
             }
 
-            bool failureLevelsEffectivelyEmpty =    _failureLevels == null
+            bool failureLevelsEffectivelyEmpty = _failureLevels == null
                                                     || _failureLevels.Count == 0
                                                     || (_failureLevels.Count == 1 && _failureLevels[0] == FailureLevel.None);
 
--- a/src/Sarif/Writers/BaseLogger.cs
+++ b/src/Sarif/Writers/BaseLogger.cs
@@ -15,36 +15,35 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
         protected BaseLogger(IEnumerable<FailureLevel> failureLevels,
             IEnumerable<ResultKind> resultKinds)
         {
-            if (failureLevels == null || failureLevels.Count() == 0)
-            {
-                _failureLevels = new List<FailureLevel> { FailureLevel.Error, FailureLevel.Warning };
-            }
-            else
-            {
-                _failureLevels = failureLevels.ToList();
-            }
-
-            if (resultKinds == null || resultKinds.Count() == 0)
-            {
-                _resultKinds = new List<ResultKind> { ResultKind.Fail };
-            }
-            else
-            {
-                _resultKinds = resultKinds.ToList();
-            }
+            _failureLevels = failureLevels?.ToList() ?? new List<FailureLevel>();
+            _resultKinds = resultKinds?.ToList() ?? new List<ResultKind>();
 
             ValidateParameters();
         }
 
         private void ValidateParameters()
         {
-            //  If we got here, both resultKinds and failurelevels can neither be null nor empty.
-            //  If resultKinds does not include "fail" then failureLevels MUST be none/zero
-            if (!_resultKinds.Contains(ResultKind.Fail))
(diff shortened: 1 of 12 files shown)


src/Sarif/Writers/BaseLogger.cs
```diff
@@ -15,36 +15,35 @@ public abstract class BaseLogger
...
 
             ValidateParameters();
         }
 
         private void ValidateParameters()
         {
-            //  If we got here, both resultKinds and failurelevels can neither be null nor empty.
-            //  If resultKinds does not include "fail" then failureLevels MUST be none/zero
-            if (!_resultKinds.Contains(ResultKind.Fail))
+            if (_resultKinds.Count==0 || (_resultKinds.Count==1 && _resultKinds[0] == ResultKind.None))
```
eddynaka, line 26: >== [](start = 34, length = 2)

use ctrl + k + d to auto ident

eddynaka:
>                 baseLoggerTestConcrete = new BaseLoggerTestConcrete(new List<FailureLevel> { FailureLevel.Error },

We could change the try/catch for something similar to Assert.ShouldNotThrow<ArgumentException>(() => call);

---
Refers to: src/Test.UnitTests.Sarif/Writers/BaseLoggerTests.cs:30 in 6957420. [](commit_id = 69574204712790ebdace0ed0f25e0049f71e102b, deletion_comment = False)

eddynaka (approved):
:shipit:

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -975,7 +1013,10 @@ private void TestForLoggingOption(LogFilePersistenceOptions loggingOption)
                 SarifLogger logger;
 
                 // Validates overload that accept a path argument.
-                using (logger = new SarifLogger(fileName, loggingOption))
+                using (logger = new SarifLogger(fileName, 
+                                                loggingOption,
+                                                levels: new List<FailureLevel> { FailureLevel.Warning, FailureLevel.Error },
```
michaelcfanning, line 1018: >levels: new List<FailureLevel> { FailureLevel.Warning, FailureLevel.Error }, [](start = 48, length = 76)

I would have expected these args to be defaults on the SarifLogger class, so you wouldn't have had to make them explicit in the tests. To provide default enumerables, check for null in the relevant ctor and initialize a new List instance on detecting null.

michaelcfanning:
>     public class SarifLoggerTests : JsonTests

I don't see any tests that prove that --kind and --level provide new granularity that allows for log files to, say, receive NotApplicable messages and Fail/Errors only.

---
Refers to: src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs:25 in 6957420. [](commit_id = 69574204712790ebdace0ed0f25e0049f71e102b, deletion_comment = False)

src/Sarif/Writers/BaseLogger.cs
jameswinkler: Executed buildandtest.cmd instead

---
In reply to: [563044550](https://github.com/microsoft/sarif-sdk/pull/2262#discussion_r563044550) [](ancestors = 563044550)

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
jameswinkler: I thought about putting default values but the validation in BaseLogger is the *only* validation that is run.  That is, BaseLogger is the one place we've defined rules for what command line options can be input.  BaseLogger, and by extension, SarifLogger isn't asking for its caller to "tell me what you think that I need to know", it's asking for the true values of the options passed so it can accurately validate them.  If we put default values in SarifLogger, we are correcting bad input.  And if we're going to correct bad input, why are we throwing exceptions at all?  Why not just overwrite their options if they ever pass options that don't make sense?

---
In reply to: [563215266](https://github.com/microsoft/sarif-sdk/pull/2262#discussion_r563215266) [](ancestors = 563215266)

jameswinkler:
>     public class SarifLoggerTests : JsonTests

Add a giant unit test that tests the two scenarios mentioned plus the default, I can add more if desired.

---
In reply to: [766236555](https://github.com/microsoft/sarif-sdk/pull/2262#issuecomment-766236555) [](ancestors = 766236555)

---
Refers to: src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs:25 in 6957420. [](commit_id = 69574204712790ebdace0ed0f25e0049f71e102b, deletion_comment = False)

jameswinkler:
>                 baseLoggerTestConcrete = new BaseLoggerTestConcrete(new List<FailureLevel> { FailureLevel.Error },

Done.

---
In reply to: [765897281](https://github.com/microsoft/sarif-sdk/pull/2262#issuecomment-765897281) [](ancestors = 765897281)

---
Refers to: src/Test.UnitTests.Sarif/Writers/BaseLoggerTests.cs:30 in 6957420. [](commit_id = 69574204712790ebdace0ed0f25e0049f71e102b, deletion_comment = False)

PR #2263, opened 2021-01-25
Reuse fileRegionsCache

eddynaka:


src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -117,8 +117,6 @@ public override PhysicalLocation VisitPhysicalLocation(PhysicalLocation node)
                     resolvedUri = artifactLocation.Uri;
                 }
 
-                if (!resolvedUri.IsAbsoluteUri) { goto Exit; }
```
eddynaka, line 120: >if (!resolvedUri.IsAbsoluteUri) { goto Exit; } [](start = 16, length = 46)

Since our Uri can be relative, I'm removing this so we can generate contextRegion

michaelcfanning (approved):
:shipit:

PR #2273, opened 2021-01-27
#2229 A little refactoring, add new class for interpretting options

### LINKED ISSUE #2229: SARIF_LEVEL_OVERRIDE, SARIF_KIND_OVERRIDE
When we complete #2228, we should add support for environment variables that completely replace anything on the command-line if they are encountered. So,

set SARIF_LEVEL_OVERRIDE=Error;Warning;Note

could be used to automatically opt into 'experimental' failure messages (which has the 'note' level set on them and would otherwise be filtered). this is useful to look at the behavior of existing analysis systems without requiring that tools be reconfigured.

Notes:
* We should think about a name convention that maps to arg names. I've chosed SARIF_[ARGNAME]_OVERRIDE. We could use this same standard for any/all arguments we decide can be overridden.
* It is pretty dangerous to conditionally drive tool behavior based on environment variables. One protection against this is to require some explicit command-line argument that allows for these variables to be read/override arguments. That requires users to update all their command-lines to enable the functionality before it works.
* Requiring opt-in to this override policy could be burdensome. If we do not require it, we should definitely emit an extremely prominent notification which can't be disabled via all our magical mechanisms that reports that an argument is being overridden. Error notifications in particular can never be disabled. If we raise an error notification for this, however, the experiment might break systems that halt on errors, preventing users from experimenting with lighting up non-default analysis. So as usual, there's more nuance here than one might think.

jameswinkler:
Fixes #2229 

Move analyzeTestOptions to same project as al other options.  Add EnvironmentVariableGetter and OptionsInterpretter classes


src/Sarif.Multitool.Library/OptionsInterpretter.cs
```diff
@@ -0,0 +1,53 @@
...
+        public OptionsInterpretter() : this (new EnvironmentVariableGetter())
+        {
+
+        }
+        
+        public OptionsInterpretter(IEnvironmentVariableGetter environmentVariableGetter)
+        {
+            if (environmentVariableGetter == null)
+            {
+                throw new ArgumentException("Required parameter", nameof(environmentVariableGetter));
```
eddynaka, line 24: >ArgumentException [](start = 26, length = 17)

Change to argumentNullException. With that, u can you just the nameof(xyz)

eddynaka: you could also do this: _environmentVariableGetter = environmentVariableGetter ??  throw new ArgumentNullException(nameof(enviromentVariableGetter));

---
In reply to: [566070993](https://github.com/microsoft/sarif-sdk/pull/2273#discussion_r566070993) [](ancestors = 566070993)

src/Sarif.Multitool.Library/OptionsInterpretter.cs
```diff
@@ -0,0 +1,53 @@
...
+            //  Level and Kind are the only two properties we're reading from env variables for now.  See if they are populated.
+            IEnumerable<ResultKind> kindsFromEnv = getOptionEnumFromEnvVar<ResultKind>(nameof(analyzeOptionsBase.Kind));
+            IEnumerable<FailureLevel> levelsFromEnv = getOptionEnumFromEnvVar<FailureLevel>(nameof(analyzeOptionsBase.Level));
+
+            //  Union in C# is distinct
+            analyzeOptionsBase.Kind = analyzeOptionsBase.Kind.Union(kindsFromEnv);
+            analyzeOptionsBase.Level = analyzeOptionsBase.Level.Union(levelsFromEnv);
+        }
+
+        private IEnumerable<T> getOptionEnumFromEnvVar<T>(string propertyName)
```
eddynaka, line 43: >getOptionEnumFromEnvVar [](start = 31, length = 23)

GetOptionEnumFromEnvVar

src/Sarif.Multitool.Library/OptionsInterpretter.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return _environmentVariableGetter.GetEnvironmentVariable(FormEnvironmentVariableName(propertyName))?.Split(';')?.Select(x => Enum.Parse(typeof(T), x, ignoreCase: true))?.ToList()?.Cast<T>();
```
eddynaka, line 45: >_environmentVariableGetter.GetEnvironmentVariable [](start = 19, length = 49)

lets wrap. too long to read in one line

src/Sarif.Multitool.Library/OptionsInterpretter.cs
```diff
@@ -0,0 +1,53 @@
...
+        private readonly IEnvironmentVariableGetter _environmentVariableGetter;
+
+        public void ConsumeEnvVarsAndInterpretOptions(AnalyzeOptionsBase analyzeOptionsBase)
+        {
+            //  Level and Kind are the only two properties we're reading from env variables for now.  See if they are populated.
+            IEnumerable<ResultKind> kindsFromEnv = getOptionEnumFromEnvVar<ResultKind>(nameof(analyzeOptionsBase.Kind));
+            IEnumerable<FailureLevel> levelsFromEnv = getOptionEnumFromEnvVar<FailureLevel>(nameof(analyzeOptionsBase.Level));
+
+            //  Union in C# is distinct
+            analyzeOptionsBase.Kind = analyzeOptionsBase.Kind.Union(kindsFromEnv);
```
eddynaka, line 39: >Union(kindsFromEnv); [](start = 62, length = 20)

do we need to check for null?

src/Sarif.Multitool.Library/OptionsInterpretter.cs
jameswinkler: Done.

---
In reply to: [566071879](https://github.com/microsoft/sarif-sdk/pull/2273#discussion_r566071879) [](ancestors = 566071879,566070993)

src/Sarif.Multitool.Library/OptionsInterpretter.cs
jameswinkler: Done.

---
In reply to: [566072089](https://github.com/microsoft/sarif-sdk/pull/2273#discussion_r566072089) [](ancestors = 566072089)

src/Sarif.Multitool.Library/OptionsInterpretter.cs
jameswinkler: Done.

---
In reply to: [566072701](https://github.com/microsoft/sarif-sdk/pull/2273#discussion_r566072701) [](ancestors = 566072701)

src/Sarif.Multitool.Library/OptionsInterpretter.cs
jameswinkler: Done.

---
In reply to: [566073724](https://github.com/microsoft/sarif-sdk/pull/2273#discussion_r566073724) [](ancestors = 566073724)

eddynaka (approved):

PR #2280, opened 2021-01-30
Adding deduplication in merge command

### LINKED ISSUE #2279: Author a 'remove duplicate results' command
Provide a mechanism for removing duplicate results from a SARIF log.  

This useful primitive operation belongs in the SARIF SDK.  Windows intends to use in several scenarios.

Duplicates from the CPP compiler are known to occur in certain situations:
- You will see duplication for template expanded code.  If the template is used for n types, you will see one issue raised per each specialization.
- You will also see duplication for inline functions from headers that are included into multiple sources.

Some duplicates may not appear until multiple per-compilation-unit SARIF logs are merged.

Removing dupes is likely less difficult than avoiding dupes.

Duplicates can be removed in certain situations.  For tools that emit source file, line, column #’s (and range?), the SDK is expected to provide reliable deduplication, at the cost of retaining some state while emitting/merging logs.

SARIF from tools that don’t emit column details are not able to be deduped.  One can never be sure whether one is actually uncovering dupes (due to a merge scenario, for example) or if the tool itself has actually found multiple issues. In general, tools are encouraged to provide intra-line regions.

eddynaka:
Fixes #2279

### DIFF
--- a/src/Sarif.Multitool.Library/MergeCommand.cs
+++ b/src/Sarif.Multitool.Library/MergeCommand.cs
@@ -26,12 +26,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
         private Channel<string> _logLoadChannel;
         private Channel<SarifLog> _mergeLogsChannel;
         private readonly Dictionary<string, Run> _ruleIdToRunsMap;
+        private readonly Dictionary<string, HashSet<Result>> _ruleIdToResultsMap;
         private readonly Dictionary<string, SarifLog> _idToSarifLogMap;
 
         public MergeCommand(IFileSystem fileSystem = null)
         {
             _fileSystem = fileSystem ?? Sarif.FileSystem.Instance;
             _ruleIdToRunsMap = new Dictionary<string, Run>();
+            _ruleIdToResultsMap = new Dictionary<string, HashSet<Result>>();
             _idToSarifLogMap = new Dictionary<string, SarifLog>();
         }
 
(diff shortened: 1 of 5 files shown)


src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -173,14 +175,17 @@ private async Task<bool> MergeSarifLogsAsync()
 
                                 if (!_ruleIdToRunsMap.TryGetValue(key, out Run splitRun))
                                 {
-                                    splitRun = _ruleIdToRunsMap[key] = new Run()
-                                    {
-                                        Tool = emptyRun.Tool,
-                                        Results = new List<Result>()
-                                    };
+                                    emptyRun.Results = new List<Result>();
+                                    splitRun = _ruleIdToRunsMap[key] = emptyRun;
```
eddynaka, line 179: >emptyRun [](start = 71, length = 8)

Instead of creating a new run, I changed to emptyRun object to preserve artifacts and other fields

michaelcfanning, line 179: >emptyRun [](start = 71, length = 8)

This might be the right start but I think we're not comprehensively handling all that we should. What are we doing about rule indexes and artifact location indexes for example? When we add a result from one run *into* another, we should be fixing up those indices, expanding the target run with new rule/artifact data if it comes in, etc. Our baselining/result matching code does this and has some utilities for it. This is all to say: we probably have more work to do in this command (which previously simply moved entire run objects around, a very simple operation). Let's open an issue noting the open work and keep moving. :)

michaelcfanning (approved):
:shipit:

PR #2285, opened 2021-02-09
Merging artifacts during merge command

### LINKED ISSUE #2284: Merge operations are zapping artifact data
We need to inline artifact data into the merged result, minimally, and zap the artifact index if we do not populate this data.

Ideally, though, we would move the artifact entry from the merge source to the merge destination (as our baseliner does) and update the merged results artifact index to accurately reflect a pointer to the merged table entry.

eddynaka:
Fixes #2284

### DIFF
--- a/src/Sarif.Multitool.Library/MergeCommand.cs
+++ b/src/Sarif.Multitool.Library/MergeCommand.cs
@@ -161,6 +161,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                             }
                         }
 
+                        run.SetRunOnResults();
                         IList<Result> cachedResults = run.Results;
                         run.Results = null;
 
@@ -172,7 +173,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                         {
                             foreach (Result result in run.Results)
                             {
-                                result.InlineArtifacts(run);
+                                result.InlineArtifacts();
                                 string key = _options.SplittingStrategy == SplittingStrategy.PerRule
                                     ? result.RuleId
                                     : string.Empty;
--- a/src/Sarif.Multitool.Library/MergeCommand.cs
+++ b/src/Sarif.Multitool.Library/MergeCommand.cs
@@ -165,12 +165,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
                         run.Results = null;
 
                         Run emptyRun = run.DeepClone();
+                        emptyRun.Artifacts = null;
                         run.Results = cachedResults;
 
                         if (run.Results != null)
                         {
                             foreach (Result result in run.Results)
                             {
+                                result.InlineArtifacts(run);
                                 string key = _options.SplittingStrategy == SplittingStrategy.PerRule
                                     ? result.RuleId
                                     : string.Empty;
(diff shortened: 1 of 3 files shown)


src/Sarif/Core/Result.cs
```diff
@@ -117,6 +123,39 @@ public bool TryIsSuppressed(out bool isSuppressed)
             return true;
         }
 
+        public void InlineArtifacts(Run run = null)
```
michaelcfanning, line 126: >InlineArtifacts [](start = 20, length = 15)

Remember that a result has an optional run object on it. And so if we find that, we should use it. Otherwise, this looks good.

We have another inlining problem, which is that a result needs to have a 'flattened' result message in cases where it otherwise has only messge args and an index to the rules table (which actually holds the message string). We have helpers elsewhere that provide this flattening, there's a visitor dedicated to it. Should make sure useful helper exists on Rule. btw - all of this stuff is helpful to enable optionally in the rewrite command.

src/Sarif/Core/Result.cs
```diff
@@ -17,9 +17,15 @@ public partial class Result
...
-        public bool ShouldSerializeWorkItemUris() { return this.WorkItemUris != null && this.WorkItemUris.Any((s) => s != null); }
+        public bool ShouldSerializeWorkItemUris()
+        {
+            return this.WorkItemUris != null && this.WorkItemUris.Any((s) => s != null);
+        }
 
-        public bool ShouldSerializeLevel() { return this.Level != FailureLevel.Warning; }
+        public bool ShouldSerializeLevel()
+        {
+            return this.Level != FailureLevel.Warning;
```
michaelcfanning, line 27: >FailureLevel.Warning [](start = 33, length = 20)

This is wrong, we should only set this if level doesn't correspond to the default rule level. Also, this value relates to kind! So, if this isn't a failure result, level == 0.

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -191,12 +195,15 @@ private async Task<bool> MergeSarifLogsAsync()
...
                                     splitLog.Runs.Add(splitRun);
                                     _ruleIdToResultsMap[key] = new HashSet<Result>(Result.ValueComparer);
+                                    _ruleIdToMergeVisitorsMap[key] = new RunMergingVisitor();
                                 }
 
                                 if (!_ruleIdToResultsMap[key].Contains(result))
                                 {
                                     _ruleIdToResultsMap[key].Add(result);
-                                    splitRun.Results.Add(result);
+                                    _ruleIdToMergeVisitorsMap[key].CurrentRun = result.Run;
```
michaelcfanning, line 204: >ruleIdToMergeVisitorsMap[key] [](start = 37, length = 29)

use a variable (of RunMergingVisitor type) for this.

src/Test.UnitTests.Sarif.Multitool.Library/TestData/MergeCommand/ExpectedOutputs/DuplicatedResults.sarif
```diff
@@ -1,11 +1,45 @@
 {
-  "version": "2.1.0",
   "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
```
michaelcfanning, line 3: >version [](start = 3, length = 7)

Looks good! Please get some example PREfast files and let's test them explicitly.

michaelcfanning (approved):
:shipit:

PR #2289, opened 2021-02-20
Splitting per fingerprint

eddynaka:


src/Sarif/Writers/RunPartitioner.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Writers
+{
+    /// <summary>
+    /// Methods to break a SARIF log file into pieces.
+    /// </summary>
+    public static class RunPartitioner
```
michaelcfanning, line 9: >RunPartitioner [](start = 24, length = 14)

You don't need this, do you?

src/Sarif/Visitors/PerRunPerFingerprintSplittingVisitor.cs
```diff
@@ -0,0 +1,86 @@
...
+                    var locations = result.Locations.ToList();
+                    foreach (var location in node.Locations)
+                    {
+                        locations.Add(location);
+                    }
+                    result.Locations = locations;
+                }
+            }
+
+            var current = sarifLog.Runs[0].Results.FirstOrDefault(f => f.Fingerprints.First().Value == fingerprint);
```
michaelcfanning, line 73: >FirstOrDefault [](start = 51, length = 14)

ouch! this is inefficient. why are we doing this? can't you just add this result anytime the trygetvalue on line 54 fails? if you can't find the fingerprint then, the result isn't in the current run, so just add it there?

src/Sarif/SplittingStrategy.cs
```diff
@@ -63,5 +63,7 @@ public enum SplittingStrategy
         /// I.e., the total number of log files created is the sum of the unique fingerprint by organization, entity type, and repository.
         /// </summary>
         PerRunPerOrgPerEntityTypePerRepositoryPerPartialFingerprint,
+
+        PerFingerprint,
```
michaelcfanning, line 67: Add a comment:

A grouping strategy that coalesces locations for results with a common fingerprint into a single result. I.e., the total number of results within a run equals the number of unique fingerprints within the run.

michaelcfanning (approved):
:shipit:

src/Sarif/SplittingStrategy.cs
eddynaka: we won't need this. will remove

---
In reply to: [579716158](https://github.com/microsoft/sarif-sdk/pull/2289#discussion_r579716158) [](ancestors = 579716158)

PR #2290, opened 2021-02-22
Improving baseliner

eddynaka:


src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -45,6 +45,14 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
             {
                 // Baseline result and current result have been matched => existing.
                 result = ConstructExistingResult(resultMatchingProperties, out originalResultMatchingProperties);
+
+                var previousResult = new Result { Locations = PreviousResult.Result.Locations };
+                var currenResult = new Result { Locations = CurrentResult.Result.Locations };
+
+                if (previousResult.ValueGetHashCode() != currenResult.ValueGetHashCode())
+                {
+                    result.BaselineState = BaselineState.Updated;
```
jameswinkler, line 54: Are you sure the location is accurate enough to perform this logic?  The sarif files I've worked on generally change every time we make any changes at all to the files we're examining.

eddynaka: this is our first logic to add the updated tag. so, it yet not precise, but it will work for our current case. probably this will not get merged since we found another way...

---
In reply to: [581470416](https://github.com/microsoft/sarif-sdk/pull/2290#discussion_r581470416) [](ancestors = 581470416)

michaelcfanning (approved):
:shipit:

PR #2293, opened 2021-02-23
Adding PerFingerprint strategy in SarifWorkItemFiler

eddynaka:


src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -232,6 +232,11 @@ public virtual IReadOnlyList<SarifLog> SplitLogFile(SarifLog sarifLog)
                             partitionFunction = (result) => result.ShouldBeFiled() ? result.GetPerRepositoryFingerprintSplittingStrategyId() : null;
                             break;
                         }
+                        case SplittingStrategy.PerFingerprint:
+                        {
+                            partitionFunction = (result) => result.ShouldBeFiled() ? result.Fingerprints.Count == 0 ? null : result.Fingerprints.First().Value : null;
```
jameswinkler, line 237: >First() [](start = 145, length = 7)

I'm fixating on "First" here.  In this splitting strategy, if there are multiple results with the same fingerprint, the rest after the first will be discarded, right?  Is that desired?

eddynaka: i didn't want to hard code a specific key...its why i added first. i could also add a parameter being the fingerprint name. 

---
In reply to: [581473453](https://github.com/microsoft/sarif-sdk/pull/2293#discussion_r581473453) [](ancestors = 581473453)

jameswinkler: I have no strong preference so long as we're preserving all the intended information :)

---
In reply to: [581474839](https://github.com/microsoft/sarif-sdk/pull/2293#discussion_r581474839) [](ancestors = 581474839,581473453)

michaelcfanning (approved):
:shipit:

src/Sarif.Multitool.Library/FileWorkItemsCommand.cs
```diff
@@ -138,6 +140,13 @@ private bool ValidateOptions(FileWorkItemsOptions options, SarifWorkItemContext
                 valid &= DriverUtilities.ReportWhetherOutputFileCanBeCreated(options.OutputFilePath, options.Force, fileSystem);
             }
 
+            if ((options.SplittingStrategy == SplittingStrategy.PerFingerprint ||
+                options.SplittingStrategy == SplittingStrategy.PerPropertyBagProperty) &&
+                string.IsNullOrEmpty(options.PropertyName))
+            {
+                valid &= false;
```
michaelcfanning, line 147: >valid &= false [](start = 16, length = 14)

You can do this, but valid = false is also fine. :)

PR #2296, opened 2021-02-24
Enabling kusto command

eddynaka:


src/Sarif.Multitool.Library/KustoCommand.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    Runs = new[]
+                    {
+                    new Run
+                    {
+                        Tool = new Tool
+                        {
+                            Driver = new ToolComponent
+                            {
+                                Name = "SARIF Kusto",
```
michaelcfanning, line 48: >ARIF Kusto [](start = 41, length = 10)

This should be 'spam' only. You can note in a property somewhere that this is spam SARIF rehydrated by the kusto command.

michaelcfanning (approved):
:shipit:

PR #2301, opened 2021-02-28
Improving memory usage

eddynaka:
For a specific run, I observed that we were caching results that weren't need based on the level/kind.

Before:
image

After:
image

During the first 40 seconds, I could see the number of objects increasing:

Before:
image

After:
image

### DIFF
--- a/src/Sarif/Writers/CachingLogger.cs
+++ b/src/Sarif/Writers/CachingLogger.cs
@@ -48,6 +48,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 throw new ArgumentNullException(nameof(result));
             }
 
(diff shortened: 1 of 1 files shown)


michaelcfanning (approved):
:shipit:

src/Sarif/Writers/CachingLogger.cs
```diff
@@ -48,6 +48,11 @@ public void Log(ReportingDescriptor rule, Result result)
                 throw new ArgumentNullException(nameof(result));
             }
 
+            if (!ShouldLog(result))
```
michaelcfanning, line 51: >ShouldLog [](start = 17, length = 9)

Ouch! Fantastic find, well done! This small change will make an enormous difference.

PR #2302, opened 2021-02-28
Disposing context after usage

eddynaka:
With this change, the quantity of GC will reduce, since we are saying: "hey, GC, you can dispose this."
If the number of calls to GC reduces, less time we will wait for GC to finish, so, in theory, the program will be faster!

Before:
image

After:
image

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -27,6 +27,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         // ways depending on whether output is captured to a log file disk or not. In the latter case,
         // the captured output is useful to verify behavior.
         internal bool _captureConsoleOutput;
+
         internal ConsoleLogger _consoleLogger;
 
         private Run _run;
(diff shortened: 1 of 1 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -276,6 +277,7 @@ private async Task<bool> WriteResultsAsync(TContext rootContext)
 
                             RuntimeErrors |= context.RuntimeErrors;
 
+                            context.Dispose();
```
michaelcfanning, line 280: >context.Dispose(); [](start = 28, length = 18)

put this in a finally block! i.e., test for null and dispose if not so. the reason is that your code currently doesn't dispose in the exceptional case. one thing, though! if you do this, remove the context = default in the exception block, as it might prevent disposal! (finally will happen after the catch). alternately, just add the dispose to the catch block. :)

michaelcfanning (approved):
:shipit:

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: I saw this BUT, since the throwExitApplication will generate a new exception and stop the application I think that is not needed.
What is ur opinion?

---
In reply to: [584349903](https://github.com/microsoft/sarif-sdk/pull/2302#discussion_r584349903) [](ancestors = 584349903)

PR #2305, opened 2021-03-02
Fixing json configuration export

eddynaka:


src/Test.UnitTests.Sarif.Multitool.Library/TestData/ValidateCommand/Configuration.json
```diff
@@ -0,0 +1,5 @@
+﻿{
+  "SARIF2002.ProvideMessageArguments.Options": {
+    "RuleEnabled": "error"
```
michaelcfanning, line 3: >error [](start = 20, length = 5)

Shouldn't this be cased as 'Error'? Also, why are we fixing the JSON configuration, how did this come up? Note that we can't actually deserialize our own JSON format!! I can explain why offline, the reason is that we can't apply the type information in the JSON (the original XML design uses element attributes for this. To make everything work, we'd need a new JSON serializer/deserializer that persist the type data as special properties at each level.

eddynaka: With this change both Error or error would work.
With this change we would be able to export the configuration as json and import.

The issue was that we were serializing the enum as value instead of string, which causes the issue.

---
In reply to: [586522294](https://github.com/microsoft/sarif-sdk/pull/2305#discussion_r586522294) [](ancestors = 586522294)

eddynaka: I have a few tests locally that i was using a custom xml. Then i was trying a custom config in json and i saw this error

---
In reply to: [586535491](https://github.com/microsoft/sarif-sdk/pull/2305#discussion_r586535491) [](ancestors = 586535491,586522294)

PR #2316, opened 2021-03-14
ConcurrentDictionary improvement

### LINKED ISSUE #2314: Replace HashSet<int> for simple int
In a multithread analysis, we store the index of seen objects in a HashSet<int>.

Since it's an ordered processing, we could replace the hashSet<int> for an int:

https://github.com/microsoft/sarif-sdk/blob/27ac6394bd6b9e768b35e66e524bae7d24f7a001/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs#L210

This would reduce the allocations when analyzing gvfs repositories.

cc: @michaelcfanning

eddynaka:
Fixes #2314


michaelcfanning (approved):
:shipit:

PR #2323, opened 2021-03-27
ObsoleteOption should not generate RuntimeError

eddynaka:


src/Test.UnitTests.Sarif/WarningsTests.cs
```diff
@@ -39,14 +39,11 @@ public void Warnings_ObsoleteOption()
             testContext.Setup(x => x.Logger).Returns(testLogger.Object);
 
             Warnings.LogObsoleteOption(testContext.Object, "testObsoleteOption");
-
             testLogger.Verify(x => x.LogConfigurationNotification(It.IsAny<Notification>()), Times.Once);
-            Assert.Equal(RuntimeConditions.ObsoleteOption, testContext.Object.RuntimeErrors);
```
michaelcfanning, line 44: >ObsoleteOption [](start = 43, length = 14)

What's this runtime condition fpor if we don't use it in this case? Should it be deleted?

eddynaka: yeah, i removed. Right now, this test only verify the LogConfigurationNotification

---
In reply to: [602733923](https://github.com/microsoft/sarif-sdk/pull/2323#discussion_r602733923) [](ancestors = 602733923)

michaelcfanning (approved):
:shipit:

PR #2327, opened 2021-04-06
Repository check should be case insensitive

eddynaka:
## Motivation
When we analyze a folder using SPAM and we pass a path that is not using the same casing, it won't fill the VersionControlProvenance because it will say that is different. With that, this will fix those cases.


michaelcfanning (approved):
:shipit:

PR #2328, opened 2021-04-06
InsertOptions should accept properties

eddynaka:


src/Sarif/GitHelper.cs
```diff
@@ -63,6 +63,12 @@ public Uri GetRemoteUri(string repoPath)
                     repoPath,
                     args: "remote get-url origin");
 
+            if (!string.IsNullOrEmpty(uriText) && uriText.Contains("@"))
```
michaelcfanning, line 66: >string [](start = 17, length = 6)

Add a comment about what we're doing here?

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -6,11 +6,15 @@
...
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.CodeAnalysis.Sarif.Visitors
 {
     public class InsertOptionalDataVisitor : SarifRewritingVisitor
     {
+        private static readonly Regex versionControlPropertiesRegex =
+            new Regex(@"runs\[(?<run>\d?)\]\.invocations\[(?<invocation>\d?)\].versionControlProvenance.properties.(?<property>[a-zA-Z]+)=(?<value>.*)", RegexOptions.Compiled);
```
michaelcfanning, line 16: >RegexOptions [](start = 153, length = 12)

should this be case-insensitive maybe? or no?

michaelcfanning (approved):
:shipit:

src/Sarif.Driver/Sdk/CommonOptionsBase.cs
```diff
@@ -83,6 +83,12 @@ public class CommonOptionsBase
             HelpText = "A count of threads that should be used for multithreaded operations.")]
         public int Threads { get; set; }
 
+        [Option(
+            "insert-property",
+            Separator = ';',
+            HelpText = "Insert properties will enhance the SARIF file.")]
```
michaelcfanning, line 89: >nsert properties will enhance the SARIF file. [](start = 25, length = 45)

A semicolon-delimited list of JSON path + property values that should be inserted into the output log. Currently, only paths that point to a version control provenance property bag is supported, e.g., 'runs[0].invocations[1].versionControlProvenance.properties.myProperty=myValue'.

michaelcfanning: By the way, Ed, we should consider allow a short-form here, myProperty=myValue should go directly into the SARIF log!!! and that's very easy. honestly, Chris could probably simply use this.

---
In reply to: [608232311](https://github.com/microsoft/sarif-sdk/pull/2328#discussion_r608232311) [](ancestors = 608232311)

eddynaka: i wouldn't add this because if we add this, in the future, we will have to handle both cases...so, i would be restrict for now....

---
In reply to: [608232927](https://github.com/microsoft/sarif-sdk/pull/2328#discussion_r608232927) [](ancestors = 608232927,608232311)

PR #2331, opened 2021-04-16
Exposing ResultLevelKind

eddynaka:


src/Sarif/Core/ResultLevelKind.cs
```diff
@@ -0,0 +1,33 @@
...
+
+                // If level has any value other than "none" and kind is present, then kind SHALL have the value "fail".
+                if (Level != FailureLevel.None)
+                {
+                    return ResultKind.Fail;
+                }
+
+                return resultKind.Value;
+            }
+            set { resultKind = value; }
```
michaelcfanning, line 28: >resultKind = value [](http://example.com/codeflow?start=18&length=18)

If someone sets a kind that isn't 'Fail', you should clear Level. Right?

src/Sarif/Core/ResultLevelKind.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public FailureLevel Level { get; set; }
```
michaelcfanning, line 31: >set [](http://example.com/codeflow?start=41&length=3)

If someone sets this value to non-zero, you should slam 'Fail' into the Kind property.

michaelcfanning (approved):
:shipit:

PR #2337, opened 2021-04-27
Improving sarif from kusto command

eddynaka:


src/Sarif.Multitool.Library/KustoCommand.cs
```diff
@@ -130,15 +130,24 @@ private void InitializeKustoClient()
...
                     }
 
-                    if (etlEntity == "Build" || etlEntity == "Release" || etlEntity == "BuildDefinition" || etlEntity == "ReleaseDefinition")
+                    if (etlEntity == "Build" || etlEntity == "BuildDefinition" ||
+                        etlEntity == "Release" || etlEntity == "ReleaseDefinition")
                     {
                         itemPath = itemPath.Replace("vsrm.visualstudio.com", "visualstudio.com");
                         itemPath = itemPath.Replace("_apis/build/Definitions/", "_build?definitionId=");
                         itemPath = itemPath.Replace("_apis/Release/definitions/", "_release?_a=releases&view=mine&definitionId=");
+                        itemPath = itemPath.Replace("_apis/wit/workItems/", "_workitems/edit/");
```
michaelcfanning, line 139: >Replace [](http://example.com/codeflow?start=44&length=7)

why is this in two places?

why would etl entity not be work item but the path would include work item?

src/Sarif.Multitool.Library/KustoCommand.cs
```diff
@@ -130,15 +130,24 @@ private void InitializeKustoClient()
...
 
                         resultMessageText += $" This pipeline can be updated on [Azure DevOps]({itemPath}) to secure the variable that exposes this secret.";
                     }
 
+                    if (etlEntity == "WorkItem")
+                    {
+                        itemPath = itemPath.Replace("_apis/wit/workItems/", "_workitems/edit/");
+
+                        resultMessageText += $" This work item can be updated on [Azure DevOps]({itemPath}) to secure the variable that exposes this secret.";
```
michaelcfanning, line 148: >resultMessageText [](http://example.com/codeflow?start=24&length=17)

why are you adding this text? i don't think we should. 

also, what does it mean to 'secure the variable' on a work item? i *think* you can only redact a secret from a work item (and that doesn't actually solve the problem, you have to invalidate it).

src/Sarif.Multitool.Library/KustoCommand.cs
eddynaka: forgot to exclude. Changin in a new pr.

src/Sarif.Multitool.Library/KustoCommand.cs
eddynaka: The idea is to add a direct link to the workitem instead of redirecting to the json file.
I removed the part to "secure the variable".

PR #2340, opened 2021-04-30
HDF to Sarif converter

shaopeng-gh:
HDF to Sarif converter


src/Sarif.Converters/HdfFile.cs
```diff
@@ -0,0 +1,576 @@
...
+using System.Globalization;
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    // <auto-generated />
+    // To parse this JSON data use var hdfFile = HdfFile.FromJson(jsonString)
+    public partial class HdfFile
```
eddynaka, line 15: >HdfFile [](http://example.com/codeflow?start=25&length=7)

create a folder called HdfModel, add one file per class

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+        private const string ToolInformationUri = "https://saf.cms.gov/#/normalize";
+
+        public override string ToolName => "Heimdall Tools";
+
+        public override void Convert(Stream input, IResultLogWriter output, OptionallyEmittedData dataToInsert)
+        {
+            input = input ?? throw new ArgumentNullException(nameof(input));
+            output = output ?? throw new ArgumentNullException(nameof(output));
+
+            var textReader = new StreamReader(input);
```
eddynaka, line 26: >StreamReader [](http://example.com/codeflow?start=33&length=12)

using  var textReader

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+                        }
+                    }
+                },
+                Results = rulesAndResults.Item2,
+            };
+
+            PersistResults(output, rulesAndResults.Item2, run);
+        }
+
+        internal new static void PersistResults(IResultLogWriter output, IList<Result> results, Run run)
```
eddynaka, line 65: >internal new static [](http://example.com/codeflow?start=8&length=19)

do we need new static?

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+                            {
+                                Name = "NIST",
+                                Guid = "AAFBAB93-5201-419E-8443-D4925C542398"
+                            }
+                        },
+                        Kinds = new List<string>() { "relevant" },
+                    }))
+            };
+
+            var results = new List<Result>();
```
eddynaka, line 133: >List<Result>() [](http://example.com/codeflow?start=30&length=14)

List<Result>(execJsonControl.Results.Count)

src/Sarif.Converters/HdfConverter.cs
shaopeng-gh: was not with the static keyword, Visual Studio show this in the message section that this can be make static, so I just use that recommended fix

src/Sarif.Converters/HdfConverter.cs
shaopeng-gh: fixed

src/Sarif.Converters/HdfConverter.cs
shaopeng-gh: fixed

src/Sarif.Converters/HdfFile.cs
shaopeng-gh: got it, will work on this.

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+                results.Add(result);
+            }
+            return (reportingDescriptor, results);
+        }
+
+        private static FailureLevel SarifLevelFromHdfImpact(double impact)
+        {
+            // level 4 & 5
+            // Hdf: Critical, High
+            if (impact >= 0.7)
```
eddynaka, line 155: >impact >= 0.7 [](http://example.com/codeflow?start=16&length=13)

check what version of c# we are using. i think we can use the  switch with conditions

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            // level 0, 1
+            // Hdf: Best_Practice, Information
+            else
+            {
+                return FailureLevel.Note;
+            }
+        }
+
+        private static double SarifRankFromHdfImpact(double impact) =>
```
eddynaka, line 179: >private [](http://example.com/codeflow?start=8&length=7)

why/when do u choose to use private/internal?

shaopeng-gh: I copied the structure from existing converter to keep it similar as possible, so I did not gave it a thought. :)

I think generally, if we know this is currently called, or in future will be called out side of class we can make it internal.
we can also minimize scope as private and open as needed.

src/Sarif.Converters/HdfConverter.cs
shaopeng-gh: below version I like need C# 9, your thoughts?

return impact switch
            {
                >= 0.7 => FailureLevel.Error,
                >= 0.3 => FailureLevel.Warning,
                _ => FailureLevel.Note
            };

Error	CS8400	Feature 'relational pattern' is not available in C# 8.0. Please use language version 9.0 or greater.	Sarif.Converters (net472), Sarif.Converters (netstandard2.0)	C:\GH\s-s\src\Sarif.Converters\HdfConverter.cs	163

src/Sarif.Converters/HdfConverter.cs
eddynaka: mark everything as private if needed. if you will need to test something specific and direct, mark internal, so it can be accessed in the test projects

src/Sarif.Converters/HdfConverter.cs
eddynaka: keep the current way :)

src/Sarif.Converters/HdfFile.cs
shaopeng-gh: done.

src/Sarif.Converters/HdfConverter.cs
eddynaka: make private static

shaopeng-gh: fixed to private new static. 
new is there because the base.

eddynaka: i think you don't need this, since this is in the base and its the same thing. so, remove this and test.

shaopeng-gh: got it, let me try.

shaopeng-gh: works great, I have removed!

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.IO;
+using System.Linq;
+
+using Microsoft.CodeAnalysis.Sarif.Converters.HdfModel;
+
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.Converters
+{
+    public class HdfConverter : ToolFileConverterBase
```
eddynaka, line 15: >HdfConverter [](http://example.com/codeflow?start=17&length=12)

can you add Hdf here: [https://github.com/microsoft/sarif-sdk/blob/main/src/Sarif.Multitool.Library/ConvertOptions.cs](https://github.com/microsoft/sarif-sdk/blob/main/src/Sarif.Multitool.Library/ConvertOptions.cs)

and here: [https://github.com/microsoft/sarif-sdk/blob/main/src/ReleaseHistory.md](https://github.com/microsoft/sarif-sdk/blob/main/src/ReleaseHistory.md)

add one new line before 2.4.6 (un-released).

shaopeng-gh: done both, thanks

eddynaka (approved):
:shipit:

PR #2357, opened 2021-05-17
Report inner exception details if available.

michaelcfanning:
@eddynaka 

We currently do not persist exception stack to logs when the exception is passed via an inner exception.


eddynaka (approved):
:shipit:

PR #2358, opened 2021-05-21
Adding support for git blame.

suvamM:
This PR adds support for enriching a SARIF log with `git blame` information. The high level algorithm is as follows:
1. For each result, if the corresponding artifact is a file, retrieve blame information associated with the file.
2. Parse the blame information to extract relevant information (such as author name, email, etc).
3. If the code location of the SARIF result overlaps with the blame hunk, enrich the result node with the extracted blame information.

The code changes have been manually tested, but the PR currently does not introduce unit tests for the feature. The problem lies with mocking appropriate external services, since the relevant code creates local objects to interact with these external services (such as `GitHelper`).

Also, the `OptionallyEmittedData.All` flag now assumes slightly different semantics, where it implies _all_ flags _except_ the overwrite and git blame flags. This was necessary since the existing test for the `OptionallyEmittedData.All` flag does not support the preconditions for git blame.

### DIFF
--- a/src/Sarif/BlameHunk.cs
+++ b/src/Sarif/BlameHunk.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal class BlameHunk : IBlameHunk
+    {
+        private readonly string name;
+        private readonly string email;
+        private readonly string commitSha;
+        private readonly int lineCount;
+        private readonly int finalStartLineNumber;
+
+        public BlameHunk(string name, string email, string commitSha, int lineCount, int finalStartLineNumber)
+        {
+            this.name = name;
+            this.email = email;
+            this.commitSha = commitSha;
+            this.lineCount = lineCount;
+            this.finalStartLineNumber = finalStartLineNumber;
+        }
+
+        public string Name { get => name; }
+        public string Email { get => email; }
+        public string CommitSha { get => commitSha; }
+
+        public int LineCount { get => lineCount; set => throw new System.NotImplementedException(); }
+
+        public int FinalStartLineNumber { get => finalStartLineNumber; }
+
+        public bool ContainsLine(int line)
+        {
+            if (line >= finalStartLineNumber && line <= finalStartLineNumber + lineCount)
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
+        }
+    }
+}
--- a/src/Sarif/BlameHunk.cs
+++ b/src/Sarif/BlameHunk.cs
@@ -24,13 +24,13 @@ namespace Microsoft.CodeAnalysis.Sarif
             this.finalStartLineNumber = finalStartLineNumber;
         }
 
-        public string Name { get => name; }
-        public string Email { get => email; }
-        public string CommitSha { get => commitSha; }
+        public string Name => name;
+        public string Email => email;
+        public string CommitSha => commitSha;
 
-        public int LineCount { get => lineCount; set => throw new System.NotImplementedException(); }
+        public int LineCount => lineCount;
 
-        public int FinalStartLineNumber { get => finalStartLineNumber; }
+        public int FinalStartLineNumber => finalStartLineNumber;
 
         public bool ContainsLine(int line)
         {
--- a/src/Sarif/BlameHunk.cs
+++ b/src/Sarif/BlameHunk.cs
@@ -25,7 +25,9 @@ namespace Microsoft.CodeAnalysis.Sarif
         }
 
         public string Name => name;
+
         public string Email => email;
+
         public string CommitSha => commitSha;
 
         public int LineCount => lineCount;
--- a/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
+++ b/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
@@ -404,8 +404,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             {
                 node.TryReconstructAbsoluteUri(_run.OriginalUriBaseIds, out uri);
             }
-
-            return uri.IsFile ? uri.GetFilePath() : null;
+            return uri.IsAbsoluteUri && uri.IsFile ? uri.GetFilePath() : null;
         }
 
         private const string RepoRootUriBaseIdStem = "REPO_ROOT";
--- a/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
+++ b/src/Sarif/Visitors/InsertOptionalDataVisitor.cs
@@ -221,23 +221,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
             if (_dataToInsert.HasFlag(OptionallyEmittedData.GitBlameInformation))
             {
                 // add git blame information to the result
-                IEnumerable<SarifTransformerUtilities.IBlameHunk> blameHunks = SarifTransformerUtilities.ParseBlameInformation(
-                    _gitHelper.GetBlame(GetRepoPath(node.Locations[0].PhysicalLocation.ArtifactLocation))
-                    ); ;
+                string filePath = GetFilePath(node.Locations[0].PhysicalLocation.ArtifactLocation);
 
-                Region region = node.Locations[0].PhysicalLocation.Region;
-                if (region != null)
+                if (filePath != null)
                 {
-                    foreach (SarifTransformerUtilities.IBlameHunk blameHunk in blameHunks)
+                    IEnumerable<SarifTransformerUtilities.IBlameHunk> blameHunks = SarifTransformerUtilities.ParseBlameInformation(
+                                                                                        _gitHelper.GetBlame(filePath));
+
+                    Region region = node.Locations[0].PhysicalLocation.Region;
+                    if (region != null)
                     {
-                        if (!blameHunk.ContainsLine(region.StartLine))
+                        foreach (SarifTransformerUtilities.IBlameHunk blameHunk in blameHunks)
                         {
-                            continue;
+                            if (!blameHunk.ContainsLine(region.StartLine))
+                            {
+                                continue;
+                            }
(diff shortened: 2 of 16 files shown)


src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -13,6 +15,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 {
     public static class SarifTransformerUtilities
     {
+        private static readonly Regex CommitShaRegex = new Regex(@"^(?<hash>[0-9a-f]{40}).*$");
```
eddynaka, line 18: >); [](http://example.com/codeflow?start=93&length=2)

Add , RegexOptions.Compiled) to all regexes

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -313,5 +323,117 @@ public static AnnotatedCodeLocationImportanceVersionOne CreateAnnotatedCodeLocat
                     return AnnotatedCodeLocationImportanceVersionOne.Important;
             }
         }
+
+        public interface IBlameHunk
```
eddynaka, line 327: >IBlameHunk [](http://example.com/codeflow?start=25&length=10)

move this and the class  below to it's own file.
you can save those inside sarif folder, just like IHttpClient interface

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -313,5 +323,117 @@ public static AnnotatedCodeLocationImportanceVersionOne CreateAnnotatedCodeLocat
...
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+            }
+        }
+
+        public static IEnumerable<IBlameHunk> ParseBlameInformation(string blameText)
```
eddynaka, line 380: >ParseBlameInformation [](http://example.com/codeflow?start=46&length=21)

can we test this?

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -313,5 +323,117 @@ public static AnnotatedCodeLocationImportanceVersionOne CreateAnnotatedCodeLocat
...
+            public BlameHunk(string name, string email, string commitSha, int lineCount, int finalStartLineNumber)
+            {
+                _Name = name;
+                _Email = email;
+                _CommitSha = commitSha;
+                _LineCount = lineCount;
+                _FinalStartLineNumber = finalStartLineNumber;
+            }
+
+            public string Name { get => _Name; set => throw new System.NotImplementedException(); }
```
eddynaka, line 359: >set => throw new System.NotImplementedException(); [](http://example.com/codeflow?start=47&length=50)

nit: remove the set from interface. with that, you would not need to create this exception

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -313,5 +323,117 @@ public static AnnotatedCodeLocationImportanceVersionOne CreateAnnotatedCodeLocat
...
+            int LineCount { get; }
+
+            int FinalStartLineNumber { get; }
+
+            bool ContainsLine(int line);
+        }
+
+        private class BlameHunk : IBlameHunk
+        {
+            private readonly string _Name;
```
eddynaka, line 344: >_Name [](http://example.com/codeflow?start=36&length=5)

can you use lower-case? _name, _email, etc.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -361,6 +393,20 @@ private string GetNextRepoRootUriBaseId()
             }
         }
 
+        private string GetFilePath(ArtifactLocation node)
+        {
+            Uri uri = node.Uri;
```
eddynaka, line 398: >Uri uri = node.Uri [](http://example.com/codeflow?start=12&length=18)

check ArtifactLocation.Resolve, that might simplify a little bit

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Done. This is a good suggestion, thanks!

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: OK, let me add a unit test to evaluate just the parser.

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Fixed.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -214,6 +218,34 @@ public override Result VisitResult(Result node)
...
+                    Region region = node.Locations[0].PhysicalLocation.Region;
+                    if (region != null)
+                    {
+                        foreach (SarifTransformerUtilities.IBlameHunk blameHunk in blameHunks)
+                        {
+                            if (!blameHunk.ContainsLine(region.StartLine))
+                            {
+                                continue;
+                            }
+                            node.SetProperty(nameof(CommitSha), blameHunk.CommitSha);
```
eddynaka, line 240: >nameof(CommitSha) [](http://example.com/codeflow?start=45&length=17)

replace by CommitSha const (just like the others below)

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Fixed.

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Fixed.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
suvamM: Fixed.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
suvamM: Fixed.

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Added unit test for the git blame text parser. The code for checking region overlap with the results is trivial, so skipping tests for that.

suvamM:
@eddynaka Made a commit incorporating your suggestions. The unit test for the git blame parser uncovered a couple of bugs! Things are fixed now. Look forward to your feedback.

src/Sarif/BlameHunk.cs
```diff
@@ -0,0 +1,47 @@
...
+            this.commitSha = commitSha;
+            this.lineCount = lineCount;
+            this.finalStartLineNumber = finalStartLineNumber;
+        }
+
+        public string Name { get => name; }
+        public string Email { get => email; }
+        public string CommitSha { get => commitSha; }
+
+        public int LineCount { get => lineCount; set => throw new System.NotImplementedException(); }
```
eddynaka, line 31: >set => throw new System.NotImplementedException(); [](http://example.com/codeflow?start=49&length=50)

i think u dont need this

src/Sarif/BlameHunk.cs
```diff
@@ -0,0 +1,47 @@
...
+        public BlameHunk(string name, string email, string commitSha, int lineCount, int finalStartLineNumber)
+        {
+            this.name = name;
+            this.email = email;
+            this.commitSha = commitSha;
+            this.lineCount = lineCount;
+            this.finalStartLineNumber = finalStartLineNumber;
+        }
+
+        public string Name { get => name; }
```
eddynaka, line 27: add a new line between each property just like linecount and others below

src/Sarif/BlameHunk.cs
suvamM: Ah sorry missed this.

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -13,6 +15,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Visitors
 {
     public static class SarifTransformerUtilities
     {
+        private static readonly Regex CommitShaRegex = new Regex(@"^(?<hash>[0-9a-f]{40}).*$", RegexOptions.Compiled);
+        private static readonly int CommitShaLength = 40;
+        private static readonly Regex AuthorTZRegex = new Regex(@"^author-tz", RegexOptions.Compiled);
+        private static readonly Regex AuthorTimeRegex = new Regex(@"^author-time", RegexOptions.Compiled);
+        private static readonly Regex AuthorMailRegex = new Regex(@"^author-mail", RegexOptions.Compiled);
+        private static readonly string authorMailString = "author-mail <>";
+        private static readonly Regex AuthorRegex = new Regex(@"^author", RegexOptions.Compiled);
```
eddynaka, line 24: Can you use the snippet above:
```csharp

        private static readonly int CommitShaLength = 40;
        private static readonly string authorMailString = "author-mail <>";
        private static readonly Regex AuthorRegex = new Regex(@"^author", RegexOptions.Compiled);
        private static readonly Regex AuthorTZRegex = new Regex(@"^author-tz", RegexOptions.Compiled);
        private static readonly Regex AuthorTimeRegex = new Regex(@"^author-time", RegexOptions.Compiled);
        private static readonly Regex AuthorMailRegex = new Regex(@"^author-mail", RegexOptions.Compiled);
        private static readonly Regex CommitShaRegex = new Regex(@"^(?<hash>[0-9a-f]{40}).*$", RegexOptions.Compiled);

```

it is ordered by textstring, so, easier to ready

src/Sarif/BlameHunk.cs
suvamM: Fixed.

src/Sarif/BlameHunk.cs
suvamM: Fixed.

src/Sarif/Visitors/SarifTransformerUtilities.cs
suvamM: Makes sense. Fixed.

eddynaka (approved):

PR #2360, opened 2021-05-30
Generate rebaseline script instead of performing the work at test time.

michaelcfanning:
@eddynaka 

Previously, to rebaseline you were required to modify a special bool, rerun tests, then reset that bool. Now we just generate a script that updates everything for you (once you've visually reviewed/approved any deltas). Saves a lot of time.


eddynaka (approved):
:shipit:

PR #2373, opened 2021-07-16
ClangTidy read line and row number from log file if exists

shaopeng-gh:


src/Sarif.Converters/ClangTidyConverter.cs
```diff
@@ -15,6 +16,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
     public class ClangTidyConverter : ToolFileConverterBase
```
eddynaka, line 16: >ClangTidyConverter [](http://example.com/codeflow?start=17&length=18)

Let's also update the md files, showing that we can read the file.yaml.log to understand the line number

eddynaka (approved):

PR #2375, opened 2021-07-30
Improving baseline command

eddynaka:


src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -67,22 +67,25 @@ public IEnumerable<SarifLog> Match(IEnumerable<SarifLog> previousLogs, IEnumerab
...
-                IEnumerable<Run> currentRuns = new Run[0];
 
+                IEnumerable<Run> currentRuns = new Run[0];
                 if (runsByToolCurrent.ContainsKey(key))
                 {
                     currentRuns = runsByToolCurrent[key];
                 }
 
-                resultToolLogs.Add(BaselineSarifLogs(baselineRuns, currentRuns));
+                if (baselineRuns.Any() && currentRuns.Any())
```
eddynaka, line 77: >baselineRuns.Any() && currentRuns.Any() [](http://example.com/codeflow?start=20&length=39)

If one of those values is empty, we will receive an exception.

To prevent that, let's skip tool names that does not match.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -94,9 +97,10 @@ public IEnumerable<SarifLog> Match(IEnumerable<SarifLog> previousLogs, IEnumerab
                 {
                     continue;
                 }
+
                 foreach (Run run in sarifLog.Runs)
                 {
-                    string toolName = run.Tool.Driver.Name;
+                    string toolName = run.Tool.Driver.Name.ToLower();
```
eddynaka, line 103: >ToLower() [](http://example.com/codeflow?start=59&length=9)

this will make all tools lower case

michaelcfanning: Don't do this! :) It is never a good idea to transform data when you are an API. Instead, please look for ways to make your comparisons case-insensitive when necessary. For example, create a hash set with a case-insensitive comparer (for the set members).

By the way, in the rare cases when ToLower() is required, you should always prefer ToLowerInvariant() in order to avoid culture-sensitive issues (for letters that don't round-trip properly). Search the web for 'turkish i problem'.

michaelcfanning (changes requested):
:clock1:

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -67,22 +67,25 @@ public IEnumerable<SarifLog> Match(IEnumerable<SarifLog> previousLogs, IEnumerab
...
+                }
             }
 
             return new List<SarifLog> { resultToolLogs.Merge(mergeEmptyLogs: true) };
         }
 
         private static Dictionary<string, List<Run>> GetRunsByTool(IEnumerable<SarifLog> sarifLogs)
         {
-            Dictionary<string, List<Run>> runsByTool = new Dictionary<string, List<Run>>();
+            var runsByTool = new Dictionary<string, List<Run>>();
```
michaelcfanning, line 88: >Dictionary [](http://example.com/codeflow?start=33&length=10)

Call the IEqualityComparer ctor here and pass StringComparer.OrdinalIgnoreCase to get case-insensitive key comparisons using .ContainsKey

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
eddynaka: got it.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
eddynaka: done

src/Sarif.Multitool.Library/Rules/SARIF2006.UrisShouldBeReachable.cs
```diff
@@ -36,7 +37,7 @@ public UrisShouldBeReachable()
         };
 
         // cache stores all visited Uris
-        private static readonly Dictionary<string, bool> s_checkedUris = new Dictionary<string, bool>();
+        private static readonly ConcurrentDictionary<string, bool> s_checkedUris = new ConcurrentDictionary<string, bool>();
```
eddynaka, line 40: >ConcurrentDictionary [](http://example.com/codeflow?start=32&length=20)

when we run multiple tests, we are facing concurrency issues. To prevent that, i'm changing to ConcurrentDictionary

michaelcfanning (approved):
:shipit:

PR #2381, opened 2021-08-18
Fix CodeQL code scanning issues

yongyan-gh:
- Dereferenced variable is always null
- Equals on incomparable types
- Container contents are never accessed
- Unsynchronized access to static collection member in non-static context

### DIFF
--- a/src/Sarif.Multitool.Library/Rules/GH1004.ReviewArraysThatExceedConfigurableDefaults.cs
+++ b/src/Sarif.Multitool.Library/Rules/GH1004.ReviewArraysThatExceedConfigurableDefaults.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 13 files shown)


src/Sarif.Multitool.Library/Rules/GH1004.ReviewArraysThatExceedConfigurableDefaults.cs
```diff
@@ -53,7 +54,7 @@ public ReviewArraysThatExceedConfigurableDefaults()
         internal static readonly string s_codeFlowsPerResultKey = $"{s_resultsPerRunKey}/0/{SarifPropertyName.CodeFlows}";
         internal static readonly string s_locationsPerThreadFlowKey = $"{s_codeFlowsPerResultKey}/0/{SarifPropertyName.ThreadFlows}/0/{SarifPropertyName.Locations}";
 
-        internal static Dictionary<string, int> s_arraySizeLimitDictionary = new Dictionary<string, int>
+        internal static ConcurrentDictionary<string, int> s_arraySizeLimitDictionary = new ConcurrentDictionary<string, int>
```
eddynaka, line 57: you can follow codeql and add readonly since no one will change it.

eddynaka (approved):

PR #2382, opened 2021-09-12
Reusing results from MultithreadedAnalyzeCommandBase when hash is enabled

eddynaka:
## Description

When someone uses `MultithreadedAnalyzeCommandBase`, the analysis will execute in parallel.
If we have the same file copied over and over it would process it multiple times since we don't have any caching mechanism that could be enabled.

Comparing with `AnalyzeCommandBase`, if we have the `--insert Hashes` argument, it would cache the result and re-use the results.
https://github.com/microsoft/sarif-sdk/blob/28b7706a4d684a07ff38dd79bdf52469d24526c2/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs#L650-L684

## Proposal

The idea is to re-use the results from a previous analysis if the hash of the file is the same.
This would mimic `AnalyzeCommandBase` caching mechanism if `--insert Hashes` is enabled. Otherwise, the cache would **not** be constructed/used.

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -40,6 +40,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         private Channel<int> _resultsWritingChannel;
         private Channel<int> _fileEnumerationChannel;
         private IDictionary<string, HashData> _pathToHashDataMap;
+        private ConcurrentDictionary<string, IAnalysisLogger> _analysisLoggerCache;
 
         public Exception ExecutionException { get; set; }
 
@@ -245,46 +246,22 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                         while (context?.AnalysisComplete == true)
                         {
-                            var cachingLogger = (CachingLogger)context.Logger;
-                            IDictionary<ReportingDescriptor, IList<Result>> results = cachingLogger.Results;
-
-                            if (results?.Count > 0)
+                            if (_computeHashes)
                             {
-                                if (context.Hashes != null)
+                                if (_analysisLoggerCache.TryGetValue(context.Hashes.Sha256, out IAnalysisLogger logger))
                                 {
-                                    Debug.Assert(_run != null);
-                                    int index = _run.GetFileIndex(new ArtifactLocation() { Uri = context.TargetUri });
-
-                                    _run.Artifacts[index].Hashes = new Dictionary<string, string>
-                                    {
-                                        { "sha-256", context.Hashes.Sha256 },
-                                    };
+                                    LogCachingLogger(rootContext, logger, context, clone: true);
                                 }
-
-                                foreach (KeyValuePair<ReportingDescriptor, IList<Result>> kv in results)
+                                else
                                 {
-                                    foreach (Result result in kv.Value)
-                                    {
-                                        rootContext.Logger.Log(kv.Key, result);
-                                    }
+                                    Debug.Assert(logger != null, "Logger should always exist in cache.");
                                 }
                             }
-
-                            if (cachingLogger.ToolNotifications != null)
+                            else
                             {
-                                foreach (Notification notification in cachingLogger.ToolNotifications)
-                                {
-                                    rootContext.Logger.LogToolNotification(notification);
-                                }
+                                LogCachingLogger(rootContext, context.Logger, context);
                             }
 
-                            if (cachingLogger.ConfigurationNotifications != null)
-                            {
-                                foreach (Notification notification in cachingLogger.ConfigurationNotifications)
-                                {
-                                    rootContext.Logger.LogConfigurationNotification(notification);
-                                }
-                            }
 
                             RuntimeErrors |= context.RuntimeErrors;
 
@@ -298,17 +275,68 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                     catch (Exception e)
                     {
                         context?.Dispose();
-                        context = default;
                         RuntimeErrors |= Errors.LogUnhandledEngineException(rootContext, e);
                         ThrowExitApplicationException(context, ExitReason.ExceptionWritingToLogFile, e);
                     }
                 }
             }
 
-            Debug.Assert(_fileContexts.Count == 0);
+            Debug.Assert(_fileContexts.IsEmpty);
             Debug.Assert(_fileContextsCount == currentIndex);
 
             return true;
+        }
+
+        private void LogCachingLogger(TContext rootContext, IAnalysisLogger logger, TContext context, bool clone = false)
+        {
+            var cachingLogger = (CachingLogger)logger;
+            IDictionary<ReportingDescriptor, IList<Result>> results = cachingLogger.Results;
+
+            if (results?.Count > 0)
+            {
+                foreach (KeyValuePair<ReportingDescriptor, IList<Result>> kv in results)
+                {
+                    foreach (Result result in kv.Value)
+                    {
+                        Result currentResult = result;
+                        if (clone)
+                        {
+                            Result clonedResult = result.DeepClone();
+
+                            UpdateLocationsAndMessageWithCurrentUri(clonedResult.Locations, clonedResult.Message, context.TargetUri);
+
+                            currentResult = clonedResult;
+                        }
+
+                        rootContext.Logger.Log(kv.Key, currentResult);
+                    }
+                }
+            }
+
+            if (cachingLogger.ToolNotifications != null)
+            {
+                foreach (Notification notification in cachingLogger.ToolNotifications)
+                {
+                    rootContext.Logger.LogToolNotification(notification);
+                }
+            }
+
+            if (cachingLogger.ConfigurationNotifications != null)
+            {
+                foreach (Notification notification in cachingLogger.ConfigurationNotifications)
+                {
+                    Notification currentNotification = notification;
+                    if (clone)
+                    {
+                        Notification clonedNotification = notification.DeepClone();
+                        UpdateLocationsAndMessageWithCurrentUri(clonedNotification.Locations, notification.Message, context.TargetUri);
+
+                        currentNotification = clonedNotification;
+                    }
+
+                    rootContext.Logger.LogConfigurationNotification(currentNotification);
+                }
+            }
         }
 
         private async Task<bool> FindFilesAsync(TOptions options, TContext rootContext)
@@ -432,15 +460,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                         context = _fileContexts[item];
 
                         DetermineApplicabilityAndAnalyze(context, skimmers, disabledSkimmers);
-
-                        if (_computeHashes && ((CachingLogger)context.Logger).Results?.Count > 0)
-                        {
-                            Debug.Assert(context.Hashes == null);
-
-                            string sha256Hash = HashUtilities.ComputeSha256Hash(context.TargetUri.LocalPath);
-
-                            context.Hashes = new HashData(md5: null, sha1: null, sha256: sha256Hash);
-                        }
                     }
                     catch (Exception e)
                     {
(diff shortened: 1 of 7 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -859,7 +896,20 @@ private SupportedPlatform GetCurrentRunningOS()
                 return context;
             }
 
+            // If computeHashes exists and we already added in the cache,
+            // we can just return the context.
+            if (_computeHashes && _analysisLoggerCache.ContainsKey(context.Hashes.Sha256))
```
eddynaka, line 901: >_computeHashes [](http://example.com/codeflow?start=16&length=14)

if we have hashes enabled and the hash already exists, we can skip the analysis

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -859,7 +896,20 @@ private SupportedPlatform GetCurrentRunningOS()
...
+            {
+                return context;
+            }
+
             context.Logger.AnalyzingTarget(context);
+
+            if (_computeHashes)
```
eddynaka, line 908: >_computeHashes [](http://example.com/codeflow?start=16&length=14)

if we have hashes enabled, we will save the logger in the cache which will be re-used in the future

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -587,6 +611,19 @@ internal AggregatingLogger InitializeLogger(AnalyzeOptionsBase analyzeOptions)
             if (filePath != null)
             {
                 context.TargetUri = new Uri(filePath);
+
+                if (_computeHashes)
```
eddynaka, line 615: >_computeHashes [](http://example.com/codeflow?start=20&length=14)

computing once per file and re-using if that already exists. We must update _pathToHashDataMap because it will be used to create the artifacts

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -447,12 +447,16 @@ private void CaptureArtifact(ArtifactLocation fileLocation)
                 catch (ArgumentException) { } // Unrecognized encoding name
             }
 
+            HashData hashData = null;
+            AnalysisTargetToHashDataMap?.TryGetValue(fileLocation.Uri.LocalPath, out hashData);
```
eddynaka, line 451: >AnalysisTargetToHashDataMap?.TryGetValue [](http://example.com/codeflow?start=12&length=40)

if we have analysisTarget, we will check if we have the hashdata. If yes, we will use it

michaelcfanning (changes requested):
:clock1:

michaelcfanning:
Did you neglect to include tests? Do you have any proof points that this change is effective, doesn't raise concurrency issues, preserves run-over-run determinism in results ordering, etc? What stress testing did you use to validate this change?

---
In reply to: [753050501](https://github.com/microsoft/sarif-sdk/pull/2382#issuecomment-753050501) [](http://example.com/codeflow?ancestors=753050501,753050501)

eddynaka:
> Did you neglect to include tests? Do you have any proof points that this change is effective, doesn't raise concurrency issues, preserves run-over-run determinism in results ordering, etc? What stress testing did you use to validate this change?
> 
> In reply to: [753050501](https://github.com/microsoft/sarif-sdk/pull/2382#issuecomment-753050501)

CodeFlow is freezing here, so, using GH for now.
I didn't add the tests because I was just creating a PoC to see if that would be a good approach.
Either way, I just pushed a few more files with one test.

From that, I discovered an inconsistency between AnalyzeCommandBase and MultithreadedCommandBase.

The AnalyzeCommandBase always emit artifacts if we are saving to a file and this does not happen for the MultithreadedCommandBase.

In `SarifLogger`, there is a flag called `_persistArtifacts` that is used for the multi-threaded flow but isn't used for the single-threaded.

Approaches:
1. We could always emit the artifacts making the multi-threaded like single-threaded
2. we could remove the artifacts from the single threaded when you don't supply the correct flags, but that is a breaking change

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1000,6 +1000,12 @@ public void AnalyzeCommandBase_CachesNotificationsWhenPersistingToLogFile()
 
             testCase.Verbose = true;
             RunResultsCachingTestCase(testCase);
+
+            testCase.Files = ComprehensiveKindAndLevelsByFilePath;
+            RunResultsCachingTestCase(testCase, multithreaded: true);
+
+            testCase.Verbose = true;
```
michaelcfanning, line 1007: >testCase.Verbose = true; [](http://example.com/codeflow?start=12&length=24)

Wasn't verbose already set to true? Is this doing anything?

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1027,9 +1033,13 @@ public void AnalyzeCommandBase_CachesResultsWhenPersistingToLogFile()
             };
 
             RunResultsCachingTestCase(testCase);
-
             testCase.Verbose = true;
             RunResultsCachingTestCase(testCase);
+
+            testCase.Files = ComprehensiveKindAndLevelsByFilePath;
+            RunResultsCachingTestCase(testCase, multithreaded: true);
+            testCase.Verbose = true;
```
michaelcfanning, line 1041: >testCase [](http://example.com/codeflow?start=12&length=8)

in your test code above you introduced vertical whitespace but you didn't do that here.

src/Sarif/Writers/SarifLogger.cs
michaelcfanning, line 451: >LocalPath [](http://example.com/codeflow?start=70&length=9)

Is this data initialized for every kind of URI? have you tried testing relative URIs? Isn't it true that local path is only available to absolute URIs?

michaelcfanning, line 451: >hashData [](http://example.com/codeflow?start=85&length=8)

You should be able to just provide an underscore here, to indicate retrieval of an unused thing. Then you don't need the unused local.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -447,12 +447,16 @@ private void CaptureArtifact(ArtifactLocation fileLocation)
...
+
             // Ensure Artifact is in Run.Artifacts and ArtifactLocation.Index is set to point to it
             int index = _run.GetFileIndex(
                 fileLocation,
                 addToFilesTableIfNotPresent: _persistArtifacts,
                 _dataToInsert,
-                encoding);
+                encoding,
+                hashData: hashData);
```
michaelcfanning, line 459: >hashData: hashData [](http://example.com/codeflow?start=16&length=18)

it's not valuable to repeat a local name that matches the name of the parameter, no additional information is provided, we've just consumed more space.

src/Sarif/Writers/SarifLogger.cs
michaelcfanning: Your comment describes the literal code change but you haven't described why the code change has been made.

src/Sarif/Writers/SarifLogger.cs
eddynaka: fixed here and in all the other places in this file.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
eddynaka: i organized everywhere to give a space between each test.

src/Sarif/Writers/SarifLogger.cs
eddynaka: since AnalysisTargetToHashDataMap can be null, I must initialize HashData variable  which will be used in the next call.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
eddynaka: fixed the logic.
I have to set the new files list + verbose to false. and, on line 1008, it will update verbose to true.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -245,46 +246,22 @@ private async Task<bool> WriteResultsAsync(TContext rootContext)
 
                         while (context?.AnalysisComplete == true)
                         {
-                            var cachingLogger = (CachingLogger)context.Logger;
-                            IDictionary<ReportingDescriptor, IList<Result>> results = cachingLogger.Results;
-
-                            if (results?.Count > 0)
+                            if (_computeHashes)
```
michaelcfanning, line 249: >_computeHashes [](http://example.com/codeflow?start=32&length=14)

are you merging the concepts of computing hashes and caching log results? is this what we want to do?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning, line 901: >_analysisLoggerCache [](http://example.com/codeflow?start=34&length=20)

Why is it correct here not to call into the logger? How are the cached results returned in this case?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: Before my change, the results were not cached. each file had its own logger.

With my change, if hash is enabled, we would re-use the logger from the first hash. Otherwise, everything is the same.

This is mimicking the single thread way but using cachingLogger

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -587,6 +613,18 @@ internal AggregatingLogger InitializeLogger(AnalyzeOptionsBase analyzeOptions)
             if (filePath != null)
             {
                 context.TargetUri = new Uri(filePath);
+
+                if (_computeHashes)
+                {
+                    context.Hashes = HashUtilities.ComputeHashes(filePath);
+                    if (_pathToHashDataMap != null)
```
eddynaka, line 620: >_pathToHashDataMap [](http://example.com/codeflow?start=24&length=18)

`_pathToHashDataMap` and `_run` are only initialized when we are outputting to a file/stream using `SarifLogger`. This check will prevent generating a `NullRefException` when we are not outputting as a file/stream

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: when this happen, we will return the context and mark as completed.

From that, the thread that is writing will look in the cache for the an AnalysisLogger with that hash.

With that, it would clone the results and log it.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -298,19 +276,70 @@ private async Task<bool> WriteResultsAsync(TContext rootContext)
...
             }
 
-            Debug.Assert(_fileContexts.Count == 0);
+            Debug.Assert(_fileContexts.IsEmpty);
             Debug.Assert(_fileContextsCount == currentIndex);
 
             return true;
         }
 
+        private void LogCachingLogger(TContext rootContext, IAnalysisLogger logger, TContext context, bool clone = false)
```
eddynaka, line 291: >LogCachingLogger [](http://example.com/codeflow?start=21&length=16)

This is the method that will either clone a previous cached result or just save it.

The clone will be used when Hashes is enabled and the logger will be the one from the cache.

This will re-use previous results that we save in the cachingLogger.

PR #2383, opened 2021-09-13
Always write out level

### LINKED ISSUE #2024: Impossible to tell whether result.level was explicitly specified
The deserialization hints on the autogenerated `Result.Level` property:

```
        [DefaultValue(FailureLevel.Warning)]
```

... mean that an SDK client can't tell whether `result.level` was present in the JSON with value `"warning"`, or whether it was absent and the deserializer assigned the default value.

This matters because the specification's algorithm for calculating `result.level` (see [§3.27.10](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317648)) depends on whether or not the property is present in the JSON.

NOTE: The algorithm actually has two bugs; see oasis-tcs/sarif-spec#470. We should implement to the corrected algorithm.

@michaelcfanning FYI

gfs:
Fix #2024

### DIFF
--- a/src/Sarif/NotYetAutoGenerated/ReportingConfiguration.cs
+++ b/src/Sarif/NotYetAutoGenerated/ReportingConfiguration.cs
@@ -49,7 +49,6 @@ namespace Microsoft.CodeAnalysis.Sarif
         /// </summary>
         [DataMember(Name = "level", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(FailureLevel.Warning)]
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
         public FailureLevel Level { get; set; }
 


michaelcfanning:
OK. It is definitely time to fix this issue. :) As you've noted, there are a few issues filed on this topic, including bugs against the SARIF spec itself. Give me a day to go review all these problems comprehensively, so I can better understand how your fix fits in. It's possible I will suggest an alternative fix, as part of attempting to resolve all known issues. If that happens, though, I'll come back with a new suggestion within 48 hours.

src/Sarif/NotYetAutoGenerated/ReportingConfiguration.cs
```diff
@@ -47,9 +47,9 @@ public SarifNodeKind SarifNodeKind
         /// <summary>
         /// Specifies the failure level for the report.
         /// </summary>
-        [DataMember(Name = "level", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "level", IsRequired = false, EmitDefaultValue = true)]
         [DefaultValue(FailureLevel.Warning)]
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
```
shaopeng-gh, line 52: >[JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)] [](http://example.com/codeflow?start=8&length=68)

(Hello! I am newly assigned to take a look at the level default value issue. I might miss some context)

just for my unsersatnding for this line only, if we are going this route this should be set to DefaultValueHandling.Include not DefaultValueHandling.Populate right, 
which eaquals to remove this line 
since it is the default of DefaultValueHandling attribute.

gfs: There are a few issues at play but this proposed PR only resolved one of them. 

1. The default value is warning. This isn't compliant with the spec. I didn't fix that here. 
2. When serializing, the default value isn't written.

I chose populate instead of include in case there were users of the sarif ask that had serialized out some files where the levels got skipped - when reading them back in would repopulate them back to warning. The downside of course being 1. That for any file not from the C# sarif sdk a blank level does not mean warning. However, changing this behavior is sort of a breaking change, so I wasn't sure if we wanted to do it at this point. 

I think a more holistic solution is as you propose to use include instead of populate and fix the default value behavior to comply with the spec.

shaopeng-gh: thanks, I created a PR, just for discussion, can not find you in the review user list to add you.
[https://github.com/microsoft/sarif-sdk/pull/2399](https://github.com/microsoft/sarif-sdk/pull/2399)

derigel23:
Folks, the result it's still missing the level property when it's a warning.

PR #2384, opened 2021-09-16
UriBaseId based on another UriBaseId should be loaded

shaopeng-gh:


src/Sarif.Multitool.Library/Rules/SARIF2007.ExpressPathsRelativeToRepoRoot.cs
```diff
@@ -58,7 +58,10 @@ protected override void Analyze(Run run, string runPointer)
                     }
                     else
                     {
-                        this.uriBaseIds.Add(run.VersionControlProvenance[i].MappedTo.UriBaseId);
+                        string validUriBaseId = run.VersionControlProvenance[i].MappedTo.UriBaseId;
+                        this.uriBaseIds.Add(validUriBaseId);
+                        run.OriginalUriBaseIds?.Where(baseId => baseId.Value?.UriBaseId == validUriBaseId).ToList()
+                            .ForEach(baseId => this.uriBaseIds.Add(baseId.Key));
```
eddynaka, line 64: I think you could replace this for:
```
if (run.OriginalUriBaseIds != null)
{
    foreach (var baseId in run.OriginalUriBaseIds(..)
    {
         this.uribaseids.add(baseid.key)
    }
}
```

with that, you won't need to call toList and generate everything at first.

src/Sarif.Multitool.Library/Rules/SARIF2007.ExpressPathsRelativeToRepoRoot.cs
```diff
@@ -58,7 +58,10 @@ protected override void Analyze(Run run, string runPointer)
                     }
                     else
                     {
-                        this.uriBaseIds.Add(run.VersionControlProvenance[i].MappedTo.UriBaseId);
+                        string validUriBaseId = run.VersionControlProvenance[i].MappedTo.UriBaseId;
+                        this.uriBaseIds.Add(validUriBaseId);
+                        run.OriginalUriBaseIds?.Where(baseId => baseId.Value?.UriBaseId == validUriBaseId).ToList()
```
eddynaka, line 63: >run.OriginalUriBaseIds [](http://example.com/codeflow?start=24&length=22)

we also have a pattern for the tests.
so, pls take a look at the endtoend tests where you can create a sarif with this and test this

michaelcfanning, line 63: >Where [](http://example.com/codeflow?start=48&length=5)

In general, please don't implement linq queries of this kind in our code base. Instead author patterns that iterate over the original collection a single time.

eddynaka (approved):
:shipit:

PR #2388, opened 2021-09-24
MultithreadedAnalysis with file hashing and caching

eddynaka:


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -384,11 +411,12 @@ private async Task<bool> FindFilesAsync(TOptions options, TContext rootContext)
                                           filePath: file)
                         );
 
-                        await _fileEnumerationChannel.Writer.WriteAsync(_fileContextsCount++);
+                        await _hashChannel.Writer.WriteAsync(_fileContextsCount++);
```
eddynaka, line 414: >_hashChannel [](http://example.com/codeflow?start=30&length=12)

adding to hashChannel so it can generate the hash if computeHash is enabled.

src/Test.UnitTests.Sarif.Driver/rewrite.coyote.release.json
```diff
@@ -0,0 +1,4 @@
+{
+  "AssembliesPath":"..\\..\\bld\\bin\\AnyCPU_Debug\\Test.UnitTests.Sarif.Driver\\netcoreapp3.1",
```
suvamM: @eddynaka This path is incorrect for releases.

src/Sarif.Sdk.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29709.97
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31728.443
```
shaopeng-gh, line 4: Question, this is version 2022 right, do we want to upgrade, should we revert this change?

eddynaka: it does not matter if u are using VS2019 or VS2022.

The nuget change is just to show when we open the sln.

PR #2390, opened 2021-10-04
Flowing suppression to baseline

eddynaka:


src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -49,15 +49,14 @@ protected virtual void ProcessBaseline(IAnalysisContext context, T driverOptions
...
-                                    Formatting.Indented :
-                                    Formatting.None
-            };
+            SarifLog baselineFile = PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(fileSystem.FileReadAllText(options.BaselineSarifFile),
+                                                                                              options.Formatting,
+                                                                                              out string _);
+
+            SarifLog currentSarifLog = PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(fileSystem.FileReadAllText(options.OutputFilePath),
+                                                                                                 options.Formatting,
+                                                                                                 out string _);
```
eddynaka, line 58: this will enable us to accept any version.

src/Sarif.Driver/Sdk/DriverCommand.cs
```diff
@@ -3,11 +3,8 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
-    public abstract class DriverCommand<T>
+    public abstract class DriverCommand<T> : CommandBase
```
eddynaka, line 6: >CommandBase [](http://example.com/codeflow?start=45&length=11)

with this, we can re-use some methods in the CommandBasae class

src/Sarif/Baseline/ResultMatching/DataStructures/MatchedResults.cs
```diff
@@ -153,6 +153,11 @@ public Result CalculateBasedlinedResult(DictionaryMergeBehavior propertyBagMerge
                 resultMatchingProperties.Add(MatchedResults.MatchResultMetadata_PreviousGuid, PreviousResult.Result.Guid);
             }
 
+            if (PreviousResult.Result.Suppressions != null && CurrentResult.Result.Suppressions == null)
+            {
+                result.Suppressions = PreviousResult.Result.Suppressions;
+            }
```
eddynaka, line 159: this will flow suppressions from the baseline file to the current file.

yongyan-gh: so here we don't consider current file's suppression?

eddynaka: Since we start from the current, we just have to compare previous with current. if previous exist and current does not, we should replace.
otherwise, we will keep the current :)

yongyan-gh (approved):
:shipit:

PR #2394, opened 2021-10-11
Enable suppress command in the multitool

eddynaka:


src/Sarif.Multitool.Library/SuppressCommand.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private bool ValidateOptions(SuppressOptions options)
+        {
+            bool valid = true;
+
+            valid &= options.Validate();
+            valid &= options.ExpiryInDays >= 0;
+            valid &= !string.IsNullOrWhiteSpace(options.Justification);
+            valid &= (options.Status == SuppressionStatus.Accepted || options.Status == SuppressionStatus.UnderReview);
+            valid &= DriverUtilities.ReportWhetherOutputFileCanBeCreated(options.OutputFilePath, options.Force, FileSystem);
```
Bpendragon, line 77: Even if every other item is false, this will return true if the last line is true, is this intended behavior?

Conversely, if every other item is true, this will return false if the last item is false. I can see why that might be useful in this case, but again, is that intended behavior?

src/Sarif/Visitors/SuppressVisitor.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (guids)
+            {
+                suppression.SetProperty("guid", Guid.NewGuid());
+            }
+
+            if (timestamps)
+            {
+                //suppression.SetProperty("timeUtc", timeUtc.ToString(SarifUtilities.SarifDateTimeFormatMillisecondsPrecision,
+                //                                                    CultureInfo.InvariantCulture));
```
shaopeng-gh, line 64: do we still need this comment

shaopeng-gh:
> 

un-needed change?

---
Refers to: src/Sarif.Multitool.Library/Sarif.Multitool.Library.csproj:66 in 31d2163. [](commit_id = 31d2163410bcdc226b29594f50882e8bd86d382e, deletion_comment = False)

src/Sarif.Multitool.Library/SuppressOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+        [Option(
+            "justification",
+            HelpText = "A string that provides the rationale for the suppressions",
+            Required = true)]
+        public string Justification { get; set; }
+
+        [Option(
+            "alias",
+            HelpText = "The account name associated with the suppression.")]
+        public string Alias { get; set; }
```
shaopeng-gh, line 22: >Alias [](http://example.com/codeflow?start=22&length=5)

just a thought, outside of Microsoft user may not immediately recognize Alias as Alias for user. do we like to change it to more common like user name, user alias, account name

src/Sarif.Multitool.Library/SuppressCommand.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                    WriteSarifFile(FileSystem, visitor.SarifLogVersionOne, actualOutputPath, options.Formatting, SarifContractResolverVersionOne.Instance);
+                }
+                else
+                {
+                    WriteSarifFile(FileSystem, reformattedLog, actualOutputPath, options.Formatting);
+                }
+
+                w.Stop();
+                Console.WriteLine($"Supress completed in {w.Elapsed}.");
```
yongyan-gh, line 58: like the idea to capture the execution time, not all commands implemented this, maybe we can move it to command base to work for all commands?

src/Sarif/Visitors/SuppressVisitor.cs
```diff
@@ -0,0 +1,78 @@
...
+                suppression.SetProperty("timeUtc", timeUtc);
+            }
+
+            if (expiryInDays > 0)
+            {
+                suppression.SetProperty("expiryUtc", expiryUtc.ToString(SarifUtilities.SarifDateTimeFormatMillisecondsPrecision,
```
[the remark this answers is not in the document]
eddynaka: just updated!
thanks

docs/multitool-usage.md
```diff
@@ -60,6 +61,9 @@ Sarif.Multitool merge C:\Input\*.sarif --recurse --output-directory=C:\Output\ -
 : Extract new Results only from New Baseline
 Sarif.Multitool query NewBaseline.sarif --expression "BaselineState == 'New'" --output Current.NewResults.sarif
 
+: Suppress Results
+Sarif.Multitool suppress current.sarif --justification "some justification" --alias "some alias" --guids --timestamps --expiryInDays 5 --output suppressed.sarif
```
yongyan-gh, line 65: >suppress [](http://example.com/codeflow?start=16&length=8)

missing a parameter "status"

eddynaka: added.

PR #2398, opened 2021-10-16
Fixing Load when deferred is enabled

### LINKED ISSUE #2397: SarifLog.Load(Stream, deferred = true) does not work
Using SarifLog.Load(Stream) always uses a JsonTextReader:
https://github.com/microsoft/sarif-sdk/blob/45e21b5671100f48807fb74e085342f196ce7513/src/Sarif/Core/SarifLog.cs#L62

But when actually getting around to the deferred deserialization it will throw an exception because it isn't using a `JsonPositionedTextReader`.
https://github.com/microsoft/sarif-sdk/blob/45e21b5671100f48807fb74e085342f196ce7513/src/Sarif/Readers/DeferredListConverter.cs#L34

eddynaka:
Fixes #2397

## Description
This change is fixing the `SarifLog.Load(stream, deferred)` implementation. To load a SARIF in deferred mode, we **must** use `JsonPositionedTextReader` to read the `stream`.
Once we fixed that, we also observed that the stream was getting closed before we read the SARIF, so, we added some protections such as `Stream.Synchronized` and `DelegatingStream`.

This will prevent the stream from being closed before needed.

`Stream.Synchronized`: https://docs.microsoft.com/dotnet/api/system.io.stream.synchronized?view=net-5.0

### DIFF
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -56,19 +56,14 @@ namespace Microsoft.CodeAnalysis.Sarif
             if (deferred)
             {
                 serializer.ContractResolver = new SarifDeferredContractResolver();
-                var jptr = new JsonPositionedTextReader(Stream.Synchronized(source));
-                return serializer.Deserialize<SarifLog>(jptr);
+                var jsonPositionedTextReader = JsonPositionedTextReader.FromStream(Stream.Synchronized(source));
+                return serializer.Deserialize<SarifLog>(jsonPositionedTextReader);
             }
-            else
+            using (var streamReader = new StreamReader(source))
+            using (var jsonTextReader = new JsonTextReader(streamReader))
             {
-                using (var sr = new StreamReader(source))
-                {
-                    using (var jtr = new JsonTextReader(sr))
-                    {
-                        // NOTE: Load with JsonSerializer.Deserialize, not JsonConvert.DeserializeObject, to avoid a string of the whole file in memory.
-                        return serializer.Deserialize<SarifLog>(jtr);
-                    }
-                }
+                // NOTE: Load with JsonSerializer.Deserialize, not JsonConvert.DeserializeObject, to avoid a string of the whole file in memory.
+                return serializer.Deserialize<SarifLog>(jsonTextReader);
             }
         }
 
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -58,7 +58,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 if (deferred)
                 {
                     serializer.ContractResolver = new SarifDeferredContractResolver();
-                    using (var jptr = new JsonPositionedTextReader(() => source))
+                    using (var jptr = new JsonPositionedTextReader(source))
                     {
                         return serializer.Deserialize<SarifLog>(jptr);
                     }
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -59,6 +59,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 var jsonPositionedTextReader = JsonPositionedTextReader.FromStream(Stream.Synchronized(source));
                 return serializer.Deserialize<SarifLog>(jsonPositionedTextReader);
             }
+
             using (var streamReader = new StreamReader(source))
             using (var jsonTextReader = new JsonTextReader(streamReader))
             {
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -56,7 +56,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             if (deferred)
             {
                 serializer.ContractResolver = new SarifDeferredContractResolver();
-                using (var jptr = new JsonPositionedTextReader(source))
+                using (var jptr = new JsonPositionedTextReader(Stream.Synchronized(source)))
                 {
                     return serializer.Deserialize<SarifLog>(jptr);
                 }
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -53,17 +53,17 @@ namespace Microsoft.CodeAnalysis.Sarif
         {
             var serializer = new JsonSerializer();
 
-            using (var sr = new StreamReader(source))
+            if (deferred)
             {
-                if (deferred)
+                serializer.ContractResolver = new SarifDeferredContractResolver();
+                using (var jptr = new JsonPositionedTextReader(source))
                 {
-                    serializer.ContractResolver = new SarifDeferredContractResolver();
-                    using (var jptr = new JsonPositionedTextReader(source))
-                    {
-                        return serializer.Deserialize<SarifLog>(jptr);
-                    }
+                    return serializer.Deserialize<SarifLog>(jptr);
                 }
-                else
+            }
+            else
+            {
+                using (var sr = new StreamReader(source))
                 {
                     using (var jtr = new JsonTextReader(sr))
                     {
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -56,10 +56,8 @@ namespace Microsoft.CodeAnalysis.Sarif
             if (deferred)
             {
                 serializer.ContractResolver = new SarifDeferredContractResolver();
-                using (var jptr = new JsonPositionedTextReader(Stream.Synchronized(source)))
-                {
-                    return serializer.Deserialize<SarifLog>(jptr);
-                }
+                var jptr = new JsonPositionedTextReader(Stream.Synchronized(source));
+                return serializer.Deserialize<SarifLog>(jptr);
             }
             else
             {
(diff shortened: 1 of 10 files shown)


eddynaka:
@gfs, this is the fix to the issue.
Let me know if that works.

Thank you

gfs:
Try the sarif log from here:

https://lgtm.com/projects/g/microsoft/OAT/?mode=list

src/Sarif/DelegatingStream.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly Stream stream;
+
+        /// <summary>
+        /// This is a wrapper for a stream that protects the underlying stream from being disposed.
+        /// </summary>
+        /// <param name="underlyingStream"></param>
+        internal DelegatingStream(Stream underlyingStream)
+        {
+            stream = underlyingStream;
+            stream.Seek(0L, SeekOrigin.Begin);
```
michaelcfanning, line 19: >Seek [](http://example.com/codeflow?start=19&length=4)

this will raise an exception if the stream isn't seekable.

src/Sarif/DelegatingStream.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+        public override bool CanRead => stream.CanRead;
+        public override bool CanSeek => stream.CanSeek;
+        public override bool CanWrite => stream.CanWrite;
+        public override long Length => stream.Length;
+        public override long Position
+        {
+            set
+            {
+                stream.Position = value;
```
michaelcfanning, line 29: >Position [](http://example.com/codeflow?start=23&length=8)

can't do this if the stream isn't seekable. there are other methods that will throw as well, SetLength, I think. check the online docs on CanSeek for more.

src/Sarif/DelegatingStream.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            get => stream.Position;
+        }
+        public override void Flush()
+        {
+            stream.Flush();
+        }
+        public override int Read(byte[] buffer, int offset, int count)
```
michaelcfanning, line 37: >override [](http://example.com/codeflow?start=15&length=8)

This file needs vertical whitespace between methods.

src/Sarif/DelegatingStream.cs
michaelcfanning: So. thinking this over, I suppose you can assume that a user will use the stream properly (i.e., understand if it isn't seekable) or otherwise get an exception when the wrapped methods are called.

You shouldn't seek here because you will break this class for all non-seekable streams. You need a unit test for this case, by the way.

Why, exactly, are you setting the stream pointer to 0? Why is that our job in this wrapper? Shouldn't it not try to interfere with the stream in any way? 

My suggestion is just delete this line, leave the rest alone, and add relevant tests.

src/Sarif/Core/SarifLog.cs
```diff
@@ -51,18 +51,24 @@ public static SarifLog Load(string sarifFilePath)
...
+                return serializer.Deserialize<SarifLog>(jptr);
             }
-
-            using (StreamReader sr = new StreamReader(source))
-            using (JsonTextReader jtr = new JsonTextReader(sr))
+            else
             {
-                // NOTE: Load with JsonSerializer.Deserialize, not JsonConvert.DeserializeObject, to avoid a string of the whole file in memory.
-                return serializer.Deserialize<SarifLog>(jtr);
+                using (var sr = new StreamReader(source))
```
michaelcfanning, line 64: >sr [](http://example.com/codeflow?start=27&length=2)

we don't use abbreviations like 'sr' and 'jtr'. Please use full words, like 'reader', 'streamReader' or 'jsonTestReader

src/Sarif/Core/SarifLog.cs
```diff
@@ -51,18 +51,24 @@ public static SarifLog Load(string sarifFilePath)
...
             {
                 serializer.ContractResolver = new SarifDeferredContractResolver();
+                var jptr = new JsonPositionedTextReader(Stream.Synchronized(source));
+                return serializer.Deserialize<SarifLog>(jptr);
             }
-
-            using (StreamReader sr = new StreamReader(source))
-            using (JsonTextReader jtr = new JsonTextReader(sr))
+            else
             {
```
michaelcfanning, line 63: you return in the previous conditional clause, so why use the else here? you can get rid of this clause and left shift the code (in terms of indentation).

michaelcfanning (changes requested):
:clock1:

src/Sarif/DelegatingStream.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal class DelegatingStream : Stream
```
michaelcfanning, line 8: >DelegatingStream [](http://example.com/codeflow?start=19&length=16)

I see no delegating stream tests.

Adding unit tests for all new classes is an important quality standard on our team.

src/Sarif/DelegatingStream.cs
gfs: Delegated reading already does not support non seekable streams (https://github.com/microsoft/sarif-sdk/blob/ea44355584c161aec784ee9d2aa3c57f84875906/src/Sarif/Readers/DeferredDictionary.cs#L69) and an exception is thrown for the argument in this PR if you try to create a jsonpositionedtextreader with a non seekable stream.

src/Sarif/Readers/JsonPositionedTextReader.cs
```diff
@@ -37,6 +37,19 @@ public class JsonPositionedTextReader : JsonTextReader
...
         { }
 
+
+        /// <summary>
+        /// Create a JsonPositionedTextReader based on a Stream and passes this to deferred collections.
+        /// </summary>
+        /// <param name="stream"></param>
+        public JsonPositionedTextReader(Stream stream) : this(() => new DelegatingStream(stream))
+        {
+            if (!stream.CanSeek)
```
michaelcfanning, line 47: >stream [](http://example.com/codeflow?start=17&length=6)

too late! you already crashed in the DelegatingStream ctor when you tried to seek to position 0. right?

src/Sarif/Core/SarifLog.cs
gfs: This code wasn't changed, just indented.

https://github.com/microsoft/sarif-sdk/blob/45e21b5671100f48807fb74e085342f196ce7513/src/Sarif/Core/SarifLog.cs#L61-L66

src/Sarif/DelegatingStream.cs
gfs: > So. thinking this over, I suppose you can assume that a user will use the stream properly (i.e., understand if it isn't seekable) or otherwise get an exception when the wrapped methods are called.
> 
> You shouldn't seek here because you will break this class for all non-seekable streams. You need a unit test for this case, by the way.

I understand the concern now. I made the class internal and it was specifically designed for use with JsonPositionedTextReader - which does not support non-seekable streams in the first place.  The constructor calls LineMappingStreamReader.ctor which calls FindBomWidth which seeks.  You can verify this by getting rid of the seek in the ctor and throwing an exception on the overloaded seek method.

> 
> Why, exactly, are you setting the stream pointer to 0? Why is that our job in this wrapper? Shouldn't it not try to interfere with the stream in any way?

The wrapper is a contribution from another project where we used it as the workaround for the current behavior. In that project we wanted to read from 0 and we always knew the source of our stream was seekable. I think you're right that here it isn't needed.

> 
> My suggestion is just delete this line, leave the rest alone, and add relevant tests.

src/Sarif/DelegatingStream.cs
gfs: What behavior would you prefer? This is essentially just allowing the exceptions being thrown by the underlying stream to surface.  The alternative would be to make a seekable wrapper around non seekable streams.

gfs:
I have a PR with some improvements: https://github.com/microsoft/sarif-sdk/pull/2405/files

src/Sarif/DelegatingStream.cs
eddynaka: Gabe added in iteration 7.

src/Sarif/DelegatingStream.cs
eddynaka: Gabe added.

src/Sarif/Core/SarifLog.cs
eddynaka: done

src/Sarif/Core/SarifLog.cs
eddynaka: Gabe just renamed.

src/Sarif/DelegatingStream.cs
michaelcfanning: @eddynaka can fix this, I say don't do anything, just make this wrapper a lightweight delegate. If someone tries to call methods illegally on a non-seekable stream they'll get the same behavior as if they called the underlying instance directly.

eddynaka: done :)

src/Sarif/Readers/JsonPositionedTextReader.cs
eddynaka: we removed the seek to position 0.

eddynaka:
@gfs, just to let you know, I'm going to close this PR and create a new one.
@michaelcfanning is facing some issues with CodeFlow.

Issue #2077, opened 2020-09-17
Non-"fail" results don't deserialize properly


lgolding:
Consider a result with `"kind"`: `"review"`, and with `"level"` omitted. The spec specifies an interaction between these properties:

> If `level` has any value other than `"none"` and `kind` is present, then `kind` **SHALL** have the value `"fail"`. [§3.27.9]

> If `kind` (§3.27.9) has any value other than `"fail"`, then if `level` is absent, it **SHALL** default to `"none"`, and if it is present, it **SHALL** have the value `"none"`. [§3.27.10]

... so this result should be equivalent to  `"kind"`: `"review", "level": "none"`. But actually, it deserializes as `"kind"`: `"fail", "level": "warning"`

The root cause is in the SDK, in the interaction between the Newtonsoft deserializer and our hand-coded (NotYetAutoGenerated) version of Result.cs. The problem (confirmed in the debugger) is that Newtonsoft first sets `Kind` to `Review`, which sets `Level` to `None`. So far, so good. But it and then sets `Level` to `Warning` (because it’s absent!), which sets `Kind` to `Fail`.

@michaelcfanning FYI

PR #2399, opened 2021-10-19
Sarif Level - just propose a way to fix, for discussion

shaopeng-gh:
Related:
https://github.com/microsoft/sarif-sdk/issues/2077
https://github.com/microsoft/sarif-sdk/issues/2024
https://github.com/microsoft/sarif-sdk/pull/2383

### DIFF
--- a/src/Sarif/Autogenerated/Result.cs
+++ b/src/Sarif/Autogenerated/Result.cs
@@ -61,44 +61,42 @@ namespace Microsoft.CodeAnalysis.Sarif
         // than fail, we automatically set Level to FailureLevel.None. Similarly, if Level is 
         // set to any value other than None we update Kind to ResultKind.Fail
 
-        private ResultKind _kind;
+        private ResultKind? _kind;
 
         /// <summary>
         /// A value that categorizes results by evaluation state.
         /// </summary>
         [DataMember(Name = "kind", IsRequired = false, EmitDefaultValue = false)]
-        [DefaultValue(ResultKind.Fail)]
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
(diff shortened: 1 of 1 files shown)


src/Sarif/Autogenerated/Result.cs
```diff
@@ -61,44 +61,42 @@ public SarifNodeKind SarifNodeKind
...
         /// <summary>
         /// A value that categorizes results by evaluation state.
         /// </summary>
         [DataMember(Name = "kind", IsRequired = false, EmitDefaultValue = false)]
-        [DefaultValue(ResultKind.Fail)]
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
         [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
-        public ResultKind Kind
+        public ResultKind? Kind
```
shaopeng-gh, line 72: >Kind [](http://example.com/codeflow?start=27&length=4)

the main change is for Level, I see it is similar, see if we want to change as well

src/Sarif/Autogenerated/Result.cs
```diff
@@ -61,44 +61,42 @@ public SarifNodeKind SarifNodeKind
...
         [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.EnumConverter))]
-        public FailureLevel Level
+        public FailureLevel? Level
         {
             get { return _level; }
             set
             {
                 _level = value;
-                if (_level != FailureLevel.None)
+                if (_level != FailureLevel.None && !_kind.HasValue)
```
shaopeng-gh, line 99: >&& !_kind.HasValue [](http://example.com/codeflow?start=48&length=18)

I think we should only set the other field, only when it is not set.

shaopeng-gh:
closing this, was just the propose fix high level concept, the actual PR is:
https://github.com/microsoft/sarif-sdk/pull/2404

PR #2402, opened 2021-10-27
Enable automation details for command base

eddynaka:
## Description

Imagine SARIF Pattern Matcher being run by someone, and we want to automatically ingest the output SARIF somehow.

With this, we could ask the user to run SARIF Pattern Matcher and pass `--automationId` and `--automationGuid`, that information would be stored in the SARIF and we would know that the SARIF was generated by a specific team.

## Follow-up change

Change the `commandBase` to accept a URI and this URI will be used to send the generated SARIF to that endpoint.

cc: @cfaucon

### DIFF
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -27,6 +27,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         internal ConsoleLogger _consoleLogger;
 
+        private Run _run;
         private Tool _tool;
         private TContext _rootContext;
         private CacheByFileHashLogger _cacheByFileHashLogger;
@@ -432,6 +433,17 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                         SarifLogger sarifLogger;
 
+                        _run = new Run();
+
+                        if (!string.IsNullOrWhiteSpace(analyzeOptions.AutomationId) || !string.IsNullOrWhiteSpace(analyzeOptions.AutomationGuid))
+                        {
+                            _run.AutomationDetails = new RunAutomationDetails
+                            {
+                                Id = analyzeOptions.AutomationId,
+                                Guid = analyzeOptions.AutomationGuid
+                            };
+                        }
+
                         if (analyzeOptions.SarifOutputVersion != SarifVersion.OneZeroZero)
                         {
                             sarifLogger = new SarifLogger(
@@ -440,7 +452,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                                     dataToInsert,
                                     dataToRemove,
                                     tool: _tool,
-                                    run: null,
+                                    run: _run,
                                     analysisTargets: targets,
                                     quiet: analyzeOptions.Quiet,
                                     invocationTokensToRedact: GenerateSensitiveTokensList(),
@@ -456,7 +468,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                                     dataToInsert,
                                     dataToRemove,
                                     tool: _tool,
-                                    run: null,
+                                    run: _run,
                                     analysisTargets: targets,
                                     invocationTokensToRedact: GenerateSensitiveTokensList(),
                                     invocationPropertiesToLog: analyzeOptions.InvocationPropertiesToLog,
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -433,16 +433,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                         SarifLogger sarifLogger;
 
-                        _run = new Run();
-
-                        if (!string.IsNullOrWhiteSpace(analyzeOptions.AutomationId) || !string.IsNullOrWhiteSpace(analyzeOptions.AutomationGuid))
+                        _run = new Run()
                         {
-                            _run.AutomationDetails = new RunAutomationDetails
+                            AutomationDetails = new RunAutomationDetails
                             {
                                 Id = analyzeOptions.AutomationId,
                                 Guid = analyzeOptions.AutomationGuid
-                            };
-                        }
+                            }
+                        };
 
                         if (analyzeOptions.SarifOutputVersion != SarifVersion.OneZeroZero)
                         {
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -112,12 +112,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         [Option(
             "automationId",
-            HelpText = "An id that will be used in the output SARIF and show in the 'automationDetails.id' under 'run'.")]
+            HelpText = "An id that will be persisted to the 'Run.AutomationDetails.Id' property. See section '3.17.3' of the SARIF specification for more information.")]
         public string AutomationId { get; set; }
 
         [Option(
             "automationGuid",
-            HelpText = "A guid that will be used in the output SARIF and show in the 'automationDetails.guid' under 'run'.")]
+            HelpText = "A guid that will be persisted to the 'Run.AutomationDetails.Guid' property. See section '3.17.4' of the SARIF specification for more information.")]
         public string AutomationGuid { get; set; }
     }
 }
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -107,7 +107,17 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         [Option(
             "baseline",
-            HelpText = "A Sarif file to be used as baseline")]
+            HelpText = "A SARIF file to be used as baseline.")]
         public string BaselineSarifFile { get; set; }
+
+        [Option(
+            "automationId",
+            HelpText = "An id that will be used in the output SARIF and show in the 'automationDetails.id' under 'run'.")]
+        public string AutomationId { get; set; }
+
+        [Option(
+            "automationGuid",
+            HelpText = "A guid that will be used in the output SARIF and show in the 'automationDetails.guid' under 'run'.")]
+        public string AutomationGuid { get; set; }
     }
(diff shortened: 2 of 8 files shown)


src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -432,6 +433,17 @@ private void InitializeOutputFile(TOptions analyzeOptions, TContext context, ISe
 
                         SarifLogger sarifLogger;
 
+                        _run = new Run();
+
+                        if (!string.IsNullOrWhiteSpace(analyzeOptions.AutomationId) || !string.IsNullOrWhiteSpace(analyzeOptions.AutomationGuid))
+                        {
```
michaelcfanning, line 439: Do you actually need this check? i.e., can't we just create the object and if both items are null it doesn't get emitted?

If that's not true, change the RunAutomationDetails ShouldSerializeXXX methods to make that happen. Add tests for this, too, if they don't exist.

michaelcfanning (changes requested):
:clock1:

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -107,7 +107,17 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
 
         [Option(
             "baseline",
-            HelpText = "A Sarif file to be used as baseline")]
+            HelpText = "A SARIF file to be used as baseline.")]
         public string BaselineSarifFile { get; set; }
+
+        [Option(
+            "automationId",
+            HelpText = "An id that will be used in the output SARIF and show in the 'automationDetails.id' under 'run'.")]
```
michaelcfanning, line 115: >An id that will be used in the output SARIF and show in the 'automationDetails.id' under 'run'. [](http://example.com/codeflow?start=24&length=95)

"An id that will be persisted to the 'Run.AutomationDetails.Id' property. See section '3.17.3' of the SARIF specification."

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -107,7 +107,17 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
...
+        public string AutomationId { get; set; }
+
+        [Option(
+            "automationGuid",
+            HelpText = "A guid that will be used in the output SARIF and show in the 'automationDetails.guid' under 'run'.")]
```
michaelcfanning, line 120: >A guid that will be used in the output SARIF and show in the 'automationDetails.guid' under 'run'. [](http://example.com/codeflow?start=24&length=98)

"A guid that will be persisted to the 'Run.AutomationDetails.Guid' property. See section '3.17.4' of the SARIF specification for more information.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -731,6 +731,15 @@ private void InitializeOutputFile(TOptions analyzeOptions, TContext context)
 
                         _run = new Run();
 
+                        if (!string.IsNullOrWhiteSpace(analyzeOptions.AutomationId) || !string.IsNullOrWhiteSpace(analyzeOptions.AutomationGuid))
+                        {
+                            _run.AutomationDetails = new RunAutomationDetails
+                            {
+                                Id = analyzeOptions.AutomationId,
+                                Guid = analyzeOptions.AutomationGuid
+                            };
```
michaelcfanning, line 740: Use this. Make sure this doesn't emit empty values if these arguments aren't provided;

_run = new Run()
 {
    AutomationDetails = new RunAutomationDetails
    {
        Id = analyzeOptions.AutomationId,
        Guid = analyzeOptions.AutomationGuid
  }
};

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1265,7 +1265,9 @@ private static void RunResultsCachingTestCase(ResultsCachingTestCase testCase, b
                 OutputFilePath = testCase.PersistLogFileToDisk ? Guid.NewGuid().ToString() : null,
                 TargetFileSpecifiers = new string[] { Guid.NewGuid().ToString() },
                 Kind = new List<ResultKind> { ResultKind.Fail },
-                Level = new List<FailureLevel> { FailureLevel.Warning, FailureLevel.Error }
+                Level = new List<FailureLevel> { FailureLevel.Warning, FailureLevel.Error },
+                AutomationId = "Test Automation",
+                AutomationGuid = Guid.NewGuid().ToString()
```
michaelcfanning, line 1270: >AutomationGuid [](http://example.com/codeflow?start=16&length=14)

So you don't have explicit tests somewhere for passing only one of these, or none.

Also, you've dropped these tests into an existing test, which is not necessarily bad, but again, we don't have a directed test. 

Did you cover the multithreaded case? Please verify with code coverage that you have coverage of all your changes.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
eddynaka: I had to add three ShouldSerialize:
1. ShouldSerializeAutomationDetails
2. ShouldSerializeId
3. ShouldSerializeGuid

if id and guid are null/empty, we should not emit automation details and that check can only occur in the 'run'.

Once we have an AutomationDetails, if one of those is empty, that property should not be emitted.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
eddynaka: added.
thanks!

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
eddynaka: added.
thanks!

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: done!

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
eddynaka: just added a new test that will cover the following scenarios:
- id only
- guid only
- id and guid

Also, the test verifies single and multithread.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1311,6 +1363,12 @@ private static void RunResultsCachingTestCase(ResultsCachingTestCase testCase, b
                 .BeEquivalentTo(runWithCaching.Invocations?[0].ToolExecutionNotifications);
         }
 
+        private static void EnhanceOptions(TestAnalyzeOptions current, TestAnalyzeOptions enhancement)
```
eddynaka, line 1368: >EnhanceOptions [](http://example.com/codeflow?start=28&length=14)

This method will be used to enhance the options with custom information. That will enable us to test multiple combinations with the same test, just changing the enhancement variable.

src/Sarif/Core/Run.cs
```diff
@@ -220,6 +220,12 @@ public bool ShouldSerializeGraphs()
             return this.Graphs.HasAtLeastOneNonDefaultValue(Graph.ValueComparer);
         }
 
+        public bool ShouldSerializeAutomationDetails()
```
michaelcfanning, line 223: >ShouldSerializeAutomationDetails [](http://example.com/codeflow?start=20&length=32)

Why did you only test for these two properties? AutomationDetails has many more.

src/Sarif/Core/RunAutomationDetails.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Information that describes a run's identity and role within an engineering system process.
+    /// </summary>
+    public partial class RunAutomationDetails
```
michaelcfanning, line 9: >RunAutomationDetails [](http://example.com/codeflow?start=25&length=20)

where are the other checks for this type?

src/Test.UnitTests.Sarif/Core/RunTests.cs
```diff
@@ -487,9 +489,88 @@
...
+
+            sarifLog.Runs[0].AutomationDetails.Id = null;
+            sarifLog.Runs[0].AutomationDetails.Guid = null;
+
+            AsserAutomationDetailsValues(sarifLog, string.Empty, string.Empty);
+
+            sarifLog.Runs[0].AutomationDetails.Id = whiteSpace;
+            sarifLog.Runs[0].AutomationDetails.Guid = whiteSpace;
+
+            AsserAutomationDetailsValues(sarifLog, whiteSpace, whiteSpace);
```
michaelcfanning, line 547: >AsserAutomationDetailsValues [](http://example.com/codeflow?start=21&length=28)

I think you meant assess. Please just use Validate, this is a very common term we use in test.

michaelcfanning, line 547: >AsserAutomationDetailsValues [](http://example.com/codeflow?start=21&length=28)

Your API is confusing. You have a log which already has an id and guid set on it. Why are you repeating those arguments instead of extracting them from the log?

If you just want the arguments, then instantiate the log file here. It's odd that you've split this work into two.

src/Sarif/Core/RunAutomationDetails.cs
eddynaka: thanks for the suggestion.
I'm adding the remaining logic :)

src/Sarif/Core/Run.cs
eddynaka: Adding the remaining properties.
thanks!

src/Test.UnitTests.Sarif/Core/RunTests.cs
eddynaka: thanks for the suggestion!
re-worked the method :)

eddynaka: Changed! thanks!

PR #2403, opened 2021-10-28
Enable postUri argument in AnalyzeCommandBase

eddynaka:
## Description

- Adding a new argument called `postUri` which will accept a URI. This URI will be used to send the outputFile to the URI using the `POST` method.
- Adding two new APIs to `SarifLog` to `POST` a SARIF to a URI.

This will facilitate in the case where you have a centralized place to send and analyze your SARIF files.

Current code coverage:
image
image
image

### DIFF
--- a/src/Sarif.Driver/DriverExtensionMethods.cs
+++ b/src/Sarif.Driver/DriverExtensionMethods.cs
@@ -169,6 +169,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             // baseline process now depends on output file
             valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.BaselineSarifFile));
 
+            // post SARIF process now depends on output file
+            valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.PostUri));
+            valid &= !(!string.IsNullOrWhiteSpace(options.PostUri) && !Uri.IsWellFormedUriString(options.PostUri, UriKind.Absolute));
+
             if (!valid)
             {
                 context.RuntimeErrors |= RuntimeConditions.InvalidCommandLineOption;
--- a/src/Sarif.Driver/DriverExtensionMethods.cs
+++ b/src/Sarif.Driver/DriverExtensionMethods.cs
@@ -164,12 +164,25 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         {
             bool valid = true;
 
-            valid &= !(options.Quiet && string.IsNullOrWhiteSpace(options.OutputFilePath));
+            if (options.Quiet && string.IsNullOrWhiteSpace(options.OutputFilePath))
+            {
+                valid = false;
+            }
 
-            valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.BaselineSarifFile));
+            if (string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.BaselineSarifFile))
+            {
+                valid = false;
+            }
 
-            valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.PostUri));
-            valid &= !(!string.IsNullOrWhiteSpace(options.PostUri) && !Uri.IsWellFormedUriString(options.PostUri, UriKind.Absolute));
+            if (string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.PostUri))
+            {
+                valid = false;
+            }
+
+            if (!string.IsNullOrWhiteSpace(options.PostUri) && !Uri.IsWellFormedUriString(options.PostUri, UriKind.Absolute))
+            {
+                valid = false;
+            }
 
             if (!valid)
             {
--- a/src/Sarif.Driver/DriverExtensionMethods.cs
+++ b/src/Sarif.Driver/DriverExtensionMethods.cs
@@ -166,10 +166,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
             valid &= !(options.Quiet && string.IsNullOrWhiteSpace(options.OutputFilePath));
 
-            // baseline process now depends on output file
             valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.BaselineSarifFile));
 
-            // post SARIF process now depends on output file
             valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.PostUri));
             valid &= !(!string.IsNullOrWhiteSpace(options.PostUri) && !Uri.IsWellFormedUriString(options.PostUri, UriKind.Absolute));
 
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -124,7 +124,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 }
                 catch (Exception ex)
                 {
-                    RuntimeErrors |= RuntimeConditions.ExceptionPostingLog;
+                    RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
                     ExecutionException = ex;
                     return FAILURE;
                 }
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -118,6 +118,17 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                     ExecutionException = ex;
                     return FAILURE;
                 }
+
+                try
+                {
+                    Export(_rootContext, options, FileSystem);
+                }
+                catch (Exception ex)
+                {
+                    RuntimeErrors |= RuntimeConditions.ExceptionExportingLogFile;
+                    ExecutionException = ex;
+                    return FAILURE;
+                }
             }
 
             if (options.RichReturnCode)
@@ -181,13 +192,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         {
             bool succeeded = true;
 
-            succeeded &= ValidateFile(context, analyzeOptions.OutputFilePath, shouldExist: null);
-            succeeded &= ValidateFile(context, analyzeOptions.ConfigurationFilePath, shouldExist: true);
-            succeeded &= ValidateFiles(context, analyzeOptions.PluginFilePaths, shouldExist: true);
-            succeeded &= ValidateFile(context, analyzeOptions.BaselineSarifFile, shouldExist: true);
+            succeeded &= ValidateFile(context, analyzeOptions.OutputFilePath, DefaultPolicyName, shouldExist: null);
+            succeeded &= ValidateFile(context, analyzeOptions.ConfigurationFilePath, DefaultPolicyName, shouldExist: true);
+            succeeded &= ValidateFiles(context, analyzeOptions.PluginFilePaths, DefaultPolicyName, shouldExist: true);
+            succeeded &= ValidateFile(context, analyzeOptions.BaselineSarifFile, DefaultPolicyName, shouldExist: true);
             succeeded &= ValidateInvocationPropertiesToLog(context, analyzeOptions.InvocationPropertiesToLog);
             succeeded &= ValidateOutputFileCanBeCreated(context, analyzeOptions.OutputFilePath, analyzeOptions.Force);
             succeeded &= analyzeOptions.ValidateOutputOptions(context);
+            succeeded &= ValidateExport(analyzeOptions.OutputFilePath, analyzeOptions.Export);
 
             if (!succeeded)
             {
@@ -195,85 +207,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 //  whenever something goes wrong. #2260 https://github.com/microsoft/sarif-sdk/issues/2260
                 ThrowExitApplicationException(context, ExitReason.InvalidCommandLineOption);
             }
-        }
-
-        private bool ValidateFiles(TContext context, IEnumerable<string> filePaths, bool shouldExist)
-        {
-            if (filePaths == null) { return true; }
-
-            bool succeeded = true;
-
-            foreach (string filePath in filePaths)
-            {
-                succeeded &= ValidateFile(context, filePath, shouldExist);
-            }
-
-            return succeeded;
-        }
-
-        private bool ValidateFile(TContext context, string filePath, bool? shouldExist)
-        {
-            if (filePath == null || filePath == DefaultPolicyName) { return true; }
-
-            Exception exception = null;
-
-            try
-            {
-                bool fileExists = FileSystem.FileExists(filePath);
-
-                if (fileExists || shouldExist == null || !shouldExist.Value)
-                {
-                    return true;
-                }
-
-                Errors.LogMissingFile(context, filePath);
-            }
-            catch (IOException ex) { exception = ex; }
-            catch (SecurityException ex) { exception = ex; }
-            catch (UnauthorizedAccessException ex) { exception = ex; }
-
-            if (exception != null)
-            {
-                Errors.LogExceptionAccessingFile(context, filePath, exception);
-            }
-
-            return false;
-        }
-
-        private static bool ValidateInvocationPropertiesToLog(TContext context, IEnumerable<string> propertiesToLog)
-        {
-            bool succeeded = true;
-
-            if (propertiesToLog != null)
-            {
-                List<string> validPropertyNames = typeof(Invocation).GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                    .Select(propInfo => propInfo.Name.ToUpperInvariant())
-                    .ToList();
-
-                foreach (string propertyName in propertiesToLog)
-                {
-                    if (!validPropertyNames.Contains(propertyName.ToUpperInvariant()))
-                    {
-                        Errors.LogInvalidInvocationPropertyName(context, propertyName);
-                        succeeded = false;
-                    }
-                }
-            }
-
-            return succeeded;
-        }
-
-        private bool ValidateOutputFileCanBeCreated(TContext context, string outputFilePath, bool force)
-        {
-            bool succeeded = true;
-
-            if (!DriverUtilities.CanCreateOutputFile(outputFilePath, force, FileSystem))
-            {
-                Errors.LogOutputFileAlreadyExists(context, outputFilePath);
-                succeeded = false;
-            }
-
-            return succeeded;
         }
 
         internal AggregatingLogger InitializeLogger(AnalyzeOptionsBase analyzeOptions)
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -119,7 +119,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                 try
                 {
-                    Export(_rootContext, options, FileSystem);
+                    ProcessPostUri(_rootContext, options, FileSystem);
                 }
                 catch (Exception ex)
                 {
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -120,7 +120,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                 try
                 {
-                    PostLogFile(_rootContext, options, FileSystem);
+                    PostLogFile(options.PostUri, options.OutputFilePath, FileSystem);
                 }
                 catch (Exception ex)
                 {
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -119,7 +119,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                 try
                 {
-                    ProcessPostUri(_rootContext, options, FileSystem);
+                    PostLogFile(_rootContext, options, FileSystem);
                 }
                 catch (Exception ex)
                 {
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -197,7 +197,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             succeeded &= ValidateInvocationPropertiesToLog(context, analyzeOptions.InvocationPropertiesToLog);
             succeeded &= ValidateOutputFileCanBeCreated(context, analyzeOptions.OutputFilePath, analyzeOptions.Force);
             succeeded &= analyzeOptions.ValidateOutputOptions(context);
-            succeeded &= ValidatePostUri(analyzeOptions.OutputFilePath, analyzeOptions.PostUri);
 
             if (!succeeded)
             {
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -5,9 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Security;
 
 using Microsoft.CodeAnalysis.Sarif.Writers;
 
@@ -125,7 +123,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 }
                 catch (Exception ex)
                 {
-                    RuntimeErrors |= RuntimeConditions.ExceptionExportingLogFile;
+                    RuntimeErrors |= RuntimeConditions.ExceptionPostingLog;
                     ExecutionException = ex;
                     return FAILURE;
                 }
@@ -199,7 +197,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             succeeded &= ValidateInvocationPropertiesToLog(context, analyzeOptions.InvocationPropertiesToLog);
             succeeded &= ValidateOutputFileCanBeCreated(context, analyzeOptions.OutputFilePath, analyzeOptions.Force);
             succeeded &= analyzeOptions.ValidateOutputOptions(context);
-            succeeded &= ValidateExportUri(analyzeOptions.OutputFilePath, analyzeOptions.ExportUri);
+            succeeded &= ValidatePostUri(analyzeOptions.OutputFilePath, analyzeOptions.PostUri);
 
             if (!succeeded)
             {
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -25,6 +25,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         internal ConsoleLogger _consoleLogger;
 
+        private Run _run;
         private Tool _tool;
         private TContext _rootContext;
         private CacheByFileHashLogger _cacheByFileHashLogger;
@@ -362,6 +363,15 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                         SarifLogger sarifLogger;
 
+                        _run = new Run()
+                        {
+                            AutomationDetails = new RunAutomationDetails
+                            {
+                                Id = analyzeOptions.AutomationId,
+                                Guid = analyzeOptions.AutomationGuid
+                            }
+                        };
+
                         if (analyzeOptions.SarifOutputVersion != SarifVersion.OneZeroZero)
                         {
                             sarifLogger = new SarifLogger(
@@ -370,7 +380,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                                     dataToInsert,
                                     dataToRemove,
                                     tool: _tool,
-                                    run: null,
+                                    run: _run,
                                     analysisTargets: targets,
                                     quiet: analyzeOptions.Quiet,
                                     invocationTokensToRedact: GenerateSensitiveTokensList(),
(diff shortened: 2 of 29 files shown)


src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -109,5 +109,10 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
             "baseline",
             HelpText = "A Sarif file to be used as baseline")]
         public string BaselineSarifFile { get; set; }
+
+        [Option(
+            "exportUri",
```
michaelcfanning, line 114: >exportUri [](http://example.com/codeflow?start=13&length=9)

Suggest you call this 'postUri'. Makes sense, right? :) Generally, we don't 'export' to web endpoints, we publish or post.

src/Sarif/RuntimeConditions.cs
```diff
@@ -49,9 +50,11 @@ public enum RuntimeConditions : uint
         ExceptionInstantiatingSkimmers = 0x8000,
         OutputFileAlreadyExists = 0x10000,
         ExceptionProcessingBaseline = 0x20000,
+        ExceptionExportingLogFile = 0x40000,
```
michaelcfanning, line 53: >ExceptionExportingLogFile [](http://example.com/codeflow?start=8&length=25)

ExceptionPostingLog

src/Sarif.Driver/Sdk/ExitReason.cs
```diff
@@ -13,6 +13,7 @@ public enum ExitReason
         NoRulesLoaded,
         NoValidAnalysisTargets,
         InvalidCommandLineOption,
-        ExceptionProcessingBaseline
+        ExceptionProcessingBaseline,
+        ExceptionExportingLogFile
```
michaelcfanning, line 17: >ExceptionExportingLogFile [](http://example.com/codeflow?start=8&length=25)

ExceptionPostingLogFile

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -37,6 +42,102 @@
             return assemblies;
         }
 
+        public bool ValidateExportUri(string outputFilePath, string exportUri)
```
michaelcfanning, line 45: >ValidateExportUri [](http://example.com/codeflow?start=20&length=17)

This isn't a good public helper. Why would our plug-in driver class has a simple URI validation API that it exposes to all classes?

michaelcfanning, line 45: >bool [](http://example.com/codeflow?start=15&length=4)

Why is this an instance method? I don't see that you are operating against instance fields.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -525,100 +536,21 @@
...
-            succeeded &= ValidateFiles(context, options.PluginFilePaths, shouldExist: true);
-            succeeded &= ValidateFile(context, options.BaselineSarifFile, shouldExist: true);
+            succeeded &= ValidateFile(context, options.OutputFilePath, DefaultPolicyName, shouldExist: null);
+            succeeded &= ValidateFile(context, options.ConfigurationFilePath, DefaultPolicyName, shouldExist: true);
+            succeeded &= ValidateFiles(context, options.PluginFilePaths, DefaultPolicyName, shouldExist: true);
+            succeeded &= ValidateFile(context, options.BaselineSarifFile, DefaultPolicyName, shouldExist: true);
             succeeded &= ValidateInvocationPropertiesToLog(context, options.InvocationPropertiesToLog);
             succeeded &= ValidateOutputFileCanBeCreated(context, options.OutputFilePath, options.Force);
             succeeded &= options.ValidateOutputOptions(context);
+            succeeded &= ValidateExportUri(options.OutputFilePath, options.ExportUri);
```
michaelcfanning, line 546: >ValidateExportUri [](http://example.com/codeflow?start=25&length=17)

Don't add this helper, instead update ValidateOutputOptions. Posting to a URI is a new kind of output.

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -37,6 +42,102 @@
...
+
+            if (exception != null)
+            {
+                Errors.LogExceptionAccessingFile(context, filePath, exception);
+            }
+
+            return false;
+        }
+
+        public bool ValidateFiles(IAnalysisContext context, IEnumerable<string> filePaths, string policyName, bool shouldExist)
```
michaelcfanning, line 90: >bool [](http://example.com/codeflow?start=15&length=4)

I'm seeing lots of public helpers in this class. Many of them look like they should be static methods. I am not seeing any granular unit tests for these classes.

First, please make sure you are adding minimal code to our SDK. Second, make sure that your accessibility is correct, either public or internal. Make sure things are static that should be. Write unit tests and report your coverage to me.

michaelcfanning (approved):
:shipit:

michaelcfanning (changes requested):
:clock1:

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
eddynaka: thanks for the suggestion :)

src/Sarif/RuntimeConditions.cs
eddynaka: renamed! :)

src/Sarif.Driver/Sdk/ExitReason.cs
eddynaka: renamed!

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
eddynaka: thanks for the suggestion, when i first created it, i followed all the other methods (all were public).

Now, i'm changing to internal or internal static.

thanks for the suggestions :)

eddynaka: thanks for the suggestion, when i first created it, i followed all the other methods (all were public).

Now, i'm changing to internal or internal static.

thanks for the suggestions :)

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
eddynaka: removed the method and updated this. thanks for the suggestion :)

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
eddynaka: Those methods were in the AnalyzeCommandBase/MultithreadCommandBase (both had the same methods). When I moved I didn't change their accessibility.

Now, I'm changing to internal or internal static.

thanks for the suggestions :)

src/Test.Utilities.Sarif/TestRuleBehaviors.cs
```diff
@@ -40,6 +41,8 @@ public enum TestRuleBehaviors
         // Assume one or more options are invalid
         RegardOptionsAsInvalid = 0x4000,
 
-        RaiseExceptionProcessingBaseline = 0x8000
+        RaiseExceptionProcessingBaseline = 0x8000,
+
+        RaiseExceptionProcessingPostUri = 0x10000
```
michaelcfanning, line 46: >RaiseExceptionProcessingPostUri [](http://example.com/codeflow?start=8&length=31)

We don't 'process' a URI, we are posting the log contents. Or we are publishing it.

RaiseExceptionPostingLogFile

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    sb.AppendLine($"The test '{testCase.Title}' was expecting '{testCase.Expected}' but found '{current}'.");
+                }
+            }
+
+            sb.Length.Should().Be(0, sb.ToString());
+        }
+
+        [Fact]
+        public void PluginDriverCommand_ProcessPostUri()
```
michaelcfanning, line 75: >PluginDriverCommand_ProcessPostUri [](http://example.com/codeflow?start=20&length=34)

PostLogFile

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -310,6 +310,22 @@ public void ExceptionRaisedProcessingBaseline()
             );
         }
 
+        [Fact]
+        public void ExceptionRasedProcessingPostUri()
```
michaelcfanning, line 314: >ExceptionRasedProcessingPostUri [](http://example.com/codeflow?start=20&length=31)

PostLogFile

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -95,5 +178,42 @@ protected virtual void ProcessBaseline(IAnalysisContext context, T driverOptions
                 };
             }
         }
+
+        protected virtual void ProcessPostUri(IAnalysisContext context, T driverOptions, IFileSystem fileSystem)
+        {
+            ProcessPostUri(driverOptions, fileSystem, new HttpClient());
+        }
+
+        internal static void ProcessPostUri(T driverOptions, IFileSystem fileSystem, HttpClient httpClient)
```
michaelcfanning, line 187: PostLogFile

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -114,6 +114,17 @@ public override int Run(TOptions options)
                     ExecutionException = ex;
                     return FAILURE;
                 }
+
+                try
+                {
+                    ProcessPostUri(_rootContext, options, FileSystem);
```
michaelcfanning, line 120: >ProcessPostUri [](http://example.com/codeflow?start=20&length=14)

PostLogFile

src/Sarif.Driver/Sdk/ExitReason.cs
```diff
@@ -13,6 +13,7 @@ public enum ExitReason
         NoRulesLoaded,
         NoValidAnalysisTargets,
         InvalidCommandLineOption,
-        ExceptionProcessingBaseline
+        ExceptionProcessingBaseline,
+        ExceptionPostingLog
```
michaelcfanning, line 17: >ExceptionPostingLog [](http://example.com/codeflow?start=8&length=19)

ExceptionPostingLogFile

Matches ExceptionCreatingLogFile, et al.

src/Sarif/RuntimeConditions.cs
michaelcfanning: Actually this needs to be ExceptionPostingLogFile, see ExitReason.cs, we should stay in sync with that casing. You can merge after making this change.

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
```diff
@@ -242,28 +242,53 @@ public void ValidatingMultipleFilesOutputOptions_ProducesExpectedResults()
...
             Assert.False(analyzeOptionsBase.ValidateOutputOptions(context));
 
-            //  quiet true, output non-empty
+            // quiet true, output non-empty
             analyzeOptionsBase.Quiet = true;
             analyzeOptionsBase.OutputFilePath = "doodle";
             Assert.True(analyzeOptionsBase.ValidateOutputOptions(context));
+
+            // postUri non-empty and outputFilePath empty = false
+            analyzeOptionsBase.PostUri = "doodle";
```
michaelcfanning, line 269: >doodle [](http://example.com/codeflow?start=42&length=6)

Why 'doodle'? You are repeating someone else's pattern but you should always ask why. 

`doodle` doesn't mean anything here, that was James Winkler's preferred term for meaningless data. In this case, you can actually use something descriptive, like '[https://NotNull.example.com](https://NotNull.example.com)'

For a file use, 'SomeFile.txt' or whatever.

These terms manage to convey something about what's going on.

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
```diff
@@ -242,28 +242,53 @@ public void ValidatingMultipleFilesOutputOptions_ProducesExpectedResults()
...
+            analyzeOptionsBase.OutputFilePath = null;
+            Assert.False(analyzeOptionsBase.ValidateOutputOptions(context));
+
+            // postUri valid non-empty and outputFilePath empty = false
+            analyzeOptionsBase.PostUri = "https://github.com/microsoft/sarif-sdk";
+            analyzeOptionsBase.OutputFilePath = null;
+            Assert.False(analyzeOptionsBase.ValidateOutputOptions(context));
+
+            // postUri non-empty and outputFilePath non-empty = true
```
michaelcfanning, line 278: >// postUri non-empty and outputFilePath non-empty = true [](http://example.com/codeflow?start=12&length=56)

These comments aren't helpful. Why are you repeating the actual code in a comment?

Again, you are repeating a pattern. But you should always consider whether that pattern is worth repeat, don't just trust it because it exists. If there's a weakness in the pattern, then you are spreading it.

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
michaelcfanning, line 269: >doodle [](http://example.com/codeflow?start=42&length=6)

Use 'InvalidUrlText' here. Get it? the contents of the string communicate what's happening in the test.

src/Test.UnitTests.Sarif.Driver/HttpMockHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+            return true;
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            Tuple<HttpRequestMessage, HttpResponseMessage> fakeResponse;
+
+            if (request.Headers.IsEmptyEnumerable())
+            {
+                fakeResponse = _fakeResponses.Find(fr =>
```
michaelcfanning, line 87: >_fakeResponses [](http://example.com/codeflow?start=31&length=14)

our code is a mixture of this. and underscore prefixed patterns./ we need to stop this inconsistency and land on a standard.

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
eddynaka: thanks!

src/Test.UnitTests.Sarif.Driver/HttpMockHelper.cs
eddynaka: let's use this. :)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -119,6 +117,17 @@ public override int Run(TOptions options)
                     ExecutionException = ex;
                     return FAILURE;
                 }
+
+                try
+                {
+                    PostLogFile(_rootContext, options, FileSystem);
```
cfaucon, line 123: What is our in memory option for posting log file for the scenarios were we do not or cannot write to disk.

For example, for scanning in memory to call AnalyzeCommand.AnalyzeTargetHelper() and to dynamically validate we just result.Visit. 

Can we add a AnalyzeCommand helper method to post a SarifLog as in memory json string to the endpoint.

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
```diff
@@ -242,27 +242,43 @@ public void ValidatingMultipleFilesOutputOptions_ProducesExpectedResults()
...
             analyzeOptionsBase.Quiet = true;
-            analyzeOptionsBase.OutputFilePath = "doodle";
+            analyzeOptionsBase.OutputFilePath = "SomeFile.txt";
+            Assert.True(analyzeOptionsBase.ValidateOutputOptions(context));
+
+            analyzeOptionsBase.PostUri = "https://NotNull.example.com";
+            analyzeOptionsBase.OutputFilePath = null;
+            Assert.False(analyzeOptionsBase.ValidateOutputOptions(context));
+
+            analyzeOptionsBase.PostUri = "https://NotNull.example.com";
```
cfaucon, line 268: Should seek to stop the pattern of having functional/unit tests that are testing more that one thing or massive test sets within a single test.  It may feel like a lot of unnecessary boiler plate breaking these tests into unique test but the benefits come in better debuggability via accurate test names and failure are specific or isolated to specific functionality.  Also discrete tests help avoid possible test environment contamination between these subtests were the functionality being tested may not be in the proper state.  For best practice, the current tests are touched, please break them out into sperate tests and use private helper methods if there is any common instantiation, cleanup, etc.

src/Test.Utilities.Sarif/TestRuleBehaviors.cs
```diff
@@ -15,6 +15,7 @@ public enum TestRuleBehaviors
 
         // Various code paths where we want to ensure behavior if an exception is raised.
         RaiseExceptionAccessingId = 0x2,
+
         RaiseExceptionAccessingName = 0x4,
```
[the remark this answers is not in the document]
eddynaka: I use an extension called CodeMaid and depending on the type of the file it adds some space.

I'm reviewing its options.
thanks for reviewing :)

src/Test.UnitTests.Sarif.Driver/DriverExtensionMethodsTests.cs
eddynaka: refactored the code to separate the previous from my current test.
made things a little easier to see and remove the duplication of the checks.

thanks for the suggestion

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -95,5 +178,42 @@ protected virtual void ProcessBaseline(IAnalysisContext context, T driverOptions
                 };
             }
         }
+
+        protected virtual void PostLogFile(IAnalysisContext context, T driverOptions, IFileSystem fileSystem)
```
michaelcfanning, line 182: >IAnalysisContext context [](http://example.com/codeflow?start=43&length=24)

this parameter isn't used.

eddynaka: just refactored to remove that.
thanks :)

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -95,5 +179,61 @@ protected virtual void ProcessBaseline(IAnalysisContext context, T driverOptions
...
         }
+
+        protected virtual void PostLogFile(string postUri, string outputFilePath, IFileSystem fileSystem)
+        {
+            PostLogFile(postUri, outputFilePath, fileSystem, new HttpClient())
+                .GetAwaiter()
+                .GetResult();
+        }
+
+        public static async Task PostLogFile(string postUri, string outputFilePath, IFileSystem fileSystem, HttpClient httpClient)
```
eddynaka, line 190: >public static [](http://example.com/codeflow?start=8&length=13)

changed the method to public, so people who uses the library can call this method.

Now, I'm wonder if we should make this an extension of the SarifLog itself.

```
Task SarifLog.Post(string postUri, string filePath);

Task SarifLog.Post(string postUri, Stream stream)
```

With that, we could keep the PostLogFile internal

@cfaucon @michaelcfanning

src/Sarif/RuntimeConditions.cs
```diff
@@ -49,9 +50,11 @@ public enum RuntimeConditions : uint
         ExceptionInstantiatingSkimmers = 0x8000,
         OutputFileAlreadyExists = 0x10000,
         ExceptionProcessingBaseline = 0x20000,
+        ExceptionPostingLogFile = 0x40000,
 
         // Non-fatal conditions
         UnassignedNonfatal = 0x01F00000,
```
michaelcfanning, line 56: >UnassignedNonfatal [](http://example.com/codeflow?start=8&length=18)

why did you insert a new line?

src/Sarif/Core/SarifLog.cs
```diff
@@ -20,12 +23,12 @@ public partial class SarifLog
         /// <returns>SarifLog instance for file</returns>
         public static SarifLog LoadDeferred(string sarifFilePath)
         {
-            JsonSerializer serializer = new JsonSerializer();
+            var serializer = new JsonSerializer();
             serializer.ContractResolver = new SarifDeferredContractResolver();
 
-            using (JsonPositionedTextReader jptr = new JsonPositionedTextReader(sarifFilePath))
+            using (var jsonPositionedTextReader = new JsonPositionedTextReader(sarifFilePath))
```
michaelcfanning, line 29: >jsonPositionedTextReader [](http://example.com/codeflow?start=23&length=24)

aren't you converting to the single line using convention?

src/Sarif/Core/SarifLog.cs
```diff
@@ -68,6 +71,67 @@ public static SarifLog Load(Stream source, bool deferred = false)
             }
         }
 
+        /// <summary>
+        /// Post the SARIF file to an URI that accepts it in a POST method.
```
michaelcfanning, line 75: >Post the SARIF file to an URI that accepts it in a POST method. [](http://example.com/codeflow?start=12&length=63)

'Post the SARIF log file to a URI.'

src/Sarif/Core/SarifLog.cs
```diff
@@ -68,6 +71,67 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown if filePath does not exist.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if postUri/filePath/stream/httpClient is null.</exception>
+        /// <exception cref="HttpRequestException">Throws an exception if the IsSuccessStatusCode property for the HTTP response is false.</exception>
+        /// <param name="postUri"></param>
+        /// <param name="filePath"></param>
+        /// <param name="fileSystem"></param>
+        /// <param name="httpClient"></param>
+        public static async Task Post(string postUri,
```
michaelcfanning, line 84: >string postUri [](http://example.com/codeflow?start=38&length=14)

why is this a string argument? i.e., why isn't it a URI?

src/Sarif/Core/SarifLog.cs
```diff
@@ -68,6 +71,67 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+
+        /// <summary>
+        /// Post the SARIF stream to an URI that accepts it in a POST method.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">Thrown if postUri/stream/httpClient is null.</exception>
+        /// <exception cref="HttpRequestException">Throws an exception if the IsSuccessStatusCode property for the HTTP response is false.</exception>
+        /// <param name="postUri"></param>
+        /// <param name="stream"></param>
+        /// <param name="httpClient"></param>
+        public static async Task Post(string postUri, Stream stream, HttpClient httpClient)
```
michaelcfanning, line 111: >string postUri [](http://example.com/codeflow?start=38&length=14)

generally we shouldn't pass strings around as URI arguments, we should use the URI type.

src/Sarif/Core/SarifLog.cs
```diff
@@ -86,7 +150,7 @@ public void Save(string sarifFilePath)
         /// <param name="stream">Stream to write SARIF to</param>
         public void Save(Stream stream)
         {
-            using (StreamWriter streamWriter = new StreamWriter(stream))
+            using (var streamWriter = new StreamWriter(stream))
```
michaelcfanning, line 153: >var [](http://example.com/codeflow?start=19&length=3)

aren't you using the single line using convention? if so, feel free to switch to it when you modify code like this. It's a nice practice to scrub an entire file if you make a single change of this kind.

there's a case for keeping things consistent (i.e., don't make a change) if the costs of updating everything are too high.

src/Sarif.Driver/DriverExtensionMethods.cs
```diff
@@ -169,6 +169,10 @@ public static bool ValidateOutputOptions(this AnalyzeOptionsBase options, IAnaly
             // baseline process now depends on output file
             valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.BaselineSarifFile));
 
+            // post SARIF process now depends on output file
```
michaelcfanning, line 172: >post SARIF process now depends on output file [](http://example.com/codeflow?start=15&length=45)

? What does this comment mean? 

can you delete it? and Delete the comment on line 170?

src/Sarif.Driver/DriverExtensionMethods.cs
```diff
@@ -169,6 +169,10 @@ public static bool ValidateOutputOptions(this AnalyzeOptionsBase options, IAnaly
...
+            valid &= !(string.IsNullOrWhiteSpace(options.OutputFilePath) && !string.IsNullOrWhiteSpace(options.PostUri));
+            valid &= !(!string.IsNullOrWhiteSpace(options.PostUri) && !Uri.IsWellFormedUriString(options.PostUri, UriKind.Absolute));
```
michaelcfanning, line 174: >string [](http://example.com/codeflow?start=24&length=6)

Please rewrite this logic to make it clearer.

Boolean patterns like '(!' are a clue this code isn't obvious.

Just use readable if {} conditions. This is a case where being clever and putting everything on a single line is working against maintainability.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -109,5 +109,10 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
             "baseline",
             HelpText = "A SARIF file to be used as baseline.")]
         public string BaselineSarifFile { get; set; }
+
+        [Option(
+            "postUri",
+            HelpText = "An URI that accepts SARIF as content in a POST method.")]
```
michaelcfanning, line 115: >An URI that accepts SARIF as content in a POST method [](http://example.com/codeflow?start=24&length=53)

'A URI to which the SARIF log file will be posted.'

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -200,6 +204,152 @@
             newSarifLog.Runs[0].Results.Count.Should().Be(sarifLog.Runs[0].Results.Count);
         }
 
+        [Fact]
+        public async Task SarifLog_PostStream_WithInvalidParameters_ShouldThrowArgumentNullException()
```
michaelcfanning, line 208: >SarifLog_PostStream_WithInvalidParameters_ShouldThrowArgumentNullException [](http://example.com/codeflow?start=26&length=74)

I don't understand this test pattern. 

Why aren't we simply using the helpers that verify an assertion is thrown?

All this test goo, creating titles, etc, isn't helpful for this simple scenario, it's a lot of complexity for not a lot of value.

Basically, all these tests could be authored by updating the code below to simply to pass parameters.

                Exception exception = await Record.ExceptionAsync(async () =>
                {
                    await SarifLog.Post(testCase.PostUri,
                                        testCase.Stream,
                                        testCase.HttpClient);
                });

Have you taken a look at the actual test output produced by this pattern? It's not great. :) That's another reason why we prefer not to use it unless necessary.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -200,6 +204,152 @@
...
+            });
+
+            exception.Should().BeOfType(typeof(ArgumentNullException));
+
+            filePath = "SomeFile.txt";
+            fileSystem
+                .Setup(f => f.FileExists(It.IsAny<string>()))
+                .Returns(false);
+
+            exception = await Record.ExceptionAsync(async () =>
```
michaelcfanning, line 277: >ExceptionAsync [](http://example.com/codeflow?start=37&length=14)

Can't we simplify this using Assert.ThrowsAsync? Or am I missing something?

 await Assert.ThrowsAsync<ArgumentNullException>(() => MethodThatThrows());

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using Moq;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Driver.Sdk
+{
+    public class PluginDriverCommandTests
+    {
+        [Fact]
+        public void PluginDriverCommand_ValidateInvocationPropertiesToLog_ShouldValidateParameters()
```
michaelcfanning, line 25: >PluginDriverCommand_ValidateInvocationPropertiesToLog_ShouldValidateParameters [](http://example.com/codeflow?start=20&length=78)

Here's another bad use of the test cases pattern. 

Please don't do this for simple test scenarios.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
eddynaka: Assert.ThrowsAsync is from MSTest. Record.Exception/ExceptionAsync is from xunit.

if we are going to use xunit everywhere, then, we should keep using record.exception.

if we dont care about using two different test patterns, then, i can move to assert.throws

:)

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
eddynaka: refactored.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
eddynaka: replaced to Assert.Throw style.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
eddynaka: just replaced to asser.throw style since we are normally using it.

src/Sarif.Driver/DriverExtensionMethods.cs
eddynaka: sure! :)

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            mockContext.SetupGet(context => context.Logger).Returns(mockLogger.Object);
+            bool current = PluginDriverCommand<string>.ValidateInvocationPropertiesToLog(mockContext.Object,
+                                                                                         null);
+            current.Should().BeTrue();
+
+            current = PluginDriverCommand<string>.ValidateInvocationPropertiesToLog(mockContext.Object,
+                                                                                    Array.Empty<string>());
+            current.Should().BeTrue();
+
+            current = PluginDriverCommand<string>.ValidateInvocationPropertiesToLog(mockContext.Object,
```
michaelcfanning, line 40: >ValidateInvocationPropertiesToLog [](http://example.com/codeflow?start=50&length=33)

I don't understand these tests. Why are you bundling all four in one unit test? There's no explanation here of why one test fails and another one passes.

michaelcfanning (approved):
:shipit:

PR #2404, opened 2021-10-28
Level Nullable Changes

shaopeng-gh:
# Description:

There were 3 issues reported related to Result.Level reported. (#2077, #2024, #2383)
My understanding of the issue is as following
- When the result level is explicitly set as warning, because we have the property default value set as warning and not written the Level json node if it is default. We will not be able to differentiate the case with user did not explicitly set the Level.
- this is also wrong, when the default configuration is e.g. Note, with the Level json node not written, it will be considered as Note instead of the warning user explicitly set.

# Fixes:

This PR propose the below fix:
- change Result Level to be nullable and always write Level Json node if user has set it. When user not set, don't write the node.
- this will be able to differentiate if user has set the Result.Level 
- this will resolve the issue above when user explicitly set to Warning but the ending Sarif file is considered with level Note 
- also fix the existing issue that ShouldLog() does not consider default configuration

Code coverage Main branch:
image

Code coverage PR:
image

image
image
image


src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1001.ValidateWithBaseline.sarif
```diff
@@ -222,6 +222,7 @@
         {
           "ruleId": "SARIF2005",
           "ruleIndex": 2,
+          "level": "warning",
```
eddynaka, line 225: >level": "warning", [](http://example.com/codeflow?start=11&length=18)

with this change, many rules will start emitting this, since we are setting the level manually to warning based on the configuration.

NuGet.Config
```diff
@@ -1,11 +1,12 @@
...
-    <clear />
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-  </packageSources>
-  <config>
-    <add key="globalPackagesFolder" value="./packages" />
-    <add key="repositoryPath" value="./packages" />
-  </config>
+	<packageSources>
+		<clear />
+		<add key="localpackages" value=".\refs\localpackages" />
```
shaopeng-gh, line 5: ><add key="localpackages" value=".\refs\localpackages" /> [](http://example.com/codeflow?start=2&length=56)

this is a temporary change,
the reason is Json Schema reference Sarif SDK and Sarif SDK reference Json Schema, without update both at the same time, the build and test will not be able to pass.
this local nuget store contains local build of the Json Schema.
after this we can publish a new version, 
then update the Json Schema to use the new version,
then finally update the Sarif SDK to use the new vesion of the Json Schema nuget, and this local nuget store can be removed at the same changeset.

shaopeng-gh: the Json Schema PR will be about the same as this one but it will be after the Sarif SDK publish a new version, to use the latest Sarif SDK nuget. 
[https://github.com/microsoft/jschema/pull/139](https://github.com/microsoft/jschema/pull/139)

yongyan-gh: just wondering about the local package thing.
why there is a reference loop between Sarif SDK and Json Schema? can we break the loop to avoid doing this temporary thing?

src/Sarif/NullableEnum.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// A wrapper class for Nullable Enum
+    /// </summary>
+    public static class NullableEnum
```
shaopeng-gh, line 12: >NullableEnum [](http://example.com/codeflow?start=24&length=12)

This is a helper class I added for simplify code related to Nullable Enum, we can add more methods if helpful

src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
```diff
@@ -2053,7 +2053,6 @@
 
         "level": {
           "description": "A value specifying the severity level of the result.",
-          "default": "warning",
```
shaopeng-gh, line 2056: >"default": "warning", [](http://example.com/codeflow?start=10&length=21)

remove the default, it does not confront to the spec as well.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1001.SyntaxError.sarif
```diff
@@ -15,7 +15,7 @@
               },
               "messageStrings": {
                 "default": {
-                  "text": "at \"{0}\": JSON syntax error: {1}"
```
shaopeng-gh, line 18: >"text": "at \"{0}\": JSON syntax error: {1}" [](http://example.com/codeflow?start=18&length=44)

all the .Sarif changes are refresh baseline.
this one is not my change, 
I think it is just was not refreshed for a while.

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1002.ValidateBaseline.NoResults.sarif
```diff
@@ -222,6 +222,7 @@
         {
           "ruleId": "SARIF2005",
           "ruleIndex": 2,
+          "level": "warning",
```
shaopeng-gh, line 225: >"level": "warning", [](http://example.com/codeflow?start=10&length=19)

the current behavior is, 
if explicitly set in the code 
".Level=FailureLevel.Warning" 
then this line will be written to Sarif. 
If not explicitly set, this line will not be written.

NuGet.Config
shaopeng-gh: Sarif SDK solution use Json Schema to validate if a json is a valid Sarif file,
in Json Schema the method to validate json use Sarif SDK to maintain a Sarif format result in memory and later to output.
the property we are changing "level" is a property used explicitly in both place.

eddynaka:
> <?xml version="1.0" encoding="utf-8"?>

we should merge both  configs in one.

---
Refers to: src/NuGet.Config:1 in 6988ce2. [](commit_id = 6988ce2ecaebcd9a7a26d2dd45fbef45b0447319, deletion_comment = False)

src/Sarif/NullableEnum.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// The Nullable Enum array matching the value array.
+        /// </returns>
+        public static IEnumerable<T> Parse<T>(string[] values)
+        {
+            if (values == null)
+            {
+                return null;
+            }
+
+            var list = new List<T>();
```
eddynaka, line 47: >() [](http://example.com/codeflow?start=34&length=2)

nit:
new List<T>(values.Length);

src/Sarif/Visitors/SarifTransformerUtilities.cs
```diff
@@ -157,16 +174,19 @@ public static ResultLevelVersionOne CreateResultLevelVersionOne(FailureLevel v2D
             {
                 case FailureLevel.Error:
                     return ResultLevelVersionOne.Error;
+
                 case FailureLevel.Note:
                     return ResultLevelVersionOne.Pass;
+
                 case FailureLevel.Warning:
                     return ResultLevelVersionOne.Warning;
+
```
eddynaka, line 183: can you revert all this empty lines?

src/Sarif/NullableEnum.cs
shaopeng-gh: fixed

src/Sarif/Visitors/SarifTransformerUtilities.cs
shaopeng-gh: fixed

shaopeng-gh:
> <?xml version="1.0" encoding="utf-8"?>

I just tried remove the one in sub folder but it will auto generate after build.
we can fix it in a separate changeset.

---
In reply to: [965166292](https://github.com/microsoft/sarif-sdk/pull/2404#issuecomment-965166292) [](http://example.com/codeflow?ancestors=965166292,965166292)

---
Refers to: src/NuGet.Config:1 in 6988ce2. [](commit_id = 6988ce2ecaebcd9a7a26d2dd45fbef45b0447319, deletion_comment = False)

michaelcfanning:
> <?xml version="1.0" encoding="utf-8"?>

Please file a bug on this issue so we don't lose it.

---
In reply to: [965330249](https://github.com/microsoft/sarif-sdk/pull/2404#issuecomment-965330249) [](http://example.com/codeflow?ancestors=965330249,965166292)

---
Refers to: src/NuGet.Config:1 in 6988ce2. [](commit_id = 6988ce2ecaebcd9a7a26d2dd45fbef45b0447319, deletion_comment = False)

shaopeng-gh:
> <?xml version="1.0" encoding="utf-8"?>

sounds good just created a bug and assigned to myself.

---
In reply to: [966725775](https://github.com/microsoft/sarif-sdk/pull/2404#issuecomment-966725775) [](http://example.com/codeflow?ancestors=966725775,965166292)

---
Refers to: src/NuGet.Config:1 in 6988ce2. [](commit_id = 6988ce2ecaebcd9a7a26d2dd45fbef45b0447319, deletion_comment = False)

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1002.ValidateBaseline.NoResults.sarif
shaopeng-gh: got some input from Michael, let me set this to draft and work on it first

src/Test.UnitTests.Sarif/Core/ResultTests.cs
```diff
@@ -43,6 +48,95 @@ public void Result_Level_ResetsKindValue()
             result.Kind.Should().Be(ResultKind.Fail);
         }
 
+        [Fact]
+        public void Result_Level_WriteCorrectValue_GeneralCases()
```
shaopeng-gh, line 52: >Result_Level_WriteCorrectValue_GeneralCases [](http://example.com/codeflow?start=20&length=43)

The same new tests are also added in a separated branch based on a main to test how it looks, not all tests can pass,
link:
[https://github.com/microsoft/sarif-sdk/pull/2409](https://github.com/microsoft/sarif-sdk/pull/2409)

src/Sarif/NotYetAutoGenerated/Result.cs
```diff
@@ -83,22 +84,47 @@ public ResultKind Kind
...
+
+                    if (currentDefaultConfiguration == null)
+                    {
+                        // if defaultConfiguration is absent, it SHALL be taken to be present and default Warning
+                        return FailureLevel.Warning;
+                    }
+                    else if (!currentDefaultConfiguration.Enabled)
+                    {
+                        // if defaultConfiguration is present, but disabled
+                        return _level;
```
shaopeng-gh, line 115: this is my understanding of the spec, let me know if we want to have this condition.

yongyan-gh: According to spec
"IF theDescriptor.defaultConfiguration.level (§3.49.14, §, §3.50.3) is present THEN
          SET level to theDescriptor.defaultConfiguration.level.
IF level has not yet been set THEN
    SET level to "warning"."

i think it may be similar to:
if (currentDefaultConfiguration != null && currentDefaultConfiguration.Enable)
{
    _level = currentDefaultConfiguration.Level;
}
return _level ?? FailureLevel.Warning;

shaopeng-gh: Thanks, you are correct. I have moved this code to extension method.

src/Sarif/Writers/BaseLogger.cs
```diff
@@ -53,13 +53,14 @@ public bool ShouldLog(Notification notification)
             return _failureLevels.Contains(notification.Level);
         }
 
-        public bool ShouldLog(Result result)
+        public bool ShouldLog(Result result, ReportingDescriptor rule)
```
shaopeng-gh, line 56: >ShouldLog [](http://example.com/codeflow?start=20&length=9)

this is another existing issue, 
Shouldlog() does not consider the default setting.

src/Sarif/ExtensionMethods.cs
```diff
@@ -401,6 +404,39 @@ public static string GetMessageText(this Result result, ReportingDescriptor rule
             return text;
         }
 
+        public static FailureLevel? GetEffectiveLevel(this Result result, ReportingDescriptor rule)
```
eddynaka, line 407: >public static [](http://example.com/codeflow?start=8&length=13)

do we need to make this public? This will probably confuse users.

src/Sarif/ExtensionMethods.cs
```diff
@@ -401,6 +404,39 @@ public static string GetMessageText(this Result result, ReportingDescriptor rule
...
+
+        public static FailureLevel? GetEffectiveLevel(this Result result, ReportingConfiguration defaultConfiguration)
+        {
+            FailureLevel? rawLevel = result.GetRawLevel();
+
+            if (rawLevel.HasValue)
+            {
+                return rawLevel;
+            }
+            else
```
eddynaka, line 420: >else [](http://example.com/codeflow?start=12&length=4)

you can remove the else

src/Sarif/NotYetAutoGenerated/Result.cs
```diff
@@ -83,28 +84,49 @@ public ResultKind Kind
...
+        public FailureLevel? Level
         {
-            get { return _level; }
+            get
+            {
+                if (_level.HasValue)
+                {
+                    return _level;
+                }
+                else
```
eddynaka, line 103: >else [](http://example.com/codeflow?start=16&length=4)

you can remove the else since we are returning on line 101.

src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -56,7 +56,7 @@ protected void LogResult(string jPointer, string formatId, params string[] args)
 
             Context.Logger.Log(this,
                 RuleUtilities.BuildResult(
-                    DefaultConfiguration.Level,
+                    null,
```
eddynaka, line 59: >null [](http://example.com/codeflow?start=20&length=4)

add the name of the variable first:
```variable: null```

with that, we will know what is the variable name that we are passing as `null`

File: src/Test.Utilities.Sarif/TestMessageLogger.cs [hunk 16]
```diff
@@ -24,6 +25,8 @@ public TestMessageLogger()
 
         public HashSet<string> NotApplicableTargets { get; set; }
 
+        public HashSet<string> NullTargets { get; set; }
```
eddynaka, line 28: >NullTargets [](http://example.com/codeflow?start=31&length=11)

where are you using this?

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1001.ValidateWithBaseline.sarif
shaopeng-gh: fixed, this is related to the existing code issue of the ShouldLog() did not consider the default configuration. 
The Reason the existing tests pass is, since it only consider the level at the result node,
when the test codes was having e.g. "note" as the default configuration, the current tests also pass the "note" to explicitly as the level at the result node the same.
Now I have fix the ShouldLog() to also consider the default configuration, I just pass null to the level at the result node.
So now you will see those "level": "note" got deleted instead, in these test files.

src/Sarif/ExtensionMethods.cs
shaopeng-gh: removed

src/Sarif/ExtensionMethods.cs
shaopeng-gh: agree, changed to internal

src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
shaopeng-gh: added.

src/Test.Utilities.Sarif/TestMessageLogger.cs
shaopeng-gh: no longer needed, removed

shaopeng-gh:
Code coverage Main branch:
image

Code coverage PR:
image

src/Sarif/NotYetAutoGenerated/Result.cs
```diff
@@ -83,26 +92,41 @@ public ResultKind Kind
...
-            set
+            get
             {
-                _level = value;
-                if (_level != FailureLevel.None)
+                if (_level.HasValue)
                 {
-                    _kind = ResultKind.Fail;
+                    return _level;
                 }
```
shaopeng-gh, line 110: this is moved to the get() of Kind, and it need to consider default configuration.

related test:
ResultTests.cs: 
Result_Level_ResetsKindValue_WithDefaultConfiguration()

src/Sarif/Autogenerated/Result.cs
```diff
@@ -394,7 +417,7 @@ public Result(Result other)
                 throw new ArgumentNullException(nameof(other));
             }
 
-            Init(other.RuleId, other.RuleIndex, other.Rule, other.Kind, other.Level, other.Message, other.AnalysisTarget, other.Locations, other.Guid, other.CorrelationGuid, other.OccurrenceCount, other.PartialFingerprints, other.Fingerprints, other.Stacks, other.CodeFlows, other.Graphs, other.GraphTraversals, other.RelatedLocations, other.Suppressions, other.BaselineState, other.Rank, other.Attachments, other.HostedViewerUri, other.WorkItemUris, other.Provenance, other.Fixes, other.Taxa, other.WebRequest, other.WebResponse, other.Properties);
+            Init(other.RuleId, other.RuleIndex, other.Rule, other.Kind, other._level, other.Message, other.AnalysisTarget, other.Locations, other.Guid, other.CorrelationGuid, other.OccurrenceCount, other.PartialFingerprints, other.Fingerprints, other.Stacks, other.CodeFlows, other.Graphs, other.GraphTraversals, other.RelatedLocations, other.Suppressions, other.BaselineState, other.Rank, other.Attachments, other.HostedViewerUri, other.WorkItemUris, other.Provenance, other.Fixes, other.Taxa, other.WebRequest, other.WebResponse, other.Properties);
```
shaopeng-gh, line 420: >_level [](http://example.com/codeflow?start=78&length=6)

this function Result(Result other)
is a deep copy of a Result object , should use the raw value.

src/Sarif/NotYetAutoGenerated/Result.cs
```diff
@@ -394,7 +417,7 @@ public Result(Result other)
                 throw new ArgumentNullException(nameof(other));
             }
 
-            Init(other.RuleId, other.RuleIndex, other.Rule, other.Kind, other.Level, other.Message, other.AnalysisTarget, other.Locations, other.Guid, other.CorrelationGuid, other.OccurrenceCount, other.PartialFingerprints, other.Fingerprints, other.Stacks, other.CodeFlows, other.Graphs, other.GraphTraversals, other.RelatedLocations, other.Suppressions, other.BaselineState, other.Rank, other.Attachments, other.HostedViewerUri, other.WorkItemUris, other.Provenance, other.Fixes, other.Taxa, other.WebRequest, other.WebResponse, other.Properties);
+            Init(other.RuleId, other.RuleIndex, other.Rule, other.Kind, other._level, other.Message, other.AnalysisTarget, other.Locations, other.Guid, other.CorrelationGuid, other.OccurrenceCount, other.PartialFingerprints, other.Fingerprints, other.Stacks, other.CodeFlows, other.Graphs, other.GraphTraversals, other.RelatedLocations, other.Suppressions, other.BaselineState, other.Rank, other.Attachments, other.HostedViewerUri, other.WorkItemUris, other.Provenance, other.Fixes, other.Taxa, other.WebRequest, other.WebResponse, other.Properties);
```
shaopeng-gh, line 420: >_level [](http://example.com/codeflow?start=78&length=6)

this function Result(Result other)
is a deep copy of a Result object , should use the raw value.

src/Sarif/NullableEnum.cs
shaopeng-gh: this is no longer needed, deleted.

shaopeng-gh:
Got some more feedbacks, after fix I will create a new PR instead, so the comments will be cleaner for review, thanks!

PR #2405, opened 2021-11-02
Response to comments

gfs:
Some formatting.
Removes seek in ctor.
Changes JsonPositionedTextReader(Stream) to be a factory method so we can enforce that the stream provided is seekable.
Adds some unit tests for DelegatingStream.
Moves DelegatingStream next to JsonPositionedTextReader since it is a helper class.
Adds test cases for deferred collections reading from the new stream API.
Adds a test file for jsonpositionedtextreader along with a test to confirm it throws when given a non seekable stream to the factory function.

### DIFF
--- a/src/Sarif/Readers/DelegatingStream.cs
+++ b/src/Sarif/Readers/DelegatingStream.cs
@@ -11,17 +11,22 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         /// <summary>
         /// This is a wrapper for a stream that protects the underlying stream from being disposed.
+        /// Disposing this object is a no-op. To dispose the underlying stream use the DisposeUnderlyingStream method.
         /// </summary>
         /// <param name="underlyingStream"></param>
         internal DelegatingStream(Stream underlyingStream)
         {
             stream = underlyingStream;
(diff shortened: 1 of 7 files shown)


src/Sarif/Readers/DelegatingStream.cs
```diff
@@ -30,26 +35,20 @@ public override long Position
...
+        public override int Read(byte[] buffer, int offset, int count) => stream.Read(buffer, offset, count);
+
+        public override long Seek(long offset, SeekOrigin origin) => stream.Seek(offset, origin);
+
+        public override void SetLength(long value) => stream.SetLength(value);
+
+        public override void Write(byte[] buffer, int offset, int count) => stream.Write(buffer, offset, count);
+
+        /// <summary>
+        /// Calls the dispose method on the wrapped stream.
```
michaelcfanning, line 50: >Calls the dispose method on the wrapped stream. [](http://example.com/codeflow?start=12&length=47)

? This comment provides no value over reading the code or reading the name 'DisposeUnderlyingStream. Please think of something additionally useful to say or delete the comment.

You will hear this from me persistently: don't author comments that are a simple recapitulation of what's already obvious in code. A comment should provide additional value.

PR #2406, opened 2021-11-02
Fixing Load when deferred is enabled

### LINKED ISSUE #2397: SarifLog.Load(Stream, deferred = true) does not work

eddynaka:
Fixes #2397

## Description
This change is fixing the `SarifLog.Load(stream, deferred)` implementation. To load a SARIF in deferred mode, we **must** use `JsonPositionedTextReader` to read the `stream`.
Once we fixed that, we also observed that the stream was getting closed before we read the SARIF, so, we added some protections such as `Stream.Synchronized` and `DelegatingStream`.

This will prevent the stream from being closed before needed.

`Stream.Synchronized`: https://docs.microsoft.com/dotnet/api/system.io.stream.synchronized?view=net-5.0


src/Sarif/Readers/DelegatingStream.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal class DelegatingStream : Stream
+    {
+        private readonly Stream stream;
+
+        /// <summary>
+        /// This is a wrapper for a stream that protects the underlying stream from being disposed.
+        /// Disposing this object is a no-op. To dispose the underlying stream use the DisposeUnderlyingStream method.
```
michaelcfanning, line 14: >DisposeUnderlyingStream [](http://example.com/codeflow?start=87&length=23)

in a 'real' doc comment, this would be a reference to the method. but we don't need that, as we're not exhaustively documenting this type.

src/Test.UnitTests.Sarif/Readers/DelegatingStreamTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using FluentAssertions;
+
+using Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Readers;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class DelegatingStreamTests
```
michaelcfanning, line 16: >DelegatingStreamTests [](http://example.com/codeflow?start=17&length=21)

Do you tests actually try to dispose the stream and ensure that it hasn't been disposed? That's the primary purpose of this wrapper, yes? So where's the test for that basic capability?

michaelcfanning (changes requested):
:clock1:

src/Test.UnitTests.Sarif/Readers/NonSeekableStream.cs
```diff
@@ -0,0 +1,60 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Readers
+{
+    internal class NonSeekableStream : Stream
```
michaelcfanning, line 11: >NonSeekableStream [](http://example.com/codeflow?start=19&length=17)

Please delete this class, which seems strange. I'm not sure that it covers all cases properly. For example, most non-seekable streams don't allow resetting the length, which you do here. 

Please go find an actual non-seekable stream in .NET and use it. 

Try System.IO.Packaging, I think their compressed streams aren't seekable.

[https://docs.microsoft.com/en-us/dotnet/api/system.io.packaging.storageinfo.createstream?view=windowsdesktop-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.io.packaging.storageinfo.createstream?view=windowsdesktop-5.0)

src/Sarif/Readers/DelegatingStream.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    internal class DelegatingStream : Stream
```
michaelcfanning, line 8: >DelegatingStream [](http://example.com/codeflow?start=19&length=16)

Rename this type NonDisposingDelegatingStream.

michaelcfanning, line 8: >DelegatingStream [](http://example.com/codeflow?start=19&length=16)

Please go author this test, instantiate this type, dispose it and dispose it again.

Does this wrapper then raise an exception? i.e., are we expecting that multiple dispose methods are a no-op on the underlying stream? is it important for this class to allow one user to dispose it and then never again (despite never touching the wrapped object)?

src/Sarif/Readers/DelegatingStream.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        private readonly Stream stream;
```
michaelcfanning, line 10: >stream [](http://example.com/codeflow?start=32&length=6)

name this underlyingStream.

src/Sarif/Readers/DelegatingStream.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// This is a wrapper for a stream that protects the underlying stream from being disposed.
+        /// Disposing this object is a no-op. To dispose the underlying stream use the DisposeUnderlyingStream method.
+        /// </summary>
+        /// <param name="underlyingStream"></param>
+        internal DelegatingStream(Stream underlyingStream)
```
michaelcfanning, line 17: >underlyingStream [](http://example.com/codeflow?start=41&length=16)

what are we supposed to do with a null argument?

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -170,6 +172,19 @@ public void SarifLog_SplitPerTarget_WithEmptyLocations()
             }
         }
 
+        [Fact]
+        public void SarifLog_LoadDeferred()
+        {
+            var random = new Random();
```
michaelcfanning, line 178: >Random [](http://example.com/codeflow?start=29&length=6)

Don't ever introduce randomness into a method without capturing the seed. Otherwise we can't reproduce test failures.

Here's what to do:

Allocate a 4 byte array
Using RandomNumberGenerator.Fill to fill it
Convert this byte array to a number (using the Convert class)
Persist this int32 value to test output
Initialize your Random instance with that seed

src/Test.UnitTests.Sarif/Readers/DeferredCollectionsTests.cs
```diff
@@ -126,6 +156,34 @@ private static void CompareReadNormalToReadDeferred(string filePath)
             Assert.Equal(contextsCopy.Count, valueCount);
         }
 
+        private static void CompareReadNormalToReadDeferred(string filePath)
+        {
+            LogModelSampleBuilder.EnsureSamplesBuilt();
+            JsonSerializer serializer = new JsonSerializer();
```
michaelcfanning, line 162: >JsonSerializer [](http://example.com/codeflow?start=12&length=14)

use 'var' here and elsewhere in the method

src/Sarif/Readers/DelegatingStream.cs
eddynaka: renamed

src/Sarif/Readers/DelegatingStream.cs
eddynaka: renamed

src/Test.UnitTests.Sarif/Readers/DeferredCollectionsTests.cs
eddynaka: done

src/Test.UnitTests.Sarif/Readers/DelegatingStreamTests.cs
eddynaka: just added.

src/Test.UnitTests.Sarif/Readers/NonSeekableStream.cs
eddynaka: that package is only for `Windows Desktop 5`.

the length isn't settable. Is is only a get.

Will search for a non-seekable stream.

src/Sarif/Readers/DelegatingStream.cs
eddynaka: changed the code to generate a ArgumentNullException

src/Sarif/Readers/DelegatingStream.cs
eddynaka: Just added a test with this behavior.
It will only thrown an exception if you call the `DisposeUnderlyingStream` and try to use the stream.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
eddynaka: done

src/Test.UnitTests.Sarif/Readers/NonSeekableStream.cs
eddynaka: Replaced  with a cryptoStream and removed the nonseekable class

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -170,6 +181,25 @@ public void SarifLog_SplitPerTarget_WithEmptyLocations()
...
 
+        [Fact]
+        public void SarifLog_LoadDeferred()
+        {
+            byte[] data = new byte[4];
+            RandomNumberGenerator.Fill(data);
+            int seed = BitConverter.ToInt32(data);
+            var random = new Random(seed);
+
+            output.WriteLine($"The seed used was: '{seed}'");
```
michaelcfanning, line 192: >output [](http://example.com/codeflow?start=12&length=6)

you need the 'this.' prefix here.

michaelcfanning, line 192: >The seed used was [](http://example.com/codeflow?start=31&length=17)

'The seed pass to the Random instance was : {seed}'

src/Test.UnitTests.Sarif/Readers/DeferredCollectionsTests.cs
```diff
@@ -126,6 +156,34 @@ private static void CompareReadNormalToReadDeferred(string filePath)
...
         }
 
+        private static void CompareReadNormalToReadDeferred(string filePath)
+        {
+            LogModelSampleBuilder.EnsureSamplesBuilt();
+            var serializer = new JsonSerializer();
+
+            Log expected;
+            Log actual;
+
```
michaelcfanning, line 166: In future, waterfall from shortest line to longest, that's our convention.

Log actual;
Log expected;

src/Test.UnitTests.Sarif/Readers/DeferredCollectionsTests.cs
```diff
@@ -126,6 +156,34 @@ private static void CompareReadNormalToReadDeferred(string filePath)
...
+            // Read normally (JsonSerializer -> JsonTextReader -> StreamReader)
+            using (var reader = new JsonTextReader(new StreamReader(filePath)))
```
michaelcfanning, line 168: why do you use the new using declaration syntax elsewhere but not here?

src/Test.UnitTests.Sarif/Readers/NonDisposingDelegatingStreamTests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            var delegatingStream = new NonDisposingDelegatingStream(stream);
+            delegatingStream.CanSeek.Should().BeFalse();
+
+            Exception exception = Record.Exception(() => new StreamReader(delegatingStream));
+            exception.Should().BeOfType(typeof(ArgumentException));
+        }
+
+        [Fact]
+        public void NonDisposingDelegatingStream_VerifyingDipose()
```
michaelcfanning, line 83: >NonDisposingDelegatingStream_VerifyingDipose [](http://example.com/codeflow?start=20&length=44)

Wahoo!

src/Test.UnitTests.Sarif/Readers/NonDisposingDelegatingStreamTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                    0x09, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16
+                };
+
+                byte[] iv = aes.IV;
+                memoryStream.Write(iv, 0, iv.Length);
+
+                var cryptoStream = new CryptoStream(
+                    memoryStream,
+                    aes.CreateEncryptor(),
+                    CryptoStreamMode.Write);
```
michaelcfanning, line 117: add a line of whitespace here.

src/Test.UnitTests.Sarif/Readers/NonDisposingDelegatingStreamTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                using (var encryptWriter = new StreamWriter(cryptoStream))
```
michaelcfanning, line 118: >using [](http://example.com/codeflow?start=16&length=5)

more inconsistency in using legacy syntax. pick one! :)

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
eddynaka: added.

src/Test.UnitTests.Sarif/Readers/DeferredCollectionsTests.cs
eddynaka: changed and renamed the variables to be able to do that.

src/Sarif/Readers/NonDisposingDelegatingStream.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin) => underlyingStream.Seek(offset, origin);
+
+        public override void SetLength(long value) => underlyingStream.SetLength(value);
+
+        public override void Write(byte[] buffer, int offset, int count) => underlyingStream.Write(buffer, offset, count);
+
+        public void DisposeUnderlyingStream()
+        {
+            underlyingStream.Dispose();
```
michaelcfanning, line 49: >underlyingStream [](http://example.com/codeflow?start=12&length=16)

missing .this.

PR #2407, opened 2021-11-04
Enable automation details for command base

eddynaka:
## Description

Imagine SARIF Pattern Matcher being run by someone, and we want to automatically ingest the output SARIF somehow.

With this, we could ask the user to run SARIF Pattern Matcher and pass `--automationId` and `--automationGuid`, that information would be stored in the SARIF and we would know that the SARIF was generated by a specific team.

## Follow-up change

Change the `commandBase` to accept a URI and this URI will be used to send the generated SARIF to that endpoint.

cc: @cfaucon

### DIFF
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -112,12 +112,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         [Option(
             "automationId",
-            HelpText = "An id that will be used in the output SARIF and show in the 'automationDetails.id' under 'run'.")]
+            HelpText = "An id that will be persisted to the 'Run.AutomationDetails.Id' property. See section '3.17.3' of the SARIF specification for more information.")]
         public string AutomationId { get; set; }
 
         [Option(
             "automationGuid",
-            HelpText = "A guid that will be used in the output SARIF and show in the 'automationDetails.guid' under 'run'.")]
+            HelpText = "A guid that will be persisted to the 'Run.AutomationDetails.Guid' property. See section '3.17.4' of the SARIF specification for more information.")]
         public string AutomationGuid { get; set; }
     }
 }
--- a/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
@@ -107,7 +107,17 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         [Option(
             "baseline",
-            HelpText = "A Sarif file to be used as baseline")]
+            HelpText = "A SARIF file to be used as baseline.")]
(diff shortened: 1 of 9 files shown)


src/Sarif/Core/Run.cs
```diff
@@ -220,6 +220,14 @@ public bool ShouldSerializeGraphs()
             return this.Graphs.HasAtLeastOneNonDefaultValue(Graph.ValueComparer);
         }
 
+        public bool ShouldSerializeAutomationDetails()
+        {
+            return this.AutomationDetails?.Description != null ||
```
michaelcfanning, line 225: >AutomationDetails?.Description [](http://example.com/codeflow?start=24&length=30)

Why should we serialize an empty description? suggest using is null or whitespace...

src/Sarif/Core/RunAutomationDetails.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Information that describes a run's identity and role within an engineering system process.
+    /// </summary>
+    public partial class RunAutomationDetails
+    {
+        public bool ShouldSerializeDescription() => this.Description != null;
```
michaelcfanning, line 11: >Description [](http://example.com/codeflow?start=57&length=11)

?? why persist an empty description? how is this useful?

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -107,7 +107,17 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
 
         [Option(
             "baseline",
-            HelpText = "A Sarif file to be used as baseline")]
+            HelpText = "A SARIF file to be used as baseline.")]
         public string BaselineSarifFile { get; set; }
+
+        [Option(
+            "automationId",
```
michaelcfanning, line 114: >automationId [](http://example.com/codeflow?start=13&length=12)

why aren't these in common options base? won't we need them for other commands? i guess we can move them later if so.

michaelcfanning (approved):
:shipit:

src/Sarif/Core/Run.cs
eddynaka: The Description isn't a string, it is a Message type. It's why I didn't change to NullOrWhiteSpace check

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
eddynaka: just moved! :)

src/Sarif/Core/RunAutomationDetails.cs
eddynaka: The Description isn't a string, it is a Message type. It's why I didn't change to NullOrWhiteSpace check

PR #2409, opened 2021-11-15
Add tests for Result.Level in current Main

shaopeng-gh:


src/Test.UnitTests.Sarif/Core/ResultTests.cs
```diff
@@ -43,6 +48,95 @@ public void Result_Level_ResetsKindValue()
...
+            Assert.True(WriteSarifThenReadLevelNode(result, defaultConfiguration) == null);
+
+            result = new Result() { };
+            defaultConfiguration = new ReportingConfiguration() { Enabled = true, Level = FailureLevel.Error };
+            overrideConfiguration = new ReportingConfiguration() { Enabled = true, Level = FailureLevel.Note };
+            Assert.True(WriteSarifThenReadLevelNode(result, defaultConfiguration, overrideConfiguration) == null);
+        }
+
+        //[Fact]
+        //public void Result_Level_WriteCorrectValue_Null()
```
shaopeng-gh, line 121: >Null [](http://example.com/codeflow?start=53&length=4)

this test is commented out on purpose, only can be un-commented out with the level change PR, I un-commented out there.

src/Test.UnitTests.Sarif/Core/ResultTests.cs
```diff
@@ -43,6 +48,95 @@ public void Result_Level_ResetsKindValue()
...
+
+        [Fact]
+        public void Result_Level_WriteCorrectValue_Warning()
+        {
+            Result result;
+            ReportingConfiguration defaultConfiguration;
+            ReportingConfiguration overrideConfiguration;
+
+            result = new Result() { Level = FailureLevel.Warning };
+            Assert.True(WriteSarifThenReadLevelNode(result) == FailureLevel.Warning);
```
shaopeng-gh, line 88: >Assert.True(WriteSarifThenReadLevelNode(result) == FailureLevel.Warning); [](http://example.com/codeflow?start=12&length=73)

this test will fail in Main, pass in the level change branch,
because in Main when it is Warning, we do not write the level node in Sarif at all.

src/Test.UnitTests.Sarif/Core/ResultTests.cs
```diff
@@ -43,6 +48,95 @@ public void Result_Level_ResetsKindValue()
...
+
+            result = new Result() { Level = FailureLevel.Warning };
+            defaultConfiguration = new ReportingConfiguration() { Enabled = true, Level = FailureLevel.Error };
+            Assert.True(WriteSarifThenReadLevelNode(result, defaultConfiguration) == FailureLevel.Warning);
+
+            result = new Result() { Level = FailureLevel.Warning };
+            defaultConfiguration = new ReportingConfiguration() { Enabled = true, Level = FailureLevel.Error };
+            overrideConfiguration = new ReportingConfiguration() { Enabled = true, Level = FailureLevel.Note };
+            Assert.True(WriteSarifThenReadLevelNode(result, defaultConfiguration, overrideConfiguration) == FailureLevel.Warning);
```
shaopeng-gh, line 97: >Assert.True(WriteSarifThenReadLevelNode(result, defaultConfiguration, overrideConfiguration) == FailureLevel.Warning); [](http://example.com/codeflow?start=12&length=118)

this test shows that, it does not matter what value is in the defaultConfiguration and overrideConfiguration, when we "write" the Sarif for Result.Level node.

the defaultConfiguration and overrideConfiguration is for when read/interpretation.

PR #2411, opened 2021-11-30
LGTM and NetAnalyzers improvements

eddynaka:
This change will fix the following warnings from .NET Analyzers:
- [CA1827: Do not use Count/LongCount when Any can be used](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1827)
- [CA1829: Use Length/Count property instead of Enumerable.Count method](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1829)
- [CA1836: Prefer IsEmpty over Count when available](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1836)
- [CA1841: Prefer Dictionary Contains methods](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1841)
- [CA1507: Use nameof in place of string
](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507)

It will also fix the HttpClient warning from LGTM.

### DIFF
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -182,9 +182,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         protected virtual void PostLogFile(string postUri, string outputFilePath, IFileSystem fileSystem)
         {
-            PostLogFile(postUri, outputFilePath, fileSystem, new HttpClient())
-                .GetAwaiter()
-                .GetResult();
+            using (var httpClient = new HttpClient())
+            {
+                PostLogFile(postUri, outputFilePath, fileSystem, httpClient)
+                    .GetAwaiter()
+                    .GetResult();
+            }
         }
 
         internal static async Task PostLogFile(string postUri, string outputFilePath, IFileSystem fileSystem, HttpClient httpClient)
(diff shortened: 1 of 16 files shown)


src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -182,9 +182,12 @@ protected virtual void ProcessBaseline(IAnalysisContext context, T driverOptions
...
         {
-            PostLogFile(postUri, outputFilePath, fileSystem, new HttpClient())
-                .GetAwaiter()
-                .GetResult();
+            using (var httpClient = new HttpClient())
+            {
+                PostLogFile(postUri, outputFilePath, fileSystem, httpClient)
+                    .GetAwaiter()
+                    .GetResult();
+            }
```
eddynaka, line 190: we need to use the old-style using to prevent it from being disposed before the method finishes.

michaelcfanning: Why are we creating a new http client? It is because we only expect a single post, for a single log file?

Also, you didn't explain why we can't use the new style using syntax. Why is that?

using var httpClient = new HttpClient();
PostLogFile(postUri, outputFilePath, fileSystem, httpClient)
                    .GetAwaiter()
                    .GetResult();

michaelcfanning (approved):
:shipit:

PR #2412, opened 2021-11-30
Fix `NullReferenceException` in WorkItemFiler when all results baseline state are unchange/absent/updated

yongyan-gh:
# Description:
BUG: Got `NullReferenceException` when WorkItemFiler tries to file a work item for the result that has baseline state should be ignored.

# Fixes:
- Handle null results case in SarifLog CreateWorkItemDescription extension method, which happens if all results are not eligible for filing work item. `result.ShouldBeFiled() == false`
- In `WorkItemFiler.FileWorkItems()`, do not create work item if all the results of target Sarif Log are not eligible for filing a work item. 

Currently the result is treated as it should not be file as work item if
  - Its baseline State is one of Unchanged, Absent, Updated
  - It's suppressed


yongyan-gh:
the issue has been resolved in commits after this.

---
In reply to: [983119894](https://github.com/microsoft/sarif-sdk/pull/2412#issuecomment-983119894) [](http://example.com/codeflow?ancestors=983119894,983119894)

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -271,6 +271,49 @@ public void SarifWorkItemExtensions_CreateWorkItemDescription_MultipleResults()
             description.Should().BeEquivalentTo($"This work item contains {numOfResult} '{toolName}' issue(s) detected in {string.Format(firstLocation, 1)} (+{additionLocationCount} locations). Click the 'Scans' tab to review results.");
         }
 
+        [Fact]
+        public void SarifWorkItemExtensions_CreateWorkItemDescription_ResultsShouldNotBeFiled()
+        {
+            string toolName = "TestToolName";
+            string firstLocation = @"C:\Test\Data\File{0}sarif";
```
shaopeng-gh, line 278: >@"C:\Test\Data\File{0}sarif" [](http://example.com/codeflow?start=35&length=28)

suggest change to not using windows specific format, like:
Path.Combine(Directory.GetCurrentDirectory(), "Test", "Data", "File{0}sarif");

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -152,7 +152,7 @@ public static string CreateWorkItemDescription(this SarifLog log, SarifWorkItemC
             Uri runRepositoryUri = log?.Runs.FirstOrDefault()?.VersionControlProvenance?.FirstOrDefault().RepositoryUri;
             Uri detectionLocationUri = !string.IsNullOrEmpty(runRepositoryUri?.OriginalString) ?
                                        runRepositoryUri :
-                                       results?.FirstOrDefault().Locations?[0].PhysicalLocation?.ArtifactLocation?.Uri;
+                                       results?.FirstOrDefault()?.Locations?[0].PhysicalLocation?.ArtifactLocation?.Uri;
```
eddynaka, line 155: >results?.FirstOrDefault()?.Locations?[0].PhysicalLocation?.ArtifactLocation?.Uri [](http://example.com/codeflow?start=39&length=80)

can u unwrap and wrap making ? and : be in separate lines?

eddynaka:
>                            .Count();

you can replace where for count and do thee unwrap/wrap as well

---
Refers to: src/Sarif.WorkItems/SarifWorkItemsExtensions.cs:166 in 9665225. [](commit_id = 9665225aa5d5675ae02aa8eef211c4ceb65fa348, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -290,6 +290,12 @@ public SarifWorkItemModel FileWorkItemInternal(SarifLog sarifLog, SarifWorkItemC
             {
                 string logId = sarifLog.GetProperty<Guid>(LOGID_PROPERTY_NAME).ToString();
 
+                if (sarifLog.Runs?.Any(run => run.Results?.Any(result => result.ShouldBeFiled()) == true) != true)
```
shaopeng-gh, line 293: >if (sarifLog.Runs?.Any(run => run.Results?.Any(result => result.ShouldBeFiled()) == true) != true) [](http://example.com/codeflow?start=16&length=98)

was thinking about the case of one run have work item and the other have empty array if this condiftion works.

do we think better to expand the nested condition, easiler to read and maintenance :)

shaopeng-gh, line 293: do we think also add a safety check in this method as well

public static bool ShouldBeFiled(this Result result)
{
    if(result ==null) return false;
    ...... 
}

shaopeng-gh (approved):
:shipit:

src/Sarif.WorkItems/SarifWorkItemFiler.cs
yongyan-gh: Added the check to ShouldBeFiled() and added test cases.

michaelcfanning:
@marmegh @EasyRhinoMSFT

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilerTests.cs
```diff
@@ -105,6 +106,41 @@ public void WorkItemFiler_ValidatesSarifLogFileLocationArgument()
...
+
+            SarifWorkItemContext context = CreateAzureDevOpsTestContext();
+
+            context.SplittingStrategy = SplittingStrategy.None;
+
+            int numberOfResults = 0;
+            context.SetProperty(ExpectedWorkItemsCount, numberOfResults);
+            context.SetProperty(ExpectedFilingResult, FilingResult.None);
+
+            TestWorkItemFiler(sarifLog, context, true);
```
michaelcfanning, line 141: >true [](http://example.com/codeflow?start=49&length=4)

As a rule don't embed literals in method calls. What does true here mean? We don't know.

Instead, qualify the literal with the parameter name. If the parameter name is good, the code will be cleaner:

TestWorkItemFiler(sarifLog, context, adoClient: true)

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilerTests.cs
```diff
@@ -105,6 +106,41 @@ public void WorkItemFiler_ValidatesSarifLogFileLocationArgument()
...
```
michaelcfanning, line 125: >true [](http://example.com/codeflow?start=49&length=4)

Is it important for these tests to verify the adoClient == false code path?

I think probably it is important, because we want to ensure the GitHub filer also doesn't file work items?

Or is the relevant code sufficiently covered by the adoClient case?

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -17,6 +18,8 @@
 {
     public class SarifWorkItemExtensionsTests
     {
+        private readonly string testSarifFilePath = Path.Combine(Directory.GetCurrentDirectory(), "Test", "Data", "File{0}.sarif");
```
michaelcfanning, line 21: >readonly [](http://example.com/codeflow?start=16&length=8)

Please add a comment here explaining that we create the test file path in this way so that we don't use a single, platform-specific file path rendering (i.e., we'll get Windows paths on Windows and Linux running xplat).

src/Sarif.WorkItems/SarifWorkItemFiler.cs
michaelcfanning, line 293: >sarifLog [](http://example.com/codeflow?start=20&length=8)

You should add logging to this code path, don't you think?

michaelcfanning:
>                             if (updatedSarifWorkItemModel == null)

Notice this 'return null' code path has good logging.

---
Refers to: src/Sarif.WorkItems/SarifWorkItemFiler.cs:329 in 5ebe0c8. [](commit_id = 5ebe0c81574c17b200c84ddbcc21e772ffaaa53a, deletion_comment = False)

src/Sarif.WorkItems/SarifWorkItemFiler.cs
michaelcfanning, line 293: >sarifLog [](http://example.com/codeflow?start=20&length=8)

I don't understand why this code path is what we want. You are going to traverse the entire log file, which could be very large. Is that what we want? What is this going to save us in work that we won't perform (when there are no issues to file)?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -12,6 +12,11 @@ public static class SarifWorkItemsExtensions
     {
         public static bool ShouldBeFiled(this Result result)
         {
+            if (result == null)
```
michaelcfanning, line 15: >result [](http://example.com/codeflow?start=16&length=6)

Why did we add this? Why wouldn't a null dereference be the expected condition?

i.e., why do we require

myObject?.ThisIsAnActualMethod()

to protect against a null variable but

myObject.ThisIsAnExtensionMethod()

should work when myObject is null?

I am asking a sincere question. But I'm not sure this change is a good design pattern, Shaopeng, what's your argument for it? :)

yongyan-gh: I think it depends on which one should be responsible for the null check, the caller or the extension method.

If the caller should handle null value then we don't need this check inside extension method.

myObject?.ThisIsAnExtensionMethod()

src/Sarif.WorkItems/SarifWorkItemFiler.cs
yongyan-gh: If the SARIF log doesn't have any eligible result to file as a work item and it doesn't return null, the following code will create a SarifWorkItemModel:

var sarifWorkItemModel = new SarifWorkItemModel(sarifLog, filingContext);

In SarifWorkItemModel code, it iterates all results and get first result which should be filed, to construct work item title. If none of result is eligible, the SairfWorkItemModel's title will be set to null, which causes NullReferenceException when calling filingClient.FileWorkItems() to file work item.

Another solution I can think is before calling FileWorkItemInternal(), in SplitLogFile method, make sure there is only eligible results in split logs. But seems traverse of the results is still needed.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
shaopeng-gh: I am not sure, my thought was similar to the IDictionary cast that we were looking at last PR, either a helper already did that for you or need to cast to IDictionary everywhere it is used, 
in this case if the extension method did the check for you, 
or where you call the extension method. 
only differ when user forgot to check null before call this extension seeing an unhandled exception because we call result.BaselineState at the first line, the other case is no exception returns false.

A quick find reference I see about 10 usages of this method.

In .net source code they also do null check at the first line of the 
extension methods, like below, our logic doe not need to throw:

public static partial class StringNormalizationExtensions
    {
        public static bool IsNormalized(this string strInput, NormalizationForm normalizationForm)
        {
            if (strInput == null)
            {
                throw new ArgumentNullException(nameof(strInput));
            }

            return strInput.IsNormalized(normalizationForm);
        }

src/Sarif.WorkItems/SarifWorkItemFiler.cs
yongyan-gh: removed in latest change

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemFilerTests.cs
yongyan-gh: Parameter adoClient controls which FilingClient mock object it generates, adoClient or githubClient. Added/updated test cases also covers githubClient.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
yongyan-gh: Like the way how .net handles in the extension method. The extension method will have similar behavior to instance method and can be protected by ? operation:

myObject?.ThisIsAnExtensionMethod()

src/Sarif.WorkItems/SarifWorkItemFiler.cs
yongyan-gh: Moved the check to SplitLogFile method, this way the sarifLog only includes eligible results to file work item, no need to traverse results. Pls let me know what do you think?

eddynaka (approved):
:shipit:

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -199,17 +199,18 @@ public virtual IReadOnlyList<SarifLog> SplitLogFile(SarifLog sarifLog)
...
-
                     PartitionFunction<string> partitionFunction = null;
 
                     Stopwatch splittingStopwatch = Stopwatch.StartNew();
 
                     switch (splittingStrategy)
                     {
+                        case SplittingStrategy.None:
+                        {
+                            // get ride of results should not be filed
```
michaelcfanning, line 210: Remove this comment. First, it does not follow convention, starting with a capital letter, ending with a period. It doesn't read very well grammatically. Not sure it provides useful additional information. Finally, this line of code is identical to other lines in this file, why are they commented? If they don't need to be, why is this line commented?

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
michaelcfanning: Here's the pattern that looks weird:

((Result)null).ShouldBeFiled();

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -12,6 +12,11 @@ public static class SarifWorkItemsExtensions
...
+            {
+                throw new ArgumentNullException(nameof(result));
+            }
+
```
michaelcfanning, line 19: I suggest removing this, because it is additional work and suggests a general pattern we should adhere that doesn't seem to have good ROI.

src/Sarif.WorkItems/SarifWorkItemsExtensions.cs
```diff
@@ -150,20 +155,19 @@ public static string CreateWorkItemDescription(this SarifLog log, SarifWorkItemC
...
-                           results
-                           .Where(r => r.Locations != null)
-                           .SelectMany(r => r.Locations)
-                           .Where(l => l.PhysicalLocation != null && l.PhysicalLocation.ArtifactLocation != null && l.PhysicalLocation.ArtifactLocation.Uri != null)
-                           .Count();
+            int locCount = results == null
+                ? 0
+                : results.Where(r => r.Locations != null)
+                         .SelectMany(r => r.Locations)
+                         .Count(l => l.PhysicalLocation?.ArtifactLocation?.Uri != null);
```
michaelcfanning, line 170: my personal opinion is this code isn't super readable and i avoid linq statements that potentially entail multiple iterations over a collection (in a pathological case).

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -17,6 +18,9 @@
 {
     public class SarifWorkItemExtensionsTests
     {
+        // Use portable file path so the tests can run on windows and other platforms.
```
michaelcfanning, line 21: > // Use portable file path so the tests can run on windows and other platforms. [](http://example.com/codeflow?start=7&length=79)

What's the utility of this comment?

src/Test.UnitTests.Sarif.WorkItems/SarifWorkItemExtensionsTests.cs
```diff
@@ -17,6 +18,9 @@
...
+        private readonly string testSarifFilePath = Path.Combine(Directory.GetCurrentDirectory(), "Test", "Data", "File{0}.sarif");
```
michaelcfanning, line 22: >testSarifFilePath [](http://example.com/codeflow?start=32&length=17)

please make this static and give it an s_ prefix

yongyan-gh:
Test coverage data:

image

michaelcfanning (approved):
:shipit:

PR #2413, opened 2021-12-01
Level Nullable Changes (2nd version)

shaopeng-gh:
# Description:

There were 3 issues reported related to Result.Level reported. (#2077, #2024, #2383)
My understanding of the issue is as following
- When the result level is explicitly set as warning, because we have the property default value set as warning and not written the Level json node if it is default. We will not be able to differentiate the case with user did not explicitly set the Level.
- this is also wrong, when the default configuration is e.g. Note, with the Level json node not written, it will be considered as Note instead of the warning user explicitly set.

# Fixes:

This PR propose the below fix:
- change Result internal field _level to be nullable and always write Level Json node if user has set it. When user not set, don't write the node. the public Level will still be not nullable since logically it will need not be.
- this will be able to differentiate if user has set the Result.Level 
- this will resolve the issue above when user explicitly set to Warning but the ending Sarif file is considered with level Note 
- also fix the existing issue that ShouldLog() does not consider default configuration

Code coverage Main branch:
image

Code coverage PR:
image

image
image
image

( there was an old PR we closed it and create this one, so will be clear to look at. https://github.com/microsoft/sarif-sdk/pull/2404 )


src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -56,7 +56,7 @@ protected void LogResult(string jPointer, string formatId, params string[] args)
 
             Context.Logger.Log(this,
                 RuleUtilities.BuildResult(
-                    DefaultConfiguration.Level,
+                    level: null,
```
shaopeng-gh, line 59: >level: null, [](http://example.com/codeflow?start=20&length=12)

this change is because we were passing the DefaultConfiguration.level and write duplicated in the Result.Level,
don't need to since we can leave out the Result.Level json node now.

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1001.ValidateWithBaseline.sarif
```diff
@@ -110,7 +110,6 @@
         {
           "ruleId": "SARIF2003",
           "ruleIndex": 0,
-          "level": "note",
```
shaopeng-gh, line 113: >"level": "note", [](http://example.com/codeflow?start=10&length=16)

these changes are the result of our fix in the SarifValidationSkimmerBase.cs file.

above we have 
"defaultConfiguration": {
    "level": "note"
},

src/Test.FunctionalTests.Sarif/v2/ConverterTestData/ContrastSecurity/WebGoat.xml.sarif
```diff
@@ -1865,6 +1865,7 @@
         {
           "ruleId": "crypto-bad-mac",
           "ruleIndex": 18,
+          "level": "warning",
```
shaopeng-gh, line 1868: >"level": "warning", [](http://example.com/codeflow?start=10&length=19)

these "level": "warning" lines added is because, in the C# code the way we using the sdk, we explicitly invoke set Level to Warning.
as by design, when user explicitly set Level to Warning we will have this json property generated.
If user does not explicitly invoke set Level, the _level will be null and this json property will not be generated.

src/Test.UnitTests.Sarif.Converters/TestData/Hdf/ExpectedOutputs/ValidResults.sarif
```diff
@@ -1204,6 +1326,7 @@
         },
         {
           "ruleId": "30002",
+          "level": "warning",
```
michaelcfanning, line 1329: >warning [](http://example.com/codeflow?start=20&length=7)

Your change in the main looks promising, but this outcome is clearly wrong, isn't it? This rule has no configuration, so the default value is warning already, and yet you've made that level explicit here. It shouldn't exist at all, isn't that correct?

Let's get on a phone call to discuss. I have an idea for you on how to test this. What you need here are very specific SARIF tests that comprehensively cover the spec. An ideal way to go about this is to author SARIF JSON and encode the expected failure level in the SARIF (via some convention in result production or as part of the property bag). Then you simply load the SARIF, verify that the API agrees the effective failure level is correct. Then serialize and deserialize the SARIF log and run the same check (to ensure that round-tripping works).

As a next step, let's author this test content and get it wired into a unit-test. This is the test-driven approach. Once we agree we've got all cases covered, well, when that test passes, you are done.

michaelcfanning (changes requested):
:clock1:

src/Test.UnitTests.Sarif.Converters/TestData/Hdf/ExpectedOutputs/ValidResults.sarif
shaopeng-gh: This rule has no configuration, so the default value is warning already, and yet you've made that level explicit here. It shouldn't exist at all, isn't that correct?
----- my understanding was user want to able able to tell if user explicitly set the level, or it is not set and took the default value.
[https://github.com/microsoft/sarif-sdk/pull/2383](https://github.com/microsoft/sarif-sdk/pull/2383)

we made the Level field internally nullable to track if the field is explicitly set by user, if set we will write to the Sarif. if user not set it, it will not be written and the reader of the Sarif file should interpret the effective level base on the Sarif spec Logic.

The problem you point out is what happen if user  explicitly set the Level, and it is the same as default configuration(exist or not)? I think if we don't write it, it will be the same problem user can not tell so I did not check to add any logic for that when we write to Sarif file. Let me know,

Let's get on a phone call to discuss --- yes!

let's author this test content and get it wired into a unit-test. ---- agree and we do need it to have round trip

gfs:
I tracked through the many layers of PR to this from this issue I filed: https://github.com/microsoft/sarif-sdk/pull/2383

It looks like this issue is still active in the SDK.  Is there another PR that is aiming to address this?

gfs:
@michaelcfanning @shaopeng-gh

shaopeng-gh:
Hi @gfs 
( Michael will able to provide more details if needed )
a quick update the status of this work item is currently parked, 
there are a few things we need to make change to the SARIF object module and we would like to bundle them as one braking change as possible, instead of several. 
Thus the change will not be a small change. 
We are working on fixing other issues before focusing on it.

michaelcfanning:
Hey, Gabe, how're things? It is hard to believe this issue is still open. :) It is a complex one. I *think* we understand it sufficiently to know what an appropriate fix is. Here's the deal, from my perspective:

- Our OM needs to express optional structs as nullable values. This is a key problem. Confusingly, SARIF specifies a sentinel value in some cases meaning 'value not present', e.g., for all properties that are indices into various arrays. I think this design was created to accommodate creation of object models in type systems where nullable structs aren't available or are considered cumbersome to use.  This format has definitely been around long enough for that to be true for C#.
- Long story short, related to above, SARIF OM needs to convert all `-1` index sentinel values to null properties. In other cases, other optional numeric types simply need to be nullable. We have another debt item, to convert these numeric types to an appropriate C# type, rather than the wholesale use of int as we have today.
- OK, so for `Level` there is special magic. This value is determined by data that may or may not exist in the rules specification. On surface, this should be a nullable thing. In practice, it needs to be a computed property, in my opinion. So, we need to convert the literal deserialization of the underlying property to a high-fidelity expression in the OM, as a nullable property, of type `FailureLevel?`. Then we provide a helper for `result.level` that computes the actual value, after consulting with the rules metadata.
- This raises *another* issue, mostly results don't have a back-pointer to their run instance. This prevents the rule config look-up I mentioned. So, what I propose to do is to update the Results retrieval pattern that's commonly used in the SDK to make sure the run instance is always pushed onto the result. We could also look into replacing the Run results enumerator with a special one that pushed the current run instance onto the result (this would be more backwards compatible).
- Although I think I have mental line-of-sight to a solution here, as you can see, it's involved and will require us to bolt some things onto the autogenerated code. Let me know if you have cycles to jump in. I can contribute some changes, we have some starter PRs, etc. The most important work done, I think, is progress on generating a comprehensive set of test data files that validate all possible scenarios. This is by far the key asset. :) If we can interpret every case properly according to the spec, brilliant, with that foothold we can move on to creating the most elegant, usable implementation that passes all tests.

gfs:
Hey Michael, 

Going well. We are going through and putting finishing touching on a major refactor of DevSkim and I rediscovered this issue.

Thanks for the extra context that helps explain the complexity. Do you have an estimate for when you may begin to tackle this?

I may be able to help with your to propose some test cases that cover the issue I had originally noticed that results with level `warning` are serialized with no level information.

gfs:
Here's a simple mstest case that demonstrates the issue. This test will fail the assertion with 1 results and 0 levels.

```csharp
using Microsoft.CodeAnalysis.Sarif;
using Newtonsoft.Json.Linq;

namespace TestProject1;

[TestClass]
public class UnitTest1
{
    [TestMethod]
    public void TestMethod1()
    {
        var resWithWarning = new Result()
        {
            Level = FailureLevel.Warning,
            RuleId = "1"
        };
        var rule = new ReportingDescriptor()
        {
            Id = "1"
        };
        rule.DefaultConfiguration = new ReportingConfiguration() { Enabled = true };
        rule.DefaultConfiguration.Level = FailureLevel.Warning;
        var run = new Run();
        run.Results = new List<Result>(){resWithWarning};
        run.Tool = new Tool();
        run.Tool.Driver = new ToolComponent();
        run.Tool.Driver.Name = "SarifSDKTests";
        run.Tool.Driver.Rules = new List<ReportingDescriptor>() { rule };
        var log = new SarifLog();
        log.Runs = new List<Run>() { run };
        string path = Path.GetTempFileName();
        log.Save(path);
        // We have to deserialize the sarif using JSON to get the raw data and not have fields repopulated by the SDK.
        var reReadLog = JObject.Parse(File.ReadAllText(path));
        IEnumerable<JToken> results = reReadLog.SelectTokens("$.runs[*].results[*]");
        IEnumerable<JToken> levels = reReadLog.SelectTokens("$.runs[*].results[*].level");
        Assert.AreEqual(results.Count(), levels.Count());
    }
}
```

gfs:
I also adapted this test into a workaround to use in DevSkim. https://github.com/microsoft/sarif-sdk/issues/2024#issuecomment-1523878929

derigel23:
Folks, what's the status of the issue? Is it merged?

PR #2415, opened 2021-12-09
BREAKING: Fix `InvalidOperationException` when using PropertiesDictionary in a multithreaded application and remove [Serializable] for it.

shaopeng-gh:
# Description:

Bug: `InvalidOperationException` when using PropertiesDictionary in a multithreaded application.
It is caused by use of Dictionary type instead of ConcurrentDictionary type.

# Fixes:

Change TypedPropertiesDictionary to use ConcurrentDictionary.
also found a existing issue with method SavePropertiesToXmlStream, fix by the way.

**BREAKING: removed [Serializable] for TypedPropertiesDictionary. 
[Serializable] is not for json and xml serialization, it is only for BinaryFormatter serialization. 
For SDK user that is currently using BinaryFormatter on TypedPropertiesDictionary object, 
will result in error: 
`SerializationException`: Type PropertiesDictionary is not marked as serializable**

# Error Details:

InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
at Microsoft.CodeAnalysis.Sarif.PropertiesDictionary.TryGetProperty[T](String key, T& value)
at Microsoft.CodeAnalysis.Sarif.PropertiesDictionary.GetProperty[T](PerLanguageOption`1 setting, Boolean cacheDefault)
at Microsoft.CodeAnalysis.Sarif.PropertiesDictionary.GetProperty[T](PerLanguageOption`1 setting)
at Microsoft.CodeAnalysis.IL.Rules.EnableSpectreMitigations.AnalyzePortableExecutableAndPdb(BinaryAnalyzerContext context) in D:\a\r1\a\src\BinSkim.Rules\PERules\BA2024.EnableSpectreMitigations.cs:line 139
at Microsoft.CodeAnalysis.IL.Rules.WindowsBinaryAndPdbSkimmerBase.Analyze(BinaryAnalyzerContext context) in D:\a\r1\a\src\BinSkim.Rules\PERules\WindowsBinaryAndPdbSkimmerBase.cs:line 63
at Microsoft.CodeAnalysis.Sarif.Driver.MultithreadedAnalyzeCommandBase`2.AnalyzeTargetHelper(TContext context, IEnumerable`1 skimmers, ISet`1 disabledSkimmers)

### DIFF
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -584,6 +585,16 @@ namespace Microsoft.CodeAnalysis.Sarif
                 // Consume the end element
                 xmlReader.Read();
             }
+        }
+
+        public static bool Add<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key, T value)
+        {
+            return concurrentDictionary.TryAdd(key, value);
+        }
+
+        public static bool Remove<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key)
+        {
+            return concurrentDictionary.TryRemove(key, out _);
         }
     }
 }
--- a/src/Sarif/ExtensionMethods.cs
+++ b/src/Sarif/ExtensionMethods.cs
@@ -587,14 +587,14 @@ namespace Microsoft.CodeAnalysis.Sarif
             }
         }
 
-        public static bool Add<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key, T value)
+        public static void Add<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key, T value)
         {
-            return concurrentDictionary.TryAdd(key, value);
+            ((IDictionary<string, T>)concurrentDictionary).Add(key, value);
         }
 
-        public static bool Remove<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key)
+        public static void Remove<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key)
         {
-            return concurrentDictionary.TryRemove(key, out _);
+            ((IDictionary<string, T>)concurrentDictionary).Remove(key);
         }
     }
 }
--- a/src/Sarif/PropertiesDictionary.cs
+++ b/src/Sarif/PropertiesDictionary.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel;
 using System.IO;
-using System.Runtime.Serialization;
 using System.Xml;
 
 using Newtonsoft.Json;
@@ -34,11 +33,6 @@ namespace Microsoft.CodeAnalysis.Sarif
             PropertiesDictionary initializer,
             IEqualityComparer<string> comparer)
             : base(initializer, comparer)
-        {
-        }
-
-        protected PropertiesDictionary(SerializationInfo info, StreamingContext context)
-            : base(info, context)
         {
         }
 
--- a/src/Sarif/PropertiesDictionary.cs
+++ b/src/Sarif/PropertiesDictionary.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel;
@@ -85,7 +86,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             if (cacheDescription)
             {
-                SettingNameToDescriptionsMap = SettingNameToDescriptionsMap ?? new Dictionary<string, string>();
+                SettingNameToDescriptionsMap ??= new ConcurrentDictionary<string, string>();
                 SettingNameToDescriptionsMap[setting.Name] = setting.Description;
             }
 
--- a/src/Sarif/PropertiesDictionary.cs
+++ b/src/Sarif/PropertiesDictionary.cs
@@ -15,7 +15,6 @@ using Newtonsoft.Json.Linq;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
-    [Serializable]
     [JsonConverter(typeof(TypedPropertiesDictionaryConverter))]
     public class PropertiesDictionary : TypedPropertiesDictionary<object>
     {
--- a/src/Sarif/PropertiesDictionaryExtensionMethods.cs
+++ b/src/Sarif/PropertiesDictionaryExtensionMethods.cs
@@ -71,10 +71,19 @@ namespace Microsoft.CodeAnalysis.Sarif
             {
                 object property = propertyBag[key];
 
+                if (settingNameToDescriptionMap != null)
+                {
+                    string description;
+
+                    if (settingNameToDescriptionMap.TryGetValue(key, out description))
+                    {
+                        writer.WriteComment(description);
+                    }
+                }
+
                 StringSet stringSet = property as StringSet;
                 if (stringSet != null)
                 {
-                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     SaveSet(writer, stringSet, key);
                     continue;
                 }
@@ -82,7 +91,6 @@ namespace Microsoft.CodeAnalysis.Sarif
                 IntegerSet integerSet = property as IntegerSet;
                 if (integerSet != null)
                 {
-                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     SaveSet(writer, integerSet, key);
                     continue;
                 }
@@ -90,12 +98,9 @@ namespace Microsoft.CodeAnalysis.Sarif
                 IDictionary pb = property as IDictionary;
                 if (pb != null)
                 {
-                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     ((IDictionary)pb).SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
                     continue;
                 }
-
-                SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
 
                 writer.WriteStartElement(PROPERTY_ID);
                 writer.WriteAttributeString(KEY_ID, key);
@@ -113,19 +118,6 @@ namespace Microsoft.CodeAnalysis.Sarif
                 writer.WriteEndElement(); // KeyValuePair
             }
             writer.WriteEndElement(); // Properties
-        }
-
-        private static void SaveSettingNameToDescriptionMap(XmlWriter writer, string key, ConcurrentDictionary<string, string> settingNameToDescriptionMap)
-        {
-            if (settingNameToDescriptionMap != null)
-            {
-                string description;
-
-                if (settingNameToDescriptionMap.TryGetValue(key, out description))
-                {
-                    writer.WriteComment(description);
-                }
-            }
         }
 
         private static string NormalizeTypeName(string typeName)
--- a/src/Sarif/PropertiesDictionaryExtensionMethods.cs
+++ b/src/Sarif/PropertiesDictionaryExtensionMethods.cs
@@ -74,6 +74,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 StringSet stringSet = property as StringSet;
                 if (stringSet != null)
                 {
+                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     SaveSet(writer, stringSet, key);
                     continue;
                 }
@@ -81,6 +82,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 IntegerSet integerSet = property as IntegerSet;
                 if (integerSet != null)
                 {
+                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     SaveSet(writer, integerSet, key);
                     continue;
                 }
@@ -88,19 +90,12 @@ namespace Microsoft.CodeAnalysis.Sarif
                 IDictionary pb = property as IDictionary;
                 if (pb != null)
                 {
+                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
                     ((IDictionary)pb).SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
                     continue;
                 }
 
-                if (settingNameToDescriptionMap != null)
-                {
-                    string description;
-
-                    if (settingNameToDescriptionMap.TryGetValue(key, out description))
-                    {
-                        writer.WriteComment(description);
-                    }
-                }
+                SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
 
                 writer.WriteStartElement(PROPERTY_ID);
                 writer.WriteAttributeString(KEY_ID, key);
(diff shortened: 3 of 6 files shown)


src/Sarif/PropertiesDictionary.cs
```diff
@@ -37,11 +36,6 @@ public PropertiesDictionary() : this(null)
         {
         }
 
-        protected PropertiesDictionary(SerializationInfo info, StreamingContext context)
-            : base(info, context)
-        {
-        }
```
eddynaka, line 43: check if the serialization will be the same comparing previous vs current

eddynaka:
>                 SettingNameToDescriptionsMap = SettingNameToDescriptionsMap ?? new Dictionary<string, string>();

I think you might need to change this as well.
Be sure to test the cacheDescription part

---
Refers to: src/Sarif/TypedPropertiesDictionary.cs:72 in bef3fbe. [](commit_id = bef3fbeb4ad38813679891ee4e8c90dceec245c8, deletion_comment = False)

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
```diff
@@ -124,6 +125,40 @@ public void PropertiesDictionary_RoundTripStringToVersionMap()
...
+                    taskList.Add(Task.Factory.StartNew(() =>
+                    properties.GetProperty(GetStringSetProperty(i))));
+                }
+
+                Task.WaitAll(taskList.ToArray());
+            });
+            Assert.Null(exception);
+        }
+
+
```
eddynaka, line 161: remove one line

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
```diff
@@ -192,6 +227,10 @@ private PropertiesDictionary RoundTripThroughJson(PropertiesDictionary propertie
             new PerLanguageOption<StringSet>(
                 FEATURE, nameof(StringSetProperty), defaultValue: () => { return STRINGSET_DEFAULT; });
 
+        public static PerLanguageOption<StringSet> GetStringSetProperty(int i) =>
```
eddynaka, line 230: >GetStringSetProperty [](http://example.com/codeflow?start=51&length=20)

I would change to StringGetProperty

shaopeng-gh: StringSet is a actually class here, let me know 
how about change to "GenerateStringSetProperty"
the Get word may be confusing

shaopeng-gh:
>                 SettingNameToDescriptionsMap = SettingNameToDescriptionsMap ?? new Dictionary<string, string>();

got it

---
In reply to: [989748904](https://github.com/microsoft/sarif-sdk/pull/2415#issuecomment-989748904) [](http://example.com/codeflow?ancestors=989748904,989748904)

---
Refers to: src/Sarif/TypedPropertiesDictionary.cs:72 in bef3fbe. [](commit_id = bef3fbeb4ad38813679891ee4e8c90dceec245c8, deletion_comment = False)

src/Sarif/PropertiesDictionary.cs
shaopeng-gh: got it, thanks

shaopeng-gh:
>                 SettingNameToDescriptionsMap = SettingNameToDescriptionsMap ?? new Dictionary<string, string>();

I think the cacheDescription part have problem before our changes. And we don't have a any test for that field.
I have made some changes, testing.

---
In reply to: [990377643](https://github.com/microsoft/sarif-sdk/pull/2415#issuecomment-990377643) [](http://example.com/codeflow?ancestors=990377643,989748904)

---
Refers to: src/Sarif/TypedPropertiesDictionary.cs:72 in bef3fbe. [](commit_id = bef3fbeb4ad38813679891ee4e8c90dceec245c8, deletion_comment = False)

src/Sarif/PropertiesDictionary.cs
shaopeng-gh: the file written are the same

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
shaopeng-gh: fixed

shaopeng-gh:
>                 SettingNameToDescriptionsMap = SettingNameToDescriptionsMap ?? new Dictionary<string, string>();

added test for the cacheDescription save.

---
In reply to: [990734565](https://github.com/microsoft/sarif-sdk/pull/2415#issuecomment-990734565) [](http://example.com/codeflow?ancestors=990734565,989748904)

---
Refers to: src/Sarif/TypedPropertiesDictionary.cs:72 in bef3fbe. [](commit_id = bef3fbeb4ad38813679891ee4e8c90dceec245c8, deletion_comment = False)

src/Sarif/PropertiesDictionaryExtensionMethods.cs
```diff
@@ -73,33 +74,28 @@ public static class PropertiesDictionaryExtensionMethods
                 StringSet stringSet = property as StringSet;
                 if (stringSet != null)
                 {
+                    SaveSettingNameToDescriptionMap(writer, key, settingNameToDescriptionMap);
```
eddynaka, line 77: >SaveSettingNameToDescriptionMap [](http://example.com/codeflow?start=20&length=31)

pls, add tests that cover this change.

src/Sarif/TypedPropertiesDictionary.cs
```diff
@@ -13,7 +13,7 @@ public interface IMarker { }
 
     [Serializable]
     [JsonConverter(typeof(TypedPropertiesDictionaryConverter))]
-    public class TypedPropertiesDictionary<T> : Dictionary<string, T>, IMarker where T : new()
+    public class TypedPropertiesDictionary<T> : ConcurrentDictionary<string, T>, IMarker where T : new()
```
eddynaka, line 16: >ConcurrentDictionary [](http://example.com/codeflow?start=48&length=20)

Update the title of the PR and add a new entry to the ReleaseHistory.

The title should be something like: "Fix InvalidOperationException when using PropertiesDictionary in a multithreaded application".

With that, if any customer checks the release history, the customer would know what motivated us to change this from Dictionary to ConcurrentDictionary.

This should also be applied to the PR description.

michaelcfanning:
At a high-level, this PR isn't well explained. I will go annotate code changes that require more clarification.

src/Sarif/PropertiesDictionary.cs
```diff
@@ -37,11 +37,6 @@ public PropertiesDictionary() : this(null)
         {
         }
 
-        protected PropertiesDictionary(SerializationInfo info, StreamingContext context)
```
michaelcfanning, line 40: >PropertiesDictionary [](http://example.com/codeflow?start=18&length=20)

Why did you remove this constructor? Preivously in .NET this ctor was required for all [Serializable] things, has that changed?

src/Sarif/ExtensionMethods.cs
```diff
@@ -585,5 +586,15 @@ internal static void ConsumeElementOfDepth(this XmlReader xmlReader, int endElem
                 xmlReader.Read();
             }
         }
+
+        public static bool Add<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key, T value)
```
michaelcfanning, line 590: >Add [](http://example.com/codeflow?start=27&length=3)

Why did you add this? What does the CurrentDictionary IDictionary.Add implementation call into? A non-thread-safe base method?

michaelcfanning, line 590: >bool [](http://example.com/codeflow?start=22&length=4)

why does this extension method return a bool? IDictionary.Add returns void.

It's not a good idea to create a strange return value for 'Add', this will confuse experienced .NET programmers.

If you need to override 'Add', so that a concurrent dictionary behaves properly when it flows around to IDictionary-accepting things, this should return void. And in the case where Add would throw, you should consult the return value of TryAdd and throw on false.

src/Sarif/PropertiesDictionaryExtensionMethods.cs
```diff
@@ -119,6 +115,19 @@ public static class PropertiesDictionaryExtensionMethods
             writer.WriteEndElement(); // Properties
         }
 
+        private static void SaveSettingNameToDescriptionMap(XmlWriter writer, string key, ConcurrentDictionary<string, string> settingNameToDescriptionMap)
```
michaelcfanning, line 118: >SaveSettingNameToDescriptionMap [](http://example.com/codeflow?start=28&length=31)

What are we doing here? Seems like you found a different issue and are oportunistically fixing it?

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
```diff
@@ -124,6 +130,68 @@ public void PropertiesDictionary_RoundTripStringToVersionMap()
...
+                    properties.GetProperty(GenerateStringSetProperty(i))));
+                }
+
+                Task.WaitAll(taskList.ToArray());
+            });
+            Assert.Null(exception);
+        }
+
+        [Fact]
+        public void PropertiesDictionary_TestCacheDescription()
```
michaelcfanning, line 167: >PropertiesDictionary_TestCacheDescription [](http://example.com/codeflow?start=20&length=41)

Well done! Love seeing these tests relevant to your change, nice work.

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
michaelcfanning: 'StringSet' is correct. :) I can completely understand why 'GenerateStringGetProperty' seems right, though! :)

michaelcfanning (changes requested):
:clock1:

src/Sarif/ExtensionMethods.cs
shaopeng-gh: just checked the source code of CurrentDictionary,
the IDictionary<TKey, TValue>.Add implementation
did call the same method as the TryAdd
so we can use it thread safely.

Also agree that we should not make the Add has a return value people will get confused. 

updated the code.

shaopeng-gh: agree and updated, see the other comment.

src/Sarif/PropertiesDictionaryExtensionMethods.cs
shaopeng-gh: there were no tests for this and when I add tests this issue show up and blocking my test pass so I should fix it together.

avoid injecting four calls to your new method ---- agree, let me take a look

src/Sarif/PropertiesDictionaryExtensionMethods.cs
shaopeng-gh: found this new method is not needed, removed.
the bug was comment not written to file,
the method I add was to test that. PropertiesDictionary_TestCacheDescription

src/Sarif/TypedPropertiesDictionary.cs
shaopeng-gh: done both, thanks

src/Sarif/PropertiesDictionaryExtensionMethods.cs
shaopeng-gh: fixed to only one copy. the method is not needed.

src/Sarif/PropertiesDictionary.cs
shaopeng-gh: the .net core concurrentdictionary is not binary serializable, so I removed it, 

question, do we even need it, it is not for json and xml serialization, it is only for BinaryFormatter serialization, in the solution I don't see code use BinaryFormatter

src/Sarif/TypedPropertiesDictionary.cs
shaopeng-gh: also added the ReleaseHistory for the other 2 PRs.

src/Sarif/PropertiesDictionary.cs
michaelcfanning: Right, so if it's used anywhere it's in unit tests, I'd guess.

If you intend for this dictionary *not* to be serializable, you should remove the [Serializable] attribute. I think that's ok to do.

michaelcfanning: That's a good question and in future, definitely try to raise this before making this kind of change. Remember we are in an SDK, so it is possible that someone is using BinaryFormatter with this public type (even though we do not as you point out).

I think probably it is unlikely we're breaking someone out there. So this change is fine. *However* we need to very clearly mark this breaking change in behavior in the release notes.

Imagine someone using BInaryFormatter with this type, using an older SDK. Now they update to the new SDK and everything is broken. Ideally, your release notes would also mention the symptom these users would receive. Make sense? IOW, try to use BinaryFormatter again PropertiesDictionary (once you remove the [Serializable] attribute as well. See what exception is raised. Make sure the release notes reference this exception in case users experience it.

By doing this, we leave an easily followed trail of clues that tell developers what happened. That's the least we can do, given that we're introducing a breaking change. :)

michaelcfanning:
>     [Serializable]

Let's remove this, if we're removing the serialization ctor.

---
Refers to: src/Sarif/PropertiesDictionary.cs:18 in 8eb304d. [](commit_id = 8eb304d1bee931243c956f8b83ee1264eb583948, deletion_comment = False)

shaopeng-gh:
>     [Serializable]

fixed.

---
In reply to: [1004422690](https://github.com/microsoft/sarif-sdk/pull/2415#issuecomment-1004422690) [](http://example.com/codeflow?ancestors=1004422690,1004422690)

---
Refers to: src/Sarif/PropertiesDictionary.cs:18 in 8eb304d. [](commit_id = 8eb304d1bee931243c956f8b83ee1264eb583948, deletion_comment = False)

src/Sarif/PropertiesDictionary.cs
shaopeng-gh: thanks for the great inputs, 
I have remove the [Serializable] attribute and ran it, the error is 
SerializationException: Type PropertiesDictionary is not marked as serializable
I have added it both in the PR desc and also the release history md file.

shaopeng-gh: @michaelcfanning 

additional question, do you want me to create a work item to check and remove all [Serializable] in the solution,
( I have not dig if any of them is used somewhere yet, just to bring up the discussion )

[https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide)

[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2300](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2300)

src/Test.UnitTests.Sarif/PropertiesDictionaryTests.cs
```diff
@@ -124,6 +130,69 @@ public void PropertiesDictionary_RoundTripStringToVersionMap()
             ((TypedPropertiesDictionary<Version>)properties[MapKey])[ValueKey].Should().Be(version);
         }
 
+        [Fact]
+        public void PropertiesDictionary_ConcurrentReadAndWrite()
+        {
+            var properties = new PropertiesDictionary();
+
+            Exception exception = Record.Exception(() =>
```
michaelcfanning, line 138: >Exception [](http://example.com/codeflow?start=41&length=9)

Did you verify that this test actually fails when the fix isn't applied?

shaopeng-gh: yes this test can repro the error before change the type.

src/Sarif/ExtensionMethods.cs
```diff
@@ -585,5 +586,15 @@ internal static void ConsumeElementOfDepth(this XmlReader xmlReader, int endElem
                 xmlReader.Read();
             }
         }
+
+        public static void Add<T>(this ConcurrentDictionary<string, T> concurrentDictionary, string key, T value)
```
michaelcfanning, line 590: >ConcurrentDictionary [](http://example.com/codeflow?start=39&length=20)

Why do we need these? Can't we just require our code to cast to IDictionary when appropriate?

michaelcfanning, line 590: >Add [](http://example.com/codeflow?start=27&length=3)

Looking at this more closely, I think what we need to do is to reintroduce IDictionary<string, T> on the TypedPropertiesDictionary class, it is currently not available because it is implemented explicitly in ConcurrentDictionary.

michaelcfanning (approved):
:shipit:

PR #2419, opened 2022-01-06
Rtm 6

michaelcfanning:


src/Sarif/Schemata/sarif-external-property-file-2.1.0-rtm.5.json
```diff
@@ -17,13 +17,13 @@
     },
 
     "guid": {
-      "description": "A stable, unique identifer for this external properties object, in the form of a GUID.",
+      "description": "A stable, unique identifier for this external properties object, in the form of a GUID.",
```
michaelcfanning, line 20: >identifier [](http://example.com/codeflow?start=39&length=10)

Oops, we probably should create an rtm6 version of this file which only has this spelling change added.

src/ESLint.Formatter/sarif.js
```diff
@@ -54,7 +54,7 @@ module.exports = function (results, data) {
 
     const sarifLog = {
         version: "2.1.0",
-        $schema: "http://json.schemastore.org/sarif-2.1.0-rtm.5",
+        $schema: "http://json.schemastore.org/sarif-2.1.0-rtm.6",
```
michaelcfanning, line 57: >$schema [](http://example.com/codeflow?start=8&length=7)

It's a very small thing but could this constant value be shared somewhere, so that we only make a single change, that flows to both the production formatter and its tests?

michaelcfanning:
>   "additionalProperties": false,

I'll need to diff this against the OASIS schema to verify that it's correct. Easiest way is to see the changes when we update schemastore.org, so will wait for that.

---
Refers to: src/Sarif/Schemata/sarif-2.1.0-rtm.6.json:6 in cb53ef4. [](commit_id = cb53ef47ea70f0e4c0569b02591fe559c2623bed, deletion_comment = False)

michaelcfanning:
>                             modifiedLog |= ApplyRtm5Changes(logObject);

So! This helper potentially makes modifications for an update to RTM5 and we don't need to do that for RTM 5. I *believe* there's no potential code level update from v5 - v6, but we should double-check a diff of the schema.

If there are no modifications required, let's just create a new empty case, of add it to .6 below (noting that v5 is treated identically to current.

If we discover a delta that could/should be handled programmatically, we will need to author it. Again, let's check this point with the schemastore.org change.

---
Refers to: src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs:86 in cb53ef4. [](commit_id = cb53ef47ea70f0e4c0569b02591fe559c2623bed, deletion_comment = False)

michaelcfanning:
Can you add Eddy and Mary to this PR so they learn about this topic? Also, let's add something to the release notes on this change. And let's add a very brief markdown doc in the \docs\ folder. 'HowToUpdateTheSaiifSchema.md. Eventually, this doc should also contain info on updating via schemastore.org (as well as the steps you've completed in this change).

michaelcfanning:
>   "$schema": "http://json-schema.org/draft-04/schema#",

Are there v5 docs we can now delete? i.e., the two old schema files.

---
Refers to: src/Sarif/Schemata/sarif-2.1.0-rtm.6.json:2 in cb53ef4. [](commit_id = cb53ef47ea70f0e4c0569b02591fe559c2623bed, deletion_comment = False)

src/Sarif/Writers/PrereleaseCompatibilityTransformer.cs
```diff
@@ -85,8 +87,8 @@ public static class PrereleaseCompatibilityTransformer
                             modifiedLog |= UpdateSarifLogVersionAndSchema(logObject);
                             break;
                         }
-                        case "http://json.schemastore.org/sarif-2.1.0-rtm.5":
-                        case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json":
+                        case "http://json.schemastore.org/sarif-2.1.0-rtm.6":
+                        case "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.6.json":
                         {
                             // Current schema version. There should be no work to do
```
michaelcfanning, line 93: >Current [](http://example.com/codeflow?start=31&length=7)

So, we should add a prerelease transformer test for this scenario, updating from rtm.5 to current. Looks like a no-op for now.

PR #2420, opened 2022-01-07
Revert Json Serialization order change

shaopeng-gh:


src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -363,14 +363,17 @@ private void InitializeOutputFile(TOptions analyzeOptions, TContext context, ISe
...
                         SarifLogger sarifLogger;
 
-                        _run = new Run()
+                        _run = new Run();
+
+                        if (!string.IsNullOrWhiteSpace(analyzeOptions.AutomationId)
+                        || !string.IsNullOrWhiteSpace(analyzeOptions.AutomationGuid))
                         {
-                            AutomationDetails = new RunAutomationDetails
+                            _run.AutomationDetails = new RunAutomationDetails
```
shaopeng-gh, line 371: >new RunAutomationDetails [](http://example.com/codeflow?start=53&length=24)

The issue is, if there is no AutomationId and AutomationGuid, we don't need to create a instance here.
instance not null with all property null will still get written with current code:
SerializeIfNotNull(_run.AutomationDetails, "automationDetails");

src/Sarif/NotYetAutoGenerated/ReportingDescriptor.cs
```diff
@@ -41,87 +41,88 @@ public SarifNodeKind SarifNodeKind
...
-        [DataMember(Name = "help", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "help", IsRequired = false, EmitDefaultValue = false, Order = 4)]
         public virtual MultiformatMessageString Help { get; set; }
 
         /// <summary>
         /// An array of objects that describe relationships between this reporting descriptor and others.
         /// </summary>
-        [DataMember(Name = "relationships", IsRequired = false, EmitDefaultValue = false)]
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        [DataMember(Name = "relationships", IsRequired = false, EmitDefaultValue = false, Order = 13)]
```
shaopeng-gh, line 117: >Order [](http://example.com/codeflow?start=90&length=5)

if JsonProperty tag exists we should also set it to the same because the default is -1 will get the filed at the first line.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF2012.ProvideRuleProperties_Invalid.sarif
```diff
@@ -18,8 +18,8 @@
             },
             {
               "id": "TEST0002",
-              "name": "This isn't pascal case",
-              "helpUri": "http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html"
+              "helpUri": "http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html",
+              "name": "This isn't pascal case"
```
shaopeng-gh, line 22: test case logic is, read this json input into c# object, and write the object back to json, and then do analyze on it.

with the order current set, helpUri is before name, so the result will have the line number +1 from 21 to 22.
To make test more readable I changed this input to be the same order so when we see the error is line 22 we can understand which line it is pointing to. Let me know your thoughts.

src/Sarif/NotYetAutoGenerated/ReportingDescriptor.cs
```diff
@@ -41,87 +41,88 @@ public SarifNodeKind SarifNodeKind
         /// <summary>
         /// A stable, opaque identifier for the report.
         /// </summary>
-        [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false, Order = 1)]
```
shaopeng-gh, line 44: >1 [](http://example.com/codeflow?start=87&length=1)

how I get the current order:

I use the version before issue, and then temp change code set all fields always written even it is null, so that I can get the full list of the ordering, and then use the same ordering.

fyi it was like below:

{
              "id": "BA2001",
              "fullDescription": {
                "text": "64-bit images should"
              },
              "helpUri": "[https://gith](https://gith)",
              "help": {
                "text": "64-bit ima"
              },
              "messageStrings": {
                "Pass": {
                  "text": "'{0}' is"
                },
                "Error": {
                  "text": "'{0}' is"
                },
                "NotApplicable_InvalidMetadata": {
                  "text": "'{0}' was"
                }
              },
              "shortDescription": null,
              "name": "LoadImageAboveFourGigabyteAddress",
              "deprecatedIds": null,
              "guid": null,
              "deprecatedGuids": null,
              "deprecatedNames": null,
              "defaultConfiguration": {},
              "relationships": null,
              "properties": {
                "equivalentBinScopeRuleReadableName": "FourGbCheck"
              }
            }

marmegh:
We really have two identical files in two different folders? I don't expect this to be changed this go around, but I'm curious Michael/Eddy: was there a reason for the repetition here?

src/Sarif/NotYetAutoGenerated/ReportingDescriptor.cs
marmegh: Thanks for being so thorough here with your explanation. I found this super helpful.

src/Sarif/Autogenerated/ReportingDescriptor.cs
```diff
@@ -41,87 +41,88 @@ public SarifNodeKind SarifNodeKind
         /// <summary>
         /// A stable, opaque identifier for the report.
         /// </summary>
-        [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
+        [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false, Order = 1)]
         public virtual string Id { get; set; }
```
marmegh, line 45: Not sure if this is covered in oru standard yet; but we probably want to consider if we're explicitly setting an order whether we want that to be how we sort files like this or if we want to do alphabetical, etc.

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Revert Json Serialization field order change and skip emit empty AutomationDetails node. [#2420](https://github.com/microsoft/sarif-sdk/pull/2420)
```
marmegh, line 5: >* BUGFIX: Revert Json Serialization field order change and skip emit empty AutomationDetails node. [](http://example.com/codeflow?start=0&length=98)

I was under the impression that the order change reversion was just until the other PR is approved and then we'll settle on a final order and consolidate the two changes.

Would this be better described accordingly:
"Standardizing Json Serialization field order and..."?

shaopeng-gh:
add my input:
we have those classes auto generated by tool [https://github.com/microsoft/jschema](https://github.com/microsoft/jschema)
(this tool is also own my our team)
since there are still some attributes/capability we are not yet supporting, for those class we make a copy from autogenerated folder to notyet folder.
and make changes needed in the notyet folder. when build, script will copy from notyet folder to autogenerated folder.
(but I suggest please keep them insync and check in both because it will be easier for local debugging)


---
In reply to: [1007846306](https://github.com/microsoft/sarif-sdk/pull/2420#issuecomment-1007846306) [](http://example.com/codeflow?ancestors=1007846306,1007846306)

src/ReleaseHistory.md
marmegh: I would consider this to be my only blocking comment at the moment.

shaopeng-gh: agree, let me change to word like "Adjust" instead of Standardizing I fear people think it is a whole solution standardizing by looking the title

shaopeng-gh:
Your question reminds me of I think I should add a note for not yet auto generated comment in code. Added.
// NOTYETAUTOGENERATED: Order attribute

---
In reply to: [1007856536](https://github.com/microsoft/sarif-sdk/pull/2420#issuecomment-1007856536) [](http://example.com/codeflow?ancestors=1007856536,1007846306)

src/ReleaseHistory.md
shaopeng-gh: updated the title.

src/Sarif/Writers/ResultLogJsonWriter.cs
```diff
@@ -318,18 +318,31 @@ public void CompleteRun()
...
+                Serialize(_run.Graphs, "graphs");
+            }
 
             // Results go here in schema order
-
-            SerializeIfNotNull(_run.AutomationDetails, "automationDetails");
+            if (_run.ShouldSerializeAutomationDetails())
+            {
+                Serialize(_run.AutomationDetails, "automationDetails");
+            }
```
eddynaka, line 332: can you create a test using resultlogjsonwriter and verify if it will emit or not.

eddynaka: also, check if the other run properties are using the shouldSerialize pattern and if yes, we should add the same logic as this.

shaopeng-gh: 1. for this one, before this PR, the test we already have
"AnalyzeCommandBase_ShouldEmitAutomationDetailsWhenIdOrGuidExists()"
is already going this path and use the resultlogjsonwriter and verify if it is correctly emitted, (can F5 to step though)

the problem was just there is no case for validate the opposite situation when it should not emit,

so I added cases for null to loop in it, and change the test method name to a generic name reflect it tests both cases.

2. agree, I just go though all fields in the run.cs those have shouldSerialize, and made the same change, updated the PR. the new ones I added are:

ShouldSerializeArtifacts
ShouldSerializeInvocations
ShouldSerializeLogicalLocations

eddynaka (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #2425, opened 2022-01-20
Remove unnecessary NuGet.Config in sub folder

shaopeng-gh:
# Description:

There were 2 NuGet.Config one in root and one in sub folder src.
When we want to make change it is confusing which one should change.

# Fixes:

the one in sub folder is not unnecessary.
it is only ref by ToDotNet. 
the difference is with this extra NuGet.Config, 
the packages microsoft.json.schema.todotnet will only be copied to {root}\src\packages.
without this extra NuGet.Config,
the packages microsoft.json.schema.todotnet will be copied to {root}\src\packages, and also {root}\packages 

extra package in the local packages cache does not have effect, I think it is better to opt to have only one NuGet.Config.

# Tests:

1. verified build locally using BuildAndTest.cmd compare to main branch can produce same set of files, until the last step for create final packages e.g. \Publish\Nuget\Release\Sarif.Sdk.2.4.12.nupkg
same size but different hash.
2. verified can build and F5 using VS as before.
3. verified can build on pipeline.


src/Sarif.Sdk.sln
```diff
@@ -47,7 +47,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{90E0CA8D-0D6
 		BinSkimConfiguration.xml = BinSkimConfiguration.xml
 		build.props = build.props
 		CommonAssemblyInfo.cs = CommonAssemblyInfo.cs
-		NuGet.Config = NuGet.Config
```
eddynaka, line 50: >NuGet.Config [](http://example.com/codeflow?start=17&length=12)

change the reference to point to the nuget.config in the root folder.

eddynaka (approved):
:shipit:

marmegh (approved):
:shipit:

src/Sarif.Sdk.sln
```diff
@@ -47,7 +47,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{90E0CA8D-0D6
...
+		..\NuGet.Config = ..\NuGet.Config
```
michaelcfanning, line 50: >NuGet.Config [](http://example.com/codeflow?start=23&length=12)

please move this to the .nuget directory

michaelcfanning (approved):
:shipit:

src/Sarif.Sdk.sln
shaopeng-gh: Nuget.config is base on a system level Nuget.config,
and then override with the parent folder Nuget.config if any,
and then override with the current folder Nuget.config if any.

the location matters, we can not move to .nuget directory, if so it will only apply to that sub folder.

PR #2426, opened 2022-01-21
Fix spelling error in rtm.5 schema

EasyRhinoMSFT:
This change results in updated auto-generated model classes and also touches a handful of functional test data.


src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
```diff
@@ -591,13 +591,13 @@
         },
 
         "guid": {
-          "description": "A stable, unique identifer for this external properties object, in the form of a GUID.",
```
eddynaka, line 594: >identifer [](http://example.com/codeflow?start=43&length=9)

is it right to make changes to a schema that is already published? not sure how the flow works.

i saw that is only english word changes, but not sure if that is correct or not

EasyRhinoMSFT: Somehow rtm.5 was published without this misspelling. We decided to think of this as a change to bring the published and checked-in schemas into sync, and would therefore not create a new version.

michaelcfanning (approved):

Issue #2428, opened 2022-01-26
'postUri' command naming convention is wrong


michaelcfanning:
This argument s/be named 'post-uri' to match other names, e.g. 'rich-return-code'

@shaopeng-gh @yongyan-gh, easy issue to fix. Can we make it a priority? This is a breaking change if we've already shipped the previous argument name.

shaopeng-gh:
got it, let me create a PR to update it.

shaopeng-gh:
I see this option was added after latest release, so will not affect user.

shaopeng-gh:
https://github.com/microsoft/sarif-sdk/pull/2429

PR #2429, opened 2022-01-26
BUGFIX: Update command line argument name from `postUri` to `post-uri`.

shaopeng-gh:
# Description:

This argument s/be named 'post-uri' to match other names, e.g. 'rich-return-code'
This is a breaking change if we've already shipped the previous argument name.

# Fixes:

Checked it was not released yet.
update the name.


src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Update command line argument name from `postUri` to `post-uri`. [#2427](https://github.com/microsoft/sarif-sdk/pull/2427)
```
marmegh, line 5: >BUGFIX [](http://example.com/codeflow?start=2&length=6)

Why is this categorized as a bugfix? Did the naming cause a problem for customers? Was this to address a particular issue?

marmegh:

src/ReleaseHistory.md
shaopeng-gh: this is created from the issue
[https://github.com/microsoft/sarif-sdk/issues/2428](https://github.com/microsoft/sarif-sdk/issues/2428)

Did the naming cause a problem for customers
---- no, because it was not yet released.

Why is this categorized as a bugfix
---- since it was not yet released, so change of the parameter name will not break user. let me know if you think I should use another keyword.

michaelcfanning: If the issue hasn't shipped as a customer issue, you don't have to document it in the release notes. Make sense? In other words, all of these notes should refer to changes strictly from v2.4.12.

Now if it *had* shipped previously, you'd need to clearly mark this as a breaking change.

So the note here is remove the release notes entry. Developers use commit history to track intra-version changes. Release notes, again, only document changes from the last shipped version.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -111,7 +111,7 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
         public string BaselineSarifFile { get; set; }
 
         [Option(
-            "postUri",
+            "post-uri",
             HelpText = "A URI to which the SARIF log file will be posted.")]
```
michaelcfanning, line 115: >HelpText [](http://example.com/codeflow?start=12&length=8)

Where is my unit test? I will always ask a question if you propose a change without a unit test. In all gentleness, I'd like to report to you that I am growing tired of giving this note in pull requests. If you don't propose a test with your change, you must please explain why you don't believe it is necessary.

This is a well-tested area, I think, but clearly we don't test a top-down test that starts with the command-line. In general this is a weaker area of our testing.

michaelcfanning (changes requested):
:clock1:

src/ReleaseHistory.md
shaopeng-gh: use commit history to track intra-version changes ---- got it, yes then I should remove it.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
shaopeng-gh: added 2 type of tests:
one is for command line, test change the string from postUri to post-uri;
the other is testing the function after command line is already parsed.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -653,6 +655,14 @@ public void AnalyzeCommandBase_DefaultEndToEndAnalysis()
             configurationNotificationCount.Should().Be(0);
         }
 
+        [Fact]
+        public void AnalyzeCommandBase_EndToEndAnalysisWithPostUri()
+        {
+            PostUriTestHelper(@"https://httpbin.org/post", TestAnalyzeCommand.SUCCESS, RuntimeConditions.None);
```
shaopeng-gh, line 661: >@"https://httpbin.org/post" [](http://example.com/codeflow?start=30&length=27)

have a question for these 2,
these are public test uri and we only post a sample Sarif without any secret.
let me know if this is ok.

src/Test.UnitTests.Sarif.Multitool/CommandLineParameterTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            validateOptions = this.CommandLineTestHelper<ValidateOptions>(
+                new string[] { "validate", sampleSarifFileName, "--post-uri", samplePostUri },
+                valid: true);
+            validateOptions.PostUri.Should().Be(samplePostUri);
+
+            validateOptions = this.CommandLineTestHelper<ValidateOptions>(
+                new string[] { "validate", sampleSarifFileName, "--postUri", samplePostUri },
+                valid: false);
+            validateOptions.Should().BeNull();
```
shaopeng-gh, line 41: test for the change of parameter name

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
michaelcfanning: This kind of off-network activity can cause issues in unit tests. Typically we would mock this. Ed has a nice mockable HTTP client.

PR #2430, opened 2022-01-26
Dispose root context before unloading process.

michaelcfanning:
This change ensures that every 'root' context that is created at analysis time is disposed in all code paths. This hasn't been a problem to-date as no tool has required an active dispose before process unload. BinSkim now collects telemetry that is finalized in a telemetry logger Dispose method, however, and so it must be definitively dispose on process exit to ensure all data is flushed and written.

@marmegh @shaopeng-gh @eddynaka @EasyRhinoMSFT @yongyan-gh


src/Sarif/Core/SarifLog.cs
```diff
@@ -87,7 +87,7 @@ public static SarifLog Load(Stream source, bool deferred = false)
 
             if (!fileSystem.FileExists(filePath))
             {
-                throw new ArgumentException(nameof(filePath));
+                throw new ArgumentException($"File path does not exist: '{filePath}'", nameof(filePath));
```
michaelcfanning, line 90: >ArgumentException [](http://example.com/codeflow?start=26&length=17)

This is a very comment API use issue, this exception type does not accept a method overload that only references the parameter name (unlike, say, NullReferenceException).

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -66,71 +66,78 @@ public string DefaultConfigurationPath
...
                 }
-            }
 
-            if (options.RichReturnCode)
+                return succeeded ? SUCCESS : FAILURE;
+            }
+            finally
             {
-                return (int)RuntimeErrors;
+                _rootContext?.Dispose();
```
michaelcfanning, line 139: >_rootContext?.Dispose [](http://example.com/codeflow?start=16&length=21)

Here's the key change. In all code paths on calling Run, we dispose the root context object if non-null.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1237,10 +1250,10 @@ private void AnalyzeScenarios(int[] scenarios)
 
                 testCase.FileSystem = null;
                 testCase.Files = multiThreadTargets;
-                Run runMultiThread = RunAnalyzeCommand(options, testCase, multithreaded: true);
+                Run runMultithreaded = RunAnalyzeCommand(options, testCase, multithreaded: true);
```
michaelcfanning, line 1253: >runMultithreaded [](http://example.com/codeflow?start=20&length=16)

Notice that I fix incidental code issues that I notice while I'm working. My goal is that *every* code change leaves the code base in a slightly better condition, even when I am extending/adding functionality to it.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -40,6 +40,19 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
             this.Output = output;
         }
 
+        [Fact]
+        public void AnalyzeCommandBase_RootContextIsDisposed()
```
michaelcfanning, line 44: >AnalyzeCommandBase_RootContextIsDisposed [](http://example.com/codeflow?start=20&length=40)

Observe my test approach. First I add a focused test for both single and multithreaded analysis. This test, however, does not come close to testing all possible exit paths (this code returns in various places to handle exceptional paths, for example). So I will go inject this validation directly into our test commands as well.

This is a general strategy of mine when adding tests to legacy code: author directed tests, but look for opportunities to generalize coverage in the test infrastructure, via dependency injection, etc.

src/Test.UnitTests.Sarif.Driver/TestMultithreadedAnalyzeCommand.cs
```diff
@@ -59,7 +61,9 @@ protected override void ValidateOptions(TestAnalyzeOptions options, TestAnalysis
 
         public int Run(AnalyzeOptionsBase options)
         {
-            return base.Run((TestAnalyzeOptions)options);
+            int result = base.Run((TestAnalyzeOptions)options);
+            this._rootContext?.Disposed.Should().BeTrue();
```
michaelcfanning, line 65: >_rootContext [](http://example.com/codeflow?start=17&length=12)

Now every test that travels this code path will enforce this invariant.

src/Test.UnitTests.Sarif.Driver/TestAnalyzeCommand.cs
```diff
@@ -70,5 +72,12 @@ protected override void ProcessBaseline(IAnalysisContext context, TestAnalyzeOpt
 
             base.ProcessBaseline(context, options, fileSystem);
         }
+
+        public override int Run(TestAnalyzeOptions options)
+        {
+            int result = base.Run(options);
+            this._rootContext?.Disposed.Should().BeTrue();
```
michaelcfanning, line 79: >_rootContext [](http://example.com/codeflow?start=17&length=12)

Now every test that travels this code path will enforce this invariant.

eddynaka (approved):
:shipit:

PR #2431, opened 2022-01-27
Artifacts should only be logged in the SARIF if we have results

eddynaka:
# Description
During some analysis, we saw that we are emitting an artifact even if we don't have a result. This is not good because it will make the SARIF gigantic in case you analyze hundreds/thousands of files.

# Proposal
We will only emit an artifact if we have a result.

# Tests
Created one 'file' of each level/kind and checked the SARIF output. It should only generate a list of artifacts equal to the locations you have in the results.

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -307,12 +307,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
             if (results?.Count > 0)
             {
-                if (_computeHashes)
-                {
-                    _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
-                                       dataToInsert: _dataToInsert,
-                                       hashData: context.Hashes);
-                }
+                _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
+                                   dataToInsert: _dataToInsert,
+                                   hashData: _computeHashes ? context.Hashes : null);
 
                 foreach (KeyValuePair<ReportingDescriptor, IList<Result>> kv in results)
                 {
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -309,7 +309,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             {
                 _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
                                    dataToInsert: _dataToInsert,
-                                   hashData: _computeHashes ? context.Hashes : null);
+                                   hashData: context.Hashes);
 
                 foreach (KeyValuePair<ReportingDescriptor, IList<Result>> kv in results)
                 {
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -307,6 +307,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
             if (results?.Count > 0)
             {
+                if (_computeHashes)
+                {
+                    _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
+                                       dataToInsert: _dataToInsert,
+                                       hashData: context.Hashes);
+                }
+
                 foreach (KeyValuePair<ReportingDescriptor, IList<Result>> kv in results)
                 {
                     foreach (Result result in kv.Value)
@@ -484,10 +491,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                             paths = new List<string>();
                             _hashToFilesMap[hashData.Sha256] = paths;
                         }
-
-                        _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
-                                           dataToInsert: _dataToInsert,
-                                           hashData: hashData);
 
                         paths.Add(localPath);
                         context.Hashes = hashData;
--- a/src/Sarif/Writers/SarifLogger.cs
+++ b/src/Sarif/Writers/SarifLogger.cs
@@ -19,7 +19,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
     {
         private readonly Run _run;
         private readonly TextWriter _textWriter;
-        private readonly bool _persistArtifacts;
         private readonly bool _closeWriterOnDispose;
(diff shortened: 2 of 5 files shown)


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -155,52 +154,17 @@ public class SarifLogger : BaseLogger, IDisposable, IAnalysisLogger
...
-                    artifact.Location = fileLocation;
-
-                    // This call will insert the file object into run.Files if not already present
-                    artifact.Location.Index = _run.GetFileIndex(
-                        artifact.Location,
-                        addToFilesTableIfNotPresent: true,
-                        dataToInsert: dataToInsert,
-                        encoding: encoding,
-                        hashData: hashData);
-                }
```
eddynaka, line 203: This code path is used by the single threaded analysis. When hashes is enabled, it was going to generate all hashes and store everything in the SARIF.

The issue is that if we analyze 1k files and we do not produce any results, we would still generate the artifacts.

As an optimization, we will only store the artifact if we have a result.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -447,12 +411,16 @@ private void CaptureArtifact(ArtifactLocation fileLocation)
                 catch (ArgumentException) { } // Unrecognized encoding name
             }
 
+            HashData hashData = null;
+            AnalysisTargetToHashDataMap?.TryGetValue(fileLocation.Uri.OriginalString, out hashData);
```
eddynaka, line 415: >AnalysisTargetToHashDataMap?.TryGetValue(fileLocation.Uri.OriginalString, out hashData); [](http://example.com/codeflow?start=12&length=88)

For the single threaded analysis, the `AnalysisTargetToHashDataMap` will have the hashes and here, we will just use the information that we already have and store in the artifacts.

The CaptureArtifact method is only used when we are logging the artifacts of a result.

eddynaka: If hashes isn't enabled, analysisTargetToHashDataMap will be null, so we are using the null check to guarantee and prevent a null reference exception.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -485,10 +492,6 @@ private async Task<bool> HashAsync()
                             _hashToFilesMap[hashData.Sha256] = paths;
                         }
 
-                        _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
-                                           dataToInsert: _dataToInsert,
-                                           hashData: hashData);
```
eddynaka, line 490: instead of adding artifacts for all files, we are moving this to the place when we have an actual result.

The hash itself will already be in the context, so, no need to calculate it again.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1258,21 +1258,93 @@ private void AnalyzeScenarios(int[] scenarios)
         }
 
         [Fact]
-        public void AnalyzeCommandBase_MultithreadedShouldUseCacheIfFilesAreTheSame()
+        public void AnalyzeCommandBase_Multithreaded_ShouldOnlyLogArtifactsWhenHashesIsEnabled()
```
eddynaka, line 1261: >AnalyzeCommandBase_Multithreaded_ShouldOnlyLogArtifactsWhenHashesIsEnabled [](http://example.com/codeflow?start=20&length=74)

This test will guarantee that we are always producing the correct number of artifacts if hashes is enabled for the multithreaded analysis.

If not enabled, the artifacts should be null.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1258,21 +1258,93 @@ private void AnalyzeScenarios(int[] scenarios)
...
+
+            // Hashes is disabled so no artifacts are expected.
+            run.Artifacts.Should().BeNull();
+            run.Results.Should().HaveCount(expectedNumberOfArtifacts);
+            run.Results.Count(r => r.Level == FailureLevel.Error).Should().Be(expectedNumberOfResultsWithErrors);
+            run.Results.Count(r => r.Level == FailureLevel.Warning).Should().Be(expectedNumberOfResultsWithWarnings);
+        }
+
+        [Fact]
+        public void AnalyzeCommmandBase_SingleThreaded_ShouldOnlyLogArtifactsWhenHashesIsEnabled()
```
eddynaka, line 1316: >AnalyzeCommmandBase_SingleThreaded_ShouldOnlyLogArtifactsWhenHashesIsEnabled [](http://example.com/codeflow?start=20&length=76)

This test will guarantee that we are always producing the correct number of artifacts if hashes is enabled for the single threaded analysis.

If not enabled, the artifacts should be null.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -335,7 +335,23 @@ public void SarifLogger_WritesFileData()
...
+                        {
+                            PhysicalLocation = new PhysicalLocation
+                            {
+                                ArtifactLocation = new ArtifactLocation
+                                {
+                                    Uri = new Uri(file)
+                                }
+                            }
+                        }
+                    };
```
eddynaka, line 352: we require a result with a location to see the artifacts.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -348,55 +364,6 @@ public void SarifLogger_WritesFileData()
...
-                    // Create a logger that inserts artifact contents.
-                    using (_ = new SarifLogger(
-                        textWriter,
-                        run: run,
-                        analysisTargets: analysisTargets,
-                        dataToInsert: OptionallyEmittedData.TextFiles,
-                        levels: new List<FailureLevel> { FailureLevel.Warning, FailureLevel.Error },
-                        kinds: new List<ResultKind> { ResultKind.Fail }))
-                    {
-                    }
```
eddynaka, line 389: no results are being emitted, with that, no artifacts should be logged. This test was wrong.

src/Sarif/Writers/SarifLogger.cs
eddynaka: Reviewed all versions that we released and I saw that all versions after 1.4.2 are emitting the artifacts IF the list of analysisTargets is not empty.

eddynaka: Looking at the other variables, looks like we tried to improve this behavior adding the _persistArtifacts, but that is occurring too late (after we already added the artifact to the run)

eddynaka:
This PR got replaced by this: https://github.com/microsoft/sarif-sdk/pull/2433

PR #2433, opened 2022-01-28
Fixing `AnalyzeCommandBase` and `MultithreadedAnalyzeCommandBase` artifacts generation

eddynaka:
# Description
The single/multi-threaded command base was generating all the artifacts even if we were not producing results.

# Test
Testing multiple files with and without the hashes flag and checking if artifacts were produced.

This PR has a dependency on this: https://github.com/microsoft/sarif-sdk/pull/2434, which will reduce the number of files to be analyzed once we merge it. That PR is just rebaselining our test files, preventing new changes from receiving all the changes we are seeing in this.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* BUGFIX: Fix `AnalyzeCommandBase` and `MultithreadedAnalyzeCommandBase` from outputting all artifacts to SARIF even if no results were produced when Hashes is enabled. [#2433](https://github.com/microsoft/sarif-sdk/pull/2433)
+* BREAKING: `AnalyzeCommandBase` previously persisted all scan target artifacts to SARIF logs rather than only persisting artifacts referenced by an analysis result, when an option to persist hashes, text file or binary information was set. `MultithreadedAnalyzeCommandBase` previously persisted all scan targets artifacts to SARIF logs in cases when hash insertion was eenabled rather than only persisting artifacts referenced by an analysis result. [#2433](https://github.com/microsoft/sarif-sdk/pull/2433)
 * BUGFIX: Adjust Json Serialization field order for ReportingDescriptor and skip emit empty AutomationDetails node. [#2420](https://github.com/microsoft/sarif-sdk/pull/2420)
 * BREAKING: Fix `InvalidOperationException` when using PropertiesDictionary in a multithreaded application, and remove `[Serializable]` from it. Now use of BinaryFormatter on it will result in `SerializationException`: Type `PropertiesDictionary` is not marked as serializable. [#2415](https://github.com/microsoft/sarif-sdk/pull/2415)
 
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix `AnalyzeCommandBase` and `MultithreadedAnalyzeCommandBase` from outputting all artifacts to SARIF even if no results were produced when Hashes is enabled. [#2433](https://github.com/microsoft/sarif-sdk/pull/2433)
 * BUGFIX: Adjust Json Serialization field order for ReportingDescriptor and skip emit empty AutomationDetails node. [#2420](https://github.com/microsoft/sarif-sdk/pull/2420)
 * BREAKING: Fix `InvalidOperationException` when using PropertiesDictionary in a multithreaded application, and remove `[Serializable]` from it. Now use of BinaryFormatter on it will result in `SerializationException`: Type `PropertiesDictionary` is not marked as serializable. [#2415](https://github.com/microsoft/sarif-sdk/pull/2415)
 
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -287,7 +287,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                 if ((options.DataToInsert.ToFlags() & OptionallyEmittedData.Hashes) != 0)
                 {
-                    if (_pathToHashDataMap?.TryGetValue(filePath, out HashData hashData) == true)
+                    if (_pathToHashDataMap != null && _pathToHashDataMap.TryGetValue(filePath, out HashData hashData))
                     {
                         context.Hashes = hashData;
                     }
--- a/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
@@ -169,7 +169,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             targets = ValidateTargetsExist(_rootContext, targets);
 
             // 5. Initialize report file, if configured.
-            InitializeOutputFile(options, _rootContext, targets);
+            InitializeOutputFile(options, _rootContext);
 
             // 6. Instantiate skimmers.
             ISet<Skimmer<TContext>> skimmers = CreateSkimmers(options, _rootContext);
@@ -287,13 +287,14 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                 if ((options.DataToInsert.ToFlags() & OptionallyEmittedData.Hashes) != 0)
                 {
-                    if (_pathToHashDataMap != null && _pathToHashDataMap.TryGetValue(filePath, out HashData hashData))
+                    if (_pathToHashDataMap?.TryGetValue(filePath, out HashData hashData) == true)
                     {
                         context.Hashes = hashData;
                     }
                     else
                     {
                         context.Hashes = HashUtilities.ComputeHashes(filePath);
+                        _pathToHashDataMap?.Add(filePath, context.Hashes);
                     }
                 }
             }
@@ -352,7 +353,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             }
         }
 
-        private void InitializeOutputFile(TOptions analyzeOptions, TContext context, ISet<string> targets)
+        private void InitializeOutputFile(TOptions analyzeOptions, TContext context)
         {
             string filePath = analyzeOptions.OutputFilePath;
             AggregatingLogger aggregatingLogger = (AggregatingLogger)context.Logger;
@@ -388,7 +389,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                                     dataToRemove,
                                     tool: _tool,
                                     run: _run,
-                                    analysisTargets: targets,
+                                    analysisTargets: null,
                                     quiet: analyzeOptions.Quiet,
                                     invocationTokensToRedact: GenerateSensitiveTokensList(),
                                     invocationPropertiesToLog: analyzeOptions.InvocationPropertiesToLog,
@@ -404,7 +405,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                                     dataToRemove,
                                     tool: _tool,
                                     run: _run,
-                                    analysisTargets: targets,
+                                    analysisTargets: null,
                                     invocationTokensToRedact: GenerateSensitiveTokensList(),
                                     invocationPropertiesToLog: analyzeOptions.InvocationPropertiesToLog,
                                     levels: analyzeOptions.Level,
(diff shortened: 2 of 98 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -307,6 +308,11 @@ private void LogCachingLogger(TContext rootContext, IAnalysisLogger logger, TCon
 
             if (results?.Count > 0)
             {
+                _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
+                                   addToFilesTableIfNotPresent: _persistArtifacts,
+                                   dataToInsert: _dataToInsert,
+                                   hashData: context.Hashes);
```
eddynaka, line 314: we should only persist if:
1. we have results
2. if we have hashes/textfiles/binaryfiles flag enabled

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -485,10 +491,6 @@ private async Task<bool> HashAsync()
                             _hashToFilesMap[hashData.Sha256] = paths;
                         }
 
-                        _run?.GetFileIndex(new ArtifactLocation { Uri = context.TargetUri },
-                                           dataToInsert: _dataToInsert,
-                                           hashData: hashData);
```
eddynaka, line 490: this will prevent us from generating artifacts for all the files

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1165,7 +1165,7 @@ public void AnalyzeCommandBase_AutomationDetailsTests()
             }
         }
 
-        [Fact(Timeout = 5000)]
+        [Fact(Timeout = 5000, Skip = "Artifacts will be different while we don't fix SarifLogger and AnalyzeCommandBase.")]
         public void AnalyzeCommandBase_ShouldGenerateSameResultsWhenRunningSingleAndMultiThread_CoyoteTest()
```
eddynaka, line 1169: >AnalyzeCommandBase_ShouldGenerateSameResultsWhenRunningSingleAndMultiThread_CoyoteTest [](http://example.com/codeflow?start=20&length=86)

This test is comparing the output of a single thread vs multithreaded command base.

With this change, both will have different outputs which will be fixed in the future.

michaelcfanning, line 1169: >AnalyzeCommandBase_ShouldGenerateSameResultsWhenRunningSingleAndMultiThread_CoyoteTest [](http://example.com/codeflow?start=20&length=86)

SingleAndMultithreaded

src/Test.UnitTests.Sarif.Multitool.Library/SuppressCommandTests.cs
```diff
@@ -170,7 +170,7 @@ private static void VerifySuppressCommand(SuppressOptions options)
 
             if (options.ExpiryInDays > 0 && suppression.TryGetProperty("expiryUtc", out DateTime expiryUtc))
             {
-                expiryUtc.Should().BeCloseTo(DateTime.UtcNow.AddDays(options.ExpiryInDays));
+                expiryUtc.Should().BeCloseTo(DateTime.UtcNow.AddDays(options.ExpiryInDays), precision: 500);
```
michaelcfanning, line 173: >500 [](http://example.com/codeflow?start=103&length=3)

used a shared constant

michaelcfanning:
Where is my release note?

src/Sarif/HashUtilities.cs
```diff
@@ -31,15 +31,14 @@ internal static IFileSystem FileSystem
 
         public static IDictionary<string, HashData> MultithreadedComputeTargetFileHashes(IEnumerable<string> analysisTargets, bool suppressConsoleOutput = false)
         {
-            if (analysisTargets == null) { return null; }
+            var fileToHashDataMap = new ConcurrentDictionary<string, HashData>();
+            if (analysisTargets == null) { return fileToHashDataMap; }
```
eddynaka, line 35: This will prevent a NullReferenceException when we use the reerence variable from the SarifLogger

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -447,12 +447,16 @@ private void CaptureArtifact(ArtifactLocation fileLocation)
                 catch (ArgumentException) { } // Unrecognized encoding name
             }
 
+            HashData hashData = null;
+            AnalysisTargetToHashDataMap?.TryGetValue(fileLocation.Uri.OriginalString, out hashData);
```
eddynaka, line 451: >AnalysisTargetToHashDataMap?.TryGetValue [](http://example.com/codeflow?start=12&length=40)

if we have the hash, let's re-use it.
this will prevent us from hashing the same file twice

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -169,7 +169,7 @@ private void Analyze(TOptions options, AggregatingLogger logger)
             targets = ValidateTargetsExist(_rootContext, targets);
 
             // 5. Initialize report file, if configured.
-            InitializeOutputFile(options, _rootContext, targets);
```
eddynaka, line 172: >InitializeOutputFile(options, _rootContext, targets); [](http://example.com/codeflow?start=12&length=53)

we won't initialize the outputfile with targets because we don't want to emit all targets in the artifacts.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -485,12 +486,9 @@ private async Task<bool> HashAsync()
...
-
                         paths.Add(localPath);
                         context.Hashes = hashData;
+                        _pathToHashDataMap?.Add(localPath, hashData);
                     }
```
eddynaka, line 492: Updating this will enable us to use the code from SarifLogger to create the artifacts

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1001.ValidateWithBaseline.sarif
```diff
@@ -97,15 +97,6 @@
...
       ],
-      "artifacts": [
-        {
-          "location": {
-            "uri": "FunctionalTestOutput.BaselineOption/ToBeValidated.sarif",
-            "uriBaseId": "TEST_DIR",
-            "index": 0
-          }
-        }
-      ],
```
eddynaka, line 108: this will go away, since we are not using any flag to emit artifacts.

src/Test.FunctionalTests.Sarif/TestData/Multitool/BaselineOption/ExpectedOutputs/TEST1001.ValidateWithBaseline.sarif
```diff
@@ -59,14 +59,14 @@
                   "text": "{0}: The tool '{1}' does not provide 'informationUri'. This property helps the developer responsible for addessing a result by providing a way to learn more about the tool."
                 }
               },
-              "helpUri": "http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html"
+              "name": "ProvideToolProperties"
             },
             {
               "id": "SARIF2012",
-              "name": "ProvideRuleProperties",
```
eddynaka, line 66: >name [](http://example.com/codeflow?start=15&length=4)

the order of the properties were enforced by shaopeng in a previous PR.

src/Test.UnitTests.Sarif.Multitool.Library/SuppressCommandTests.cs
eddynaka: done!

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
eddynaka: done!

eddynaka: already fixed in this pr

src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* BUGFIX: Fix `AnalyzeCommandBase` and `MultithreadedAnalyzeCommandBase` from outputting all artifacts to SARIF even if no results were produced when Hashes is enabled. [#2433](https://github.com/microsoft/sarif-sdk/pull/2433)
```

* BREAKING: `AnalyzeCommandBase` previously persisted all scan target artifacts to SARIF logs rather than only persisting artifacts referenced by an analysis result, when an option to persist hashes, text file or binary information was set. `MultithreadedAnalyzeCommandBase` previously persisted all scan targets artifacts to SARIF logs in cases when hash insertion was eenabled rather than only persisting artifacts referenced by an analysis result.

src/Sarif/HashUtilities.cs
```diff
@@ -31,15 +31,14 @@ internal static IFileSystem FileSystem
 
         public static IDictionary<string, HashData> MultithreadedComputeTargetFileHashes(IEnumerable<string> analysisTargets, bool suppressConsoleOutput = false)
         {
-            if (analysisTargets == null) { return null; }
```
michaelcfanning, line 34: >null [](http://example.com/codeflow?start=50&length=4)

Please revert this change. You should be careful about allocating objects to indicate no results. Why not just have the caller handle null? This code succeeded previously, so you probably have only just introduced the requirement for this to be non-null.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
michaelcfanning, line 172: >targets [](http://example.com/codeflow?start=56&length=7)

Revert this change please it's unnecessarily breaking customers.

src/Sarif/HashUtilities.cs
eddynaka: reverted

src/Sarif/OptionallyEmittedData.cs
```diff
@@ -80,6 +80,9 @@ public enum OptionallyEmittedData : int
         // Enrich SARIF log with git blame information
         GitBlameInformation = 0x1000,
 
+        // Enrich with artifacts only.
+        Artifacts = 0x2000,
```
eddynaka, line 84: >Artifacts [](http://example.com/codeflow?start=8&length=9)

To keep the same behavior as before, I introduced a new flag that will be able to emit the artifacts only

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
eddynaka: This API isn't exposed to customers. Only tools that use it will see a difference.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF2012.ProvideRuleProperties_Invalid.sarif
```diff
@@ -120,7 +120,7 @@
                   "index": 0
                 },
                 "region": {
-                  "startLine": 22,
```
eddynaka, line 123: >22 [](http://example.com/codeflow?start=31&length=2)

we faced a similar issue in this PR:
[https://github.com/microsoft/sarif-sdk/pull/2420/files#diff-277d16bd3a7355417dd787cfc0961db73b7f32e73ad49650468d4f4bb6e5c477](https://github.com/microsoft/sarif-sdk/pull/2420/files#diff-277d16bd3a7355417dd787cfc0961db73b7f32e73ad49650468d4f4bb6e5c477)

We should investigate this.

michaelcfanning: Yes, let's open an issue on it.

michaelcfanning (approved):
:shipit:

PR #2435, opened 2022-01-30
Add suppression support

Yiwei-Ding:
This PR is to add suppression support for the sarif formatter of ESLint.

### Suppression support in ESLint

In eslint/eslint#15459, the suppression support feature was merged into ESLint. When disabling suppressions by inline comments, the violations will be put into `suppressedMessages` in the output instead of being discarded. The text after two or more adjacent dashes will be considered as the justification. Detailed spec can be seen in [rfcs/2021-suppression-support](https://github.com/eslint/rfcs/tree/main/designs/2021-suppression-support).

For example:
```js
foo=1 // eslint-disable-line no-undef -- for test
```
```shell
> eslint -f json 1.js
```
```json
[
  {
    ...
    "messages":[],
    "suppressedMessages":[{
      "ruleId":"no-undef",
      "severity":2,
      "message":"'foo' is not defined.",
      "line":1,
      "column":1,
      "nodeType":"Identifier",
      "messageId":"undef",
      "endLine":1,
      "endColumn":4,
      "suppressions":[{"kind":"directive","justification":"for test"}]}],
    ...
  }
]
```
**NOTE**: currently only inline suppressions are supported, so `kind` of `suppressions` is always "directive". Reason for why "directive" not "inSource" could be seen in this [comment](https://github.com/eslint/rfcs/pull/82#discussion_r695306329).

### What this PR does

In this PR, I modify the formatter to convert `suppressedMessages` from ESLint to `results` with the `suppressions` property in the format of SARIF.

As for the example above, if we run `eslint -f @microsoft/sarif 1.js`, we will get:
```json
{
  "version": "2.1.0",
  "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.5",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "ESLint",
          "informationUri": "https://eslint.org",
          "rules": [
            {
              "id": "no-undef",
              "helpUri": "https://eslint.org/docs/rules/no-undef",
              "properties": {},
              "shortDescription": {
                "text": "disallow the use of undeclared variables unless mentioned in `/*global */` comments"
              }
            }
          ],
          "version": "8.6.0"
        }
      },
      "artifacts": [
        {
          "location": {
            "uri": "file:///D:/test/1.js"
          }
        }
      ],
      "results": [
        {
          "level": "error",
          "message": {
            "text": "'foo' is not defined."
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "file:///D:/test/1.js",
                  "index": 0
                },
                "region": {
                  "startLine": 1,
                  "startColumn": 1
                }
              }
            }
          ],
          "ruleId": "no-undef",
          "ruleIndex": 0,
          "suppressions": [
            {
              "kind": "inSource",
              "justification": "for test"
            }
          ]
        }
      ]
    }
  ]
}
```

### DIFF
--- a/src/ESLint.Formatter/sarif.js
+++ b/src/ESLint.Formatter/sarif.js
@@ -132,9 +132,11 @@ module.exports = function (results, data) {
                     console.log(err);
                 }
             }
-            if (result.messages.length > 0) {
 
-                result.messages.forEach(message => {
+            const messages = result.messages.concat(result.suppressedMessages);
+
+            if (messages.length > 0) {
+                messages.forEach(message => {
 
                     const sarifRepresentation = {
                         level: getResultLevel(message),
@@ -208,7 +210,7 @@ module.exports = function (results, data) {
                         }
                         if (message.column > 0) {
                             sarifRepresentation.locations[0].physicalLocation.region.startColumn = message.column;
-                        };
+                        }
                     }
 
                     if (message.source) {
@@ -218,6 +220,15 @@ module.exports = function (results, data) {
                         sarifRepresentation.locations[0].physicalLocation.region.snippet = {
                             text: message.source
                         };
+                    }
+
+                    if (message.suppressions) {
+                        sarifRepresentation.suppressions = message.suppressions.map(suppression => {
+                            return {
+                                kind: suppression.kind === "directive" ? "inSource" : "external",
+                                justification: suppression.justification
+                            }
+                        });
                     }
 
                     if (message.ruleId) {
--- a/src/ESLint.Formatter/sarif.js
+++ b/src/ESLint.Formatter/sarif.js
@@ -133,7 +133,7 @@ module.exports = function (results, data) {
                 }
             }
 
-            const messages = result.messages.concat(result.suppressedMessages);
+            const messages = result.suppressedMessages ? result.messages.concat(result.suppressedMessages) : result.messages;
 
             if (messages.length > 0) {
                 messages.forEach(message => {
@@ -229,6 +229,8 @@ module.exports = function (results, data) {
                                 justification: suppression.justification
                             }
                         });
+                    } else {
+                        sarifRepresentation.suppressions = [];
                     }
 
                     if (message.ruleId) {
--- a/src/ESLint.Formatter/test/sarif.js
+++ b/src/ESLint.Formatter/test/sarif.js
@@ -127,6 +127,8 @@ describe("formatter:sarif", () => {
             assert(log.runs[0].results[0].locations[0].physicalLocation.artifactLocation.uri.startsWith(uriPrefix));
             assert(log.runs[0].results[0].locations[0].physicalLocation.artifactLocation.uri.endsWith("/" + sourceFilePath1));
             assert.strictEqual(log.runs[0].results[0].locations[0].physicalLocation.artifactLocation.index, 0);
+            assert.isDefined(log.runs[0].results[0].suppressions);
+            assert.lengthOf(log.runs[0].results[0].suppressions, 0);
         });
     });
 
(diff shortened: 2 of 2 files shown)


src/ESLint.Formatter/sarif.js
```diff
@@ -132,9 +132,11 @@ module.exports = function (results, data) {
                     console.log(err);
                 }
             }
-            if (result.messages.length > 0) {
 
-                result.messages.forEach(message => {
+            const messages = result.messages.concat(result.suppressedMessages);
```
jeffersonking, line 136: Is `suppressedMessages` guaranteed to be not `undefined`?

jeffersonking (changes requested):

src/ESLint.Formatter/sarif.js
Yiwei-Ding: Yes. The official doc is [here](https://eslint.org/docs/developer-guide/nodejs-api#-lintresult-type).

[_distinguishSuppressedMessages()](https://github.com/eslint/eslint/blob/6f940c3ce715327f282c197d0f71b91848e5d83d/lib/linter/linter.js#L1830) in [linter](https://github.com/eslint/eslint/blob/main/lib/linter/linter.js) is to preserve the latest suppressed message list, which is initially an empty list instead of `undefined` or `null`. This [comment](https://github.com/eslint/eslint/pull/15459#discussion_r790648587) could also corroborate this.

jeffersonking: Thanks @Yiwei-Ding. Any backwards compatibility concerns? Would our formatter ever receive output from an earlier version of ESLint that does not output `suppressedMessages` property?

Everything else looks great. Thanks for the contribution.

Yiwei-Ding: Thanks for reminding! It's good for me to add incompatibility check.

```js
let messages = result.messages;

if (result.suppressedMessages) {
    messages = messages.concat(result.suppressedMessages);
}

...
```

src/ESLint.Formatter/sarif.js
```diff
@@ -220,6 +222,15 @@ module.exports = function (results, data) {
                         };
                     }
 
+                    if (message.suppressions) {
```
michaelcfanning, line 225: >message.suppressions [](http://example.com/codeflow?start=24&length=20)

An issue to keep in mind: in the case of *any* suppression utilization, SARIF requires that all messages populate this element (with an empty array in cases the result is unsuppressed. Spec text follows.

Many systems only conditionally emit or otherwise process suppression details. The .NET SuppressMessageAttribute is a good example: this in-source data is only produced in the presence of a certain #define at compilation time.

If ESLint *always* provides a suppression mechanism, one choice you have is to always emit an empty array for every result (at the expense of increasing log file size).

Alternately, if there's an efficient way to determine that *no* message in an analysis is suppressed, you can omit this data.,

```
The suppressions values for all result objects in theRun SHALL be either all null or all non-null.
NOTE: The rationale is that an engineering system will generally evaluate all results for suppression, or none of them. Requiring that the suppressions values be either all null or all non-null enables a consumer to determine whether suppression information is available for the run by examining a single result object.
```

Yiwei-Ding: Fixed. Thanks.

src/ESLint.Formatter/sarif.js
jeffersonking: Thanks for addressing that!

jeffersonking (approved):

EasyRhinoMSFT (approved):
:shipit:

michaelcfanning (approved):
:shipit:

michaelcfanning:
Nice improvement! Thanks for the contribution!

scalvert:
Sorry to be late to reviewing this. We've moved the source to the [`sarif-js-sdk`](https://github.com/microsoft/sarif-js-sdk), but I was remiss in removing the source from this repo in favor of that, and republishing from that repo.

We should determine a plan to:

- decommission the source from this repo
- publish a new version from that repo (it can likely just be a patch, since it'll only be changing the repository path

This should not break/interrupt consumers, since the packages will be functionally the same.

PR #2437, opened 2022-02-01
`SarifLogger` now emits an artifacts table entry if `artifactLocation` is not null for tool configuration and tool execution notifications.

eddynaka:
# Description
After we merged this [PR](https://github.com/microsoft/sarif-sdk/pull/2433) when we log a notification, we stopped it from logging to the artifacts. But, this is a bug, since a notification is also a result and that result might have an `artifactLocation` which can point to the artifacts table.

# Test
Let's create notifications with/without locations and check if the artifacts table is being updated.


src/ReleaseHistory.md
```diff
@@ -5,6 +5,7 @@
 * BREAKING: `AnalyzeCommandBase` previously persisted all scan target artifacts to SARIF logs rather than only persisting artifacts referenced by an analysis result, when an option to persist hashes, text file or binary information was set. `MultithreadedAnalyzeCommandBase` previously persisted all scan targets artifacts to SARIF logs in cases when hash insertion was eenabled rather than only persisting artifacts referenced by an analysis result. [#2433](https://github.com/microsoft/sarif-sdk/pull/2433)
 * BUGFIX: Adjust Json Serialization field order for ReportingDescriptor and skip emit empty AutomationDetails node. [#2420](https://github.com/microsoft/sarif-sdk/pull/2420)
 * BREAKING: Fix `InvalidOperationException` when using PropertiesDictionary in a multithreaded application, and remove `[Serializable]` from it. Now use of BinaryFormatter on it will result in `SerializationException`: Type `PropertiesDictionary` is not marked as serializable. [#2415](https://github.com/microsoft/sarif-sdk/pull/2415)
+* BUGFIX: Notifications should emit an artifact if artifactLocation is not null. [#2437](https://github.com/microsoft/sarif-sdk/pull/2437)
```
michaelcfanning, line 8: >Notifications [](http://example.com/codeflow?start=10&length=13)

Please encapsulate API names, etc. in back ticks as necessary/

michaelcfanning, line 8: >Notifications should emit an artifact if artifactLocation is not null [](http://example.com/codeflow?start=10&length=69)

You didn't document what type's behavior changed, it's SarifLogger. You need this information for customers to assess whether the change impacts them.

`SarifLogger` now emits an artifacts table entry is `artifactLocation` is not null for tool configuration and tool execution notifications.

eddynaka: Just updated!
thanks for the suggestion!

michaelcfanning (approved):
:shipit:

PR #2440, opened 2022-02-02
BUGFIX: Properly report `JSON1006.ValueTooLarge` in the `results` node when validating Sarif file with `id` of `location` greater than Int32.MaxValue.

shaopeng-gh:
# Description:

request from GitHub,

Below is the conversation:
Hi gang, do you maintain the sarif validator? One thing that seems to slip through is that the ID field accepts an integer but there's a max limit on the GitHub side, we're asking for the limit to be documented but it might be a small addition to the validator to ensure that the ID value is > max_value. If you maintain it I can circle back once I know the max ID value.

I've attached an example invalid sarif file. On line 85, you'll see:
"id":311971300974507712963699621624531493277,
That number is way larger than an integer. I think the max value depends on the language/compiler, but it looks like the general integer max value is 2,147,483,647.

# Fixes:

1. Because Deserialize() will result in error,
We already have a rule `JSON1006.ValueTooLarge`, the `JSONXXXX` rules runs before we try to Deserialize(),
so it is a good place to do the validate.
The Schema will need to add the max value for this to happen.

2. added tests for when we expect it to success and fail, 
current code does not support the case of fail so a few modification made for it.

### DIFF
--- a/src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
+++ b/src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
@@ -81,7 +81,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
 
         protected override string IntermediateTestFolder => @"Multitool";
 
-        protected override string ConstructTestOutputFromInputResource(string inputResourceName, object parameter)
+        protected override string ConstructTestOutputFromInputResource(string inputResourceName, object parameter, int expectedReturnCode = SUCCESS)
         {
             string testName = parameter as string;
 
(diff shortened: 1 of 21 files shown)


src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
```diff
@@ -81,7 +81,7 @@ private void RunBaselineOptionTest(string testFileName, string testName, bool in
 
         protected override string IntermediateTestFolder => @"Multitool";
 
-        protected override string ConstructTestOutputFromInputResource(string inputResourceName, object parameter)
+        protected override string ConstructTestOutputFromInputResource(string inputResourceName, object parameter, int expectedReturnCode = SUCCESS)
```
shaopeng-gh, line 84: >, int expectedReturnCode = SUCCESS [](http://example.com/codeflow?start=113&length=34)

because I changed the method signature in base, 
so a few other files in this PR the change is just need to keep the same signature since they override it.

shaopeng-gh:
will update and resend PR

PR #2441, opened 2022-02-02
BUGFIX: Properly report `JSON1006.ValueTooLarge` in the `results` node when validating Sarif file with `id` of `location` greater than Int32.MaxValue.

shaopeng-gh:
# Description:

request from GitHub,

Below is the conversation:
Hi gang, do you maintain the sarif validator? One thing that seems to slip through is that the ID field accepts an integer but there's a max limit on the GitHub side, we're asking for the limit to be documented but it might be a small addition to the validator to ensure that the ID value is > max_value. If you maintain it I can circle back once I know the max ID value.

I've attached an example invalid sarif file. On line 85, you'll see:
"id":311971300974507712963699621624531493277,
That number is way larger than an integer. I think the max value depends on the language/compiler, but it looks like the general integer max value is **2,147,483,647**.

# Fixes:

1. Because Deserialize() will result in error,
We already have a rule `JSON1006.ValueTooLarge`, the `JSONXXXX` rules runs before we try to Deserialize(),
so it is a good place to do the validate.
The Schema will need to add the max value for this to happen.

2. Ideally we want to support values [-1, Int32.Max], 
and anything larger than Int32.Max will result in `JSON1006.ValueTooLarge`. 
and smaller than -1 will result in `JSON1008.ValueTooSmall`.

3. added tests


src/ReleaseHistory.md
```diff
@@ -7,6 +7,7 @@
 * BREAKING: Fix `InvalidOperationException` when using PropertiesDictionary in a multithreaded application, and remove `[Serializable]` from it. Now use of BinaryFormatter on it will result in `SerializationException`: Type `PropertiesDictionary` is not marked as serializable. [#2415](https://github.com/microsoft/sarif-sdk/pull/2415)
 * BREAKING: `SarifLogger` now emits an artifacts table entry if `artifactLocation` is not null for tool configuration and tool execution notifications. [#2437](https://github.com/microsoft/sarif-sdk/pull/2437)
 * BUGFIX: Fix `ArgumentException` when `--recurse` is enabled and two file target specifiers generates the same file path. [#2438](https://github.com/microsoft/sarif-sdk/pull/2438)
+* BUGFIX: Properly report `JSON1006.ValueTooLarge` in the `results` node when validating Sarif file with `id` of `location` greater than Int32.MaxValue. [#2441](https://github.com/microsoft/sarif-sdk/pull/2441)
```
eddynaka, line 10: >Sarif [](http://example.com/codeflow?start=89&length=5)

SARIF

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1008.ValueTooSmall_LocationId_Valid.sarif
```diff
@@ -0,0 +1,57 @@
...
+      },
+      "results": [
+        {
+          "ruleId": "TEST1001",
+          "message": {
+            "text": "Locations that specify file paths."
+          },
+          "locations": [
+            {
+              "id": 0,
```
eddynaka, line 21: >0 [](http://example.com/codeflow?start=20&length=1)

add -1, the max allowed value

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
```diff
@@ -0,0 +1,50 @@
...
+          },
+          "locations": [
+            {
+              "id": 2147483648,
+              "physicalLocation": {
+                "artifactLocation": {
+                  "uri": "file1.cs"
+                }
+              }
+            },
```
eddynaka, line 27: remove this

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
```diff
@@ -0,0 +1,50 @@
...
+            {
+              "id": 2147483647000,
```
eddynaka, line 29: >2147483647000 [](http://example.com/codeflow?start=20&length=13)

you don't need to add more than one since the one above is enough to prove ourselves that we are validating it.

src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
```diff
@@ -1292,6 +1292,7 @@
           "description": "Value that distinguishes this location from all other locations within a single result object.",
           "type": "integer",
           "minimum": -1,
+          "maximum": 2147483647,
```
eddynaka, line 1295: >maximum [](http://example.com/codeflow?start=11&length=7)

nit: add the other fields from SARIF that only accepts non-negative integers besides -1, for example, itemCount, offset, etc.

eddynaka: This will prevent us from changing over and over due to a max int value

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1008.ValueTooSmall_LocationId_Valid.sarif
shaopeng-gh: will add the max
I think we should not add the -1;

-1 is the 3rd case in this test, 
"If id is absent, it SHALL default to -1"

and the spec does allow explicit -1:
 "id whose value is a non-negative integer"

[https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317672](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317672)

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
shaopeng-gh: as discussed I keep this one but changed to Int 64 max.

src/ReleaseHistory.md
shaopeng-gh: fixed

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1008.ValueTooSmall_LocationId_Valid.sarif
shaopeng-gh: added the max

src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
shaopeng-gh: agree, I have created a issue to track
[https://github.com/microsoft/sarif-sdk/issues/2442](https://github.com/microsoft/sarif-sdk/issues/2442)

there can be many to go over, 
after we agree with this proposed way of fix in this PR, I can work on it.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
```diff
@@ -0,0 +1,50 @@
...
+                "artifactLocation": {
+                  "uri": "Test.Int32.Max+1.cs"
+                }
+              }
+            },
+            {
+              "id": 9223372036854775807,
+              "physicalLocation": {
+                "artifactLocation": {
+                  "uri": "Test.Int64.Max.cs"
```
shaopeng-gh, line 32: >Test.Int64.Max [](http://example.com/codeflow?start=26&length=14)

I use this as a place to add comments of the current test case.

Want to point out another issue,
In our other repo which is used here to do the JSON validations
[https://github.com/microsoft/jschema/blob/main/src/Json.Schema.Validation/Validator.cs](https://github.com/microsoft/jschema/blob/main/src/Json.Schema.Validation/Validator.cs)
in the whole solution we choose to use type `long` and `double`.
(in this case downsize from `long` to `int`)
This means field value should not exceed the range of `long` and `double`.
technically we can choose to replace `long` and `double` with larger range types,
but not sure if it is necessary,
and whatever range we increase to, user will still be able to add more digits to 
exceed the range, unless we make fields as string we can not prevent that.
I think the code choose to use `long` and `double` was a design choice not a bug, let me know.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
```diff
@@ -0,0 +1,104 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
```
michaelcfanning, line 6: >tool [](http://example.com/codeflow?start=7&length=4)

It is very inefficient to add 8 log files to our tests for this specific bug fix. Please create a pattern in the unit tests for validating this problem.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
```diff
@@ -0,0 +1,104 @@
...
+        "driver": {
+          "name": "SARIF Functional Testing",
+          "rules": [
+            {
+              "id": "JSON1006",
```
michaelcfanning, line 11: >JSON1006 [](http://example.com/codeflow?start=21&length=8)

The rule factoring here is not wise. Why is it important to have one rule to detect something is too large and another rule for something that is too small?  Does this really help the user? Would they, for example, need two different help topics to clarify this distinction?

We should have created something like 'ValueOutOfRange' and collapsed all this analysis into it. 

Please open an issue for this improvement.

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -47,6 +47,22 @@
         public void JSON1002_DeserializationError()
             => RunTest("JSON1002.DeserializationError.sarif");
 
+        [Fact]
+        public void JSON1006_ValueTooLarge_LocationId_Valid()
```
michaelcfanning, line 51: >JSON1006_ValueTooLarge_LocationId_Valid [](http://example.com/codeflow?start=20&length=39)

This is too much test infrastructure for the work that you are doing. Are we going to have hundreds or thousands of test files in order to test every possible numeric value in the format? 

Please just create at most two test files, one for reach rule.

Please don't qualify your test name by Location id, is this the only numeric value we are testing? If so, why are we only testing the specific corner case we have found here?

As always, please don't just focus on the specific issue that's been flagged. Please don't just accept the code design patterns as you've found them and make things more complicated.

Instead, look for efficient analysis and testing that comprehensively addresses the problem that's been found.

michaelcfanning:
>             catch (JsonSerializationException)

Why are we swallowing these exceptions?

---
Refers to: src/Sarif.Multitool.Library/ValidateCommand.cs:87 in bf846f0. [](commit_id = bf846f0556b1cae1a76021432a7fdd58cef9f3aa, deletion_comment = False)

src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
michaelcfanning: You can't just update this schema in place, it has to be versioned, this is an official output of the SARIF technical committee. 

Why have you chosen this max value? Does it correspond to what the SARIF spec allows?

michaelcfanning (changes requested):
:clock1:

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
shaopeng-gh: agree, created an issue in jschema repo:
[https://github.com/microsoft/jschema/issues/142](https://github.com/microsoft/jschema/issues/142)

shaopeng-gh:
>             catch (JsonSerializationException)

I see we are swallowing these exceptions 
so that it will only show in the result node, 
instead of both in result node + tool error (ERR999.UnhandledEngineException).

talked to Eddy we don't want to show tool error as well, 
so I added similar swallowing exceptions code under it.


---
In reply to: [1029154865](https://github.com/microsoft/sarif-sdk/pull/2441#issuecomment-1029154865) [](http://example.com/codeflow?ancestors=1029154865,1029154865)

---
Refers to: src/Sarif.Multitool.Library/ValidateCommand.cs:87 in bf846f0. [](commit_id = bf846f0556b1cae1a76021432a7fdd58cef9f3aa, deletion_comment = False)

src/Sarif/Schemata/sarif-2.1.0-rtm.5.json
shaopeng-gh: Why have you chosen this max value? Does it correspond to what the SARIF spec allows?  
------ the spec just say `non-negative integer`.
[https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317672](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317672)

I don't see in the spec, but do we have definition of the term `integer` or is it always assume to be 32 bit,
I see our C# code object model all use `int` which is 32 bit:
public virtual `int` Id { get; set; }
looks like we should have defined some where,

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
shaopeng-gh: Please just create at most two test files, one for each rule.
----- yes, I created a file for valid and a file for invalid, and in each file there can be more than 1 case, I will merge the valid and invalid to 1 file.

Please don't qualify your test name by Location id, is this the only numeric value we are testing?  ---- It is, I only added the max for Location Id in the Definition file, so I can only test for it. 
I should also look for other fields those have similar issue and add the max attribute,
**But I see your other comments, is the way I am adding the max attribute the right way to go? Need some guidance here.**

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/JSON1006.ValueTooLarge_LocationId_Invalid.sarif
shaopeng-gh: yes, will change to only one 1 input file for each rule according to your other comment

shaopeng-gh:
Start another one trying Big Integer

PR #2459, opened 2022-02-18
Revert simple Join-Path syntax

shaopeng-gh:
# Description:

I did noticed `CodeQL` task does not recognize simplified syntax and report as fail,
but there is also a `.compliance` task will fail, look like because of this `CodeQL` fail.
Suggest just revert that simplified syntax. 

image

It looks that that task is running in old version of PowerShell:
image

# Fixes:

Revert changes made for simplified syntax of `Join-Path`.


michaelcfanning (approved):
:shipit:

PR #2460, opened 2022-02-24
Updating jschema version

eddynaka:
This change is updating the jschema packages to be compatible with the UNIX environments.
Previously, we were using an unsupported version (netcoreapp2.0).


michaelcfanning (approved):
:shipit:

Issue #2457, opened 2022-02-16
MultithreadAnalyzeCommandBase is throwing exception when path provided in target specifiers does not exist


eddynaka:
Hi,

If we use `MultithreadAnalyzeCommandBase`, and pass a path that does not exist in the targets specifiers, it will throw an exception:
```
error ERR999.UnhandledEngineException : System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\__w\1\CPU\Release\net472'. 
at System.IO.Enumeration.FileSystemEnumerator`1.CreateDirectoryHandle(String path, Boolean ignoreNotFound) 
at System.IO.Enumeration.FileSystemEnumerator`1.Init() 
at System.IO.Enumeration.FileSystemEnumerator`1..ctor(String directory, Boolean isNormalized, EnumerationOptions options) 
at System.IO.Enumeration.FileSystemEnumerable`1..ctor(String directory, FindTransform transform, EnumerationOptions options, Boolean isNormalized) 
at System.IO.Enumeration.FileSystemEnumerableFactory.UserFiles(String directory, String expression, EnumerationOptions options) 
at System.IO.Directory.InternalEnumeratePaths(String path, String searchPattern, SearchTarget searchTarget, EnumerationOptions options) 
at System.IO.Directory.EnumerateFiles(String path, String searchPattern, SearchOption searchOption) 
at Microsoft.CodeAnalysis.Sarif.FileSystem.DirectoryEnumerateFiles(String path, String searchPattern, SearchOption searchOption) in /_/src/Sarif/FileSystem.cs:line 370 
at Microsoft.CodeAnalysis.Sarif.Driver.MultithreadedAnalyzeCommandBase`2.FindFilesAsync(TOptions options, TContext rootContext) in /_/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs:line 397 
```

This does not happen with `AnalyzeCommandBase` because we validate before navigating:
https://github.com/microsoft/sarif-sdk/blob/ea44355584c161aec784ee9d2aa3c57f84875906/src/Sarif.Driver/FileSpecifier.cs#L70

Should we keep the same behavior as before?

FYI: @michaelcfanning 
@marmegh , this issue is happening sometimes at BinSkim. They can fix it by using the correct paths. But this is a breaking change.

PR #2461, opened 2022-02-25
Handle non-existent directories the same for Multi/Single-threaded analysis

marmegh:
Updated multithreaded analysis to incorporate a directory validation as already implemented for single-threaded analysis. Also, added a test to prevent regressions for either. This is to account for any customers using sarif that may be dependent on the single threaded implementation. This is to address issue #2457.


eddynaka:
@marmegh , what is the status of this PR?

src/ReleaseHistory.md
```diff
@@ -9,6 +9,7 @@
 * BUGFIX: Fix `ArgumentException` when `--recurse` is enabled and two file target specifiers generates the same file path. [#2438](https://github.com/microsoft/sarif-sdk/pull/2438)
 * BUGFIX: Fix 'InvalidOperationException' with message `Collection was modified; enumeration operation may not execute` in `MultithreadedAnalyzeCommandBase`, which is raised when analyzing with the `--hashes` switch. [#2447](https://github.com/microsoft/sarif-sdk/pull/2447)
 * BUGFIX: Fix `Merge` command produces empty SARIF file in Linux when providing file name only without path. [#2408](https://github.com/microsoft/sarif-sdk/pull/2408)
+* BUGFIX: Fix `UnhandledEngineException` when target path does not exist for multithreaded application by validating directories as is done for singlethreaded analysis. []()
```
michaelcfanning, line 12: >UnhandledEngineException [](http://example.com/codeflow?start=15&length=24)

This change looks good but you'll need to update your release notes, since we shipped v.2.14

eddynaka (approved):
:shipit:

PR #2463, opened 2022-03-02
BREAKING: `Id` field of `Location` changed from `int`(32bit) to `BigInteger`(unlimited)

shaopeng-gh:
# Description:

We found that in SARIF spec, there is no definition of the max length of integer type, whether it is 32 bit, 64 bit or no limit.
Our current SDK code assumes 32 bit, which should be wrong assumption, should be no limit since spec does not mention any limit.

# Fixes:

Spot fix the `Location.Id` to use `BigInteger` which is no limit.

Note that we should also change the jschema to use `BigInteger` across the solution. created a issue there:
https://github.com/microsoft/jschema/issues/144


src/Sarif/NotYetAutoGenerated/Location.cs
```diff
@@ -0,0 +1,237 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// A location within a programming artifact.
+    /// </summary>
+    [DataContract]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "1.1.0.0")]
+    public partial class Location : PropertyBagHolder, ISarifNode
```
shaopeng-gh, line 20: >Location [](http://example.com/codeflow?start=25&length=8)

created a copy of the Location.cs to 
NotYetAutoGenerated folder to make modification.

src/Sarif/Autogenerated/Location.cs
```diff
@@ -40,7 +41,12 @@ public virtual SarifNodeKind SarifNodeKind
         [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(-1)]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public virtual int Id { get; set; }
+        public virtual BigInteger Id { get; set; }
+
+        public bool ShouldSerializeId()
+        {
+            return Id != -1;
+        }
```
shaopeng-gh, line 49: this is needed because DefaultValue does not work for BigInteger.

then this need 2 things,
1. in the constructor to set it, which we already have.
2. ShouldSerializeXXX to handle the serialization.

src/ReleaseHistory.md
```diff
@@ -11,6 +11,7 @@
 * BUGFIX: Fix `Merge` command produces empty SARIF file in Linux when providing file name only without path. [#2408](https://github.com/microsoft/sarif-sdk/pull/2408)
 * BUGFIX: Fix `NullReferenceException` when filing work item with a SARIF file which has no filable results. [#2412](https://github.com/microsoft/sarif-sdk/pull/2412)
 * BUGFIX: Fix missing `endLine` and `endColumn` properties and remove vulnerable packages for ESLint SARIF formatter. [#2458](https://github.com/microsoft/sarif-sdk/pull/2458)
+* BREAKING: `Id` field of `Location` changed from `int`(32bit) to `BigInteger`(unlimited) [#2463](https://github.com/microsoft/sarif-sdk/pull/2463)
```
michaelcfanning, line 14: >field [](http://example.com/codeflow?start=17&length=5)

use 'property' not 'field' as a term. Add something about the exception message you see on deserializing a id value larger than 2^31

src/Sarif/Autogenerated/Location.cs
```diff
@@ -40,7 +41,12 @@ public virtual SarifNodeKind SarifNodeKind
         [DataMember(Name = "id", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(-1)]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public virtual int Id { get; set; }
+        public virtual BigInteger Id { get; set; }
+
+        public bool ShouldSerializeId()
+        {
+            return Id != -1;
```
michaelcfanning, line 48: >return Id != -1 [](http://example.com/codeflow?start=12&length=15)

return Id >-1;

michaelcfanning: Please add a unit test for this.

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -85,5 +92,68 @@ public void Location_LogicalLocation_WhenSetToNull_RemovesLogicalLocationsArray(
...
+            location.Id = 0;
+            AssertShouldSerializeId(location, true);
+
+            location.Id = 1;
+            AssertShouldSerializeId(location, true);
+
+            location.Id = int.MaxValue;
+            AssertShouldSerializeId(location, true);
+
+            location.Id = long.MaxValue;
```
michaelcfanning, line 114: >long [](http://example.com/codeflow?start=26&length=4)

add ulong max value

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -85,5 +92,68 @@ public void Location_LogicalLocation_WhenSetToNull_RemovesLogicalLocationsArray(
...
         }
+
+        [Fact]
+        public void Location_SerializeId()
+        {
+            var location = new Location();
+            AssertShouldSerializeId(location, false);
+
+            location.Id = -1;
+            AssertShouldSerializeId(location, false);
```
michaelcfanning, line 103: add test for long.MinValue - 1
add test for long.MinValue
add test for int.MinValue
add test for -2

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -85,5 +92,68 @@ public void Location_LogicalLocation_WhenSetToNull_RemovesLogicalLocationsArray(
 
             location.LogicalLocations.Should().BeNull();
         }
+
+        [Fact]
+        public void Location_SerializeId()
```
michaelcfanning, line 97: >Location_SerializeId [](http://example.com/codeflow?start=20&length=20)

These tests may not provoke round-tripping issues?

michaelcfanning (approved):
:shipit:

src/Sarif/NotYetAutoGenerated/Location.cs
shaopeng-gh: added

src/Test.UnitTests.Sarif/Core/LocationTests.cs
shaopeng-gh: added all

src/Test.UnitTests.Sarif/Core/LocationTests.cs
shaopeng-gh: change to round trip

src/ReleaseHistory.md
shaopeng-gh: updated

src/ReleaseHistory.md
```diff
@@ -1,12 +1,14 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
 ## UNRELEASED
+
 * BUGFIX: Eliminate dispose of stream and `StreamWriter` arguments passed to `SarifLog.Save` helpers. This would result in `ObjectDisposedException` being raised on attempt to access streams after save.
+* BREAKING: `Id` property of `Location` changed from `int`(32bit) to `BigInteger`(unlimited) to fix `Newtonsoft.Json.JsonReaderException: JSON integer XXXXX is too large or small for an Int32.` [#2463](https://github.com/microsoft/sarif-sdk/pull/2463)
```
shaopeng-gh, line 6: removed the very first part "error ERR999.UnhandledEngineException" which only apply to the SDK validate tool.

the rest error message apply to both SDK validate tool and directly using:
`JsonConvert.DeserializeObject<SarifLog>(content)`

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -85,5 +92,160 @@ public void Location_LogicalLocation_WhenSetToNull_RemovesLogicalLocationsArray(
...
+            VerifyIdRoundTripFromJsonHelper(jsonLocation, deserializedId: ulong.MaxValue,
+                shouldSerialize: true, reconstructedLocationId: ulong.MaxValue);
+
+            jsonLocation = $"{{\"id\":{new BigInteger(ulong.MaxValue) + 1}}}";
+            VerifyIdRoundTripFromJsonHelper(jsonLocation, deserializedId: new BigInteger(ulong.MaxValue) + 1,
+                shouldSerialize: true, reconstructedLocationId: new BigInteger(ulong.MaxValue) + 1);
+        }
+
+        [Fact]
+        public void Location_VerifyAbleToDeserializeWithBigIntegerId()
```
shaopeng-gh, line 212: >public void Location_VerifyAbleToDeserializeWithBigIntegerId() [](http://example.com/codeflow?start=8&length=62)

added a new test for `JsonConvert.DeserializeObject<SarifLog>(content)`
from a SARIF file.

src/Test.UnitTests.Sarif/TestData/JsonConverters/LocationId_BigInteger.sarif
```diff
@@ -0,0 +1,148 @@
...
+      ],
+      "results": [
+        {
+          "message": {
+            "text": "This app does not have capabilities to prevent tapjacking attacks. An attacker can hijack the user's taps and tricks him into performing some critical operations that he did not intend to."
+          },
+          "level": "note",
+          "locations": [
+            {
+              "id": 31197130097450771296369962162453149327732752356239421572342053257324632475324,
```
shaopeng-gh, line 33: >31197130097450771296369962162453149327732752356239421572342053257324632475324 [](http://example.com/codeflow?start=20&length=77)

data in this file contains very long int, 64 int max +1, 32 int max +1 and a few normal ones.

PR #2468, opened 2022-03-14
Update Code generation `SerializedPropertyInfo`

yongyan-gh:
# Description

Currently there is an issue in JSchema EqualityComparerGenerator that doesn't correctly handle comparison of data model type element of n dictionary. A patch fix was introduced in [src/Sarif/NotYetAutoGenerated/ReportingDescriptorEqualityComparer.cs](https://github.com/microsoft/sarif-sdk/blob/main/src/Sarif/NotYetAutoGenerated/ReportingDescriptorEqualityComparer.cs)

https://github.com/microsoft/sarif-sdk/blob/b2c7e399cfaf1fde47b336bbb4976f723263efa3/src/Sarif/NotYetAutoGenerated/ReportingDescriptorEqualityComparer.cs#L278-L287

Already have a fix in JSchema for this issue. To make it work the way to generate codes related to `SerializedPropertyInfo` needs to be the same way as other Sarif entities.

Sarif.SDK uses JSchema to auto generate SARIF entities using the information in CodeGenHints.json.

`SerializedPropertyInfo` is the type of property bags (`IDictionary<string, SerializedPropertyInfo>`) of all SARIF entities. It's not auto generated by JSchema, its defined in Sarif project, and JSchema generates the code reference to `SerializedPropertyInfo` using the information in CodeGenHints.json.

The change is to make the codeGenHints configurations for 'properties', the dictionary of `SerializedPropertyInfo`, to be the same way as all other dictionaries of Sarif entities.

# Changes

Updated the configurations for 'properties in CodeGenHints.json to correct values, same as all other Sarif entities:
- `comparisonKind` from `ObjectEquals` to `EqualityComparerEquals`.
- `hashKind` from `ScalarReferenceType` to `ObjectModelType`

Added `SerializedPropertyInfoComparer : IComparer<SerializedPropertyInfo>` so that it will be ready for auto generated comparer for all Sarif entities.

Added unit tests.

### DIFF
--- a/src/Sarif/Core/SerializedPropertyInfoEqualityComparer.cs
+++ b/src/Sarif/Core/SerializedPropertyInfoEqualityComparer.cs
@@ -24,17 +24,12 @@ namespace Microsoft.CodeAnalysis.Sarif
                 return false;
             }
 
-            return left.SerializedValue.Equals(right.SerializedValue);
+            return left.Equals(right);
         }
 
         public int GetHashCode(SerializedPropertyInfo serializedPropertyInfo)
         {
-            if (serializedPropertyInfo.SerializedValue == null)
-            {
-                return 0;
-            }
-
-            return serializedPropertyInfo.SerializedValue.GetHashCode();
+            return serializedPropertyInfo.GetHashCode();
         }
     }
 }
--- a/src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
+++ b/src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
@@ -1,0 +1,171 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+using FluentAssertions;
+
+using Microsoft.CodeAnalysis.Sarif;
+
+using Newtonsoft.Json;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Core
+{
+    public class SerializedPropertyInfoTests
+    {
+        [Fact]
+        public void SerializedPropertyInfo_ComparerTests()
+        {
+            Guid testGuid = Guid.NewGuid();
+            DateTime now = DateTime.UtcNow;
+
+            var testCases = new[]
+            {
+                new
+                {
+                    Left = new SerializedPropertyInfo(null, isString:false),
+                    Right = new SerializedPropertyInfo(null, isString:false),
+                    ExpectedEqualsResult = true,
+                    ExpectedCompareResult = 0
+                },
+                new
+                {
+                    Left = new SerializedPropertyInfo(null, isString:false),
+                    Right = new SerializedPropertyInfo(null, isString:true),
+                    ExpectedEqualsResult = false,
+                    ExpectedCompareResult = -1
+                },
+                new
+                {
+                    Left = new SerializedPropertyInfo(JsonConvert.SerializeObject(true), isString:false),
+                    Right = new SerializedPropertyInfo(JsonConvert.SerializeObject(true), isString:false),
+                    ExpectedEqualsResult = true,
+                    ExpectedCompareResult = 0
+                },
+                new
+                {
+                    Left = new SerializedPropertyInfo(JsonConvert.SerializeObject(true), isString:false),
+                    Right = new SerializedPropertyInfo(JsonConvert.SerializeObject(false), isString:false),
+                    ExpectedEqualsResult = false,
+                    ExpectedCompareResult = 1
+                },
+                new
+                {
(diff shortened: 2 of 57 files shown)


src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                new
+                {
+                    Left = new SerializedPropertyInfo(JsonConvert.SerializeObject(true), isString:false),
+                    Right = new SerializedPropertyInfo(JsonConvert.SerializeObject(false), isString:false),
+                    ExpectedEqualsResult = false,
+                    ExpectedCompareResult = 1
+                },
+                new
+                {
+                    Left = new SerializedPropertyInfo(JsonConvert.SerializeObject(10), isString:false),
```
shaopeng-gh, line 56: >new SerializedPropertyInfo(JsonConvert.SerializeObject(10), isString:false), [](http://example.com/codeflow?start=27&length=76)

some combination of below that you can consider:

new SerializedPropertyInfo(JsonConvert.SerializeObject(10), isString:false),
new SerializedPropertyInfo(JsonConvert.SerializeObject(10), isString:true),
new SerializedPropertyInfo(JsonConvert.SerializeObject("10"), isString:false),
new SerializedPropertyInfo(JsonConvert.SerializeObject("10"), isString:true),
new SerializedPropertyInfo(JsonConvert.SerializeObject("\"10\""), isString:false),
new SerializedPropertyInfo(JsonConvert.SerializeObject("\"10\""), isString:true),

And if user use SetProperty() instead, will there be problem

src/Sarif/Core/SerializedPropertyInfoEqualityComparer.cs
```diff
@@ -24,17 +24,12 @@ public bool Equals(SerializedPropertyInfo left, SerializedPropertyInfo right)
...
+            return left.Equals(right);
         }
 
         public int GetHashCode(SerializedPropertyInfo serializedPropertyInfo)
         {
-            if (serializedPropertyInfo.SerializedValue == null)
-            {
-                return 0;
-            }
-
```
shaopeng-gh, line 36: will the logic change after we remove this, do we still return same 0 when it is null

src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
```diff
@@ -0,0 +1,171 @@
...
+                new
+                {
+                    Left = new SerializedPropertyInfo(JsonConvert.SerializeObject(new[] { "a", "b", "c" }), isString:false),
+                    Right = new SerializedPropertyInfo(JsonConvert.SerializeObject(new[] { "c", "b", "a" }), isString:false),
+                    ExpectedEqualsResult = false,
+                    ExpectedCompareResult = -1
+                },
+                new
+                {
+                    // serialzed string: "{\"a\":1,\"b\":true}"
```
shaopeng-gh, line 135: >// serialzed string: "{\"a\":1,\"b\":true}" [](http://example.com/codeflow?start=20&length=43)

by "addtional info that is not in the code" I think these comments can be removed

src/Sarif/Core/SerializedPropertyInfoEqualityComparer.cs
yongyan-gh: The `SerializedPropertyInfoEqualityComparer.GetHashCode()` was not used, `serializedPropertyInfo.GetHashCode()` is the one been called. Here the change is to make SerializedPropertyInfoEqualityComparer to point to right code.

src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        public void SerializedPropertyInfo_ComparerTests()
+        {
+            Guid testGuid = Guid.NewGuid();
+            DateTime now = DateTime.UtcNow;
+
+            var testCases = new[]
+            {
+                new
+                {
+                    Left = new SerializedPropertyInfo(null, isString:false),
```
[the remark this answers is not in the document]
yongyan-gh: serialized value of below cases:
input: null -> serializedValue: null
input: string.Empty -> serializedValue: string.Empty
intput: JsonConvert.SerializeObject(null) -> serializedValue: "null"
intput: JsonConvert.SerializeObject(string.Empty) -> serializedValue: "\"\""

added test cases

src/Test.UnitTests.Sarif/Core/SerializedPropertyInfoTests.cs
yongyan-gh: added the test cases

shaopeng-gh (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #2470, opened 2022-03-22
Create codeql-js-analysis.yml

eddynaka:


EasyRhinoMSFT (approved):
:shipit:

michaelcfanning (approved):
:shipit:

.github/workflows/codeql-js-analysis.yml
```diff
@@ -0,0 +1,39 @@
+name: "CodeQL JS"
+
+on:
+  push:
+    branches: "**"
```
marmegh, line 5: >"**" [](http://example.com/codeflow?start=14&length=4)

Did you validate this change in the IDE? I expect there to be a type mismatch exception with branches because this should be an array and not a string.

eddynaka: its not an ide thing. its gh thing.
it can be a string or an array. both works :)

PR #2471, opened 2022-03-24
CodeQL Fix "This cast is redundant because the expression already has type"

shaopeng-gh:
# Description:

CodeQL reports "This cast is redundant because the expression already has type"

# Fixes:

remove the redundant cast

### DIFF
--- a/src/Sarif/PropertiesDictionaryExtensionMethods.cs
+++ b/src/Sarif/PropertiesDictionaryExtensionMethods.cs
@@ -98,7 +98,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 IDictionary pb = property as IDictionary;
                 if (pb != null)
                 {
-                    ((IDictionary)pb).SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
+                    pb.SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
                     continue;
                 }
 
--- a/src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
+++ b/src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
@@ -100,7 +100,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             }
 
             ITestAnalyzeCommand command = multithreaded
-                ? (ITestAnalyzeCommand)new TestMultithreadedAnalyzeCommand()
+                ? new TestMultithreadedAnalyzeCommand()
                 : (ITestAnalyzeCommand)new TestAnalyzeCommand();
 
             command.DefaultPluginAssemblies = plugInAssemblies;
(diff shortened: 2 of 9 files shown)


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -100,7 +100,7 @@ public void AnalyzeCommandBase_RootContextIsDisposed()
             }
 
             ITestAnalyzeCommand command = multithreaded
-                ? (ITestAnalyzeCommand)new TestMultithreadedAnalyzeCommand()
+                ? new TestMultithreadedAnalyzeCommand()
                 : (ITestAnalyzeCommand)new TestAnalyzeCommand();
```
eddynaka, line 104: >(ITestAnalyzeCommand)new TestAnalyzeCommand() [](http://example.com/codeflow?start=18&length=45)

cant you apply the same in this line?

shaopeng-gh: That one can not be removed, will have compile error:

Error CS8957 
Conditional expression is not valid in language version 8.0 because a common type was not found between 'TestMultithreadedAnalyzeCommand' and 'TestAnalyzeCommand'. 
To use a target-typed conversion, upgrade to language version 9.0 or greater. 
Test.UnitTests.Sarif.Driver (netcoreapp3.1) C:\GH\s-s\src\Test.UnitTests.Sarif.Driver\Sdk\AnalyzeCommandBaseTests.cs 102

src/Sarif/PropertiesDictionaryExtensionMethods.cs
```diff
@@ -98,7 +98,7 @@ public static class PropertiesDictionaryExtensionMethods
                 IDictionary pb = property as IDictionary;
                 if (pb != null)
                 {
-                    ((IDictionary)pb).SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
+                    pb.SavePropertiesToXmlStream(writer, settings, key, settingNameToDescriptionMap);
```
michaelcfanning, line 101: >SavePropertiesToXmlStream [](http://example.com/codeflow?start=23&length=25)

Some of the code changes clearly have no potential for runtime impact and we therefore have a high degree of confidence in taking them. This is one because pb is explicitly typed as IDictionary (and there are others in your change). What we need to look for are situations where someone may have overloaded a cast operator, or where a change you made may have changed precedence ordering/evaluation.

One thing you should do here is to generate code coverage and see where we have coverage. If we do not have coverage in a place where the change looks like it has potential to change tool behavior, we need new tests.

You should always describe why you are not submitting a test, if you submit a pull request without them. @eddynake, your help providing this feedback in a review context would be appreciated as well.

michaelcfanning (approved):
:shipit:

src/Sarif/PropertiesDictionaryExtensionMethods.cs
michaelcfanning: OK, I reviewed this carefully and have convinced myself there are no behavioral implications of this change, so I won't block on testing. Please do bear in mind what I said: if you have no tests, please tell the reviewers why you believe they are not required. Core expectation.

shaopeng-gh: Agree with the inputs here, sorry did not mention anything.
Just added in the top description why tests are not authored together with the change.

PR #2473, opened 2022-04-05
Test infrastructure cleanup preparatory to making `result.level` nullable.

michaelcfanning:
This work is test infrastructure clean-up preparatory to resolving #2024, #2383 and #2077 (making result level nullable to satisfy the SARIF spec). As part of looking at a fix for this complex issue, I observed that our test code is messy in SARIF-SDK, with lots of organically evolved patterns and utilities for various test suites.

So this change is a time-boxed investment in cleaning up tests to make them more standard.

@marmegh @EasyRhinoMSFT @yongyan-gh @shaopeng-gh @eddynaka @cfaucon @suvamM


src/Test.EndToEnd.Baselining/Test.EndToEnd.Baselining.csproj
```diff
@@ -14,5 +14,6 @@
     <OutputType>Exe</OutputType>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <IsPackable>false</IsPackable>
+    <RootNamespace>Test.EndToEnd.Baselining</RootNamespace>
```
michaelcfanning, line 17: >RootNamespace [](http://example.com/codeflow?start=5&length=13)

The root namespace is used as a common prefix by the resource compiler. And so we create a root namespace that matches the binary name. This makes it very easy at runtime to locate embedded resources, etc., we simply look at the Assembly binary name to do that.

There is an unpleasant consequence of this 'trick', whenever you create a new file, this will be used as the default namespace (and it never should be retained). So that's an inconvenience for sure.

VS has some knobs for controlling the resource manifest name, e.g., 'Custom Tool Namespace' in the properties pane. This approach can be tedious to use and VS wants to get involved. I haven't found a clean approach. If someone else knows more than I, let me know. :)

src/Test.FunctionalTests.Sarif/Test.FunctionalTests.Sarif.csproj
```diff
@@ -129,6 +130,41 @@
     <Content Include="v1\ConverterTestData\StaticDriverVerifier\sdv-harness.c" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Update="v1\ConverterTestData\StaticDriverVerifier\cancelspinlock_bug1.cs">
```
michaelcfanning, line 133: >Compile [](http://example.com/codeflow?start=5&length=7)

Honestly not sure why this code was updated. :)

src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
```diff
@@ -87,18 +91,18 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
             string testName = parameter as string;
 
             string fileToBeValidated = "ToBeValidated.sarif";
-            string filePathToBeValidated = Path.Combine(this.OutputFolderPath, fileToBeValidated);
-            string logText = GetResourceText($"Inputs.{fileToBeValidated}");
```
michaelcfanning, line 91: >GetResourceText [](http://example.com/codeflow?start=29&length=15)

Previously, we would explicitly retrieve a resource knowing that it lived in disk in a directory named 'Inputs' by qualifying a file name with 'Inputs.'. This matches the resource manifest path created from the directory name. 

Now we ask explicitly for an 'input' file (there is also special knowledge of the 'expected output' files. This will drive us to consistency across tests, i.e., storing files in consistent directory names.

Note that by redacting 'Inputs.' from the file name passed to the actual SDK API, we change resulting URLs for the fake file names in checked in SARIF content. This results in many changes in the TestData directory, later.

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -465,15 +465,15 @@ private string MakeTestFileName(ReportingDescriptor rule, string testFileNameSuf
...
-            string inputLogDirectory = this.OutputFolderPath;
+            string inputLogDirectory = this.TestOutputDirectory;
             string inputLogFileName = Path.GetFileName(inputResourceName);
-            string inputLogFilePath = Path.Combine(this.OutputFolderPath, inputLogFileName);
+            string inputLogFilePath = Path.Combine(this.TestOutputDirectory, inputLogFileName);
 
             string actualLogFilePath = Guid.NewGuid().ToString();
 
-            string ruleUnderTest = Path.GetFileNameWithoutExtension(inputLogFilePath).Split('.')[1];
+            string ruleUnderTest = Path.GetFileNameWithoutExtension(inputLogFilePath).Split('.')[0];
```
michaelcfanning, line 476: >0 [](http://example.com/codeflow?start=97&length=1)

We no longer have the 'Inputs.' prefix in the file names.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/GH1001.ProvideRequiredLocationProperties_Invalid.sarif
```diff
@@ -41,7 +41,7 @@
       "artifacts": [
         {
           "location": {
-            "uri": "FunctionalTestOutput.ValidateCommand/Inputs.GH1001.ProvideRequiredLocationProperties_Invalid.sarif",
+            "uri": "FunctionalTestOutput.ValidateCommand/GH1001.ProvideRequiredLocationProperties_Invalid.sarif",
```
michaelcfanning, line 44: >GH1001 [](http://example.com/codeflow?start=57&length=6)

Here's the change mentioned earlier: we now elide 'Inputs.' from the manufactured test file name.

I have manually checked all the modified files in this directory to confirm that this is the sole change for each of them. (By all means feel free to double-check my work).

Note how much useful information is encoded in this URL about our test scenario, rather than being something like, 'foo.sarif', or 'mytest.sarif'.

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
```diff
@@ -465,15 +465,20 @@ private string MakeTestFileName(ReportingDescriptor rule, string testFileNameSuf
...
-            string inputLogDirectory = this.OutputFolderPath;
+            string inputLogDirectory = this.TestOutputDirectory;
             string inputLogFileName = Path.GetFileName(inputResourceName);
-            string inputLogFilePath = Path.Combine(this.OutputFolderPath, inputLogFileName);
+            string inputLogFilePath = Path.Combine(this.TestOutputDirectory, inputLogFileName);
 
             string actualLogFilePath = Guid.NewGuid().ToString();
 
-            string ruleUnderTest = Path.GetFileNameWithoutExtension(inputLogFilePath).Split('.')[1];
+            // Splits a test file name such as 'GH1001.ProvideRequiredLocationProperties.cs'
```
michaelcfanning, line 476: >Splits [](http://example.com/codeflow?start=15&length=6)

Please note that I've added a good comment to the code explaining this mysterious parse and retrieval of the 0th token. T

src/Test.UnitTests.Sarif/JsonTests.cs
```diff
@@ -18,8 +18,6 @@ public abstract class JsonTests
         protected static readonly Tool DefaultTool = new Tool() { Driver = new ToolComponent { Name = "DefaultTool" } };
         protected static readonly Result DefaultResult = new Result { Message = new Message { Text = "Some testing occurred." } };
 
-        protected static readonly ResourceExtractor ResourceExtractor = new ResourceExtractor(typeof(JsonTests));
```
michaelcfanning, line 21: >ResourceExtractor [](http://example.com/codeflow?start=52&length=17)

I removed this in preference of cleaning up the single derived type that needs test extraction.

src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
```diff
@@ -23,6 +23,10 @@ public class BaselineOptionTests : FileDiffingFunctionalTests
     {
         private bool IsInline = false;
 
+        public static readonly TestAssetResourceExtractor BaselineNamespaceExtractor =
```
michaelcfanning, line 26: >BaselineNamespaceExtractor [](http://example.com/codeflow?start=58&length=26)

Basic pattern, a test can allocate a 'test resource extractor' that retrieves test data from the resource table. We work against the resource table because it minimizes disk activity and improves tests performance. Probably, a standard test resource extractor could be allocated by a test base class, but that's a refactoring for another day, this change is already huge.

src/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+  </packageSources>
+</configuration>
```
shaopeng-gh, line 6: This file was removed in other changeset likely a merge issue, 
should we removed this file under src and leave only one at the root?

src/Sarif/Autogenerated/ReportingDescriptorEqualityComparer.cs
```diff
@@ -190,7 +190,7 @@ public bool Equals(ReportingDescriptor left, ReportingDescriptor right)
                         return false;
                     }
 
-                    if (!object.Equals(value_2.Value, value_3))
+                    if (!SerializedPropertyInfo.ValueComparer.Equals(value_2.Value, value_3))
```
shaopeng-gh, line 193: >if (!SerializedPropertyInfo.ValueComparer.Equals(value_2.Value, value_3)) [](http://example.com/codeflow?start=20&length=73)

just a note that when we change the NotYetAutoGenerated one we should always also change and checkin the AutoGenerated one to keep them in sync.

I just did a comparison of all currently what we have, this file is the only one not in sync.

NuGet.Config
```diff
@@ -1,11 +1,6 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
```
shaopeng-gh, line 4: > <clear /> [](http://example.com/codeflow?start=3&length=10)

the <clear/> is needed for security reasons.

NuGet.Config - Missing <clear/> statement.

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
shaopeng-gh, line 476: >.cs [](http://example.com/codeflow?start=88&length=3)

should this example be changed to .sarif

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
```diff
@@ -108,7 +122,17 @@ public void SarifLog_PropertyBagProperties_CanBeRoundTripped()
 
             string roundTrippedContents = JsonConvert.SerializeObject(log, Formatting.Indented);
 
-            roundTrippedContents.Should().Be(originalContents);
+            roundTrippedContents.Should().Be(sarifLogText);
+
+            RunTestHelper(testFileName);
+        }
```
shaopeng-gh, line 128: >} [](http://example.com/codeflow?start=8&length=1)

nit: add a blank line in between

src/Test.UnitTests.Sarif/RoundTrippingTests.cs
```diff
@@ -108,7 +122,17 @@ public void SarifLog_PropertyBagProperties_CanBeRoundTripped()
...
+        private void RunTestHelper(string testFileName)
+        {
+            string sarifLogText = GetInputSarifTextFromResource(testFileName);
+            SarifLog sarifLog = JsonConvert.DeserializeObject<SarifLog>(sarifLogText);
+
+            // If this test succeeds
```
shaopeng-gh, line 134: >// If this test succeeds [](http://example.com/codeflow?start=12&length=24)

see if we want to remove this comment

shaopeng-gh:
I am seeing some strange behavior, when I switch to this branch it will keep pop up many times the security warning, and each time click yes one of those files got generated as pending changes.
likely related to the changes in the file src/Test.FunctionalTests.Sarif/Test.FunctionalTests.Sarif.csproj

image

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Core
     public class LocationTests
     {
         private const string Id = "\"Id\"";
+        private static readonly TestAssetResourceExtractor Extractor = new TestAssetResourceExtractor(typeof(LocationTests));
```
shaopeng-gh, line 20: >Extractor [](http://example.com/codeflow?start=59&length=9)

the other file has:
private static readonly IResultMatcher s_matcher = new V2ResultMatcher();
should we renaming to have s_ similarly

src/Test.UnitTests.Sarif/Map/JsonMapBuilderTests.cs
```diff
@@ -9,13 +9,13 @@
 {
     public class JsonMapBuilderTests
     {
-        private static readonly ResourceExtractor Extractor = new ResourceExtractor(typeof(JsonMapBuilderTests));
+        private static readonly TestAssetResourceExtractor Extractor = new TestAssetResourceExtractor(typeof(JsonMapBuilderTests));
```
shaopeng-gh, line 12: >Extractor [](http://example.com/codeflow?start=59&length=9)

s_ for static

src/Test.UnitTests.Sarif/TestData/ExpectedOutputs/NoRuleDefaultConfiguration.sarif
```diff
@@ -0,0 +1,55 @@
...
+          "ruleId": "NoDefaultConfigurationLevel/ShouldBeNote",
+          "level": "note",
+          "message": {
+            "text": "Level should be 'note' as it is directly expressed on the result."
+          }
+        },
+        {
+          "ruleId": "NoDefaultConfigurationLevel/ShouldBeWarning",
+          "message": {
+            "text": "Level is 'warning' because kind is implictly 'fail' (so 'warning' is the default when there's no default configuration level')."
```
shaopeng-gh, line 41: >Level is 'warning' because kind is implictly 'fail' [](http://example.com/codeflow?start=21&length=51)

I like how the message is used for information.

Should we also add a case:
Level is 'warning' because kind is **explicit** 'fail'

src/Test.UnitTests.Sarif/TestData/RoundTripping/Inputs/InvocationOverridesRuleDefaultConfigurationOfError.sarif
```diff
@@ -0,0 +1,61 @@
...
+        },
+        {
+          "ruleId": "ExplicitDefaultLevel/ShouldBeNone",
+          "kind": "open",
+          "message": { "text": "Level should be 'none' because kind is not 'fail' and level is missing (so 'none' is the default)." }
+        },
+        {
+          "ruleId": "ExplicitDefaultLevel/ShouldBeNote",
+          "message": { "text": "Level should be 'note' as level is missing and 'note' is the invocation override configuration default." }
+        },
```
shaopeng-gh, line 46: I think we can also add a explicit kind fail test,
we know it should have the same result but if we add we know it is tested.

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -687,7 +687,8 @@ public void InsertOptionalDataVisitor_FlattensMessageStringsInFix()
         public void InsertOptionalDataVisitor_ResolvesOriginalUriBaseIds()
         {
             string inputFileName = "InsertOptionalDataVisitor.txt";
-            string testDirectory = GetTestDirectory("InsertOptionalDataVisitor") + Path.DirectorySeparatorChar;
+            string inputFileText = GetResourceText(inputFileName);
+            string testDirectory = Path.GetDirectoryName(GetResourceDiskPath(inputFileName));
```
shaopeng-gh, line 691: >string testDirectory = Path.GetDirectoryName(GetResourceDiskPath(inputFileName)); [](http://example.com/codeflow?start=12&length=81)

testDirectory is used for originalUriBaseIds,
it should be end with a `/`, the current code Path.GetDirectoryName() will not have a `/` at the end.
Also need the `+ Path.DirectorySeparatorChar;`

src/Test.UnitTests.Sarif.Converters/ConverterTestsBase.cs
```diff
@@ -10,6 +10,8 @@
 
 using Newtonsoft.Json;
 
+using Xunit.Abstractions;
```
shaopeng-gh, line 13: this file change can be reverted.

src/Test.UnitTests.Sarif.Multitool/TestData/QueryCommand/property-bag-queries.sarif
```diff
@@ -62,7 +62,7 @@
           },
           "properties": {
             "comment": "Case-insensitive match.",
-            "name": "terisa"
+            "name": "Calliope"
```
shaopeng-gh, line 65: >Calliope [](http://example.com/codeflow?start=21&length=8)

question, it was 1 terisa with lower t and the other one was with upper T

src/Test.UnitTests.Sarif.Multitool.Library/ExportValidationRulesMetadataCommandTests.cs
```diff
@@ -27,15 +27,14 @@ public void BuildRule_GeneratesExpectedMarkdown()
...
+            var resourceExtractor = new TestAssetResourceExtractor(this.GetType());
             foreach (KeyValuePair<string, SarifValidationSkimmerBase> test in tests)
             {
                 var sb = new StringBuilder();
                 var command = new ExportValidationRulesMetadataCommand();
                 command.BuildRule(test.Value, sb);
 
-                string expectedMarkdown = resourceExtractor.GetResourceText(test.Key);
-
+                string expectedMarkdown = resourceExtractor.GetResourceText($"ExpectedOutputs.{test.Key}");
```
shaopeng-gh, line 37: >GetResourceText [](http://example.com/codeflow?start=60&length=15)

can this  also use the `GetResourceExpectedOutputsText` and remove the `ExpectedOutputs.`

src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
```diff
@@ -285,7 +308,10 @@ protected virtual string ConstructTestOutputFromInputResource(string inputResour
                 sb.AppendLine(
                     "To rebaseline with current behavior:");
 
-                string testDirectory = Path.Combine(ProductTestDataDirectory, "ExpectedOutputs");
+                // TODO I suspect this baseline command works in some contexts (when we have a specific type under
```
shaopeng-gh, line 311: >TODO [](http://example.com/codeflow?start=19&length=4)

do we add name with todo item

michaelcfanning:
I see. Interesting. So possibly someone redacted that CSPROJ content to prevent this from occurring? i.e., active TT transformation. I will revert those changes and we'll see if everything continues to work.

---
In reply to: [1089577373](https://github.com/microsoft/sarif-sdk/pull/2473#issuecomment-1089577373) [](http://example.com/codeflow?ancestors=1089577373,1089577373)

src/Sarif/Autogenerated/ReportingDescriptorEqualityComparer.cs
michaelcfanning: OK! So I will retain this file.

src/NuGet.Config
michaelcfanning: Yes, sir good catch.

NuGet.Config
michaelcfanning: Yes, another good catch, another merge problem. :)

src/Test.FunctionalTests.Sarif/Multitool/ValidateCommandTests.cs
michaelcfanning: yes!

src/Test.UnitTests.Sarif/Core/LocationTests.cs
michaelcfanning: Scrubbed for this.

src/Test.UnitTests.Sarif/TestData/ExpectedOutputs/NoRuleDefaultConfiguration.sarif
michaelcfanning: Great question! The answer is no for a strange reason: doing this won't round-trip. :) If you mark a result explicitly as kind == 'fail', that explicit information should be removed on saving the log file (because it is the default value). And so that saved version won't match the input. 

Your test case is a good one, but we should include it elsewhere, i.e., our serialization logic should allow for this value to be explicitly set. Maybe we can then leverage the validation that we'll use here for the round-tripping to make sure everything is still correct.

src/Test.UnitTests.Sarif/TestData/RoundTripping/Inputs/InvocationOverridesRuleDefaultConfigurationOfError.sarif
michaelcfanning: See other comment. These files *must* be rendered precisely as we expect them to be when our serializer writes them to disk (according to the spec, eliding default values). That invariant is required in order to author simple, general purpose validation logic.

1) We read the log file
2) We validate that all result level/kind values are correct
3) We save the log file again to disk
4) We read that saved copy and validate nothing has changed.

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: Excellent comment! If these tests pass after my mistake, that means we didn't validate the resulting SARIF! I think we should scrub all tests, potentially, to add this additional logic (wonder what the perf impact would be).

src/Test.UnitTests.Sarif.Multitool/TestData/QueryCommand/property-bag-queries.sarif
michaelcfanning: Yes. Not sure how I missed this one, thanks.

src/Test.UnitTests.Sarif.Multitool.Library/ExportValidationRulesMetadataCommandTests.cs
michaelcfanning: Yes! you understand the pattern. :)

src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
michaelcfanning: Thanks! I actually think I located and fixed the bug, there was a suite in this repo that showed the potential issue (related to having an intermediate directory storing test files).

shaopeng-gh:
signed off, assume these will be taken care of
1. this strange issue
2. test green checkmark
3. CodeQL green checkmark

---
In reply to: [1090710906](https://github.com/microsoft/sarif-sdk/pull/2473#issuecomment-1090710906) [](http://example.com/codeflow?ancestors=1090710906,1089577373)

shaopeng-gh (approved):
:shipit:

src/Test.EndToEnd.Baselining/Test.EndToEnd.Baselining.csproj
eddynaka: I think, the issue is related to this:
[https://github.com/microsoft/sarif-sdk/blob/ed38f391b85e9a26b5e6079ad29fdd9f2b342e5e/src/build.props#L40-L41](https://github.com/microsoft/sarif-sdk/blob/ed38f391b85e9a26b5e6079ad29fdd9f2b342e5e/src/build.props#L40-L41)

And, each test project we are also using the same build.props, which is changing the namespace.

We could create a condition in that build props to only apply that if we have a product code or create a separate props file to test projects.

michaelcfanning: Thanks, Eddy, something for the future.

eddynaka (approved):
:shipit:

shaopeng-gh:
@michaelcfanning fyi I think the Linux test fail are just because the git.exe usage is increased to the methods that did not depends on them before, during the refactoring. You can consider just add windows only flag to it. This issue will only be truly resolved after we have cross-platform git related code. (I can investigate that if you like me to)

michaelcfanning:
@shaopeng-gh, can you do me a favor and fix the Linux test break? It's a directory separator problem I think.

shaopeng-gh:
@michaelcfanning will check and fix it.

---
In reply to: [1104510890](https://github.com/microsoft/sarif-sdk/pull/2473#issuecomment-1104510890) [](http://example.com/codeflow?ancestors=1104510890,1104510890)

shaopeng-gh:
@michaelcfanning investigated, the refactoring work we have here 
removed the method that has no other dependency only using pure path manipulation, 
and replaced with ProductRootDirectory which invoke the DirectoryHelpers.GetEnlistmentRoot() which is using git.exe, 
which we know will only work in Windows. 
This is not a new issues and we do all tests depended on git.exe mark as Windows only already. 
I think the refactoring work we have is right direction,
I have mark this one RebaseUriCommand_InjectsRegions() as Window Only as well.

---
In reply to: [1104512336](https://github.com/microsoft/sarif-sdk/pull/2473#issuecomment-1104512336) [](http://example.com/codeflow?ancestors=1104512336,1104510890,1104510890)

PR #2479, opened 2022-04-21
Removing Sarif.TypeScript and Typescript folders

eddynaka:
This change will delete the following folders:
- src/Sarif.TypeScript
- src/Typescript

As per Jeff's comments:

> On closer inspection, It looks to be unfinished work from https://github.com/microsoft/sarif-sdk/commit/33f83d68790966050c4b46f9da50ed1050087f23 and https://github.com/microsoft/sarif-sdk/commit/6c9dc2347a305dbb1ef7ef664b58d3a2d0765ee8.

The code was moved to this: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sarif/index.d.ts


src/TypeScript/Sarif.SDK/package.json
```diff
@@ -30,7 +30,7 @@
     "@types/chai": "^4.1.3",
     "@types/mocha": "^5.2.0",
     "chai": "^4.1.2",
-    "mocha": "^5.1.1",
+    "mocha": "^9.2.2",
```
eddynaka, line 33: >mocha": "^9.2.2 [](http://example.com/codeflow?start=5&length=15)

Hi @jeffersonking / @EasyRhinoMSFT,

do you know how can I test this?

src/TypeScript/Sarif.SDK/package-lock.json
```diff
@@ -1,8 +1,1252 @@
...
+      "integrity": "sha512-f5dXGzOJycyzSMdaXVhiBhauL4dYydXwVpavfQ1mVCaGjR56a9QfklXObUxlIY9bGTmCPHEEZ04I16BZ/8w5ww==",
+      "dev": true
+    },
+    "node_modules/@types/mocha": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@types/mocha/-/mocha-5.2.0.tgz",
+      "integrity": "sha512-YeDiSEzznwZwwp766SJ6QlrTyBYUGPSIwmREHVTmktUYiT/WADdWtpt9iH0KuUSf8lZLdI4lP0X6PBzPo5//JQ==",
+      "dev": true
+    },
+    "node_modules/@types/sarif": {
```
michaelcfanning, line 35: >sarif [](http://example.com/codeflow?start=25&length=5)

first, this seems like an awful lot of churn in this file, is it expected? Also, why would we have a reference to the NPM SARIF types package? isn't that exactly what we're building in this code base?

michaelcfanning:
There's no release note for this potentially customer-impacting change, let's please add one.

src/TypeScript/Sarif.SDK/package-lock.json
EasyRhinoMSFT: This file is generated by npm.

jeffersonking: @michaelcfanning Yes, this churn is par for the course. `package-lock.json` is practically considered opaque. I can dig deeper if you want to know why there's a self-reference.

src/TypeScript/Sarif.SDK/package.json
jeffersonking: Should be `npx mocha`.

src/TypeScript/Sarif.SDK/package-lock.json
michaelcfanning: You folks are the experts, I am mostly here to learn and to try to nudge this one forward. If we think it's safe, let's add a release note and close on it.

Ed's asked a good question around what kind of test story we have to ensure we don't regress due to update. ?

jeffersonking: At a glance, the `TypeScript` folder/project does not produce the TypeScript typings nor any other meaningful output. I'm not sure why this is here. I presume this was the beginning of an effort to migrate the TypeScript types generator from the `sarif-vscode-extension` codebase to here.

We should probably delete this folder. The typings generator should migrate to the `sarif-js-sdk` now that it exists.

eddynaka: @jeffersonking , with what you said in mind, is it safe to delete then?

jeffersonking: @eddynaka Yes, safe.

On closer inspection, It looks to be unfinished work from 33f83d68790966050c4b46f9da50ed1050087f23 and 6c9dc2347a305dbb1ef7ef664b58d3a2d0765ee8. @michaelcfanning would know this history, but I believe Everett's attempt was superseded by Rusty's.

michaelcfanning (approved):
:shipit:

PR #2480, opened 2022-04-21
Releasing 2.4.14

eddynaka:
This change is a preparation to release the version 2.4.14.


src/build.props
```diff
@@ -20,8 +20,8 @@
     having all the relevant values in one place. This property is actually used by the PowerShell
     script that hides ("delists") the previous package versions on nuget.org.
     -->
-    <VersionPrefix>2.4.13</VersionPrefix>
-    <PreviousVersionPrefix>2.4.12</PreviousVersionPrefix>
+    <VersionPrefix>2.4.14</VersionPrefix>
+    <PreviousVersionPrefix>2.4.13</PreviousVersionPrefix>
```
michaelcfanning, line 24: >PreviousVersionPrefix [](http://example.com/codeflow?start=5&length=21)

This comment says that the 'delist' command uses this property. Is that true? If not, let's please delete all this stuff.

If we *could* delist in future, please update this source code comment with an accurate state of the world.

src/ReleaseHistory.md
```diff
@@ -1,6 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
-## UNRELEASED
+## **v2.4.14** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/2.4.14) | [Driver](https://www.nuget.org/packages/Sarif.Driver/2.4.14) | [Converters](https://www.nuget.org/packages/Sarif.Converters/2.4.14) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/2.4.14) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/2.4.14)
```
michaelcfanning, line 3: >. [](http://example.com/codeflow?start=9&length=1)

What's prompting us to release at this specific point in time?

eddynaka: We are updating the sarif website with the latest package. If we use the Location Id change, we would be able to parse SARIF files that are using bigInteger values.

Also, we closed this: [https://github.com/microsoft/sarif-sdk/issues/2397](https://github.com/microsoft/sarif-sdk/issues/2397) but we never shipped the actual fix.

src/build.props
eddynaka: Removed the comment.

michaelcfanning (approved):
:shipit:

PR #2482, opened 2022-04-26
Fix `ArgumentNullException` when `PropertiesDictionary` is used and comparer is null

### LINKED ISSUE #2481: PropertiesDictionary is throwing ArgumentNullException when comparer is null
After we changed the TypedPropertiesDictionary from Dictionary to ConcurrentDictionary, we changed the behavior based on this:
Dictionary: <https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=net-6.0#system-collections-generic-dictionary-2-ctor(system-collections-generic-ienumerable((system-collections-generic-keyvaluepair((-0-1)))))>

ConcurrentDictionary: <https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.-ctor?view=netframework-4.5.2#system-collections-concurrent-concurrentdictionary-2-ctor(system-collections-generic-iequalitycomparer((-0)))>

ConcurrentDictionary will throw an exception when the comparer is null.
Dictionary will NOT throw an exception and will default to the EqualityComparer<T>.Default: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.equalitycomparer-1.default?view=net-6.0

@Rohan-Khemani , this is the issue that you are facing.

The PR that changed the behavior: https://github.com/microsoft/sarif-sdk/pull/2415

eddynaka:
Fixes #2481
This change will fix the ArgumentNullException when passing a null comparer to the ConcurrentDictionary when using the TypedPropertiesDictionary.

The issue:
1. https://github.com/microsoft/sarif-sdk/blob/ea44355584c161aec784ee9d2aa3c57f84875906/src/Sarif.WorkItems/SarifWorkItemContext.cs#L12-L22
2. https://github.com/microsoft/sarif-sdk/blob/ea44355584c161aec784ee9d2aa3c57f84875906/src/Sarif/PropertiesDictionary.cs#L20-L31
3. https://github.com/microsoft/sarif-sdk/blob/503fe2946bda129bf0900dfa20ffdfcbe88e0be9/src/Sarif/TypedPropertiesDictionary.cs#L16-L31

With that, on (3), the second constructor will receive a null comparer, generating the ArgumentNullException.


src/Test.UnitTests.WorkItems/Test.UnitTests.WorkItems.csproj
```diff
@@ -10,6 +10,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net48</TargetFrameworks>
```
eddynaka, line 13: >net48 [](http://example.com/codeflow?start=76&length=5)

To simulate the issue, we need to change to targetFramework 48. It will happen every time.

eddynaka: As a side note, I also tried to add net48 for all the other test projects to enhance our testing capabilities BUT the pipeline is timing out. So, we have to investigate why is that happening.

src/Sarif/TypedPropertiesDictionary.cs
```diff
@@ -19,7 +19,7 @@ public TypedPropertiesDictionary() : this(null, StringComparer.Ordinal)
         {
         }
 
-        public TypedPropertiesDictionary(PropertiesDictionary initializer, IEqualityComparer<string> comparer) : base(comparer)
+        public TypedPropertiesDictionary(PropertiesDictionary initializer, IEqualityComparer<string> comparer) : base(comparer ?? StringComparer.Ordinal)
```
eddynaka, line 22: >StringComparer [](http://example.com/codeflow?start=130&length=14)

As I mentioned in the description, when this is instantiated by the SarifWorkItemContext, the comparer will be null.

So, to use the same strategy as the constructor above, I'm defaulting the comparer to StringComparer.Ordinal.

src/Test.UnitTests.WorkItems/Test.UnitTests.WorkItems.csproj
eddynaka: Also included a simple and focused test to prevent us from facing this issue again.

src/ReleaseHistory.md
```diff
@@ -1,6 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
 ## Unreleased
+
+* BUGFIX: Fix `ArgumentNullException` when `PropertiesDictionary` is used and comparer is null. [#2482](https://github.com/microsoft/sarif-sdk/pull/2482)
```
michaelcfanning, line 5: >used [](http://example.com/codeflow?start=69&length=4)

'is instantiated with a null comparer.'

src/Sarif/TypedPropertiesDictionary.cs
michaelcfanning: Yes! Good fix. Did you confirm that this is the behavior for Dictionary<string, T> by examining the reference implementation? Ideally, we would match the behavior of that concrete type.

michaelcfanning (approved):
:shipit:

PR #2483, opened 2022-04-26
Releasing v2.4.15

eddynaka:
This change will release a new package with the following fixes:
* BUGFIX: Fix `ArgumentNullException` when `PropertiesDictionary` is instantiated with a null comparer. [#2482](https://github.com/microsoft/sarif-sdk/pull/2482)
* BUGFIX: Fix `UnhandledEngineException` when target path does not exist for multithreaded application by validating directories as is done for singlethreaded analysis. [#2461](https://github.com/microsoft/sarif-sdk/pull/2461)

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,6 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
-## Unreleased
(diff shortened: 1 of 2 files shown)


michaelcfanning (approved):
:shipit:

Issue #2424, opened 2022-01-20
Comprehensive sorting supports


yongyan-gh:
An issue opened in JSchema repo to enable auto generated IComparer for all SARIF types.
https://github.com/microsoft/jschema/issues/141

The auto generated Comparer implementation should covers all types while comparing. Once it's in place, we should use them for sorting.

PR #2485, opened 2022-05-05
Replace existing comparers with autogenerated comparers by JSchema

yongyan-gh:
# Description


src/Sarif/CodeGenHints.json
```diff
@@ -4,7 +4,6 @@
       "kind": "DictionaryHint",
       "arguments": {
         "valueTypeName": "SerializedPropertyInfo",
-        "namespaceName": "Microsoft.CodeAnalysis.Sarif.Readers",
```
yongyan-gh, line 7: >Microsoft.CodeAnalysis.Sarif.Readers [](http://example.com/codeflow?start=26&length=36)

This is to cleanup unused namespace in autogenerated codes.
"*.Properties" applies to all entities in Sarif, but only a few classes need the namespace "Microsoft.CodeAnalysis.Sarif.Readers".
Instead adding the namespace for all classes, just explicitly adding namespace for the classes require it.

src/Sarif/NotYetAutoGenerated/ReportingDescriptorEqualityComparer.cs
```diff
@@ -1,339 +0,0 @@
...
-
-using Microsoft.CodeAnalysis.Sarif.Readers;
-
-namespace Microsoft.CodeAnalysis.Sarif
-{
-    /// <summary>
-    /// Defines methods to support the comparison of objects of type ReportingDescriptor for equality.
-    /// </summary>
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "1.1.3.0")]
-    internal sealed class ReportingDescriptorEqualityComparer : IEqualityComparer<ReportingDescriptor>
```
yongyan-gh, line 16: >ReportingDescriptorEqualityComparer [](http://example.com/codeflow?start=26&length=35)

This is removed because we fixed the issue for Dictionary type of data model (`MultiformatMessage` in this case) in EqualityComparer code generator, so the path version of this EqualityComparer is not required any more.

src/Sarif/NotYetAutoGenerated/AddressComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.CodeDom.Compiler;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    /// <summary>
+    /// Defines methods to support the comparison of objects of type Address for sorting.
+    /// </summary>
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "1.1.4.0")]
+    internal sealed class AddressComparer : IComparer<Address>
```
yongyan-gh, line 14: >AddressComparer [](http://example.com/codeflow?start=26&length=15)

This class is need because the NotYetAutoGenerated version of Address replaced couple properties of int type to int?. The nullable type is not supported yet for Comparer code generator.

src/Sarif/NotYetAutoGenerated/AddressComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                return compareResult;
+            }
+
+            if (left.RelativeAddress.TryReferenceCompares(right.RelativeAddress, out compareResult))
+            {
+                return compareResult;
+            }
+
+            compareResult = left.RelativeAddress.Value.CompareTo(right.RelativeAddress.Value);
```
eddynaka, line 40: >left.RelativeAddress.Value [](http://example.com/codeflow?start=28&length=26)

I think this will generate a null reference exception when RelativeAddress is null.

Can you add some tests as:
left.RelativeAddress = right.relativeAddress = null
left.RelativeAddres = null, right.relativeAddress=  1
left.RelativeAddress = 1, right.relativeAddress = null
left.relativeAddress = 1, right.relativeaddress = 1

src/Sarif/Comparers/ArtifactComparer.cs
```diff
@@ -1,114 +1,110 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.CodeDom.Compiler;
 using System.Collections.Generic;
 
-/// <summary>
-/// Note: This comparer may not have all properties compared. Will be replaced by a comprehensive
-/// comparer generated by JSchema as part of EqualityComparer in a planned comprehensive solution.
-/// Tracking by issue: https://github.com/microsoft/jschema/issues/141
```
marmegh, line 9: Does this PR address this issue?

yongyan-gh: Correct. Previously manually created some comparer classes for Sarif entities. With the new change in JSchema, comprehensive auto-generated comparer classes, those manually created ones can be removed.

also related issue: #2424

marmegh: Awesome. If you're removing a comment like this in the future, it would be helpful to add a comment in the PR to link it all together.
If this PR fixes the Issue #2424, you probably want to note that in the description.

michaelcfanning (approved):
:shipit:

PR #2494, opened 2022-07-01
Resolve unhandled exceptions when analyzing folders that contain large files

marmegh:
`OutOfMemory` and `NullReferenceException` exceptions are being thrown when analyzing folders containing large files. This change adopts a file size limit in kilobytes and defaults to 1024.

Added unit testing to cover combinations of file sizes and file size limits.

This is related to [#621](https://github.com/microsoft/sarif-pattern-matcher/pull/621).


src/Sarif/IAnalysisContext.cs
```diff
@@ -29,5 +29,11 @@ public interface IAnalysisContext : IDisposable
         bool AnalysisComplete { get; set; }
 
         DefaultTraces Traces { get; set; }
+
+        /// <summary>
+        /// Gets or sets the maximum file size (in kilobytes) that will be analyzed.
+        /// If not set, it will analyze all sizes.
+        /// </summary>
+        int FileSizeInKilobytes { get; set; }
```
eddynaka, line 37: >FileSizeInKilobytes [](http://example.com/codeflow?start=12&length=19)

since we are here, can we update this to MaxFileSizeInKilobytes or something that actually explains why is this needed?

marmegh: This was mentioned in the other PR, it would be a breaking change for SPAM. Is that acceptable?

eddynaka: I think, if that would make things clearer and we tell people, yes.

What are your thoughts?

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -754,6 +754,135 @@ public void MultithreadedAnalyzeCommandBase_EndToEndMultithreadedAnalysis()
             }
         }
 
+        [Fact]
+        public void MultithreadedAnalyzeCommandBase_TargetFileSizeTestCases()
+        {
+            Random random = new Random();
```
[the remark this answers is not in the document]
eddynaka: Looking closer, you just need to use the TestRule.s_seed

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -41,6 +41,13 @@ public IEnumerable<Assembly> RetrievePluginAssemblies(IEnumerable<Assembly> defa
             return assemblies;
         }
 
+        public bool IsTargetWithinFileSizeLimit(string path, int maxFileSize)
+        {
+            long fileSize = FileSystem.GetFileSize(path) / 1024;
+
+            return (maxFileSize == -1 || fileSize < maxFileSize);
```
eddynaka, line 48: >fileSize < maxFileSize [](http://example.com/codeflow?start=41&length=22)

I think a few cases are missing:
maxfile = -1 vs  -1/0/1/big
maxfile = file = 10 (a reasonable number)

src/Sarif.Multitool.Library/SarifValidationContext.cs
```diff
@@ -96,6 +96,13 @@ public Uri TargetUri
 
         public DefaultTraces Traces { get; set; }
 
+
```
eddynaka, line 99: remove empty line

src/Sarif/IAnalysisContext.cs
eddynaka: Another strategy could be the following:
1. we rename the C# property
2. we keep the exposed argument.

So, internally, we know the good name. and we tell that in the next release we will rename the argument :)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -754,6 +754,135 @@
...
+                },
+                new {
+                    expectedExitReason = ExitReason.InvalidCommandLineOption,
+                    fileSize = (long)ulong.MinValue,
+                    maxFileSize = (int)0
+                },
+                new {
+                    expectedExitReason = ExitReason.None,
+                    fileSize = (long)ulong.MinValue,
+                    maxFileSize = (int)0 + 1
```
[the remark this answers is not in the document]
michaelcfanning: CodeQL, what is going on with you? You must see something in the IL that we do not observe here.

src/Sarif/IAnalysisContext.cs
michaelcfanning: Please do not take this change now. A breaking change of this kind needs to come with a thoughtful implementation, this is new work that is slowing down the immediate change.

So, go ahead and file an issue. We have other patterns of deprecating command-line args (like --hashes is support to be obsoleted!!). We do this in a manner such as Eddy describes, retain the old and mark it obsolete and/or update the code to properly transform old command-lines to new. 

So, Ed, be sure to capture your excellent thoughts in the issue that you file. :)

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -251,7 +252,14 @@
...
                 var fileSpecifier = new FileSpecifier(normalizedSpecifier, recurse: analyzeOptions.Recurse, fileSystem: FileSystem);
-                foreach (string file in fileSpecifier.Files) { targets.Add(file); }
+                foreach (string file in fileSpecifier.Files)
+                {
+                    // Only include files that are below the max size limit.
+                    if (IsTargetWithinFileSizeLimit(file, _rootContext.FileSizeInKilobytes))
+                    {
+                        targets.Add(file);
+                    }
+                }
```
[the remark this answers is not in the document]
michaelcfanning: Wow. I actually like this result, it is, in fact, flagging a use case I agree with, a filtered, single-pass iteration over a collection. 

Please feel free to take this suggestion!

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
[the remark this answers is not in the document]
michaelcfanning: Please do step through other tests that show the output that we expect for any random use. Eddy, be sure to explain this sort of thing in detail, it's important for others to understand completely (you may have done this already offline).  Basically, any test that uses a random value generator MUST emit the seed used to initialize the PRNG in its output. For tests that fail only due to specific randomized values, this seed can be hard-coded on the dev box to ensure reproducibility of the problem.

By the way! I recently fixed a test bug of this kind in an MS service. I was co-developing a change and my build failed, test rerun succeeded. When I looked at history, I could clearly see this was a flaky test, but it only failed once every several hundred pipeline builds. :) The issue was a random value generator that only provoked a negative condition in a very small % of cases. Ouch! That bug had lived in this code base for over a year...

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -754,6 +754,135 @@ public void MultithreadedAnalyzeCommandBase_EndToEndMultithreadedAnalysis()
...
         }
 
+        [Fact]
+        public void MultithreadedAnalyzeCommandBase_TargetFileSizeTestCases()
+        {
+            Random random = new Random();
+            int randomMaxFileSize = random.Next(1, int.MaxValue - 1);
+            long randomFileSize = (long)random.Next(2, int.MaxValue - 1);
+
+
```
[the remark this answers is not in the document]
michaelcfanning: I would so dearly love to not see this kind of feedback in code reviews, i.e., for all style suggestions to mostly be resolved by an automated code formatter. Is there not an autofix for this in a Roslyn analyzer or command-line tool? It's important that we stay clean and detail-oriented, so I appreciate all of us pointing this sort of thing out, to be clear. Just encouraging us to remove as many issues of this kind from ever occurring, wherever we can.

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -122,6 +129,11 @@ internal bool ValidateOutputFileCanBeCreated(IAnalysisContext context, string ou
             return succeeded;
         }
 
+        internal bool ValidateFileSizeInKilobytes(int fileSizeInKilobytes)
```
michaelcfanning, line 132: >ValidateFileSizeInKilobytes [](http://example.com/codeflow?start=22&length=27)

This factoring looks questionable to be. Your method name is as long as its implementation. :) 

In general, a utility method only has value if it contains some specialized logic that you don't want to maintain in multiple places. A simple 'greater than 0' comparison doesn't seem to meet that test.

There is another subtle issue 'ValidateFileSizeInKilobytes' is suggestive, it suggests more validation is done. You encode 'kilobytes' in the name and 'file size' and yet the implementation of this helper seems to be 'VerifyIsGreaterThanZero', and that's it. 

Having said all this, I don't even see a caller for this method. :) So maybe I'm making a whole lot of pointing out you left some intermediate code in the file. :)

Final point, though, just to keep beating this to death, this helper doesn't access instance data and, if it were to survive, could be static.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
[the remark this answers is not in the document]
marmegh: We usually have a step that removes these extra lines as part of buildAndTest.cmd for most of our repos.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -754,6 +754,135 @@ public void MultithreadedAnalyzeCommandBase_EndToEndMultithreadedAnalysis()
...
+                },
+                new {
+                    expectedExitReason = ExitReason.None,
+                    fileSize = (long)ulong.MinValue,
+                    maxFileSize = 1
+                },
+                new {
+                    expectedExitReason = ExitReason.None,
+                    fileSize = (long)ulong.MinValue,
+                    maxFileSize = randomMaxFileSize
```
michaelcfanning, line 785: >randomMaxFileSize [](http://example.com/codeflow?start=34&length=17)

Don't do this. :) there really isn't a good use of a random value in this test pattern, in my opinion. Why is that? Because the test matrix of use values is so clear. You need all boundaries (max & min), you need zero, -1, and then a couple values within the interior range. 

Using random to generate something just won't buy you much, and meantime you've opted into a more complex test pattern which, for example, leads others to require you to follow specific pattern. Don't do that unless you really think you derive value from it.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -754,6 +754,135 @@ public void MultithreadedAnalyzeCommandBase_EndToEndMultithreadedAnalysis()
...
+                                                                    It.IsAny<SearchOption>())).Returns(files);
+                mockFileSystem.Setup(x => x.FileOpenRead(It.IsAny<string>())).Returns(mockStream.Object);
+                mockFileSystem.Setup(x => x.FileExists(tempFile.Name)).Returns(true);
+                mockFileSystem.Setup(x => x.GetFileSize(It.IsAny<string>())).Returns(testCase.fileSize);
+
+                Output.WriteLine($"The seed that will be used is: {TestRule.s_seed}");
+
+                var options = new TestAnalyzeOptions
+                {
+                    Threads = 10,
```
michaelcfanning, line 857: >Threads = 10 [](http://example.com/codeflow?start=20&length=12)

Why are you opting into this behavior, and enabling hashes, etc? This is a nice test, but it's doing a lot, why is that? Can't we make this a bit more focused on the matter at hand, file size validation?

When you start adding things, you increase performance costs, but you also risk introducing non-obvious breaks in tests. So, let's say we had a problem introduced by our hashing argment to --insert. This test may fail, and we'd all be confused why the file size test broken.

This is just an example. But the principle here is that we want targeted unit tests that directly focus on some test work, as far as possible

It isn't always possible to create a clean union. But for this example, I definitely don't see why we'd explicitly set any args, would prefer defaults and minimal options otherwise.

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
marmegh: Updated test cases and added some explicit follow up confirming this method separately.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
marmegh: Hashes, etc were part of the original customer escalation. It's fairly conclusive that with and without these the repro and fix are the same. I've removed.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Resolve OutofMemoryException and NullReferenceException' failures by adding `--file-size-in-kb` argument to specify a max file size to analyze or default to 1024KB. [#2494](https://github.com/microsoft/sarif-sdk/pull/2494)
```
michaelcfanning, line 5: >OutofMemoryException [](http://example.com/codeflow?start=18&length=20)

enquote this and the other exception kind.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -412,10 +412,10 @@ private async Task<bool> FindFilesAsync(TOptions options, TContext rootContext)
                     }
 #endif
 
-                    foreach (string file in FileSystem.DirectoryEnumerateFiles(directory,
-                                                                               filter,
-                                                                               SearchOption.TopDirectoryOnly))
+                    foreach (string file in FileSystem.DirectoryEnumerateFiles(directory, filter, SearchOption.TopDirectoryOnly)
+                            .Where(file => IsTargetWithinFileSizeLimit(file, _rootContext.FileSizeInKilobytes)))
```
michaelcfanning, line 416: >Where [](http://example.com/codeflow?start=29&length=5)

Please put your predicate/handling within the collection, the rule is to avoid chained collection traversal.

michaelcfanning, line 416: >Where [](http://example.com/codeflow?start=29&length=5)

put this filter predicate into the code block.

src/Sarif.Multitool/AnalyzeTestContext.cs
```diff
@@ -30,6 +30,8 @@ public class AnalyzeTestContext : IAnalysisContext
 
         public DefaultTraces Traces { get; set; }
 
+        public int FileSizeInKilobytes { get; set; } = -1;
```
michaelcfanning, line 33: >-1; [](http://example.com/codeflow?start=55&length=3)

remove this. only the attribute should set the default value.

src/Sarif.Multitool.Library/SarifValidationContext.cs
```diff
@@ -96,6 +96,12 @@ public Uri TargetUri
 
         public DefaultTraces Traces { get; set; }
 
+        /// <summary>
+        /// Gets or sets the maximum file size (in kilobytes) that will be analyzed.
+        /// If not set, it will analyze all sizes.
+        /// </summary>
+        public int FileSizeInKilobytes { get; set; } = -1;
```
michaelcfanning, line 103: >-1 [](http://example.com/codeflow?start=55&length=2)

Remove this. Did you author a test for the validation command that this is honored?

src/Test.Utilities.Sarif/TestAnalysisContext.cs
```diff
@@ -35,6 +35,12 @@ public class TestAnalysisContext : IAnalysisContext
 
         public bool Disposed { get; private set; }
 
+        /// <summary>
+        /// Gets or sets the maximum file size (in kilobytes) that will be analyzed.
+        /// If not set, it will analyze all sizes.
+        /// </summary>
+        public int FileSizeInKilobytes { get; set; } = -1;
```
michaelcfanning, line 42: >-1 [](http://example.com/codeflow?start=55&length=2)

Remove this.

michaelcfanning:
>     public class AnalyzeTestContext : IAnalysisContext

Why is this in here? See if you can delete it. I don't understand why a test file is in the production code, looks like it might be mistake.

---
Refers to: src/Sarif.Multitool/AnalyzeTestContext.cs:9 in eb91a3a. [](commit_id = eb91a3ab9bc1cc28919f2780a532656dbb459593, deletion_comment = False)

michaelcfanning (approved):
:shipit:

src/ReleaseHistory.md

FEATURE: Add `file-size-in-kb` argument that filters allows filtering scan targets by file size.

marmegh:
>     public class AnalyzeTestContext : IAnalysisContext

Eddy provided the [PR](https://github.com/microsoft/sarif-sdk/pull/2146) for historical context. Not part of this change, but this and related files should eventually be moved to test project.

---
In reply to: [1176781298](https://github.com/microsoft/sarif-sdk/pull/2494#issuecomment-1176781298) [](http://example.com/codeflow?ancestors=1176781298,1176781298)

---
Refers to: src/Sarif.Multitool/AnalyzeTestContext.cs:9 in eb91a3a. [](commit_id = eb91a3ab9bc1cc28919f2780a532656dbb459593, deletion_comment = False)

eddynaka (approved):
:shipit:

PR #2500, opened 2022-07-05
Add boilerplate helper to provide logging when no plugins are configured

michaelcfanning:
Malformed plugin command-line arguments can result in 'no rules loaded' messages, masking the actual issue, which is that no plugins were properly specified.

This change adds a new helper for this condition. We still return 'no rules loaded' for this case, as it's not clear why we need a more precise return value (we halt because there's no work to do). 

We did not have a unit tests file, I've created one and a boilerplate test,

### DIFF
--- a/src/Sarif/Errors.cs
+++ b/src/Sarif/Errors.cs
@@ -15,13 +15,14 @@ namespace Microsoft.CodeAnalysis.Sarif
         // Configuration errors:
         private const string ERR997_MissingFile = "ERR997.MissingFile";
         private const string ERR997_NoRulesLoaded = "ERR997.NoRulesLoaded";
+        private const string ERR997_NoPluginsConfigured = "ERR997.NoPluginsConfigured";
         private const string ERR997_ExceptionLoadingPlugIn = "ERR997.ExceptionLoadingPlugIn";
         private const string ERR997_NoValidAnalysisTargets = "ERR997.NoValidAnalysisTargets";
         private const string ERR997_ExceptionAccessingFile = "ERR997.ExceptionAccessingFile";
-        private const string ERR997_MissingReportingConfiguration = "ERR997.MissingReportingConfiguration";
         private const string ERR997_ExceptionCreatingLogFile = "ERR997.ExceptionCreatingLogFile";
         internal const string ERR997_AllRulesExplicitlyDisabled = "ERR997.AllRulesExplicitlyDisabled";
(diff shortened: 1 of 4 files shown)


src/Sarif/SdkResources.resx
```diff
@@ -178,7 +178,7 @@
     <value>Could not load analysis target '{0}'.</value>
   </data>
   <data name="ERR997_ExceptionLoadingPlugIn" xml:space="preserve">
-    <value>Could not load the plug-in '{0}'.</value>
+    <value>Could not load the plugin '{0}'.</value>
```
michaelcfanning, line 181: >plugin [](http://example.com/codeflow?start=30&length=6)

generally, we prefer the term 'plugin' these days, not the hyphenated form.

src/Sarif/Errors.cs
```diff
@@ -128,6 +129,26 @@ public static void LogAllRulesExplicitlyDisabled(IAnalysisContext context)
 
             context.RuntimeErrors |= RuntimeConditions.NoRulesLoaded;
         }
+        public static void LogNoPluginsConfigured(IAnalysisContext context)
```
yongyan-gh, line 132: >public [](http://example.com/codeflow?start=8&length=6)

missing a blank line above.

src/Sarif/Errors.cs
```diff
@@ -128,6 +129,26 @@ public static void LogAllRulesExplicitlyDisabled(IAnalysisContext context)
...
+                    Errors.CreateNotification(
+                        context.TargetUri,
+                        ERR997_NoPluginsConfigured,
+                        ruleId: null,
+                        FailureLevel.Error,
+                        exception: null,
+                        persistExceptionStack: false,
+                        messageFormat: null));
+
+            context.RuntimeErrors |= RuntimeConditions.NoRulesLoaded;
```
yongyan-gh, line 150: >NoRulesLoaded [](http://example.com/codeflow?start=55&length=13)

should we add a dedicated RuntimConditions value for no plugin error?

shaopeng-gh: dedicated RuntimCondition may add some value for person who is troubleshooting.

src/Sarif/Errors.cs
shaopeng-gh, line 132: >public [](http://example.com/codeflow?start=8&length=6)

do we need to add to history md file for a new public method.

shaopeng-gh (approved):
:shipit:

PR #2503, opened 2022-07-13
* BREAKING: `absoluteAddress` property of `address` changed from `int`(32bit) to `BigInteger`(unlimited) to fix `Newtonsoft.Json.JsonReaderException: JSON integer XXXXX is too large or small for an Int32.`

shaopeng-gh:
# Description:

We found that in SARIF spec, there is no definition of the max length of integer type, whether it is 32 bit, 64 bit or no limit.
Our current SDK code assumes 32 bit, which should be wrong assumption, should be no limit since spec does not mention any limit.

# Fixes:

Spot fix the `Address.AbsoluteAddress` to use `BigInteger` which is no limit.

Also identify the other fields in the same `Address` class (not in this PR):

  1. int? RelativeAddress
  if present, is an integer containing the relative address; 
  If absent, it SHALL default to null
  
  2. int? Length
  if present, is an integer whose absolute value specifies the number of bytes in the range of addresses specified by this object; 
  Can be negative; 
  if absent, it SHALL default to null
  
  3. int? OffsetFromParent
   if present, is an integer containing the offset of this address from the absolute address of theParent
   if absent, it SHALL default to null
   
  4. int Index
  If present, its value SHALL be the index of thisObject within theRun.addresses;
  does not mention what if absent, what it SHALL default to
  
  5. int ParentIndex
   If present, its value SHALL be whose value is the array index (§3.7.4) of theParent within theRun.addresses;
   does not mention what if absent, what it SHALL default to


src/Test.UnitTests.Sarif/TestData/Address/Address_BigInteger.sarif
```diff
@@ -0,0 +1,88 @@
...
+          "message": {
+            "text": "This app does not have capabilities to prevent tapjacking attacks. An attacker can hijack the user's taps and tricks him into performing some critical operations that he did not intend to."
+          },
+          "level": "note",
+          "locations": [
+            {
+              "id": 1,
+              "physicalLocation": {
+                "address": {
+                  "absoluteAddress": 31197130097450771296369962162453149327732752356239421572342053257324632475324
```
shaopeng-gh, line 36: >absoluteAddress [](http://example.com/codeflow?start=19&length=15)

data in this file contains very long int, 64 int max +1, 32 int max +1 and a few normal ones.

src/Sarif/Autogenerated/Address.cs
```diff
@@ -42,7 +43,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "absoluteAddress", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(-1)]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public int AbsoluteAddress { get; set; }
+        public BigInteger AbsoluteAddress { get; set; }
+
+        public bool ShouldSerializeAbsoluteAddress()
+        {
+            return AbsoluteAddress >= 0;
+        }
```
shaopeng-gh, line 51: this is needed because DefaultValue does not work for BigInteger.

then this need 2 things,
1. in the constructor to set it, which we already have.
2. ShouldSerializeXXX to handle the serialization.

src/ReleaseHistory.md
```diff
@@ -4,6 +4,7 @@
 
 * FEATURE: Add `max-file-size-in-kb` argument that allows filtering scan targets by file size. [#2494](https://github.com/microsoft/sarif-sdk/pull/2494)
 * BUGFIX: Fix false positive for `SARIF1002.UrisMustBeValid` for file URIs that omit the `authority`. [#2501](https://github.com/microsoft/sarif-sdk/pull/2501)
+* BREAKING: `absoluteAddress` property of `address` changed from `int`(32bit) to `BigInteger`(unlimited) to fix `Newtonsoft.Json.JsonReaderException: JSON integer XXXXX is too large or small for an Int32.` [#2502](https://github.com/microsoft/sarif-sdk/pull/2502)
```
shaopeng-gh, line 7: This PR follows the same fix as the one we did before for Location.Id.

Also listed the other int fields in the same `Address` class (not fixed in current PR), in the `description.txt` of the PR.

Let me know if we want this PR to fix the reported issue and publish a version then a separated PR to fix other fields we choose to. If this PR is reviewed all other properties can follow the same pattern.

marmegh: How about:
```Fix `Newtonsoft.Json.JsonReaderException: JSON integer XXXXX is too large or small for an Int32` by changing the `absoluteAddress` property of `address` from `int`(32bit) to `BigInteger`(unlimited).```

Comment (shaopeng-gh) [reply src/ReleaseHistory.md]: agree, but now I notice that both of them may confuse the people looking at this checkin history fixed all integer in the solution,

How about something like:
BREAKING: Fix `Newtonsoft.Json.JsonReaderException: JSON integer XXXXX is too large or small for an Int32` 
for `absoluteAddress` property of `address` 
by changing it from `int`(32bit) to `BigInteger`(unlimited).

Comment (marmegh) [reply src/ReleaseHistory.md]: Agreed.

**Comment by marmegh** [comment]
>         }

Just to confirm, this is the constructor implementation you mentioned above?

---
Refers to: src/Sarif/Autogenerated/Address.cs:121 in 88ec7d1. [](commit_id = 88ec7d184090ef6f742471f4b17077eaaea6ce4e, deletion_comment = False)

File: src/Sarif/Autogenerated/Address.cs [hunk 4]
```diff
@@ -42,7 +43,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "absoluteAddress", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(-1)]
```
marmegh, line 44: >[DefaultValue(-1)] [](http://example.com/codeflow?start=8&length=18)

If DefaultValue doesn't work for BigInteger, should this be removed?

src/Sarif/NotYetAutoGenerated/Address.cs
```diff
@@ -42,7 +43,12 @@ public SarifNodeKind SarifNodeKind
         [DataMember(Name = "absoluteAddress", IsRequired = false, EmitDefaultValue = false)]
         [DefaultValue(-1)]
```
marmegh, line 44: >[DefaultValue(-1)] [](http://example.com/codeflow?start=8&length=18)

Again, if DefaultValue isn't supported for BigInteger, should this be removed?

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
```diff
@@ -23,88 +23,6 @@
...
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\OneRunWithAllReportingDescriptors.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\ExpectedOutputs\ResultLocationsAsEmptyArray.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\MinimumWithLanguage.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\NestedInnerExceptionsInNotifications.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\NonDottedQuadFileVersion.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\OneRunWithAllReportingDescriptors.sarif" />
-    <None Remove="TestData\SarifVersionOneToCurrentVisitor\Inputs\ResultLocationsAsEmptyArray.sarif" />
-    <None Remove="TestData\UriConverter\UriConverterTests.json" />
-    <None Remove="TestData\V2ResultMatcher\elfie-arriba.sarif" />
-    <None Remove="xunit.runner.json" />
```
marmegh, line 105: Was it intentional or unintentional that all of these were removed from the .csproj?

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,201 @@
...
+            address.AbsoluteAddress++;
+            Address_VerifyRoundTripFromObjectHelper(address,
+                shouldSerialize: true, reconstructedAbsoluteAddress: address.AbsoluteAddress);
+
+            address.AbsoluteAddress = ulong.MaxValue;
+            Address_VerifyRoundTripFromObjectHelper(address,
+                shouldSerialize: true, reconstructedAbsoluteAddress: address.AbsoluteAddress);
+            address.AbsoluteAddress++;
+            Address_VerifyRoundTripFromObjectHelper(address,
+                shouldSerialize: true, reconstructedAbsoluteAddress: address.AbsoluteAddress);
```
marmegh, line 82: This looks like a list of test cases all being put through the same helper method. Can we consolidate this test a little so it conforms to "Arrange/Act/Assert" a little more cleanly?

src/Sarif/Autogenerated/Address.cs
shaopeng-gh: tested, no it is needed when a json string that does not have the AbsoluteAddress node at all, to be able to construct an c# object that has AbsoluteAddress property as -1

shaopeng-gh:
>         }

yes it is.

---
In reply to: [1183804034](https://github.com/microsoft/sarif-sdk/pull/2503#issuecomment-1183804034) [](http://example.com/codeflow?ancestors=1183804034,1183804034)

---
Refers to: src/Sarif/Autogenerated/Address.cs:121 in 88ec7d1. [](commit_id = 88ec7d184090ef6f742471f4b17077eaaea6ce4e, deletion_comment = False)

src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj
shaopeng-gh: yes intentional, I had a previous PR comment from Michael that <ItemGroup>that has only <None Remove> are not needed should be removed.

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: looking

src/ReleaseHistory.md
shaopeng-gh: updated as discussed.

yongyan-gh:
How do we handle JSchema code auto-generation for BigInt data type?
Looks like Jason Schema only support 'integer' data type, need additional information to tell JSchema to generate BigInt type.

```
        "absoluteAddress": {
          "description": "The address expressed as a byte offset from the start of the addressable region.",
          "type": "integer",
          "minimum": -1,
          "default": -1
        },
```

shaopeng-gh:
yes, if to auto gen JSchema project will need some work to figure out. Also does not support nullable int, the current file is already in NotYetAutoGenerated folder due to other properties.

---
In reply to: [1184801887](https://github.com/microsoft/sarif-sdk/pull/2503#issuecomment-1184801887) [](http://example.com/codeflow?ancestors=1184801887,1184801887)

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: checked and made the code clearer.

when I have the pattern in mind it make me realize that a few var name is not clear e.g. a AbsoluteAddress that is the expected result to compare to, after the "Act", I have changed the name to "expectedAbsoluteAddress" to make it more clear.

yongyan-gh:
great i see the issue you already created in JSchema [https://github.com/microsoft/jschema/issues/144](https://github.com/microsoft/jschema/issues/144)

---
In reply to: [1184958816](https://github.com/microsoft/sarif-sdk/pull/2503#issuecomment-1184958816) [](http://example.com/codeflow?ancestors=1184958816,1184801887)

src/Sarif/Autogenerated/Address.cs
eddynaka: Normally, this is placed in the core folder:
[https://github.com/microsoft/sarif-sdk/blob/41c12efef265094d9fb0dcf35eb890fc35079337/src/Sarif/Core/Run.cs#L194-L211](https://github.com/microsoft/sarif-sdk/blob/41c12efef265094d9fb0dcf35eb890fc35079337/src/Sarif/Core/Run.cs#L194-L211)

src/Sarif/NotYetAutoGenerated/Address.cs
```diff
@@ -42,7 +43,12 @@ public SarifNodeKind SarifNodeKind
...
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public int AbsoluteAddress { get; set; }
+        public BigInteger AbsoluteAddress { get; set; }
+
+        public bool ShouldSerializeAbsoluteAddress()
```
eddynaka, line 48: >ShouldSerializeAbsoluteAddress [](http://example.com/codeflow?start=20&length=30)

move this to the core (check my previous comment)

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,303 @@
...
+            address.AbsoluteAddress = long.MinValue;
+
+            Address_VerifyRoundTripFromObjectHelper(address,
+                expectedShouldSerializeAbsoluteAddress: false, expectedAbsoluteAddress: -1);
+
+            // Arrange
+            address.AbsoluteAddress--;
+
+            Address_VerifyRoundTripFromObjectHelper(address,
+                expectedShouldSerializeAbsoluteAddress: false, expectedAbsoluteAddress: -1);
```
eddynaka, line 41: we have a lot of duplication here.

can't you create a struct/object/class that contains the initial values and do a foreach just like all the other places?

marmegh: @eddynaka and @shaopeng-gh: Something along these lines?
```// Arrange
var testCases = new[]
{
    new {
        absoluteAddress = <>,
        expectedShouldSerializeAbsoluteAddress = <>
    },
    etc.
}

var address = new Address();

foreach(var testCase in testCases)
{
    address.AbsoluteAddress = testCase.absoluteAddress;
    int expectedAbsoluteAddress = testCase.absoluteAddress < 0 ? -1 : absoluteAddress;
    // Act..Assert (combined in a single helper method)
    Address_VerifyRoundTripFromObjectHelper(
        address,
        testCase.expectedShouldSerializeAbsoluteAddress, 
        expectedAbsoluteAddress 
        
}```

File: src/Sarif/Autogenerated/Address.cs (hunk 2 above)
Comment (shaopeng-gh) [reply src/Sarif/Autogenerated/Address.cs:51]: Just notice this, thanks for the info!

File: src/Test.UnitTests.Sarif/Core/AddressTests.cs (hunk 9 above)
Comment (shaopeng-gh) [reply src/Test.UnitTests.Sarif/Core/AddressTests.cs]: updated to use loops.

File: src/Test.UnitTests.Sarif/Core/AddressTests.cs [hunk 10]
```diff
@@ -0,0 +1,227 @@
...
+    {
+        private const string AbsoluteAddress = "\"AbsoluteAddress\"";
+        private static readonly TestAssetResourceExtractor s_extractor = new TestAssetResourceExtractor(typeof(AddressTests));
+
+        [Fact]
+        public void Address_VerifyRoundTripFromObject()
+        {
+            // Arrange
+            var testCases = new[] {
+                new { absoluteAddress = (BigInteger?)null, expectedShouldSerializeAbsoluteAddress = false },
```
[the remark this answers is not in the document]
shaopeng-gh: false, will not build.

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,227 @@
...
+
+                BigInteger expectedAbsoluteAddress =
+                    (!testCase.absoluteAddress.HasValue || testCase.absoluteAddress.Value < 0) ? -1 : testCase.absoluteAddress.Value;
+
+                // Act & Assert
+                Address_VerifyRoundTripFromObjectHelper(address,
+                    expectedShouldSerializeAbsoluteAddress: testCase.expectedShouldSerializeAbsoluteAddress,
+                    expectedAbsoluteAddress: expectedAbsoluteAddress);
+            }
+
```
eddynaka, line 61: remove extra new line.

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                if (testCase.absoluteAddress.HasValue)
+                {
+                    address.AbsoluteAddress = testCase.absoluteAddress.Value;
+                }
+
+                BigInteger expectedAbsoluteAddress =
+                    (!testCase.absoluteAddress.HasValue || testCase.absoluteAddress.Value < 0) ? -1 : testCase.absoluteAddress.Value;
+
+                // Act & Assert
+                Address_VerifyRoundTripFromObjectHelper(address,
```
eddynaka, line 57: >Address_VerifyRoundTripFromObjectHelper [](http://example.com/codeflow?start=16&length=39)

do you need to add Address here? do you think we would have something that is not address?

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                    expectedShouldSerializeAbsoluteAddress: testCase.expectedShouldSerializeAbsoluteAddress,
+                    expectedAbsoluteAddress: expectedAbsoluteAddress);
```
eddynaka, line 59: >expectedAbsoluteAddress: expectedAbsoluteAddress [](http://example.com/codeflow?start=20&length=48)

you don't need to repeat yourself

src/Test.UnitTests.Sarif/Core/AddressTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    address.AbsoluteAddress = testCase.absoluteAddress.Value;
+                }
+
+                BigInteger expectedAbsoluteAddress =
+                    (!testCase.absoluteAddress.HasValue || testCase.absoluteAddress.Value < 0) ? -1 : testCase.absoluteAddress.Value;
+
+                // Act & Assert
+                Address_VerifyRoundTripFromObjectHelper(address,
+                    expectedShouldSerializeAbsoluteAddress: testCase.expectedShouldSerializeAbsoluteAddress,
```
eddynaka, line 58: >expectedShouldSerializeAbsoluteAddress [](http://example.com/codeflow?start=20&length=38)

the idea to use this is if you are passing a bool or a magic number, so people would know what is the variable and what it does.

Since the property from the testCase already have the same information, you could remove this.

src/Test.UnitTests.Sarif/Core/AddressTests.cs
eddynaka: this applies for the other tests as well.

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: Oh got it, so I will change all from:
expectedShouldSerializeAbsoluteAddress: testCase.expectedShouldSerializeAbsoluteAddress,
to just:
testCase.expectedShouldSerializeAbsoluteAddress,

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: will change it from
Address_VerifyRoundTripFromObjectHelper
to:
VerifyRoundTripFromObjectHelper

src/Test.UnitTests.Sarif/Core/AddressTests.cs
marmegh: Exactly

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: fixed

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: done

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: done

src/Test.UnitTests.Sarif/Core/AddressTests.cs
shaopeng-gh: Thanks for the explanation in the other comment.

src/Sarif/Autogenerated/Address.cs
```diff
@@ -42,7 +43,7 @@ public SarifNodeKind SarifNodeKind
...
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public int AbsoluteAddress { get; set; }
+        public BigInteger AbsoluteAddress { get; set; }
```
shaopeng-gh, line 46: >BigInteger [](http://example.com/codeflow?start=15&length=10)

After this is merged there will be follow up PR(s) to apply the same change to else appropriate.

shaopeng-gh:
Closed this PR because we are looking at if the `jschema` tool can generate `BigInteger` for all, instead of fixing one by one.

Issue #2486, opened 2022-06-15
[SARIF Multitool] match-results-forward generates invalid SARIF when results contain SubRuleId


eddynaka:
# How to simulate:

Use the attached [example.zip](https://github.com/microsoft/sarif-sdk/files/8912141/example.zip) and execute the command: `sarif match-results-forward .\original.sarif --previous .\previous.sarif -o baselined.sarif --force`

# Current behavior:

The rules are duplicated after executing the command, generating an invalid SARIF.

# Expected behavior:

After executing the command, the SARIF must have the baselined data and be a valid SARIF.

eddynaka:
@marmegh this is one of the issues that I saw when running the E2E pipeline.

michaelcfanning:
@marmegh @EasyRhinoMSFT these issues filed by Eddy are important for working scenarios and it would be good to make them a priority. do we have a result matching expert blessed? this area is going to get some attention moving forward, for sure.

marmegh:
@yongyan-gh, this is one of the issues discussed today.
cc: @EasyRhinoMSFT, @shaopeng-gh

yongyan-gh:
The issue of invalid Sarif produced by `match-results-forward` command is that the Sarif contains duplicated rule definitions in `rules` array, if the Sarif to be merged contain results which use a sub `ruleId`.

E.g. if there are 2 results' `ruleId` is "TESTRULE/001/SUB001", both reference to parent rule "TESTRULE/001", the command will produce a Sarif which `rules` array contains 2 duplicated rules of "TESTRULE/001".

Have a fix in #2505 pls review

Issue #2487, opened 2022-06-15
[SARIF multitool] match-results-forward removes relevant data from original file


eddynaka:
# How to simulate:

Use the attached [example.zip](https://github.com/microsoft/sarif-sdk/files/8912197/example.zip)
 and execute the command: `sarif match-results-forward .\original.sarif --previous .\previous.sarif -o baselined.sarif --force`

# Current behavior:

The baselined SARIF does not keep `VersionControlProvenance`.

# Expected behavior:

All relevant data from the original SARIF should be preserved.

eddynaka:
@marmegh this is one of the issues that I saw when running the E2E pipeline.

marmegh:
@yongyan-gh, this is one of the issues discussed today.
cc: @EasyRhinoMSFT, @shaopeng-gh

PR #2505, opened 2022-07-20
Fix bug: `match-results-forward` command generates invalid Sarif (duplicated rules).

yongyan-gh:
# Description

Fixes for below issue:
match-results-forward command generates invalid SARIF when results contain SubRuleId (#2486)
- When merging results and rules in multiple runs, we create a rule-id to index mapping (Dictioanry) to keep tracking all referenced rules. It uses ruleId from result entity as the key. If there are multiple unique sub ruleIds of a parent rule in the results, it caches the same parent rule for each unique sub ruleId, which caused the invalid Sarif output.


src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -78,7 +78,23 @@ public override Result VisitResult(Result node)
...
+                if (possilbeParentRuleId != null &&
+                    ruleId.StartsWith(possilbeParentRuleId + SarifConstants.HierarchicalComponentSeparator))
+                {
+                    ruleId = possilbeParentRuleId;
+                }
+            }
+
             if (!string.IsNullOrEmpty(ruleId))
             {
                 if (RuleIdToIndex.TryGetValue(ruleId, out int ruleIndex))
```
yongyan-gh, line 100: >RuleIdToIndex [](http://example.com/codeflow?start=20&length=13)

It uses this rule-id to index mapping to track all referenced rules. 
If there are multiple unique sub ruleIds of a same parent rule in the results, it caches the same parent rule for each unique sub ruleId, which caused the invalid Sarif output.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -278,6 +278,7 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
             run.Graphs = graphs;
             run.Invocations = invocations;
             run.Properties = properties;
+            run.VersionControlProvenance = MergeVersionControlDetails(currentRuns.Concat(previousRuns));
```
eddynaka, line 281: >MergeVersionControlDetails [](http://example.com/codeflow?start=43&length=26)

can you verify if we have other relevant properties that we should keep?

src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -78,7 +78,23 @@ public override Result VisitResult(Result node)
...
+                ruleId.IndexOf(SarifConstants.HierarchicalComponentSeparator) > 0)
+            {
+                rule = node.GetRule(CurrentRun);
+                string possilbeParentRuleId = rule?.Id;
+                if (possilbeParentRuleId != null &&
+                    ruleId.StartsWith(possilbeParentRuleId + SarifConstants.HierarchicalComponentSeparator))
+                {
+                    ruleId = possilbeParentRuleId;
+                }
+            }
```
eddynaka, line 96: can't we replace this for a smaller logic?

ReportingDescriptor rule = result.GetRule(CurrentRun);
use rule everywhere when possible.

src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -87,15 +103,14 @@ public override Result VisitResult(Result node)
...
                     int newIndex = Rules.Count;
-                    Rules.Add(rule);
+                    Rules.Add(rule ?? node.GetRule(CurrentRun));
 
                     RuleIdToIndex[ruleId] = newIndex;
                     node.RuleIndex = newIndex;
                 }
 
-                node.RuleId = ruleId;
+                node.RuleId = resultRuleId;
```
eddynaka, line 113: >resultRuleId [](http://example.com/codeflow?start=30&length=12)

why do we have to change this?

eddynaka: can't we remove this entire line or just replace if node.ruleid is null/empty?

marmegh:
We also need a release history update for this.

src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -78,7 +78,23 @@ public override Result VisitResult(Result node)
             }
 
             // Cache RuleId and set Result.RuleIndex to the (new) index
-            string ruleId = node.ResolvedRuleId(CurrentRun);
+            string resultRuleId = node.ResolvedRuleId(CurrentRun);
+            string ruleId = resultRuleId;
```
marmegh, line 82: Do we really need both of these? Why isn't this just:
`string ruleId = node.ResolvedRuleId(CurrentRun);`?

src/Sarif/Visitors/RunMergingVisitor.cs
yongyan-gh: The `resultRuleId` keeps the value from result entity, `ruleId` is the value from referenced rule entity. They may be different if result uses a sub rule id of the rule.
e.g. 
in `rules` array there is a rule id is "SEC105/003", value of `ruleId` should be "SEC105/003"
a result has rule id "SEC105/003/maybenull", which is a sub rule of above rule, `resultRuleId` keeps this id.
`ruleId` should be used as key of the cache `RuleIdToIndex`.
`resultRuleId` used to set on result rule id.

src/Sarif/Visitors/RunMergingVisitor.cs
yongyan-gh: agree. The additional checks is to make sure the case is a rule with sub rule. I can trust result.GetRule(...) and remove these checks.

src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -87,15 +90,14 @@ public override Result VisitResult(Result node)
                 }
                 else
                 {
-                    ReportingDescriptor rule = node.GetRule(CurrentRun);
                     int newIndex = Rules.Count;
-                    Rules.Add(rule);
+                    Rules.Add(rule ?? node.GetRule(CurrentRun));
```
eddynaka, line 94: > ?? node.GetRule(CurrentRun) [](http://example.com/codeflow?start=34&length=28)

you can remove this, because is exact the same as line 82

eddynaka:
Also, the title of the PR should tell more about the bug. What bug are you fixing? Is it one? Many? Should be in the same PR or should be separate?

---
In reply to: [1192800154](https://github.com/microsoft/sarif-sdk/pull/2505#issuecomment-1192800154) [](http://example.com/codeflow?ancestors=1192800154,1192800154)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: there are couple properties of run like Addresses/WebRequest/WebResponse etc are not been set.
I am not sure if any special requirements to merge these properties. Can I just combine the values of the properties in the final run object?

src/Sarif/Visitors/RunMergingVisitor.cs
yongyan-gh: updated with simple implementation

src/Sarif/Visitors/RunMergingVisitor.cs
yongyan-gh: this was to keep the value of the result's rule id and can be used later, i updated the implemenation pls see the latest iteration.

src/Sarif/Visitors/RunMergingVisitor.cs
yongyan-gh: updated the implementation pls see latest iteration

yongyan-gh:
updated release history. since both issue are related to `result-match-foward` command, i put the fix in same pr.

---
In reply to: [1193015158](https://github.com/microsoft/sarif-sdk/pull/2505#issuecomment-1193015158) [](http://example.com/codeflow?ancestors=1193015158,1192800154)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -343,5 +344,24 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
                 }
             }
         }
+
+        private static IList<VersionControlDetails> MergeVersionControlDetails(IEnumerable<Run> runs)
+        {
+            // use HashSet with EqualityComparer to avoid duplicated entries
```
shaopeng-gh, line 350: >u [](http://example.com/codeflow?start=15&length=1)

Cap and add . at the end

eddynaka: I would remove the entire comment. I don't see the difference between comment and the actual code.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fix issue `match-results-forward` command generates invalid Sarif when results contain sub ruleId. [#2486](https://github.com/microsoft/sarif-sdk/pull/2486)
```
eddynaka, line 5: >BUGFIX: Fix issue `match-results-forward` [](http://example.com/codeflow?start=2&length=41)

I would either separate in two PRs so each PR has its own fix. Or, make an improvement in the text to be one entry that explains the current behavior for both issues and point to the correct PullRequest Id.

yongyan-gh:
remove fix for #2487 to another PR. keep only fix for #2486

---
In reply to: [1194682619](https://github.com/microsoft/sarif-sdk/pull/2505#issuecomment-1194682619) [](http://example.com/codeflow?ancestors=1194682619,1192800154)

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: removed the comments.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: moved this change to another PR so each PR. resolving this comment.

src/ReleaseHistory.md
yongyan-gh: Separated into 2 PRs.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
...
```
marmegh, line 5: >BUGFIX: Fix issue `match-results-forward` command generates invalid Sarif when results contain sub ruleId. [#2486](https://github.com/microsoft/sarif-sdk/pull/2486) [](http://example.com/codeflow?start=2&length=164)

I'm not sure if this is just unclear or not descriptive enough. What was invalid about the SARIF results before this fix?

src/Sarif/Visitors/RunMergingVisitor.cs
```diff
@@ -78,7 +78,10 @@ public override Result VisitResult(Result node)
             }
 
             // Cache RuleId and set Result.RuleIndex to the (new) index
-            string ruleId = node.ResolvedRuleId(CurrentRun);
+            string resultRuleId = node.ResolvedRuleId(CurrentRun);
+            ReportingDescriptor rule = node.GetRule(CurrentRun);
+            string ruleId = rule?.Id ?? resultRuleId;
```
marmegh, line 83: >string ruleId = rule?.Id ?? resultRuleId; [](http://example.com/codeflow?start=12&length=41)

Is there a case where rule.Id might not be null and resultRuleId may be another value?

src/ReleaseHistory.md
yongyan-gh: The issue of invalid Sarif produced before the fix is it contains duplicated rule definitions in rules array if the Sarif results contain results have sub ruleId.

E.g. if there are 2 results' ruleId is "TESTRULE/001", the rules array creates 2 duplicated rules of "TESTRULE".

Will update the description with the duplicated rules issue, and update the issue with the detail errors. What do you think?

marmegh: So, this is fixing an issue where the `match-results-forward` command generates Sarif with duplicated results if there is a sub `ruleId`?
Yes, please update both.

src/Test.UnitTests.Sarif/Visitors/RunMergingVisitorTests.cs
```diff
@@ -359,5 +360,90 @@ private void VerifyRuleMatches(Run mergedRun, Run sourceRun, int firstResultInde
...
+            mergedRun.Results.Count.Should().Be(11);
+            foreach (Result result in mergedRun.Results)
+            {
+                string subRuleId = result.RuleId;
+                string parentRuleId = result.GetRule(mergedRun).Id;
+                if (subRuleId != parentRuleId)
+                {
+                    subRuleId.StartsWith(parentRuleId).Should().BeTrue();
+                }
+            }
```
marmegh, line 446: We can consolidate all of this:
```
int rule_001Count = 0;
int rule_002Count = 0;
int rule_003Count = 0;
int rule004Count = 0;
int rule_005Count = 0;
int rule_006Count = 0;
int rule_DoesNotExistCount = 0;

mergedRun.Results.Count.Should().Be(11); //Even better if you have a way of calculating this expected value from the inputs above.

foreach (Result result in mergedRune.Results)
{
    ...// keep everything the same and add something like this:
   if(result.Id == "Rule/001")
   {
       rule_001Count+=1;
   }
}

rule_001Count.Should().Be(expectedRule001Count);
//etc
```

We're trying to avoid iterating over the same lists multiple times. Let me know if this makes sense.

marmegh:
>                 throw new InvalidOperationException("RemapIndicesVisitor requires CurrentRun to be set before Visiting Results.");

Code coverage (via Live Unit Testing) shows this is the only segment of code in VisitResults that is not covered by a test. Stepped through the added tests and confirmed they cover remaining if/else blocks. All changed code appears fully covered.

If at all possible to sneak in a unit test to cover this section, that would help improve our code coverage overall.

---
Refers to: src/Sarif/Visitors/RunMergingVisitor.cs:77 in 9548a62. [](commit_id = 9548a62486d516a1ae73a71e28a1eb31a423f605, deletion_comment = False)

src/ReleaseHistory.md
marmegh: And just to be clear, also update the release note here. If the above captures everything, you can copy/paste most of it. Don't forget the backticks around ruleId.

src/Test.UnitTests.Sarif/Visitors/RunMergingVisitorTests.cs
yongyan-gh: agree, updated & removed duplicated iterations

yongyan-gh:
>                 throw new InvalidOperationException("RemapIndicesVisitor requires CurrentRun to be set before Visiting Results.");

test case added for the case CurrentRun is null.

---
In reply to: [1211334197](https://github.com/microsoft/sarif-sdk/pull/2505#issuecomment-1211334197) [](http://example.com/codeflow?ancestors=1211334197,1211334197)

---
Refers to: src/Sarif/Visitors/RunMergingVisitor.cs:77 in 9548a62. [](commit_id = 9548a62486d516a1ae73a71e28a1eb31a423f605, deletion_comment = False)

shaopeng-gh:
Need to resolve the conflicts.

shaopeng-gh (approved):
:shipit:

yongyan-gh:
resolved

---
In reply to: [1218847548](https://github.com/microsoft/sarif-sdk/pull/2505#issuecomment-1218847548) [](http://example.com/codeflow?ancestors=1218847548,1218847548)

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fix an issue where the `match-results-forward` command generates Sarif with duplicated rules if there are results with sub `ruleId`. [#2486](https://github.com/microsoft/sarif-sdk/pull/2486)
```
michaelcfanning, line 5: >Fix an issue where the `match-results-forward` command generates Sarif with duplicated rules if there are results with sub `ruleId [](http://example.com/codeflow?start=10&length=130)

Please change this to: 

Remove duplicated rule definitions when executing `match-results-forward` commands for results with sub-rule ids.

michaelcfanning (approved):
:shipit:

src/ReleaseHistory.md
yongyan-gh: updated

PR #2512, opened 2022-07-27
Fix bug: `match-results-forward` command removes relevant data from original file

yongyan-gh:
# Description

Fixes for below issue:

`match-results-forward` command removes relevant data from original file (#2487)

- Add code to handle merging of missing data `VersionControlProvenance`


src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -343,5 +347,24 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
                 }
             }
         }
+
+        private static IList<T> MergeList<T>(IEnumerable<T> source1, IEnumerable<T> source2, IEqualityComparer<T> comparer)
+        {
+            var elementSet = new HashSet<T>(comparer);
+
+            List<T> combinedList = new List<T>();
+            combinedList.AddRange(source1 ?? Enumerable.Empty<T>());
```
eddynaka, line 356: >?? Enumerable.Empty<T>() [](http://example.com/codeflow?start=42&length=24)

You could remove this since the caller already verifies it.

src/Test.UnitTests.Sarif/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
```diff
@@ -387,6 +387,53 @@ public void SarifLogResultMatcher_PreservesPropertiesProperly()
             matchedLog.Runs[0].Results?.Where((r) => { return r.GetProperty("Key") == currentPropertyValue; }).Count().Should().Be(matchedLog.Runs[0].Results.Count);
         }
 
+        [Fact]
+        public void SarifLogResultMatcher_PreservesVersionControlProvenanceProperly()
```
eddynaka, line 391: >SarifLogResultMatcher_PreservesVersionControlProvenanceProperly [](http://example.com/codeflow?start=20&length=63)

You need to include some null scenarios as well.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fix issue `match-results-forward` command generates Sarif missing VersionControlDetails data. [#2487](https://github.com/microsoft/sarif-sdk/pull/2487)
```
michaelcfanning, line 5: >VersionControlDetails [](http://example.com/codeflow?start=76&length=21)

Named terms should be encapsulated in back ticks.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
michaelcfanning: I'd actually suggest removing the caller's handling, as it strengthens this method to have this code.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -343,5 +347,24 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
                 }
             }
         }
+
+        private static IList<T> MergeList<T>(IEnumerable<T> source1, IEnumerable<T> source2, IEqualityComparer<T> comparer)
```
michaelcfanning, line 351: >MergeList [](http://example.com/codeflow?start=32&length=9)

I don't see unit tests for this generic helper.

This is extremely generic code, do you have it in the right place? Should it live elsewhere?

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -278,6 +278,10 @@ private List<ExtractedResult> ExtractResultsFromRuns(IEnumerable<Run> sarifRuns)
             run.Graphs = graphs;
             run.Invocations = invocations;
             run.Properties = properties;
+            run.VersionControlProvenance = MergeList(
+                currentRuns?.SelectMany(run => run.VersionControlProvenance ?? Enumerable.Empty<VersionControlDetails>()),
```
michaelcfanning, line 282: >currentRuns [](http://example.com/codeflow?start=16&length=11)

this is a poor practice, to iterate over these collections and the pass that result to a helper that will iterate again. please rewrite to avoid that. looks like you already provide handling in the callee for the null run.VersionControlProvenance case.

michaelcfanning (changes requested):
:clock1:

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
shaopeng-gh: looks like you already provide handling in the callee for the null run.VersionControlProvenance case.  
  ------  Just to comment on this line, with the current code if there is one run.VersionControlProvenance is null, the C# SelectMany() will have NullRef exception. 

This is by design of the SelectMany() when any one of the selector returns null.

This can not be handled by null check inside MergeList() method, because the error happens at SelectMany().

for SelectMany people usually write this way:
`AAAA
.Where(x => x.property1 != null)
.SelectMany(x => x.property1)`

(Just for discussion don't make changes according to my comment :) )

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
shaopeng-gh: 1. I think if a helper method can handle null case inside it would be safer and may save some duplicated code for the person calling it.

2. another way is put different case of null check at the first few lines of the helper method which people don't like,
if source1 is not null and source2, do we even need to loop or just return source1

3. This method not just do merge but also do distinct, if source1 has 3 items and source2 is null, the result only have 2 items, the person call this method will be surprised. Should we call it like MergeAndDistinctList

(Just for discussion don't directly make changes according to my comment :) )

yongyan-gh: refactor/simplified the method into an extension method, now it doesn't need to join the lists into 1 and null case handling is easier, pls see latest change.

src/Test.UnitTests.Sarif/Baseline/ResultMatching/SarifLogResultMatcherTests.cs
yongyan-gh: added null value test cases.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: refactor this function avoid the duplicated iteration of the collection, also no need to handle the null value in the enumerable the old function has to deal with. 
please see the latest change.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: refactor the function and unit tests added pls see the latest change.

src/Sarif/ExtensionMethods.cs
```diff
@@ -514,6 +514,45 @@
...
+
+            var elementSet = new HashSet<T>(equalityComparer);
+
+            foreach (T element in list)
+            {
+                if (elementSet.Add(element))
+                {
+                    yield return element;
+                }
+            }
```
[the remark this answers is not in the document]
yongyan-gh: Right, in this case the suggestion doesn't make sense because it has to iterate all elements and yield return the element, change to `Where` call make it difficult to understand.

In some cases the `Where` may be useful to make a complicated filtering condition to be a short, easy to read standard query, with little performance trade-off.

I think we can keep the codeQL rule, and check specific case if its applies. :grinning:

src/Sarif/ExtensionMethods.cs
```diff
@@ -514,6 +514,45 @@
             return propertyBag;
         }
 
+        public static IEnumerable<T> MergeWithList<T>(
```
shaopeng-gh, line 517: >MergeWithList [](http://example.com/codeflow?start=37&length=13)

This is a common method, with the current name,
This method not just do merge but also do distinct, 
if original list has 3 items and the second list contains 0 item, 
the result returns only 1 item, the person calling this method will be surprised. 

Not sure if there is a way to make the name clear 
e.g. 
`DistictMergeWithList`

yongyan-gh: updated

shaopeng-gh (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fix issue `match-results-forward` command generates Sarif missing `VersionControlDetails` data. [#2487](https://github.com/microsoft/sarif-sdk/pull/2487)
```
michaelcfanning, line 5: >Fix issue [](http://example.com/codeflow?start=10&length=9)

Please use this:

'Resolve issue where `match-results-forward` command fails to generate `VersionControlDetails` data.

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
```diff
@@ -273,11 +274,29 @@
...
+                    versionControls = versionControls.DistinctMerge(
+                        previousRun.VersionControlProvenance,
+                        VersionControlDetails.ValueComparer).ToList();
+                }
             }
 
             run.Graphs = graphs;
             run.Invocations = invocations;
             run.Properties = properties;
+            run.VersionControlProvenance = versionControls.Any() ? versionControls : null;
```
michaelcfanning, line 299: >versionControls [](http://example.com/codeflow?start=43&length=15)

Just use versionControls.FirstOrDefault() here.

michaelcfanning (approved):
:shipit:

src/Sarif/Baseline/ResultMatching/SarifLogResultMatcher.cs
yongyan-gh: This check is to prevent `run.VersionControlProvenance` to be set to an empty list. 

If it's an empty list, the output sarif file contains the entry.
```
      "versionControlProvenance": [ ],
```
if it's null the output file doesn't contain it.

The empty and null VersionControlProvenance are same semantically, but the empty entry will break some file diffing tests. What do you think?

Issue #2488, opened 2022-06-15
[SARIF Multitool] merge command generates SARIF with multiple runs


eddynaka:
# How to simulate:

Use the attached [example.zip](https://github.com/microsoft/sarif-sdk/files/8912215/example.zip) and execute the command: `sarif merge *.sarif --output-file merged.sarif --force`

# Current behavior:

The merged SARIF file creates one run with one rule for each distinct rules/results/tool you have in the SARIF.

# Expected behavior:

The merged SARIF should contain:
- multiple runs if we have multiple versions of tools
- single run if we have multiple SARIFs with the same tool/versioning but different results

eddynaka:
@marmegh this is one of the issues that I saw when running the E2E pipeline.

marmegh:
@yongyan-gh, this is one of the issues discussed today.
cc: @EasyRhinoMSFT, @shaopeng-gh

yongyan-gh:
There is a command argument `merge-runs` for merge command is to merge multiple runs by same tool + version. But it's not working properly.
https://github.com/microsoft/sarif-sdk/blob/698adb6365a242c6bb75adde56e3bd4be39c21d7/src/Sarif.Multitool.Library/MergeOptions.cs#L33-L37

So I created the fix PR based on the assumption:
- By default, the merge command merges multiple runs by `rule id + tool + version`.
- If `merge-runs` argument is set, the merge command merges multiple runs by `tool + version`.

Please review detail fix in the PR.

Thanks

PR #2513, opened 2022-07-28
Fix bug: `merge` command generates multiple runs

yongyan-gh:
# Description 

Fixes for below issues:

- merge command generates SARIF with multiple runs (https://github.com/microsoft/sarif-sdk/issues/2488)
By default, the results are grouped into multiple runs per rule id + tool version.
When `merge-runs` argument is set, group the results by tool version, this will place all results of runs produced by same tool same version into a single run.

- Another issue it fixes is `System.IO.IOException`/`System.IO.DirectoryNotFoundException` exception when execute `merge` command with `--split PerRule` option, and the Sarif files to be merged contains rule Id has invalid character for a file name. E.g. a sub rule id "SEC105/001". 
With `--split PerRule` option, `merge` command creates separate Sarif files for each rule id. E.g. "SEC105/001_merged.sarif". The character "/" in the rule Id is an invalid file name character, it will cause the error to create the merged file.


src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -117,7 +121,7 @@ public int Run(MergeOptions mergeOptions)
                     mergedLog.SchemaUri = mergedLog.Version.ConvertToSchemaUri();
 
                     FileSystem.DirectoryCreateDirectory(outputDirectory);
-                    outputFilePath = Path.Combine(outputDirectory, GetOutputFileName(_options, key));
+                    outputFilePath = Path.Combine(outputDirectory, ReplaceInvalidCharInFileName(GetOutputFileName(_options, key), "."));
```
yongyan-gh, line 124: >ReplaceInvalidCharInFileName [](http://example.com/codeflow?start=67&length=28)

If PerRule split strategy is specified, it creates a SarifLog file per rule, the file name is {ruleId}_merged.sarif.

If the ruleId is a sub rule Id, e.g. `SEC105/001`, it contains invalid file name char '/' and will cause saving the log file failed.

This is the fix for this issue.

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -179,7 +183,14 @@ private async Task<bool> MergeSarifLogsAsync()
                                     };
                                 }
 
-                                key = CreateRuleKey(result.RuleId, run);
+                                if (_options.MergeRuns)
```
yongyan-gh, line 186: >MergeRuns [](http://example.com/codeflow?start=45&length=9)

Acccording to `merge-runs` argument description
```
 [Option( 
     "merge-runs", 
     HelpText = "Merge runs of the same tool + verion combination (requires " + 
                "eliding run-specific details such as invocations data.")] 
 public bool MergeRuns { get; set; } 
```
If `merge-runs` is false (default), the merged log runs are aggregated by key `ruleId+tool name+tool version`.
if it is set to true, the merged log runs will be aggregated by key `tool name+tool version`. This means all results produced by same tool and same version will be merged into 1 single run.

src/Sarif.Multitool.Library/MergeCommand.cs
michaelcfanning: Where is the unit test for this issue? Did I miss it?

src/ReleaseHistory.md
```diff
@@ -1,5 +1,9 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fix issue `merge` command generates SARIF with multiple runs [#2488](https://github.com/microsoft/sarif-sdk/pull/2488)
```
michaelcfanning, line 5: >Fix [](http://example.com/codeflow?start=10&length=3)

This description does not clearly describe the issue and the fix. @marmegh or @easyrhinomsft can you please help Yong with this?

There's no reference in the release notes to the invalid characters issue.

src/Sarif.Multitool.Library/MergeCommand.cs
michaelcfanning: Yes, your understanding is correct. Do we really have the term `verion` burned into our help? If so, can you please correct this to `version`?

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -307,6 +318,11 @@
 
             return prefix ?? string.Empty;
         }
+
+        private static string ReplaceInvalidCharInFileName(string fileName, string replacement)
```
michaelcfanning, line 322: >ReplaceInvalidCharInFileName [](http://example.com/codeflow?start=30&length=28)

You provided no unit tests for this helper. Generally, a helper with a single line of code with a single upstream caller doesn't warrant dedicating its own method.

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -31,6 +32,9 @@
         private readonly Dictionary<string, HashSet<Result>> _ruleIdToResultsMap;
         private readonly Dictionary<string, RunMergingVisitor> _ruleIdToMergeVisitorsMap;
 
+        private static readonly string s_invalidFileNameCharRegexString = $"[{Regex.Escape(new string(Path.GetInvalidFileNameChars()))}]";
```
michaelcfanning, line 35: >s_invalidFileNameCharRegexString [](http://example.com/codeflow?start=39&length=32)

This functionality seems very general and not something at all specific to merging SARIF logs. Should it be somewhere else? A file utilities class? Should it simply be an extension method somewhere?

src/Test.UnitTests.Sarif.Multitool.Library/MergeCommandUnitTests.cs
```diff
@@ -0,0 +1,281 @@
...
+            };
+
+            for (int i = 1; i <= 6; i++)
+            {
+                mergedLogs[i - 1].Runs[0].Tool.Driver.Name.Should().Be("TestTool");
+                mergedLogs[i - 1].Runs[0].Results[0].RuleId.Should().Be($"TESTRULE/00{i}");
+            }
+        }
+
+        private static Run CreateTestRun(int numberOfResult, string toolName = null, bool createSubRule = false)
```
michaelcfanning, line 231: >CreateTestRun [](http://example.com/codeflow?start=27&length=13)

Don't we have other code somewhere that autogenerates test files? If so, can we please us it instead of creating this duplicative helper?

src/Test.UnitTests.Sarif.Multitool.Library/MergeCommandUnitTests.cs
```diff
@@ -0,0 +1,281 @@
...
+                mergedLog.Runs[i - 1].Tool.Driver.Name.Should().Be("TestTool");
+                mergedLog.Runs[i - 1].Results[0].RuleId.Should().Be($"TESTRULE00{i}");
+            }
+        }
+
+        [Fact]
+        public void MergeCommand_WhenSplitPerRule_LogShouldAggregateByRuleToolVersion()
+        {
+            // 2 logs 2 runs, same tool version, 6 rules
+            SarifLog sarifLog1 = new SarifLog { Runs = new[] { CreateTestRun(6, null, true) } };
```
michaelcfanning, line 166: >SarifLog [](http://example.com/codeflow?start=12&length=8)

this code doesn't meet our style standard, you should be using 'var' instead of an explicit type, because the type itself is clearly specified after the `new` keyword.

Why did this code pass build? Are we not enforcing this standard in our diagnostics configuration?

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -179,7 +183,14 @@
...
 
-                                key = CreateRuleKey(result.RuleId, run);
+                                if (_options.MergeRuns)
+                                {
+                                    key = CreateRuleKey(null, run);
+                                }
+                                else
+                                {
+                                    key = CreateRuleKey(result.RuleId, run);
+                                }
```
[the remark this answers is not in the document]
michaelcfanning: This is a good suggestion. I'd also improve this code to remove the standalone use of the literal 'null', generally providing a named parameter is preferable.

string ruleId = _options.MergeRuns? null : result.RuleId;
key = CreateRuleKey(ruleId, run);

michaelcfanning:
Um, this is an excellent question. We support multiple varieties of intra-logfile link, one of which is a simply reference to a location id (which I think should generally be stable as I believe it's associated with an id). The support for the SARIF schema, I also think, is mostly non-existent. And so I'd guess we have no code whatsoever to assist with this. Yong, we should create an issue for this?

---
In reply to: [1199064853](https://github.com/microsoft/sarif-sdk/pull/2513#issuecomment-1199064853) [](http://example.com/codeflow?ancestors=1199064853,1199064853)

src/ReleaseHistory.md
michaelcfanning: Has someone built/run your tests and examined code coverage? Do you have any information for us on the completeness of code & data coverage for your tests?

src/Sarif.Multitool.Library/MergeCommand.cs
yongyan-gh: dedicated unit tests added. pls review.

src/Test.UnitTests.Sarif.Multitool.Library/MergeCommandUnitTests.cs
yongyan-gh: Fixed. Its no enforced. In VS there are messages (not warning or erro) in error list showing this kind of issue.

src/Test.UnitTests.Sarif.Multitool.Library/MergeCommandUnitTests.cs
yongyan-gh: Moved the RandomSarifLogGenerator from Test.UnitTests.Sarif to Test.Utilities.Sarif so can re-use its functionalities.

src/Sarif.Multitool.Library/MergeCommand.cs
yongyan-gh: dedicated unit tests added pls review

src/Sarif.Multitool.Library/MergeCommand.cs
yongyan-gh: yes, moved the function to PathExtensions changed it to an extension method of string.

src/ReleaseHistory.md
yongyan-gh: I ran code coverage analysis in VS. All code paths I made change were covered. In term of `MergeCommand` coverage, it was 84%, added couple test cases its now 94%

src/Sarif.Multitool.Library/MergeCommand.cs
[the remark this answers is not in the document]
yongyan-gh: updated

src/ReleaseHistory.md
yongyan-gh: Updated pls review. cc @VCOLAWTON

src/Sarif.Multitool.Library/MergeCommand.cs
yongyan-gh: fixed typo.

yongyan-gh:
Right, the `merge` command does not update the pointers. Instead it calls `FixupVisitor` to resolve result's artifact location if it has index pointer points to artifacts array, and remove the artifact array. This needs further efforts, created issue [https://github.com/microsoft/sarif-sdk/issues/2520](https://github.com/microsoft/sarif-sdk/issues/2520) for tracking.


---
In reply to: [1200469594](https://github.com/microsoft/sarif-sdk/pull/2513#issuecomment-1200469594) [](http://example.com/codeflow?ancestors=1200469594,1199064853)

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
```
VCOLAWTON, line 1: Can you send me a few sentences that explain what this document is used for? I can refine the language for you.

VCOLAWTON (changes requested):
Need an overview, please see comment. Thanks, Yong! @yongyan-gh

src/ReleaseHistory.md
marmegh: So, let's look at what we're fixing here.

The problem: the merge command was generating a sarif with a run per rule instead of a run per tool.

The description here does not capture that. We need to tell the customers what behavior they may have been experiencing that is now changed. Specific first, concise second.

"Fixed `merge` command producing runs per rule to the expected runs per tool."

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
...
 
+## Unreleased
+
+* BUGFIX: Fix the issue `merge` command generates SARIF with multiple runs even specified `--merge-runs` option [#2488](https://github.com/microsoft/sarif-sdk/pull/2488)
+* BUGFIX: Fix `System.IO.IOException` and `System.IO.DirectoryNotFoundException` when merge Sarif files contains invalid file name characters in the rule id with `--split PerRule` option.
```
marmegh, line 6: I don't see any other references to these exceptions in this change? Was this broken out into one of your other PRs? Or was this included and not documented in a closed PR?

yongyan-gh: This error was not reported by any issue, but found during my testing for `merge` command. My other 2 PRs are for `match-result-forward` command. Its a straightforward fix, so I think it can be fixed along with the multiple runs issue.

src/ReleaseHistory.md
yongyan-gh: Thanks for the suggestions. Updated both, pls review.

src/ReleaseHistory.md
marmegh: Can you point me to the specific test implemented to prevent regression?

yongyan-gh: sure, in this test `MergeCommandUnitTests.MergeCommand_WhenSplitPerRule_LogShouldAggregateByRuleToolVersion', it creates sarif with id contains invalid char '/', e.g. 'TESTRULE/001', and the fix replaced the invalid char and generate valid file name e.g. 'TESTRULE.001_merged.sarif'

src/Sarif.Driver/PathExtensions.cs
```diff
@@ -19,6 +20,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
     /// <summary>Path extension functions.</summary>
     public static class PathExtensions
     {
+        private static readonly Regex s_invalidFileNameCharRegex = new Regex($"[{Regex.Escape(new string(Path.GetInvalidFileNameChars()))}]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
shaopeng-gh, line 23: >Path.GetInvalidFileNameChars() [](http://example.com/codeflow?start=105&length=30)

not an actual issue,
just to note that with current way we will possibly generate different name in Linux and Windows, 
and a name without any invalid char (win+Linux) would be 
working in both os, easier to share.

the other way of cause is to hard code the combination of them.
each has its pros and cons, just to call it out.

shaopeng-gh (approved):
:shipit:

shaopeng-gh:
Need to resolve the conflicts.

yongyan-gh:
resolved

---
In reply to: [1218848244](https://github.com/microsoft/sarif-sdk/pull/2513#issuecomment-1218848244) [](http://example.com/codeflow?ancestors=1218848244,1218848244)

src/Test.Utilities.Sarif/RandomSarifLogGenerator.cs
```diff
@@ -13,7 +13,7 @@
     /// <summary>
     /// Generate random sarif logs for testing.
     /// </summary>
-    internal static class RandomSarifLogGenerator
+    public static class RandomSarifLogGenerator
```
yongyan-gh, line 16: >public [](http://example.com/codeflow?start=4&length=6)

Move this helper to the common Teset.Utilities.Sarif project so it can be used by other test projects.

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## Unreleased
+
+* BUGFIX: Fixed `merge` command producing runs per rule to the expected runs per tool. [#2488](https://github.com/microsoft/sarif-sdk/pull/2488)
```
michaelcfanning, line 5: >er rule to the expected r [](http://example.com/codeflow?start=48&length=25)

this doesn't describe the adverse condition precisely or the state once resolved

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
...
+* BUGFIX: Fixed `System.IO.IOException` or `System.IO.DirectoryNotFoundException` when `merge` command producing per rule files.
```
michaelcfanning, line 6: >command producing per rule files [](http://example.com/codeflow?start=95&length=32)

I don't understand this

src/Sarif.Multitool.Library/MergeCommand.cs
```diff
@@ -117,7 +116,9 @@ public int Run(MergeOptions mergeOptions)
                     mergedLog.SchemaUri = mergedLog.Version.ConvertToSchemaUri();
 
                     FileSystem.DirectoryCreateDirectory(outputDirectory);
-                    outputFilePath = Path.Combine(outputDirectory, GetOutputFileName(_options, key));
+                    outputFilePath = Path.Combine(
+                        outputDirectory,
+                        GetOutputFileName(_options, key).ReplaceInvalidCharInFileName("."));
```
michaelcfanning, line 121: >GetOutputFileName [](http://example.com/codeflow?start=24&length=17)

Why are you doing this replacement? It doesn't seem like a good idea that we would arbitrarily create a file name that the user didn't request. Why didn't you simply throw on receiving an argument with an invalid char? Is there a good reason why we're trying to manage this adverse condition?

src/ReleaseHistory.md
michaelcfanning, line 5: >Fixed [](http://example.com/codeflow?start=10&length=5)

Use this:

Fix `merge` command failure to produce runs by tool and version when passed `merge-runs`.

src/ReleaseHistory.md
michaelcfanning, line 6: >Fixed `System.IO.IOException [](http://example.com/codeflow?start=10&length=28)

Please use this:

'Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids).

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        private readonly ITestOutputHelper _outputHelper;
+
+        public PathExtensionsTests(ITestOutputHelper outputHelper)
+        {
+            this._outputHelper = outputHelper;
+            outputHelper.WriteLine($"TestName: {nameof(PathExtensionsTests)} has random seed {randomSeed}");
+        }
+
+        [Theory]
```
michaelcfanning, line 34: >Theory [](http://example.com/codeflow?start=9&length=6)

Please stop introducing [Theory] usage into the code base, this functionality uses reflection and historically has caused performance issues.

All this is really buying is is the need not to inject a foreach in the method that calls into your helper.

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
```diff
@@ -0,0 +1,180 @@
...
+        [MemberData(nameof(TestData))]
+        [Trait(TestTraits.WindowsOnly, "true")]
+        public void ReplaceInvalidCharInFileName_ShouldCorrectFilePath(string fileName, string replacement, string expectFileName)
+        {
+            VerifyReplaceInvalidCharInFileName(fileName, replacement, expectFileName);
+        }
+
+        [Theory]
```
michaelcfanning, line 42: >Theory [](http://example.com/codeflow?start=9&length=6)

No more [Theory] usage.

michaelcfanning (approved):
:shipit:

src/ReleaseHistory.md
yongyan-gh: updated as suggested

src/ReleaseHistory.md
yongyan-gh: updated the note as suggested

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
yongyan-gh: changed to Fact pattern

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
yongyan-gh: Changed to [Fact] test pattern

PR #2515, opened 2022-07-29
Newtonsoft update.

michaelcfanning:
We recently introduced some unfortunate changes spread across several releases. First, we introduced many SDK breaking changes without properly versioning the assembly (e.g., release 2.4.9 should have reversioned as 2.5 to reflect its breaking change).

We recently introduced a requirement for the SARIF SDK to utilize Newtonsoft.JSON 13.0 at a minimum. This isn't appropriate for a component (unless the component itself directly provokes a security issue that can't be resolved unless running against the  most recent version). An API component should permit the lowest compatible utilization, while also protecting itself as far as possible from well-known exploitable issues. In this case, we take a targeted change that resolves GitHub advisory GHSA-5crp-9r3c-p9vr (if we observe that no other actor has initialized relevant default data). This is a useful defensive measure protecting naive consumers of our API. Any SDK consumers has the opportunity/capability to resolve this CVE in their own code (or to misconfigure the API to introduce the vulnerability). IOW, the ultimate security responsibility lies with the consumer. [Again, we minimize risk by proactively making the relevant fix ourselves, because that's feasible].

.NET framework support has been dropped for various down-level versions. This change actually increases the # of .net frameworks we allow linking to (for compatibility reasons). Again, our API does not have a strong motivation/accountability to force upgrade to .NET 4.8 (the current supported versions). 

The Newtonsoft.JSON upgrade strategy was as follows:
* Drop SDK min version requirement to 6.0.8 for .NET FX, 9.0.1 otherwise
* Enforce 12.0.1 for the work item assemblies (as there is an implied min-version for these components due to TFS utilizations).
* Reference 13.0 in test code (simply to validate that binding to the most recent version doesn't introduce any behavioral issues).

The down-level of Newstonsoft.JSON required some minor code updates, mostly along the lines of adding explicit notations where a more recent version allows for more efficient expression/invocation of the functionality (e.g., by specifying a camel-casing naming strategy, instead of requiring every method to express the desired json camel-cased property name, instead of picking up the conventional .NET pascal-cased property name).

@marmegh @cfaucon @EasyRhinoMSFT @eddynaka @hulonjenkins


src/Sarif/Sarif.csproj
```diff
@@ -20,7 +20,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;</TargetFrameworks>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net472;net461;net452</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net452;net462;net472;net48</TargetFrameworks>
```
michaelcfanning, line 23: >net462 [](http://example.com/codeflow?start=83&length=6)

So, I've updated our net461 target to net462. Will this cause user issues? Is it helpful to have both to ensure maximal back-compat?

scripts/Projects.psm1
```diff
@@ -10,7 +10,7 @@
 $Frameworks = @{}
 
 # .NET Framework versions for which we build.
-$Frameworks.NetFx = @("net452", "net461", "net472")
+$Frameworks.NetFx = @("net48")
```
michaelcfanning, line 13: >net48 [](http://example.com/codeflow?start=23&length=5)

I need to explore this script further to make sure it's correct. in-progress. TBD

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -19,6 +19,12 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// </summary>
     public class PropertyBagHolder : IPropertyBagHolder
```
michaelcfanning, line 20: >PropertyBagHolder [](http://example.com/codeflow?start=17&length=17)

This class is a base type for all our core OM. This means that any utilization of an OM type will invoke this static cctor.

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Processors
     [Serializable]
     public class SarifLogPipeline
     {
-        [JsonRequired]
+        //[JsonRequired]
```
michaelcfanning, line 18: >JsonRequired [](http://example.com/codeflow?start=11&length=12)

Still looking at this. Minimally, we can permit it for netstandard2.0 and elide otherwise.

src/Sarif/Sarif.csproj
eddynaka: Right now, only versions greater than 462 are supported by microsoft ([https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework](https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework)).

Based on that, should we keep net452 and net461?

You would only need to add net4xx if you have any special condition for those, otherwise, net framework is able to load those normally.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -9,7 +9,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net48</TargetFrameworks>
```
eddynaka, line 12: >net48 [](http://example.com/codeflow?start=76&length=5)

if we bump to net48, you won't be able to use sarifconverters from a netframework 462, for example.

in this case, my suggestion is to change to 462 (the lowest supported version).

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## UNRELEASED (will be 2.5.0)
```
marmegh, line 3: Can we start following this standard of noting when a minor version bump is needed if you're introducing a breaking change? Help prompt whoever may be coming along for the release.

scripts/Projects.psm1
EasyRhinoMSFT, line 13: >net48 [](http://example.com/codeflow?start=23&length=5)

Build.props has a version specified as well (4.7.2)

src/Sarif/Sarif.csproj
michaelcfanning: If I've understood you correctly, you've misconstrued the value we're trying to provide, which is to not require developers to recompile their code/update their package references. Make sense? Someone who compiles against net461 can successfully run on a net462 runtime deployment without recompilation. If we prevent those users from consuming our package because our metadata requires they explicitly retarget/recompile, we've broken builds and compromised the .NET compatibility feature. 

And so, it sounds odd, but we should be providing the lowest possible version that allow people to consume it and updates from that version. I'm actually wondering if this might do it.

<TargetFrameworks>netstandard2.0;net45;</TargetFrameworks>

[https://docs.microsoft.com/en-us/dotnet/standard/frameworks](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)

src/Sarif.Converters/Sarif.Converters.csproj
michaelcfanning: Yes, this was a mistake. What I'm going to do is drop this to net45. As mentioned previously, although net462 is the lowest supported version, it is not appropriate for our dependency to require users to upgrade to the supported version, that is their business decision to make.

src/Sarif/Sarif.csproj
michaelcfanning: OK, so looking at NuGet version data, it appears there are the explicitly specified versions and the inferred compatible versions. If you specify net472, for example, there is an inferred compatibility both with net48 but also down to net461 (there's some hard break at net461 that related to some specific platform, Mono or Silverlight or something).

If I'm correct, specifying net48 would work, except for users who tried to drop down to net45. I'm still going to proceed with and test net45, which I believe will imply compatibility from that version all the way through net48.

src/ReleaseHistory.md
michaelcfanning: Yes, agreed. We could consider reversioning immediate with a preview extension, as per semver. this would facilitate advance ship/testing as well.

EasyRhinoMSFT: Should we be following the Semantic Versioning standard? It states that breaking changes should prompt a major version increment.
[https://semver.org/](https://semver.org/)

marmegh: Adding some emphasis to Chris's recommendation above, breaking changes as major version increments was also expressed as a requirement moving forward for VS. Not sure if this is something we can unilaterally decide or if it needs to be reviewed with the TC, though.

scripts/BuildAndTest.ps1
```diff
@@ -192,10 +197,13 @@ if (-not $NoClean) {
 Install-VersionConstantsFile
 
 if (-not $NoRestore) {
-    Write-Information "Restoring NuGet packages for $SampleSolutionFile..."
-        & $NuGetExePath restore -ConfigFile $NuGetConfigFile -Verbosity $NuGetVerbosity -OutputDirectory $NuGetSamplesPackageRoot (Join-Path $SourceRoot $SampleSolutionFile)
-    if ($LASTEXITCODE -ne 0) {
-        Exit-WithFailureMessage $ScriptName "NuGet restore failed for $SampleSolutionFile."
+    if (-not $NoSamples) {
```
marmegh, line 200: I don't see an else for `-not $NoSamples` is there a reason these couldn't be combined as `if (-not $NoRestore -and -not $NoSamples)`?

scripts/BuildAndTest.ps1
```diff
@@ -214,7 +222,10 @@ if (-not $?) {
 if (-not $NoBuild) {
     Invoke-DotNetBuild $SolutionFile
     if ($ENV:OS) {
-        Invoke-DotNetBuild $sampleSolutionFile
+        if (-not $NoSamples) {
```
marmegh, line 225: Same idea as above, any reason these couldn't be combined into a single if statement?

scripts/Projects.psm1
marmegh: I see this is closed, did you have any notes besides the inline comments on your exploration?

src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -21,14 +21,14 @@
...
     <!-- The subfolder layout for the different TargetFrameworks is intentionally inconsistent -->
-    <!-- net472 and netcoreapp3.1 pack to tools\TargetFramework\* for backcompat -->
-    <file src="bld\bin\$platform$_$configuration$\Sarif.Multitool\Publish\net472\Sarif.Multitool.exe.config"
-          target="tools\net472" />
-    <file src="bld\bin\$platform$_$configuration$\Sarif.Multitool\Publish\net472\**"
-          target="tools\net472"
-          exclude="bld\bin\$platform$_$configuration$\Sarif.Multitool\Publish\net472\Sarif*;bld\bin\$platform$_$configuration$\Sarif.Multitool\Publish\net472\WorkItems.dll;" />
-    <file src="bld\bin\Signing\net472\**"
-          target="tools\net472" />
+    <!-- net462 and netcoreapp3.1 pack to tools\TargetFramework\* for backcompat -->
```
marmegh, line 24: Is this just to bring in line with our build settings for supported versions?

src/Sarif/Sarif.csproj
```diff
@@ -29,17 +29,26 @@
...
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="6.0.8" />
+        <Reference Include="System.Web" />
+        <Reference Include="System.Net.Http" />
+      </ItemGroup>
+    </Otherwise>
+  </Choose>
```
marmegh, line 45: Is Newtonsoft.Json guaranteed to be compatible with all of our supported versions besides netstandard 2.0?

Also, glad to see an example of this. @EasyRhinoMSFT, is this along the lines of what you were experimenting with the other week?

src/Sarif/Core/PropertyBagHolder.cs
marmegh, line 20: >PropertyBagHolder [](http://example.com/codeflow?start=17&length=17)

Looking forward, what would be the appropriate analog for this in Jschema? Do we have a comparable base class in that repo to extend this same type of fix to?

src/Sarif.Multitool/Sarif.Multitool.csproj
```diff
@@ -30,7 +30,10 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <!-- 
+      net462 is the current minimal supported .NET framework.
+    -->
```
marmegh, line 35: Is there a reason this comment is split to 3 lines? The others in the file appear to be single line comments.

src/Test.Utilities.Sarif/Test.Utilities.Sarif.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
marmegh, line 9: Why is this test project specifically using Newtonsoft 12.0? I see Test.UnitTests.Sarif.csproj is using 13.0 as of this commit.

src/WorkItems/WorkItems.csproj
```diff
@@ -29,11 +30,11 @@
...
     <PackageReference Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.20.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.2" />
-    <PackageReference Include="Microsoft.Json.Schema" Version="1.1.5" />
-    <PackageReference Include="Microsoft.Json.Schema.Validation" Version="1.1.5" />
+    <PackageReference Include="Microsoft.Json.Schema" Version="1.1.4" />
+    <PackageReference Include="Microsoft.Json.Schema.Validation" Version="1.1.4" />
     <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.170.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
marmegh, line 37: ><PackageReference Include="Newtonsoft.Json" Version="12.0.3" /> [](http://example.com/codeflow?start=4&length=63)

WorkItems doesn't need to be on v9.0?

src/ReleaseHistory.md
michaelcfanning: This is a good note, let's rev the major version before we ship.

src/Nuget/Sarif.Multitool.nuspec
michaelcfanning: this is to ensure that our own exes runs on the lowest possible fully supported .net fx.

src/Sarif/Core/PropertyBagHolder.cs
michaelcfanning: Great question. We must 'go see'.

src/Sarif/Core/PropertyBagHolder.cs
```diff
@@ -19,6 +19,12 @@ namespace Microsoft.CodeAnalysis.Sarif
...
     {
+        static PropertyBagHolder()
+        {
+            // Mitigation for Newtonsoft.Json v12 vulnerability GHSA-5crp-9r3c-p9vr
+            JsonConvert.DefaultSettings ??= () => new JsonSerializerSettings { MaxDepth = 64 };
```
michaelcfanning, line 25: I still need to verify that this setting is, in fact, null in the default case.

src/Sarif/Map/JsonMapNode.cs
```diff
@@ -17,25 +17,27 @@ namespace Microsoft.CodeAnalysis.Sarif.Map
     ///  structure of another JSON document compactly to enable constructing
     ///  subsets of it quickly.
     /// </summary>
-    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
```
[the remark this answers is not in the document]
michaelcfanning: It is precisely as you have inferred, the CamelCaseStrategy feature isn't available in downlevel newstonsoft.json, so I've implemented the equivalent functionality using this more verbose (but entirely equivalent) approach.

src/Sarif/Processors/Log/SarifLogPipeline.cs
```diff
@@ -15,7 +15,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Processors
     [Serializable]
     public class SarifLogPipeline
     {
+#if !NET45
```
michaelcfanning, line 18: >NET45 [](http://example.com/codeflow?start=5&length=5)

Make this #if NETSTANDARD_20 instead.

michaelcfanning:
>     <PackageReference Include="CommandLineParser" Version="2.7.82" />

The multitool itself should force an upgrade to 13.0

---
Refers to: src/Sarif.Multitool/Sarif.Multitool.csproj:40 in 2abc20f. [](commit_id = 2abc20f5021d62324dd7e30badb4f3f94bec52ae, deletion_comment = False)

eddynaka:
>     <TargetFrameworks>netcoreapp3.1;</TargetFrameworks>

For each project where targetFramework got changed, we should make a change here to see if everything works correctly and prevent more issues.

---
Refers to: src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj:11 in 2abc20f. [](commit_id = 2abc20f5021d62324dd7e30badb4f3f94bec52ae, deletion_comment = False)

src/WorkItems/WorkItems.csproj
michaelcfanning: It can't run on it, due to our utilization of Microsoft.TeamFoundationServer.Client, which requires 12.x

cfaucon (approved):
Feedback reviewed and provided offline.

scripts/NuGetUtilities.psm1
```diff
@@ -17,8 +17,8 @@ $NugetExePath = "nuget"
 if ($ENV:OS) {
     $NugetExePath = "$RepoRoot\.nuget\NuGet.exe"
 }
-$NuGetPackageRoot = Join-Path $SourceRoot "packages"
-$NuGetSamplesPackageRoot = Join-Path (Join-Path $SourceRoot "Samples") "packages"
+$NuGetPackageRoot = "$RepoRoot\bld\packages\"
+$NuGetSamplesPackageRoot = "$RepoRoot\src\Samples\packages\"
```
shaopeng-gh, line 21: Can you explain the reason of the change,
1. I believe packages is not under bld
2. would Join-Path syntax better? The current code may works in Linux but with Join-Path the benefit is don't have to think or test if \ actually works in Linux and macOS

src/ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## 2.5.0.beta-1
+* BUGFIX: Loosen Newtonsoft.JSON minimum version requirement to 6.0.8 (for .NET framework) or 9.0.1 (for all other compilations) for Sarif.Sdk. Sarif.Converts requires 8.0.1, minimally, for .NET framework compilations.
```
shaopeng-gh, line 4: >BUGFIX [](http://example.com/codeflow?start=2&length=6)

I think both upgrade and downgrade Newtonsoft.JSON is a BREAKING change,
if we want to downgrade,
A user using old version of Sarif SDK, which was 12.0.3 or 10.0.3
(ignore the reverted 13.0.1 change)
Depending on his configuration he may be actually using Newtonsoft.JSON 6.0.8, after the dependent bot/manually bumps the Sarif SDK package.
1. he may have code that have issue with Newtonsoft.JSON 6.0.8
2. he may have pipeline with security scan got pipeline blocked with security issues between 6.0.8 and 12.0.3

src/Sarif/Core/PropertyBagHolder.cs
EasyRhinoMSFT: https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_MaxDepth.htm

src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -21,14 +21,14 @@
...
+    <file src="bld\bin\$platform$_$configuration$\Sarif.Multitool\Publish\net462\Sarif.Multitool.exe.config"
```
shaopeng-gh, line 25: >\net462\ [](http://example.com/codeflow?start=73&length=8)

I tried get the branch, cleaning everything, build, this is the folder I have:
sarif-sdk\bld\bin\AnyCPU_Release\Sarif.Multitool\net472
There is no net462 folder.
Some other changes needed to generate to the net462 instead?

shaopeng-gh:
I tried build and then check the \sarif-sdk\bld\Publish\Nuget\Release\Sarif.Sdk.2.4.16.nupkg,
this is in the Sarif.Sdk.nuspec inside, with **13.0.1**:

```
<dependencies>
      <group targetFramework=".NETFramework4.5.2">
        <dependency id="Newtonsoft.Json" version="13.0.1" exclude="Build,Analyzers" />
        <dependency id="System.Collections.Immutable" version="5.0.0" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETFramework4.6.1">
        <dependency id="Newtonsoft.Json" version="13.0.1" exclude="Build,Analyzers" />
        <dependency id="System.Collections.Immutable" version="5.0.0" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETFramework4.7.2">
        <dependency id="Newtonsoft.Json" version="13.0.1" exclude="Build,Analyzers" />
        <dependency id="System.Collections.Immutable" version="5.0.0" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Newtonsoft.Json" version="13.0.1" exclude="Build,Analyzers" />
        <dependency id="System.Collections.Immutable" version="5.0.0" exclude="Build,Analyzers" />
      </group>
    </dependencies>
```

src/Samples/SarifDeferredSample/SarifDeferredSample.csproj
```diff
@@ -34,8 +34,8 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Newtonsoft.Json, Version=13.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\Newtonsoft.Json.13.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
+    <Reference Include="Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\packages\Newtonsoft.Json.11.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
```
shaopeng-gh, line 38: >Newtonsoft.Json.11.0.2 [](http://example.com/codeflow?start=28&length=22)

to <Reference> version 11
need to have <PackageReference> for that version

<PackageReference Include="Newtonsoft.Json" Version="" />

src/Sarif/Sarif.csproj
michaelcfanning: Yes, compatible.

src/Sarif/Core/PropertyBagHolder.cs
michaelcfanning: Good thought. The question is whether JsonConvert.DefaultSettings is null by default or not. The doc doesn't say.

Based on your help info, though, I can make a defensive code fix, thanks very much.

[https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonConvert_DefaultSettings.htm](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonConvert_DefaultSettings.htm)

michaelcfanning: OK, forgot this thing is a delegate. It *is*, in fact, null on start-up. My thought around initializing MaxDepth defensively was bogus, since the settings object is the output of a call to the delegate, when non-null. Anyway, I confirmed this code is correct.

src/Sarif.Multitool/Sarif.Multitool.csproj
michaelcfanning: The reason is 'I am only human' and prone to inconsistencies. :) Fixed, thanks for the note.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -37,6 +37,21 @@
...
+    <When Condition="$(TargetFramework) == 'netstandard2.0'">
+      <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="8.0.1" />
+        <Reference Include="System.Web" />
+        <Reference Include="System.Net.Http" />
```
shaopeng-gh, line 50: these should be not needed

src/Test.Utilities.Sarif/Test.Utilities.Sarif.csproj
michaelcfanning: Good question! The reason is to get diversity of utilization of Newtonsoft.Json.

I've updated all the samples to consume multiple versions as well. I'll comment this code.

scripts/NuGetUtilities.psm1
michaelcfanning: OK, great note. Couldn't get the join-path syntax working previously and hard-coded this change. Let me revert and see whether everything still works.

Thanks again, this occurred to me previously (thanks to your advocacy on this issue) and then I forgot about the potential issue again. :)

src/Sarif/Core/PropertyBagHolder.cs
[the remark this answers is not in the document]
michaelcfanning: JsonReader.MaxDepth is instance data which is, as far as I know, initialized either from an internal Newtonsoft default *or* as a result of creating an alternate default settings object using JsonConvert.DefaultSettings. 

Let me go take a look at the Newtonsoft code, though, to prove that assumption or prove me wrong. :)

scripts/Projects.psm1
shaopeng-gh: set as "Active" for marmegh's comment for better visibility.

src/Sarif/Core/PropertyBagHolder.cs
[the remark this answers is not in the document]
shaopeng-gh: if it is don' t think fix needed, just to note.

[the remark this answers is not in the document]
michaelcfanning: OK, look at the following code. The default serializer acquires a default settings object from JsonConvert. During serialization, the serializer will override the JsonReader max depth if it differs from the default setting. This approach has existed for several years.

Please review yourself and confirm whether you think I'm right about things. :)

[https://github.com/JamesNK/Newtonsoft.Json/blob/94ff24f815a777269920f5a7b13f60db50c5eeb7/Src/Newtonsoft.Json/JsonSerializer.cs#L623](https://github.com/JamesNK/Newtonsoft.Json/blob/94ff24f815a777269920f5a7b13f60db50c5eeb7/Src/Newtonsoft.Json/JsonSerializer.cs#L623)

[https://github.com/JamesNK/Newtonsoft.Json/blob/94ff24f815a777269920f5a7b13f60db50c5eeb7/Src/Newtonsoft.Json/JsonSerializer.cs#L958](https://github.com/JamesNK/Newtonsoft.Json/blob/94ff24f815a777269920f5a7b13f60db50c5eeb7/Src/Newtonsoft.Json/JsonSerializer.cs#L958)

src/ReleaseHistory.md
michaelcfanning: OK, not sure I understand this. If the only dependency on Newtonsoft comes from the SDK assembly, we're ok because only the SDK functionality is using it.

If there is another dependency on Newtonsoft, then that dependency *must* bet at least what it was before, say 12 or 13.0. If it isn't, the build won't work.

And so, when we drop the version, that already existing dependency version will still exist and SARIF will *still* be updated to the more recent version.

scripts/Projects.psm1
michaelcfanning: Nope, the comments stand. Let me know if they're insufficiently clear.

Also, I should stop 'closing' items and 'resolve' them so others can close. Sorry for this lapse in process! I just did it to one of your notes as well, Shaopeng. :)

michaelcfanning:
>     <TargetFrameworks>netcoreapp3.1;</TargetFrameworks>

I don't understand this comment, sorry.

---
In reply to: [1203218610](https://github.com/microsoft/sarif-sdk/pull/2515#issuecomment-1203218610) [](http://example.com/codeflow?ancestors=1203218610,1203218610)

---
Refers to: src/Test.UnitTests.Sarif/Test.UnitTests.Sarif.csproj:11 in 2abc20f. [](commit_id = 2abc20f5021d62324dd7e30badb4f3f94bec52ae, deletion_comment = False)

src/Samples/SarifDeferredSample/SarifDeferredSample.csproj
michaelcfanning: indeed, thanks, fixed up!

src/ReleaseHistory.md
shaopeng-gh: If the only dependency on Newtonsoft comes from the SDK assembly  ---- my understanding is if so, unless the project set explicitly otherwise, we control the Newtonsoft version, and by using our sdk, if there is code interacting with the input and output json or json class we produce, it will be code that use Newtonsoft dll.

to repo, create an empty c# project, only add SARIF SDK package. 
Newtonsoft.Json v13 is added with the current published version.

At this point, if his code use V13 feature and works on V13, he checks it in.
We downgrade,
his build pipeline will get V6, it may break.

michaelcfanning: OK, so how did 'his code use V13 features' without adding an explicit reference to V13 in that separate code project? You have to directly reference Newtonsoft in order to consume the V13 API. Am I still misunderstanding? Can you try to create a repro/example for me to look at?

src/Sarif/Core/PropertyBagHolder.cs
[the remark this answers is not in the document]
shaopeng-gh: is that JsonReader internal inside serializer
I see this was the PR
https://github.com/JamesNK/Newtonsoft.Json/pull/2462
the change both in the file JsonReader.cs and JsonSerializerSettings
```
protected JsonReader()
        {
            _currentState = State.Start;
            _dateTimeZoneHandling = DateTimeZoneHandling.RoundtripKind;
            _dateParseHandling = DateParseHandling.DateTime;
            _floatParseHandling = FloatParseHandling.Double;
            _maxDepth = 128;

            CloseInput = true;
        }
```
the usage of the JsonReader is like:
`using (JsonTextReader jsonReader = new JsonTextReader(streamReader))`
it may not related to the serializer.
(what the source code is like in V6 I have not looked)

Bear with me if I am wrong :)

src/ReleaseHistory.md
shaopeng-gh: I mean when you ref a package, all dependency of the package will be added to the project, you can start using all functions in all of the dependency dlls. You do not need to manually add any version of Newtonsoft , if you add ref to SARIF SDK.

image

to repo, create an empty c# project, only add SARIF SDK package, 

image

shaopeng-gh: And then if look at the .csproj it only has:
`<PackageReference Include="Sarif.Sdk" Version="2.4.16" />`

image

marmegh (approved):
:shipit:

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -37,6 +37,19 @@
...
 
+  <Choose>
+    <When Condition="$(TargetFramework) == 'netstandard2.0'">
+      <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json" Version="8.0.1" />
```
shaopeng-gh, line 50: >8.0.1 [](http://example.com/codeflow?start=61&length=5)

here we are not using 6, but 8 
I see the reason is we have a attribute that does not exists in 8 and will not build.
If not for this reason we will be also using 6 like other projects.
Let me know if I understand correctly.

michaelcfanning: Yes, exactly. The HDF JSON definitions in particular consume a new attribute enumeration property that doesn't exist in down-level Newtonsoft.Json (it is an attribute that specifies that a property is *not* required but also cannot be null when present). Accepting 6.0.8 support would have required an intrusive change to resolve this and also would have compromised enforcement for this specific serialization condition. We can discuss more offline, now that this PR is merged.

shaopeng-gh: Accepting 6.0.8 support would have required an intrusive change to resolve  --- agree, just wanted to make sure I understand the reason of the choice we made.

PR #2526, opened 2022-08-04
Fix release pipeline error

marmegh:
The release pipeline is failing while running `scripts\BuildPackagesFromSigningDirectory.ps1` with this error:
```
2022-08-03T23:27:08.2825834Z Could not find a part of the path 'D:\a\r1\a\bld\bin\Signing\netstandard2.1'.
2022-08-03T23:27:08.3945948Z BuildPackagesFromSigningDirectory: Sarif.Multitool NuGet package creation failed.
2022-08-03T23:27:08.3949421Z BuildPackagesFromSigningDirectory FAILED.
2022-08-03T23:27:08.4872582Z ##[error]PowerShell exited with code '1'.
```
The same error is reproducible locally and resolves with updating the below nuspec file to remove the obsolete reference.

image

This change is limited to fixing the nuget release pipeline.

### DIFF
--- a/src/Nuget/Sarif.Multitool.nuspec
+++ b/src/Nuget/Sarif.Multitool.nuspec
@@ -38,8 +38,6 @@
           target="tools\netcoreapp3.1\any" />
     <file src="bld\bin\Signing\netstandard2.0\**"
           target="tools\netcoreapp3.1\any" />
-    <file src="bld\bin\Signing\netstandard2.1\**"
-          target="tools\netcoreapp3.1\any" />
 
     <file src="triskelion.png" />
     <file src="src\ReleaseHistory.md" />


EasyRhinoMSFT (approved):
:shipit:

marmegh:
@EasyRhinoMSFT and @michaelcfanning, do we want to bump the version number and treat this as a whole new release? Or do we want to absorb this into the beta1 release as there are no changes here outside of unblocking the build pipeline? What is the best practice?

src/Nuget/Sarif.Multitool.nuspec
```diff
@@ -38,8 +38,6 @@
           target="tools\netcoreapp3.1\any" />
     <file src="bld\bin\Signing\netstandard2.0\**"
           target="tools\netcoreapp3.1\any" />
-    <file src="bld\bin\Signing\netstandard2.1\**"
```
michaelcfanning, line 41: >netstandard2 [](http://example.com/codeflow?start=31&length=12)

btw - the need for this change indicates some technical debt in our early build/test scripts. we could/should consider a richer build (that compiles this package and also the NPM package) that runs at PR time.

michaelcfanning:
We only reversion once we successfully publish. Until that moment, there is no need to revise, as no customer can take a dependency on the changed bits. Also, in this case, there's no literal behavior change/customer impact. We simply are bug-fixing a broken process. In general, our source control history is used for developer tracking purposes, and our release notes and actual published version are for customer tracking purposes.

---
In reply to: [1204621340](https://github.com/microsoft/sarif-sdk/pull/2526#issuecomment-1204621340) [](http://example.com/codeflow?ancestors=1204621340,1204621340)

PR #2527, opened 2022-08-05
Beta2

michaelcfanning:
- Reversion to beta2. 
- Drop SARIF SDK NETFX support to net461 (as this is the minimal version that can be compiled with a downloadable Microsoft dev pack).
- Drop proactive security mitigation fix for the core SDK assembly (as updating this code in the static constructor causes type load exceptions on executing this static constructor in some test scenarios, where the assemblies are loaded into a new AppDomain where binding redirects appear not to be applied in an expected way).

@cfaucon  @EasyRhinoMSFT @marmegh @shaopeng-gh @eddynaka

### DIFF
--- a/scripts/Projects.psm1
+++ b/scripts/Projects.psm1
@@ -10,10 +10,10 @@
 $Frameworks = @{}
 
 # .NET Framework versions for which we build.
-# net45 is the minimal framework we build the SDK against.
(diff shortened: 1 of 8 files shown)


EasyRhinoMSFT (approved):
:shipit:

PR #2533, opened 2022-08-15
Lowering System.Collections.Immutable to 1.5.0

eddynaka:
This change will drop to a previous version of `System.Collections.Immutable` that we used to use (1.5.0).
We did this update nearly two years ago: https://github.com/microsoft/sarif-sdk/pull/2146 but we didn't have any reason to do it.

Approach:
- Removed the dependency from the test projects since the tests can use the binaries from the actual product
- Lowered the dependency from the product
- Ran tests

Tests:
- All tests are passing
- BinSkim tests are passing
- Sarif.PatternMatcher tests are passing

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,9 +1,12 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v3.1.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/3.1.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/3.1.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/3.1.0) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/3.1.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/3.1.0) 
+
+* DEPENDENCY BREAKING: SARIF.SDK now requires `System.Collections.Immutable` 1.5.0. [#2504](https://github.com/microsoft/sarif-sdk/pull/2533)
+
 ## **v3.0.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/3.0.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/3.0.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/3.0.0) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/3.0.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/3.0.0)
 * BUGFIX: Loosen Newtonsoft.JSON minimum version requirement to 6.0.8 (for .NET framework) or 9.0.1 (for all other compilations) for Sarif.Sdk. Sarif.Converts requires 8.0.1, minimally, for .NET framework compilations.
 * BUGFIX: Broaden set of supported .NET frameworks for compatibility reasons. Sarif.Sdk, Sarif.Driver and Sarif.WorkItems requires net461.
-* BUGFIX: Set default stack limit in Newtonsoft.JSON utilization (if `JsonConvert.Defaults` is not already configured) to address GitHub advisory [GHSA-5crp-9r3c-p9vr](https://github.com/advisories/GHSA-5crp-9r3c-p9vr).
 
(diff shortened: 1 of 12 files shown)


src/ReleaseHistory.md
```diff
@@ -1,9 +1,12 @@
...
+
+* DEPENDENCY BREAKING: SARIF.SDK now requires `System.Collections.Immutable` 1.5.0. [#2504](https://github.com/microsoft/sarif-sdk/pull/2533)
+
 ## **v3.0.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/3.0.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/3.0.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/3.0.0) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/3.0.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/3.0.0)
 * BUGFIX: Loosen Newtonsoft.JSON minimum version requirement to 6.0.8 (for .NET framework) or 9.0.1 (for all other compilations) for Sarif.Sdk. Sarif.Converts requires 8.0.1, minimally, for .NET framework compilations.
 * BUGFIX: Broaden set of supported .NET frameworks for compatibility reasons. Sarif.Sdk, Sarif.Driver and Sarif.WorkItems requires net461.
-* BUGFIX: Set default stack limit in Newtonsoft.JSON utilization (if `JsonConvert.Defaults` is not already configured) to address GitHub advisory [GHSA-5crp-9r3c-p9vr](https://github.com/advisories/GHSA-5crp-9r3c-p9vr).
```
eddynaka, line 6: >BUGFIX: Set default stack limit in Newtonsoft.JSON utilization [](http://example.com/codeflow?start=2&length=62)

Removing this because we are not setting any limit anymore.

src/ReleaseHistory.md
```diff
@@ -1,9 +1,12 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **v3.1.0-beta1** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/3.1.0-beta1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/3.1.0-beta1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/3.1.0-beta1) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/3.1.0-beta1) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/3.1.0-beta1)
+
+* DEPENDENCY BREAKING: SARIF.SDK now requires `System.Collections.Immutable` 1.5.0. [#2504](https://github.com/microsoft/sarif-sdk/pull/2533)
```
marmegh, line 5: Is there a compatibility issue between 1.5.0 and 5.0.0? If 5.0.0 was previously required, is this really a breaking change or are we actually loosening restrictions for the sarif sdk?

marmegh:
Can you explain in the description the reason for downgrading this dependency? Is it for VS compatibility like the newtonsoft downgrade? Is there an external motivation for this? Or are we simply applying a principle of lowest required version for all dependencies?

src/ReleaseHistory.md
eddynaka: we have other systems that uses sarif-sdk which is causing major breaking changes in other projects.

So, with that in mind, I want to reduce to the previous version that we were using.

src/ReleaseHistory.md
marmegh: We're no longer setting a limit? Which PR was this removed in and why?

eddynaka: This got removed in this:
[https://github.com/microsoft/sarif-sdk/pull/2527](https://github.com/microsoft/sarif-sdk/pull/2527)

eddynaka: We discussed today in the IcMs meeting about this. The point is: the consume will need to fix and the sarif.sdk, since its the lower level, it cannot enforce the fix on anyone.

src/Test.FunctionalTests.Sarif/Test.FunctionalTests.Sarif.csproj
```diff
@@ -49,7 +49,6 @@
     <PackageReference Include="FluentAssertions" Version="5.10.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.0-preview-20220707-01" />
     <PackageReference Include="Moq" Version="4.13.1" />
-    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
```
eddynaka, line 52: >System.Collections.Immutable [](http://example.com/codeflow?start=31&length=28)

I'm removing the dependency from the tests, since it will use the same version of the product. This will facilitate new updates in the future if required.

src/ReleaseHistory.md
marmegh: Got it. Thanks for the context.

marmegh (approved):
:shipit:

EasyRhinoMSFT (approved):

PR #2540, opened 2022-09-07
Fix a test which fails randomly

yongyan-gh:
# Description
Test `SarifLogTests.SarifLog_SplitPerRun` fails randomly. Can be reproduced using random seed 1529710912.

This is because the `RandomSarifLogGenerator` may generate test Sarif log run contains 0 result. But test assumes all runs have non-empty results.


shaopeng-gh (approved):
:shipit:

EasyRhinoMSFT (approved):

michaelcfanning (approved):
:shipit:

PR #2542, opened 2022-09-08
FEATURE: Add `ignore-non-fatal-runtime-conditions` argument that allows program to exit as code `0` instead of `1`.

shaopeng-gh:
## Problem description

Today, when we run Binskim in AzureDevOps, if the target folder is empty, it will generate a runtime error called `NoValidAnalysisTargets`.

With that, it will return exit code as `1`.
This means that it will 'break' the run since for Guardian, everything that isn't returning `0` is a failure.

## Proposal

Similar to what we have last time for `ExceptionLoadingPdb`: 
Log the issue in the SARIF and return code `0`.

This time we will use a more generic flag so that we don't need to add another flag if there is new similar request.
`--ignore-non-fatal-runtime-conditions NoValidAnalysisTargets;XXX`

## Current behavior

Log the issue in the SARIF and return code `1`.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -6,6 +6,7 @@
 * BUGFIX: Remove duplicated rule definitions when executing `match-results-forward` commands for results with sub-rule ids. [#2486](https://github.com/microsoft/sarif-sdk/pull/2486)
 * BUGFIX: Update `merge` command to properly produce runs by tool and version when passed the `--merge-runs` argument. [#2488](https://github.com/microsoft/sarif-sdk/pull/2488)
 * BUGFIX: Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids).
+* FEATURE: Add `ignore-non-fatal-runtime-conditions` argument that allows program to exit as code `0` instead of `1`. Initial supported value `NoValidAnalysisTargets`. [#2542](https://github.com/microsoft/sarif-sdk/pull/2542)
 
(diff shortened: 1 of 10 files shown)


src/ReleaseHistory.md
```diff
@@ -6,6 +6,7 @@
 * BUGFIX: Remove duplicated rule definitions when executing `match-results-forward` commands for results with sub-rule ids. [#2486](https://github.com/microsoft/sarif-sdk/pull/2486)
 * BUGFIX: Update `merge` command to properly produce runs by tool and version when passed the `--merge-runs` argument. [#2488](https://github.com/microsoft/sarif-sdk/pull/2488)
 * BUGFIX: Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids).
+* FEATURE: Add `ignore-non-fatal-runtime-conditions` argument that allows program to exit as code `0` instead of `1`. Initial supported value `NoValidAnalysisTargets`. [#2542](https://github.com/microsoft/sarif-sdk/pull/2542)
```
michaelcfanning, line 9: >FEATURE [](http://example.com/codeflow?start=2&length=7)

There's a core problem with the way this feature is expressed. Currently, 'No valid analysis targets' *is* a fatal condition. That's why we return 1 instead of 0. And so, all the language around the command-line argument, etc. is confusing.

This condition is fatal due to historical problems with misconfiguring security-related static analysis tools. The classic problem is a failure to omit the --recursive switch. It is a devastating outcome when teams discover that a security analysis deployed for months or more, which is a component of satisfying internal security policy, meeting federal security certification standards, etc., simply hasn't been executing.

And so, in order to be secure-by-default, we fail in the common case where a misconfigured command-line returns no valid analysis targets. It is generally trivial for scripting/other environments to perform simple checks (such as 'does this directory contain any files?') which conditionally lead to BinSkim execution (or not).

Further, the tool does have a mechanism where users can selectively ignore this condition: they simply invoke the rich return code option and process that return value, ignoring the `NoValidAnalysisTargets` (0x400) bit and any other *actual* non-fatal conditions observed.

michaelcfanning (changes requested):
:clock1:

src/ReleaseHistory.md
shaopeng-gh: Thanks, synced with Mary we will close this PR for now,
will reopen if changes are needed after discussion with Guardian.

PR #2546, opened 2022-09-15
Add output warning and runtime condition when files are skipped due to size

michaelcfanning:
@shaopeng-gh, do you have time to add tests and finish this PR?

Currently, our driver tools silently fail when skipping a file due to size limits. We need to warn users of this circumstance.

Please make sure these warnings end up in logs, as tool configuration warnings.


shaopeng-gh:
> @shaopeng-gh, do you have time to add tests and finish this PR?
> 
> Currently, our driver tools silently fail when skipping a file due to size limits. We need to warn users of this circumstance.
> 
> Please make sure these warnings end up in logs, as tool configuration warnings.

@michaelcfanning sure,

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -439,6 +445,11 @@
 
             _hashChannel.Writer.Complete();
 
+            if (_ignoredFilesCount > 0)
```
shaopeng-gh, line 448: >Warnings.LogOneOrMoreFilesSkippedDueToSize(rootContext); [](http://example.com/codeflow?start=16&length=56)

@michaelcfanning question, do you think we should also change the same in file `AnalyzeCommandBase.cs` ?
If so I can work on the change together.

shaopeng-gh:
Added a test following current pattern.

---
In reply to: [1248354509](https://github.com/microsoft/sarif-sdk/pull/2546#issuecomment-1248354509) [](http://example.com/codeflow?ancestors=1248354509,1248354509)

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning: For sure if we're missing this elsewhere, let's fix it. Here's a release note entry:

FEATURE: provide explicit logging (and new runtime condition flag) when any file is skipped due to exceeding file size limits.

shaopeng-gh: Got it, working on it.

shaopeng-gh: Synced with Mary, working on that internal BinSkim rule request first, will return to this once PR is published.

shaopeng-gh (approved):
:shipit:

PR #2547, opened 2022-09-29
Update WorkItemsResources.resx

markphip:
When users do not have a Scans tab available they are unclear how to discover more information. Directing them to view the Sarif file will allow them to discover more details

### DIFF
--- a/src/Sarif.WorkItems/WorkItemsResources.resx
+++ b/src/Sarif.WorkItems/WorkItemsResources.resx
@@ -143,6 +143,6 @@
     <value> (and other locations)</value>
   </data>
   <data name="WorkItemBodyTemplateText" xml:space="preserve">
-    <value>This work item contains {0} {1} issue(s) detected in {2}{3}. Click the 'Scans' tab to review results.</value>
+    <value>This work item contains {0} {1} issue(s) detected in {2}{3}. Click the 'Scans' tab to review results. If the 'Scans' tab is not present, please open the attached SARIF in one of the viewers below.</value>
   </data>
 </root>


src/Sarif.WorkItems/WorkItemsResources.resx
```diff
@@ -143,6 +143,6 @@
     <value> (and other locations)</value>
   </data>
   <data name="WorkItemBodyTemplateText" xml:space="preserve">
-    <value>This work item contains {0} {1} issue(s) detected in {2}{3}. Click the 'Scans' tab to review results.</value>
+    <value>This work item contains {0} {1} issue(s) detected in {2}{3}. Click the 'Scans' tab to review results. If the 'Scans' tab is not present, please open the attached SARIF in one of the viewers below.</value>
```
michaelcfanning, line 146: >If [](http://example.com/codeflow?start=113&length=2)

This is certainly an improvement. We should also consider adding a link to the ADO extension that provides the scans tab, to the default template body.

michaelcfanning (approved):
:shipit:

src/Sarif.WorkItems/WorkItemsResources.resx
michaelcfanning: Mark, sorry about this, I made you a repo contributor, can you branch off main and submit a second PR? Our pipeline won't accept PRs from forks, for security reasons and our policy, currently, is to strictly require this pipeline validation for all changes. We will need to look more carefully at this situation.

@jeffersonking can you provide me the link to the ADO extension? so we can add it to the work item body?

markphip: No problem. Will close this and create a PR. I was not sure if this was supposed to be ADO specific based on some of what I had read. The proposed text is what I had seen suggested.

PR #2549, opened 2022-10-06
Resolve 'collection modified' exception when logging results

michaelcfanning:
This change introduces thread-safe utilization of a single CachingLogger instance in cases when multiple files with the same file hash are analyzed.

In these circumstances, the scan should not occur again, rather, the cached results (if any) should be updated (to reflect the current file path of the duplicated file) and 'replayed' into the log.

This is tricky because the scan pipeline design is intended for every context object (per scan target) to be accessed only on a single thread.

To make this work, we introduce a slim semaphore in the caching logger, which allows access to only a single thread at a time. This semaphore's `Wait` method is invoked on receiving the notice that a new scan target is being analyzed.

We also have a new ReleaseLock method, which is called after scan completes. This method releases the semaphore and also puts the caching logger into a 'committed' state, meaning that the first copy, by file hash, of a set of duplicated targets has been analyzed and we now have a fixed set of results that can be 'replayed' into the log.

@suvamM

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -28,7 +28,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         internal bool _captureConsoleOutput;
 
         internal ConsoleLogger _consoleLogger;
-        internal ConcurrentDictionary<string, IAnalysisLogger> _analysisLoggerCache;
+        internal ConcurrentDictionary<string, CachingLogger> _analysisLoggerCache;
 
         private Run _run;
         private Tool _tool;
@@ -265,15 +265,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                         while (context?.AnalysisComplete == true)
                         {
-                            if (_computeHashes)
-                            {
-                                bool cache = _analysisLoggerCache.TryGetValue(context.Hashes.Sha256, out IAnalysisLogger logger);
-                                LogCachingLogger(rootContext, logger ?? context.Logger, context, clone: cache);
-                            }
-                            else
-                            {
-                                LogCachingLogger(rootContext, context.Logger, context);
-                            }
+                            LogCachingLogger(rootContext, context, clone: _computeHashes);
 
                             RuntimeErrors |= context.RuntimeErrors;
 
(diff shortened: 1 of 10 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -190,12 +189,18 @@
                 : (Debugger.IsAttached) ? 1 : Environment.ProcessorCount;
 
             var channelOptions = new BoundedChannelOptions(2000)
+            {
+                SingleWriter = true,
+                SingleReader = true,
+            };
+            readyToHashChannel = Channel.CreateBounded<int>(channelOptions);
```
michaelcfanning, line 196: >readyToHashChannel [](http://example.com/codeflow?start=12&length=18)

Currently the hashing channel is single-threaded. There is an opportunity here, I think, the improve performance by making this phase multi-threaded.

src/Sarif/Writers/CachingLogger.cs
```diff
@@ -24,6 +28,14 @@ public CachingLogger(IEnumerable<FailureLevel> levels, IEnumerable<ResultKind> k
 
         public IList<Notification> ToolNotifications { get; set; }
 
+        /// <summary>
+        /// Gets or sets a boolean value that indicates whether the Results
+        /// object associated with this logger is fixed and ready to replay.
+        /// </summary>
+        public bool CacheFinalized { get; private set; }
```
marmegh, line 35: >CacheFinalized [](http://example.com/codeflow?start=20&length=14)

Nice!

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -190,12 +189,18 @@ private void Analyze(TOptions analyzeOptions, AggregatingLogger logger)
...
```
marmegh, line 196: Should we have a line break here and six lines down to improve readability?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -206,26 +211,39 @@ private void Analyze(TOptions analyzeOptions, AggregatingLogger logger)
...
 
-            Task<bool> hashFiles = HashAsync();
-            Task<bool> findFiles = FindFilesAsync(options, rootContext);
-            Task<bool> writeResults = WriteResultsAsync(rootContext);
-
-            // FindFiles is single-thread and will close its write channel
-            findFiles.Wait();
-            hashFiles.Wait();
+            // 4: A single-threaded consumer watches for completed scans
+            //    and logs results, if any. This operation is singlle-threaded
```
marmegh, line 233: >singlle [](http://example.com/codeflow?start=62&length=7)

typo -> `single`

marmegh (approved):
:shipit:

PR #2554, opened 2022-10-13
Revert removal of _persistArtifacts variable

HulonJenkins:


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -442,7 +448,7 @@ private void CaptureArtifact(ArtifactLocation fileLocation)
 
             // Ensure Artifact is in Run.Artifacts and ArtifactLocation.Index is set to point to it
             int index = _run.GetFileIndex(fileLocation,
-                                          addToFilesTableIfNotPresent: true,
+                                          addToFilesTableIfNotPresent: _persistArtifacts || hashData != null,
```
HulonJenkins, line 451: >|| hashData != null [](http://example.com/codeflow?start=89&length=19)

Do we want this? Or just the plain _persistArtifacts

PR #2559, opened 2022-10-17
Updating file paths to be relative instead of using $(SolutionDir)

eddynaka:
## Description

This change is updating file paths to be relative to the current project instead of using $(SolutionDir).

## Why do we need this change?

If you use the sarif.sdk project as submodule and you want to include some unit test projects, it won't compile because the $(SolutionDir) will be different from what it expects, generating an issue like:
```
error MSB3030: Could not copy the file "C:\Microsoft\sarif-pattern-matcher-legacy\src\Test.FunctionalTests.Sarif\v2\SpecExamples\Comprehensive.sarif" because it was not f
ound. [C:\Microsoft\sarif-pattern-matcher-legacy\src\sarif-pattern-matcher\Src\sarif-sdk\src\Test.UnitTests.Sarif\Test.UnitTests.Sarif.csproj::TargetFramework=netcoreapp3.1]
```

### DIFF
--- a/src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
+++ b/src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
@@ -25,7 +25,7 @@
 
   <Target Name="CopyFunctionalTestData" AfterTargets="Build">
     <ItemGroup>
-      <TestFiles Include="$(SolutionDir)FunctionalTestData\**\*" />
+      <TestFiles Include="..\FunctionalTestData\**\*" />
     </ItemGroup>
     <Copy SourceFiles="@(TestFiles)" DestinationFolder="$(OutputPath)\FunctionalTestData\%(RecursiveDir)" />
   </Target>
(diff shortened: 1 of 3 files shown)


src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
```diff
@@ -25,7 +25,7 @@
 
   <Target Name="CopyFunctionalTestData" AfterTargets="Build">
     <ItemGroup>
-      <TestFiles Include="$(SolutionDir)FunctionalTestData\**\*" />
+      <TestFiles Include="..\FunctionalTestData\**\*" />
```
michaelcfanning, line 28: >FunctionalTestData [](http://example.com/codeflow?start=29&length=18)

A more advanced change would be to use a special property that is $(SolutionDir) by default but which is overwritten when this repo is included as a submodule.

We can take this improved change later, if there's a need.

michaelcfanning (approved):
:shipit:

PR #2572, opened 2022-11-08
* BREAKING: Properties changed from `int`(32bit) to `BigInteger`(unlimited).

shaopeng-gh:
* BREAKING: Properties changed from `int`(32bit) to `BigInteger`(unlimited): `Address.AbsoluteAddress`, `Address.RelativeAddress`, `Address.Length`, `Address.OffsetFromParent`, `Artifact.Offset`, `Artifact.Length`, `Invocation.ExitCode`, `Invocation.ExitSignalNumber`, `LocationRelationship.Target`, `Region.ByteOffset`, `Region.ByteLength`, `Result.OccurrenceCount`. 

We had made the change before for 1 property, this change is to use the new feature in JS Schema to batch change all we choose to use Big Integer. Unless in the schema definition file it has limit set, all JSON `number` are be generated as `BigInteger`.

As we make the schema definition same as the C# generated, we need to first to make sure the schema file has the correct limitations set if we need it to be int 32.

We set all of those index type properties to int 32 because in C# both array and list are int32 index based.


src/ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## **v3.1.0** (UNRELEASED)
 
+* BREAKING: Properties changed from `int`(32bit) to `BigInteger`(unlimited): `Address.AbsoluteAddress`, `Address.RelativeAddress`, `Address.Length`, `Address.OffsetFromParent`, `Artifact.Offset`, `Artifact.Length`, `Invocation.ExitCode`, `Invocation.ExitSignalNumber`, `LocationRelationship.Target`, `Region.ByteOffset`, `Region.ByteLength`, `Result.OccurrenceCount`. [#2572](https://github.com/microsoft/sarif-sdk/pull/2572)
```
shaopeng-gh, line 5: >`Invocation.ExitCode`, `Invocation.ExitSignalNumber` [](http://example.com/codeflow?start=215&length=52)

Let me know if we want to use int32 instead for:
`Invocation.ExitCode`, `Invocation.ExitSignalNumber`
I don't think there is clear indication of limitation in SARIF spec,
e.g. if user has a method that has return value defined as C# `BigInteger`  and he use that return value as `ExitSignalNumber`.
So I currently use `BigInteger`

src/Sarif/ToDotNet/ToDotNet.targets
```diff
@@ -28,8 +28,8 @@
...
-    <Exec Command="dotnet &quot;$(JsonSchemaToDotNetPath)&quot; --schema-name Sarif --schema-file-path &quot;%(JsonSchemaFile.FullPath)&quot; --output-directory &quot;$(_ObjectModelOutputDirectory)&quot; --force-overwrite --namespace-name %(JsonSchemaFile.Namespace) --root-class-name %(JsonSchemaFile.RootClassName) --copyright-file-path &quot;%(JsonSchemaFile.CopyrightFilePath)&quot; --hints-file-path &quot;%(JsonSchemaFile.HintsFilePath)&quot; --generate-cloning-code --protected-init-methods --virtual-members" Condition="$(OS) != 'Windows_NT'" />
+    <Exec Command="&quot;$(JsonSchemaToDotNetPath)&quot; --schema-name Sarif --schema-file-path &quot;%(JsonSchemaFile.FullPath)&quot; --output-directory &quot;$(_ObjectModelOutputDirectory)&quot; --force-overwrite --namespace-name %(JsonSchemaFile.Namespace) --root-class-name %(JsonSchemaFile.RootClassName) --copyright-file-path &quot;%(JsonSchemaFile.CopyrightFilePath)&quot; --hints-file-path &quot;%(JsonSchemaFile.HintsFilePath)&quot; --generate-cloning-code --protected-init-methods --virtual-members --generate-json-integer-as auto" Condition="$(OS) == 'Windows_NT'" />
```
shaopeng-gh, line 31: > --generate-json-integer-as auto [](http://example.com/codeflow?start=508&length=32)

the batch change is triggered from here

src/Sarif/Visitors/SarifCurrentToVersionOneVisitor.cs
```diff
@@ -202,9 +202,9 @@ private FileDataVersionOne CreateFileDataVersionOne(Artifact v2FileData)
                 fileData = new FileDataVersionOne
                 {
                     Hashes = CreateHashVersionOneListFromV2Hashes(v2FileData.Hashes),
-                    Length = v2FileData.Length == -1 ? 0 : v2FileData.Length,
+                    Length = v2FileData.Length == -1 ? 0 : (int)v2FileData.Length,
```
shaopeng-gh, line 205: >Length = v2FileData.Length == -1 ? 0 : (int)v2FileData.Length, [](http://example.com/codeflow?start=20&length=62)

we will still not be able to convert to v1 without error if the data is too large since we are not changing v1.
So I just use direct cast here.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1009.IndexPropertiesMustBeConsistentWithArrays_Invalid.sarif
```diff
@@ -123,7 +123,7 @@
                   "index": 0
                 },
                 "region": {
-                  "startLine": 154,
+                  "startLine": 155,
```
shaopeng-gh, line 126: I have looked,
the 155 line is the correct line that is 
`"index": 2`
so the after change is correct.

src/ReleaseHistory.md
```diff
@@ -5,6 +5,7 @@
...
 * BREAKING: Correct casing of `LogMissingreportingConfiguration` helper to `LogMissingReportingConfiguration`. [#2599](https://github.com/microsoft/sarif-sdk/pull/2599)
 * BREAKING: Change type of `MaxFileSizeInKilobytes` from int to long in `IAnalysisContext` and other classes.  [#2599](https://github.com/microsoft/sarif-sdk/pull/2599)
+* BREAKING: Properties changed from `int`(32bit) to `BigInteger`(unlimited): `Address.AbsoluteAddress`, `Address.RelativeAddress`, `Address.Length`, `Address.OffsetFromParent`, `Artifact.Offset`, `Artifact.Length`, `Invocation.ExitCode`, `Invocation.ExitSignalNumber`, `LocationRelationship.Target`, `Region.ByteOffset`, `Region.ByteLength`, `Result.OccurrenceCount`. [#2572](https://github.com/microsoft/sarif-sdk/pull/2572)
 * BREAKING: For `Guid` properties defined in SARIF spec, updated Json schema to use `uuid`, and updated C# object model to use `Guid?` instead of `string`. [#2555](https://github.com/microsoft/sarif-sdk/pull/2555)
```
suvamM, line 9: @shaopeng-gh Curious: are there any performance implications for switching from Int32 to BigInteger?

src/Samples/SarifToCsv/CsvWriter.cs
```diff
@@ -101,6 +102,11 @@ public void Write(int value)
             Write(value.ToString());
         }
 
+        public void Write(BigInteger value)
+        {
+            Write(value.ToString());
```
michaelcfanning, line 107: >value [](http://example.com/codeflow?start=18&length=5)

is there an overload for specifying the invariant culture here>

michaelcfanning: Prefer CultureInfo.InvariantCulture unless default culture is already getting picked up (IOW don't introduce a breaking change in behavior).

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
```diff
@@ -97,14 +97,16 @@
           "description": "The index within run.addresses of the cached object for this address.",
           "type": "integer",
           "default": -1,
-          "minimum": -1
+          "minimum": -1,
+          "maximum": 2147483647
```
michaelcfanning, line 101: >2147483647 [](http://example.com/codeflow?start=21&length=10)

You can't do this, it's not specified.

michaelcfanning, line 101: >2147483647 [](http://example.com/codeflow?start=21&length=10)

Create a second file to specify .NET

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1009.IndexPropertiesMustBeConsistentWithArrays_Invalid.sarif
michaelcfanning: Please figure out why this changed.

src/Test.UnitTests.Sarif/Query/EvaluatorTests.cs
```diff
@@ -57,6 +61,7 @@ public void ComplexEvaluator_Basics()
                 item.ID = i;
                 item.State = (State)(i % 4);
                 item.Uri = i.ToString();
+                item.AbsoluteAddress = BigInteger.Parse(ulong.MaxValue.ToString() + 1000000 * i);
```
michaelcfanning, line 64: >1000000 * i) [](http://example.com/codeflow?start=84&length=12)

lift this int a variable and tostring it, or use parens to clarify the code.

michaelcfanning (approved):
:shipit:

src/Samples/SarifToCsv/CsvWriter.cs
shaopeng-gh: the soucre code is 
```csharp
public override String ToString() {
            return BigNumber.FormatBigInteger(this, null, NumberFormatInfo.CurrentInfo);
        }
```

src/ReleaseHistory.md
michaelcfanning: This is good information. Allocating a StringBuilder instance is a very heavyweight operation at local code execution scope.

We should definitely test deserialization behavior slowdown. You'd likely benefit from testing a synthetic file, i.e., create a SARIF file that very intensively produces BigInteger properties (not a real world example).

src/ReleaseHistory.md
michaelcfanning: Excellent points. I have the exact same opinion.

src/Samples/SarifToCsv/CsvWriter.cs
michaelcfanning: Got it. So we should actually not use this helper. The reason is that you're persisting to CSV, where we should prefer a stable format.

Imagine someone who produces this CSV on a German system and reading it in the US. Commas and decimals in code may be reversed and cause problems.

Instead of ToString(), use this: BigNumberFormatBigInteger(value, null, NumberFormatInfo.InvariantInfo.

src/Samples/SarifToCsv/CsvWriter.cs
```diff
@@ -101,6 +102,11 @@ public void Write(int value)
             Write(value.ToString());
```
michaelcfanning, line 102: >value [](http://example.com/codeflow?start=18&length=5)

Please use

value.ToString(CultureInfo.InvariantCulture)

here. Also, please review any other numeric values written in this code to scrub for the same problem.

src/Samples/SarifToCsv/CsvWriter.cs
shaopeng-gh: I checked we have codes like 
`region.EndLine.ToString(CultureInfo.InvariantCulture)`
I used the same.
This file is actually a sample project code.

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
shaopeng-gh: fixed to use rtm.5.dotnet
since we are using a version for dotnet specific implementation, I think we don't need to bump the version number to 6 now.

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1009.IndexPropertiesMustBeConsistentWithArrays_Invalid.sarif
shaopeng-gh: still looking this one

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/ExpectedOutputs/SARIF1005.UriMustBeAbsolute_Invalid.sarif
```diff
@@ -59,7 +59,7 @@
                 },
                 "region": {
                   "startLine": 2,
-                  "startColumn": 53
+                  "startColumn": 60
```
shaopeng-gh, line 62: >60 [](http://example.com/codeflow?start=33&length=2)

added `.dotnet` which is 7 in length.

Issue #2531, opened 2022-08-14
Feature: Nessus & CIS CAT Converter


ejohn20:
New converter transforming:

* CIS CAT Pro JSON into SARIF
* Nessus XML schema into SARIF

michaelcfanning:
We believe SARIF has succeeded sufficiently to warrant direct authoring, not new or updated converters. Closing as not planned — please reopen if a compelling current need emerges.

PR #2574, opened 2022-11-08
FEATURE: Nessus + CIS CAT Converters

ejohn20:
Creates a new converter + unit tests for Nessus (XML format) and CIS CAT (JSON format) covered in https://github.com/microsoft/sarif-sdk/issues/2531


yongyan-gh:
@ejohn20 thanks for sharing the sample CisCat/Nessus SARIF files for review. Have couple suggestions pls review

yongyan-gh:
1. Both converted CisCat and Nessus SARIF files' results don't contain `location` object in `locations` property. 

>>>
GitHub Advanced Security code scanning will not display a result unless it provides a location that specifies the URI of the artifact that contains the result.

If plan to be ingested by GHAS pls consider generate a location for the results.

yongyan-gh:
2. Considering adding a help URI for each rule through `helpUri` property of the rule object if possible.

Providing a URI where users can find detailed information about the rule helps users to understand the result and how they can best address it. The varies SARIF viewers can render a hyperlink for users easily navigate to the Uri contains detailed information.

E.g. the rule id `RUSTSEC-2019-0001` has a help Uri '[https://rustsec.org/advisories/RUSTSEC-2019-0001](https://rustsec.org/advisories/RUSTSEC-2019-0001)'

yongyan-gh:
3. Please provide version information through 'version', 'semanticVersion', 'dottedQuadFileVersion' properties of `driver` object for Nessus SARIF log.

Providing version information enables the log file consumer to determine whether the file was produced by an up to date version, and to avoid accidentally comparing log files produced by different tool versions.

src/Sarif.Converters/CisCatConverter.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            //Set the tool details
+            run.Tool = new Tool();
+            run.Tool.Driver = CreateDriver(log);
+
+            //Set the list of tool rules
+            run.Tool.Driver.Rules = new List<ReportingDescriptor>();
+            foreach (CisCatRule rule in log.Rules)
+            {
+                run.Tool.Driver.Rules.Add(CreateReportDescriptor(rule));
```
yongyan-gh, line 43: >run.Tool.Driver.Rules.Add(CreateReportDescriptor(rule)); [](http://example.com/codeflow?start=16&length=56)

Suggest to only log a rule if its referenced by a result and avoid logging duplicated rules.

This can be done in the results loop below and checking if the rule already exists in the `log.Rules' list.

michaelcfanning:
Picking this up to see how close it is to landing.

**Status now:**

* Merges `dev` cleanly; full solution builds with 0 warnings, 0 errors.
* Pushed a baseline-refresh commit — the converter tests fail purely on stale `` URL and string-array formatting introduced by serializer changes since 2023. After regenerating the 5 expected SARIF files via the multitool, `Test.UnitTests.Sarif.Converters` filtered to the two converters is 10/10 green.

**Still open (from @yongyan-gh's Jan-2023 review):**

1. **Results have no `locations[]`.** GHAS code scanning won't render results without an artifact location. Nessus has the host (`targetId`) and CIS CAT has the benchmarked system; either could synthesize a `physicalLocation.artifactLocation.uri` (e.g. `host://<hostname>`) so the results are renderable.
2. **Rules have no `helpUri`.** Nessus has CVEs available — `https://nvd.nist.gov/vuln/detail/<CVE>` is a natural mapping. CIS CAT could point at the CIS Workbench rule URL.
3. **Nessus driver is missing version metadata.** `version` is best-effort from `sc_version`; `semanticVersion` and `dottedQuadFileVersion` are unset. CIS CAT already sets both from `BenchmarkVersion`.

Each is a small, well-bounded change.

@ejohn20 — are you in a position to address these and update the PR, or would you like us to take it over?

ejohn20:
@michaelcfanning Wow, it's been a minute since we opened this 😂 If you've got bandwidth, feel free to take it over. 

I've got a deadline a couple weeks out before I'd have time to finish this one out. Let me know if you'd rather wait and I'll get it on my radar.

michaelcfanning:
Right, @ejohn20, can you believe it? Me and my AI friends just thought maybe this backlog needed clearing. :)  A general point about converters (and a rationale for closing some items in the backlog) is that SARIF is pretty clearly an industry standard today, supported by cross-platform tooling, and the GH and MS alert management systems. And it's worth asking, do we need a converter or should we push for direct export? 

Ongoing rationales are strongest for government related standards, historical commercial standards with deep maturity, etc., where rate of change or simply the load of back-compat/ROI to change argue otherwise. I think these rationales exist for your work, would be interested to get your confirmation/additional thoughts there.

I'll see if I can get this cleaned up and turn you into a reviewer. We'll want human eyes on this line-by-line for reasons that don't need stating.

michaelcfanning:
@ejohn20 picking this up to close out the three review items from the original review — thanks for the offer to hand it off. Pushed and synced with `dev` (clean merge, converter tests green). Summary of what changed and the judgment calls worth a look:

**Review items addressed**
1. **Results now carry `locations[]`.** Nessus uses `host://<hostName>`; CIS CAT uses `benchmark://<benchmarkId>`.
2. **Nessus rules carry `helpUri`** → the Tenable plugin page (`https://www.tenable.com/plugins/nessus/<pluginId>`).
3. **Nessus driver carries `version`/`semanticVersion`** from the scan's `sc_version` server preference.

**Judgment calls (please challenge any of these)**
- **CIS CAT location = `benchmark://<benchmarkId>`** rather than a host. The CIS-CAT object model carries no per-host coordinate (only benchmark/profile/rules), so the assessed benchmark is the closest stable artifact identity.
- **CIS CAT `helpUri` skipped.** I could not derive a reliable public per-rule URL from `rule-id`. Left unset rather than guess.
- **Nessus `helpUri` = Tenable plugin page, not an NVD CVE link.** A rule maps 1:1 to a plugin (always present); CVEs are 0..many per plugin, so the plugin page is the dependable target.
- **Nessus version fields = `version` + `semanticVersion` only** (no `dottedQuadFileVersion`); `sc_version` is not reliably an `n.n.n.n` quad, and CIS CAT doesn't set it either.

**One behavioral note on the baselines:** CIS CAT persists via the `Run` overload, so `AddFileReferencesVisitor` interns its synthesized location into `run.artifacts` (with an `index` and a normalized trailing-slash URI). Nessus persists via the `SarifLog` overload and is not interned. That asymmetry is pre-existing in the two converters' persist paths — I left it as-is rather than expand scope here, but flagging in case you'd prefer them consistent.

Shipping under **v5.0.10** (5.0.9 has shipped).

---
_[This comment authored with Copilot]_

michaelcfanning:
Merged latest `dev` and resolved the `ReleaseHistory.md` conflict. Heads up on the version: this branch had introduced a standalone **v5.0.11** section, but `dev` has since moved to **v5.1.0** (`build.props` `VersionPrefix`). I folded the `NessusConverter` / `CisCatConverter` `NEW:` bullet into the existing **v5.1.0** section and dropped the v5.0.11 header; the merge takes `dev`'s 5.1.0 `build.props`. Code merged cleanly. Built `Sarif.Converters` and ran `NessusConverterTests` + `CisCatConverterTests` (10/10 pass). Now MERGEABLE.

_[This comment authored with Copilot]_

PR #2530, opened 2022-08-12
Feature: Suppression Expression Support

ejohn20:
Pull request adds the features described in #2529

Issue #2529, opened 2022-08-12
Feature: Suppression Expression Support


ejohn20:
The SARIF multitool `suppress` command adds a suppression to all results in the SARIF file. Enhancements to the SARIF Multitool `suppress` command should support:

* Providing one to many guid(s) (e.g. instance ids) for suppressing individual instances
* Providing an expression to suppress all results matching a given rule id, baseline state (new), and other supported expression fields
* Support the ability to query for results that are suppressed (or not suppressed), and include results that have expired based on the `expiryUtc` property value using the `--expiryInDays` argument.

PR #2575, opened 2022-11-08
Feature: Suppression Expression Support (#2530)

ejohn20:
Pull request adds the features described in https://github.com/microsoft/sarif-sdk/issues/2529


ejohn20:
@michaelcfanning Still working through a few modifications required due to the result guid changing from a string to a Guild?. I'll report back when this is stable.

msandhaus:
@ejohn20 , any way to push this along? It seems the `IsSuppressed` option snuck into the [docs](https://github.com/microsoft/sarif-sdk/blob/main/docs/query-mode.md) anyhow. We were really hoping to use that as a filter.

PR #2576, opened 2022-11-15
Coyote testing

suvamM:
This PR makes the following changes:

1. Moves Coyote testing out of the standard SDK tests, and is encapsulated into its own class, and uses `MSTest` instead of `xUnit`. Coyote tests now are in the category `CoyoteTest` and are skipped during normal `BuildAndTest` since these are long running.
2. Moving Coyote tests to its own class and using `MSTest` improves the logging. 
3. The `BuildAndTest` has a new parameter to run Coyote tests in isolation.
4. A GitHub action is added, which runs Coyote on the `CoyoteTest` tests upon a pull request.


scripts/BuildAndTest.ps1
```diff
@@ -218,11 +223,21 @@ if (-not $NoBuild) {
     }
 }
 
+if ($CoyoteTesting) {
+echo "Run Coyote Tests"
```
eddynaka, line 227: >echo [](http://example.com/codeflow?start=0&length=4)

nit: align

scripts/BuildAndTest.ps1
```diff
@@ -218,11 +223,21 @@ if (-not $NoBuild) {
...
+   dotnet test --filter TestCategory=CoyoteTest $SourceRoot\$SolutionFile --configuration Release
```
eddynaka, line 228: >dotnet test [](http://example.com/codeflow?start=3&length=11)

Similar to the command bellow, can we add --no-build?

.github/workflows/coyote-testing.yml
```diff
@@ -0,0 +1,19 @@
+# Run Coyote for systematically testing multithreaded code
+name: "Coyote Testing"
+
+on:
+  pull_request:
+    branches: [ main ]
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
```
eddynaka, line 10: >jobs [](http://example.com/codeflow?start=0&length=4)

Do we need to specify the image? windows-latest?

scripts/BuildAndTest.ps1
```diff
@@ -218,11 +223,21 @@ if (-not $NoBuild) {
...
+echo "Run Coyote Tests"
+   dotnet test --filter TestCategory=CoyoteTest $SourceRoot\$SolutionFile --configuration Release 
+    if ($LASTEXITCODE -ne 0) {
+        Exit-WithFailureMessage $ScriptName "Tests failed."
+    }
+    exit 0
+}
+
 if (-not $NoTest) {
+    echo "Normal build and test"
```
eddynaka, line 236: >Normal build [](http://example.com/codeflow?start=10&length=12)

We are just testing in this code path.

Issue #2569, opened 2022-11-03
Adding support for partialFingerprints


mpoqq:
It would be nice if a partialFingerprint can be generated and added to the SARIF log file.

There are several ways of calculating such a partial fingerprint. The CodeQL Team approach is to hash the context around the location (see https://github.com/github/codeql-action/blob/main/src/fingerprints.ts).

Since the logic to add context region snippets already exists, I would recommend using the same approach and just hash the snippet.

I have an almost running version and I am also willing to contribute.

PR #2577, opened 2022-11-16
Adding support for partialFingerprints

mpoqq:
Adding partialFingerprint inseret option. Hash contextRegion and add value to sarif. MD5 is used to hash the context. https://github.com/microsoft/sarif-sdk/issues/2569


mpoqq:
I can add a section to [docs/multitool-usage.md](https://github.com/microsoft/sarif-sdk/blob/main/docs/multitool-usage.md) about the rewrite modes and all insert options (hashes, TextFiles, Snippets, PartialFingerprints, ...). Alternatively, we could improve the documentation about the rewrite mode in a seperate issue/pr.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -253,6 +237,32 @@ public override Result VisitResult(Result node)
...
 
+            if (_dataToInsert.HasFlag(OptionallyEmittedData.PartialFingerprints))
+            {
+                Location primaryLocation = node.Locations?.FirstOrDefault();
+                if (primaryLocation != null)
+                {
+                    PhysicalLocation physicalLocation = primaryLocation.PhysicalLocation;
+                    Uri resolvedUri = GetResolvedArtifactLocationUri(physicalLocation.ArtifactLocation);
+
+                    Region expandedRegion = _fileRegionsCache.PopulateTextRegionProperties(physicalLocation.Region, resolvedUri, false);
```
yongyan-gh, line 248: >Region expandedRegion [](http://example.com/codeflow?start=20&length=21)

Since this code runs after `base.VisitResult(node);`, the region/contextRegion snippets may have already been created.

You can check whether `physicalLocation.Region.Snippet` already exists to avoid computing the value again.

Then this comes with a design question: if `OptionallyEmittedData.PartialFingerprints` is flagged, does it implies `OptionallyEmittedData.RegionSnippets` and `OptionallyEmittedData.ContextRegionSnippets` also are flagged? since to populate the partial fingerprint it has to read the region and context region.

@michaelcfanning what do you think?

src/Sarif/OptionallyEmittedData.cs
```diff
@@ -79,6 +79,9 @@ public enum OptionallyEmittedData : int
 
         // Enrich SARIF log with git blame information
         GitBlameInformation = 0x1000,
+        
+        // Enrich SARIF log with partial fingerprint
+        PartialFingerprints = 0x2000,
```
yongyan-gh, line 84: >PartialFingerprints [](http://example.com/codeflow?start=8&length=19)

PartialFingerprints can contains multiple entries, the context region hash is just one of them.

The enum entry name should be more specified, e.g. "ContextRegionSnippetHashFingerprints"

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
```diff
@@ -253,6 +237,32 @@ public override Result VisitResult(Result node)
...
+                    Region contextRegion = _fileRegionsCache.ConstructMultilineContextSnippet(expandedRegion, resolvedUri);
+
+                    if (contextRegion?.Snippet?.Text != null)
+                    {
+                        string contextRegionHash = HashUtilities.ComputeStringSha256Hash(contextRegion.Snippet.Text);
+
+                        node.PartialFingerprints ??= new Dictionary<string, string>();
+
+                        if (!node.PartialFingerprints.ContainsKey(ContextRegionHash) ||
+                            _dataToInsert.HasFlag(OptionallyEmittedData.OverwriteExistingData))
```
yongyan-gh, line 258: The opposite of these conditions can be placed at beginning so it can avoid unnecessary computing the region snippets and hashes.

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
mpoqq: > design question
I think we should not also emit `ContextRegionSnippets` and `RegionSnippets` because the sarif file will be unnecessarily bloated. The user may want to use the fingerprint only for result matching and do not want to send "Code" to a result management system.

src/Sarif/OptionallyEmittedData.cs
mpoqq: What do you think about `ContextRegionSnippetPartialFingerprints`? I would like to have `PartialFingerprints` in the name to find this feature faster.

yongyan-gh: Sounds good!

src/Sarif/Visitors/InsertOptionalDataVisitor.cs
yongyan-gh: @mpoqq had a discussion with the team, this is a valid scenario. Thanks for bringing it up.

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -246,6 +253,13 @@ public void InsertOptionalDataVisitor_ContextRegionSnippets_DoesNotFail_TopLevel
                 OptionallyEmittedData.ContextRegionSnippets);
         }
 
+        [Fact]
+        public void InsertOptionalDataVisitor_PartialFingerprint_DoesNotFail_TopLevelOriginalUriBaseIdUriMissing()
```
yongyan-gh, line 257: >InsertOptionalDataVisitor_PartialFingerprint_DoesNotFail_TopLevelOriginalUriBaseIdUriMissing [](http://example.com/codeflow?start=20&length=92)

seems this is same as above test case. please lets know if its not. we try to not add test SARIF files if not necessary, to avoid increasing tests run time since these tests need disk I/O.

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
mpoqq: Yes you are right, it is almost the same.

mpoqq:
@microsoft-github-policy-service agree [company="Vector Informatik GmbH"]

mpoqq:
@microsoft-github-policy-service agree company="Vector Informatik GmbH"

michaelcfanning:
@mpoqq, thanks for the contribution. I'm adding you as a write contributor to the repo.

Can you create a branch off main with this contribution? Our workflow will then run and we can merge. If that's a bother, we can take care of the merge ourselves. If you do what I suggest, however, your GitHub badges will properly update to reflect your contribution to this project. :)

mpoqq:
@michaelcfanning Thank you for adding me as contributor. I created a new branch and pull request https://github.com/microsoft/sarif-sdk/pull/2610.

So I guess I can now close this pull request and you will take over the merge process of the new pull request.

mpoqq:
Resolved in https://github.com/microsoft/sarif-sdk/pull/2610

PR #2578, opened 2022-11-17
FEATURE: Allow external set of `MaxFileSizeInKilobytes`

shaopeng-gh:
The SDK user https://github.com/microsoft/sarif-website is not able to set the `MaxFileSizeInKilobytes`.
It is limited to 1024KB.
With this enabled, SDK users should be able to set it as needed.
e.g.
``` CSharp
var validateOptions = new ValidateOptions
                {
                    OutputFilePath = outputFilePath,
                    Force = true,
                    PrettyPrint = true,
                    Level = new FailureLevel[] {
                        FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note
                    },
                    Kind = new ResultKind[] {
                        ResultKind.Fail, ResultKind.Informational, ResultKind.NotApplicable, ResultKind.Open, ResultKind.Open, ResultKind.Pass, ResultKind.Review
                    },
                    ConfigurationFilePath = configFilePath,
                    TargetFileSpecifiers = new string[] { inputFilePath },
                    // MaxFileSizeInKilobytes = XXX,
                };
```


src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -119,6 +119,6 @@ public abstract class AnalyzeOptionsBase : CommonOptionsBase
             "max-file-size-in-kb",
             HelpText = "The maximum file size (in kilobytes) that will be analyzed.",
             Default = 1024)]
```
shaopeng-gh, line 121: >1024 [](http://example.com/codeflow?start=22&length=4)

I also have a question here, we probably have a context so we added this parameter.

The other way is,
unless user specifically limit it, we don't put limit.
If so the default can be int.Max.

Or a reasonable bigger default value like 128MB.

Let me know if you like to change it and what value I can update the PR.

EasyRhinoMSFT: I think changing the default would be a breaking change. So perhaps just making the setter public is the right fix.

shaopeng-gh: I think changing the default would be a breaking change --- Yes!

EasyRhinoMSFT (approved):

michaelcfanning (approved):
:shipit:

PR #2581, opened 2022-12-06
FEATURE: Add `--normalize-for-ghas` argument to the `rewrite` command

shaopeng-gh:
This adds the function to rewrite a SARIF file to meet requirement of GitHub / GHAS.


src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -12,42 +13,19 @@ public class GitHubIngestionVisitor : SarifRewritingVisitor
...
-                        if (result.Level == FailureLevel.Error)
-                        {
-                            errors.Add(result);
-                        }
-
-                        if (errors.Count == s_MaxResults) { break; }
-                    }
-
-                    node.Results = errors;
-                }
```
shaopeng-gh, line 49: please also see the test file "NonErrorResults.sarif'.
This does not seem right, why do we want to remove warning results?

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -161,6 +138,32 @@ public override Result VisitResult(Result node)
             return base.VisitResult(node);
         }
 
+        public override Message VisitMessage(Message node)
+        {
+            if (node.Text == null)
```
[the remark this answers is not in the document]
shaopeng-gh: this is form GitHub, I think GitHub requires .Text only

src/ReleaseHistory.md
```diff
@@ -12,6 +12,7 @@
 * BUGFIX: Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids). [#2513](https://github.com/microsoft/sarif-sdk/pull/2513)
 * BUGFIX: Fix classes inside NotYetAutoGenerated folder missing `virtual` keyword for public methods and properties, by regenerate and manually sync the changes. [#2537](https://github.com/microsoft/sarif-sdk/pull/2537)
 * FEATURE: Allow external set of `MaxFileSizeInKilobytes`, which will allow SDK users to change the value. (Default value is 1024) [#2578](https://github.com/microsoft/sarif-sdk/pull/2578)
+* FEATURE: Add `--normalize-for-github` argument to the `rewrite` command to get GitHub supported SARIF format. [#2581](https://github.com/microsoft/sarif-sdk/pull/2581)
```
michaelcfanning, line 15: Good release note item but let me add a bit more detail, try this:

* FEATURE: Add `--normalize-for-github` argument to the `rewrite` command to ensure rewritten SARIF is compatible with GitHub Advanced Security ingestion requirements. [#2581](https://github.com/microsoft/sarif-sdk/pull/2581)

src/Sarif/Visitors/GitHubIngestionVisitor.cs
michaelcfanning: Perhaps GitHub only used to process errors? Have you tested this in their system?

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -12,47 +13,24 @@
...
 
         public override Run VisitRun(Run node)
         {
+            this.run = node;
             this.artifacts = node.Artifacts;
             this.threadFlowLocations = node.ThreadFlowLocations;
 
             if (node.Results != null)
             {
-                int errorsCount = 0;
```
michaelcfanning, line 25: >errorsCount [](http://example.com/codeflow?start=20&length=11)

I wouldn't delete this until you confirm that GH can accept warnings. Also, if this visitor was correct, the validator should also fire on any warnings in a log file (warning that they will be stripped).

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -144,6 +119,8 @@
 
         public override Result VisitResult(Result node)
         {
+            this.ruleIndex = node.RuleIndex;
+
             if (node.Fingerprints != null)
             {
                 // GitHub appears to require that fingerprints be emitted to the
```
michaelcfanning, line 126: >VisitMessage [](http://example.com/codeflow?start=32&length=12)

We have a message string look-up helper that you should use. As Kalle notes, this is a pretty complex operation.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -161,6 +138,32 @@ public override Result VisitResult(Result node)
...
+            {
+                ReportingDescriptor rule = this.ruleIndex != -1 ? this.run.Tool.Driver.Rules[this.ruleIndex] : null;
+
+                if (rule != null &&
+                    rule.MessageStrings != null &&
+                    rule.MessageStrings.TryGetValue(node.Id, out MultiformatMessageString formatString))
+                {
+                    node.Text = node.Arguments?.Count > 0
+                        ? rule.Format(node.Id, node.Arguments)
+                        : formatString?.Text;
```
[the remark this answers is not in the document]
shaopeng-gh: the result is put in `.Text`.
`.Text` is for plain text, consumer should not try to expand any placeholders in it.

should this reset node.Arguments to null ---- currently, .text is added side by side with .id and .arguments.
The information is duplicated, the benefit is we can un-flatten if needed by simply remove the .text field.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
michaelcfanning: OK, nope. It appears that we don't. So the correct course here is to please create a common helper for both this code and the insert optional data visitor to call. Second, create an issue for the problem that Kalle notes: 'Message creation logic does not properly search tool extensions for format strings.'

src/Sarif/Visitors/GitHubIngestionVisitor.cs
shaopeng-gh: This is from the current SDK repo:
image

src/Sarif/Visitors/GitHubIngestionVisitor.cs
shaopeng-gh: got it, will work on this.

src/Sarif.Multitool.Library/RewriteCommand.cs
```diff
@@ -64,6 +65,25 @@ public int Run(RewriteOptions options)
                     reformattedLog = new SortingVisitor().VisitSarifLog(reformattedLog);
                 }
 
+                if (options.NormalizeForGitHub)
+                {
+                    if ((reformattedLog.Runs?.Any(r => r.Artifacts?.Any(a => a.Location?.Uri?.IsAbsoluteUri == true) == true) == true) ||
```
michaelcfanning, line 70: >reformattedLog [](http://example.com/codeflow?start=25&length=14)

So I don't like these two lines of code which will traverse the log file *twice*.

Can't we simply invoke the RebaseUriVisitor in cases when those args are non-null.

And then make the GitHubIngestionVisitor throw new arg exception on encountering the any absolute uri as you detect here.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
shaopeng-gh: done both add helper method and create an issue
[https://github.com/microsoft/sarif-sdk/issues/2582](https://github.com/microsoft/sarif-sdk/issues/2582)

src/ReleaseHistory.md
shaopeng-gh: thanks, fixed

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -161,6 +138,32 @@ public override Result VisitResult(Result node)
             return base.VisitResult(node);
         }
 
+        public override Message VisitMessage(Message node)
+        {
+            if (node.Text == null)
+            {
+                ReportingDescriptor rule = this.ruleIndex != -1 ? this.run.Tool.Driver.Rules[this.ruleIndex] : null;
```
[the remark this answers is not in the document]
shaopeng-gh: you are correct, created an issue to track.

src/Sarif.Multitool.Library/RewriteCommand.cs
shaopeng-gh: made the suggested change.
Though we can not simply throw in GitHubIngestionVisitor, because it is a same method `VisitArtifactLocation` to visit both of it in result and also the originalUriBaseIds,
we allow one and not the other to be absolute.
Currently I does not check and throw, let me know your instruction.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
shaopeng-gh: yes i think it supports warning, please see my other reply with screenshot.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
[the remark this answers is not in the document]
shaopeng-gh: With more reading I think you are right! The `Text` can contain placeholder again!

Within both plain text and formatted message strings, the characters “{” and “}” SHALL be represented by the character sequences “{{” and “}}” respectively. ----- so, I think I will replace { with {{ and } with }}, let me know if sounds good to you.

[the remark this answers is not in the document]
shaopeng-gh: I will replace { with {{ and } with }} -- this is done and updated the test case.

src/Sarif.Multitool.Library/RewriteOptions.cs
```diff
@@ -16,5 +16,33 @@ public class RewriteOptions : SingleFileOptionsBase
             Default = false,
             HelpText = "Sort results in the final output file.")]
         public bool SortResults { get; set; }
+
+        [Option(
+            "normalize-for-github",
```
yongyan-gh, line 21: >"normalize-for-github" [](http://example.com/codeflow?start=12&length=22)

Question: the `Convert` command has the same `normalize-for-github` parameter, can we use `Convert` command to achieve the same?

michaelcfanning: Convert means 'convert from some other non-SARIF format to SARIF.' This operation is a SARIF->SARIF transformation, not a 'conversion'.

One thing, we could use 'normalize-for-ghas' instead as a more precise designation (though 'ghas' may be more opaque to users.

I'm fine with either choice.

shaopeng-gh: use 'normalize-for-ghas' instead as a more precise designation
 ---- sounds good, will change it.

shaopeng-gh: use 'normalize-for-ghas' instead as a more precise designation
---- done.

yongyan-gh (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -28,6 +28,8 @@
 * FEATURE : Allow initialization of file regions cache in `InsertOptionalDataVisitor` (previously initialized exclusively from `FileRegionsCache.Instance`).
 * FEATURE : Provide 'RuleScanTime` trace and emitted timing data. Provide `ScanExecution` trace with no utilization.
 * FEATURE : Populate associated rule data in `LogToolNotification` as called from `SarifLogger`. [#2604](https://github.com/microsoft/sarif-sdk/pull/2604)
+* FEATURE : Add `--normalize-for-ghas` argument to the `rewrite` command to ensure rewritten SARIF is compatible with GitHub Advanced Security (GHAS) ingestion requirements. [#2581](https://github.com/microsoft/sarif-sdk/pull/2581)
+* BREAKING: Rename `--normalize-for-github` argument to `--normalize-for-ghas` for `convert` command. [#2581](https://github.com/microsoft/sarif-sdk/pull/2581)
```
michaelcfanning, line 32: >BREAKING [](http://example.com/codeflow?start=2&length=8)

move this to top

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -161,6 +137,15 @@
             return base.VisitResult(node);
         }
 
+        public override Message VisitMessage(Message node)
+        {
```
michaelcfanning, line 141: >return [](http://example.com/codeflow?start=12&length=6)

add a blank line.

src/Sarif/Visitors/GitHubIngestionVisitor.cs
```diff
@@ -144,6 +118,8 @@
 
         public override Result VisitResult(Result node)
         {
+            this.ruleIndex = node.RuleIndex;
+
             if (node.Fingerprints != null)
             {
                 // GitHub appears to require that fingerprints be emitted to the
```
michaelcfanning, line 125: >VisitMessage [](http://example.com/codeflow?start=32&length=12)

add unit test for result with rule id only, no index

src/Sarif/Visitors/VisitorHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.CodeAnalysis.Sarif.Visitors
+{
+    public static class VisitorHelper
+    {
+        public static void FlattenMessage(Message node, int ruleIndex, Run run)
```
michaelcfanning, line 12: >node [](http://example.com/codeflow?start=50&length=4)

name this message

src/Sarif.Multitool.Library/ConvertOptions.cs
```diff
@@ -24,8 +24,8 @@ public class ConvertOptions : SingleFileOptionsBase
         public string PluginAssemblyPath { get; set; }
 
         [Option(
-            "normalize-for-github",
-            HelpText = "Normalize converted output to conform to GitHub Advanced Security code scanning ingestion requirements.")]
-        public bool NormalizeForGitHub { get; set; }
```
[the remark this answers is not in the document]
michaelcfanning: let's go and remove it, per hulon.

src/Sarif.Multitool.Library/RewriteOptions.cs
```diff
@@ -16,5 +16,33 @@ public class RewriteOptions : SingleFileOptionsBase
             Default = false,
             HelpText = "Sort results in the final output file.")]
         public bool SortResults { get; set; }
+
+        [Option(
+            "normalize-for-ghas",
```
suvamM, line 21: >"normalize-for-ghas" [](http://example.com/codeflow?start=12&length=20)

@shaopeng-gh Sorry if I missed this: this is a required argument?

src/ReleaseHistory.md
```diff
@@ -28,6 +28,8 @@
 * FEATURE : Allow initialization of file regions cache in `InsertOptionalDataVisitor` (previously initialized exclusively from `FileRegionsCache.Instance`).
 * FEATURE : Provide 'RuleScanTime` trace and emitted timing data. Provide `ScanExecution` trace with no utilization.
 * FEATURE : Populate associated rule data in `LogToolNotification` as called from `SarifLogger`. [#2604](https://github.com/microsoft/sarif-sdk/pull/2604)
+* FEATURE : Add `--normalize-for-ghas` argument to the `rewrite` command to ensure rewritten SARIF is compatible with GitHub Advanced Security (GHAS) ingestion requirements. [#2581](https://github.com/microsoft/sarif-sdk/pull/2581)
```
suvamM, line 31: >* FEATURE : Add `--normalize-for-ghas` argument to the `rewrite` command to ensure rewritten SARIF is compatible with GitHub Advanced Security (GHAS) ingestion requirements. [](http://example.com/codeflow?start=0&length=173)

@shaopeng-gh Can you add some more details on what these requirements are? That would make the PR self-contained, otherwise it is a little opaque to me why we are making these changes.

src/Sarif.Multitool.Library/RewriteOptions.cs
michaelcfanning, line 21: >normalize-for-ghas [](http://example.com/codeflow?start=13&length=18)

don't delete the previous option, mark it as obsolete and make sure it continues to work. update the release notes.

src/Sarif.Multitool.Library/RewriteOptions.cs
```diff
@@ -16,5 +16,33 @@ public class RewriteOptions : SingleFileOptionsBase
...
+            HelpText = "Variable to use for the base path token (e.x. 'SRCROOT')"
+            )]
+        public string BasePathToken { get; set; }
+
+        [Option(
+            "rebase-relative-uris",
+            Default = false,
+            HelpText = "All relative uris will be rebased to the base-path-value if true. Default is false."
+            )]
+        public bool RebaseRelativeUris { get; set; }
```
michaelcfanning, line 46: >RebaseRelativeUris [](http://example.com/codeflow?start=20&length=18)

as a next step, you need to make sure rewrite always honors this command in all cases.

src/ReleaseHistory.md
shaopeng-gh: per Yong in the PR meeting, it is from the task 
"Binskim action send log file to GHAS".

src/Sarif.Multitool.Library/ConvertOptions.cs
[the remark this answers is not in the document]
shaopeng-gh: added it back as obsolete, per the last decision from Michael:
image

michaelcfanning (approved):
:shipit:

src/Sarif.Multitool.Library/RewriteOptions.cs
shaopeng-gh: fixed

PR #2590, opened 2022-12-13
Log additional debugging information for analyze command.

shaopeng-gh:
### Notes: if you previously took a look that this PR, we were only adding these additional log when `--verbose`. It is now changed to always log based on the feedback.
------------------------------------------------------------
 
Some user facing `StackOverflowException`.
We are unable to repro the issue, adding the ability to log more information for troubleshooting.

with artificially injecting a stack overflow exception, this is what it looks like in the log:
image

Logging on start and end of processing file,
and log mem usage
image


src/ReleaseHistory.md
```diff
@@ -12,6 +12,7 @@
 * BUGFIX: Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids). [#2513](https://github.com/microsoft/sarif-sdk/pull/2513)
 * BUGFIX: Fix classes inside NotYetAutoGenerated folder missing `virtual` keyword for public methods and properties, by regenerate and manually sync the changes. [#2537](https://github.com/microsoft/sarif-sdk/pull/2537)
 * FEATURE: Allow external set of `MaxFileSizeInKilobytes`, which will allow SDK users to change the value. (Default value is 1024) [#2578](https://github.com/microsoft/sarif-sdk/pull/2578)
+* FEATURE: Log additional debugging information for analyze command when provided `--level` value includes `none`. [#2590](https://github.com/microsoft/sarif-sdk/pull/2590)
```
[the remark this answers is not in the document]
shaopeng-gh: additional logging currently I set to happen when user pass `--verbose` or `--level` includes `note`.

I think you are not looking at the current commit in CodeFlow

src/ReleaseHistory.md
```diff
@@ -12,6 +12,7 @@
 * BUGFIX: Eliminate `IOException` and `DirectoryNotFoundException` exceptions thrown by `merge` command when splitting by rule (due to invalid file characters in rule ids). [#2513](https://github.com/microsoft/sarif-sdk/pull/2513)
 * BUGFIX: Fix classes inside NotYetAutoGenerated folder missing `virtual` keyword for public methods and properties, by regenerate and manually sync the changes. [#2537](https://github.com/microsoft/sarif-sdk/pull/2537)
 * FEATURE: Allow external set of `MaxFileSizeInKilobytes`, which will allow SDK users to change the value. (Default value is 1024) [#2578](https://github.com/microsoft/sarif-sdk/pull/2578)
+* FEATURE: Log additional debugging information for analyze command when provided `--verbose`, or when provided `--level` value including `note`. [#2590](https://github.com/microsoft/sarif-sdk/pull/2590)
```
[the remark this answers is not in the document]
suvamM: @shaopeng-gh Pinging about this

src/Test.UnitTests.Sarif.Driver/TestMultithreadedAnalyzeCommand.cs
```diff
@@ -66,13 +66,13 @@ public int Run(AnalyzeOptionsBase options)
             return result;
         }
 
-        protected override TestAnalysisContext DetermineApplicabilityAndAnalyze(TestAnalysisContext context, IEnumerable<Skimmer<TestAnalysisContext>> skimmers, ISet<string> disabledSkimmers)
+        protected override TestAnalysisContext DetermineApplicabilityAndAnalyze(TestAnalyzeOptions options, TestAnalysisContext context, IEnumerable<Skimmer<TestAnalysisContext>> skimmers, ISet<string> disabledSkimmers)
```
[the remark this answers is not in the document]
suvamM: @shaopeng-gh ping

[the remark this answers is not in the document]
shaopeng-gh: yes, in test I have `TestRuleBehaviors.RaiseStackOverflowException`,
`RaiseStackOverflowException_WithLevelNote` gets the StackOverflowException in the captured output,
`RaiseStackOverflowException_WithoutLevelNote` does not (current logic. if we decided to change to always log as your other comment, I will update this test accordingly)

src/ReleaseHistory.md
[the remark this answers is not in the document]
shaopeng-gh: @michaelcfanning talked to Suvam, he propose we always add these additional logging and no need to check for level, so we don't rely on the ability of the customer to add additional parameters to give us enough telemetry. (e.g. --verbose)
Sounds good to me. Let me know your thoughts.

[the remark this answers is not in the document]
shaopeng-gh: @michaelcfanning Suvam asked me to run this design choice by you, whether always add these additional logs, or only when `--verbose` (or equivalent level note) is passed.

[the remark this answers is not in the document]
shaopeng-gh: Synced with @suvamM, we changed it to always log.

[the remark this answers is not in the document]
michaelcfanning: As a principle, it should not be possible to 'hide` catastrophic errors, they should always display.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -545,7 +545,17 @@ private async Task<bool> ScanTargetsAsync(IEnumerable<Skimmer<TContext>> skimmer
                     {
                         context = _fileContexts[item];
 
-                        DetermineApplicabilityAndAnalyze(context, skimmers, disabledSkimmers);
+                        if (options.Level.Contains(FailureLevel.None))
```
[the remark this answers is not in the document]
shaopeng-gh: changed to always log, closing this comment.

src/Sarif/IAnalysisLogger.cs
```diff
@@ -11,6 +11,10 @@ public interface IAnalysisLogger
 
         void AnalyzingTarget(IAnalysisContext context);
 
+        void CompletedAnalyzingTarget(IAnalysisContext context);
```
michaelcfanning, line 14: >CompletedAnalyzingTarget [](http://example.com/codeflow?start=13&length=24)

In general, modifying this interface is painful and so we must carefully rationalize all additions.

I think I understand the rationalize for 'CompletedAnalyzingTarget' but we should call this 'TargetAnalysisCompleted' to match other overloads.

I don't think LogMemoryUsage' is appropriate, shouldn't this be something that's done by the engine at appropriate junctures? This would simply be a new kind of notification, i.e., a memory utilization notification.

Wouldn't it?

src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -90,7 +91,34 @@ public void AnalyzingTarget(IAnalysisContext context)
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            WriteLineToConsole(string.Format(CultureInfo.CurrentCulture,
+                    SdkResources.MSG_CompletedAnalyzingTarget,
+                        context.TargetUri.LocalPath));
+        }
+
+        public void LogMemoryUsage(IAnalysisContext context)
```
michaelcfanning, line 109: >LogMemoryUsage [](http://example.com/codeflow?start=20&length=14)

A better pattern would be for the context object to indicate whether memory logging is in play and to emit this data when we either start or stop analyzing a target.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -604,18 +604,19 @@
                 rootContext.Policy,
                 target);
 
+            rootContext.Logger.AnalyzingTarget(context);
+            rootContext.Logger.LogMemoryUsage(context);
```
michaelcfanning, line 608: >LogMemoryUsage [](http://example.com/codeflow?start=31&length=14)

What we should do here is consult data somewhere that tells us whether memory logging is enabled. If so, we should generate an informational notification, it's a tool execution notification.

Note that this all leads to a problem: how do users specify verbose output that is specific to a certain logging domain? i.e., they might not want to enable fully verbose logging but want to see memory messages only, or pdb trace messages only, etc. We should establish and follow a clear pattern for this.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -545,7 +545,12 @@ private async Task<bool> ScanTargetsAsync(IEnumerable<Skimmer<TContext>> skimmer
                     {
                         context = _fileContexts[item];
 
-                        DetermineApplicabilityAndAnalyze(context, skimmers, disabledSkimmers);
+                        rootContext.Logger.AnalyzingTarget(context);
```
michaelcfanning, line 548: >AnalyzingTarget [](http://example.com/codeflow?start=43&length=15)

It seems off you have to introduce this, why is that? weren't we calling into analyzing target before?

shaopeng-gh: I have debugged, it is only logged once so it was not called before.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
shaopeng-gh: What we should do here is consult data somewhere that tells us whether memory logging is enabled.
We should establish and follow a clear pattern for this.
----- Yes, I found we have a `Trace`, I added it there.

src/Sarif/Writers/ConsoleLogger.cs
shaopeng-gh: Yes, see the other comment.

src/ReleaseHistory.md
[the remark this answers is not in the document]
shaopeng-gh: Agree

src/Sarif/IAnalysisLogger.cs
shaopeng-gh: we should call this 'TargetAnalysisCompleted' 
 ---- fixed.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -592,7 +592,16 @@ private async Task<bool> ScanTargetsAsync(IEnumerable<Skimmer<TContext>> skimmer
...
+
+                        if (options.Traces.Contains(nameof(DefaultTraces.MemoryUsage), StringComparer.OrdinalIgnoreCase))
+                        {
+                            rootContext.Logger.LogMemoryUsage(context);
+                        }
+
+                        DetermineApplicabilityAndAnalyze(options, context, skimmers, disabledSkimmers);
+
+                        rootContext.Logger.TargetAnalysisCompleted(context);
```
yongyan-gh, line 604: >rootContext.Logger.TargetAnalysisCompleted(context); [](http://example.com/codeflow?start=24&length=52)

when the analysis throws exception do we need the memory usage information? Should we move this call to finally block?

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -462,6 +463,27 @@ public void AnalyzingTarget(IAnalysisContext context)
...
 
+        public void TargetAnalysisCompleted(IAnalysisContext context)
+        {
+        }
+
+        public void LogMemoryUsage(IAnalysisContext context)
+        {
+            var currentProcess = Process.GetCurrentProcess();
+
+            LogConfigurationNotification(
```
yongyan-gh, line 474: >LogConfigurationNotification [](http://example.com/codeflow?start=12&length=28)

The memory usage is more like a tool execution notification to me.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -462,6 +463,27 @@ public void AnalyzingTarget(IAnalysisContext context)
         {
         }
 
+        public void TargetAnalysisCompleted(IAnalysisContext context)
+        {
+        }
+
+        public void LogMemoryUsage(IAnalysisContext context)
+        {
+            var currentProcess = Process.GetCurrentProcess();
```
yongyan-gh, line 472: >var currentProcess = Process.GetCurrentProcess(); [](http://example.com/codeflow?start=12&length=49)

nit: wrap this call in `using` as it allocates a lot resources

shaopeng-gh: fixed

src/Sarif/Writers/SarifLogger.cs
shaopeng-gh: fixed.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
shaopeng-gh: moved to finally block.

suvamM (approved):

michaelcfanning (changes requested):
:clock1:

src/Sarif/IAnalysisLogger.cs
```diff
@@ -11,6 +11,10 @@ public interface IAnalysisLogger
 
         void AnalyzingTarget(IAnalysisContext context);
 
+        void TargetAnalysisCompleted(IAnalysisContext context);
+
+        void LogMemoryUsage(IAnalysisContext context);
```
michaelcfanning, line 16: >LogMemoryUsage [](http://example.com/codeflow?start=13&length=14)

Why is this interface change still here? You should eliminate it and instead generate a tool notification using the logger at the appropriate place.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -462,6 +463,28 @@ public void AnalyzingTarget(IAnalysisContext context)
...
 
+        public void TargetAnalysisCompleted(IAnalysisContext context)
+        {
+        }
+
+        public void LogMemoryUsage(IAnalysisContext context)
+        {
+            using (var currentProcess = Process.GetCurrentProcess())
+            {
+                LogToolNotification(
```
michaelcfanning, line 474: >LogToolNotification [](http://example.com/codeflow?start=16&length=19)

you have implemented duplicate output in two different loggers. this isn't right. instead, you generate a notification upstream in the aggregating logger, which will be passed to the sarif logger and console logger, as every notification is.

src/Sarif/IAnalysisLogger.cs
michaelcfanning: i would prefer not to make an interface change at all. can you prove to me why we must do so?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning: I think you are saying we had a bug before. did you author a unit test that flags the problem and which then passes thanks to your change?

src/Test.Utilities.Sarif/TestRuleBehaviors.cs
```diff
@@ -41,6 +41,7 @@ public enum TestRuleBehaviors
         RegardOptionsAsInvalid = 0x4000,
 
         RaiseExceptionProcessingBaseline = 0x8000,
-        RaiseExceptionPostingLogFile = 0x10000
+        RaiseExceptionPostingLogFile = 0x10000,
+        RaiseStackOverflowException = 0x20000
```
michaelcfanning, line 45: >RaiseStackOverflowException [](http://example.com/codeflow?start=8&length=27)

You have mixed three things in this PR, stack overflow creation, interface/loggings changes, a new trace, and fixed a functional bug in the analysis engine.

Can you go back and just implement the test behaviors change? I will provide a clean example of tracing for now.

Issue #2591, opened 2022-12-13
Support 'IncompatibleRuleIds' in the infrastructure


yongyan-gh:
Please see more details in #2580

>Rules should be able to declare an `IncompatibleRuleIds` set or something similar, All the rule does is to publish this data.
>
>Then we update the analysis engine, which, when instantiating skimmers, makes sure that no new enabled skimmers have a rule id that conflicts with the set of loaded skimmers so far. The analysis engine could raise an error-level configuration notification and exit.
>
>This approach would require users to properly configure analysis to account for the configuration problem. I think everything would drop in pretty cleanly by doing this, we need a low-level update to rules metadata and a simple update to the default skimmer loading behavior.

shaopeng-gh:
Closing the issue with the feature Yong implemented.
https://github.com/microsoft/sarif-sdk/pull/2597

PR #2592, opened 2022-12-20
Adding support for Incompatible rules

yongyan-gh:
# Description

Adding the `IncompatibleRuleIds` and `IncompatibleRuleHandling` properties to `Skimmer` according to #2591


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -2069,6 +2069,134 @@ public IFileSystem FileSystem
...
 
+        [Fact]
+        public void CheckIncompatibleRules_DisableIncompatibleRuleAndContinueAnalysis()
+        {
+            IncompatibleTestRule[] skimmers = new[]
+            {
+                new IncompatibleTestRule("TEST1001", null),
+                new IncompatibleTestRule("TEST1002", null),
+                new IncompatibleTestRule("TEST1003", new HashSet<string> { "TEST1001" },  IncompatibleRuleHandling.DisableAndContinueAnalysis),
+            };
```
shaopeng-gh, line 2080: suggestion:

if rule 1 is incompatible with rule 2,
the incompatible flag can be set at rule 1, or rule 2.
And they should be equivalent, is it correct.

What if they are both set, may add a test for that.
And then a test for rule 1 -> rule 2 -> rule 3 -> rule 1, circular case.

src/Sarif/SdkResources.resx
```diff
@@ -284,4 +284,10 @@ You can also refer to properties in the result's property bag with 'properties.&
   <data name="WRN997_OneOrMoreFilesSkippedDueToSize" xml:space="preserve">
     <value>One or more files were skipped for analysis due to exceeding size limits (currently configured as {0} KB). The 'max-file-size-in-kb' command-line argument can be used to increase this threshold.</value>
   </data>
+  <data name="ERR998_IncompatibleRuleDetected" xml:space="preserve">
+    <value>Detected incompatible rules. Rule '{0}' is incompatible with rule '{1}'.</value>
+  </data>
+  <data name="Wrn998_IncompatibleRuleDetected" xml:space="preserve">
```
shaopeng-gh, line 290: >Wrn998 [](http://example.com/codeflow?start=14&length=6)

just question, 
how do we choose 998 or 997, I see a few 997 above.

src/Sarif/Warnings.cs
```diff
@@ -189,5 +190,32 @@ public static void LogObsoleteOption(IAnalysisContext context, string obsoleteOp
...
+
+            context.Logger.LogConfigurationNotification(
+                new Notification
+                {
+                    Descriptor = new ReportingDescriptorReference
+                    {
+                        Id = Wrn998_IncompatibleRuleDetected
+                    },
+                    Message = new Message { Text = message },
+                    Level = FailureLevel.Warning,
```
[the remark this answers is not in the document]
shaopeng-gh: I found it:  for `DisableAndContinueAnalysis`.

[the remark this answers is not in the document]
shaopeng-gh: Is it something user can choose what option he wants, if so what command line parameter to pass for it:
`switch (originalSkimmer.IncompatibleRuleHandling)`

src/Sarif/SdkResources.resx
LingZhou-gh: "Wrn998_IncompatibleRuleDetected" should capitalize the first 3 letters to be more consistent with other name.

src/Sarif.Driver/Sdk/IncompatibleRuleHandling.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Driver
+{
+    public enum IncompatibleRuleHandling
+    {
+        Ignore = 0,
+        DisableAndContinueAnalysis,
+        ExitAnalysis,
+    }
```
LingZhou-gh, line 11: The comma in "ExitAnalysis," could be removed.

src/Sarif/Warnings.cs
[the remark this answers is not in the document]
yongyan-gh: My understanding is that people who creates the new rules should be responsible to set the `IncompatibleRuleIds` and `IncompatibleRuleHandling` or just leave them as default.

Its not possible to set the behavior for each rule through command line right?

src/Sarif/SdkResources.resx
yongyan-gh: It is in Warnings.cs line 21, the WRN998 is the category for "// (Non-catastrophic) conditions that result in rules disabling themselves."

src/Sarif/RuntimeConditions.cs
```diff
@@ -52,6 +52,7 @@ public enum RuntimeConditions : uint
         ExceptionPostingLogFile = 0x40000,
 
         // Non-fatal conditions
+        RuleIsIncompatibleWithAnotherRule = 0x00800000,
```
michaelcfanning, line 55: Change this to 'OneOrMoreRulesAreIncompatible'

src/Sarif/Errors.cs
```diff
@@ -26,6 +26,7 @@ public static class Errors
         private const string ERR997_ExceptionLoadingAnalysisTarget = "ERR997.ExceptionLoadingAnalysisTarget";
         private const string ERR997_ExceptionInstantiatingSkimmers = "ERR997.ExceptionInstantiatingSkimmers";
         private const string ERR997_OutputFileAlreadyExists = "ERR997.OutputFileAlreadyExists";
+        internal const string ERR998_IncompatibleRuleDetected = "ERR997.IncompatibleRuleDetected";
```
michaelcfanning, line 29: >ERR998_IncompatibleRuleDetected [](http://example.com/codeflow?start=30&length=31)

ERR998_IncompatibleRulesDetected

Note the plural form. It takes two rules to be incompatible. :)

src/Sarif/RuntimeConditions.cs
```diff
@@ -52,6 +52,7 @@ public enum RuntimeConditions : uint
         ExceptionPostingLogFile = 0x40000,
 
         // Non-fatal conditions
```
michaelcfanning, line 54: >Non [](http://example.com/codeflow?start=11&length=3)

Why isn't this a fatal condition? How do we proceed if the configured rules are not compatible?

src/Sarif/Errors.cs
```diff
@@ -487,6 +488,28 @@ public static void LogUnhandledExceptionInitializingRule(IAnalysisContext contex
             context.RuntimeErrors |= RuntimeConditions.ExceptionInSkimmerAnalyze;
         }
 
+        public static void LogIncompatibleRule(IAnalysisContext context, string ruleId, string incompatibleRuleId)
```
michaelcfanning, line 491: >LogIncompatibleRule [](http://example.com/codeflow?start=27&length=19)

LogIncompatibleRules

Note plural form.

src/Sarif.Driver/Sdk/ExitReason.cs
```diff
@@ -11,6 +11,7 @@ public enum ExitReason
         UnhandledExceptionInstantiatingSkimmers,
         UnhandledExceptionInEngine,
         NoRulesLoaded,
+        IncompatibleRulesDetected,
```
michaelcfanning, line 14: >IncompatibleRulesDetected [](http://example.com/codeflow?start=8&length=25)

This name is good!

src/Sarif.Driver/Sdk/IncompatibleRuleHandling.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Driver
+{
+    public enum IncompatibleRuleHandling
```
michaelcfanning, line 6: >u [](http://example.com/codeflow?start=29&length=1)

You have implemented these choices cleanly. But can you explain why you think we need them? :)

If someone specifies 'ignore', then why would they bother to publish their incompatible rule ids?

Is it clear to us that a skimmer would have a single setting that would apply to all incompatible rules? 

Why would a rule choose to disable and continue analysis?

What happens if two incompatible rules have different settings? If either rule chooses 'exit analysis' then we will exit. If both rules are disabled, then analysis isn't complete.

This additional machinery would be interesting if it could be specified via configuration. But if a user can specify configuration, they could simply choose to selectively enable/disable the rule they want to run.

I am interested in your thoughts, of course, but at this point I wonder whether we shouldn't simply take the clean 'exit analysis' behavior, which would be a smaller change. We could improve on it later. What's your opinion?

yongyan-gh: The reason I am thinking about adding the `DisableAndContinueAnalysis` is, adding a new rule with existing rule as incompatible rule should not break existing analysis.

For example: The SARIF web validator scenario. If we add the new `GH1007` ProvideFullyFormattedMessageStrings, its incompatible with 'SARIF2002' ProvideMessageArguments. The validator allows users to run all SARIF* rules with GH* rule together. After the new GH1007 rule added, users run the SARIF and GH rules together, they will get a tool configuration exception indicates there is a incompatible rule, they will not see any results. The validator needs to define 2 set of configuration files to enable SARIF2002 and disable GH1007 or the opposite, and switch configuration file between these 2. So I think an option to just disable the incompatible rule and continue analysis may be useful in this case. 
What do you think

michaelcfanning: I think the exception is appropriate. When you add a new GH rule, you will be depending on existing configuration. The existing rule can't reasonably be expected to defensively turn itself off if a future rule marks it as incompatible (and what if you want the new rule to run by default, as we do with the GH analysis??).

Really, we can't handle the full matrix of configuration cases in a reasonable way, I think. Let's remove this extra complexity and just fail-fast in the event of incompatibility. For sure, we need to mark new rules with incompatibilities as a breaking change. 

Please push back if you disagree and/or see deeper into this topic than I. :)

yongyan-gh: ok i agree, there are more complicated cases of handling the conflicting rules, we can start with the throwing exception now.

yongyan-gh:
New PR created #2597 so close this one

PR #2595, opened 2022-12-22
Update CommandLineParser from 2.7.82 to 2.9.1 in sarif-sdk

LingZhou-gh:
Update CommandLineParser from 2.7.82 to 2.9.1 for sarif-sdk

### DIFF
--- a/src/Sarif.Driver/Sarif.Driver.csproj
+++ b/src/Sarif.Driver/Sarif.Driver.csproj
@@ -37,7 +37,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="CommandLineParser" Version="2.7.82" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
     <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="3.1.9" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="System.Composition" Version="5.0.0" />
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -47,6 +47,7 @@
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
   </ItemGroup>
 
   <ItemGroup Label="Packaging">
--- a/src/Sarif/Sarif.csproj
+++ b/src/Sarif/Sarif.csproj
@@ -46,8 +46,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-    <PackageReference Include="CommandLineParser" Version="2.9.1" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />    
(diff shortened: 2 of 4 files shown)


src/Sarif/Sarif.csproj
```diff
@@ -47,6 +47,7 @@
   <ItemGroup>
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
```
yongyan-gh, line 50: >CommandLineParser [](http://example.com/codeflow?start=31&length=17)

why its required here?

yongyan-gh:
I recall one principle we summarized from the NewtonSoft.Json upgrade: as a low level sdk, it should keep reference version as lowest as possible.
Do we have to upgrade the version of command line parser?

LingZhou-gh:
The error which is met while *just* updating BinSKim CommandLineParser package is as follows:
image

It might due to the referenced projects sarif-sdk still using the old version dll, based on some article here:
https://stackoverflow.com/questions/8058832/system-missingmethodexception-method-not-found

src/Sarif/Sarif.csproj
LingZhou-gh: Actually not sure, but it is the only project reference by Sarif.Driver.csproj. 
image

yongyan-gh: Sarif.Driver depends on CommandLineParser but Sarif project doesn't have a dependency on it right?

This is new dependency added to Sarif. It is not required in my mind.

yongyan-gh:
According to the trace  log, the exception was thrown from BinSkim. You can just update BinSkim project to use the latest CommandLine Parser.

---
In reply to: [1363532416](https://github.com/microsoft/sarif-sdk/pull/2595#issuecomment-1363532416) [](http://example.com/codeflow?ancestors=1363532416,1363532416)

LingZhou-gh:
https://github.com/microsoft/binskim/pull/644/commits
PR 644 just got updated this afternoon (with only BinSkim project using the latest CommandLineParser). A new test was triggered and shows the fail message.

LingZhou-gh:
I just **repro** the same exception on my local machine **with *only* BinSkim project using the latest CommandLineParser package** (2.9.1) while the others still using the old one.
image

With all the other related projects updated to use the latest CommandLineParser package, the issue is gone.

src/Sarif/Sarif.csproj
LingZhou-gh: Verified that the change in Sarif.csproj is not needed. Remove it.

shaopeng-gh:
Cool, if the problem that BinSkim was not able to auto upgrade this ref is just due to submodule sdk need to upgrade as well,
then I think this would be a simple change, putting my signoff.

---
In reply to: [1363621525](https://github.com/microsoft/sarif-sdk/pull/2595#issuecomment-1363621525) [](http://example.com/codeflow?ancestors=1363621525,1363621525)

shaopeng-gh (approved):
:shipit:

michaelcfanning (approved):
:shipit:

src/Sarif.Driver/Sarif.Driver.csproj
```diff
@@ -37,7 +37,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="CommandLineParser" Version="2.7.82" />
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
```
michaelcfanning, line 40: >Include [](http://example.com/codeflow?start=22&length=7)

OK, hm, I thought there was an API change to make. I guess not (or maybe it shows up in another client tool).

michaelcfanning: Now I understand from the discussion, everything works, but you have to upgrade everyone at once. :)

PR #2596, opened 2022-12-27
Fix several user reported issue `NullReferenceException`

shaopeng-gh:
Fix `NullReferenceException` when file hashing fails (due to file locked or other errors reading the file).

One way to repro:
Scan the root folder and put the output SARIF file inside the folder, the output .sarif file will be locked,
for any reason can’t get hash (locked in this case), hashes will be null and then we will get unhandled null exception.


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -513,11 +513,11 @@ private async Task<bool> HashFilesAndPutInAnalysisQueueAsnc()
 
                         loggerCache ??= new Dictionary<string, CachingLogger>();
 
-                        if (!loggerCache.TryGetValue(hashData.Sha256, out CachingLogger logger))
-                        {
-                            logger = loggerCache[hashData.Sha256] = (CachingLogger)context.Logger;
-                        }
-                        context.Logger = logger;
+                        context.Logger = hashData?.Sha256 == null
+                            ? (CachingLogger)context.Logger
```
yongyan-gh, line 517: >? (CachingLogger)context.Logger [](http://example.com/codeflow?start=28&length=31)

if no hash, have to set the context.Logger to itself? can we skip in this case?

michaelcfanning: Yes, let's just wrap this code in a check that hashData.Sha256 is non-null, will be more readable.

Could we get this change in and then also try to mock up a test for this condition? 

I'm concerned we're not necessarily handling this negative condition comprehensively.

shaopeng-gh: Could we get this change in and then also try to mock up a test for this condition? ---- done fixing the code, will need to mock up a test.

michaelcfanning (approved):
:shipit:

PR #2597, opened 2022-12-28
Support Incompatible rules

yongyan-gh:
# Description

Adding `IncompatibleRuleIds` properties to `Skimmer` class.

The rule creator can set the id of the rules, which conflict with the new added rule, as the value of `IncompatibleRuleIds` property.

The `AnalyzeCommandBase` virtual function `CheckIncompatibleRules` checks if any rule is incompatible with other rules, if yes it logs a tool configuration notification with the conflict rule ids and exit the analysis.

Related to #2591


src/Sarif.Driver/Sdk/ExitReason.cs
```diff
@@ -11,6 +11,7 @@ public enum ExitReason
         UnhandledExceptionInstantiatingSkimmers,
         UnhandledExceptionInEngine,
         NoRulesLoaded,
+        IncompatibleRulesDetected,
```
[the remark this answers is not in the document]
michaelcfanning: Good catch! We need this value to be appended to the end of the enum.

Alternately, if you wanted to preserve some specific order, you could explicitly define the value of each enum. 

If you simply add the value to the end, please add this comment at the top of the class.

// Be sure to append new values for this enum to avoid changing existing member values.

src/Sarif/Errors.cs
```diff
@@ -509,6 +510,28 @@ public static RuntimeConditions LogUnhandledEngineException(IAnalysisContext con
             return RuntimeConditions.ExceptionInEngine;
         }
 
+        public static void LogIncompatibleRule(IAnalysisContext context, string ruleId, string incompatibleRuleId)
```
michaelcfanning, line 513: >LogIncompatibleRule [](http://example.com/codeflow?start=27&length=19)

This should be plural in form, 'LogIncompatibleRules'.

Also, you haven't added the string value for this message as a comment in this code. Please observe other LogXXX members to see the pattern.

src/Sarif/Errors.cs
```diff
@@ -509,6 +510,28 @@ public static RuntimeConditions LogUnhandledEngineException(IAnalysisContext con
...
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.Logger.LogConfigurationNotification(
```
michaelcfanning, line 520: >LogConfigurationNotification [](http://example.com/codeflow?start=27&length=28)

Add the updated string as a comment here. Please look at the code in this file to observe how carefully these comments are formatted (they are 'columnized').

The current configuration enables rules that are not compatible ('{0}' has declared that it is not compatible with '{1}'). You can selectively disable one of the rules using an updated XML configuration (passed by the --config argument).

src/Sarif/Errors.cs
michaelcfanning, line 513: >LogIncompatibleRule [](http://example.com/codeflow?start=27&length=19)

In general, when you update a user-facing message, please provide an example of the message from a real world test, so that we can assess its quality and completeness.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -792,6 +795,38 @@
             };
         }
 
+        protected virtual void CheckIncompatibleRules(IEnumerable<Skimmer<TContext>> skimmers, TContext context, ISet<string> disabledSkimmers)
```
michaelcfanning, line 798: >protected [](http://example.com/codeflow?start=8&length=9)

Can you shut down accessibility of this method for now? i.e., mark it private

This class is pretty carefully constructed to follow a sequence of specific overridable methods to function.

I understand why you made it protected virtual, because imposing exist is pretty serious. :) But let's leave it until we can do a more thorough review of the way this API is evolving (i.e., we should be cautious about introducing new exposed methods).

Also, can you tell me, does your new functionality work for the multithreaded command as well?

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -792,6 +795,38 @@
...
+                }
+
+                foreach (string incompatibleRuleId in entry.Value.IncompatibleRuleIds)
+                {
+                    if (availableRules.ContainsKey(incompatibleRuleId))
+                    {
+                        Errors.LogIncompatibleRule(context, entry.Key, incompatibleRuleId);
+                        ThrowExitApplicationException(context, ExitReason.IncompatibleRulesDetected);
+                    }
+                }
```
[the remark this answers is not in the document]
michaelcfanning: Mostly, it's ok to simply ignore these CodeQL results, as we only see ones that are added incrementally.

This rule happens to be one that's ok to fix, Yong, if you want (and if you contrive a pattern that's equally or more readable to this one). :)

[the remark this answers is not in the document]
michaelcfanning: Well, I tend to agree. First off, if you can't use the construct to make the code more maintainable/readable, why would you make a change with greater runtime costs? I tend not to worry as much about this simpler suggestion to implement .Where in preference of for + conditional, I definitely get very agitated when I see devs chain enumerations together in obviously non-performant ways.

Yong, let's do leave this code alone. And if we can disable this check in CodeQL workflow, I agree with Kalle, good change.

src/Sarif.Driver/Sdk/ExitReason.cs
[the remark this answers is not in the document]
yongyan-gh: move the new enum to the end. comment added.

src/Sarif/Errors.cs
yongyan-gh: fixed

src/Sarif/SdkResources.resx
```diff
@@ -284,4 +284,7 @@ You can also refer to properties in the result's property bag with 'properties.&
   <data name="WRN997_OneOrMoreFilesSkippedDueToSize" xml:space="preserve">
     <value>One or more files were skipped for analysis due to exceeding size limits (currently configured as {0} KB). The 'max-file-size-in-kb' command-line argument can be used to increase this threshold.</value>
   </data>
+  <data name="ERR997_IncompatibleRulesDetected" xml:space="preserve">
+    <value>Detected incompatible rules. Rule '{0}' is incompatible with rule '{1}'.</value>
```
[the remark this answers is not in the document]
yongyan-gh: Thanks for the suggestion

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
yongyan-gh: Understand, changed to internal so it can be accessed by the unit tests.

The same code and tests is added for `MultithreadedAnalyzeCommandBase`, could not find a common place for it, thinking about extract the common code of these 2 classes into a common base class.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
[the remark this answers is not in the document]
yongyan-gh: sure will look into that. Created #2598

src/Sarif/Errors.cs
michaelcfanning: Nice!

michaelcfanning (approved):
:shipit:

PR #2599, opened 2022-12-28
Enqueue files filter

michaelcfanning:
This is a significant change that touches on some core driver concepts, worth a careful review. 

* Updates max file size in kb type from int to long.
* Marks `AnalyzeCommandBase` as obsolete, we should delete it soon.
* We now emit a warning anytime a file is skipped due to exceeding size threshold.
* Updated the file size option to allow setting it to any value 0 or great. The rationale here is that setting the threshold to 0 would allow a tool to collect a log file that documents what files *would* have been scanned but were not.
* Found and fixed at least one bug accounting for the hangs observed in Linux analysis (due to exceptions enumerating files).
* Provide an overridable mechanism for updating file filtering logic (being file size). This will allow tools like BinSkim to examine and discard files before the hashing phase (because they aren't Windows PEs or other scan targets). This will resolve the horrible perf slowdowns observed in pipelines that naively apply wildcard specifiers at analysis time.
* I've converted the max file size in kb argument to an actual context property. This means it can be configured using exported XML. We should convert *all* command-line arguments in this way. i.e., you should be able to pass an XML file only to drive analysis.

@shaopeng-gh @LingZhou-gh @EasyRhin0 @suvamM @yongyan-gh 

@rtaket @cfaucon, heads up: this SDK will emit skipped file warnings, as requested.


src/Sarif/Errors.cs
```diff
@@ -46,8 +46,10 @@ public static void LogExceptionLoadingTarget(IAnalysisContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
+            context.RuntimeErrors |= RuntimeConditions.ExceptionLoadingTargetFile;
```
michaelcfanning, line 49: >context [](http://example.com/codeflow?start=12&length=7)

We now set RuntimeErrors early in every method, so that this data is persisted in the event that an unhandled exception occurs in the remainder of the method (as happened during my current dev/test loop).

src/Sarif/Warnings.cs
```diff
@@ -22,6 +23,30 @@ public static class Warnings
...
 
+        public static void LogFileSkippedDueToSize(IAnalysisContext context, string skippedFile, long fileSizeInKb)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // '{0}' was skipped as its size ({1} kilobytes) exceeds the currently configured threshold ({2} kilobytes).
+            context.Logger.LogConfigurationNotification(
```
michaelcfanning, line 34: >context [](http://example.com/codeflow?start=12&length=7)

Produces a console warning at runtime like so. Note that we include the full path name of the skipped file in the output. I produced this message by configuring the max file in kb argument via a config file, demonstrating that this is functional using actual client tool.

BINSKIM : warning WRN997.FileSkippedDueToSize : 'D:\repros\NetperfServiceReleaseFolder\VMAgents\Linux\NetPerfCanaryAppOrchestrator_artifact\NetPerfOrchestrator\resources\scripts\SecondaryNic.sh' was not analyzed as its size (4 kilobytes) exceeds the currently configured threshold (1 kilobytes).

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -12,6 +12,7 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.Driver
 {
+    [Obsolete("AnalyzeCommandBase will be deprecated entirely soon. Use MultithreadedAnalyzeCommandBase instead.")]
```
michaelcfanning, line 15: >AnalyzeCommandBase [](http://example.com/codeflow?start=15&length=18)

It grows wearisome to maintain this obsolete single-threaded analysis, let's zap it next release.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -94,15 +95,16 @@
~                        ExecutionException = ex;
~                        return FAILURE;
~                    }
                     catch (Exception ex)
```
michaelcfanning, line 98: >LogUnhandledEngineException [](http://example.com/codeflow?start=48&length=27)

This helper previously returned an error condition to OR into RuntimeErrors. I've updated the helper to be consistent with other utility methods. We now retrieve the RuntimeErrors in the finally block.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -214,7 +213,7 @@
 
             // 1: First we initiate an asynchronous operation to locate disk files for
             // analysis, as specified in analysis configuration (file names, wildcards).
-            Task<bool> enumerateFilesOnDisk = EnumerateFilesOnDiskAsync(options, rootContext);
+            Task<bool> enumerateFilesOnDisk = EnumerateFilesOnDiskAsync(options);
```
michaelcfanning, line 216: >EnumerateFilesOnDiskAsync [](http://example.com/codeflow?start=46&length=25)

This method already access instance data, so we'll just utilize _rootContext.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -363,107 +367,129 @@
             }
```
michaelcfanning, line 367: >ShouldEnqueue [](http://example.com/codeflow?start=31&length=13)

This diff is all screwed up here (in codeflow). This is the core new feature, we provide an overridable mechanism for filtering files at file enumeration time. This will allow binskim to check to see whether a file can be scanned. Currently, we let all files in and then they all get hashed (and then discarded), resulting in egregious slowdowns.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -363,107 +367,129 @@
...
+
+                            await readyToHashChannel.Writer.WriteAsync(_fileContextsCount++);
+                        }
                     }
                 }
             }
-
-            readyToHashChannel.Writer.Complete();
+            catch (Exception e)
+            {
```
michaelcfanning, line 475: OK, so here's the source of some of the hangs that have been reported, if file enumeration raised an unhandled exception, we never closed the write channel. Oops.

src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
```diff
@@ -115,7 +115,6 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
                 Quiet = true,
                 PrettyPrint = true,
                 Optimize = true,
-                Kind = new List<ResultKind> { ResultKind.Fail },
```
michaelcfanning, line 118: >Kind [](http://example.com/codeflow?start=16&length=4)

This is now a default in the options ctor.

src/Test.Utilities.Sarif/TestAnalyzeOptions.cs
```diff
@@ -9,12 +9,6 @@ namespace Microsoft.CodeAnalysis.Sarif
 {
     public class TestAnalyzeOptions : AnalyzeOptionsBase
     {
-        public TestAnalyzeOptions()
```
michaelcfanning, line 12: >TestAnalyzeOptions [](http://example.com/codeflow?start=15&length=18)

set now in the base ctor.

suvamM (approved):

src/Sarif/Errors.cs
yongyan-gh: Curious what exception you got, guess its thrown by logger? Should we apply the same change to `Warnings`?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -89,13 +89,14 @@
                         if (!(ex is ExitApplicationException<ExitReason>))
                         {
                             // These exceptions escaped our net and must be logged here
-                            RuntimeErrors |= Errors.LogUnhandledEngineException(_rootContext, ex);
+                            Errors.LogUnhandledEngineException(_rootContext, ex);
```
yongyan-gh, line 92: >Errors.LogUnhandledEngineException [](http://example.com/codeflow?start=28&length=34)

so this call may throw exception, should we add a 'catch' block in the first 'try' block to catch it?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -363,107 +366,133 @@ private void LogCachingLogger(TContext rootContext, TContext context, bool clone
...
+                            if (ShouldEnqueue(file, _rootContext))
+                            {
+                                sortedFiles.Add(file);
+                                continue;
+                            }
+
+                            if (!IsTargetWithinFileSizeLimit(file, _rootContext.MaxFileSizeInKilobytes, out long fileSizeInKb))
+                            { 
+                                _ignoredFilesCount++;
+                            }
```
shaopeng-gh, line 458: question, 
we use the ShouldEnqueue() to see if to add to sortedFiles list.
we use the IsTargetWithinFileSizeLimit to see if to add to _ignoredFilesCount.

Should we use the same logic, by just using the `else`

We can add more logics to `ShouldEnqueue` in the future.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -118,7 +129,7 @@
         [Option(
             "max-file-size-in-kb",
             HelpText = "The maximum file size (in kilobytes) that will be analyzed.",
-            Default = 1024)]
-        public int MaxFileSizeInKilobytes { get; set; } = 1024;
+            Default = -1)]
```
shaopeng-gh, line 132: >-1 [](http://example.com/codeflow?start=22&length=2)

the benefit of have 1024 in the CommonOption here is it will be shown to user. with changed to -1, it is actually 1024 as we coded inside c# but just is not as clear to user.

shaopeng-gh (approved):
:shipit:

PR #2600, opened 2022-12-29
BUGFIX  : Resolve hangs due to unhandled exceptions during multithreaded analysis file hashing phase.

shaopeng-gh:
When exception happens, below 2 lines will not run:

image

I can reproduce this when debugging below line will be waiting, not ending at all:

image


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -882,6 +882,45 @@ public void MultithreadedAnalyzeCommandBase_TargetFileSizeTestCases()
             }
         }
 
+        [Fact]
+        public void MultithreadedAnalyzeCommandBase_ErrorWhenHashing()
```
shaopeng-gh, line 886: >MultithreadedAnalyzeCommandBase_ErrorWhenHashing [](http://example.com/codeflow?start=20&length=48)

this is to add a test for previous pr about `NullReferenceException`.  
without that fix, this test will fail with  `NullReferenceException`

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -520,41 +520,51 @@ private async Task<bool> HashFilesAndPutInAnalysisQueueAsnc()
...
-            readyToScanChannel.Writer.Complete();
+            catch (Exception e)
+            {
+                Errors.LogUnhandledEngineException(_rootContext, e);
+                ThrowExitApplicationException(_rootContext, ExitReason.UnhandledExceptionInEngine);
+            }
+            finally
+            {
+                readyToScanChannel.Writer.Complete();
+            }
```
shaopeng-gh, line 567: used the same code for enumeration phase.

michaelcfanning (approved):
Yes meant to note that we need to review all channel writes and reads to ensure that completion occurs in any code path with unhandled exception.

PR #2601, opened 2022-12-30
FEATURE : Provide overridable `ShouldComputeHashes` predicate method to prevent files from hashing.

shaopeng-gh:
This change only provides overridable for the derived class.


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -378,6 +378,11 @@ protected virtual bool ShouldEnqueue(string file, TContext context)
             return shouldEnqueue;
         }
 
+        protected virtual bool ShouldComputeHashes(string file, TContext context)
```
shaopeng-gh, line 381: >ShouldComputeHashes [](http://example.com/codeflow?start=31&length=19)

This change enables override of `ShouldComputeHashes` instead of `ShouldEnqueue`

speed test of my local files folder with 7k files, roughly:
1. previous PR using `ShouldEnqueue`: 00:03:47
2. similar to before any change, directly return true: 00:02:48
3. this PR and override with the check we desire(if is binary): 00:02:13

after this merged next will send PR for BinSkim using it.

michaelcfanning (approved):

PR #2604, opened 2023-01-11
Trace notifications

michaelcfanning:


src/Sarif/IAnalysisLogger.cs
```diff
@@ -24,7 +24,10 @@ public interface IAnalysisLogger
         /// <param name="notification">
         /// The notification to log.
         /// </param>
-        void LogToolNotification(Notification notification);
+        /// <param name="associatedRule">
+        /// The scan rule, if any, associated with the notification.
+        /// </param>
+        void LogToolNotification(Notification notification, ReportingDescriptor associatedRule = null);
```
michaelcfanning, line 30: >associatedRule = null [](http://example.com/codeflow?start=80&length=21)

An optional argument in an interface contract is an interesting thing. Any caller to the interface does not need to provide this argument. Anyone implementing this interface can override the optional value.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -305,6 +305,11 @@ private ISet<string> ValidateTargetsExist(TContext context, ISet<string> targets
                 Policy = policy ?? new PropertiesDictionary()
             };
 
+            context.Traces =
```
michaelcfanning, line 308: >context [](http://example.com/codeflow?start=12&length=7)

This handling can now be removed from upstream driver tools that are currently required to do the same thing.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -479,6 +479,16 @@ public void LogToolNotification(Notification notification)
             _run.Invocations[0].ExecutionSuccessful &= notification.Level != FailureLevel.Error;
 
             CaptureFilesInNotification(notification);
+
+            if (associatedRule != null)
```
michaelcfanning, line 483: >associatedRule [](http://example.com/codeflow?start=16&length=14)

With this change, the rule metadata for any notification associated with a rule will be persisted to the rules table in the log.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1041,7 +1056,31 @@ public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TCo
 
                 try
                 {
+                    Stopwatch stopwatch = context.Traces.HasFlag(DefaultTraces.RuleScanTime)
```
michaelcfanning, line 1059: >stopwatch [](http://example.com/codeflow?start=30&length=9)

Shaopeng, here's a clean pattern for producing timing data on a per-rule + target file basis. We can also consider adding memory consumption information in a similar change, if helpful.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -663,11 +663,100 @@ public void AnalyzeCommandBase_ReportsWarningOnUnsupportedPlatformForRuleAndNoRu
             return run;
         }
 
+        [Fact]
+        public void AnalyzeCommand_Traces()
```
michaelcfanning, line 667: >AnalyzeCommand_Traces [](http://example.com/codeflow?start=20&length=21)

Good things about this test: there are minimal literal constant 'magic values', it runs a comprehensive set of tests and will report on all break (instead of failing on the first one), it nicely buckets all trace testing in one place and can easily be extended. This change also makes some nice incremental improvements to the general test methodology, for example, returning the Run object from the test helper to assist in validation.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -663,11 +663,100 @@ public void AnalyzeCommandBase_ReportsWarningOnUnsupportedPlatformForRuleAndNoRu
...
+                                                         expectedResultCode: SUCCESS,
+                                                         expectedResultCount: WARNING_COUNT + NOTE_COUNT,
+                                                         options);
+
+
+                int resultCount = 0;
+                int executionNotificationsCount = 0;
+                int configurationNotificationCount = 0;
+
+                SarifHelpers.ValidateRun(
```
michaelcfanning, line 692: >ValidateRun [](http://example.com/codeflow?start=29&length=11)

Note that this validation will halt the test early if it fails. That's ok. If this fails, then the follow-on validation is unlikely to succeed, because we already know that we don't have the expected results + notifications data generated.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -663,11 +663,100 @@ public void AnalyzeCommandBase_ReportsWarningOnUnsupportedPlatformForRuleAndNoRu
...
+                        // We expected timing data for every rule.
+                        if (executionNotificationsCount != expectedNotificationsCount)
+                        {
+                            sb.AppendLine($"\t{trace} : expected {expectedNotificationsCount} notifications but saw {executionNotificationsCount}.");
+                            continue;
+                        }
+
+                        if (executionNotifications?.Where(t => t.Message.Text.Contains("elapsed")).Count() != expectedNotificationsCount)
+                        {
+                            sb.AppendLine($"\t{trace} : did not observe term 'elapsed' in rule timing notifications.");
```
michaelcfanning, line 741: >\t [](http://example.com/codeflow?start=44&length=2)

Notice how this code indents individual reporting lines to maximize readability in the VS test window.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -663,11 +663,100 @@ public void AnalyzeCommandBase_ReportsWarningOnUnsupportedPlatformForRuleAndNoRu
...
+                        if (executionNotifications?.GroupBy(t => t.AssociatedRule.Id).Count() != rulesCount)
+                        {
+                            sb.AppendLine($"\t{trace} : did not observe timing notifications for every rule.");
+                        }
+
+                        break;
+                    }
+                }
+            }
+            sb.Length.Should().Be(0, $"test cases failed : {Environment.NewLine}{sb}");
```
michaelcfanning, line 753: > $"test cases failed : {Environment.NewLine}{sb}" [](http://example.com/codeflow?start=36&length=49)

notice this construct which maximizes readability in the VS test window. this reporting subtlety should be captured in some test helpers sometime soon.

michaelcfanning:
@shaopeng, can you look at this for me? :) I'm still working on the release notes.

shaopeng-gh:
@michaelcfanning 
ah I see wrong shaopeng was added here should be shaopeng-gh so I missed this pr.
I will read the pattern and work on the memory consumption information in a similar change.

---
In reply to: [1381184325](https://github.com/microsoft/sarif-sdk/pull/2604#issuecomment-1381184325) [](http://example.com/codeflow?ancestors=1381184325,1381184325)

michaelcfanning:
Very sorry! Please review this change and let me know if you have feedback. :) we've got other worked blocked by this, so I merged already. I will remember to use the proper account reference : @shaopeng-gh . :)

PR #2607, opened 2023-01-16
* FEATURE : Provide `MemoryUsage` trace and emitted memory usage data.

shaopeng-gh:
Based on pattern of Michael's newest check-in, now it does not need interface changing and is straightforward.

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -1060,6 +1060,28 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                         ? Stopwatch.StartNew()
                         : null;
 
+                    if (context.Traces.HasFlag(DefaultTraces.MemoryUsage))
+                    {
+                        using (var currentProcess = Process.GetCurrentProcess())
+                        {
+                            string file = context.TargetUri.LocalPath;
+                            string directory = Path.GetDirectoryName(file);
+                            file = Path.GetFileName(file);
+                            string memoryUsage = $"'{file}' : current process memory usage {currentProcess.WorkingSet64 / 1024 / 1024}MB : '{skimmer.Id}.{skimmer.Name}' : at '{directory}'. Peak memory usage: {currentProcess.PeakWorkingSet64 / 1024 / 1024}MB";
+
+                            context.Logger.LogToolNotification(
+                            new Notification
+                            {
+                                Level = FailureLevel.Note,
+                                Message = new Message
+                                {
+                                    Text = memoryUsage,
+                                },
+                            },
+                            skimmer);
+                        }
+                    }
+
                     skimmer.Analyze(context);
 
                     if (stopwatch != null && context.TargetUri.IsAbsoluteUri)
@@ -1067,12 +1089,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                         string file = context.TargetUri.LocalPath;
                         string directory = Path.GetDirectoryName(file);
                         file = Path.GetFileName(file);
-                        string timing = $"'{file}' : elapsed {stopwatch.Elapsed} : '{skimmer.Name}' : at '{directory}'";
+                        string timing = $"'{file}' : elapsed {stopwatch.Elapsed} : '{skimmer.Id}.{skimmer.Name}' : at '{directory}'";
 
                         context.Logger.LogToolNotification(
                             new Notification
                             {
-                                Level = FailureLevel.Warning,
+                                Level = FailureLevel.Note,
                                 Message = new Message
                                 {
                                     Text = timing,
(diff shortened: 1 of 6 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1060,19 +1060,41 @@ public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TCo
...
                         string directory = Path.GetDirectoryName(file);
                         file = Path.GetFileName(file);
-                        string timing = $"'{file}' : elapsed {stopwatch.Elapsed} : '{skimmer.Name}' : at '{directory}'";
+                        string timing = $"'{file}' : elapsed {stopwatch.Elapsed} : '{skimmer.Id}.{skimmer.Name}' : at '{directory}'";
 
                         context.Logger.LogToolNotification(
                             new Notification
                             {
-                                Level = FailureLevel.Warning,
+                                Level = FailureLevel.Note,
```
shaopeng-gh, line 1097: >Note [](http://example.com/codeflow?start=53&length=4)

The other overall time log is `Note`, suggest to use `Note` as well

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1060,19 +1060,41 @@ public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TCo
...
                         : null;
 
+                    if (context.Traces.HasFlag(DefaultTraces.MemoryUsage))
+                    {
+                        using (var currentProcess = Process.GetCurrentProcess())
+                        {
+                            string file = context.TargetUri.LocalPath;
+                            string directory = Path.GetDirectoryName(file);
+                            file = Path.GetFileName(file);
+                            string memoryUsage = $"'{file}' : current process memory usage {currentProcess.WorkingSet64 / 1024 / 1024}MB : '{skimmer.Id}.{skimmer.Name}' : at '{directory}'. Peak memory usage: {currentProcess.PeakWorkingSet64 / 1024 / 1024}MB";
```
michaelcfanning, line 1070: >memoryUsage [](http://example.com/codeflow?start=35&length=11)

Can you please reply here with a real world example of the new output?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1060,19 +1060,41 @@ public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TCo
                         ? Stopwatch.StartNew()
                         : null;
 
+                    if (context.Traces.HasFlag(DefaultTraces.MemoryUsage))
```
michaelcfanning, line 1063: >context [](http://example.com/codeflow?start=24&length=7)

should we move this earlier in the method, before starting the stop watch? i.e., we don't need to include this code in the timing data.

shaopeng-gh: fixed

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
shaopeng-gh: yes, I tried to use the same format as the one you added,

[this one:]
BINSKIM : note 'BinaryParsers.dll' : current process memory usage 39MB : 'BA2004.EnableSecureSourceCodeHashing' : at 'C:\GH\binskim\bld\bin\x64_Debug\netcoreapp3.1'. Peak memory usage: 39MB

[the one you added before:]
BINSKIM : note 'BinaryParsers.dll' : elapsed 00:00:00.0263046 : 'BA2004.EnableSecureSourceCodeHashing' : at 'C:\GH\binskim\bld\bin\x64_Debug\netcoreapp3.1'

src/Sarif/DefaultTraces.cs
```diff
@@ -10,7 +10,8 @@ public enum DefaultTraces
     {
         None,
         ScanTime = 0x01,
-        ScanExecution = 0x2,
+        ScanExecution = 0x02,
         RuleScanTime = 0x04,
+        MemoryUsage = 0x08,
```
michaelcfanning, line 15: >MemoryUsage [](http://example.com/codeflow?start=8&length=11)

Should we have a global memory usage vs. peak mem data point for the end-to-end scan? And the one per rule/per target? Who is asking for this feature, by the way?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
shaopeng-gh: Now that this is changed to log only once the new format is :
`BINSKIM : note Done. 1,386 files scanned, elapsed time 00:00:25.6855079.`
`BINSKIM : note Peak memory usage: 1213MB.`

src/Sarif/DefaultTraces.cs
shaopeng-gh: fixed to only log once per end of the PR review.

PR #2605, opened 2023-01-12
SARIF has per-line rolling (partial) hash support

suvamM:
This PR adds a new hash utility to compute per-line rolling (partial) hashes for a given file. The PR also provides a port of a [numeric library](https://github.com/dcodeIO/long.js/tree/main) for representing and operating on 64-bit two's complement integers.

PR #2610, opened 2023-01-19
Adding support for partialFingerprints

mpoqq:
Adding partialFingerprint inseret option. Hash contextRegion and add value to sarif. MD5 is used to hash the context. https://github.com/microsoft/sarif-sdk/issues/2569


michaelcfanning (approved):

michaelcfanning:
@mpoqq do you know why build is failing? would you like one of us to investigate?

mpoqq:
@michaelcfanning Sorry I had a bug in one test.
Is it normal that I am unable to see the Azure Pipeline output (401 not authorized - logged in via GitHub)?

michaelcfanning:
Hm, I'm not sure how to address your 401, will research. I did see that one of your tests was failing, glad you've resolved. Time to merge! 

Thanks for this excellent contribution.

mpoqq:
@michaelcfanning Thank you for merging the PR. 
I saw that you also merged [SARIF has per-line rolling (partial) hash support #2605 ](https://github.com/microsoft/sarif-sdk/pull/2605). Would this approach be more suitable for the calculation of  a partial fingerprint? Additionally, as you mentioned, that CodeQL use this approach as well, it would be compatible with CodeQL.

michaelcfanning:
I'd first note that if your log file contains context regions, any consumer could produce a context region hash from that data. The benefit of your feature is that a producer might not want to emit the text snippet or any context region at all (resulting in a slimmer log file).

The CodeQL rolling hash is more sophisticated and part of larger result matching approach that has proved very effective. The key advantage at the rolling hash level is that they handle edits for both sparse and dense content better than the context region hash (which will change completely, for example, if a user simply deletes an erroneous empty line in the region).

See below for the CodeQL (previously Semmle) paper on their end-to-end results matching technique.

The SARIF committee has some very solid content on result matching, in particular see section 5.3.4, in [EN06 SARIF Result Matching](https://github.com/oasis-tcs/sarif-spec/blob/main/Documents/SupplementalContent/EN06%20SARIF%20Result%20Matching.docx) which has a nice, concise summary of the CodeQL approach.

All those details and more are here:
https://semmle.com/wp-content/uploads/2015/01/team-insight.pdf

mpoqq:
Thank you for the detailed answer and the additional information provided.
I now understand that CodeQL's matching approach is much more complex than I assumed.

Now that the rolling hash is included in the SDK, would you appreciate a contribution to the multiool to add the rolling hash partialFingerprint as well?

Unfortunately you provided link to the Semmel paper is not reachable anymore, but I think you meant the following paper: [Tracking Static Analysis Violations over Time to Capture Developer Characteristics](https://ieeexplore.ieee.org/document/7194595)

PR #2614, opened 2023-01-31
Add rolling has optional data enum value. Remove SarifLogger tool parameter.

michaelcfanning:


michaelcfanning:
That's correct, not yet wired up. The plan is to integrate this with the FileRegionsCache, following the same pattern as handling the new line indices in that type. We have a large-scale pipeline design change coming that might lead us to reconsider all hash production, though, which is why I delayed the work.

---
In reply to: [1412629616](https://github.com/microsoft/sarif-sdk/pull/2614#issuecomment-1412629616) [](http://example.com/codeflow?ancestors=1412629616,1412629616)

PR #2616, opened 2023-02-13
Missing context regions for single-line 

michaelcfanning:
@suvamM @HulonJenkins @shaopeng-gh


src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
```diff
@@ -595,14 +714,14 @@
...
                 CharLength = 600,
             };
-
+           
             var fileRegionsCache = new FileRegionsCache();
             region = fileRegionsCache.PopulateTextRegionProperties(region, uri, true, fileContent);
 
             Region multilineRegion = fileRegionsCache.ConstructMultilineContextSnippet(region, uri);
 
-            // 114 (charoffset) + 600 (charlength) + 128 (grabbing right content)
```
michaelcfanning, line 604: > // 114 (charoffset) + 600 (charlength) + 128 (grabbing right content) [](http://example.com/codeflow?start=11&length=70)

This code comment should have been a clue to the problem! We proposed to create context regions that exceeded the size of the analysis target!

This test enforced that. :)

src/Sarif/FileRegionsCache.cs
```diff
@@ -142,6 +142,9 @@ private Region PopulateTextRegionProperties(NewLineIndex lineIndex, Region input
             return region;
         }
 
+        internal const int BIGSNIPPETLENGTH = 512;
```
michaelcfanning, line 145: >BIGSNIPPETLENGTH [](http://example.com/codeflow?start=27&length=16)

Note that I've lifted out these magic values to make them more explicit in test use, and to allow product changes to flow more naturally to test validation.

src/Sarif/FileRegionsCache.cs
```diff
@@ -156,11 +159,10 @@ public Region ConstructMultilineContextSnippet(Region inputRegion, Uri uri, stri
                 return null;
             }
 
-            const int bigSnippetLength = 512;
-            const int smallSnippetLength = 128;
+            fileText ??= newLineIndex.Text;
 
             // Generating full inputRegion to prevent issues.
-            Region originalRegion = this.PopulateTextRegionProperties(inputRegion, uri, populateSnippet: true);
+            Region originalRegion = this.PopulateTextRegionProperties(inputRegion, uri, populateSnippet: true, fileText);
```
michaelcfanning, line 165: >fileText [](http://example.com/codeflow?start=111&length=8)

We were not consistent around passing this optional argument around. This likely would break certain in-memory analysis operations that aren't highly exercised/well-tested today.

src/Sarif/FileRegionsCache.cs
```diff
@@ -184,17 +186,15 @@ public Region ConstructMultilineContextSnippet(Region inputRegion, Uri uri, stri
             region.EndColumn = 0;
             region.StartLine = 0;
             region.EndLine = 0;
-            region.CharOffset = originalRegion.CharOffset < smallSnippetLength
-                ? 0
-                : originalRegion.CharOffset - smallSnippetLength;
+            region.CharOffset = Math.Max(0, originalRegion.CharOffset - SMALLSNIPPETLENGTH);
```
michaelcfanning, line 189: >CharOffset [](http://example.com/codeflow?start=19&length=10)

Any prepended context region must start at the beginning of the file, any value that precedes this is invalid.

src/Sarif/FileRegionsCache.cs
```diff
@@ -184,17 +186,15 @@ public Region ConstructMultilineContextSnippet(Region inputRegion, Uri uri, stri
...
 
-            region.CharLength = originalRegion.CharLength + region.CharOffset + smallSnippetLength < newLineIndex.Text.Length
-                ? originalRegion.CharLength + smallSnippetLength + Math.Abs(region.CharOffset - originalRegion.CharOffset)
-                : newLineIndex.Text.Length - region.CharOffset;
+            region.CharLength =
```
michaelcfanning, line 191: >CharLength [](http://example.com/codeflow?start=19&length=10)

Similarly, any context region that blows past the end of the actual file is invalid.

michaelcfanning:
>         public void FileRegionsCache_IncreasingToLeftAndRight()

This test really should have been built out more. It tests one idiosyncratic case in our idiosyncratic system.

If the case when have a context region that itself exceeds the large snippet size (of 512 chars), our solution will add up to 128 chars of additional text to it, starting with the prepended side. This test shows that only the overflow from the unused 128 chars is appended to the region.

All of this behavior should first be reexamined and then tested in directed tests. This change is a hammer to ensure that all cases, in fact, produce context regions. We should ship this and then later fix this:

[https://github.com/microsoft/sarif-sdk/issues/2617](https://github.com/microsoft/sarif-sdk/issues/2617)



---
Refers to: src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs:707 in 520eb13. [](commit_id = 520eb131776fa427eb69453bebbcc7088bde10af, deletion_comment = False)

src/Sarif/FileRegionsCache.cs
shaopeng-gh: do we prefer put at the top of the file with other definitions

shaopeng-gh (approved):
:shipit:

src/Sarif/FileRegionsCache.cs
michaelcfanning: Yes, will do this in a follow-on change, thanks for the review!!

PR #2618, opened 2023-02-14
Threading fixes

michaelcfanning:
Ouch. We previously created all tasks on a common thread. We have a mile-wide bug in our unit-testing which we need to close.

@shaopeng-gh


shaopeng-gh (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -11,6 +11,8 @@
 * BRK: For `Guid` properties defined in SARIF spec, updated Json schema to use `uuid`, and updated C# object model to use `Guid?` instead of `string`. [#2555](https://github.com/microsoft/sarif-sdk/pull/2555)
 * BRK: Mark `AnalyzeCommandBase` as obsolete. This type will be removed in the next significant update.  [#2599](https://github.com/microsoft/sarif-sdk/pull/2599)
 * BRK: `LogUnhandledEngineException` no longer has a return value (and updates the `RuntimeErrors` context property directly as other helpers do).  [#2599](https://github.com/microsoft/sarif-sdk/pull/2599)
+* BUG: Populate missing context region data for small, single-line scan targets. [#2616](https://github.com/microsoft/sarif-sdk/pull/2616)
+* BUG: Increase parallelism in `MultithreadedAnalyzeCommandBase` by correcting task creation. []#2618](https://github.com/microsoft/sarif-sdk/pull/2618)
```
shaopeng-gh, line 15: >[] [](http://example.com/codeflow?start=94&length=2)

typo

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -218,20 +218,20 @@ private void Analyze(TOptions analyzeOptions, AggregatingLogger logger)
...
             // 1: First we initiate an asynchronous operation to locate disk files for
             // analysis, as specified in analysis configuration (file names, wildcards).
-            Task<bool> enumerateFilesOnDisk = EnumerateFilesOnDiskAsync(options);
+            Task<bool> enumerateFilesOnDisk = Task.Run(() => EnumerateFilesOnDiskAsync(options));
 
             // 2: Files found on disk are put in a specific sort order, after which a 
             // reference to each scan target is put into a channel for hashing,
             // if hashing is enabled.
-            Task<bool> hashFilesAndPutInAnalysisQueue = HashFilesAndPutInAnalysisQueueAsnc();
+            Task<bool> hashFilesAndPutInAnalysisQueue = Task.Run(() => HashFilesAndPutInAnalysisQueueAsnc());
```
suvamM, line 226: Spelling fix for `Async`

suvamM (approved):
@michaelcfanning This looks good to me. But I had a question about the procedure: this an open PR. There is already a release for `sarif-pattern-matcher` which references a commit in this PR. Ideally, we should have dependencies from the `main` branch of `sarif-sdk` and not dev branches.

michaelcfanning:
We don't currently sim-ship the SDK when we ship the scanner. This is because the scanner is intended to be a client tool that consumes the OSS source as source, i.e., via submodule. We don't take a NuGet package dependency and so there isn't pressure to keep the NuGet delivery in sync with sarif pattern matcher. Open to your thoughts on this topic. 

---
In reply to: [1299909340](https://github.com/microsoft/sarif-sdk/pull/2618#issuecomment-1299909340) [](http://example.com/codeflow?ancestors=1299909340,1299909340)

PR #2619, opened 2023-02-16
Memory tuning

michaelcfanning:
This change makes some tactical changes to the current threading model in order to maximize CPU utilization and limit memory consumption, while preserving the deterministic output required for certain current client tool users.

@kelewis has pointed out that building analysis around a limited number of flattened records designed to enable efficient results streaming would be much more efficient. This seems like the right approach to me; I am only uncertain about how we would retrofit determinism for those who need it. We could create a post-processing sorting API. We could design a new binary format for serialized records, optimized for reconstructing into sorted SARIF. Looking forward to the conversation.

As it turns out, our runaway memory usage was do to a simple bug, the context creating channel wasn't throttled on write. Oops.

Tactical changes include:
* Throttle context creation to, at most, 25k objects. I will make this configurable in the short-term.
* Conversion of failure level and result kind to immutable sets, rather than list instances. We now prefer constant values for the default. These allocations were unnecessary and cluttered profiling results, so good to clear.
* Added a simple measure of working set consumed at analysis time.
* Eliminated `HashUtilities.FileSystem`, an unnecessary duplication of the `FileSystem.Instance` test knob. Taking this change required adding an `IFileSystem` instance to relevant API (a non-breaking change).
* Lowered individual 'file skipped due to size' message to note. Generally, we have a couple of issues with diagnostics reporting: these values can't be individually controlled via the failure levels. They likely should be separated from them entirely, as trace values. The trace messages arguably should always be emitted, when configured. Would love to discuss this further with the team.
* Updated 'one or more files skipped' warning to include a count of files skipped. Also moved this warning to the end of analysis (rather than embedding it in the middle of analysis results, when file enumeration is complete).
* This change aggressively flushes SARIF json to disk, so that a concentration of results doesn't increase memory consumption. This writing occurs on a single thread and the incremental loss of IO efficiency is unlikely to be observeable as it's amortized alongside the multithreaded scanning.
* We also now no longer take an exclusive lock on the SARIF log. This allows users to open an in-progress log or us a tool like `tail.exe` to watch output as it's generated.
* I have disabled the caching/replay of file results by hash. This functionality isn't of general value and it definitely shouldn't be tied to the configuration to emit file hashes for files that are referenced in SARIF. This capability will come back, therefore, as a new config knob. In the meantime, we've dropped the hashing channel, which is helpful for lowering memory usage.


suvamM:
@michaelcfanning A PR description would be super helpful.

src/Sarif/DefaultTraces.cs
```diff
@@ -10,7 +10,7 @@ public enum DefaultTraces
     {
         None,
         ScanTime = 0x01,
-        ScanExecution = 0x2,
```
michaelcfanning, line 13: >ScanExecution [](http://example.com/codeflow?start=8&length=13)

'ScanExecution' previously unused.

src/Sarif.Driver/Sdk/AnalyzeCommandBase.cs
```diff
@@ -30,7 +30,6 @@ public abstract class AnalyzeCommandBase<TContext, TOptions> : PluginDriverComma
         private Run _run;
         private Tool _tool;
         internal TContext _rootContext;
-        private CacheByFileHashLogger _cacheByFileHashLogger;
```
michaelcfanning, line 33: >_cacheByFileHashLogger [](http://example.com/codeflow?start=38&length=22)

I've clobbered caching by file hash in general with this change. It will not come back for the single-threaded analysis, as this engine is getting dropped entirely.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -100,7 +104,21 @@
             HelpText = "Execution traces, expressed as a semicolon-delimited list, that " +
                        "should be emitted to the console and log file (if appropriate). " +
                        "Valid values: ScanTime.")]
-        public virtual IEnumerable<string> Traces { get; set; }
+        public virtual IEnumerable<string> Trace { get; set; }
+
+        private DefaultTraces? defaultTraces;
+        public DefaultTraces Traces
```
michaelcfanning, line 110: >Traces [](http://example.com/codeflow?start=29&length=6)

This pattern may be worth pursuing aggressively, i.e., some class surface area for interacting directly with the commandline library vs. reconstructed data more suitable for direct consumption by the engine.

I keep wondering whether options could be merged entirely with the context object...

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -193,21 +191,14 @@ private void Analyze(TOptions analyzeOptions, AggregatingLogger logger)
...
 
-            var channelOptions = new BoundedChannelOptions(2000)
-            {
-                SingleWriter = true,
-                SingleReader = true,
-            };
-            readyToHashChannel = Channel.CreateBounded<int>(channelOptions);
-
-            channelOptions = new BoundedChannelOptions(2000)
+            var channelOptions = new BoundedChannelOptions(25000)
```
michaelcfanning, line 194: >25000 [](http://example.com/codeflow?start=59&length=5)

We allow the scan phase to build up 25k targets, as candidates, or completed files that haven't triggered the logging thread.

This all designed to maximize CPU utilization for the scan phase.

PR #2625, opened 2023-02-22
New model

michaelcfanning:
Foundational work for updated threading model that better utilizes CPU while throttling memory use and allows for cancellation and timeout. The code is refactored to allow API consumers to plug-in at the highest level possible (rather than calling into individual methods to instantiate skimmers, analyze targets, etc.). Consumers can drive analysis strictly from command-line options (mirroring client tool behavior closely) or simply create a new (greatly expanded) context object that is sufficient to drive end-to-end analysis.

As a result of the context work, all parameterization can be expressed as an XML configuration file as well.

* BRK: `MultithreadedAnalyzeCommandBase` IDispose implementation now manages logging dispose. Be sure to call `base.Dispose()` in any derived type implementations. [#2614](https://github.com/microsoft/sarif-sdk/pull/2614)
* BRK: Rename `LogFilePersistenceOptions` to `FilePersistenceOptions` (due to its general applicability in other file persistence contexts other than output logs).[#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* BRK: Many breaking changes in `IAnalysisContext` and `AnalyzeContextBase`. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* BUG: Eliminate per-context allocations contributing to unnecessary memory use. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* NEW: Rewrite  `MultithreadedAnalyzeCommandBase` pipeline to allow for timeout, cancellation, and better API-driven use. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* NEW: Move large amounts of scan data to the context object, to streamline pipeline and allow for XML-driven configuration. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* NEW: Switch file processing to an `ArtifactProvider` model where enumerated artifacts consist of URI and optional content. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* NEW: Add new `FailureLevelSet` and `ResultKindSet` types that are compatible with XML-based configuration. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
* NEW: Add `PeakWorkingSet` to `--trace` command to report maximum working set value during analysis. [#2619](https://github.com/microsoft/sarif-sdk/pull/2619)

This is a large change and it's going in with many `TBD` comments I will address next week. Everyone will have further opportunity to review as it goes into `sarif-pattern-matcher`, where the new pipeline is highly exercised.


src/Sarif/AggregatingArtifactsProvider.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class AggregatingArtifactsProvider : IArtifactProvider
```
michaelcfanning, line 9: >AggregatingArtifactsProvider [](http://example.com/codeflow?start=17&length=28)

An aggregating artifacts provider is an artifacts provider that itself aggregates scan targets generated by one or more providers. Directly analogous to our aggregating logger.

Note that individual providers own problems such as skipping files which aren't appropriate for scanning (because, for example, they exceed a size threshold) and for reporting those that were skipped.

src/Sarif/AnalyzeContextBase.cs
```diff
@@ -16,42 +19,254 @@ public AnalyzeContextBase()
         // All of these properties are persisted to configuration XML and can be
         // passed using that mechanism. All command-line arguments are 
         // candidates to follow this pattern.
-        public IEnumerable<IOption> GetOptions()
+        public virtual IEnumerable<IOption> GetOptions()
```
michaelcfanning, line 22: >GetOptions [](http://example.com/codeflow?start=44&length=10)

This is the way that a class declares it has serializable properties. This array is not complete. 

An 'export-config' command will use MEF to retrieve all instances of `IOptionsProvider` and subsequently dump all published properties to an XML representation. What's serialized is what's expressed in the underlying static descriptors.

src/Sarif/AnalyzeContextBase.cs
```diff
@@ -16,42 +19,254 @@ public AnalyzeContextBase()
...
+        public virtual IAnalysisLogger Logger { get; set; }
+        public virtual RuntimeConditions RuntimeErrors { get; set; }
+        public virtual bool AnalysisComplete { get; set; }
+
+        public bool Minify => OutputFileOptions.HasFlag(FilePersistenceOptions.Minify);
+        public bool Optimize => OutputFileOptions.HasFlag(FilePersistenceOptions.Optimize);
+        public bool PrettyPrint => OutputFileOptions.HasFlag(FilePersistenceOptions.PrettyPrint);
+        public bool ForceOverwrite => OutputFileOptions.HasFlag(FilePersistenceOptions.ForceOverwrite);
+
+        public virtual string AutomationId
```
michaelcfanning, line 64: >AutomationId [](http://example.com/codeflow?start=30&length=12)

This is the way we create class 'surface area' for properties that are backed by the policy property bag. That dictionary can be kept empty, in which case any retrieval of the static gets the default value from the static descriptor.

This also means that the policy dictionary only contains data values that override defaults, which allows for very compact serialized representation.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class EnumeratedArtifact : IEnumeratedArtifact
```
michaelcfanning, line 10: >EnumeratedArtifact [](http://example.com/codeflow?start=17&length=18)

EnumeratedArtifact is a thing.

It may exist only as a URI reference, in which case the URI can be resolved to retrieve it.

It can be populated with a stream, in which case there's no need to retrieve/download/load it.

It can be pre-populated with a string representation. We allow this because in the majority of cases we are in the business of analyzing strings. We might need to reconsider this approach, though, as other scan targets won't always be a string. 

Probably, this type needs to be generic, with EnumeratedArtifact<string> ultimately resolving to a string.

src/Sarif/Errors.cs
```diff
@@ -84,7 +88,7 @@ public static void LogExceptionLoadingTarget(IAnalysisContext context)
             // Could not instantiate skimmers from the following plugins: {0}
             context.Logger.LogConfigurationNotification(
                 CreateNotification(
-                    context.TargetUri,
```
michaelcfanning, line 87: >TargetUri [](http://example.com/codeflow?start=28&length=9)

there is no target uri for this and some other error handlers i cleaned up.

src/Sarif/ExtensionMethods.cs
```diff
@@ -190,15 +190,16 @@ public static string ToAndPhrase(this List<string> words)
             }
         }
 
-        public static OptionallyEmittedData ToFlags(this IEnumerable<OptionallyEmittedData> optionallyEmittedData)
+        public static TFlags ToFlags<TFlags>(this IEnumerable<TFlags> enumeratedFlags) where TFlags : Enum, IConvertible
```
michaelcfanning, line 193: >ToFlags [](http://example.com/codeflow?start=29&length=7)

made this work for any enum type that is int32

src/Sarif/FailureLevelSet.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    [Serializable]
+    [JsonConverter(typeof(TypedPropertiesDictionaryConverter))]
+    public class FailureLevelSet : HashSet<FailureLevel>
```
michaelcfanning, line 14: >FailureLevelSet [](http://example.com/codeflow?start=17&length=15)

Enable clean serialization of this set type to and from XML.

src/Sarif/Notes.cs
```diff
@@ -43,9 +22,9 @@ public static void LogFileSkippedDueToSize(IAnalysisContext context, string skip
             // '{0}' was skipped as its size ({1} kilobytes) exceeds the currently configured threshold ({2} kilobytes).
             context.Logger.LogConfigurationNotification(
                 Errors.CreateNotification(
-                    context.TargetUri,
+                    new Uri(skippedFile, UriKind.RelativeOrAbsolute),
                     Msg002_FileSkippedDueToSize,
-                    ruleId: Msg002_FileSkippedDueToSize,
+                    ruleId: null,
```
michaelcfanning, line 27: >null [](http://example.com/codeflow?start=28&length=4)

oops, this is an associated rule id, not the notification id, and should have been null.

src/Sarif/PropertiesDictionaryExtensionMethods.cs
```diff
@@ -95,6 +95,20 @@ public static class PropertiesDictionaryExtensionMethods
                     continue;
                 }
 
+                FailureLevelSet failureLevelSet = property as FailureLevelSet;
```
michaelcfanning, line 98: >failureLevelSet [](http://example.com/codeflow?start=32&length=15)

changes in this file allow for clean serialization and deserialization of sets that contain failure levels and result kinds, highly used to configure analysis.

src/Sarif/Core/Notification.cs
```diff
@@ -12,14 +14,19 @@ public override string ToString()
         {
             var sb = new System.Text.StringBuilder();
 
-            sb.Append(this.Locations?[0].PhysicalLocation?.ArtifactLocation?.Uri);
-            sb.Append(" : ").Append(this.Descriptor.Id);
-            sb.Append(" : ").Append(this.AssociatedRule?.Id);
-            sb.Append(" : ").Append(this.Level);
+            if (this.TimeUtc != default) { sb.Append($"{this.TimeUtc}"); }
```
michaelcfanning, line 17: >this [](http://example.com/codeflow?start=16&length=4)

Provides a very nice way in the debugger to see what unexpected stuff is getting persisted to configuration and tool notifications, often something like an unhandled engine exception.

src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -15,7 +15,7 @@ public class ConsoleLogger : BaseLogger, IAnalysisLogger
         //  TODO:  We directly instantiate this logger in two classes, creating 
         //  unamanged dependencies.  Fix this pattern with dependency injection or a factory.
         //  #2272 https://github.com/microsoft/sarif-sdk/issues/2272
-        public ConsoleLogger(bool quietConsole, string toolName, IImmutableSet<FailureLevel> levels = null, IImmutableSet<ResultKind> kinds = null) : base(levels, kinds)
```
michaelcfanning, line 18: >IImmutableSet [](http://example.com/codeflow?start=65&length=13)

immutable set can't easily be serialized/deserialized, so we create alias types to these constructs.

src/Sarif/FailureLevelSet.cs
michaelcfanning: Note that we could make these sets immutable, could be desirable.

PR #2627, opened 2023-02-27
Update pipeline for multithreaded analysis.

michaelcfanning:
This change deprecates a duplicative single exception for runtime failures expressed as `MulthreadedAnalyzeCommandBase.EngineException` and `IAnalysisContext.RuntimeException`, in favor of a new property, `IAnalysisContext.RuntimeExceptions`, that can hold multiple exceptions.

This allows for tracking errors across multiple threads, e.g., an unhandled exception in a rule as well as a timeout exception.

Otherwise, this change cleans up error handling in the core threading model.

### DIFF
--- a/src/Sarif/AnalyzeContextBase.cs
+++ b/src/Sarif/AnalyzeContextBase.cs
@@ -53,7 +53,7 @@ namespace Microsoft.CodeAnalysis.Sarif
         public virtual IEnumeratedArtifact CurrentTarget { get; set; }
         public virtual string MimeType { get; set; }
         public virtual HashData Hashes { get; set; }
-        public virtual Exception RuntimeException { get; set; }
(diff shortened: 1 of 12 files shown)


src/Sarif/AnalyzeContextBase.cs
```diff
@@ -53,7 +53,7 @@ public virtual IEnumerable<IOption> GetOptions()
         public virtual IEnumeratedArtifact CurrentTarget { get; set; }
         public virtual string MimeType { get; set; }
         public virtual HashData Hashes { get; set; }
-        public virtual Exception RuntimeException { get; set; }
+        public virtual IList<Exception> RuntimeExceptions { get; set; }
```
michaelcfanning, line 56: >RuntimeExceptions [](http://example.com/codeflow?start=40&length=17)

We will retain exceptions across multiple threads in some cases.

PR #2629, opened 2023-02-28
Snippet production

michaelcfanning:
We now create context region snippets that are restricted to 512 bytes in size. In the event that the original snippet is larger than this, we simply recapitulate that snippet in the log data (i.e., the context region is identical).

### DIFF
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -164,6 +164,11 @@ namespace Microsoft.CodeAnalysis.Sarif
             // Generating full inputRegion to prevent issues.
             Region originalRegion = this.PopulateTextRegionProperties(inputRegion, uri, populateSnippet: true, fileText);
 
+            if (originalRegion.CharLength >= BIGSNIPPETLENGTH)
+            {
+                return originalRegion.DeepClone();
+            }
+
             int maxLineNumber = newLineIndex.MaximumLineNumber;
 
             var region = new Region
@@ -188,9 +193,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             region.EndLine = 0;
             region.CharOffset = Math.Max(0, originalRegion.CharOffset - SMALLSNIPPETLENGTH);
 
-            region.CharLength =
-                Math.Min(originalRegion.CharLength + region.CharOffset + SMALLSNIPPETLENGTH,
-                         fileText.Length - region.CharOffset);
+            region.CharLength = Math.Min(BIGSNIPPETLENGTH, fileText.Length - region.CharOffset);
 
             // Generating multiline region with 128 characters to the left and right from the
             // originalRegion if possible.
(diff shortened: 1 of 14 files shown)


michaelcfanning:
Need a review asap on this. @suvamM @EasyRhinoMSFT @shaopeng-gh @yongyan-gh @HulonJenkins

src/Sarif/FileRegionsCache.cs
```diff
@@ -164,6 +164,11 @@ public Region ConstructMultilineContextSnippet(Region inputRegion, Uri uri, stri
             // Generating full inputRegion to prevent issues.
             Region originalRegion = this.PopulateTextRegionProperties(inputRegion, uri, populateSnippet: true, fileText);
 
+            if (originalRegion.CharLength >= BIGSNIPPETLENGTH)
```
EasyRhinoMSFT, line 167: >= [](http://example.com/codeflow?start=43&length=1)

So 512 is considered big?

suvamM: +1. This should be >

src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
```diff
@@ -745,8 +740,7 @@ public void FileRegionsCache_PopulatesWithOneLine_IncreasingToTheRight()
             region = fileRegionsCache.PopulateTextRegionProperties(region, uri, true, content);
             Region multilineRegion = fileRegionsCache.ConstructMultilineContextSnippet(region, uri);
 
-            // CharLength + 128 to the right = 428 characters
-            multilineRegion.CharLength.Should().Be(300 + 128);
+            multilineRegion.CharLength.Should().Be(512);
```
suvamM, line 743: Curious, with the bug regarding `>` above, how does this test pass?

suvamM:
Just wanted to comment on the fantastic PR description: brief and informative. Thanks @michaelcfanning

src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
schlaman-ms, line 743: And shouldn't that be  a constant?

suvamM: +1 to Craig's comment. We should have an `Assume` section in the test defining the constants.

michaelcfanning:
:)

---
In reply to: [1448659817](https://github.com/microsoft/sarif-sdk/pull/2629#issuecomment-1448659817) [](http://example.com/codeflow?ancestors=1448659817,1448659817)

src/Sarif/FileRegionsCache.cs
michaelcfanning: why? if the size of the original region is as big as our bigsnippet maximum, why would attempt to generate it? what am i missing?

src/Test.UnitTests.Sarif/FileRegionsCacheTests.cs
michaelcfanning: We now pick it up from the product, good call.

michaelcfanning: I don't understand the bug that you see. big snippet is defined as 512.

src/Sarif/FileRegionsCache.cs
EasyRhinoMSFT: I was verifying that the max length is 511, since the >= operator excludes 512. This seems to be an off-by-one considering the statement that snippets are "restricted to 512 bytes in size"

PR #2632, opened 2023-03-03
If HDF CodeDesc is empty, use Desc as Message

candrews:
Sometimes, the HDF CodeDesc is empty. Currently, that results in the SARIF Message being set to "." which isn't good.

To fix that issue, if CodeDesc is empty, use the Desc.

I've reported one case where the HDF `code_desc` is set to the empty string at https://github.com/mitre/saf/issues/1163

### DIFF
--- a/src/Sarif.Converters/HdfConverter.cs
+++ b/src/Sarif.Converters/HdfConverter.cs
@@ -126,7 +126,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                     RuleId = execJsonControl.Id,
                     Message = new Message
                     {
-                        Text = AppendPeriod(controlResult.CodeDesc),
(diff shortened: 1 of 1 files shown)


michaelcfanning:
ok! this makes sense as a mitigation, particularly as you have done the due diligence of pursuing an upstream solution to the root problem. :)

if you will create a branch off main with your change and update the release note with an appropriate summary of the goodness, we will approve and merge.

thanks for the contribution!

candrews:
> ok! this makes sense as a mitigation, particularly as you have done the due diligence of pursuing an upstream solution to the root problem. :)

Fantastic! That's my sentiment precisely :)

> if you will create a branch off main with your change and update the release note with an appropriate summary of the goodness, we will approve and merge.

I've submitted a new PR as you described: https://github.com/microsoft/sarif-sdk/pull/2635

> 
> thanks for the contribution!

You're welcome!

PR #2635, opened 2023-03-03
If HDF CodeDesc is empty, use Desc as Message

candrews:
Sometimes, the HDF CodeDesc is empty. Currently, that results in the SARIF Message being set to "." which isn't good.

To fix that issue, if CodeDesc is empty, use the Desc.

I've reported one case where the HDF `code_desc` is set to the empty string at https://github.com/mitre/saf/issues/1163

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -4,6 +4,7 @@
 * BRK: Eliminate `MulthreadedAnalyzeCommandBase.EngineException` and `IAnalysisContext.RuntimeException` properties in favor of `IAnalysisContext.RuntimeExceptions`. [#2627](https://github.com/microsoft/sarif-sdk/pull/2627)
 * BRK: Rename `LogFilePersistenceOptions` to `FilePersistenceOptions` (due to its general applicability in other file persistence contexts other than output logs).[#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
 * BRK: Many breaking changes in `IAnalysisContext` and `AnalyzeContextBase`. [#2625](https://github.com/microsoft/sarif-sdk/pull/2625)
+* BUG: In `HDFConverter` if `code_desc` is empty, use `desc` as the SARIF `message`. [#2632](https://github.com/microsoft/sarif-sdk/pull/2632)
 * BUG: Eliminate creation of extremely large context region snippets (now always restricted to 512 chars). https://github.com/microsoft/sarif-sdk/pull/2629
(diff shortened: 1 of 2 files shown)


candrews:
@michaelcfanning how does this PR look?

I'm eager to get this merged and see the improvement in a release :-)

src/Sarif.Converters/HdfConverter.cs
```diff
@@ -126,7 +126,7 @@ private static (ReportingDescriptor, IList<Result>) SarifRuleAndResultFromHdfCon
                     RuleId = execJsonControl.Id,
                     Message = new Message
                     {
-                        Text = AppendPeriod(controlResult.CodeDesc),
+                        Text = AppendPeriod(string.IsNullOrWhiteSpace(controlResult.CodeDesc) ? execJsonControl.Desc : controlResult.CodeDesc),
```
michaelcfanning, line 129: >IsNullOrWhiteSpace [](http://example.com/codeflow?start=51&length=18)

are you sure this logic is what you want? I think you have the arguments reversed maybe?

ReleaseHistory.md
```diff
@@ -1,5 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.1.0** UNRELEASED
+* BUG: If HDF `code_desc` is empty, use `desc` as the SARIF `message`. [#2632](https://github.com/microsoft/sarif-sdk/pull/2632)
```
michaelcfanning, line 3: >BUG [](http://example.com/codeflow?start=2&length=3)

All BUG entries belong together in the list below, since I asked you to review another possible change could you make this one?

Also let's reference `HDFConverter` to make things a little more explicit:

BUG: If `HDFConverter` `code_desc`... etc.

candrews: I've made that update - how's it look now?

src/Sarif.Converters/HdfConverter.cs
michaelcfanning: Um, yes. :) Haven't had my coffee yet. The good news is my mistake squeezed a release note improvement out of you. :)

ReleaseHistory.md
michaelcfanning: Looks great! I'm going to ship this right now for you via NuGet.

michaelcfanning:
Your changes are published to NuGet. Thanks for the contribution! Come back again. :)

https://www.nuget.org/packages/Sarif.Sdk

PR #2637, opened 2023-03-07
Target analysis completed

michaelcfanning:
The primary point of this PR is to restore callbacks to `IAnalysisLogger.AnalyzingTarget` implementations (removed in recent pipeline rewrite). While the hood is up, added a new `IAnalysisLogger.TargetAnalyzed` callback.

`AnalyzingTarget` is called from the target enumeration thread. Strictly speaking, this isn't when analysis begins. The scan threads do not have a pointer to the global context logger, however, so changing this would require more thought.

`TargetAnalyzed` is called from the logging thread and so synchronization is required again on the global context object. 

As part of this change, I noticed some console logger messages that never fired (because the callback wasn't executed) tracking the start of analysis. This is verbose logging, so I put it behind a new trace flag (and added analogous handling in the target analyzed callback).

@eddynaka @suvamM @shaopeng-gh @easyrhino-gh @cfaucon

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -242,7 +242,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             context.ResultKinds = options.ResultKinds;
             context.FailureLevels = options.FailureLevels;
             context.OutputFileOptions = options.OutputFileOptions.ToFlags();
-            context.MaxFileSizeInKilobytes = options.MaxFileSizeInKilobytes;
+            context.MaxFileSizeInKilobytes = options.MaxFileSizeInKilobytes != null ? options.MaxFileSizeInKilobytes.Value : context.MaxFileSizeInKilobytes;
             context.PluginFilePaths = options.PluginFilePaths?.ToImmutableHashSet();
             context.InsertProperties = InitializeStringSet(options.InsertProperties);
             context.TargetFileSpecifiers = InitializeStringSet(options.TargetFileSpecifiers);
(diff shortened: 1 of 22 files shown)


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -36,6 +36,31 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
...
+            {
+                TargetFileSpecifiers = new List<string>(new[] { "*" })
+            };
+
+            var logger = new TestMessageLogger();
+
+            var context = new TestAnalysisContext
+            { 
+                Logger = logger,
+                MaxFileSizeInKilobytes = long.MaxValue,
```
michaelcfanning, line 54: >MaxValue [](http://example.com/codeflow?start=46&length=8)

Set the max value so that we don't skip any files on enumerating the current working directory.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -242,7 +242,7 @@ public virtual TContext InitializeContextFromOptions(TOptions options, ref TCont
             context.ResultKinds = options.ResultKinds;
             context.FailureLevels = options.FailureLevels;
             context.OutputFileOptions = options.OutputFileOptions.ToFlags();
-            context.MaxFileSizeInKilobytes = options.MaxFileSizeInKilobytes;
+            context.MaxFileSizeInKilobytes = options.MaxFileSizeInKilobytes != null ? options.MaxFileSizeInKilobytes.Value : context.MaxFileSizeInKilobytes;
```
michaelcfanning, line 245: >MaxFileSizeInKilobytes [](http://example.com/codeflow?start=94&length=22)

This pattern is required in order to permit a combination of options- and context-specified configuration. We need to make a general scrub for this pattern I've implemented for max file size only.

cfaucon (approved):

PR #2639, opened 2023-03-13
API/other improvements for improved in-memory analysis.

michaelcfanning:


src/Sarif/SarifExtensionMethods.cs
```diff
@@ -15,8 +15,18 @@
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
-    public static class ExtensionMethods
+    public static class SarifExtensions
```
michaelcfanning, line 18: >SarifExtensions [](http://example.com/codeflow?start=24&length=15)

For publicly exposed extension methods classes, it's good to qualify the type name, to avoid collisions with this general name 'ExtensionMethods'.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -101,11 +100,6 @@ public class SarifLogger : BaseLogger, IDisposable, IAnalysisLogger
             RuleToIndexMap = new Dictionary<ReportingDescriptor, int>(ReportingDescriptor.ValueComparer);
             ExtensionGuidToIndexMap = new Dictionary<Guid, int>();
 
-            if (dataToInsert.HasFlag(OptionallyEmittedData.Hashes))
```
michaelcfanning, line 104: >dataToInsert [](http://example.com/codeflow?start=16&length=12)

I'm deleting this aggressive hashing of targets data on initialization for now. First, we previously had at least three mechanisms for hashing, this one, the hash delegate and on-demand hashing via the insert data visitor. This particular mechanism always held all this data in memory, as opposed to the file regions cache mechanism, which only holds that previous 100 files hashed...

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -197,12 +192,13 @@
 
                 foreach (string target in analysisTargets)
                 {
-                    Uri uri = new Uri(UriHelper.MakeValidUri(target), UriKind.RelativeOrAbsolute);
+                    string uriText = UriHelper.MakeValidUri(target);
```
michaelcfanning, line 195: >MakeValidUri [](http://example.com/codeflow?start=47&length=12)

We call this helper inconsistently and really should call it everywhere.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -262,10 +258,6 @@
             _run.Invocations.Add(invocation);
         }
 
-        public Func<Uri, HashData> ComputeHashData { get; set; }
```
michaelcfanning, line 265: >ComputeHashData [](http://example.com/codeflow?start=35&length=15)

These mechanisms deleted in preference of a share file regions cache.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -113,36 +111,18 @@
         [Option(
             "level",
             Separator = ';',
-            Default = new FailureLevel[] { FailureLevel.Error, FailureLevel.Warning },
             HelpText = "A semicolon delimited list to filter output of scan results to one or more failure levels. Valid values: Error, Warning and Note.")]
         public IEnumerable<FailureLevel> Level { get; set; }
 
-        private FailureLevelSet failureLevels;
```
michaelcfanning, line 120: >FailureLevelSet [](http://example.com/codeflow?start=15&length=15)

Providing defaults in the options classes is problematic because it interferes with the context object default value functionality (which is much richer/close to the analysis model.). In the options classes, we should actually return null for all cases where the property isn't explicit on the command-line, otherwise reflect the command-line precisely. That's the model. There's more clean-up to do.

src/Sarif.Driver/Sdk/CommandBase.cs
```diff
@@ -95,11 +95,9 @@ public static HashSet<string> CreateTargetsSet(IEnumerable<string> targetSpecifi
 
                 if (Uri.TryCreate(specifier, UriKind.RelativeOrAbsolute, out Uri uri))
                 {
-                    if (uri.IsAbsoluteUri && (uri.IsFile || uri.IsUnc))
-                    {
-                        normalizedSpecifier = uri.LocalPath;
-                    }
+                    normalizedSpecifier = uri.GetFileName();
```
michaelcfanning, line 98: >GetFileName [](http://example.com/codeflow?start=46&length=11)

Here's another clean-up item, we need to call the GetFilePath and GetFileName helpers consistently across the code base (to properly process things like relative URLs).

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -38,13 +38,13 @@
 
         public static bool RaiseUnhandledExceptionInDriverCode { get; set; }
 
-        protected virtual Tool Tool { get; set; }
+        public virtual Tool Tool { get; set; }
```
michaelcfanning, line 41: >public [](http://example.com/codeflow?start=8&length=6)

This property needs to be public in order to interact with it properly in scenarios where users proactively create a SARIF logger (and where skimmers may be loaded as extensions).

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -236,24 +236,27 @@
             context.OutputFilePath = options.OutputFilePath;
             context.AutomationGuid = options.AutomationGuid;
             context.BaselineFilePath = options.BaselineFilePath;
-            context.Traces = InitializeStringSet(options.Trace);
+            context.Traces = options.Trace != null ? InitializeStringSet(options.Trace) : context.Traces;
```
michaelcfanning, line 239: >Trace [](http://example.com/codeflow?start=37&length=5)

This pattern needs to be pushed through-out this code. 

If options return null consistently, then we can override an existing config object with updated values but ONLY if the user has expressed an override on the command-line.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -511,7 +511,7 @@ public void SarifLogger_ScrapesFilesFromResult()
             {
                 using (var sarifLogger = new SarifLogger(textWriter,
                                                          analysisTargets: null,
-                                                         dataToInsert: OptionallyEmittedData.Hashes,
+                                                         dataToInsert: OptionallyEmittedData.None,
```
michaelcfanning, line 514: >None [](http://example.com/codeflow?start=93&length=4)

Interesting bug! Because we didn't have sufficient handling for the `dataToInsert` property, we missed a case where this data was requested but not available. The fix now prompts request for missing data. Since that's not the purpose of this test, I just dropped the request for hashes data.

src/Test.UnitTests.Sarif.Driver/SarifHelpers.cs
```diff
@@ -1,48 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-
-using Xunit;
-
-namespace Microsoft.CodeAnalysis.Sarif.Driver
```
michaelcfanning, line 8: >Microsoft [](http://example.com/codeflow?start=10&length=9)

Turns out that test code doesn't use this delegate action injection mechanism highly, so I just ripped it out.

Moving forward we should be updating tests to use the new improved analysis model (which is designed to allow solid testing using the core analyze command mechanism, just as the actual client tool does). So this sort of testing model should wither away, rather than being built on.

PR #2640, opened 2023-03-13
Exception thrown during DetermineFromFileExtension when given http with query

kelewis:
When sending web uris with query parameters as "filenames" the mime type detection often would generate an exception.  Specifically, the string would fail all the extension checks and fall into the local Directory check, which fails with the https: uri and throws an invalid char in filepath exception.

Changed the detection function to only take Uri (not string filenames) and uses more of the Uri properties to avoid landing on the failing check.

Example string:
https://dev.azure.com/1edu/270ece61-e250-4844-8782-16f9d77c2be4/_apis/build/Definitions/357?revision=61


src/Sarif/Writers/MimeType.cs
```diff
@@ -27,43 +30,22 @@ public static string DetermineFromFileExtension(string path)
...
-        /// its extension.</returns>
-        public static string DetermineFromFileExtension(Uri fileUri)
-        {
-            if (fileUri == null)
-            {
-                throw new ArgumentNullException(nameof(fileUri));
-            }
-
-            string fileName = fileUri.ToString();
-
```
michaelcfanning, line 58: So, this issue is a well-known weakness in our API, our handling of URIs is very inconsistent across schema, relative vs. absolute, etc. 

The issue with the change that you've proposed is that I'm not sure it is resilient to all kinds of inputs (such as relative paths). I see that we only have one upstream caller to this API in Artifact. Create, which looks like it is over-aggressively generating the mime type. That caller is also a place we could consider a fix (unless we are convinced this API is useful in general, which I am not)

Try this, create a new branch, and for this line I've highlighted:

    string fileName = fileUri.ToString().

See if this line succeeds:

    string fileName = fileUri.GetFileName();

This helper is supposed to elide all differences in URL kinds and retrieve what is effectively the file name.

This helper lives in Sarif/ExtensionMethods.cs. I'm guessing this one line change may fix the issue... But looking forward to whether you discover if this is so.

If you've spent enough time on this break and would prefer we finish the investigation let me know. :) Thanks for your help so far, I appreciate it.

src/Sarif/Writers/MimeType.cs
```diff
@@ -54,7 +54,7 @@ public static string DetermineFromFileExtension(Uri fileUri)
                 throw new ArgumentNullException(nameof(fileUri));
             }
 
-            string fileName = fileUri.ToString();
+            string fileName = fileUri.GetFileName();
```
michaelcfanning, line 57: >GetFileName [](http://example.com/codeflow?start=38&length=11)

You can change this to GetFilePath

kelewis: Yep that's on the way.

michaelcfanning (approved):
:shipit:

PR #2641, opened 2023-03-14
Update HttpMockHelper.cs

HulonJenkins:
Fix what I suppose is a typo. Have BadGateway return BadGatewayResponse.


michaelcfanning (approved):
:shipit:

PR #2642, opened 2023-03-15
Zap `FileRegionsCache.Instance` singleton

michaelcfanning:
In general, our static singletons work well for use in a client tool model, where executions lights up, completes, and exits. This pattern is creating complications for API-driven scenarios where consumers want to take more control of thread & state management.

Removing `FileRegionsCache.Instance` first as its existence clearly permits threading issues (even if we haven't encountered them in the wild yet). We will hunt down and remove all other occurrences of this pattern as follow-on changes.

### DIFF
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -16,7 +16,6 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// </summary>
     public class FileRegionsCache
     {
-        public static readonly FileRegionsCache Instance = new FileRegionsCache();
         public const int DefaultCacheCapacity = 100;
         private readonly IFileSystem _fileSystem;
 
--- a/src/Sarif/FileRegionsCache.cs
+++ b/src/Sarif/FileRegionsCache.cs
@@ -99,7 +99,7 @@ namespace Microsoft.CodeAnalysis.Sarif
             _newLineIndexCache.Clear();
         }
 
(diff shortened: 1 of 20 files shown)


src/Sarif/FileRegionsCache.cs
```diff
@@ -100,7 +99,7 @@ public void ClearCache()
             _newLineIndexCache.Clear();
         }
 
-        private Region PopulateTextRegionProperties(NewLineIndex lineIndex, Region inputRegion, string fileText, bool populateSnippet)
+        private static Region PopulateTextRegionProperties(NewLineIndex lineIndex, Region inputRegion, string fileText, bool populateSnippet)
```
michaelcfanning, line 102: >static [](http://example.com/codeflow?start=16&length=6)

When I touch a file I go ahead and fix the open style issues flagged by VS.

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -104,12 +104,12 @@ public class SarifLogger : BaseLogger, IDisposable, IAnalysisLogger
                 dataToInsert.HasFlag(OptionallyEmittedData.ContextRegionSnippets))
             {
                 _insertOptionalDataVisitor = new InsertOptionalDataVisitor(dataToInsert,
+                                                                           fileRegionsCache,
```
michaelcfanning, line 107: >fileRegionsCache [](http://example.com/codeflow?start=75&length=16)

this is a required argument so had to be moved in the parameters list. This is therefore documented as a breaking change in the SDK.

eddynaka (approved):
:shipit:

PR #2650, opened 2023-03-28
SARIF OM changes for nullable int batch 1

shaopeng-gh:
* NEW: Update `Microsoft.Json.Schema.*` packages from 2.1.0.0 to 2.3.0.0. [#2650](https://github.com/microsoft/sarif-sdk/pull/2650)
* BRK: Update `Artifact.Offset`, `Invocation.ExitCode`, `Invocation.ExitSignalNumber`, `Invocation.ProcessId`, `Result.OccurrenceCount`, `EdgeTraversal.StepOverEdgeCount`, `Notification.ThreadId`, `StackFrame.ThreadId`, `ThreadFlowLocation.NestingLevel`, and `WebResponse.StatusCode` properties from `int` to `int?` type. [#2650](https://github.com/microsoft/sarif-sdk/pull/2650)

The reason of this change is, the upgrade of the `Microsoft.Json.Schema.*` packages exposed an issue that a few int properties should be nullable but it is current not nullable in our C# code.

Due to impact of the 4 line/columns numbers for `Region`, proposed for a separated PR and will need a confirmation for the details before start.


src/Samples/SarifToCsv/SarifCsvColumnWriters.cs
```diff
@@ -106,7 +106,7 @@ public static Action<WriteContext> GetWriter(string columnName)
             writers["Kind"] = (c) => { c.Writer.Write(c.Result.Kind.ToString()); };
             writers["Level"] = (c) => { c.Writer.Write(c.Result.Level.ToString()); };
             writers["Message.Text"] = (c) => { c.Writer.Write(c.Result.Message?.Text ?? ""); };
-            writers["OccurrenceCount"] = (c) => { c.Writer.Write(c.Result.OccurrenceCount); };
+            writers["OccurrenceCount"] = (c) => { c.Writer.Write(c.Result.OccurrenceCount?.ToString() ?? ""); };
```
shaopeng-gh, line 109: >writers["OccurrenceCount"] = (c) => { c.Writer.Write(c.Result.OccurrenceCount?.ToString() ?? ""); }; [](http://example.com/codeflow?start=12&length=100)

note, this file is from sample solution.

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
```diff
@@ -1740,6 +1744,7 @@
         "charLength": {
```
shaopeng-gh, line 1744: >charLength [](http://example.com/codeflow?start=9&length=10)

for `charLength` and `byteLength`, in the spec they indeed have the default value 0.

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
```diff
@@ -1709,24 +1709,28 @@
         "startLine": {
           "description": "The line number of the first character in the region.",
           "type": "integer",
+          "default": 0,
```
shaopeng-gh, line 1712: >"default": 0, [](http://example.com/codeflow?start=10&length=13)

for the 4 number properties, we will do in a separated PR.
This change is temporary just for skip them so they will not be generated as nullable, and the PR is against the dev branch, we will get to it before to main, so this change will not show up in main at all.

src/Sarif/Autogenerated/Region.cs
```diff
@@ -39,24 +39,32 @@ public virtual SarifNodeKind SarifNodeKind
         /// The line number of the first character in the region.
         /// </summary>
         [DataMember(Name = "startLine", IsRequired = false, EmitDefaultValue = false)]
+        [DefaultValue(0)]
```
shaopeng-gh, line 42: >[DefaultValue(0)] [](http://example.com/codeflow?start=8&length=17)

this file is generated from the RTM6 json schema, please see my comment on that file in this PR.

src/Sarif/CodeGenHints.json
```diff
@@ -221,6 +221,16 @@
       }
     }
   ],
+  "ExternalProperties.Schema": [
```
michaelcfanning, line 224: >Schema [](http://example.com/codeflow?start=22&length=6)

how did we miss this?

src/Sarif/CodeGenHints.json
```diff
@@ -1075,6 +1095,26 @@
       }
     }
   ],
+  "TranslationMetadata.DownloadUri": [
```
michaelcfanning, line 1098: >DownloadUri [](http://example.com/codeflow?start=23&length=11)

Good changes, but hm. how'd you find them?

michaelcfanning:
>         [DefaultValue(-1)]

why is this still here? :) when will it go away?

---
Refers to: src/Sarif/Autogenerated/Address.cs:43 in fefed9a. [](commit_id = fefed9a7ff09fd0d9a33d1c94a201b14bbe67f04, deletion_comment = False)

src/Sarif/Autogenerated/ExternalProperties.cs
```diff
@@ -38,6 +38,7 @@ public virtual SarifNodeKind SarifNodeKind
         /// The URI of the JSON schema corresponding to the version of the external property file format.
         /// </summary>
         [DataMember(Name = "schema", IsRequired = false, EmitDefaultValue = false)]
+        [JsonConverter(typeof(Microsoft.CodeAnalysis.Sarif.Readers.UriConverter))]
         public virtual Uri Schema { get; set; }
```
michaelcfanning, line 42: >Uri [](http://example.com/codeflow?start=23&length=3)

There's a 1x1 mapping between this type, URI and the JsonConverter attribute. That could have been the way we handled this in the codegen hints + jschema (i.e., provide a way to emit a specific attribute for every instance of a property with a specific return type). Just an idea, can you confirm this feature does not already exist?

src/Sarif/Autogenerated/Region.cs
michaelcfanning: Right, so we're making the same mistake though! '0' is a default value that effectively means it isn't there, so we should have a nullable type here, again.

This is more painful from an API use standpoint, maybe too painful! But everything would be consistent. 

I do see from the spec that the region class in particular has complex/unusual default value semantics...

michaelcfanning, line 42: >DefaultValue [](http://example.com/codeflow?start=9&length=12)

have you proven this is required? 

the value of 0 is a default for start line, so is the ignore and populate attribute sufficient here? i.e., no need to specify the zero value?

src/Sarif/NotYetAutoGenerated/Address.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// A physical or virtual address, or a range of addresses, in an 'addressable region' (memory or a binary file).
     /// </summary>
     [DataContract]
-    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "2.1.0.0")]
+    [GeneratedCode("Microsoft.Json.Schema.ToDotNet", "2.3.0.0")]
     public partial class Address : PropertyBagHolder, ISarifNode
```
michaelcfanning, line 19: >partial [](http://example.com/codeflow?start=11&length=7)

so do you know for sure not that every file in notyetautogenerated contains other code changes that need to be preserved? or can we delete any of these?

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
michaelcfanning: so what's odd about this is we have a default of -1 for other things, why not this?

how can a default be lower than the minimum value? seems odd.

michaelcfanning:
>           "default": -1,

observe that the minimum value here matches the default value!

---
Refers to: src/Sarif/Schemata/sarif-2.1.0-rtm.6.json:1740 in fefed9a. [](commit_id = fefed9a7ff09fd0d9a33d1c94a201b14bbe67f04, deletion_comment = False)

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
michaelcfanning, line 1712: >default [](http://example.com/codeflow?start=11&length=7)

something is wrong here. if this is the default, then minimum should be set as 0 as well.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
```diff
@@ -225,6 +225,11 @@ private void ProcessLine(string logFileLine, ref int nestingLevel, Result result
                     }
                 }
 
+                if (threadFlowLocation.NestingLevel == 0)
```
michaelcfanning, line 228: >NestingLevel [](http://example.com/codeflow?start=39&length=12)

why is this correct?

how do we distinguish between not present vs. an actual nesting level of 0?

src/Test.FunctionalTests.Sarif/TestData/Multitool/ValidateCommand/Inputs/SARIF1002.UrisMustBeValid_Valid.sarif
```diff
@@ -43,7 +43,7 @@
         },
         {
           "location": {
-            "uri": "file:/c:/src/file.c",
+            "uri": "file:///c:/src/file.c",
```
michaelcfanning, line 46: >file [](http://example.com/codeflow?start=20&length=4)

i see!

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -680,7 +680,7 @@ public void SarifLogger_LogsStartAndEndTimesByDefault()
             // Other properties should be empty.
             invocation.CommandLine.Should().BeNull();
             invocation.WorkingDirectory.Should().BeNull();
-            invocation.ProcessId.Should().Be(0);
+            invocation.ProcessId.Should().Be(null);
```
michaelcfanning, line 683: >Be(null) [](http://example.com/codeflow?start=42&length=8)

Just use .BeNull() here?

src/Sarif/Schemata/sarif-2.1.0-rtm.6.json
```diff
@@ -1709,24 +1709,28 @@
...
           "minimum": 1
```
michaelcfanning, line 1713: >"minimum": 1 [](http://example.com/codeflow?start=10&length=12)

set all minimums to default and then let's please make every int value that has a default in the schema be a nullable item.

of course, we are skipping region for now. I wonder if we should build a special region deserializer! :)

src/Sarif/CodeGenHints.json
shaopeng-gh: not sure how we missed them.
I found them by test not eye.

src/Sarif/CodeGenHints.json
shaopeng-gh: replied in the other comment.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
shaopeng-gh: correct, batch fixed all existing ones.

src/Sarif/Autogenerated/ExternalProperties.cs
shaopeng-gh: The UriConverter class is from SDK not JSchama so we can not directly always generate this attribute when Uri type in JSchema.

We could have support hint that base on type, not just name, then we can have all this declared once. I checked the hint is only based on name.

An other way is we could have still use  base on name but support partial wild card e.g. `*.*.Uri*` then we can have it declared declared once, except this kind of special one `Schema` that does no have Uri in the name. 
But I checked we only support * at the first part only: `*.Uri`.

```CSharp
private static string MakeHintDictionaryKey(string typeName, string propertyName)
        {
            return typeName + "." + propertyName.ToPascalCase();
        }
```

Let me know I can create a work item and work on it.

src/Sarif/NotYetAutoGenerated/Address.cs
shaopeng-gh: Yes I did have compared that every file in the notyetautogenerated folder have difference between the raw auto generated one, when work on these I believe some tests are needed, I think for the separate change.

src/Sarif/Autogenerated/Region.cs
```diff
@@ -116,8 +128,14 @@ public virtual SarifNodeKind SarifNodeKind
         /// </summary>
         public Region()
         {
+            StartLine = 0;
```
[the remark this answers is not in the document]
shaopeng-gh: correct, there are 2 thing here:
1. the line column numbers will be gone after we work on it and regenerates, for this PR we can ignore all changes related to them for now,
2.  CharLength and ByteLength is a real change here. It is autogenerated.

src/Sarif.Converters/StaticDriverVerifierConverter.cs
shaopeng-gh: you are correct, this is code for converting StaticDriverVerifier result to SARIF, the NestingLevel here is converted so if it is 0 then it means 0, not unknown. I have reverted this change, and fixed the test expected outputs to have this property.

src/Sarif/Query/Evaluators/SarifEvaluators.cs
```diff
@@ -29,7 +29,7 @@ public static IExpressionEvaluator<Result> ResultEvaluator(TermExpression term)
                 case "message.text":
                     return new StringEvaluator<Result>(r => r.Message?.Text, term, StringComparison.OrdinalIgnoreCase);
                 case "occurrencecount":
-                    return new LongEvaluator<Result>(r => r.OccurrenceCount, term);
+                    return new LongEvaluator<Result>(r => r.OccurrenceCount ?? 1, term);
```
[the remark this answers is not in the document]
shaopeng-gh: wonder if we should include this in a spec errata ---- I agree with this and will need to talk to you if we consider "the spec is missing this default as 1 ". If so we can keep the OccurrenceCount as not nullable and have default as 1.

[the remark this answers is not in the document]
michaelcfanning: All types with a default must be nullable!! We can discuss in stand-up today. :)

shaopeng-gh:
Thanks, I see your comments about the changes about the 4 line/column numbers.
You are correct that all changes related to them are wrong.
The reason is I want to skip these 4 fields and keep them not nullable now, and because they are auto generated, the way I skip them is to add a default value to them. Can be any integer but why I choose 0 which is less than the minimum 1? Because it is actually the effective default, if not most of the tests will fail. 
Those temporary changes related to them are auto generated which will be auto generated again once we work on these 4 fields.
So we can ignore any changes related to them now.
Also this will go to DEV branch first so will not get merged to main before we get to these 4 fields.
I put the comment where and  will close those ones, to avoid duplicated comments to look.

shaopeng-gh:
>         [DefaultValue(-1)]

Will need to fix this in the future changes, this change set is only those triggered by the new version of JSchema, must act or else will not build.

---
In reply to: [1491078338](https://github.com/microsoft/sarif-sdk/pull/2650#issuecomment-1491078338) [](http://example.com/codeflow?ancestors=1491078338,1491078338)

---
Refers to: src/Sarif/Autogenerated/Address.cs:43 in fefed9a. [](commit_id = fefed9a7ff09fd0d9a33d1c94a201b14bbe67f04, deletion_comment = False)

PR #2654, opened 2023-04-10
Update `--log` help text

yongyan-gh:
To be align with the latest change merging `--inline`, `--force`, `pretty-print` and `minify` into `--log` parameters.

No code change.

### DIFF
--- a/docs/multitool-usage.md
+++ b/docs/multitool-usage.md
@@ -35,7 +35,7 @@ Sarif.Multitool @microsoft/sarif-multitool help
 Sarif.Multitool convert Current.fpr -tool FortifyFpr -output Current.sarif
 
 : Add file contents from analyzed files and snippets from result regions to SARIF
-Sarif.Multitool rewrite Current.sarif --insert "TextFiles,RegionSnippets" --log "Inline"
+Sarif.Multitool rewrite Current.sarif --insert "TextFiles;RegionSnippets" --log "Inline"
 
 : Remove codeFlows from results regions to SARIF
 Sarif.Multitool rewrite Current.sarif --remove "CodeFlows" --log "Inline"
@@ -102,4 +102,4 @@ This will generate an extra report.yml.log, leave in the same folder with the in
 | Inline | Inline outputs to files where appropriate. |
 | PrettyPrint | Produce pretty-printed JSON output rather than compact form. |
 | Minify | Produce compact JSON output (all white space removed) rather than pretty-printed output. |
-| Optimize | Producing a smaller but non-human-readable log omitting redundant properties. |
+| Optimize | Produce a smaller but non-human-readable log omitting redundant properties. |
--- a/docs/multitool-usage.md
+++ b/docs/multitool-usage.md
@@ -35,7 +35,7 @@ Sarif.Multitool @microsoft/sarif-multitool help
 Sarif.Multitool convert Current.fpr -tool FortifyFpr -output Current.sarif
 
 : Add file contents from analyzed files and snippets from result regions to SARIF
-Sarif.Multitool rewrite Current.sarif --insert "TextFiles;RegionSnippets" --log "Inline"
+Sarif.Multitool rewrite Current.sarif --insert TextFiles;RegionSnippets --log "Inline"
 
 : Remove codeFlows from results regions to SARIF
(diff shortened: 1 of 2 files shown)


docs/multitool-usage.md
```diff
@@ -94,10 +94,12 @@ clang-tidy --checks=* --header-filter=.* --system-headers --export-fixes=report.
...
 | Name | Purpose |
 | ---- | ------- |
-| pretty-print | Produce pretty-printed JSON output rather than compact form. |
-| minify | Produce compact JSON output (all white space removed) rather than pretty-printed output. |
-| force | Force overwrite of output file if it exists. |
+| ForceOverwrite | Force overwrite of output file if it exists. |
+| Inline | Inline outputs to files where appropriate. |
+| PrettyPrint | Produce pretty-printed JSON output rather than compact form. |
+| Minify | Produce compact JSON output (all white space removed) rather than pretty-printed output. |
+| Optimize | Producing a smaller but non-human-readable log omitting redundant properties. |
```
michaelcfanning, line 105: >Producing [](http://example.com/codeflow?start=13&length=9)

This s/be 'produce' here and anywhere else you observe this bug.

src/Sarif.Driver/Sdk/CommonOptionsBase.cs
```diff
@@ -28,7 +28,7 @@ public class CommonOptionsBase : PropertiesDictionary
             "log",
             Separator = ';',
             HelpText =
-            "One or more settings for governing output files. Valid values include Force, PrettyPrint, Minify or Optimize.")]
+            "One or more settings for governing output files. Valid values include ForceOverwrite, Inline, PrettyPrint, Minify or Optimize.")]
```
michaelcfanning, line 31: >One or more settings [](http://example.com/codeflow?start=13&length=20)

S/be 'one or more semicolon-delimited settings...

docs/multitool-usage.md
```diff
@@ -35,10 +35,10 @@ Sarif.Multitool @microsoft/sarif-multitool help
 Sarif.Multitool convert Current.fpr -tool FortifyFpr -output Current.sarif
 
 : Add file contents from analyzed files and snippets from result regions to SARIF
-Sarif.Multitool rewrite Current.sarif --insert "TextFiles,RegionSnippets" --inline
+Sarif.Multitool rewrite Current.sarif --insert "TextFiles,RegionSnippets" --log "Inline"
```
michaelcfanning, line 38: >TextFiles,RegionSnippets [](http://example.com/codeflow?start=48&length=24)

This command-line arg is wrong, everything is semicolon delimeted, can you adjust this to 

--insert TextFiles;RegionSnippets

Note that by switching to semicolon as a delimiter, you don't need to enquote args anymore, so switch to this as well:

--log Inline

Finally, if I'm correct here, all this indicates that you didn't actually attempt to execute these command-lines that you are changing. :) For anything you touch, can you do that?

Thanks!

yongyan-gh: Corrected the separator char to ";"

If multiple options used for the parameter value, its doesn't need to be quoted.

If only 1 option is used, still needs to be quoted, otherwise it give the error:
```
ERROR(S):
  Option 'log' is defined with a bad format.
```

michaelcfanning: Looks strange. Maybe related to an idiosyncratic Commandline library behavior? 

Try moving the positional argument `current.sarif` to the end. Does this work? Note the absence of quotation marks.

Sarif.Multitool rewrite --log Inline --insert TextFiles,RegionSnippets Current.sarif

yongyan-gh: Tested the commands I updated both works

yongyan-gh: the same error:

```
C:\repo\github.com\sarif-sdk\bld\bin\AnyCPU_Debug\Sarif.Multitool\netcoreapp3.1>Sarif.Multitool.exe rewrite --insert TextFiles;RegionSnippets --log Inline Current.sarif
SARIF Multitool 4.1.0
c Microsoft Corporation. All rights reserved.

ERROR(S):
  Option 'log' is defined with a bad format.
  A required value not bound to option name is missing.
```

michaelcfanning: I am confused. You say 'both works', you also say 'same error' for both.

PR #2656, opened 2023-04-14
--force message updates. handle 422 code from log post endpoint.

michaelcfanning:
* BUG: Update user messages and code comments that refer to `--force` (replaced by `--log ForceOverwrite`).
* BUG: Handle return code 422 `UnprocessableEntity` when validating that log file POST endpoint is available.


HulonJenkins (approved):
:shipit:

PR #2657, opened 2023-04-18
File regions cache for notifications

michaelcfanning:
BUG: Eliminate `NullReferenceException` in `HashUtilities.ComputeHashesForText` when processing a tool notification in a scan target with no results previously fired.

This bug identified in (and fully tested by) the sarif pattern matcher tool. There is clearly more testing/hardening we can do with the new pipeline model.

@HulonJenkins


HulonJenkins (approved):
:shipit:

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -485,6 +485,10 @@ private static void LogCachingLogger(TContext globalContext, CachingLogger cachi
 
             if (results?.Count > 0)
             {
+                globalContext.Logger.FileRegionsCache = cachingLogger.FileRegionsCache ?? new FileRegionsCache();
```
suvamM, line 488: @michaelcfanning Why are we duplicating this thrice inside the next three if-blocks. Could we not do this once immediately after IEnumerated artifact = globalContext.currentTarget?

PR #2660, opened 2023-04-30
Reauthor runtime errors.

michaelcfanning:
* BRK: `RuntimeConditions` now of type `long` to permit more flag values. Many literal values have changed for individual members. [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)
* BRK: `RuntimeConditions.OneOrMoreFilesSkippedDueToSize` renamed to `OneOrMoreFilesSkippedDueToExceedingSizeLimits`. [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)
* BUG: Properly report skipping empty files (rather than reporting file was skipped due to exceeding size limits). [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)
* NEW: Provide convenience enumerator at the `SarifLog` level that iterates over all results in all runs in the log. [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)
* NEW: Provide `Notes.LogEmptyFileSkipped` helper for reporting zero-byte files skipped at scan time. [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -32,7 +32,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         internal ConsoleLogger _consoleLogger;
 
         private uint _fileContextsCount;
-        private uint _ignoredFilesCount;
+        private long _filesExceedingSizeLimitCount;
         private Channel<uint> _resultsWritingChannel;
         private Channel<uint> readyToScanChannel;
         private ConcurrentDictionary<uint, TContext> _fileContexts;
@@ -386,9 +386,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             enumerateTargets.Wait();
             logResults.Wait();
 
-            if (_ignoredFilesCount > 0)
+            if (_filesExceedingSizeLimitCount > 0)
             {
-                Warnings.LogOneOrMoreFilesSkippedDueToSize(globalContext, _ignoredFilesCount);
+                Warnings.LogOneOrMoreFilesSkippedDueToExceedingSizeLimit(globalContext, _filesExceedingSizeLimitCount);
             }
 
             Console.WriteLine();
@@ -549,15 +549,19 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 readyToScanChannel.Writer.Complete();
             }
 
-            if (context.TargetsProvider.Skipped?.Count > 0)
+            if (context.TargetsProvider.Skipped != null)
             {
                 foreach (IEnumeratedArtifact artifact in context.TargetsProvider.Skipped)
                 {
-                    Notes.LogFileSkippedDueToSize(context, artifact.Uri.GetFilePath(), (long)artifact.SizeInBytes);
-                }
+                    if (artifact.SizeInBytes > 0)
+                    {
+                        _filesExceedingSizeLimitCount++;
+                        Notes.LogFileExceedingSizeLimitSkipped(context, artifact.Uri.GetFilePath(), (long)artifact.SizeInBytes);
+                        continue;
+                    }
 
-                //TBD resolve type mismatch
-                _ignoredFilesCount += (uint)context.TargetsProvider.Skipped.Count;
+                    Notes.LogEmptyFileSkipped(context, artifact.Uri.GetFilePath());
+                }
             }
 
             if (_fileContextsCount == 0)
--- a/src/Sarif/Notes.cs
+++ b/src/Sarif/Notes.cs
@@ -10,9 +10,32 @@ namespace Microsoft.CodeAnalysis.Sarif
     public static class Notes
     {
         public const string Msg001AnalyzingTarget = "MSG001.AnalyzingTarget";
-        public const string Msg002_FileSkippedDueToSize = "MSG002.FileSkippedDueToSize";
+        public const string Msg002_EmptyFileSkipped = "MSG002.EmptyFileSkipped";
+        public const string Msg002_FileExceedingSizeLimitSkipped = "MSG002.FileExceedingSizeLimitSkipped";
 
-        public static void LogFileSkippedDueToSize(IAnalysisContext context, string skippedFile, long fileSizeInKb)
+        public static void LogEmptyFileSkipped(IAnalysisContext context, string skippedFile)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // '{0}' was skipped because it is empty (zero bytes in size).
+            context.Logger.LogConfigurationNotification(
+                Errors.CreateNotification(
+                    new Uri(skippedFile, UriKind.RelativeOrAbsolute),
+                    Msg002_EmptyFileSkipped,
+                    ruleId: null,
+                    FailureLevel.Note,
+                    exception: null,
+                    persistExceptionStack: false,
+                    messageFormat: null,
+                    skippedFile,
+                    context.MaxFileSizeInKilobytes.ToString()));
+
+            context.RuntimeErrors |= RuntimeConditions.OneOrMoreEmptyFilesSkipped;
+        }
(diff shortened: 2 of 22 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -549,15 +549,19 @@ private async Task<bool> EnumerateTargetsAsync(TContext context)
...
+                    if (artifact.SizeInBytes > 0)
+                    {
+                        _filesExceedingSizeLimitCount++;
+                        Notes.LogFileExceedingSizeLimitSkipped(context, artifact.Uri.GetFilePath(), (long)artifact.SizeInBytes);
+                        continue;
+                    }
 
-                //TBD resolve type mismatch
-                _ignoredFilesCount += (uint)context.TargetsProvider.Skipped.Count;
+                    Notes.LogEmptyFileSkipped(context, artifact.Uri.GetFilePath());
```
michaelcfanning, line 563: >LogEmptyFileSkipped [](http://example.com/codeflow?start=26&length=19)

Note that we don't track/report out on skipping 0-byte files, though we do set a non-fatal runtime condition. How interesting is it that we don't scan a 0-byte file?

I suppose we still have a problem, though, there's at least a theoretical case that someone's analysis might be intended to flag all zero-byte files. :) We could allow this lower-bound threshold to be configurable and the core engine code would probably need to be updated.

src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -302,62 +302,69 @@ private static SarifLog GetSarifLogWithMinimalUniqueData()
         }
 
         [Fact]
-        public async Task SarifLog_Post_WithValidParameters_ShouldNotThrownAnExceptionWhenRequestIsValid()
```
michaelcfanning, line 305: >SarifLog_Post_WithValidParameters_ShouldNotThrownAnExceptionWhenRequestIsValid [](http://example.com/codeflow?start=26&length=78)

These tests were poorly factored and as a result made it difficult to locate an underlying test helper bug. It's a good object lesson for why we factor tests the way that we do, it's to lower costs in resolving regressions (in this case introduced by an update to net6.0 from netcoreapp3.1).

src/Test.UnitTests.Sarif/Core/StackTests.cs
```diff
@@ -38,7 +38,7 @@ public void Stack_CreateFromException()
             {
                 File.Create(Path.GetInvalidFileNameChars()[0].ToString(), 0);
             }
-            catch (ArgumentException exception)
+            catch (IOException exception)
```
michaelcfanning, line 41: >IOException [](http://example.com/codeflow?start=19&length=11)

Change in behavior from netcoreapp3.1, File.Create now raises an IOException for illegaly chars (which makes sense).

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -845,13 +845,13 @@ public void SarifLogger_EnhancesRunWithAdditionalAnalysisTargets()
         [Fact]
         public void SarifLogger_AcceptsOverrideOfDefaultEncoding()
         {
-            const string Utf8 = "UTF-8";
-            const string Utf7 = "UTF-7";
```
michaelcfanning, line 849: >UTF [](http://example.com/codeflow?start=33&length=3)

UTF7 is entirely deprecated in net6.0 as an insecure format to support.

src/Test.Utilities.Sarif/HttpMockHelper.cs
```diff
@@ -20,6 +20,11 @@ public static StringContent AnyContent()
             return new StringContent(AnyContentText);
         }
 
+        public static HttpResponseMessage CreateOKResponse()
```
michaelcfanning, line 23: >CreateOKResponse [](http://example.com/codeflow?start=42&length=16)

So, this code illustrates some of the dangers of statics. There is no such thing as a static response, once you read the stream from it, that stream is disposed.

I'm surprised we haven't gotten burned by this before. We need to revisit this test class. I took a small change along the lines of what I think we'll need, helpers that dynamically create new instances of the default data.

Though honestly, a helper that wraps a single line of code may not be worth implementing at all. :)

src/Test.Utilities.Sarif/TestUtilitiesExtensions.cs
```diff
@@ -24,11 +24,9 @@ public static void ValidateCommandExecution(this IAnalysisContext context, int r
 
             // For application exist exceptions, e.g., an unhandled exception in a rule,
             // the inner exception has the most useful details.
-            context.RuntimeExceptions?[0].InnerException.Should().BeNull();
-            context.RuntimeExceptions?[0].Should().BeNull();
-
+            context.RuntimeExceptions?[0].InnerException?.ToString().Should().BeNull();
```
michaelcfanning, line 27: >ToString [](http://example.com/codeflow?start=58&length=8)

These are subtle changes that publish more actionable data to the test logs when failures occur. Now we'll get all of ex.ToString() (which is very rich and useful) rather than a limited report that's constrained to the exception type.

HulonJenkins (approved):
:shipit:

src/Sarif/Notes.cs
```diff
@@ -10,9 +10,32 @@ namespace Microsoft.CodeAnalysis.Sarif
...
+                    ruleId: null,
+                    FailureLevel.Note,
+                    exception: null,
+                    persistExceptionStack: false,
+                    messageFormat: null,
+                    skippedFile,
+                    context.MaxFileSizeInKilobytes.ToString()));
+
+            context.RuntimeErrors |= RuntimeConditions.OneOrMoreEmptyFilesSkipped;
+        }
```
yongyan-gh, line 37: nit: add a empty line after this function

src/Sarif/Warnings.cs
```diff
@@ -11,31 +11,31 @@ public static class Warnings
...
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            // {0} file(s)s were skipped for analysis due to exceeding size limits
+            // {0} file(s)s were skipped for analysis due to exceeding size limit
             // (currently configured as {1} kilobytes). The 'max-file-size-in-kb'
             // command-line argument can be used to increase this threshold.
             context.Logger.LogConfigurationNotification(
```
yongyan-gh, line 35: >LogConfigurationNotification [](http://example.com/codeflow?start=27&length=28)

may move the common code into a function and resuse it

schlaman-ms (approved):
:shipit:

src/Test.FunctionalTests.Sarif/SarifValidatorTests.cs
```diff
@@ -197,7 +197,7 @@ public void ValidatesAllTestFiles()
             Path.Combine("v2", "ConverterTestData"),
             Path.Combine("v2", "SpecExamples"),
             Path.Combine("v2", "ObsoleteFormats"),
-            Path.Combine("..", "..", "Test.UnitTests.Sarif", "netcoreapp3.1", "TestData")
+            Path.Combine("..", "..", "Test.UnitTests.Sarif", "net6.0", "TestData")
```
shaopeng-gh, line 200: >net6.0 [](http://example.com/codeflow?start=62&length=6)

fyi Suvam mentioned to me coyote testing does not support 6.0 yet, see if we want to keep in 3.1

shaopeng-gh (approved):
:shipit:

src/Sarif/RuntimeConditions.cs
```diff
@@ -11,51 +11,98 @@ namespace Microsoft.CodeAnalysis.Sarif
     // code paths. These conditions are a combination of fatal
     // and non-fatal circumstances
     [Flags]
-    public enum RuntimeConditions : uint
+    public enum RuntimeConditions : long
```
shaopeng-gh, line 14: >long [](http://example.com/codeflow?start=36&length=4)

@michaelcfanning we changed this to long, should we change the usage of it to use long as well, 
e.g. I see the run result was int, should we also change to long?
``` CSharp
abstract public int Run(T options);
```

PR #2661, opened 2023-05-01
sarif-sdk: Update text for verbose, recurse, quiet, environment, rich-return-code

shaopeng-gh:
The other part is BinSkim:
https://github.com/microsoft/binskim/pull/880

### DIFF
--- a/docs/multitool-usage.md
+++ b/docs/multitool-usage.md
@@ -56,7 +56,7 @@ Sarif.Multitool export-validation-config validation.xml
 Sarif.Multitool export-validation-rules ValidationRules.md
 
 : Merge multiple SARIF files into one
-Sarif.Multitool merge C:\Input\*.sarif --recurse --output-directory=C:\Output\ --output-file=MergeResult.sarif
+Sarif.Multitool merge C:\Input\*.sarif --recurse true --output-directory=C:\Output\ --output-file=MergeResult.sarif
 
 : Extract new Results only from New Baseline
 Sarif.Multitool query NewBaseline.sarif --expression "BaselineState == 'New'" --output Current.NewResults.sarif
--- a/src/Sarif/SdkResources.resx
+++ b/src/Sarif/SdkResources.resx
@@ -226,10 +226,10 @@
     <value>Unexpected fatal runtime condition(s) observed: </value>
   </data>
   <data name="MSG_OneOrMoreInvalidTargets" xml:space="preserve">
-    <value>One or more targets was skipped entirely as it was determined to be an invalid target for analysis. Include all supported values for both --kind and --level on the command-line for more information.</value>
+    <value>One or more targets was skipped entirely as it was determined to be an invalid target for analysis. Include 'NotApplicable' on an explicit '--kind' command-line argument (e.g., '--kind Fail;NotApplicable') for more information.</value>
   </data>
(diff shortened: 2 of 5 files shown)


src/Sarif/SdkResources.resx
```diff
@@ -226,10 +226,10 @@
     <value>Unexpected fatal runtime condition(s) observed: </value>
   </data>
   <data name="MSG_OneOrMoreInvalidTargets" xml:space="preserve">
-    <value>One or more targets was skipped entirely as it was determined to be an invalid target for analysis. Pass --verbose on the command-line for more information.</value>
+    <value>One or more targets was skipped entirely as it was determined to be an invalid target for analysis. Include all supported values for both --kind and --level on the command-line for more information.</value>
```
michaelcfanning, line 229: >Include all supported values for both --kind and --level [](http://example.com/codeflow?start=111&length=56)

I believe that `--kind Fail;NotApplicable` is the specific value required. You don't need `--level` here as this severity level isn't relevant except for kind == Fail. Can you test this?

Here's a new sentence for you if I'm correct in what I assume:

Include `NotApplicable` on an explicit `--kind` command-line argument (e.g., `--kind Fail;NotApplicable`) for more information.

shaopeng-gh: tested and updated. 
MSG_OneOrMoreInvalidTargets and 
MSG_OneOrMoreNotApplicable is different,
I have tested the specific value required.

shaopeng-gh: oh btw I did not use the markup backtick, we don't use markup in the console output currently and i think that is correct.

suvamM (approved):
:shipit:

shaopeng-gh:
@michaelcfanning @suvamM 
Please review with this together to get the full change I am making:
https://github.com/microsoft/sarif-sdk/pull/2661

docs/multitool-usage.md
```diff
@@ -56,7 +56,7 @@ Sarif.Multitool export-validation-config validation.xml
 Sarif.Multitool export-validation-rules ValidationRules.md
 
 : Merge multiple SARIF files into one
-Sarif.Multitool merge C:\Input\*.sarif --recurse --output-directory=C:\Output\ --output-file=MergeResult.sarif
+Sarif.Multitool merge C:\Input\*.sarif --recurse true --output-directory=C:\Output\ --output-file=MergeResult.sarif
```
michaelcfanning, line 59: >true [](http://example.com/codeflow?start=49&length=4)

This shouldn't be literally required, you *can* provide an explicit true but you can also exclude it. Do you not observe that this is correct in your testing?

src/Test.UnitTests.Sarif/HashUtilitiesTests.cs
```diff
@@ -72,7 +72,7 @@ public void HashUtilities_ShouldRespectQuietFlag()
                 //  Do nothing, this is the desired behavior.
             }
 
-            //  Mimic the --quiet flag.  We should not use the console this time.
+            //  Mimic the '--quiet true'.  We should not use the console this time.
```
michaelcfanning, line 75: >quiet [](http://example.com/codeflow?start=29&length=5)

Again all of these values s/b legal:

--quiet (s/be interpreted as --quiet true)
-- quiet true
--quiet false

michaelcfanning (approved):
:shipit:

PR #2663, opened 2023-05-02
File Size enumeration exception

suvamM:
This PR makes the following fixes:

1. `EnumeratedArtifact` class throws an exception while analyzing certain Linux files. An example file path which reproduces the error is `D:/test/file`. The issue arises in the `get_SizeInBytes` method:
```
this.sizeInBytes = Uri!.IsFile
                    ? (ulong)FileSystem.FileInfoLength(Uri.LocalPath)
                    : (ulong?)null;
```
The example file path above is treated as a relative URI, which causes exceptions in `Uri.IsFile` and `Uri.LocalPath`. The change in this PR checks if the path associated with `Uri.GetFilePath()` has the attributes of a file or directory (which is expected work across Windows and Linux). I tested this change by doing an analysis in Linux, and no exceptions are thrown.

2. The post-build task of Coyote binary instrumentation does not work in Linux, since it is a powershell script. This causes a build failure in Linux. This PR adds a condition in the post-build task to perform the rewriting only in Windows.

### DIFF
--- a/src/Sarif/EnumeratedArtifact.cs
+++ b/src/Sarif/EnumeratedArtifact.cs
@@ -62,9 +62,14 @@ namespace Microsoft.CodeAnalysis.Sarif
             {
                 if (sizeInBytes != null) { return sizeInBytes.Value; };
 
-                this.sizeInBytes = Uri!.IsFile
-                    ? (ulong)FileSystem.FileInfoLength(Uri.LocalPath)
-                    : (ulong?)null;
+                if (!File.GetAttributes(Uri.GetFilePath()).HasFlag(FileAttributes.Directory))
+                {
+                    this.sizeInBytes = (ulong)FileSystem.FileInfoLength(Uri.GetFilePath());
+                }
+                else
+                { 
+                    this.sizeInBytes = (ulong?)null; 
+                }
 
                 return this.sizeInBytes;
             }
--- a/src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
+++ b/src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
@@ -70,7 +70,7 @@
     </EmbeddedResource>
   </ItemGroup>
 
-  <Target Name="CoyoteRewriting" AfterTargets="AfterBuild">
+  <Target Name="CoyoteRewriting" AfterTargets="AfterBuild" Condition=" '$(OS)' == 'Windows_NT' ">
     <Exec Command="$(ExecCommand)" />
   </Target>
 </Project>


schlaman-ms:
> 

What about this location for same issue?

---
Refers to: src/Sarif/EnumeratedArtifact.cs:43 in 1909293. [](commit_id = 1909293f63b23109b6947ecce63dbd866011cd98, deletion_comment = False)

schlaman-ms:
> 

In another method, FormatForVisualStudio(), there is this:

 string path = uri.IsAbsoluteUri && uri.IsFile ? uri.LocalPath : uri.ToString();

---
Refers to: src/Sarif/EnumeratedArtifact.cs:43 in 1909293. [](commit_id = 1909293f63b23109b6947ecce63dbd866011cd98, deletion_comment = False)

suvamM:
> > 
> 
> In another method, FormatForVisualStudio(), there is this:
> 
> string path = uri.IsAbsoluteUri && uri.IsFile ? uri.LocalPath : uri.ToString();
> 
> Actually in all the other locations where .IsFile is called it has the above pattern for uri.IsAbsoluteUri && uri.IsFile
> 
> Isn't this one an issue as well?
> 
> Refers to: src/Sarif/EnumeratedArtifact.cs:43 in [1909293](https://github.com/microsoft/sarif-sdk/commit/1909293f63b23109b6947ecce63dbd866011cd98). [](commit_id = [1909293](https://github.com/microsoft/sarif-sdk/commit/1909293f63b23109b6947ecce63dbd866011cd98), deletion_comment = False)

@schlaman-ms Do we have an example where this causes an issue?

schlaman-ms (approved):
:shipit:

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -62,9 +62,14 @@
             {
                 if (sizeInBytes != null) { return sizeInBytes.Value; };
 
-                this.sizeInBytes = Uri!.IsFile
-                    ? (ulong)FileSystem.FileInfoLength(Uri.LocalPath)
-                    : (ulong?)null;
+                if (!File.GetAttributes(Uri.GetFilePath()).HasFlag(FileAttributes.Directory))
```
shaopeng-gh, line 65: >!File.GetAttributes(Uri.GetFilePath()).HasFlag(FileAttributes.Directory) [](http://example.com/codeflow?start=20&length=72)

thanks for explaining the reason in the description.
The fix itself doesn't look straight forward, I will need to get the branch and try it.
I think, at least if the result of the Uri.GetFilePath() not a valid file path then File.GetAttributes should throw unhandled exception here.

michaelcfanning: Right none of this code is hardened for scenarios such as in-memory analysis, scanning from zip archives, etc. These won't be file scenarios.

But I'm not sure how to handle it.

Suvam, I see that I was very off-base in our conversation, the root issue appears to be someone attempting to retrieve a file size from a directory, not an attempt to retrieve file size from a relative path causing an exception.

michaelcfanning (approved):
:shipit:

michaelcfanning (changes requested):
:clock1:

PR #2664, opened 2023-05-04
Fix error in Linux `ERR999.UnhandledEngineException` : `System.InvalidOperationException`: This operation is not supported for a relative URI.

shaopeng-gh:
# Test: 
have added the test in BinSkim after this is merged in and sub-module update,
the failing test that repro the issue (https://github.com/microsoft/binskim/pull/884) will pass see below:

image

image

image

# Reason: 
Below code will output false in Linux, means not absolute Uri.
But I think it is wrong, I don't know if it is .NET bug or not. The url should be a absolute  uri.

``` CSharp
string path = "/home/__w/1/s/src/build/_deps/cxx-src/WORKSPACE";
Uri uri = new Uri(path, UriKind.RelativeOrAbsolute);
Console.WriteLine("uri.IsAbsoluteUri:" + uri.IsAbsoluteUri);
```

The place fix is made, if you look at the code above, it is always a Absolute uri before reaching the line. So we don't leave .NET to do the guessing, we just create a Absolute uri.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -14,7 +14,7 @@
 * BRK: Remove unused `quiet` parameter from `SarifLogger`. [#2639]https://github.com/microsoft/sarif-sdk/pull/2639
 * BRK: Remove `ComputeHashData` and `AnalysisTargetToHashDataMap` properties from `SarifLogger` (in preference of new `fileRegionsCache` parameter. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
 * BRK: Eliminate proactive hashing of artifacts in `SarifLogger` constructor when `OptionallyEmittedData.Hashes` is specified. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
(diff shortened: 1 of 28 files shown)


ReleaseHistory.md
```diff
@@ -14,6 +14,7 @@
 * BRK: Remove unused `quiet` parameter from `SarifLogger`. [#2639]https://github.com/microsoft/sarif-sdk/pull/2639
 * BRK: Remove `ComputeHashData` and `AnalysisTargetToHashDataMap` properties from `SarifLogger` (in preference of new `fileRegionsCache` parameter. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
 * BRK: Eliminate proactive hashing of artifacts in `SarifLogger` constructor when `OptionallyEmittedData.Hashes` is specified. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
+* BUG: Fixed `ERR999.UnhandledEngineException`: `System.InvalidOperationException`: "This operation is not supported for a relative URI" when running in Linux with files skipped due to zero byte size. [#2664](https://github.com/microsoft/sarif-sdk/pull/2664)
```
shaopeng-gh, line 17: >BUG [](http://example.com/codeflow?start=2&length=3)

please view the PR description in GitHub so it has the screenshots.

suvamM: Put the whole thing from ERR999 to relative URI inside backticks,

suvamM (approved):

PR #2667, opened 2023-05-12
Fix Hash issue in 4.1.0

shaopeng-gh:
Fix Hash issue in 4.1.0, if point BinSkim to this version, the exact changes for hashes in https://github.com/microsoft/binskim/pull/853 will be reverted to the original hash, which is expected:
https://github.com/microsoft/binskim/pull/891

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -14,6 +14,7 @@
 * BRK: Remove unused `quiet` parameter from `SarifLogger`. [#2639]https://github.com/microsoft/sarif-sdk/pull/2639
 * BRK: Remove `ComputeHashData` and `AnalysisTargetToHashDataMap` properties from `SarifLogger` (in preference of new `fileRegionsCache` parameter. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
 * BRK: Eliminate proactive hashing of artifacts in `SarifLogger` constructor when `OptionallyEmittedData.Hashes` is specified. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
+* BUG: Fixed incorrect hashes being returned when `--insert hashes` is enabled. [#2667](https://github.com/microsoft/sarif-sdk/pull/2667)
 * BUG: Fixed `ERR999.UnhandledEngineException: System.InvalidOperationException: This operation is not supported for a relative URI` when running in Linux with files skipped due to zero byte size. [#2664](https://github.com/microsoft/sarif-sdk/pull/2664)
 * BUG: Properly report skipping empty files (rather than reporting file was skipped due to exceeding size limits). [#2660](https://github.com/microsoft/sarif-sdk/pull/2660)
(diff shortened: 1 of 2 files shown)


ReleaseHistory.md
```diff
@@ -14,6 +14,7 @@
 * BRK: Remove unused `quiet` parameter from `SarifLogger`. [#2639]https://github.com/microsoft/sarif-sdk/pull/2639
 * BRK: Remove `ComputeHashData` and `AnalysisTargetToHashDataMap` properties from `SarifLogger` (in preference of new `fileRegionsCache` parameter. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
 * BRK: Eliminate proactive hashing of artifacts in `SarifLogger` constructor when `OptionallyEmittedData.Hashes` is specified. [#2639](https://github.com/microsoft/sarif-sdk/pull/2639)
+* BUG: Fixed incorrect hashes being returned when `--insert hashes` is enabled. [#2667](https://github.com/microsoft/sarif-sdk/pull/2667)
```
michaelcfanning, line 17: >incorrect [](http://example.com/codeflow?start=13&length=9)

No unit test? What actually is the problem here, the release note doesn't tell us the symptom.

shaopeng-gh: the symptom --- for example the https://github.com/microsoft/binskim/pull/853 changed the hash from 
```
"md5": "2400EE99A6BB7B792A67A03A59385018",
            "sha-1": "525CB4CBA8B8F80539BBC86C2D78A4275279D8F6",
            "sha-256": "41BF336F8AB7D9A6246D37A6EC7210D4BE07893836890BAB0BFEDA98ACEDE87D"
```
to:

```
            "md5": "C60E01E6A2660A3359EACF5D00799FA0",
            "sha-1": "3E0D3A9216B0AA9D248A89C1C4398215A415B638",
            "sha-256": "BA60EF8C65F0E9006620D357113E40205B201DFF4935C55D9101FDF0702A9E29"
```
which is incorrect, 

No unit test ---- I only tested the hashes will be back with original values in https://github.com/microsoft/binskim/pull/891, good point, let me work on add tests.

michaelcfanning: Yes, but why is it incorrect? We were hashing the wrong data clearly, what's the difference between the broken and fixed version?

Please don't overinvest in tests. Write a targeted regression test that captures the bug. I assume it is due to entering the file regions cache with different code paths (one of which hashed the empty string maybe?)

shaopeng-gh: why is it incorrect? We were hashing the wrong data clearly, what's the difference between the broken and fixed version?
          ----- the code we added in the new pipeline, we used the cached `_fileTextCache` which is reading file as text.
reading a binary file as **text** will not retain the correct **all** bytes, with that, the hash will be different. That is why in the new pipeline change you noticed hash changed.

the symptom 
          ----- above is the cause, the symptom is what user will see, they just see the hash we provided is wrong, if they verify. If they have some kind of baseline mechanism, they will see the hash changed to a wrong value.

PR #2677, opened 2023-05-25
Updated events model.

michaelcfanning:


src/Sarif/AnalyzeContextBase.cs
```diff
@@ -221,6 +225,24 @@ public virtual void Dispose()
             var disposableLogger = this.Logger as IDisposable;
             disposableLogger?.Dispose();
             this.Logger = null;
+
+            if (TraceEventSession != null)
+            {
+                TraceEventSession?.Flush();
```
suvamM, line 231: >TraceEventSession? [](http://example.com/codeflow?start=16&length=18)

Is the null check required here?

src/Sarif/AnalyzeContextBase.cs
```diff
@@ -221,6 +225,24 @@ public virtual void Dispose()
...
+
+                if (TraceEventSession?.EventsLost > 0)
```
suvamM, line 233: >TraceEventSession? [](http://example.com/codeflow?start=20&length=18)

Same point here: the null check is redundant

src/Sarif/AnalyzeContextBase.cs
```diff
@@ -75,6 +77,8 @@ public virtual IEnumerable<IOption> GetOptions()
 
         public Regex CompiledGlobalFileDenyRegex { get; set; }
 
+        public TraceEventSession TraceEventSession { get; set; }
```
suvamM, line 80: >TraceEventSession [](http://example.com/codeflow?start=15&length=17)

Is this a standard class or something that we are adding? I think the session is called "Event Trace Session (ets)", so if this is a class we are creating, maybe worth renaming to align with the ets naming.

src/Sarif.Driver/Sdk/DriverEventNames.cs
```diff
@@ -9,10 +9,20 @@ public sealed class DriverEventNames
...
+        public const string EnumerateArtifactsStart = $"{EnumerateArtifacts}/Start";
+
         public const string ReadArtifact = nameof(ReadArtifact);
         public const string ReadArtifactStop = $"{ReadArtifact}/Stop";
         public const string ReadArtifactStart = $"{ReadArtifact}/Start";
 
+        public const string Scanned = nameof(Scanned);
+
+        public const string ScanArtifact = nameof(ScanArtifact);
+        public const string ScanArtifactStop = $"{ScanArtifact}/Stop";
```
suvamM, line 23: >Stop [](http://example.com/codeflow?start=64&length=4)

Any value in doing a public const string StartEvent = nameof(Start), and a similar one for Stop?

src/Sarif/AnalyzeContextBase.cs
michaelcfanning: Not our class! I agree it's odd how the word order switches in various contexts. :)

src/Sarif/AnalyzeContextBase.cs
michaelcfanning: Yes, thanks, will follow up on this in next change.

src/Sarif.Driver/Sdk/DriverEventNames.cs
michaelcfanning: Sure, I think so, simply because it's repeated everywhere. These terms aren't actively referenced separately but using a constant will protect against typos. Next change! Thanks for the note.

src/Sarif/AnalyzeContextBase.cs
michaelcfanning: Had to reformat code anyway, so worked in your notes, thanks for the feedback.

PR #2684, opened 2023-06-14
Fix GetBlame: Fixes #2683

AvremelM:
Fix for inserting git blame information. 

Fixes #2683

PR #2686, opened 2023-06-19
Exception retrieving zip archive length

michaelcfanning:
+many other corollary changes. This issue is .net 48 only and so we now invoke additional testing for that platform. This resulted in provoking some of the differences between net4.8 and later frameworks.


src/build.props
```diff
@@ -10,8 +10,8 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF SDK</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix>4.2.0</VersionPrefix>
-    <PreviousVersionPrefix>4.1.0</PreviousVersionPrefix>
+    <VersionPrefix>4.2.1</VersionPrefix>
```
michaelcfanning, line 13: >1 [](http://example.com/codeflow?start=23&length=1)

If we only have a behavioral bug fix, then this versioning change is good. Will keep my eyes open for a more serious change that might argue for upgrade to 4.3

src/.editorconfig
```diff
@@ -1,6 +1,15 @@
 # Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
+# Ignore paths
```
michaelcfanning, line 4: >Ignore [](http://example.com/codeflow?start=2&length=6)

FYI, I have pushed an upstream version of .editorconfig to SARIF SDK and will delete others. I have tinkered with this file a bit to enforce some standards we can tighten up. We continue to discuss the standard as a team and will continue to adjust.

src/Sarif/GitHelper.cs
```diff
@@ -87,7 +87,7 @@ public string GetCurrentCommit(string repoPath)
         public string GetBlame(string filePath)
         {
             return GetSimpleGitCommandOutput(
-                Path.GetDirectoryName(filePath),
+                GetTopLevel(filePath),
```
michaelcfanning, line 90: >GetTopLevel [](http://example.com/codeflow?start=16&length=11)

@AvremelM, here is your fix, thanks for providing it.

src/Test.UnitTests.Sarif/ArtifactProviderTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using Xunit;
+
+namespace Test.UnitTests.Sarif
+{
+    public class ArtifactProviderTests
+    {
+
+        [Fact]
+        public void Test1()
```
HulonJenkins, line 20: >Test1 [](http://example.com/codeflow?start=20&length=5)

A more descriptive test name would be nice here

src/Sarif/GitHelper.cs
```diff
@@ -87,7 +87,7 @@ public string GetCurrentCommit(string repoPath)
...
                 args: $"blame -f --porcelain {Path.GetFileName(filePath)}",
```
AvremelM, line 91: This line is also required for this change to work, because `filePath` includes the intervening nested directory path between the repo root and the file.
```suggestion
                args: $"blame -f --porcelain {filePath}",
```

src/Test.UnitTests.Sarif/GitHelperTests.cs
```diff
@@ -178,7 +178,7 @@ public void GitExePath_WhenPathDoesNotExist_SettingManuallyShouldWork()
         [Fact]
         public void GetTopLevel_WhenRepoPathIsToAFile()
         {
-            string pathToFile = typeof(GitHelperTests).Assembly.Location;
+            string pathToFile = Environment.CurrentDirectory;
```
michaelcfanning, line 181: >Environment [](http://example.com/codeflow?start=32&length=11)

When tests execute under .net the product/test binaries are copied to a location outside of the repo.

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.CodeAnalysis.Test.UnitTests.Sarif.Core
 {
     public class LocationTests
     {
-        private const string Id = "\"Id\"";
+        private const string Id = "\"id\"";
```
michaelcfanning, line 19: >id [](http://example.com/codeflow?start=37&length=2)

There was no clear reason to make this property name pascal-cased.

src/Test.UnitTests.Sarif/Core/LocationTests.cs
```diff
@@ -226,14 +226,14 @@ private void VerifyIdRoundTripFromObjectHelper(Location location, bool shouldSer
         {
             location.ShouldSerializeId().Should().Be(shouldSerialize);
             string jsonLocation = JsonConvert.SerializeObject(location);
-            jsonLocation.Contains(Id, StringComparison.InvariantCultureIgnoreCase).Should().Be(shouldSerialize);
+            jsonLocation.Contains(Id).Should().Be(shouldSerialize);
```
michaelcfanning, line 229: >Contains [](http://example.com/codeflow?start=25&length=8)

.NET 4.8 does not provide a Contains overload that accepts a string comparison arg. We don't need case-insensitivity here and elsewhere in any case.

src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
```diff
@@ -8,7 +8,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net48;net6.0</TargetFrameworks>
```
michaelcfanning, line 11: >net48 [](http://example.com/codeflow?start=22&length=5)

A key change. We need to run more tests under .NET 4.8 because this framework is a primary production use case for us.

src/Sarif/GitHelper.cs
AvremelM: @michaelcfanning  Thank you! But this is only half of it, please see: 
https://github.com/microsoft/sarif-sdk/pull/2684#issuecomment-1597431868
as well as my Review suggestion on this PR

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
```diff
@@ -48,7 +48,7 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
             // to the expected output.
             string enlistmentRoot = GitHelper.Default.GetRepositoryRoot(Environment.CurrentDirectory, useCache: false);
 
-            if (inputResourceName == "CoreTests-Relative.sarif")
+            if (inputResourceName == "Rel.sarif")
```
HulonJenkins, line 51: >Rel [](http://example.com/codeflow?start=38&length=3)

verbose "Relative" here is easier to understand imo. Why was it shortened?

src/Test.UnitTests.Sarif/Core/StackTests.cs
```diff
@@ -38,8 +38,9 @@ public void Stack_CreateFromException()
             {
                 File.Create(Path.GetInvalidFileNameChars()[0].ToString(), 0);
             }
-            catch (IOException exception)
+            catch (Exception exception)
```
LingZhou-gh, line 41: Exception to our Exception handling guidelines ? ([https://dev.azure.com/mseng/1ES/_wiki/wikis/Security%20Tools%20Detections%20Team/34629/Team-coding-standards-guidelines](https://dev.azure.com/mseng/1ES/_wiki/wikis/Security%20Tools%20Detections%20Team/34629/Team-coding-standards-guidelines)) 

There is one article about catch multiple exceptions:
[https://stackoverflow.com/questions/136035/catch-multiple-exceptions-at-once](https://stackoverflow.com/questions/136035/catch-multiple-exceptions-at-once)

AvremelM (changes requested):

src/Test.UnitTests.Sarif/GitHelperTests.cs
```diff
@@ -187,7 +187,7 @@ public void GetTopLevel_WhenRepoPathIsToAFile()
         [Fact]
         public void GetTopLevel_WhenRepoPathIsToADirectory()
         {
-            string pathToDirectory = Path.GetDirectoryName(typeof(GitHelperTests).Assembly.Location);
+            string pathToDirectory = Environment.CurrentDirectory;
```
LingZhou-gh, line 190: >Environment.CurrentDirectory; [](http://example.com/codeflow?start=37&length=29)

pathToFile (line 181) and pathToDirecotory (line 190) are same after this change. Is it right?

src/Test.UnitTests.Sarif/GitHelperTests.cs
yongyan-gh: should the `pathToFile` point to a file?

src/Test.UnitTests.Sarif/Core/StackTests.cs
```diff
@@ -82,8 +84,9 @@ public void Stack_CreateFromAggregatedExceptionWithInnerException()
...
```
yongyan-gh, line 87: >Exception exception [](http://example.com/codeflow?start=19&length=19)

nit: may use this syntax if don't want to duplicate the exception handling code:

```
catch (Exception ex) when(ex is ArgumentException || ex is IOException)
{
     ...
}
```

ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.2.1** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/4.2.1) | [Driver](https://www.nuget.org/packages/Sarif.Driver/4.2.1) | [Converters](https://www.nuget.org/packages/Sarif.Converters/4.2.1) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/4.2.1) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/4.2.0)
```
shaopeng-gh, line 2: >4.2.0 [](http://example.com/codeflow?start=345&length=5)

4.2.1

src/Test.UnitTests.Sarif/Core/StackTests.cs
michaelcfanning: Well, this is a weird one. This test is designed to raise an exception and it doesn't really matter which one. It also isn't a concern that we swallow all exceptions because we're not in product code, we're testing the literal exceptional path.

So it is the 'exception that proves the rule.' Haha. Still, please note that you review feedback to add an exception filter that's constrained to the IOException and ArgumentException *does* meet our strict exception standard. A different exception type will fail the test... So thanks for the note!

src/Test.UnitTests.Sarif/Core/StackTests.cs
michaelcfanning: Yes! Nice and this resolves Ling's concern around the test code having a general Exception handler. There's benefit in *never* allowing that because that underscores to devs that it is never desirable in the product itself.

ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
...
+* BUG: Resolve `NotSupportedException` thrown on accessing `DeflateStream.Length` from `MultithreadedZipArchiveArtifactProvider.SizeInBytes` property.
```
shaopeng-gh, line 3: >thrown [](http://example.com/codeflow?start=39&length=6)

should we add the info above the bug ".net 4.8 and earlier", somewhere in the line

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
michaelcfanning: I mentioned this elsewhere will add the note here too. thanks!

ReleaseHistory.md
michaelcfanning: yes! great addition, nice job.

src/Sarif/GitHelper.cs
michaelcfanning: ok, added a simple unit test. we need to go back and do more exhaustive testing soon.

schlaman-ms (approved):
:shipit:

src/Test.UnitTests.Sarif/ArtifactProviderTests.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public class ArtifactProviderTests
+    {
+
+        [Fact]
+        public void MultithreadedZipArchiveArtifactProvider_RetrieveSizeInBytes()
+        {
+            string entryContents = "test";
+            ZipArchive zip = CreateZipArchive("test.zip", entryContents);
+            var artifactProvider = new MultithreadedZipArchiveArtifactProvider(zip, FileSystem.Instance);
```
suvamM, line 24: Curious why we would use `var` here, and the explicit `ZipArchive` type in the previous line.

yongyan-gh (approved):
:shipit:

suvamM (approved):

shaopeng-gh:
Build and test failed, please check, I see the new tests and some old tests are failing.

src/Test.UnitTests.Sarif/GitHelperTests.cs
shaopeng-gh: >Environment.CurrentDirectory [](http://example.com/codeflow?start=32&length=28)

will `Environment.CurrentDirectory` be the same as Assembly location? I think that depends on how it is executed, if the test run local and on server both work then I think it is good.

src/Test.UnitTests.Sarif/Visitors/InsertOptionalDataVisitorTests.cs
shaopeng-gh: for windows, default is something like
`C:\Users\%USERNAME%\AppData\Roaming`
if it related to the length of username as well, then a longer alias person will be easier to have problem.

PR #2687, opened 2023-06-19
Disable CodeQL analysis.

michaelcfanning:
This workflow crashes on every PR. And we're training engineers to just ignore those failures. That's not good.

I am removing the workflow for now. The team has not built the maturity yet of acting on these analysis results.

Some right-thinking engineer with a high standard of excellence will surely re-enable this in the future.


suvamM (approved):

HulonJenkins (approved):
:shipit:

shaopeng-gh (approved):
:shipit:

.github/workflows/codeql-analysis.yml
```diff
@@ -1,48 +0,0 @@
...
-on:
-  push:
-    # The branches below must be a subset of the branches above
-    branches: "**"
-  pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [ main ]
-  schedule:
-    - cron: '0 0 * * 0'
-  workflow_dispatch:
```
shaopeng-gh, line 23: >workflow_dispatch: [](http://example.com/codeflow?start=2&length=18)

suggestion: instead of remove, how about change it to on demand only? That way we can test if need.

michaelcfanning: the reason is that I have no motivation to spend more time on this topic than I have. if you have a commit to propose you think will do what you suggest, contribute it here. otherwise this is going away because i have lots of other stuff competing for attention. :) i suspect you can empathise.

shaopeng-gh: yes! I know `workflow_dispatch` is the on demand one so I quickly changed it. I think ready to merge now.

Issue #2695, opened 2023-07-21
HDF->SARIF converter output doesn't map status to kind/level/rank correctly


candrews:
The SARIF generated by this tool from HDF does not have HDF's `status` mapped to SARIF's [`kind`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317647) and [`level`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317648) correctly.

HDF's `status` can have the following values and these are the `kind`/`level` values I believe they should map to:
* `passed` should map to `pass` and `level` either unset or `none`
* `failed` should map to `fail`  and `level` either `warning`, `error`, or `note` which is already implemented by [`SarifLevelFromHdfImpact`](https://github.com/microsoft/sarif-sdk/blob/274ed7e9f32b48229ca214c31c3b27c9576642b6/src/Sarif.Converters/HdfConverter.cs#L139)
* `error` should map to `review` and `level` either unset or `none`
* `skipped` should map to `notApplicable` and `level` either unset or `none`

Also, if `kind` is set to `fail`, then [`rank`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317663) can be set. If `kind` is any value other than `fail`, then `rank` must not be set.

To reproduce:
1. Take this sample HDF: [openscap-report.hdf.json.gz](https://github.com/microsoft/sarif-sdk/files/12134335/openscap-report.hdf.json.gz)
2. Use this project to convert it to SARIF: `npx @microsoft/sarif-multitool@4.2.2 convert -t Hdf -o "openscap-report.sarif" "openscap-report.hdf.json"` Here's the resulting sarif (I pretty formatted it for readability): 
[openscap-report.sarif.gz](https://github.com/microsoft/sarif-sdk/files/12134347/openscap-report.sarif.gz)
3. Open `openscap-report.sarif`

Expected:
A variety of `kind` and `level` values should be present.
The SARIF is basically useless at the moment as every result shows as a finding, whether it passed or not.
For some example:
* `xccdf_org.ssgproject.content_rule_accounts_no_uid_except_zero` has an HDF `status` of `passed` but in the SARIF has a `level` of `error` (`level` should be `none` or absent, and `kind` should be set to `passed`)
[...]

Issue #2694, opened 2023-07-21
HDF->SARIF converters missing location, making invalid for use by GitHub


candrews:
The SARIF generated by this tool from HDF does not pass GitHub validation.

To reproduce:
1. Take this sample HDF: [openscap-report.hdf.json.gz](https://github.com/microsoft/sarif-sdk/files/12134335/openscap-report.hdf.json.gz)
2. Use this project to convert it to SARIF: `npx @microsoft/sarif-multitool@4.2.2 convert -t Hdf -o "openscap-report.sarif" "openscap-report.hdf.json"` Here's the resulting sarif (I pretty formatted it for readability): 
[openscap-report.sarif.gz](https://github.com/microsoft/sarif-sdk/files/12134347/openscap-report.sarif.gz)
3. Use the validator at https://sarifweb.azurewebsites.net/Validation to validate the `openscap-report.sarif` file with the "GitHub ingestion rules" option selected.

Expected:
Successful validation.

Actual:
[GH1001](http://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html): Each result location must provide the property 'physicalLocation.artifactLocation.uri'. GitHub Advanced Security code scanning will not display a result whose location does not provide the URI of the artifact that contains the result.

When trying to use the SARIF in practice with GitHub's `github/codeql-action/upload-sarif@v2` action as documented at https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#example-workflow-for-sarif-files-generated-outside-of-a-repository GitHub fails to process the SARIF as expected due to the missing required location data. For example failing GitHub action, see https://github.com/candrews/jumpstart/actions/runs/5603707977/job/15181171825?pr=884#step:10:22 the failure is:
```
Error: Code Scanning could not process the submitted SARIF file:
[...]

PR #2696, opened 2023-07-24
HdfConverter: map status to SARIF kind/level/rank correctly (#2695)

candrews:
* rank should only be output when kind=fail
* level should only be output when kind=fail
* handle kind values of passed, faild, error, and skipped


michaelcfanning:
Merged. I'll drop in a release note in a subsequent change. Do you need a NuGet release?

candrews:
> Do you need a NuGet release?

I'm hoping to get an MR in for https://github.com/microsoft/sarif-sdk/issues/2694 very soon (:crossed_fingers: today), after that, a release would be amazing :)

PR #2705, opened 2023-07-28
HdfConverter: Set security-severity property used by GitHub

candrews:
See: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#reportingdescriptor-object

### DIFF
--- a/src/Sarif.Converters/HdfConverter.cs
+++ b/src/Sarif.Converters/HdfConverter.cs
@@ -134,6 +134,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Converters
                         Kinds = new List<string>() { "relevant" },
                     }))
             };
+            reportingDescriptor.SetProperty("security-severity", SarifSecuritySeverityFromHdfImpact(execJsonControl.Impact).ToString());
 
             var results = new List<Result>(execJsonControl.Results.Count);
             foreach (ControlResult controlResult in execJsonControl.Results)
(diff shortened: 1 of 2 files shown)


michaelcfanning:
We can take this change if you're confident GitHub consumes this data. I *thought* that security severity was only relevant for the rules metadata. Someone else, actually, had raised the issue with us (and GitHub) that security-severity can't currently be overridden at the result level. Take a look at the content below, see how it requires adding this data to the `reportingDescriptorReference`? 

https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#runautomationdetails-object

michaelcfanning:
>               "id": "7",

This is the level where you need to add security-severity, I think. If security-severity is a dynamically computed per-result value that can change within a rule, then we need to work with GitHub to get that functionality going for you (it's already been requested by someone else as well).

---
Refers to: src/Test.UnitTests.Sarif.Converters/TestData/HdfConverter/ExpectedOutputs/ValidResults.sarif:4949 in e1ca167. [](commit_id = e1ca167d3b10e4be12d6c31ca6e4c1877486ffc9, deletion_comment = False)

candrews:
> I _thought_ that security severity was only relevant for the rules metadata.

You're right...

> This is the level where you need to add security-severity, I think.

Agreed.

I've updated this PR accordingly.

michaelcfanning (approved):
:shipit:

PR #2706, opened 2023-08-01
Add realized configuration peristence mechanism. Namespace removal.

michaelcfanning:
Provides a mechanism to persist all realized configuration, as specified on the command-line (including any serialized config file that's been loaded) to an output file.

This persisted configuration file can be round-tripped to the tool to render the identical analysis configuration.

This allows users to render a very complex command-line, and then to capture it in a single file that can provide a simplified command-line experience (i.e., you simply pass the XML via the `-c` switch) or so that the detailed configuration can be manually adjusted (for example, to create a specialized min-bar of analysis checks, by disabling certain rules). Currently not all of our settings (and rule disablement is a good example) is available on the command-line.

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -825,12 +825,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                     : null;
             }
 
-            if (!File.Exists(configurationFilePath)) 
-            { 
+            if (!File.Exists(configurationFilePath))
+            {
                 string fileName = Path.GetFileNameWithoutExtension(configurationFilePath);
                 string spamDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
                 fileName = Path.Combine(spamDirectory, $"{fileName}.xml");
-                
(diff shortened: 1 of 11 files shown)


src/Sarif/AnalyzeContextBase.cs
```diff
@@ -112,7 +113,7 @@ public virtual int ChannelSize
             set => this.Policy.SetProperty(ChannelSizeProperty, value);
         }
 
-        public virtual Guid? AutomationGuid
+        public virtual Guid AutomationGuid
```
michaelcfanning, line 116: >AutomationGuid [](http://example.com/codeflow?start=28&length=14)

Our XML serializer can't round-trip nullable types currently...

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -824,12 +825,25 @@ internal string GetConfigurationFileName(string configurationFilePath, IFileSyst
                     : null;
             }
 
+            if (!File.Exists(configurationFilePath))
```
michaelcfanning, line 828: This is preparatory content for publishing standard configurations in our tool output directory. I'd like to use this to do things like 'run the standard security rule set'.

PR #2715, opened 2023-09-13
Http mock test helper improvement

shaopeng-gh:
The problem is the timestamp difference between mocking and retrieving the test result, causing that we may not be able to retrieve the correct result form the cache list, because it compares all headers and the headers contains the timestamp, and also the other fields that generated based on the timestamp.

In most cases, we set up one request and response pair.
The header matching mechanism is only needed for those situations where we need more than one pair.
With this change, 1 to 1 mocking (which is the vast majority) will no longer need to worry about the timestamp.
The rest of the code path is kept for those cases need it.

### DIFF
--- a/src/Test.Utilities.Sarif/HttpMockHelper.cs
+++ b/src/Test.Utilities.Sarif/HttpMockHelper.cs
@@ -15,6 +15,8 @@ namespace Microsoft.CodeAnalysis.Sarif
     {
         public const string AnyContentText = "29f8354b-8b0d-4d21-91ac-bd04c47b85fb";
 
+        private int callCount = 0;
+
         public static StringContent AnyContent()
         {
             return new StringContent(AnyContentText);
@@ -60,8 +62,8 @@ namespace Microsoft.CodeAnalysis.Sarif
             return new HttpResponseMessage(HttpStatusCode.NonAuthoritativeInformation);
         }
 
-        private readonly Queue<HttpResponseMessage> mockedResponses =
-            new Queue<HttpResponseMessage>();
+        private readonly List<HttpResponseMessage> mockedResponses =
+            new List<HttpResponseMessage>();
 
         public static HttpResponseMessage GetResponseForStatusCode(HttpStatusCode statusCode)
         {
@@ -82,17 +84,20 @@ namespace Microsoft.CodeAnalysis.Sarif
 
         public void Mock(HttpRequestMessage httpRequestMessage, HttpStatusCode httpStatusCode, HttpContent responseContent)
         {
-            this.mockedResponses.Enqueue(
+            callCount++;
+            this.mockedResponses.Add(
                 new HttpResponseMessage(httpStatusCode) { RequestMessage = httpRequestMessage, Content = responseContent });
         }
 
         public void Mock(HttpResponseMessage httpResponseMessage)
         {
-            this.mockedResponses.Enqueue(httpResponseMessage);
+            callCount++;
+            this.mockedResponses.Add(httpResponseMessage);
         }
 
         public void Clear()
         {
+            callCount = 0;
             this.mockedResponses.Clear();
         }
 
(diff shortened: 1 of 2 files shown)


src/Test.Utilities.Sarif/HttpMockHelper.cs
```diff
@@ -116,6 +116,13 @@ private static bool CompareHeaders(HttpRequestHeaders headers1, HttpRequestHeade
 
         protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
+            // In most cases, we set up one request and response pair.
+            // The matching mechanism is only used for those situations where we need more than one pair.
+            if (this.mockedResponses.Count == 1 && mockedResponses[0].Item1.RequestUri != null)
```
shaopeng-gh, line 121: >mockedResponses[0].Item1.RequestUri != null [](http://example.com/codeflow?start=51&length=43)

1. Please check the pr description for the reason of this change.
2. This also fast return on the case of that we expect to have unhandled exception.
3.  All SPM tests still pass with this change.
4. If this is merged, I can work on change the tests to removed the need of using same timestamps
 e.g. from:
`signer.SignRequest(defaultRequest, "region1", "storage1", timestamp);`
to 
`signer.SignRequest(defaultRequest, "region1", "storage1");`
or just keep them? they works, change or not.
5. the earlier i was thinking about hard code a list of time related strings to skip when compare header, that was discarded: [https://github.com/microsoft/sarif-sdk/pull/2714](https://github.com/microsoft/sarif-sdk/pull/2714)

michaelcfanning:
>         public static readonly HttpResponseMessage OKResponse =

Please get rid of this property, we should only have CreateOkResponse. Once this single response is read, it is no longer valid, so publishing instance fields doesn't make sense.

---
Refers to: src/Test.Utilities.Sarif/HttpMockHelper.cs:28 in a5d8664. [](commit_id = a5d8664e0d60c5659b43a60092bff19f6f9d5f1a, deletion_comment = False)

michaelcfanning:
>         public static readonly HttpResponseMessage NotFoundResponse =

Please convert this field and all others to a CreateXXX method, so this one should be CreateNotFoundResponse(). Then delete this public static field. Convert all tests to use the creation methods.

---
Refers to: src/Test.Utilities.Sarif/HttpMockHelper.cs:31 in a5d8664. [](commit_id = a5d8664e0d60c5659b43a60092bff19f6f9d5f1a, deletion_comment = False)

src/Test.Utilities.Sarif/HttpMockHelper.cs
```diff
@@ -116,6 +116,13 @@ private static bool CompareHeaders(HttpRequestHeaders headers1, HttpRequestHeade
 
         protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
michaelcfanning, line 117: >SendAsync [](http://example.com/codeflow?start=53&length=9)

There is an easier change. Update this class to contain an array/list of response. Maintain a counter of calls to SendAsync and simply return the response associated with that index.

In other words, return 'mockedResponse[callCount++]'

That's it. Now you can delete the other items from the response tuple. Now we can stop matching as below.

The reason this works is that in *all* our tests, the calls to SendAsync are sequential and ordered.

michaelcfanning (changes requested):
:clock1:

src/Test.Utilities.Sarif/HttpMockHelper.cs
shaopeng-gh: This approach came to me first but I was not sure 
1. the possibility of Async call will return same order as executed, e.g. the first land on a core that is more congested than a second one.
2. tests to be executed in parallel

@michaelcfanning  let me know

michaelcfanning: Every mock helper should be created per rule that is under test. I think all our rules effectively make these calls synchronously on the same thread, i.e., they await the result of the first before moving on. So i think the approach i suggest will be ok.

shaopeng-gh: 1. I have fixed the other 2 comments about using CreateXXX method. CodeFlow freeze if I try to reply them.
2. I have tried using 3 ways: maintain a counter with List, Queue without counter, ConcurrentQueue without counter,
only ConcurrentQueue works better than the other two(many errors in the other repo), 
because it is async and async methods typically span on a different thread to avoid blocking the calling thread. 
3. **please check with the PR that actually applies it, in the other repo**.

michaelcfanning (approved):
:shipit:

PR #2717, opened 2023-09-29
Fixing crash in dump-events

scottoneil-ms:
I'm not 100% sure why this event started getting into my etl logs and triggering crashes, but I suspect the right answer is to treat it as another way of recognizing session start.

### DIFF
--- a/src/Sarif.Driver/DumpEventsCommand.cs
+++ b/src/Sarif.Driver/DumpEventsCommand.cs
@@ -199,7 +199,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
                         case "RuleFired":
                         {
-                            FailureLevel level = (FailureLevel)(int)(uint)traceEvent.PayloadByName("level");
+                            var level = (FailureLevel)(int)(uint)traceEvent.PayloadByName("level");
                             data1 = level;
                             data2 = traceEvent.PayloadByName("matchIdentifier");
                             ruleId = (string)traceEvent.PayloadByName(nameof(ruleId));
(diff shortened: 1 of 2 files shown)


michaelcfanning (approved):
:shipit:

src/Sarif.Driver/DumpEventsCommand.cs
```diff
@@ -297,8 +297,9 @@ public int Run(DumpEventsOptions options)
                         case "ManifestData":
                         {
                             return;
-
                         }
+
+                        case "EventTrace/PartitionInfoExtension":
```
michaelcfanning, line 302: >PartitionInfoExtension [](http://example.com/codeflow?start=41&length=22)

well, okey-doke. :) you collected the legacy form, it appears.

michaelcfanning:
Suggest adding a release note on this in ReleaseNotes.md, similar to following. Please replace `NullReferenceException` with the actual type of the exception you see when we have this mismatched start/stop pair.

BUG: Resolve trace parsing `NullReferenceException` exception by updating `dump-events` command to process `PartitionInfoExtension` session event as we do `PartitionInfoExtensionV2`.

src/Sarif.Driver/DumpEventsCommand.cs
scottoneil-ms: If it's obvious what I did to trigger legacy form mode instead of V2, I'd love to be corrected.  This is the command I'm using.  [1]

[1] Sarif.PatternMatcher.cli.exe analyze ..\\..\\..\\..\\..\\..\\AzureDevOps\ --recurse true --max-file-size-in-kb 999999999 --output .\ado.sarif --level "Error;Warning;Note" --plugin ".\SEC101.SecurePlaintextSecretsInternal.json" --log ForceOverwrite --insert ContextRegionSnippets --etw .\ADO-Run.etl

michaelcfanning (approved):

PR #2722, opened 2023-10-09
Fix net462 version of the Multitool not working issue

shaopeng-gh:
* BUG: Fix `Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'Sarif.Multitool.Library, Version=...` when using net462 version of the Multitool.

The issue is from we changed from net472 to 461 and 462,
most of them are built for 461, we need to copy them or will have them missing in the package.

image


ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.3.5** UNRELEASED
```
michaelcfanning, line 2: >UNRELEASED [](http://example.com/codeflow?start=14&length=10)

Maybe we should just switch to build this as a dotnet tool!

michaelcfanning (approved):
:shipit:

michaelcfanning (approved):

PR #2724, opened 2023-10-24
Newline index: drop support for unicode line and paragraph separators.

michaelcfanning:
* BRK: Remove processing of `/u2028` (Unicode line separator) and `/u2029` (Unicode paragraph separator) from `NewLineIndex`.
* BUG: Resolve `KeyNotFoundException: The given key was not present` exception when scanning content that contains Unicode line and paragraph separators (`/u2028` and `/u2029`) when enabling `OptionallyEmittedData.RollingHashPartialFingerprints`.

### DIFF
--- a/src/Sarif/NewLineIndex.cs
+++ b/src/Sarif/NewLineIndex.cs
@@ -19,7 +19,7 @@ namespace Microsoft.CodeAnalysis.Sarif
         // which is the line in the file at index n.
         private readonly ImmutableArray<int> _lineOffsetStarts;
 
-        // Realistically, no IDE creates start line that honor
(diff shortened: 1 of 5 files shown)


src/Sarif/NewLineIndex.cs
```diff
@@ -19,12 +19,14 @@ public class NewLineIndex
         // which is the line in the file at index n.
         private readonly ImmutableArray<int> _lineOffsetStarts;
 
+        // Realistically, no IDE create line locations based on the
```
suvamM, line 22: "creates"

suvamM (approved):

PR #2725, opened 2023-10-26
Preorder file traversal 

scottoneil-ms:
This drops time to first scan from 5s to .5s in one of our large-ish repos.  Could also have systematic benefits, as it "streams" the targets out instead of maintaining a list of all directories in memory.  This could be significant for very large repos. AB#{2117606}

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## UNRELEASED
+* PRF: Switch file system traversal to pre-order with producer-consumer to accelerate time to scan first artifact.
+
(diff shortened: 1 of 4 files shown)


ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## UNRELEASED
+* PRF: Switch file system traversal to pre-order with producer-consumer to accelerate time to scan first artifact.
```
michaelcfanning, line 3: >Switch [](http://example.com/codeflow?start=7&length=6)

I feel like we will require some testing for this. Did you look at the existing tests for the ordered file specifier class?

scottoneil-ms: Just dug into the existing coverage a bit.  We do have functional coverage.  Does stuff like checking that we're hitting the right number of files.  I also ran coverage.  [1]  I'm not sure why I'm showing line coverage instead of block, (and need to be shifting focus presently) but it's about 85% from the Sarif.Driver tests.  Running more of our existing tests probably won't get this number up any because based on code highlighting, pretty much all of the missed lines are in error paths.

Those are in two categories: 1) the CheckFaulted() return statements on the worker thread, and 2) the Exception handler in EnumeratedArtifacts() that deals with Exceptions coming out of the Channel and cancellation.

If I Moq CheckFaulted, I probably get it over 90% but maybe fall a bit short of 95%.  I don't think it makes sense to try to hit every early-return path out of the worker thread.

[1] 
image

michaelcfanning:
Thanks for the analysis, ok with getting this one in.

PR #2726, opened 2023-11-01
Refactoring environment setup docs, and suppressing json package warnings.

scottoneil-ms:
AB#2119933

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,29 +18,11 @@ In addition to the object model, the SARIF SDK provides a set of helper classes 
 ## Building the SDK
 
 If you want to build the SDK from source, rather than consuming the NuGet package,
-proceed as follows:
+please refer to [Devbox Setup](docs/devbox-setup.md) for help setting up a compilation
+environment.  
 
-1. Install .NET Core SDK 2.1 and 3.1 from https://dotnet.microsoft.com/download
-
-2. Ensure that Visual Studio 2019 is installed on your machine.
-
-    You can build in VS 2017 as well.
-
-3. Ensure that your Visual Studio installation includes the components that support
-    - C# development
-
-4. Open a Visual Studio 2019 Developer Command Prompt Window.
-
-5. From the root directory of your local repo, run the command `BuildAndTest.cmd`.
-    This restores all necessary NuGet packages, builds the SDK, and runs all the tests.
-
-    All build output appears in the `bld\` subdirectory of the repo root directory.
-
-    NOTE: You must run `BuildAndTest.cmd` once _before_ attempting to build in
-    Visual Studio, to ensure that all required NuGet packages are available.
-
-6. After you have run `BuildAndTest.cmd` once, you can open any of the solution files
-in the `src\` directory in Visual Studio 2017, and build them by running **Rebuild Solution**.
+Then you should be able to build in Visual Studio or on the Command Line using 
+`BuildAndTest.cmd`.
 
 
 ## Accomplishing common tasks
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -18,29 +18,11 @@ In addition to the object model, the SARIF SDK provides a set of helper classes
 ## Building the SDK
 
 If you want to build the SDK from source, rather than consuming the NuGet package,
```
scottoneil-ms, line 20: I didn't find this stuff in onboarding.  Readme is obviously good to have, but making it thinner with links into docs creates more paths to finding the content.  If you go to the Readme, you'll still find it via the link.  But if you're scanning the "docs" folder to see what's there, you'll now see "devbox-setup.md" whereas before you wouldn't see it at all.

src/Sarif.WorkItems/SarifWorkItemFiler.cs
```diff
@@ -336,7 +336,9 @@ public SarifWorkItemModel FileWorkItemInternal(SarifLog sarifLog, SarifWorkItemC
 
                     if (string.IsNullOrWhiteSpace(sarifWorkItemModel.Title) || string.IsNullOrWhiteSpace(sarifWorkItemModel.BodyOrDescription))
                     {
-                        this.Logger.LogWarning("Attempt to call work item client with invalid work item values.", logId);
+                        this.Logger.LogWarning("Attempt to call work item client with invalid work item values. Title: '{sarifWorkItemModel.Title}', BodyOrDescription:{sarifWorkItemModel.BodyOrDescription}",
```
scottoneil-ms, line 339: This is something I was seeing as an error, but is probably a warning for other people.  It's a pretty obvious "just fix it" case.

I want to say I'm going to be methodical about rest of the warning work.  I'm only frontloading these two issues on an ad-hoc basis because a) they're causing pain locally, and b) the resolutions are obvious and should be noncontroversial.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -40,14 +40,19 @@
   <Choose>
     <When Condition="$(TargetFramework) == 'netstandard2.0'">
       <ItemGroup>
-        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+        <!-- We have to ship pre-patch versions of NewtonSoft for VisualStudio SDK.
```
michaelcfanning, line 43: >NewtonSoft [](http://example.com/codeflow?start=51&length=10)

Actually, for this case, v9.0.1 is required to maximize down-level compatibility for this assembly.  Specifically, v9 is the first newtonsoft version to support .NET standard 1.0.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -40,14 +40,19 @@
...
+        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" NoWarn="NU1903" />
+
         <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
         <PackageReference Include="System.Text.Encoding.Extensions" Version="4.3.0" />
       </ItemGroup>
     </When>
     <Otherwise>
       <ItemGroup>
-        <PackageReference Include="Newtonsoft.Json" Version="8.0.1" />
+        <!-- We have to ship pre-patch versions of NewtonSoft for VisualStudio SDK.
```
michaelcfanning, line 53: >patch [](http://example.com/codeflow?start=33&length=5)

This reference is for .net framework builds. For this condition, I think our intent was to take the maximal version for .NET framework only. But does that even make sense? Also, I see that even for this standard, 8.0.3 is the last version.

So minimally, we should update this reference to 8.0.3. I honestly couldn't say without more research whether we actually need to drop to 8.X for this case.

michaelcfanning (approved):

PR #2735, opened 2023-11-21
Version conflicts

scottoneil-ms:
It looks like I never opened this PR.  It is intended to clear up build warnings and especially version conflict issues at the sarif-sdk level.  A lot of this change is removing extraneous references.  Since everything is working well at the consumer level, it looks like this work isn't strictly necessary to get to clean there (I worked this change in the same enlistment so it wasn't clear as I was working if they were severable.)  However, I think we should definitely want to take this one especially to get the change that turns on warnings as errors for assembly conflict issues.


michaelcfanning (approved):
:shipit:

PR #2738, opened 2023-11-29
Fix component compliance issues

yongyan-gh:
## Description

- CVE-2023-36414:
  - Explicitly add reference to `Azure.Identity` 1.10.2 to fix the vulnerability in 1.3.0 via Microsoft.Azure.Kusto.Data 10.0.3.

- CVE-2022-41064:
  - Explicitly add reference to `Microsoft.Data.SqlClient` 2.1.2 to fix the vulnerability in 2.1.1 via Microsoft.Azure.Kusto.Data 10.0.3.
  - Explicitly add reference to `System.Data.SqlClient` 4.8.5 to fix the vulnerability in 4.4.2 via `Microsoft.TeamFoundationServer.Client` 16.170.0

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **Unreleased**
+* DEP: Explicitly add `Azure.Identity` 1.10.2 in `Sarif.Multitool.Library` and `WorkItems` to fix the vulnerability in 1.3.0 via `Microsoft.Azure.Kusto.Data` 10.0.3.
+* DEP: Explicitly add `Microsoft.Data.SqlClient` 2.1.2 in `Sarif.Multitool.Library` and `WorkItems` to fix the vulnerability in version 2.1.1 via `Microsoft.Azure.Kusto.Data` 10.0.3.
+* DEP: Explicitly add `System.Data.SqlClient` 4.8.5 in `WorkItems` to fix the vulnerability in version 4.2.2 via `Microsoft.TeamFoundationServer.Client` 16.170.0
+
 ## **v4.3.7** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.3.7) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.3.7) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.3.7)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.3.7) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.3.)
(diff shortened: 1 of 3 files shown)


ReleaseHistory.md
```diff
@@ -1,5 +1,10 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
+## **Unreleased**
+* DEP: Explicitly add `Azure.Identity` 1.10.2 in `Sarif.Multitool.Library` and `WorkItems` to fix the vulnerability in 1.3.0 via `Microsoft.Azure.Kusto.Data` 10.0.3.
```
michaelcfanning, line 4: >the [](http://example.com/codeflow?start=98&length=3)

can we say 'a' vulnerability here as I suspect that the one in question is not the sole bug in 1.3.0.

Also should we explicitly say 'fix a security vulnerability' ?

src/Test.UnitTests.Sarif.Driver/Test.UnitTests.Sarif.Driver.csproj
```diff
@@ -35,7 +35,7 @@
     <PackageReference Include="Microsoft.Coyote" Version="$(CoyoteVersion)" />
     <PackageReference Include="Microsoft.Coyote.Test" Version="$(CoyoteVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.1" />
-    <PackageReference Include="Moq" Version="4.20.69" />
+    <PackageReference Include="Moq" Version="4.18.4" />
```
michaelcfanning, line 38: >Version [](http://example.com/codeflow?start=36&length=7)

down-leveled??

michaelcfanning: your description doesn't describe why we're making this change. is there a security issue in latest moq that requires us to downlevel?

scottoneil-ms: Good catch.

yongyan-gh: Added that into description. 

According to [https://dev.azure.com/mseng/1ES/_componentGovernance/Spmi/alert/155580?typeId=252438&pipelinesTrackingFilter=0](https://dev.azure.com/mseng/1ES/_componentGovernance/Spmi/alert/155580?typeId=252438&pipelinesTrackingFilter=0) :

"This package exfiltrates data unexpectedly during installation. Do not use this version of the component. Pin to version 4.18.4."

ReleaseHistory.md
yongyan-gh: how about "... to avoid the vulnerable package per compliance requirements"?

PR #2739, opened 2023-11-30
Add binary vs. textual file enumeration.

michaelcfanning:
* BRK: `EnumeratedArtifact` now sniffs artifacts to distinguish between textual and binary data. The `Contents` property will be null for binary files (use `Bytes` instead).
* BRK: `EnumeratedArtifact` raises `InvalidOperationException` for non-seekable streams (unless the `SupportNonSeekableStreams` property is set). Non-seekable streams result in performance inefficiencies when expanding.
* BRK: `MultithreadedZipArchiveArtifactProvider` now distinguishes binary vs. textual data using a hard-coded binary files extensions list. This data will be made configurable in a future change.
* NEW: `EnumeratedArtifact` now automatically detects and populates a `Bytes` property for binary files such as executables and certificates.

Previously, artifact enumeration assumed all data consisted of text. This breaks attempts to load/examine binary data.

In `EnumeratedArtifact`, we now examine the beginning of all seekable streams to detect whether we are seeing textual or binary data. We do so using an OSS snippet that looks for null/control characters in that data. This code also attempts to retrieve a range of text file encodings but this capability is not verified. Binary files are retrieved as byte arrays. Text files go to the existing `Contents` property. 

If we have a non-seekable stream, what we have to do is retrieve the file as a byte array and then subsequently convert it to text if we determine that it is textual. This is a performance degradation from our previous implementation and so we require users to explicitly set a `SupportNonSeekableStreams` property to ensure they know what they are opting into. 

Separately, the `MultithreadedZipArchiveArtifactProvider` *does* enumerate non-seekable streams, since they comprise compressed data. This provider uses a specialized `ZipArchiveArtifact` kind. Rather than accepting the costs to double-transform a text stream to bytes and then a decoded string, we simply hard-code a set of binary file extensions and only retrieve these as bytes. This property is settable. This isn't a really ideal solution but works for our immediate scenarios. You could imagine updating this class to optionally allow for the non-performant behavior (expand as bytes, sniff to classify as binary or text, convert text to strings). This would allow for higher fidelity classification without requiring advance knowledge of file extensions.

Code coverage information:

94.38% : `EnumeratedArtifact`
97.22% : `MultithreadedZipArtifactProvider`
93.59% : `ZipArchiveArtifact`

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -737,7 +737,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
                     DriverEventSource.Log.ReadArtifactStart(filePath);
                     // Reading the length property faults in the file contents.
-                    long sizeInBytes = perFileContext.CurrentTarget.Contents.Length;
+                    long sizeInBytes = perFileContext.CurrentTarget.SizeInBytes.Value;
                     DriverEventSource.Log.ReadArtifactStop(filePath, sizeInBytes);
 
                     DetermineApplicabilityAndAnalyze(perFileContext, skimmers, disabledSkimmers);
--- a/src/Sarif/ArtifactProvider.cs
+++ b/src/Sarif/ArtifactProvider.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
-using System.Text;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
@@ -24,109 +23,5 @@ namespace Microsoft.CodeAnalysis.Sarif
         public virtual IEnumerable<IEnumeratedArtifact> Artifacts { get; set; }
 
         public IFileSystem FileSystem { get; set; }
-    }
-
-    public class SinglethreadedZipArchiveArtifactProvider : ArtifactProvider
-    {
-        public SinglethreadedZipArchiveArtifactProvider(ZipArchive zipArchive, IFileSystem fileSystem) : base(fileSystem)
-        {
-            var artifacts = new List<IEnumeratedArtifact>();
-
-            foreach (ZipArchiveEntry entry in zipArchive.Entries)
-            {
-                artifacts.Add(new EnumeratedArtifact(Sarif.FileSystem.Instance)
-                {
-                    Uri = new Uri(entry.FullName, UriKind.RelativeOrAbsolute),
-                    Contents = new StreamReader(entry.Open()).ReadToEnd()
-                });
-            }
-
-            Artifacts = artifacts;
-        }
-    }
-
-    public class MultithreadedZipArchiveArtifactProvider : ArtifactProvider
-    {
-        private readonly ZipArchive zipArchive;
-
-        public MultithreadedZipArchiveArtifactProvider(ZipArchive zipArchive, IFileSystem fileSystem) : base(fileSystem)
-        {
-            this.zipArchive = zipArchive;
-        }
-
-        public override IEnumerable<IEnumeratedArtifact> Artifacts
-        {
-            get
-            {
-                foreach (ZipArchiveEntry entry in this.zipArchive.Entries)
-                {
-                    yield return new ZipArchiveArtifact(this.zipArchive, entry);
-                }
-            }
-        }
-    }
-
-    public class ZipArchiveArtifact : IEnumeratedArtifact
-    {
-        private ZipArchiveEntry entry;
-        private readonly ZipArchive archive;
-        private readonly Uri uri;
-        private string contents;
-
-        public ZipArchiveArtifact(ZipArchive archive, ZipArchiveEntry entry)
-        {
-            this.entry = entry;
-            this.archive = archive;
-            this.uri = new Uri(entry.FullName, UriKind.RelativeOrAbsolute);
-        }
-
-        public Uri Uri => this.uri;
-
-        public Stream Stream
-        {
-            get
-            {
-                lock (this.archive)
-                {
-                    return entry != null
-                        ? entry.Open()
-                        : null;
-                }
-            }
-            set => throw new NotImplementedException();
-        }
-
-        public Encoding Encoding { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
-
-        public string Contents
-        {
-            get
-            {
-                lock (this.archive)
-                {
-                    if (this.contents != null) { return this.contents; }
-                    this.contents = new StreamReader(Stream).ReadToEnd();
-                    this.entry = null;
-                    return this.contents;
-                }
-            }
-            set => throw new NotImplementedException();
-        }
-
-        public long? SizeInBytes
-        {
-            get
-            {
-                lock (this.archive)
-                {
-                    if (this.entry != null)
-                    {
-                        return this.entry.Length;
-                    }
-                    return this.contents?.Length;
-                }
-            }
-            set => throw new NotImplementedException();
-        }
     }
 }
--- a/src/Sarif/EnumeratedArtifact.cs
+++ b/src/Sarif/EnumeratedArtifact.cs
@@ -93,7 +93,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 }
 
                 // This is our client-side, disk-based file retrieval case.
-                Stream = FileSystem.FileOpenRead(Uri.LocalPath);
+                this.Stream = FileSystem.FileOpenRead(Uri.LocalPath);
             }
 
             if (Stream.CanSeek)
@@ -105,7 +105,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 RetrieveDataFromNonSeekableStream();
             }
 
-            Stream = null;
+            this.Stream = null;
 
             return (this.contents, this.bytes);
         }
--- a/src/Sarif/EnumeratedArtifact.cs
+++ b/src/Sarif/EnumeratedArtifact.cs
@@ -16,44 +16,152 @@ namespace Microsoft.CodeAnalysis.Sarif
             FileSystem = fileSystem;
         }
 
+        internal byte[] bytes;
         internal string contents;
+
+        private bool? isBinary;
+        private Encoding encoding;
 
         public Uri Uri { get; set; }
 
+        public bool IsBinary 
+        { 
+            get
+            {
+                return this.isBinary ?? IsArtifactBinary();
+            }
+        }
+
+        internal bool IsArtifactBinary()
+        {
+            if (isBinary.HasValue) 
+            {
+                return isBinary.Value;
+            }
+
+            GetArtifactData();
+            this.isBinary = this.contents == null;
+            return this.isBinary.Value;
+        }
+
         public Stream Stream { get; set; }
 
-        public Encoding Encoding { get; set; }
+        public Encoding Encoding
+        {
+            get
+            {
+                if (encoding == null)
+                {
+                    GetArtifactData();
+                }
+                return this.encoding;
+            }
+
+            set => this.encoding = value;
+        }
 
         internal IFileSystem FileSystem { get; set; }
 
+        /// <summary>
+        /// Gets or sets a property that determines whether the enumerated artifact
+        /// can obtain data from a non-seekable stream. This operation will be
+        /// less performant in any case where an artifact is a textual file (as
+        /// the file will first be converted to a byte array to determine
+        /// whether it is textual and subsequently turned into a string if it is).
+        /// </summary>
+        public bool SupportNonSeekableStreams { get; set; }
+
         public string Contents
         {
-            get => GetContents();
+            get => GetArtifactData().text;
             set => this.contents = value;
         }
 
-        private string GetContents()
+        public byte[] Bytes
         {
-            if (this.contents != null) { return this.contents; }
+            get => GetArtifactData().bytes;
+            set => this.bytes = value;
+        }
 
-            if (Stream == null && this.contents == null)
+        private (string text, byte[] bytes) GetArtifactData()
+        {
+            if (this.contents != null)
+            { 
+                return (this.contents, bytes: null); 
+            }
+
+            if (this.bytes != null) 
             {
-                // TBD we actually have no validation URI is non-null yet.
-                contents = Uri!.IsFile
-                    ? FileSystem.FileReadAllText(Uri.LocalPath)
-                    : null;
+                return (text: null, this.bytes);
+            }
 
-                this.sizeInBytes = (long?)this.contents?.Length;
+            if (Stream == null && this.contents == null && this.bytes == null)
+            {
+                if (Uri == null || (Uri.IsAbsoluteUri && !Uri.IsFile)) 
+                {
+                    throw new InvalidOperationException();
+                }
+
+                // This is our client-side, disk-based file retrieval case.
+
+                Stream = FileSystem.FileOpenRead(Uri.LocalPath);
+            }
+
+            bool isText = false;
+
+            if (Stream.CanSeek) 
+            { 
+                // Reset to beginning of stream in case caller neglected to do so.
+                this.Stream.Seek(0, SeekOrigin.Begin);
+
+                byte[] header = new byte[4096];
+                int length = this.Stream.Read(header, 0, header.Length);
+                isText = FileEncoding.CheckForTextualData(header, 0, length, out this.encoding);
+
+                if (isText) 
+                { 
+                    // If we have textual data and the encoding was null, we are UTF8
+                    // (which will be a perfectly valid encoding for ASCII as well).
+                    this.encoding ??= Encoding.UTF8;
+                }
+
+                this.Stream.Seek(0, SeekOrigin.Begin);
+
+                if (isText)
+                {
+                    using var contentReader = new StreamReader(Stream);
+                    this.contents = contentReader.ReadToEnd();
+                }
+                else
+                {
+                    this.bytes = new byte[Stream.Length];
+                    this.Stream.Read(this.bytes, 0, bytes.Length);
+                }
             }
             else
             {
-                if (Stream.CanSeek) { this.Stream.Seek(0, SeekOrigin.Begin); }
-                using var contentReader = new StreamReader(Stream);
-                this.contents = contentReader.ReadToEnd();
-                Stream = null;
+                if (!SupportNonSeekableStreams) 
+                {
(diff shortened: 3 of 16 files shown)


src/Sarif/ArtifactProvider.cs
```diff
@@ -25,108 +24,4 @@ public ArtifactProvider(IEnumerable<IEnumeratedArtifact> artifacts)
 
         public IFileSystem FileSystem { get; set; }
     }
-
-    public class SinglethreadedZipArchiveArtifactProvider : ArtifactProvider
```
michaelcfanning, line 29: >SinglethreadedZipArchiveArtifactProvider [](http://example.com/codeflow?start=17&length=40)

Notoriously, this single file has included multiple class definitions, I finally took a few minutes and extracted all of them.

src/Sarif/FileEncodings.cs
```diff
@@ -0,0 +1,137 @@
...
+            return nullSequences == 0 && (controlSequences <= (rawData.Length / 10));
+        }
+
+        /// <summary>
+        /// Detects if data has bytes order mark to indicate its encoding for textual data.
+        /// </summary>
+        /// <param name="rawData">The raw data.</param>
+        /// <param name="start">The start.</param>
+        /// <returns></returns>
+        public static bool CheckForByteOrderMark(byte[] rawData, out Encoding encoding, int start = 0)
```
michaelcfanning, line 103: >CheckForByteOrderMark [](http://example.com/codeflow?start=27&length=21)

We do not have explicit testing/use of BOM detection. I have a stubbed test but haven't nailed down how to generate useful test cases. I'm leaving this (and a stub test) for the future.

In practice, detecting/handling esoteric encodings is not a prominent scenario for most analysis scenarios.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,162 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
 
+        public bool IsBinary
+        {
+            get
+            {
+                return this.isBinary ?? IsArtifactBinary();
+            }
+        }
+
+        internal bool IsArtifactBinary()
```
scottoneil-ms, line 35: Suggest: make private to narrow API surface to IsBinary?

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,162 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+
+        private void RetrieveDataFromNonSeekableStream()
+        {
+            bool isText;
+            if (!SupportNonSeekableStreams)
+            {
+                throw new InvalidOperationException("Stream is not seekable. Provide a seekable stream or set the 'SupportNonSeekableStreams' property.");
+            }
+
+            this.bytes = new byte[Stream.Length];
```
scottoneil-ms, line 133: Why is it the case that we can categorize a Nonseekable stream in 4096 bytes, but we need to read in the full stream for a Seekable one?  If 4096 is sound, I suspect we'd both love to take a one-block peek in the Nonseekable case, and use that as the starting buffer for a full Stream or byte[] read depending on what is appropriate.  So I probably just don't see a fatal flaw yet.

Where is 4096 coming from in the Seekable case?  Looking at the implementation of FileEncoding, I'd optimistically credit that we probably don't get much more accuracy in the classifier after this many bytes if that's coming from some tribal knowledge or guidance stemming from other users of this code.  But right now that's just a brittle prior without any experience on with so many of these file extensions, so I'm really curious what the author guidance or community discourse looks like.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,162 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+        internal bool IsArtifactBinary()
+        {
+            if (isBinary.HasValue)
+            {
+                return isBinary.Value;
+            }
+
+            GetArtifactData();
+            this.isBinary = this.contents == null;
+            return this.isBinary.Value;
```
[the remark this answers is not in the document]
michaelcfanning: Yes! and just delete the backing field.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,162 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+
+        private void RetrieveDataFromSeekableStream()
+        {
+            bool isText;
+
+            // Reset to beginning of stream in case caller neglected to do so.
+            this.Stream.Seek(0, SeekOrigin.Begin);
+
+            byte[] header = new byte[4096];
+            int length = this.Stream.Read(header, 0, header.Length);
```
scottoneil-ms, line 155: I am fully cognizant that this is almost certainly out of scope for what you've set out to accomplish but I'm duty-bound to wonder aloud when I see more synchronous code being crafted that does full file reads: do you see a path to the Task Programming Mode?

src/Sarif/EnumeratedArtifact.cs
scottoneil-ms: Here's maybe a way to put a finger on what I'm getting at:

The snippet code we're taking asserts that if you see no more than 10% of the bytes as being non-textual, then that's a good-enough heuristic.  I buy that.  Not second-guessing it.  But that's the kind of suspiciously-round number someone just yanked out of their brain one day and it just worked well enough in the real world that it stuck and became an institution.

Maybe 4096 is the same way and we don't need to do full file scans to classify in any scenario?  This is what I'm wondering/hoping.  (Not asserting.)

src/Sarif/SinglethreadedZipArchiveArtifactProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class SinglethreadedZipArchiveArtifactProvider : ArtifactProvider
+    {
+        public SinglethreadedZipArchiveArtifactProvider(ZipArchive zipArchive, IFileSystem fileSystem) : base(fileSystem)
+        {
+            var artifacts = new List<IEnumeratedArtifact>();
+
+            foreach (ZipArchiveEntry entry in zipArchive.Entries)
```
scottoneil-ms, line 16: Will this indirectly recurse?  Do we want it to?

(This might become obvious to me later, but tossing it out there in case it doesn't and I forget.)

src/Sarif/ZipArchiveArtifact.cs
```diff
@@ -0,0 +1,105 @@
...
+            this.uri = new Uri(entry.FullName, UriKind.RelativeOrAbsolute);
+        }
+
+        public Uri Uri => this.uri;
+
+        public Stream Stream
+        {
+            get
+            {
+                lock (this.archive)
```
scottoneil-ms, line 36: Maybe strengthen error checking in the constructor to preclude the null condition and elide the lock?

src/Sarif/ZipArchiveArtifact.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (this.contents == null && this.bytes == null)
+            {
+                lock (this.archive)
+                {
+                    if (this.contents == null && this.bytes == null)
+                    {
+                        string extension = Path.GetExtension(Uri.ToString());
+                        if (this.binaryExtensions.Contains(extension))
+                        {
+                            this.bytes = new byte[Stream.Length];
```
scottoneil-ms, line 71: Do all our underlying Streams support .Length?

It is more defensive to promote these .bytes fields to MemoryStreams and use Stream.Copy.  If I'm reading the docs right: wherever the underlying Stream supports .CanSeek, the MemoryStream can be precisely presized by safely querying .Length.  Otherwise, .Length can throw for some implementations (but not others.)

The tradeoff of doing this is that if we ever needed a precisely-sized byte[] for an artifact later, we would potentially need to take a copy.  (But not in cases where we were able to size it precisely.)

src/Sarif/ZipArchiveArtifact.cs
scottoneil-ms: Bonus: if we do need locking, it would be nice to be locking the entry as opposed to the archive.  (Assuming you don't need a whole archive locked to safely open an entry and such.)

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,162 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+                return (this.contents, bytes: null);
+            }
+
+            if (this.bytes != null)
+            {
+                return (text: null, this.bytes);
+            }
 
-            if (Stream == null && this.contents == null)
+            if (Stream == null && this.contents == null && this.bytes == null)
```
yongyan-gh, line 98: >Stream [](http://example.com/codeflow?start=16&length=6)

nit: have both `Stream` and `this.Stream`. could be consistent by using 1 style `this.Stream`

src/Test.FunctionalTests.Sarif/Multitool/BaselineOptionTests.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Mime;
+using System.Text;
 
 using FluentAssertions;
```
scottoneil-ms, line 10: Esp. since there are some new files here and you're a fellow appreciator: have you measured code coverage?  If so: where did it land?

src/Sarif/FileEncodings.cs
```diff
@@ -0,0 +1,137 @@
...
+            // http://www.gnu.org/software/diffutils/manual/html_node/Binary.html
+            // count the number od null bytes sequences
+            // considering only sequences of 2 0s: "\0\0" or control characters below 10
+            int nullSequences = 0;
+            int controlSequences = 0;
+            for (int i = start + 1; i < count; i++)
+            {
+                if (rawData[i - 1] == 0 && rawData[i] == 0)
+                {
+                    if (++nullSequences > 1)
```
yongyan-gh, line 82: >>  [](http://example.com/codeflow?start=40&length=2)

`++nullSequences > 1` requires finding of 2 '\0\0' sequences. 
Only 1 should be enough? `>= 1` or `> 0`

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -16,44 +16,164 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+
+        private void RetrieveDataFromNonSeekableStream()
+        {
+            bool isText;
+            if (!SupportNonSeekableStreams)
+            {
+                throw new InvalidOperationException("Stream is not seekable. Provide a seekable stream or set the 'SupportNonSeekableStreams' property.");
+            }
+
+            this.bytes = new byte[Stream.Length];
```
yongyan-gh, line 133: >bytes [](http://example.com/codeflow?start=17&length=5)

should we consider adding lock to the bytes array in case a thread tries to read dirty data?

src/Sarif/EnumeratedArtifact.cs
michaelcfanning: 'internal' is still private, i tend to do this to make things testable but not public. As it happens I can now delete this helper entirely.

It now occurs to me: this new property value is untested!! Need to add it to the test validations. So I'm glad you pointed to it. :)

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -737,7 +737,7 @@ private async Task ScanTargetsAsync(TContext globalContext, IEnumerable<Skimmer<
 
                     DriverEventSource.Log.ReadArtifactStart(filePath);
                     // Reading the length property faults in the file contents.
-                    long sizeInBytes = perFileContext.CurrentTarget.Contents.Length;
+                    long sizeInBytes = perFileContext.CurrentTarget.SizeInBytes.Value;
```
yongyan-gh, line 740: >SizeInBytes.Value [](http://example.com/codeflow?start=68&length=17)

...SizeInBytes?.Value ?? 0

SizeInBytes may be null in some weird condition.

src/Sarif/EnumeratedArtifact.cs
michaelcfanning: OK, so first, 4096 is *way* too much data and I dropped it to 1024. This is probably *still* more data than we need.

Let's talk offline further about this (in particular about whether we can address the inefficiency in the non-seekable case of reading the whole thing). I'm just not aware of how we can handle a string decoding from the original non-seekable string if we've already obtained 1024 bytes from it.

src/Test.UnitTests.Sarif/EnumeratedArtifactTests.cs
```diff
@@ -16,7 +20,223 @@ namespace Test.UnitTests.Sarif
...
+                var artifact = new EnumeratedArtifact(FileSystem.Instance)
+                {
+                    Stream = entry.Open(),
+                };
+
+                Assert.Throws<InvalidOperationException>(() => ValidateTextArtifact(artifact, guid.Length));
+            }
+        }
+
+        private void ValidateBinaryArtifact(EnumeratedArtifact artifact, int sizeInBytes)
```
scottoneil-ms, line 219: Inferring standards from practice: are we okay with sequences of assertions wherever the [Fact] is validating a single input/case, as opposed to a multiplicity of them?

I'm hypothesizing that if this method were ValidateBinaryArtifact**s** then it would be using a StringBuilder.

src/Test.UnitTests.Sarif/EnumeratedArtifactTests.cs
```diff
@@ -16,7 +20,223 @@ namespace Test.UnitTests.Sarif
...
+            // that reliably produces disk files with appropriate BOMs/file
+            // formats to properly test the encoding detection from the new
+            // helpers we have added. This test passes but is very incomplete.
+            // If/when we have additional demands to support esoteric encodings
+            // in analysis, we can build this out. Currently, things are not
+            // guaranteed to work reliably except in ASCII, UTF8, Unicode.
+
+            string guid = Guid.NewGuid().ToString();
+
+#pragma warning disable SYSLIB0001 // Type or member is obsolete
```
scottoneil-ms, line 35: 🧐
Which type or member is obsolete and why can't we use the shiny new stuff?

src/Sarif/EnumeratedArtifact.cs
michaelcfanning: Sure thing! It is a SMOP. Just a 'small matter of programming'. :)

src/Sarif/SinglethreadedZipArchiveArtifactProvider.cs
michaelcfanning: So, this should *not*. This is the zip archive's own enumerator, which handles traversing all zip contents. As far as I'm aware, the zip archive format does not allow features (such as a kind of symbolic link) that would result in any cycles/recursion.

src/Sarif/ZipArchiveArtifact.cs
michaelcfanning: You have to lock on the archive! Too bad. We tested the other case. There must be (makes sense) state maintained at the archive level about decompression/other entry state. I'm not sure why we permit a null entry. Will try to rip that out.

src/Sarif/ZipArchiveArtifact.cs
michaelcfanning: ZipArchive supports providing a decompressed length. What's the performance impact of creating a memory stream as you describe? Will it effectively use the wrapped stream as its buffer? And so the length computation is computable before the memory stream did it? :)

This sounds like a nice next clean-up. It also makes me wonder if I've been over-concerned about handling non-seekable streams. This case may be entirely unremarkable based on what you've pointed out: we should simply read the original stream into a byte array, sniff it and then, if we need to decode, we wrap it in a memory stream (which will use the byte array as its buffer and not create a new copy).

src/Sarif/FileEncodings.cs
michaelcfanning: The problem, I believe, is that some BOM sequences has a '\0' character (but no more than 1). So your heuristic might make use believe that a text file with a BOM is binary.

src/Sarif/EnumeratedArtifact.cs
michaelcfanning: made this consistent for this file. We should discuss this style point, whether properties have a `this.` prefix in code is very inconsistent across the code base.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning: Yes, and we should not provoke those conditions. :) There is a unit test for this. It's non-obvious, I'll try to find it.

scottoneil-ms:
Oh, you're right.  I'm just proposing a copy with extra steps.  Sorry.

What about disposing the stream and opening a new one to avoid the full copy?  Call it "Seek(Origin), cave-man style."

src/Sarif/SinglethreadedZipArchiveArtifactProvider.cs
scottoneil-ms: I was unclear and imprecise.  The zip archive doesn't know anything about the contents of its individual entries.  It's counterproductive to zip up a file that is already zipped, but you can.  And it probably happens in some cases where people are automating bulk archiving where other processes and/or user behavior may have caused some of the contents to be zipped up.

To get super concrete: if we're ever working through the entrys in "BulkArchive.zip", and find an entry in it such that entry.Name == "doublezipped.zip" --will we crack open doublezipped.zip?

src/Sarif/EnumeratedArtifact.cs
scottoneil-ms: Actually: if the zip archive is always already in memory, maybe closing/reopening is trivially cheap?

src/Sarif/ZipArchiveArtifact.cs
michaelcfanning: OK. But for this specific location, ZipArchive, these streams are never seekable. Do you have a have suggestion or do you agree this code can ship as is? :)

src/Test.UnitTests.Sarif/EnumeratedArtifactTests.cs
michaelcfanning: Yes! For a true, more atomic unit test, it's fine to break early on the first condition that fails. There's two interesting things to consider, one is what failure information is more useful to the debugger. The second is to ensure that the literal output of the test is illustrative.

Some teams have stringent standards around emitting textual messages with all failures. Seeing 'Expected '17' but observed '13'' isn't always helpful.

PR #2741, opened 2023-12-05
Replace file encodings

michaelcfanning:
* NEW: `FileEncoding.IsTextualData` utility can effectively distinguish between binary and textual data.

Code coverage: 100% (this really makes the point of how critical *data* coverage is as well as code coverage. :)

This change pivots our textual data detection mechanism. What we do now is attempt to decode as UTF8, Windows 1252 and UTF32. If any of these attempts results in a Unicode replacement character or observes an embedded `NUL` (after the first character, where a textual BOM might generate one), we classify the data as binary.

Otherwise it's text.

This utility is running against a 3M file test data set. It is looking generally effective. I am concerned about performance, we need to look at this closely.


src/Sarif/FileEncoding.cs
```diff
@@ -0,0 +1,101 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    public class FileEncoding
```
michaelcfanning, line 11: >FileEncoding [](http://example.com/codeflow?start=17&length=12)

There's nothing in here yet really about detecting file encodings, a very hard problem. We strictly attempt classification but the use of of three specific encodings is critical to this.

We use UTF8 as a general mechanism for sniffing ASCII and various Unicode. We use Windows code page as this CP allows for certain extended ASCII characters that are replaced when decoded as UTF8. Finally, UTF32 (an unusual encoding) is handled specially so that we can use the trick of classifying data as binary when it contains '\0' chars.

Currently, I've used this to classify around 175k files from the world's largest GitHub repository. It's done a pretty impressive job of sorting files for this data set. I found many examples of files whose extension didn't match reality (mostly test data). 

One minor concern, sometimes developers save source code with the Unicode replacement char in it (due to round-tripping the file, perhaps, in an editor that doesn't support some copied text, typically in a code comment). 

We could attempt to address this moving forward by other heuristics, we could provide an extensions list ('always treat these as text/binary') or we could attempt to compute a % of characters in the sniff that were invalid.

I continue to be concerned about the perf impact of introducing this helper in a general scan framework. However, it has also occurred to me that some scanners could be attempting to scan large amounts of binary data as textual (this certainly happens in one of our current tools). And so this classifier might allow all of that analysis to go away. We'll see. I'll test real world data sets today.

src/Sarif/FileEncoding.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                    if (ch == -1)
+                    {
+                        break;
+                    }
+
+                    // Because we enable 'detectEncodingFromByteOrderMarks' we will skip past any NUL
+                    // characters in the file that might result from being BOM-prefixed. So any
+                    // evidence of this character is an indicator of binary data.
+                    if (encoding != Encoding.UTF8 && ch == '\0')
```
michaelcfanning, line 66: >encoding [](http://example.com/codeflow?start=24&length=8)

This isn't a good check for UTF8 as for UTF32 in particular produces '\0' when decoded.

src/Sarif.Converters/Sarif.Converters.csproj
```diff
@@ -52,7 +52,7 @@
       <ItemGroup>
         <!-- We have to ship pre-patch versions of NewtonSoft for VisualStudio SDK. 
              We mitigate risk by limiting nesting depth. -->
-        <PackageReference Include="Newtonsoft.Json" Version="8.0.3" NoWarn="NU1903" />
+        <PackageReference Include="Newtonsoft.Json" Version="9.0.1" NoWarn="NU1903" />
```
scottoneil-ms, line 55: If we're taking this as part of essentially a hotfix, I'd recommend avoiding (if possible --may not be) dependency updates.  

If we do have to take this: is the comment still accurate/NoWarn still needed?  Just want to be sure.

src/Sarif/FileEncoding.cs
```diff
@@ -0,0 +1,101 @@
...
+                var span = new Span<byte>(bytes, start, count);
+                using var reader = new StreamReader(new MemoryStream(span.ToArray()), encoding, detectEncodingFromByteOrderMarks: true);
+                reader.BaseStream.Seek(start, SeekOrigin.Begin);
+
+                bool isTextual = true;
+                bool continueProcessing = true;
+
+                for (int i = start; i < count; i++)
+                {
+                    int ch = reader.Read();
```
scottoneil-ms, line 56: Since this code is iterating over a byte[] it felt to me like the memory stream could be elided and I went down a small rabbit hole trying to figure out how to do that.  I strongly suspect it's not worth trying for an urgent PR.  But if you wanted to play around with another way in your copious free time*, I think this API could do the job?

MemoryMarshal.Read<T>(ReadOnlySpan<byte>)

* (a cheeky euphemism for "probably never" on some teams.)

scottoneil-ms (approved):

src/Sarif.Converters/Sarif.Converters.csproj
michaelcfanning: Yes, good catch. First, I had an unpushed release note entry on this, will resolve that. 

This upgrade to `9.0.1` is implied by the move to `net462`. The latest VS updates should allow for this (we may even be able to be more stringent). 

Push release notes update.

src/Sarif/FileEncoding.cs
michaelcfanning: Oops, spoke too soon. :) Doesn't work with MemoryStream. I reverted but I also added a simple optimization not to create a new array unless the caller has asked only for a partial sniff against the buffer. Will prevent copy in the common case.

PR #2745, opened 2023-12-08
* BUG: Emit `WRN997.OneOrMoreFilesSkippedDueToExceedingSizeLimit` when no valid analysis targets are detected (due to exceeding size limits).

michaelcfanning:
* BUG: Emit `WRN997.OneOrMoreFilesSkippedDueToExceedingSizeLimit` when no valid analysis targets are detected (due to exceeding size limits).

This bug has burned me repeatedly so I finally fixed it. In cases where I attempt to analyze a single file, and it is too large to analysis given current configuration, the console only reports 'no valid targets found', leading me to assume I mistyped the file path or something else: 

> Analyzing...
THREADS: 36
SPMI : error ERR997.NoValidAnalysisTargets : No valid analysis targets were specified.

*Now* we emit the explicit warning that one or more files were skipped, as we do in the case when we find at least one valid target: 

> Analyzing...
THREADS: 36
SPMI : error ERR997.NoValidAnalysisTargets : No valid analysis targets were specified.
SPMI : warning WRN997.OneOrMoreFilesSkippedDueToExceedingSizeLimit : 1 file(s) were skipped for analysis due to exceeding size limit (currently configured as 1024 KB). The 'max-file-size-in-kb' command-line argument can be used to increase this threshold.

Added a solid unit test for this. I was glad to see this test was easy to author, we're getting much better in that regard.


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -36,6 +36,44 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
...
+
+            logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Note).Count().Should().Be(1);
+            Notification note = logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Note).First();
+            note.Descriptor.Id.Should().Be(Notes.Msg002_FileExceedingSizeLimitSkipped);
+
+            logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Warning).Count().Should().Be(1);
+            Notification warning = logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Warning).First();
+            warning.Descriptor.Id.Should().Be(Warnings.Wrn997_OneOrMoreFilesSkippedDueToExceedingSizeLimits);
+
+            logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Note).Count().Should().Be(1);
```
shaopeng-gh, line 72: >FailureLevel.Note [](http://example.com/codeflow?start=68&length=17)

I think the purpose of this change is, 
the visibility of file skipped is low to user (for user that has Note).
It is ok (I think not too bad), 
but when all files are oversized, 
people will appreciate more visibility of it.

In other tools that generates SARIF file, 
we generates warning and promote to error in some condition.
How about we do similar promote the visibility from note to warning, 
instead of producing another message? 

The information included in the note has all and more.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -36,6 +36,44 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
             Output.WriteLine($"The seed that will be used is: {TestRule.s_seed}");
         }
 
+        [Fact]
+        public void AnalyzeCommandBase_ScanWithNoValidTargetsAndOneOrMoreSkippedFilesEmitsSkippedFilesWarning()
```
shaopeng-gh, line 40: >ScanWithNoValidTargetsAndOneOrMoreSkippedFilesEmitsSkippedFilesWarning [](http://example.com/codeflow?start=39&length=70)

throw a much less precise, but easier to read version: `ScanWithOnlyOversizedFilesEmitsFilesSkippedWarning`

it is not precise but it meets what is actually tested in this test.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
shaopeng-gh, line 72: >Note [](http://example.com/codeflow?start=81&length=4)

this should be a copy paste error

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -36,6 +36,44 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
...
+
+            context.RuntimeErrors.Should().Be(conditions);
+
+            logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Note).Count().Should().Be(1);
+            Notification note = logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Note).First();
+            note.Descriptor.Id.Should().Be(Notes.Msg002_FileExceedingSizeLimitSkipped);
+
+            logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Warning).Count().Should().Be(1);
+            Notification warning = logger.ConfigurationNotifications.Where(n => n.Level == FailureLevel.Warning).First();
+            warning.Descriptor.Id.Should().Be(Warnings.Wrn997_OneOrMoreFilesSkippedDueToExceedingSizeLimits);
```
shaopeng-gh, line 70: >Should [](http://example.com/codeflow?start=34&length=6)

suggest a simple wrap so all failure can surface
using (new AssertionScope()){   }

shaopeng-gh (changes requested):
:clock1:

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
shaopeng-gh: I found whether 
`all files are too big`
or
`at least one is within limit`
we will always output this message, so to understand, this effectively can be shorten to understand as 
"always report a warning, if at least one file skipped"

if this behavior is what we want, then we can remove the other copy of  `if (_filesExceedingSizeLimitCount > 0){Warnings.LogOneOrMoreFilesSkippedDueToExceedingSizeLimit(globalContext, filesExceedingSizeLimitCount); }`
and move your copy outside of the check `if (_fileContextsCount == 0)`
for less copy of same code

shaopeng-gh (approved):
:shipit:

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
michaelcfanning: I always forget about this, THANK YOU.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
shaopeng-gh: Without look into code details my understanding the purpose of the change at first sight was wrong. :)

The test names looks great!

PR #2746, opened 2023-12-18
Better UX emitting `FailureLevel.Note` debug messages on the console.

michaelcfanning:
Emit `FailureLevel.Note` messages with label `info` (rather than `fail`) in `ConsoleLogger`.

Previously, for debug (i.e., `FailureLevel.Note` messages) we emitted strings like:

`temp.txt(1,1-108): fail TST101/016: '…xxxDid' is an apparent false positive (because it does not contain at least one digit and letter).`

Now we emit `fail` as a more appropriate label for debug messages:

`temp.txt(1,1-108): info TST101/016: '…xxxDid' is an apparent false positive (because it does not contain at least one digit and letter).`

### DIFF
--- a/src/Sarif/Writers/ConsoleLogger.cs
+++ b/src/Sarif/Writers/ConsoleLogger.cs
@@ -10,9 +10,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
 {
     public class ConsoleLogger : BaseLogger, IAnalysisLogger
     {
-        //  TODO:  We directly instantiate this logger in two classes, creating 
-        //  unamanged dependencies.  Fix this pattern with dependency injection or a factory.
-        //  #2272 https://github.com/microsoft/sarif-sdk/issues/2272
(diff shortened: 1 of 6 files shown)


src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -125,7 +122,7 @@ public void Log(ReportingDescriptor rule, Result result, int? extensionIndex = n
             WriteLineToConsole(GetMessageText(_toolName, physicalLocation?.ArtifactLocation?.Uri, physicalLocation?.Region, result.RuleId, message, result.Kind, result.Level));
         }
 
-        private static string GetMessageText(
+        public static string GetMessageText(
```
michaelcfanning, line 125: >public [](http://example.com/codeflow?start=8&length=6)

I've elevated this helper to public to unblock another change. My sense though is there would be a better api, probably driven by a single result instance that would be better as a public offering.

And so I'm not documenting this change in the release notes. :) I will go back to it.

stacywray (approved):
:shipit:

PR #2748, opened 2023-12-21
End to end diffing

michaelcfanning:
This change permits capturing user-facing output as part of test diffing.

This change isn't ideal/fully realized: it's authored in this haphazard way to permit an upstream submodule consumer to accept it.

After shipping in main, we should push these changes deeper into the core test API (ensuring that *this* repos tests continue to function).

### DIFF
--- a/src/Sarif/EnumeratedArtifact.cs
+++ b/src/Sarif/EnumeratedArtifact.cs
@@ -16,132 +16,56 @@ namespace Microsoft.CodeAnalysis.Sarif
             FileSystem = fileSystem;
         }
 
-        internal byte[] bytes;
         internal string contents;
-
-        private Encoding encoding;
 
         public Uri Uri { get; set; }
 
-        public bool IsBinary
-        {
-            get
-            {
-                GetArtifactData();
-                return this.contents == null;
-            }
-        }
-
         public Stream Stream { get; set; }
 
-        public Encoding Encoding
-        {
-            get
-            {
-                if (encoding == null)
-                {
-                    GetArtifactData();
-                }
-                return this.encoding;
-            }
-
-            set => this.encoding = value;
-        }
+        public Encoding Encoding { get; set; }
 
         internal IFileSystem FileSystem { get; set; }
 
+        public bool IsBinary { get; set; }
+
         public string Contents
         {
-            get => GetArtifactData().text;
+            get => GetContents();
             set => this.contents = value;
         }
 
         public byte[] Bytes
         {
-            get => GetArtifactData().bytes;
-            set => this.bytes = value;
+            get
+            {
+                GetContents();
+                return Encoding.UTF8.GetBytes(this.contents);
+            }
+            set => this.contents = Encoding.UTF8.GetString(value);
         }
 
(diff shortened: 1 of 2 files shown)


src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
```diff
@@ -22,6 +22,8 @@
 
 using Xunit.Abstractions;
 
+using YamlDotNet.Core.Tokens;
```
shaopeng-gh, line 25: is this needed or can be deleted

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -19,131 +19,78 @@ public EnumeratedArtifact(IFileSystem fileSystem)
         internal byte[] bytes;
         internal string contents;
 
-        private Encoding encoding;
```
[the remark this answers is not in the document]
michaelcfanning: Oops, should have cherry-picked this specific commit, fixing.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -19,131 +19,78 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
-                    (Uri.IsAbsoluteUri && !Uri.IsFile))
-                {
-                    throw new InvalidOperationException("An absolute URI pointing to a file location was not available.");
-                }
-
-                // This is our client-side, disk-based file retrieval case.
-                this.Stream = FileSystem.FileOpenRead(Uri.LocalPath);
-            }
+                // TBD we actually have no validation URI is non-null yet.
+                contents = Uri!.IsFile
```
shaopeng-gh, line 51: >IsFile [](http://example.com/codeflow?start=32&length=6)

not sure if applies here,
Uri.IsFile will throws an InvalidOperationException when it is a relative Uri

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -19,131 +19,78 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
 
-            if (Stream.CanSeek)
-            {
-                RetrieveDataFromSeekableStream();
+                this.sizeInBytes = (long?)this.contents?.Length;
             }
             else
             {
-                RetrieveDataFromNonSeekableStream();
+                if (Stream.CanSeek) { this.Stream.Seek(0, SeekOrigin.Begin); }
```
shaopeng-gh, line 59: >Stream [](http://example.com/codeflow?start=20&length=6)

just a note the else part of the if logic itself, 
does not guarantee Stream is not null here

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -19,131 +19,78 @@ public EnumeratedArtifact(IFileSystem fileSystem)
...
+                this.sizeInBytes = Stream.Read(this.bytes, 0, this.bytes.Length);
+                Stream.Close();
+                Stream = null;
             }
+
+            return this.bytes;
         }
 
+
         public long? sizeInBytes;
```
shaopeng-gh, line 94: extra empty line can be removed

PR #2749, opened 2023-12-22
User facing output to file diff

michaelcfanning:
This change retrieves and persists user-facing output generated by functional tests. These strings are to be persisted alongside the generated `.sarif` files in the `Expected` directory in `TestData`. We are experimenting with this capability in one SARIF Driver framework use case. Once it's proven/refined, we should push that code deeper into the test API.

### DIFF
--- a/src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
+++ b/src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 
 using FluentAssertions;
 
@@ -21,6 +22,8 @@ using Newtonsoft.Json.Linq;
 using Newtonsoft.Json.Serialization;
 
 using Xunit.Abstractions;
+
+using YamlDotNet.Core.Tokens;
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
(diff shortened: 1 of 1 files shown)


src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
```diff
@@ -228,15 +229,27 @@ protected virtual string ConstructTestOutputFromInputResource(string inputResour
 
             // Reify the list of keys because we're going to modify the dictionary in place.
             List<string> keys = expectedSarifTextDictionary.Keys.ToList();
+            var userFacingTexts = new Dictionary<string, string>();
 
             foreach (string key in keys)
             {
+                string actualSarifText = actualSarifTextDictionary[key];
+                SarifLog actualSarifLog = JsonConvert.DeserializeObject<SarifLog>(actualSarifText);
+                if (actualSarifLog.Runs?[0].TryGetProperty("consoleOut", out string userFacingText) == true)
```
michaelcfanning, line 238: >TryGetProperty [](http://example.com/codeflow?start=44&length=14)

So here's the trick upstream users of this base class can do. If they'd like, they can populate the run with `consoleOut` to capture user-facing strings. We can (and should) build this deeper into the test infra but it will break everyone's tests (which will require baselining). So we're taking this spotfix for now.

shaopeng-gh:
Will the generated .txt file same when running on different machines, does it contains the part that is the absolute path of the file that is having problem which will be different on the machine

michaelcfanning:
The non-deterministic repo root is normalized to a variable, e.g.:

$(SRC_ROOT)\src\Plugins\Tests\TestData\Example\Inputs\TST101_001.TestRule.ps1(4,43-69): warning STST101/001: '…111111' was flagged as a potential problem.

---
In reply to: [1867931239](https://github.com/microsoft/sarif-sdk/pull/2749#issuecomment-1867931239) [](http://example.com/codeflow?ancestors=1867931239)

shaopeng-gh (approved):
:shipit:

PR #2751, opened 2023-12-29
Allow consumers to output SARIF results with both ruleId and rule.id

### LINKED ISSUE #2721: [REQUEST] Allow users of the SARIF SDK to emit SARIF results with both ruleId and rule.id
The ShouldSerializeRuleId() method in https://github.com/microsoft/sarif-sdk/blob/b8c063659520e72c5bcc96db9f2061378fcb2333/src/Sarif/Core/Result.cs#L20 disallows serializing ruleId when rule.id is present.

For consumers of the SARIF SDK who wish to emit both ruleId and rule.id (I imagine there a few implementations of reading SARIF files only expecting ruleId?), it'd be nice if the ShouldSerializeRuleId() method was virtual. Then a consumer could derive from the Result class and override to always serialize the ruleId property

dotpaul:
Fixes #2721


michaelcfanning (approved):
:shipit:

src/Sarif/Core/Result.cs
```diff
@@ -17,7 +17,7 @@ public partial class Result
         /// </remarks>
         public Run Run { get; set; }
 
-        public bool ShouldSerializeRuleId()
+        public virtual bool ShouldSerializeRuleId()
```
michaelcfanning, line 20: >virtual [](http://example.com/codeflow?start=15&length=7)

Nice change and probably one that we should push everywhere else in the SDK.

michaelcfanning (approved):

PR #2753, opened 2024-01-10
Use new 'PeekableStream' to do binary categorization.

scottoneil-ms:
PeekableStream offers a very weak Seek(Origin) capability that expires once a configurable amount of Stream has been read.


src/Sarif/EnumeratedArtifact.cs
```diff
@@ -87,50 +91,28 @@ public byte[] Bytes
...
-        private void RetrieveDataFromNonSeekableStream()
+        private void RetrieveDataFromStream()
         {
-            bool isText;
-
-            this.bytes = new byte[Stream.Length];
-            int length = this.Stream.Read(this.bytes, 0, this.bytes.Length);
-            isText = FileEncoding.IsTextualData(this.bytes, 0, length);
+            // Reset to beginning of stream in case caller neglected to do so.
+            TryRewindStream();
```
scottoneil-ms, line 104: I would love to remove this call.  It can't be necessary, right?  But I sort of inherited it from the Seekable variant.  Would love to hear I can delete.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -5,12 +5,16 @@
 using System.IO;
 using System.Text;
 
+using Microsoft.CodeAnalysis.Sarif.Readers;
+
```
michaelcfanning, line 9: as discussed, leave at root of `Sarif` namespace

src/Sarif/EnumeratedArtifact.cs
michaelcfanning: Sure, let's zap it.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -87,50 +91,28 @@ public byte[] Bytes
...
-            }
+            RetrieveDataFromStream();
 
             this.Stream = null;
 
             return (this.contents, this.bytes);
         }
 
-        private void RetrieveDataFromNonSeekableStream()
+        private void RetrieveDataFromStream()
```
michaelcfanning, line 101: >RetrieveDataFromStream [](http://example.com/codeflow?start=21&length=22)

Nice abstraction you've delivered! Look how clean and readable this code is now.

src/Test.UnitTests.Sarif/PeekableStreamTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using FluentAssertions;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    public class PeekableStreamTests
+    {
+        [Fact]
+        public void PeekableStream_BoundaryRewinds()
```
michaelcfanning, line 16: >PeekableStream_BoundaryRewinds [](http://example.com/codeflow?start=20&length=30)

So, can we assume that there wasn't a need to produce other directed unit tests because the enumerated artifact class and maybe zip archive already exercise what's interesting? I think that's true.

michaelcfanning:
Release note, please! For this excellent change.

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.CodeAnalysis.Sarif.Readers
+{
+    /// <summary>
+    /// This stream provides a very limited Seekability for Streams which would benefit from an ersatz Seek(Origin)
```
michaelcfanning, line 10: >Seekability [](http://example.com/codeflow?start=44&length=11)

lower-case? NIT

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// capability, but which only need it at the beginning of a Stream.  Ie: to categorize binary versus textual
```
michaelcfanning, line 11: >Ie [](http://example.com/codeflow?start=74&length=2)

s/be rendered as `i.e.` if you prefer this. Also this is a sentence fragment, maybe go ahead and append to the previous sentence after a comma.

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// This stream provides a very limited Seekability for Streams which would benefit from an ersatz Seek(Origin)
+    /// capability, but which only need it at the beginning of a Stream.  Ie: to categorize binary versus textual 
+    /// data.
+    /// </summary>
+    internal class PeekableStream : Stream
+    {
+        private readonly Stream underlyingStream;
+        private byte[] rewindBuffer;
+        private int cursor;
+        public PeekableStream(Stream underlyingStream, int initialPeekWindow)
```
michaelcfanning, line 19: NIT: introduce a line before declaring your ctor to separate from the field decls.

michaelcfanning:
I know the class is internal and we're swapping out for function that should be equivalent. We'll want signal in the release notes of when we made the switch-over though, in case users report an issue in the enumerated artifact class.

---
In reply to: [1885935208](https://github.com/microsoft/sarif-sdk/pull/2753#issuecomment-1885935208) [](http://example.com/codeflow?ancestors=1885935208)

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public void Rewind()
+        {
+            if (this.rewindBuffer == null)
+            {
+                throw new InvalidOperationException("This stream has read too many bytes and can not longer be rewound.");
```
michaelcfanning, line 114: >InvalidOperationException [](http://example.com/codeflow?start=26&length=25)

is there a unit test for this condition?

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                // We are rewound.  Service the read out of the rewind buffer.
+                int safeReadCount = Math.Min(count, this.rewindBuffer.Length - cursor);
+                safeReadCount = Math.Min(safeReadCount, (int)(this.underlyingStream.Position - cursor));
+                Array.Copy(this.rewindBuffer, cursor, buffer, offset, safeReadCount);
+
+                cursor += safeReadCount;
+                return safeReadCount;
+            }
+            else
```
michaelcfanning, line 87: >else [](http://example.com/codeflow?start=12&length=4)

NIT, I would tend to elise this `else` statement, just leaving the throw to close the function. IMO, you don't need an // unreachable comment when you handle it this way. The last statement is a throw, pretty clear.

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+                // Buffer cursor is "caught up" to stream position, so we are not in a "rewound" state.
+                if (this.cursor == this.rewindBuffer.Length)
+                {
+                    // A previous read has brought us to the brink of exceeding the "peekability window", and we are
+                    // now reading more.  So we abandon the buffer, and serve the read as a thin wrapper.
+                    this.rewindBuffer = null;
+                    this.cursor = int.MaxValue;
+                    return underlyingStream.Read(buffer, offset, count);
+                }
+                else
```
michaelcfanning, line 64: >else [](http://example.com/codeflow?start=16&length=4)

Again, I'd elide this `else`. when your previous if statement returns, just get rid of it. it makes it *much* clearer that there's no possibility of a sneaky return at the end of this `else` clause. Also, doing this left shifts a significant chunk of code, and less nesting == more readable, again IMO

src/Sarif/PeekableStream.cs
scottoneil-ms: Not yet.

src/Sarif/PeekableStream.cs
```diff
@@ -0,0 +1,129 @@
...
+            get => Math.Min(underlyingStream.Position, cursor);
+            set => throw new System.NotImplementedException();
+        }
+
+        public override void Flush()
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
```
michaelcfanning, line 46: >Read [](http://example.com/codeflow?start=28&length=4)

ok! I think I understand. your peekable stream is kinda like a pocket knife with a 'stop' before it fully opens. users can read the header bytes and you will 'stop' when you hit the transition point that will require you to abandon the rewind buffer. this gives the caller the change to easily fully inspect the header without worrying about an over-read. At this point, either the caller presumably rewinds (to get the entire stream by passing it to another reader) *OR* moves forward.

Two things: 

1) this is weird, but for readability, I might have ordered this code to reflect these three states reflecting where we are in the stream, 1) within the rewind buffer, 2) at the transition point, 3) past the transition point.

Simple because this makes it easier to build a mental model of what's going on by reading your commented code. 

Second thing, consider a large comment at the top that describes the general approach. For fun, ask Copilot to generate it. :) haha.

michaelcfanning (approved):
:shipit:

src/Test.UnitTests.Sarif/PeekableStreamTests.cs
scottoneil-ms: I think this covers a lot, but I've thought of some other cases to hit and you added one as well.

src/Sarif/PeekableStream.cs
scottoneil-ms: The "stop" point is more an implementation detail to make implementation easier than an API feature.  The caller is in control of the buffer size.  That's 100% of what they need to know to use the class correctly.  I would not want them forming dependencies on the partial read behavior.

What the partial reads do, is let me operate over one byte range per Read() call.  That simplifies things a bit.  It would be harder to read and get correct (I claim) if I got hung up on delivering all the requested bytes, and had to compose both an internal buffer read and a call into the underlying stream, in the same call.

PR #2757, opened 2024-01-12
Add support for throwing HttpRequestException("No such host is known") in HttpMockHelper

LingZhou-gh:
Add support for throwing HttpRequestException("No such host is known") in HttpMockHelper.
No release note added for this test change.


LingZhou-gh:
It turns out this change is not needed anymore.

PR #2759, opened 2024-01-18
Demote nuget security advisories to warnings.

scottoneil-ms:
We have warnings-as-errors enabled for most of our repositories because it keeps out new warnings/regressions.  This doesn't work for CVEs, which are exogenous.  These should be treated as warnings and not break the build or create work stoppages.


michaelcfanning (approved):
:shipit:

PR #2762, opened 2024-01-19
Do not send empty log

shaopeng-gh:
* NEW: Command-line argument `--post-uri` will now skip file posting step if the result is empty.

Also found a bug and fixed so that: Even if there are fatal errors, if the log file is generated, we can upload it with the ToolExecutionNotifications.

User facing strings:

image

image

Coverage
image

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -8,7 +8,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
-* NEW: Command-line argument `--post-uri` will now skip file posting step if the result is empty.
+* NEW: `--post-uri` will skip sending the SARIF log to the configured endpoint if the file contains no results or fatal execution errors.
 
 ## **v4.4.1 UNRELEASED
 * DEP: Update reference to `System.Collections.Immutable` 5.0.0 for `Sarif` and `Sarif.Converters`.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -7,6 +7,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
+* NEW: Command-line argument `--post-uri` will now skip file posting step if the result is empty.
 
 ## **v4.4.1 UNRELEASED
 * DEP: Update reference to `System.Collections.Immutable` 5.0.0 for `Sarif` and `Sarif.Converters`.
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -225,7 +225,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
         protected virtual void PostLogFile(IAnalysisContext globalContext)
         {
-            if (string.IsNullOrEmpty(globalContext.PostUri)) { return; }
+            if (string.IsNullOrWhiteSpace(globalContext.PostUri) ||
+                string.IsNullOrWhiteSpace(globalContext.OutputFilePath) ||
+                !globalContext.FileSystem.FileExists(globalContext.OutputFilePath))
+            {
+                return;
+            }
 
             try
             {
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -231,11 +231,18 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             {
                 using (var httpClient = new HttpClient())
                 {
-                    PostLogFile(globalContext.PostUri, globalContext.OutputFilePath, globalContext.FileSystem, httpClient)
+                    bool logFilePosted = PostLogFile(globalContext.PostUri, globalContext.OutputFilePath, globalContext.FileSystem, httpClient)
                         .GetAwaiter()
                         .GetResult();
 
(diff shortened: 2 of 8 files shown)


stacywray (approved):
:shipit:

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -249,14 +256,14 @@ protected virtual void PostLogFile(IAnalysisContext globalContext)
...
+        internal static async Task<bool> PostLogFile(string postUri, string outputFilePath, IFileSystem fileSystem, HttpClient httpClient)
         {
             if (string.IsNullOrWhiteSpace(postUri))
             {
-                return;
+                return false;
             }
 
-            await SarifLog.Post(new Uri(postUri), outputFilePath, fileSystem, httpClient);
+            return await SarifLog.Post(new Uri(postUri), outputFilePath, fileSystem, httpClient);
```
scottoneil-ms, line 271: I think this is correct right now, but it's a maintenance hazard.  If someone reasonably adds another reason to return false below, this line will continue reporting empty log files that might not be empty.

Can you move the Console output down to where the post-or-not decision is being made?

scottoneil-ms: Oh, if you do this, I think you can remove the return code, which simplifies things nicely.

src/Sarif/Errors.cs
```diff
@@ -697,5 +698,27 @@ internal static void LogAnalysisTimedOut(IAnalysisContext context)
...
         }
+
+        public static void LogErrorPostingLogFile(IAnalysisContext context, string postUri)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
```
scottoneil-ms, line 709: Move above the nullcheck?

shaopeng-gh: can not, context.RuntimeErrors requires context not null

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
shaopeng-gh: Console output down to where the post-or-not decision is being made ---- sounds good, fixed

src/Sarif/Core/SarifLog.cs
```diff
@@ -91,8 +92,17 @@ public static SarifLog Load(Stream source, bool deferred = false)
                 throw new ArgumentException($"File path does not exist: '{filePath}'", nameof(filePath));
             }
 
-            using Stream fileStream = fileSystem.FileOpenRead(filePath);
-            await Post(postUri, fileStream, httpClient);
+            byte[] fileBytes = fileSystem.FileReadAllBytes(filePath);
```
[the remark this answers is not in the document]
shaopeng-gh: the empty SARIF was a shorten saying of SARIF with with 0 results, the file structure will be there so the file will be the size of a SARIF file with all nodes other than results node. the all other nodes size can vary depend on what user put, so there is not a good size to compare to.
We could have a check for e.g. size > 20 bytes, but the step of upload file is after the SARIF generated by our tool and our tool does generate valid files so the check may not be necessary.

scottoneil-ms (approved):

scottoneil-ms (approved):

src/Sarif/Core/SarifLog.cs
```diff
@@ -237,5 +248,49 @@ internal static void ComputePolicies(Dictionary<string, FailureLevel> localCache
...
+                    if (run.Results?.Count > 0)
+                    {
+                        return true;
+                    }
+
+                    if (run.Invocations?.Count > 0)
+                    {
+                        foreach (Invocation invocation in run.Invocations)
+                        {
+                            if (invocation.ToolConfigurationNotifications?.Count > 0)
```
michaelcfanning, line 267: >ToolConfigurationNotifications [](http://example.com/codeflow?start=43&length=30)

remove this block. a tool configuration notification is not something we need to receive: this condition indicates that the person running the tool has a problem.

make sense?

1) we receive results because those are findings that will be routed to the code owners
2) we receive tool execution notifications because those will be routed to the tool owner to fix.

we do not receive tool configuration notifications because those are to be resolved by the automation/person running the tool. presumably this can be fixed on their side without a tool change.

ReleaseHistory.md
```diff
@@ -8,6 +8,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
+* NEW: Command-line argument `--post-uri` will now skip file posting step if the result is empty.
```
michaelcfanning, line 11: > Command-line argument `--post-uri` will now skip file posting step if the result is empty. [](http://example.com/codeflow?start=6&length=91)

Please use this:

`--post-uri` will skip sending the SARIF log to the configured endpoint if the file contains no results or fatal execution errors.

src/Sarif/Core/SarifLog.cs
```diff
@@ -91,8 +92,18 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
             }
 
-            using Stream fileStream = fileSystem.FileOpenRead(filePath);
-            await Post(postUri, fileStream, httpClient);
+            byte[] fileBytes = fileSystem.FileReadAllBytes(filePath);
+            SarifLog sarifLog = Load(new MemoryStream(fileBytes));
+
+            if (!ShouldSendLog(sarifLog))
+            {
+                Console.WriteLine($"Posting of the log file to {postUri} was skipped because the result is empty.");
```
michaelcfanning, line 100: >Posting of the log file to {postUri} was skipped because the result is empty. [](http://example.com/codeflow?start=36&length=77)

Change to :

"Post of log file to {postUri} was skipped because the log contains no results."

src/Sarif/Core/SarifLog.cs
shaopeng-gh: Got it. Wasn't sure if we want the log for troubleshooting config issues. Removed.

michaelcfanning (approved):
:shipit:

PR #2764, opened 2024-01-22
Add `IsBinary` property to `IEnumeratedArtifact`

yongyan-gh:
Implement the `IsBinary` property for `ZipArchiveArtifact`

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -8,6 +8,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
+* NEW: Add `IsBinary` property to `IEnumeratedArtifact` and implement the property in `ZipArchiveArtifact`.
 
 ## **v4.4.1 UNRELEASED
 * DEP: Update reference to `System.Collections.Immutable` 5.0.0 for `Sarif` and `Sarif.Converters`.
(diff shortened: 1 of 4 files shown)


ReleaseHistory.md
```diff
@@ -8,6 +8,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
+* NEW: Add `IsBinary` property to `IEnumeratedArtifact` and implement the property in `ZipArchiveArtifact`.
```
stacywray, line 11: What does this new feature "now" do? See other NEW entries for the standard.

src/Sarif/ZipArchiveArtifact.cs
```diff
@@ -30,6 +30,15 @@ public ZipArchiveArtifact(ZipArchive archive, ZipArchiveEntry entry, ISet<string
...
+        public bool IsBinary
+        {
+            get
+            {
+                string extension = Path.GetExtension(Uri.ToString());
+                return this.binaryExtensions.Contains(extension);
+            }
+        }
+
         public Stream Stream
```
scottoneil-ms, line 42: I don't know why we're splitting this change into an IsBinary implementation based on extension, followed up by a Peekable-backed check in a separate change.  That's not an objection.  I'm just worried I don't see a risk others see.  Is there a reason why ZipArchiveArtifact.IsBinary can't follow the pattern in EnumeratedArtifact, and immediately call GetArtifactData()?

yongyan-gh: ye the final goal should be using the peakable stream for binary sniffing. this is just a quick fix to release the bug fix to unblock for ADO service team. Will create work item for it.

ReleaseHistory.md
yongyan-gh: I think it should not be a new feature. maybe a BUG fix?

src/Sarif/ZipArchiveArtifact.cs
scottoneil-ms: Thanks for that.  If you can move it to active so it's obvious it's at the front of my plate, I'd appreciate it.

scottoneil-ms: Also, thanks for the clarification on the motivation for the sequencing.   This makes a lot of sense to me as a QFE.  Part of me wondered why I shouldn't be offering to quickly add the Peekable stuff into this change.  It shouldn't be hard, but it would not be as safe of a change as this conservative approach.

yongyan-gh: User Story 2142116: Update `ZipArchiveArtifact` to reuse peakable steam implementation in `EnumeratedArtifact`

michaelcfanning (approved):
:shipit:

PR #2765, opened 2024-01-22
Explicitly add related package to avoid version downgraded build error.

yongyan-gh:
# Description

In order to keep SDK retains use of the lowest compatible version, the version of `System.Text.Encoding.CodePages` package used in `Sarif` project is 4.3.0. It caused build errors in `Sarif.Driver` and `Sarif.Multitool` because couple packages version downgraded detected.

The fix is to add lowest required version of those packages explicitly in `Sarif` project.


LingZhou-gh (approved):
:shipit:

michaelcfanning (approved):

PR #2767, opened 2024-01-24
Fix unsupported exception when checking the Length property on Streams created by ZipArchiveArtifact

scottoneil-ms:


yongyan-gh (approved):
looks good! thanks for the fix!

PR #2768, opened 2024-01-25
Fix issue `MultithreadedAnalyzeCommandBase` does not return rich return code

yongyan-gh:


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -146,7 +146,7 @@ public virtual int Run(TOptions options, ref TContext globalContext)
             }
 
             return
-                globalContext.RichReturnCode == true
+                options.RichReturnCode == true
```
yongyan-gh, line 149: >options [](http://example.com/codeflow?start=16&length=7)

the globalContext value is populated  in line 78 `InitializeGlobalContextFromOptions`.

if exception was thrown before the RichReturnCode was set, the original code will not return rich return code even `--rich-return-code` option is set.

found by one of test cases.

scottoneil-ms: I'm looking at `InitializeGlobalContextFromOptions`, and I'm not seeing where globalContext.RichReturnCode is being set at all.  Is it possible that the problem here is that we simply never set the globalContext property in that function?

If the problem is really that we're throwing in there, then that raises a large question of which _other_ GlobalContext properties aren't being set properly because of that selfsame Exception.

scottoneil-ms: I synced with Yong on this offline.  This is a deceptively complicated case, with multiple issues at play.

I think this is a defensible place to land, so I merged it.  But we tugged on the threads a bit more after it went in, and we think we see a small change that will resolve the tensions better.  We're planning on taking another pass at it tomorrow.

ReleaseHistory.md
```diff
@@ -7,6 +7,7 @@
 * DEP: Update `System.Data.SqlClient` reference from 4.8.5 to 4.8.6 in `WorkItems` to resolve [CVE-2024-0056](https://github.com/advisories/GHSA-98g6-xh36-x2p7).
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * BUG: Fix `UnsupportedOperationException` in `ZipArchiveArtifact`.
+* BUG: Fix issue the `MultithreadedAnalyzeCommandBase` does not return rich return code with the `--rich-return-code` option.
```
stacywray, line 10: Suggested rewrite: Fix the `MultithreadedAnalyzeCommandBase` issue to return rich return code with the `--rich-return-code` option.

PR #2769, opened 2024-01-25
Restructuring initialization handling to harden, redirecting RichReturnCode to GlobalContext

scottoneil-ms:
Discussion about previous change (which breaks a pattern in a small way) revealed a larger problem around config initialization.  This is a modest change to how we do config init that hardens all the easy cases without changing error behavior for any hard cases.  It is enough to ensure config state that is essential for basic IO expectations even in failure cases (ie: RichReturnCode) is respected even in failure scenarios where we might not do much in the way of scanning.


yongyan-gh (approved):
:shipit:

PR #2771, opened 2024-01-29
Post the response from the endpoint when log file posting succeed or failed

shaopeng-gh:
The response from the endpoint is useful for troubleshooting.

### DIFF
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -102,19 +102,15 @@ namespace Microsoft.CodeAnalysis.Sarif
             }
 
             HttpResponseMessage response = await Post(postUri, new MemoryStream(fileBytes), httpClient);
-            string responseText = response.Content.ReadAsStringAsync().Result;
+            string responseText = await response.Content.ReadAsStringAsync();
 
-            try
+            if (!response.IsSuccessStatusCode)
             {
-                response.EnsureSuccessStatusCode();
-            }
-            catch (Exception)
-            {
-                Console.WriteLine($"Error Posting log file to: {postUri}. Endpoint provided message: {responseText}");
+                Console.WriteLine($"Error Posting log file to: {postUri}. Endpoint provided status code {response.StatusCode} and message: {responseText}");
                 return false;
             }
 
-            Console.WriteLine($"Posted log file successfully to: {postUri}. Endpoint provided message: {responseText}");
+            Console.WriteLine($"Posted log file successfully to: {postUri}. Endpoint provided status code {response.StatusCode} and message: {responseText}");
             return true;
         }
 
--- a/src/Sarif/Core/SarifLog.cs
+++ b/src/Sarif/Core/SarifLog.cs
@@ -76,7 +76,7 @@ namespace Microsoft.CodeAnalysis.Sarif
         /// <param name="filePath"></param>
         /// <param name="fileSystem"></param>
         /// <param name="httpClient"></param>
-        /// <returns>If the SarifLog has been posted.</returns>
+        /// <returns>If the SarifLog has been posted successfully.</returns>
         public static async Task<bool> Post(Uri postUri,
                                       string filePath,
                                       IFileSystem fileSystem,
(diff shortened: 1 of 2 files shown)


src/Sarif/Core/SarifLog.cs
```diff
@@ -101,8 +101,20 @@ public static SarifLog Load(Stream source, bool deferred = false)
                 return false;
             }
 
-            await Post(postUri, new MemoryStream(fileBytes), httpClient);
-            Console.WriteLine($"Posted log file successfully to: {postUri}");
+            HttpResponseMessage response = await Post(postUri, new MemoryStream(fileBytes), httpClient);
```
shaopeng-gh, line 104: copied the code section,
but made some changes to the message, please review.

src/Sarif/Core/SarifLog.cs
```diff
@@ -101,8 +101,20 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+            string responseText = response.Content.ReadAsStringAsync().Result;
```
scottoneil-ms, line 105: This method is actually async, so you can await this Task instead of doing .Result.  (And that is better.)

src/Sarif/Core/SarifLog.cs
```diff
@@ -101,8 +101,20 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+            try
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            catch (Exception)
+            {
+                Console.WriteLine($"Error Posting log file to: {postUri}. Endpoint provided message: {responseText}");
+                return false;
+            }
+
```
scottoneil-ms, line 116: It seems better to check response.IsSuccessStatusCode.

scottoneil-ms: That is, as opposed to doing a try/catch here.

src/Sarif/Core/SarifLog.cs
```diff
@@ -101,8 +101,20 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+            Console.WriteLine($"Posted log file successfully to: {postUri}. Endpoint provided message: {responseText}");
             return true;
```
scottoneil-ms, line 118: I'm not sure what the format of the responseText is going to be and what it contains.  Is it possible to run this locally and provide a sample?

src/Sarif/Core/SarifLog.cs
shaopeng-gh: And I also think status code is useful to user, I think we should add it to the output as well

src/Sarif/Core/SarifLog.cs
shaopeng-gh: This is generic, 
the server that takes the SARIF file is to be set by the user,
and that server could be with any design by the user. 
It can returns anything, e.g. correlation id.

stacywray (approved):
:shipit:

src/Sarif/Core/SarifLog.cs
```diff
@@ -101,8 +101,16 @@ public static SarifLog Load(Stream source, bool deferred = false)
...
+            HttpResponseMessage response = await Post(postUri, new MemoryStream(fileBytes), httpClient);
+            string responseText = await response.Content.ReadAsStringAsync();
+
+            if (!response.IsSuccessStatusCode)
+            {
+                Console.WriteLine($"Error Posting log file to: {postUri}. Endpoint provided status code {response.StatusCode} and message: {responseText}");
+                return false;
+            }
+
+            Console.WriteLine($"Posted log file successfully to: {postUri}. Endpoint provided status code {response.StatusCode} and message: {responseText}");
```
scottoneil-ms, line 113: Thanks.  I'm realizing we should revisit this code soon.  We probably shouldn't be writing directly to the Console, and instead refactor the APIs at this layer and a few more up above to pass a Context down, so we can access a proper logger.   I'll leave it for Michael to decide if we need that in this change.  I can see cases in both directions.

scottoneil-ms (approved):
:shipit:

scottoneil-ms (approved):

PR #2772, opened 2024-01-31
Improve testing for post log. Do not emit post log messages when '--quiet' is enabled.

michaelcfanning:
* BRK: Change return value of `SarifLog.Post` from `bool` to `(bool, string)` to allow return of status code and content of post response.

### DIFF
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -256,7 +256,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             catch (Exception ex)
             {
                 Console.WriteLine(ex.ToString());
-                globalContext.RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
+                Errors.LogErrorPostingLogFile(globalContext, globalContext.PostUri);
+                globalContext.RuntimeExceptions ??= new List<Exception>();
+                globalContext.RuntimeExceptions.Add(ex);
                 throw new ExitApplicationException<ExitReason>(DriverResources.MSG_UnexpectedApplicationExit, ex)
                 {
                     ExitReason = ExitReason.ExceptionPostingLogFile
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -253,6 +253,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 {
                     Console.WriteLine(result.Item2);
                 }
+
+                if (result.Item1 == false && !result.Item2.Contains("skipped"))
+                {
+                    globalContext.RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
+                }
             }
             catch (Exception ex)
             {
--- a/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
+++ b/src/Sarif.Driver/Sdk/PluginDriverCommand.cs
@@ -220,6 +220,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 };
             }
         }
+
         protected virtual void PostLogFile(IAnalysisContext globalContext)
         {
             using var httpClient = new HttpClientWrapper();
@@ -256,7 +257,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             catch (Exception ex)
             {
                 Console.WriteLine(ex.ToString());
-                Errors.LogErrorPostingLogFile(globalContext, globalContext.PostUri);
+                globalContext.RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
                 globalContext.RuntimeExceptions ??= new List<Exception>();
                 globalContext.RuntimeExceptions.Add(ex);
                 throw new ExitApplicationException<ExitReason>(DriverResources.MSG_UnexpectedApplicationExit, ex)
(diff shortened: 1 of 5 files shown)


src/Test.UnitTests.Sarif/Core/SarifLogTests.cs
```diff
@@ -276,59 +276,68 @@ public async Task SarifLog_PostFile_PostTests()
...
-            logPosted = await SarifLog.Post(postUri: postUri,
-                                    filePath,
-                                    fileSystem.Object,
-                                    httpClient: new HttpClient(httpMock));
-            logPosted.Should().BeTrue("with error level ToolExecutionNotifications, but the server returns BadRequest");
+            logPosted = await SarifLog.Post(postUri,
+                                            filePath,
+                                            fileSystem.Object,
+                                            httpClient: new HttpClient(httpMock));
+            logPosted.Item1.Should().BeFalse("the server returns a BadRequest even though there are error level ToolExecutionNotifications");
```
shaopeng-gh, line 339: >the server returns a BadRequest even though there are error level ToolExecutionNotifications [](http://example.com/codeflow?start=46&length=92)

the new message will be like:
`Expected logPosted.Item1 to be false because the server returns a BadRequest even though there are error level ToolExecutionNotifications, but found True.`

michaelcfanning: Try this:

"a status code of 'BadRequest' should fail the call even though we posted error-level 'ToolExcecutionNotifications'".

For a fully constructed line like so:

`Expected logPosted.Item1 to be false because a status code of 'BadRequest' should fail the call even though we posted error-level 'ToolExcecutionNotifications', but found True.`

shaopeng-gh: Look better! Updated.

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
```diff
@@ -60,26 +60,93 @@ public void PluginDriverCommand_ValidateInvocationPropertiesToLog_ShouldReturnFa
...
-            exception.Should().BeOfType(typeof(ArgumentNullException));
+                PostUri = postUri,
+                OutputFilePath = outputFilePath,
+            };
+
+            Exception exception = Record.Exception(() =>
+                    PluginDriverCommand<AnalyzeOptionsBase>.SarifPost(context,
+                                                                      mockHttpClient.Object));
+
+            exception.Should().BeOfType(typeof(ArgumentException));
```
michaelcfanning, line 149: >ArgumentException [](http://example.com/codeflow?start=47&length=17)

why is this raising argument exception?

this doesn't conform to the name of this unit test.

where's the ExitApplicationException?

also, after this call we would expect to see certain runtime conditions on the context.RuntimeErrors property and also the InvalidOperationException in the RuntimeExceptions object.

can you please add these check?

shaopeng-gh, line 149: >ArgumentException [](http://example.com/codeflow?start=47&length=17)

I changed this to make sure if all tests can pass, but will need to check if this is as expected, or a bug.

shaopeng-gh: changed a few issues,
1. `HttpClientWrapper` requires a default constructor  
`HttpClientWrapper(HttpClient httpClient = null)` will not do.
2. missing mock FileSystem will throw error
3. missing context.Logger will throw error 
4. missing results, will skip instead of getting exceptions.
5. added the asserts you mentioned.

src/Sarif.Driver/Sdk/PluginDriverCommand.cs
```diff
@@ -234,32 +238,38 @@ protected virtual void PostLogFile(IAnalysisContext globalContext)
...
-                    PostLogFile(globalContext.PostUri, globalContext.OutputFilePath, globalContext.FileSystem, httpClient)
-                        .GetAwaiter()
-                        .GetResult();
+                    Console.WriteLine(result.Item2);
+                }
+
+                if (result.Item1 == false && !result.Item2.Contains("skipped"))
+                {
+                    globalContext.RuntimeErrors |= RuntimeConditions.ExceptionPostingLogFile;
                 }
```
shaopeng-gh, line 260: After I added `RuntimeExceptions` below when there is unhandled Exception, I realized a question, 
here when there is no unhandled Exception but the server is not "working", should we set `RuntimeErrors` without setting `RuntimeExceptions` ? I use the last commit for this change, if not we can just revert the commit 7.

src/Test.UnitTests.Sarif.Driver/Sdk/PluginDriverCommandTests.cs
```diff
@@ -60,26 +62,125 @@ public void PluginDriverCommand_ValidateInvocationPropertiesToLog_ShouldReturnFa
...
+                PostUri = null,
+                OutputFilePath = outputFilePath,
+                FileSystem = mockFileSystem.Object
+            };
+
+            PluginDriverCommand<AnalyzeOptionsBase>.SarifPost(context, httpClient);
+        }
+
+        [Fact]
+        public void PluginDriverCommand_ThrowsExitApplicationExceptionOnUnhandledException()
```
michaelcfanning: >PluginDriverCommand_ThrowsExitApplicationExceptionOnUnhandledException [](http://example.com/codeflow?start=20&length=70)

Excellent test! Well done!

PR #2775, opened 2024-02-07
New AdvSec & GHAS - Enable using config file

shaopeng-gh:
@EasyRhinoMSFT 
This is to enable using config file to change settings for New AdvSec & GHAS work.

### DIFF
--- a/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
+++ b/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
@@ -319,7 +319,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 context.TargetFileSpecifiers = options.TargetFileSpecifiers?.Any() == true ? InitializeStringSet(options.TargetFileSpecifiers) : context.TargetFileSpecifiers;
                 context.InvocationPropertiesToLog = options.InvocationPropertiesToLog?.Any() == true ? InitializeStringSet(options.InvocationPropertiesToLog) : context.InvocationPropertiesToLog;
                 context.Traces = options.Trace.Any() ? InitializeStringSet(options.Trace) : context.Traces;
-                context.RuleKinds = options.RuleKinds;
+                context.RuleKinds = options.RuleKindOption != null ? options.RuleKinds : context.RuleKinds;
             }
 
             // Less-obviously throw-safe. We don't do these in the finally block because we'd prefer not to mask an earlier Exception during logger initialization. 
--- a/src/Sarif/RuleKindSet.cs
+++ b/src/Sarif/RuleKindSet.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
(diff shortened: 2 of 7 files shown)


src/Sarif/RuleKindSet.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif
+{
+    [Serializable]
+    [JsonConverter(typeof(TypedPropertiesDictionaryConverter))]
+    public class RuleKindSet : HashSet<RuleKind>
```
EasyRhinoMSFT, line 14: >RuleKindSet [](http://example.com/codeflow?start=17&length=11)

Why couldn't we use a HashSet?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -319,7 +319,7 @@ public virtual TContext InitializeGlobalContextFromOptions(TOptions options, ref
                 context.TargetFileSpecifiers = options.TargetFileSpecifiers?.Any() == true ? InitializeStringSet(options.TargetFileSpecifiers) : context.TargetFileSpecifiers;
                 context.InvocationPropertiesToLog = options.InvocationPropertiesToLog?.Any() == true ? InitializeStringSet(options.InvocationPropertiesToLog) : context.InvocationPropertiesToLog;
                 context.Traces = options.Trace.Any() ? InitializeStringSet(options.Trace) : context.Traces;
-                context.RuleKinds = options.RuleKinds;
+                context.RuleKinds = options.RuleKindOption != null ? options.RuleKinds : context.RuleKinds;
```
EasyRhinoMSFT, line 322: >options.RuleKindOption != null ? options.RuleKinds : context.RuleKinds [](http://example.com/codeflow?start=36&length=70)

Can we use this simpler syntax?

options.RuleKindOption ?? context.RuleKinds

shaopeng-gh: no we can not,
options.RuleKindOption is not the same type as options.RuleKinds

src/Sarif/RuleKindSet.cs
shaopeng-gh: this is following the same pattern as other fields we have, the ResultKind.
The commanline nuget package the options have some limitation so there is some conversions needed

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
shaopeng-gh: if you see lines above
`context.ResultKinds = options.Kind != null ? options.ResultKinds : context.ResultKinds;`
Some tricky here, the `options.Kind` is not the same type as `options.ResultKinds`.

p.s. I think the naming before is not good for `Kind` for `ResultKinds` , it is better to name using xxxOption, or some pattern. People will not know `Kind` is the command line option for `ResultKinds`

EasyRhinoMSFT (approved):
:shipit:

PR #2777, opened 2024-02-09
fix: restore Post's public API accepting HttpClient

rwoll:
Fixes the error:

```
error CS1503: Argument 3: cannot convert from 'System.Net.Http.HttpClient' to 'Microsoft.CodeAnalysis.Sarif.HttpClientWrapper'
```

which was introduced by an inadvertant public signature change in https://github.com/microsoft/sarif-sdk/pull/2772

NB: The return type change from `Task<bool>` to `Task<(bool, string)>` was intentional and is left in tact.


shaopeng-gh (approved):
:shipit:

scottoneil-ms (approved):

PR #2780, opened 2024-02-09
Fixing textual classifier's handling of misaligned data

scottoneil-ms:
The classifier attempts to decode an input buffer as both an Uft32 and Utf16.  It takes a count of "characters" to read.  In practice, this seems to be bytes to read, and callers tend to pass in the buffer length.  This works fine when the byte[] is sized to an array of length equivalent to zero modulo four (the number of bytes in a UTF32 buffer.)  But when the alignments are less favorable, it is possible for one encoding to return a false-binary detection, and the other to partially read off the end of the buffer in a way that fails the decoding.

Proposed fix is to have the classifier ensure the decoding is aligned to zero modulo 4.

### DIFF
--- a/src/Test.UnitTests.Sarif/FileEncodingTests.cs
+++ b/src/Test.UnitTests.Sarif/FileEncodingTests.cs
@@ -89,12 +89,19 @@ namespace Microsoft.CodeAnalysis.Sarif
             using var assertionScope = new AssertionScope();
 
             var sb = new StringBuilder();
-            string unicodeText = "американец";
-
-            foreach (Encoding encoding in new[] { Encoding.Unicode, Encoding.UTF8, Encoding.BigEndianUnicode, Encoding.UTF32 })
+            string[] unicodeTexts = new[] 
             {
-                byte[] input = encoding.GetBytes(unicodeText);
-                FileEncoding.IsTextualData(input).Should().BeTrue(because: $"'{unicodeText}' encoded as '{encoding.EncodingName}' should not be classified as binary data");
+                "американец",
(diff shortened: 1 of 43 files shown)


src/Test.UnitTests.Sarif/FileEncodingTests.cs
```diff
@@ -89,12 +89,19 @@ public void FileEncoding_UnicodeDataIsTextual()
             using var assertionScope = new AssertionScope();
 
             var sb = new StringBuilder();
-            string unicodeText = "американец";
+            string[] unicodeTexts = new[]
+            {
+                "американец",
+                "Generates a warning and an error for each of :  foo  foo \r\n" // Challenging for the classifer; found by accident.
```
scottoneil-ms, line 95: \nNULL seems to break decoding as text for UTF16.  I did not fully isolate which part of the string failed us in UTF32 decoding, but I do know the classifier determines this thing to be binary for UTF32 due to the check on line 69 (new code), which is a block that I doubt we're hitting in any other way.

michaelcfanning: Not sure what you mean. First is this data '\n\0' that you are referring to?

Second, any input fewer than 4 bytes will result in 0 characters read from the content, if it comes to it, correct?

I think I may be confused, reach out offline if it's faster.

michaelcfanning: ok! Now I understanding, you are referring to the very final chars of the test string. :) Oops.

PR #2783, opened 2024-02-12
Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character. Scan file with URL-decodable name without throwing exceptions.

LingZhou-gh:
When the path of the scan target contains invalid character, the exception 'ArgumentException: Illegal characters in path' was raised for 100++ rule checks, as reported in a bug.

This fix check the path of the scan target before invoking any rule skimmer. If the path contains invalid character or URL encoded character, return `WRN997_InvalidTarget`.

Unit tests are added/updated.

====More detailed description===
In Microsoft.CodeAnalysis.Sarif.RuntimeConditions enum, there are at least 2 different target-related condition:
Fatal one (**Error**):        NoValidAnalysisTargets
Non-fatal (**Warning**): TargetNotValidToAnalyze
If we have 2 files to scan, one is a **normal** file, another is a file with **weird** filename (invalid char or contains %2F)
We won't see NoValidAnalysisTargets, as we do have a valid target.

But when file with weird name continues to flow through the code, it might cause trouble somewhere.
1) if the filename contains URL encoded character, such as "%2F", when URI is made out of the filename. **The filename changed!**, which means it won't be located later.
Eg: Incoming-email-not-synchronized-%2D-%22Ignoring-item%22.md will be changed to Incoming-email-not-synchronized---"Ignoring-item".md
 
2) if the filename contains invalid character as defined by **System.IO.Path.GetInvalidPathChars,** it could throw exception as documented by [Path.GetExtension Method (System.IO) | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getextension?view=net-8.0)

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,7 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.5.1 UNRELEASED
-* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid characters.
-* BUG: When the path of the scan target contains URL Percent-encoding characters, `uri.GetFilePath` should return the original string.
+* BUG: When the path of the scan target contains invalid characters, report `WRN997_InvalidTarget` before analyzing it against any rule, otherwise `ArgumentException: Illegal characters in path` will be thrown.
+* BUG: When the path of the scan target contains URL Percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.
 
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
 * DEP: Downgrade `System.Text.Encoding.CodePages` from 8.0.0 to 4.3.0 in `Sarif`.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,6 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.5.1 UNRELEASED
-* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character or encoded character.
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid characters or encoded characters.
 
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
 * DEP: Downgrade `System.Text.Encoding.CodePages` from 8.0.0 to 4.3.0 in `Sarif`.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,6 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.5.1 UNRELEASED
-* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid characters or encoded characters.
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid characters.
+* BUG: When the path of the scan target contains URL Percent-encoding characters, `uri.GetFilePath` should return the original string.
 
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
 * DEP: Downgrade `System.Text.Encoding.CodePages` from 8.0.0 to 4.3.0 in `Sarif`.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.5.1 UNRELEASED
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character or encoded character.
+
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
 * DEP: Downgrade `System.Text.Encoding.CodePages` from 8.0.0 to 4.3.0 in `Sarif`.
 * DEP: Remove explicit versioning for `System.Memory` and `System.Runtime.CompilerServices.Unsafe`.
@@ -9,7 +12,6 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * BUG: Fix `UnsupportedOperationException` in `ZipArchiveArtifact`.
 * BUG: Fix `MultithreadedAnalyzeCommandBase` to return rich return code with the `--rich-return-code` option.
-* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character or encoded character.
 * NEW: Add `IsBinary` property to `IEnumeratedArtifact` and implement the property in `ZipArchiveArtifact`.
 * NEW: Switch to content-based `IsBinary` categorization for `ZipArchiveArtifact`s.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,6 +1,6 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.5.3 UNRELEASED
-* BUG: When the path of the scan target contains invalid characters, report `WRN997_InvalidTarget` before analyzing it against any rule, otherwise `ArgumentException: Illegal characters in path` will be thrown.
+* BUG: When the path of the scan target contains invalid characters, report `WRN997_InvalidTarget` before analyzing it against any rule to avoid `ArgumentException: Illegal characters in path`.
 * BUG: When the path of the scan target contains URL percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.
 
 ## **v4.5.2 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.2) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.2) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.2)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.2) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.2)
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -8,6 +8,7 @@
 * BUG: Update `Stack.Create` method to populate missing `PhysicalLocation` instances when stack frames reference relative file paths.
 * BUG: Fix `UnsupportedOperationException` in `ZipArchiveArtifact`.
 * BUG: Fix `MultithreadedAnalyzeCommandBase` to return rich return code with the `--rich-return-code` option.
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character or encoded character.
 * NEW: Add `IsBinary` property to `IEnumeratedArtifact` and implement the property in `ZipArchiveArtifact`.
 * PRF: Change default `max-file-size-in-kb` parameter to 10 megabytes.
 * PRF: Add support for efficiently peeking into non-seekable streams for binary/text categorization.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,7 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 ## **v4.5.1 UNRELEASED
 * BUG: When the path of the scan target contains invalid characters, report `WRN997_InvalidTarget` before analyzing it against any rule, otherwise `ArgumentException: Illegal characters in path` will be thrown.
-* BUG: When the path of the scan target contains URL Percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.
+* BUG: When the path of the scan target contains URL percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.
 
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
 * DEP: Downgrade `System.Text.Encoding.CodePages` from 8.0.0 to 4.3.0 in `Sarif`.
--- a/src/Sarif.Driver/OrderedFileSpecifier.cs
+++ b/src/Sarif.Driver/OrderedFileSpecifier.cs
@@ -48,10 +48,9 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             string normalizedSpecifier = Environment.ExpandEnvironmentVariables(this.specifier);
 
             // 'normalizedSpecifier' should not be changed even if it contains URL Percent-encoding characters,
-            // thus 'uri.OriginalString' is used below.
             if (Uri.TryCreate(this.specifier, UriKind.RelativeOrAbsolute, out Uri uri))
             {
-                normalizedSpecifier = uri.OriginalString;
+                normalizedSpecifier = uri.GetFilePath();
             }
 
             string filter = Path.GetFileName(normalizedSpecifier);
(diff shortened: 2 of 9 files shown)


LingZhou-gh:
@stacywray is added to the review. #Closed

LingZhou-gh:
@stacywray is added to the review. #Closed

LingZhou-gh:
@stacywray is added to the review. #Closed

ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.5.1 UNRELEASED
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid character or encoded character.
+
```
stacywray, line 4: Add "an" before "invalid character" and before "encoded character"

LingZhou-gh: I change 'character' to 'characters' instead, as the encoded character could mean several character combination such as %2F.

stacywray (approved):
:shipit:

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -47,6 +47,7 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
         {
             string normalizedSpecifier = Environment.ExpandEnvironmentVariables(this.specifier);
 
+            // normalizedSpecifier might be changed if it contains URL Percent-Encoding characters.
```
michaelcfanning, line 50: Please try to remember to use single quotes to indicate named things.

i.e., normalizedSpecifier is not an English word, it is the name of a variable. So for readability say. Note I fixed a casing problem as well:

// 'normalizedSpecifier' might be changed if it contains URL percent-encoded characters.

I repeat this code review note a lot. If you do not understand it, let's please talk about this language subtlety at our next stand-up.

@Stacy Wray, FOCR

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
```diff
@@ -31,6 +31,17 @@ public PathExtensionsTests(ITestOutputHelper outputHelper)
...
         }
 
+        [Fact]
+        [Trait(TestTraits.WindowsOnly, "true")]
+        public void ContainsInvalidPathChar_test()
+        {
+            string validPath = "test.md";
+            validPath.ContainsInvalidPathChar().Should().BeFalse();
+
+            string invalidPath = "test\t.md";
```
yongyan-gh, line 41: >\t [](http://example.com/codeflow?start=38&length=2)

consider get an invald char from `System.IO.Path.GetInvalidPathChars()` so it should work on both windows/linux/mac

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1222,6 +1223,18 @@ protected virtual void AnalyzeTarget(TContext context, IEnumerable<Skimmer<TCont
 
         public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TContext>> skimmers, ISet<string> disabledSkimmers)
         {
+            // This might the public API invoked by other tools
+            Uri uri = context.CurrentTarget.Uri;
+            string originalStr = uri.OriginalString;
+            string filePath = uri.GetFilePath();
+            var URLPercentEncodingRegex = new Regex(@"%[0-9][0-9A-Fa-f]", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
michaelcfanning, line 1230: >URLPercentEncodingRegex [](http://example.com/codeflow?start=16&length=23)

you should cache this regex. why create it on every call?

you shouldn't be using .net here for this pattern matching, please use the context regex engine.

michaelcfanning:
>         public void AnalyzeCommand_EncodedPaths()

this change isn't right!

we authored this test because these aren't legal file paths, they are relative paths.

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:835 in 7340d37. [](commit_id = 7340d376bb832c77894b6bd2509cfa3b3dd79d27, deletion_comment = False)

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
michaelcfanning, line 1230: >URLPercentEncodingRegex [](http://example.com/codeflow?start=16&length=23)

forbidding URL percent-encoding isn't right.

this is a subtle topic, we need to talk about this more.

i think you will drop analysis of scan targets with your approach and we should be careful about that

michaelcfanning (changes requested):
:clock1:

src/Sarif.Driver/OrderedFileSpecifier.cs
LingZhou-gh: Update the comment

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
LingZhou-gh: Good suggestion. Update the regex.

LingZhou-gh: The new changes now allow scan targets with URL percent-encoding path. It won't be dropped anymore.

LingZhou-gh:
>         public void AnalyzeCommand_EncodedPaths()

With new commit, this test case can keep unchanged. Add a new test case instead.

---
In reply to: [1942193638](https://github.com/microsoft/sarif-sdk/pull/2783#issuecomment-1942193638) [](http://example.com/codeflow?ancestors=1942193638,1939774900)

---
Refers to: src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs:835 in 7340d37. [](commit_id = 7340d376bb832c77894b6bd2509cfa3b3dd79d27, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
```diff
@@ -31,6 +31,49 @@ public PathExtensionsTests(ITestOutputHelper outputHelper)
...
+            if (invalidPathChars.Length > 0)
+            {
+                string invalidPath = "test" + invalidPathChars[0] + ".md";
+                this._outputHelper.WriteLine($"The invalid path is: {invalidPath}.");
+                invalidPath.ContainsInvalidPathChar().Should().BeTrue();
+            }
+        }
+
+        [Fact]
+        public void UriGetFilePath_ShouldNotChangeFileName()
```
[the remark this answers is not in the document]
LingZhou-gh: Sure, 'test%2D.md' is a relative Uri. The next one 'C:\\test%2D.md' is an absolute Uri with other path components.

src/Sarif/SarifExtensionMethods.cs
```diff
@@ -258,7 +259,19 @@ public static string GetFileName(this Uri uri)
 
         public static string GetFilePath(this Uri uri)
         {
-            return uri.IsAbsoluteUri ? uri.LocalPath : uri.OriginalString;
+            if (!uri.IsAbsoluteUri)
```
scottoneil-ms, line 262: C:\Local%20DFile\File.txt

scottoneil-ms: Path.GetFullPath(OriginalString)

LingZhou-gh:
@eddy is added to the review. #Closed

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -865,6 +865,36 @@ public void AnalyzeCommand_EncodedPaths()
             sarifLog.Runs[0].Results.Count.Should().Be(1);
         }
 
+        [Fact]
+        public void AnalyzeCommand_InvalidPaths()
+        {
+            var logger = new MemoryStreamSarifLogger(dataToInsert: OptionallyEmittedData.Hashes);
```
scottoneil-ms, line 871: Subdirectory test.  Ie: Direct%20dory\filename.txt doesn't confuse the encoder/decoder.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1241,17 +1250,14 @@ public static void AnalyzeTargetHelper(TContext context, IEnumerable<Skimmer<TCo
...
-                    Uri uri = context.CurrentTarget.Uri;
+                    DriverEventSource.Log.RuleStop(filePath, skimmer.Id, skimmer.Name);
 
                     if (stopwatch != null)
                     {
-                        string file = uri.GetFilePath();
-                        string directory = Path.GetDirectoryName(file);
-                        file = Path.GetFileName(file);
+                        string directory = Path.GetDirectoryName(filePath);
+                        string file = Path.GetFileName(filePath);
```
LingZhou-gh, line 1261: There is no functionality change here, just re-use the variable added at the beginning of the function.

src/Test.UnitTests.Sarif.Driver/PathExtensionsTests.cs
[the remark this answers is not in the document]
LingZhou-gh: More tests are added.

src/Sarif/SarifExtensionMethods.cs
LingZhou-gh: more test cases are added, including this style.

LingZhou-gh: Tried, but Path.GetFullPath(...) add a lot of unexpected prefix, which is not right.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
LingZhou-gh: The test case above this one is of the style 'Direct%20dory\filename.txt'.

src/Sarif/Writers/ConsoleLogger.cs
```diff
@@ -291,10 +291,10 @@ private static string ConstructPathFromUri(Uri uri)
 
             if (uri != null)
             {
-                // If a path refers to a URI of form file://blah, we will convert to the local path           
+                // If a path refers to a URI of form file://blah, we will convert to the local path
```
michaelcfanning, line 294: >path [](http://example.com/codeflow?start=95&length=4)

when you see comments without a period at the end of it, that does not meet our standard, please remember to add one. @stacywray , FOCR. do we have a code comment style guideline?

src/Sarif.Driver/PathExtensions.cs
```diff
@@ -159,6 +159,26 @@ public static string PathUnescapePath(string sourcePath)
             }
         }
 
+        /// <summary>
+        /// Determines whether the input string contains any invalid Path character.
```
michaelcfanning, line 163: >alid Path char [](http://example.com/codeflow?start=64&length=14)

this casing isn't correct.

src/Sarif.Driver/PathExtensions.cs
```diff
@@ -159,6 +159,26 @@ public static string PathUnescapePath(string sourcePath)
...
+        /// </summary>
+        /// <param name="path">The input string.</param>
+        /// <returns>true if the input string contains invalid Path character; otherwise, false.</returns>
```
michaelcfanning, line 166: >Path [](http://example.com/codeflow?start=63&length=4)

NIT casing is not correct.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -836,7 +836,7 @@ public void AnalyzeCommand_EncodedPaths()
         {
             var logger = new MemoryStreamSarifLogger(dataToInsert: OptionallyEmittedData.Hashes);
             var command = new TestMultithreadedAnalyzeCommand();
-            var uri = new Uri("/new folder/0%1%20%.txt", UriKind.Relative);
+            var uri = new Uri("/new%20folder/0%1%20%.txt", UriKind.Relative);
```
michaelcfanning, line 839: >20folder [](http://example.com/codeflow?start=36&length=8)

why did you have to change this?

i am concerned that you've changed some subtle behavior in our service usage.

this is a lousy test because it doesn't describe any expectations of the test. why was it authored? what was it protecting against?

src/Sarif/SarifExtensionMethods.cs
```diff
@@ -258,7 +260,25 @@ public static string GetFileName(this Uri uri)
...
+                return originalStringWithoutQuery;
+            }
+
+            if (Regex.IsMatch(originalString, URLPercentEncodingPattern))
+            {
+                // If the 'originalString' contains URL percent-encoding characters,
+                // 'uri.LocalPath' will decode them thus changing the path unexpectedly.
+                return originalStringWithoutQuery;
+            }
+            else
```
michaelcfanning, line 278: >else [](http://example.com/codeflow?start=12&length=4)

you don't need an else block here, the previous conditional always returns.

ReleaseHistory.md
```diff
@@ -1,4 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.5.1 UNRELEASED
+* BUG: Report `WRN997_InvalidTarget` if the path of the scan target contains invalid characters.
```
michaelcfanning, line 3: >Report [](http://example.com/codeflow?start=7&length=6)

this doesn't describe the symptoms of the original string.

ReleaseHistory.md
```diff
@@ -1,4 +1,8 @@
...
+* BUG: When the path of the scan target contains URL Percent-encoding characters, `uri.GetFilePath` should return the original string.
```
michaelcfanning, line 4: >When [](http://example.com/codeflow?start=7&length=4)

this is not a good release note, it describes the code fix.
 
how does this help a customer understand whether we have fixed a problem they have seen?

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
scottoneil-ms, line 839: Recommend breaking this out into a loop that tests the directory case and the file case separately rather than together.

src/Sarif/SarifExtensionMethods.cs
scottoneil-ms: What about 
Path.Combine(
Path.GetDirectoryName(uri.OriginalString),
Path.GetFileName(uri.OriginalString));
?

Generally speaking: System.IO.Path is how .NET exposes ways of manipulating local file system paths which are expressed as strings.  (In a system-independent way.)  I believe such a string must be what was originally passed into the Uri constructor.  If that's right, then the most promising strategy is going to be trying to figure out how to leverage System.IO.Path to manipulate that original string into the form that we need.

Attempting to correctly reverse the encoding performed by the URI constructor ourselves may not be strictly impossible, but it entails a lot more risk of being bit again later by something we do not understand about the intersection between Uri encoding and/or file system paths on one of the platforms we target.

ReleaseHistory.md
LingZhou-gh: Updated with symptom.

ReleaseHistory.md
LingZhou-gh: Updated with symptom.

src/Sarif/SarifExtensionMethods.cs
```diff
@@ -258,7 +260,25 @@ public static string GetFileName(this Uri uri)
...
+            string originalString = uri.OriginalString;
+            string originalStringWithoutQuery = originalString.Split('?')[0];
+
+            // 'uri.LocalPath' is not supported for RelativeUri.
+            if (!uri.IsAbsoluteUri)
+            {
+                return originalStringWithoutQuery;
+            }
+
+            if (Regex.IsMatch(originalString, URLPercentEncodingPattern))
```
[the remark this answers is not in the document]
LingZhou-gh: Updated to use a compiled regex.

src/Sarif/Writers/ConsoleLogger.cs
LingZhou-gh: Updated.

src/Sarif.Driver/PathExtensions.cs
LingZhou-gh: Updated

src/Sarif.Driver/PathExtensions.cs
LingZhou-gh: Updated.

src/Sarif/SarifExtensionMethods.cs
LingZhou-gh: Remove the else block.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
LingZhou-gh: Add more scan targets.

LingZhou-gh: Restored the original target, which is a relativeUriWithEncodedFileName. (relative Uri will return the original string in the path, so no problem before the change)

Add one more target, which is a absoluteUriWithEncodedFileName.

My understanding of the purpose of this test case is to ensure no exception is throwing before proceeding with Test rule analyzing.

ReleaseHistory.md
```diff
@@ -1,4 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+## **v4.5.1 UNRELEASED
+* BUG: When the path of the scan target contains invalid characters, report `WRN997_InvalidTarget` before analyzing it against any rule, otherwise `ArgumentException: Illegal characters in path` will be thrown.
+* BUG: When the path of the scan target contains URL Percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.
+
```
stacywray, line 5: "Percent" should be lowercase: "percent-encoded"

LingZhou-gh: updated

src/Sarif/SarifExtensionMethods.cs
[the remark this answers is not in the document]
LingZhou-gh: [https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.urldecode?view=net-8.0#system-web-httputility-urldecode(system-string](https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.urldecode?view=net-8.0#system-web-httputility-urldecode(system-string))

src/Sarif.Driver/PathExtensions.cs
```diff
@@ -159,6 +159,26 @@ public static string PathUnescapePath(string sourcePath)
             }
         }
 
+        /// <summary>
+        /// Determines whether the input string contains any invalid path character.
+        /// </summary>
+        /// <param name="path">The input string.</param>
+        /// <returns>true if the input string contains invalid path character; otherwise, false.</returns>
+        public static bool ContainsInvalidPathChar(this string path)
```
scottoneil-ms, line 167: Looking at this again, I'm not as sure this is the right answer.  Could you take a Todo and open a bug to consider not checking in advance for invalid path characters, and instead catch the Exception that eventually pops out?

This isn't an enormous string, but it's a lot of passes through it.  It's almost certainly going to be more performant to just handle the exceptions that pop out for invalid paths.

scottoneil-ms: That's definitely a complication, but I think it just means we shouldn't be catching that Exception and proceeding in the Skimmer.  We could catch it in MultithreadedAnalyzeCommandBase instead.

scottoneil-ms: Again: my suggestion here is not to take a change at this late hour.  I'm suggesting tracking a todo to revisit.

PR #2787, opened 2024-02-20
Expose utility method for building a disabled skimmers set.

scottoneil-ms:
This is useful code to have handy if you're invoking via AnalyzeTargetHelper().

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -4,7 +4,7 @@
   `System.Diagnostics.Debug` 4.3.0,
   `System.IO.FileSystem.Primitives` 4.3.0,
   `System.Text.Encoding.Extensions` 4.3.0.
-* NEW: Expose `MultithreadedAnalyzeCommandBase.BuildDisabledSkimmersSet`, a utility function which extracts a disabled skimmer set from a TContext.
+* NEW: Expose `MultithreadedAnalyzeCommandBase.BuildDisabledSkimmersSet`, a utility function which extracts a disabled skimmer set from a `TContext`.
 
 ## **v4.5.0 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.0)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.0)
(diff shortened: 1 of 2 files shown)


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1075,6 +1075,21 @@ private static SupportedPlatform GetCurrentRunningOS()
...
+
+            if (disabledSkimmers.Count == skimmers.Count())
+            {
+                Errors.LogAllRulesExplicitlyDisabled(context);
+                ThrowExitApplicationException(ExitReason.NoRulesLoaded);
+            }
+
+            this.CheckIncompatibleRules(skimmers, context, disabledSkimmers);
+
+            MultithreadedAnalyzeTargets(context, skimmers, disabledSkimmers);
```
michaelcfanning, line 1088: >MultithreadedAnalyzeTargets [](http://example.com/codeflow?start=12&length=27)

the real question, since you are in this code is why do we pass the skimmers as an argument when the context object contains a skimmer property?

the second question I understand, why aren't the disabled skimmers a property on the context? well, they could be. but the disabled skimmers by definition are a cross-thread thing and we are nervous about cross-thread data in the context object.

but we could simply do this. add the disabled skimmers onto the context object and pass it everywhere.

you've still built/exposed a nice helper.

ReleaseHistory.md
```diff
@@ -4,6 +4,7 @@
   `System.Diagnostics.Debug` 4.3.0,
   `System.IO.FileSystem.Primitives` 4.3.0,
   `System.Text.Encoding.Extensions` 4.3.0.
+* NEW: Expose `MultithreadedAnalyzeCommandBase.BuildDisabledSkimmersSet`, a utility function which extracts a disabled skimmer set from a TContext.
```
michaelcfanning, line 7: >TContext [](http://example.com/codeflow?start=138&length=8)

wrap this (and all named things, generally) in back ticks. Looks better in markdown. FCOR. :) :) :)

michaelcfanning (approved):
:shipit:

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1075,6 +1075,21 @@ private static SupportedPlatform GetCurrentRunningOS()
...
+        }
+
+        public static ISet<string> BuildDisabledSkimmersSet(TContext context, IEnumerable<Skimmer<TContext>> skimmers)
```
suvamM, line 1091: @scottoneil-ms This is a user facing method now. There should be tests for it, and XML docs.

PR #2793, opened 2024-02-26
Fix MessageResourceNames merging

EasyRhinoMSFT:
The previous architecture seemed to be sort of incompatible with the existing skimmer platform in some cases. This change sets up lists of message strings in the rule and base classes, with the Rule.MessageResourceNames property returning a merged list. It also cleans up some overlooked messiness.


EasyRhinoMSFT:
Anyone else getting an error when you open this in CodeFlow?

shaopeng-gh (approved):

PR #2794, opened 2024-02-27
Fix issue of EnumeratedArtifact with problematic file path (containing either legal character or URL percent-encoding characters)

LingZhou-gh:
**Code Change:**
1. When the path of the `EnumeratedArtifact` contains URL percent-encoding characters, it should proceed with analyzing without throwing `System.IO.FileNotFoundException`.

A new property 'CleanPath' was added to the 'EnumeratedArtifact' class to get a clean path of the associated URI.
- It starts with the 'URI.OriginalString' **without decoding**.
- If the 'OrignalString' starts with file scheme such as "file:///path", it mimics the behavior of URI.LocalPath but **without decoding**, by using regex match and extra the parts between directory separators.

The new property 'CleanPath' ensures no change in the file name and correct format (without file scheme prefix) before retrieve the length and content of the `EnumeratedArtifact`. 

2. When the path of the `EnumeratedArtifact` contains invalid characters, skip it as scan target to avoid `System.IO.IOException: The filename, directory name, or volume label syntax is incorrect`.

It doesn't seem to have a way to access a file using System.IO.FileInfo class if the file name contains illegal character, so this change will skip the file as scan target.

**Test:**
Three unit-tests were added to 'EnumeratedArtifactTests' class.
Three e2e tests were added to 'AnalyzeCommandBaseTests' class.
They test the following cases:
(1) File path with illegal character
(2) File path with encoded characters
(3) File path with encoded characters and staring with file scheme prefix "file:/"


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -666,18 +666,30 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
...
+                }
+
+                if (artifactSize == 0)
                 {
                     DriverEventSource.Log.ArtifactNotScanned(filePath, DriverEventNames.EmptyFile, 00, data2: null);
                     Notes.LogEmptyFileSkipped(globalContext, filePath);
                     continue;
                 }
 
-                if (!IsTargetWithinFileSizeLimit(artifact.SizeInBytes.Value, globalContext.MaxFileSizeInKilobytes))
```
LingZhou-gh, line 676: There is no functionality change in this paragraph. It just re-uses the variables defined before those statements.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -64,6 +66,39 @@ public byte[] Bytes
...
+            }
+
+            if (originalString.StartsWith("file:"))
+            {
+                Match match = AbsoluteFilePathWithFileSchemeRegex.Match(originalString);
+                if (match.Success)
+                {
+                    string separator = Path.DirectorySeparatorChar.ToString();
+                    this.cleanPath = match.Groups[1].Value.Replace("/", separator);
+                    return this.cleanPath;
```
LingZhou-gh, line 91: An example is as follows:
If the originalString is **file:///C:/Users/v-zhouling/AppData/Local/Temp/New%2DYapeAzureCertificateForWinRMOverHttpsInKeyVault.md**,
the transferred string (in Windows) will be: **C:\Users\v-zhouling\AppData\Local\Temp\New%2DYapeAzureCertificateForWinRMOverHttpsInKeyVault.md**

src/Sarif/Notes.cs
```diff
@@ -37,7 +37,7 @@ public static void LogFileSkipped(IAnalysisContext context, string skippedFile,
                     Path.GetFileName(skippedFile),
                     reason));
 
-            context.RuntimeErrors |= RuntimeConditions.OneOrMoreFilesSkippedDueToExceedingSizeLimits;
+            context.RuntimeErrors |= RuntimeConditions.OneOrMoreFilesSkipped;
```
LingZhou-gh, line 40: I believe this is a bug if you look at the line 89 below.

michaelcfanning (dismissed):
:shipit:

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -666,18 +666,30 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
                     continue;
                 }
 
-                if (artifact.SizeInBytes == 0)
+                long artifactSize = 0;
+                try
+                {
+                    artifactSize = artifact.SizeInBytes.Value;
+                }
+                catch (Exception e) when (e is IOException || e is ArgumentException)
```
nguerrera, line 674: Why `ArgumentException`? I hope there's not a property getter (all that's in the try block) that throws that.

Also, is the API in the try block owned by us? If it does IO and can therefore fail, it shouldn't have been a property.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -64,6 +66,39 @@ public byte[] Bytes
...
+                }
+
+                throw new IOException("Invalid file path format.");
```
nguerrera, line 94: This doesn't look like an IO exception, no IO was performed to get to it.
 
Also, CleanPath shouldn't be property if the getter can throw this.

nguerrera:
I am reviewing and will have comments. Putting a waiting early.

src/Sarif/EnumeratedArtifact.cs
nguerrera: Isn't that `%2D` a url-encoded `-` that should be url-decoded when converting to a windows path?

Instead of using the original string + regex + string replace, `Path.GetFullPath(uri.AbsolutePath)`

https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6ASQHkBuPDIgBSIG88iOTMA2EgFiIBZAIYBLKAAoAlK3ad5AN2FgiEMKKIBeIlACmAdyIBVdRIBEAM1EAbXSAD0jgMIPgugM7B7AOQMBSdAARAE1dZTIAWzgzKTpceXlSYmsAewBjYWtGYWAACy0mHNyyAHFdYAAxCGssook1UTIAQQAjdxTrCDdsvNi5BK4ATglUjNreuPkAXzwpoA===

nguerrera (changes requested):

src/Sarif/EnumeratedArtifact.cs
nguerrera: I apologize, I missed the description and jumped to the code.

I would like to understand more why we have URL's that have encoded strings, but shouldn't be decoded. 

The original URLs are wrong. Where did they come from like that?

src/Test.UnitTests.Sarif/EnumeratedArtifactTests.cs
```diff
@@ -34,6 +34,43 @@ public void EnumeratedArtifact_BinaryFile_OnDisk()
...
+
+            artifact.CleanPath.Should().Be(originalString);
+        }
+
+
+        [Fact]
+        public void EnumeratedArtifact_FilePath_EncodedChars()
+        {
+            string originalString = "C:\\Users\\%28test%29\\Downloads\\New%2DYear.md";
+            var uri = new Uri(originalString, UriKind.Absolute);
```
nguerrera, line 53: I see the culprit is the Uri constructor. :(

I'm paging stuff back in here. I tried discussing it with Copilot and after 6 of its suggestions failed, it started to suggest making a `CustomUri` class. 🤣

nguerrera: I will contact my dear old friends on the BCL team about this.

nguerrera: They were still asleep and I was sniped. Here is a helper I just wrote to make a correct URI from a local absolute path. 

[Demo](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgBgAJVMA6AEQEsBDAcygHsBnYKgY2YG49CTyAkgHkeuXun4B2PAG88RBfwBsJACxEAsjSpQAFAEp5iubkVn+xAA41gACyIBeEqgBEAYRAAdYAFNWngDkfAHcAUnQKAE0fGjAXUXNFADdYogAjGjgAVTAqRyIoEKIcql1rOxhi3IBpHTgyAEE05kYAGwhffQTEohSwIjpGRmzc/JKAMUYwJpb23wAFG1sypa68I0TSAE5dABIXctsQIhlDgF8XNdMe7YNuzcwd/YyRqmOZF5KLq57+J5dBsMSu9Aa9vvdzLcfjdHnsXJ9cmQADKMdg0VqLOzvBFUZGo9GY2zgjaQ2H7UElPFojFLEFDV5UglLYnXBRndasogAeh5RAAPMwIABbIWxACeAD4STyuUQ3GAYr4iDRigAlAREABmYEYQuVUGVzTaHR8WqorVNhzIRGlvKyzFNdiozCIOlYMTgREYmv5DtN7AVmocLkpJV0pAIlRKtSgcH0Li5EqIwEYRFsNFjFqIAHJQtnXQbNebLUsXewpgr2MBWmKyLbZXyuYKReKpZzSCoSlUqJNpka5j5CeHMFZViSTL8qD6VnZHE4oBBWq1DJyzBPfoo7DrggUig0wHRhT4oMAAovWgBRAAe7B8ljYjD0UBoQp83pntn00MS7LEq8UU5ELoACEhJkAIzCEqqQy+HAH5fiSa6IT0W6MDuhQ7vuh6vie163veVCProLiEkQQoQKw6SmjQ/YmmQLiVM+r7vocX4QmYv7IQovIBFMoqtFQABeJazhmnoUaampTMEsSeswrQ0Mwth+FxRCHPkYEAOI+MA4znkOrFkKqd4KbeujZp4njZpU2ZctmVyqbyDRwJ6FqZDodBEPJin2CmRAAOp1GhLo6owwDWj4ZB0Na7ggFyvisC4RAALRJgmHjxX4wAuKpgEgVaADKwCxMAzCBXYxFcpcK6/OuG5qUs+QJkliANXY7GKJx/7cbKWSqkiyXHuWcA+JUaQdEQADWPh3l5ClKX4+qeuWrSPswqnqU4lIXswaKWD4FA2DQRW5FAdDwUZJk0GZLjhOMDFEM1+iXZYpk+MRoQAMwNA9LggJcoiObKpHOgUaHKr06JUJ6fUaup5ZCpYj7HsAS1EME/oZukEDmp6KpdowaQAFY+FWWo6nqVDhapqCSLuO4lAAQjjrQjf0MhEAV7DKa+TVFhaD0ABIsKjTguA9pFOG19hnGQJQQr+XV4EAA)

```cs
 /// <summary>
 /// Create a URI from an absolute file path. 
 /// Use this instead of <see cref="Uri.Uri(string, UriKind)"/> to handle '%' in file paths correctly.
 /// </summary>
 static Uri UriForAbsolutePath(string path)
 {
     if (path == null)
     {
         throw new ArgumentNullException(nameof(path));
     }

     if (!Path.IsPathRooted(path))
     {
         throw new ArgumentException("Path must be absolute.", nameof(path));
     }

     // Normalize path and use forward slashes
     path = Path.GetFullPath(path).Replace('\\', '/');

     // Add leading slash to Windows root. e.g. "C:/test" -> "/C:/test"
     if (!path.StartsWith("/"))
     {
         path = "/" + path;
     }

     // URL-encode, but keep slashes and colons
     path = Uri.EscapeDataString(path).Replace("%2F", "/").Replace("%3A", ":");

     // Path is now a valid URI path component and we can build a Uri object from it.
     return new UriBuilder { Scheme = "file", Host = "", Path =  path }.Uri;
 }
 ```

nguerrera: We should use this wherever we construct these URIs from paths and then we can safely use Uri.LocalPath.

michaelcfanning: Great! Can you please just create a clean PR with your work and tests?

michaelcfanning: Ling will be fine with this, I am sure. This PR is months old and she is busy with other things. I ran into this issue again doing some explorations around zip scanning (where we encounter unusual characters). I remembered the open PR, and bumped it. Looking it the implementation, however, I was not certain we can accept it as is. Sounds like you could help move this over the finish line. I'm closing this PR for  now.

nguerrera: Will do!

PR #2796, opened 2024-02-28
Do not merge: Second working fork of Ling's work for file encoding

scottoneil-ms:


src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -656,6 +656,20 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
 
                 string filePath = artifact.Uri.GetFilePath();
 
+                long artifactSizeInBytes = -1;
```
LingZhou-gh, line 659: NIT: Should this block be moved a little down just before the statement 669, where the resulting artifactSizeInBytes will be evaluated (whether it is 0 or whether it is exceeding the size threshold?

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -37,6 +37,99 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
...
+        [Fact]
+        public void AnalyzeCommand_AbsoluteFileURIWithPercentInDirectory()
+        {
+            var logger = new TestMessageLogger();
+            var command = new TestMultithreadedAnalyzeCommand();
+
+            string dirPath = Path.Combine(Path.GetTempPath(), "New%2DYearDir");
+            if (!Directory.Exists(dirPath))
+            {
+                Directory.CreateDirectory(dirPath);
```
LingZhou-gh, line 108: There is no big difference between this test case and the one above. IMO, you can combine them into one, that is both the directory, and the filename contains encoded characters.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -446,11 +539,9 @@ public void ExceptionRaisedInvokingInitialize()
         [Fact]
         public void FileUri()
         {
-            var uri = new Uri(this.GetType().Assembly.Location);
-
             var options = new TestAnalyzeOptions()
             {
-                TargetFileSpecifiers = new string[] { uri.ToString() },
+                TargetFileSpecifiers = new string[] { this.GetType().Assembly.Location },
```
LingZhou-gh, line 544: Don't know the original purpose of this test case. 
But before the change, 'uri.ToString()' will returl a string starting with "file:/". Do we still plan to support that, which is absolute file URI starting with file scheme prefix? If we don't plan to the support that? The essence of the fix is replacing 'uri.LocalPath' to 'uri.OriginalPath', that is it.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -78,11 +82,10 @@ public byte[] Bytes
 
             if (Stream == null && this.contents == null && this.bytes == null)
             {
-                if (Uri == null ||
-                    !Uri.IsAbsoluteUri ||
-                    (Uri.IsAbsoluteUri && !Uri.IsFile))
+                if (Uri == null || !FileSystem.FileExists(Uri.OriginalString))
```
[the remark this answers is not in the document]
LingZhou-gh: FileSystem.FileExists will return false if the 'Uri.OriginalString' starting with "file:/". 
So this change won't support file path staring with "file:/".
If that is the decision, shall we explicitly call it out in the **release note**?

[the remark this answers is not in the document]
scottoneil-ms: Hot-damn, nice catch.  Shakes my faith in the whole change.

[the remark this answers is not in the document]
scottoneil-ms: Thanks!

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
scottoneil-ms: I hope that placing it up at the top of the method like this sends the message to future maintainers that they should tend to add new functionality _below_ it.  This reduces the risk of a file access that would trigger one of the exceptions winding up before the check that's guarded by the catch blocks.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -666,7 +680,7 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
                     continue;
                 }
 
-                if (artifact.SizeInBytes == 0)
+                if (artifactSizeInBytes == 0)
```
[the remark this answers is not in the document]
scottoneil-ms: Thanks. Fixed.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
scottoneil-ms: I would say the idea is we always call into things that open files using variants that use OriginalString.  We know this is fine because someone passed us the string into a URI in an original form that came from... well .. something.  If that something is a Directory.GetFiles() call, then it's going to be in a format that the .NET APIs can handle for opening.  If that something is directly off the command line, then maybe someone will pass in file://whatever and an open call could fail.  So yes, we'd be dropping support for that very specific case to make the problem of handling percents in file paths more tractable.

Does anyone do that?  I doubt it.  So it seems like a favorable tradeoff.  We'll see if it passes CR. :)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
scottoneil-ms: Fixed.  Thanks!

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -47,11 +47,6 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
         {
             string normalizedSpecifier = Environment.ExpandEnvironmentVariables(this.specifier);
 
-            if (Uri.TryCreate(this.specifier, UriKind.RelativeOrAbsolute, out Uri uri))
```
michaelcfanning, line 50: >e [](http://example.com/codeflow?start=25&length=1)

I would expect a unit test in `OrderedFileSpecifierTests` for a change like this.

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -47,11 +47,6 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
...
-            {
-                normalizedSpecifier = uri.GetFilePath();
-            }
-
             string filter = Path.GetFileName(normalizedSpecifier);
```
michaelcfanning, line 50: >filter [](http://example.com/codeflow?start=19&length=6)

This code now never expands a specifier to a full path. That means that things like directory name might remain something like `..\..\`. Is this intentional? It makes me concerned that we are introducing some subtle change in behavior because we no longer fully expand.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -160,15 +163,15 @@ private void TryRewindStream()
                 }
                 else if (this.Stream != null)
                 {
-                    this.SizeInBytes = (long)this.Stream.Length;
+                    this.sizeInBytes = (long)this.Stream.Length;
                 }
                 else if (Uri != null && Uri.IsAbsoluteUri && Uri.IsFile)
                 {
-                    this.sizeInBytes = (long)FileSystem.FileInfoLength(Uri.LocalPath);
```
michaelcfanning, line 167: >LocalPath [](http://example.com/codeflow?start=75&length=9)

are you sure this change is required now that you've removed the call in ordered file specifier to remove the call to `GetFile`?

It doesn't make sense to me that `LocalPath` would be broken unless something that intervenes between the file enumeration and this code path has done something unexpected to the URI.

Thinking it all over, I now believe the root cause is for relative file paths that work against the file system. Or at least I'd be very sure to look at this case.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -656,28 +656,42 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
 
                 string filePath = artifact.Uri.GetFilePath();
 
+                long artifactSizeInBytes = -1;
+                try
+                {
+                    artifactSizeInBytes = artifact.SizeInBytes.Value;
+                }
+                catch (Exception e) when (e is ArgumentException || (e is IOException && e.Message.IndexOf("The filename, directory name, or volume label syntax is incorrect.") != -1))
```
michaelcfanning, line 664: >Exception [](http://example.com/codeflow?start=23&length=9)

this line of code is very long, perhaps create a string const for your message literal.

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -656,28 +656,42 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
...
+                {
+                    // These exceptions can come out of the FileSystem when we're passing in invalid characters.  We need to catch and
```
michaelcfanning, line 666: >exceptions [](http://example.com/codeflow?start=29&length=10)

NIT: please 'columnize this comment for better readability, i.e., prefer line lengths that are a bit closer to being equal in size.

michaelcfanning:
>         public long? SizeInBytes

this code really should be commented if it raises exceptions for invalid file paths. you didn't create this technical debt but i can no longer remain silent about it.

---
Refers to: src/Sarif/EnumeratedArtifact.cs:147 in a7f7257. [](commit_id = a7f7257c7d9b0753c93513329c9e043b21af5f4c, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -37,6 +37,85 @@ public AnalyzeCommandBaseTests(ITestOutputHelper output)
...
+        [Fact]
+        public void AnalyzeCommand_AbsoluteFileURIWithPathsContainsIllegalCharacters()
+        {
+            var logger = new TestMessageLogger();
+            var command = new TestMultithreadedAnalyzeCommand();
+
+            // No content is set for the following uri, so when its length or content is retrieved,
+            // the file path format will be evaluated.
+            var uri1 = new Uri("C:\\test" + Path.GetInvalidPathChars()[0] + "ABC.md", UriKind.Absolute);
+            var target1 = new EnumeratedArtifact(FileSystem.Instance) { Uri = uri1 };
```
michaelcfanning, line 49: >EnumeratedArtifact [](http://example.com/codeflow?start=30&length=18)

this doesn't this code crash? doesn't that make more sense as a design?

i.e., why should we allow a URI with invalid characters to be successfully created? is it so that users can have the convenience of creating illegal file URIs in cases where they bring their own stream? is that really a compelling user scenario?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -656,28 +656,42 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
 
                 string filePath = artifact.Uri.GetFilePath();
 
+                long artifactSizeInBytes = -1;
+                try
+                {
+                    artifactSizeInBytes = artifact.SizeInBytes.Value;
```
michaelcfanning, line 662: I'm hostile to this change. it doesn't make sense to me that we permit a code path of this kind with a file name that is deemed invalid at runtime. Further, although this code is nicely commented, it's surprising that simply obtaining the size in bytes property provokes this exception.

src/Sarif.Driver/OrderedFileSpecifier.cs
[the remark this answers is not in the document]
scottoneil-ms: This isn't setting this.Uri.  It's creating a temp URI (which will be URI-encoded) so we can subsequently call GetFilePath on it and update the normalized specifier.  Uri should still be created as normal. (Wherever that happens.)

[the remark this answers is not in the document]
michaelcfanning: this path normalization had impact on the filter and directory value we realized after this. now that normalization isn't occurring.

it's basically like eliminating a Path.GetFullPath call I suspect. i.e., constructs like `.\` and `..\..\` aren't resolved.

it's always dangerous to change this kind of behavior.

this code path is part of the core problem. the behavior here will differ depending on whether URI believes it has a file URI or not. that means for any relative paths, it will assume we have a web URI not a file URI. this is probably what's leading to encoding confusion

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
scottoneil-ms: Yeah, I have misgivings, too, but I want to defend this.  For background: Ling's initial PR handled this quite differently.  She had some code that scanned the path for invalid characters before we start doing IOs.  It's not cheap.  For every character that's an invalid filepath character, you scan through the string.  HashSets and similar can get that to what is formally O(n), but with a significant constant overhead.

I definitely don't want to pay that cost during file enumeration.  We could pay it after file enumeration (at higher cost).  But why pay it at all?  Lower layers are going to do this check for us, whether we like it or not.  It's a super rare condition.  So the optimal thing from a perf perspective is to let those layers enforce this for us (probably in optimized native code), and in the extremely rare cases where we have an invalid character in the path: catch and handle the Exception.

The place to deliberately provoke the exception is just before you're sure you need to be interfacing with the filesystem, anyway.  But also as late as possible subject to that constraint because if you're going to screen for file extensions and such, you'd rather do that first.  And you'd like to provoke the Exception in a way that gets you metadata (because cheaper) that you need anyway later.

So, that's how this code wound up in this form and in this place.  Which outcome would you prefer?
* Document that reasoning better in the code, 
* Pay the cost of provoking the Exception earlier at a higher layer (and possibly waste it)
* Do explicit file path scrubbing at a higher layer

src/Sarif.Driver/OrderedFileSpecifier.cs
scottoneil-ms: There is no such suite AFAICT.  After investigating this more, I think all our coverage of this class comes as pass-through coverage via UTs that use `MultiThreadedAnalyzeCommandBase`.   `ValidateCommandTests` does.  Possibly others.  I'll make sure it's covered by the next iteration via such a mechanism, at least.  If you want to see a new suite for `OrderedFileSpecifierTests` in that iteration, just yell and I'll make sure to get that in.

src/Sarif.Driver/OrderedFileSpecifier.cs
scottoneil-ms: Not intentional.  It was puzzling me *why* we did this temporary URI parsing.  You've answered that question.  

I'm of two minds on this:
1) Restoring the absolute path normalization (w/o using Uri class) is a safeguard against introducing a regression, for the reasons you're saying.  I'm also a little worried that trying to do this could introduce a different subtle error, or partially-mask a problem we would otherwise expose by going with #2.  But on balance, I think that's the conservative option.
2) We could lean into what the rest of this change is doing.  Declare we use unmodified strings for interfacing with .NET, and Uri.GetFilePath() for logging things.  GetFilePath will clean up the relative paths for the logs, and I think we can reasonably expect .NET to not choke on them.

They're both defensible.  I'll definitely at least try #1.

PR #2800, opened 2024-03-12
Fix hangs when file path is provided with wildcard without recurse flag

shaopeng-gh:
* BUG: Resolve hangs when a file path is provided with wildcard but without `-r` (recurse) flag during the multithreaded analysis file enumeration phase.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,4 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+
+## UNRELEASED
+* BUG: Resolve hangs when file path is provided with wildchar but without `-r` (recurse) flag during multithreaded analysis file enumeration phase.
+
 ## **v4.5.4 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.4) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.4) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.4)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.4) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.4)
 * BUG: Fix incorrect base class in rule ADO2012.
 
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -1,7 +1,7 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
 
 ## UNRELEASED
-* BUG: Resolve hangs when file path is provided with wildchar but without `-r` (recurse) flag during multithreaded analysis file enumeration phase.
+* BUG: Resolve hangs when a file path is provided with wildcard but without `-r` (recurse) flag during the multithreaded analysis file enumeration phase.
 
 ## **v4.5.4 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.4) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.4) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.4)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.4) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.4)
 * BUG: Fix incorrect base class in rule ADO2012.
--- a/src/Sarif.Driver/OrderedFileSpecifier.cs
+++ b/src/Sarif.Driver/OrderedFileSpecifier.cs
@@ -103,10 +103,18 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
             }
             else
             {
-                WriteFilesInDirectoryToChannel(directory, filesToProcessChannel, filter, new SortedSet<string>(StringComparer.Ordinal));
-
-                filesToProcessChannel.Writer.Complete();
-                directoryEnumerationTask = Task.CompletedTask;
+                directoryEnumerationTask = Task.Run(() =>
+                {
+                    try
+                    {
+                        WriteFilesInDirectoryToChannel(directory, filesToProcessChannel, filter, new SortedSet<string>(StringComparer.Ordinal));
+                    }
+                    finally
+                    {
+                        filesToProcessChannel.Writer.Complete();
+                        directoryEnumerationTask = Task.CompletedTask;
+                    }
+                });
             }
 
             ChannelReader<string> reader = filesToProcessChannel.Reader;


src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -103,10 +103,18 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
             }
             else
             {
-                WriteFilesInDirectoryToChannel(directory, filesToProcessChannel, filter, new SortedSet<string>(StringComparer.Ordinal));
-
-                filesToProcessChannel.Writer.Complete();
-                directoryEnumerationTask = Task.CompletedTask;
+                directoryEnumerationTask = Task.Run(() =>
```
shaopeng-gh, line 106: This is the only fix needed to work, I tested run with it modified directly in SPMI.

michaelcfanning, line 106: >directoryEnumerationTask [](http://example.com/codeflow?start=16&length=24)

where is the test for this?

shaopeng-gh: where is the test for this --- working on it next. Want to give an early update that the issue has been found, or else will need Scott's view since he made this change.

ReleaseHistory.md
```diff
@@ -1,4 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+
+## UNRELEASED
+* BUG: Resolve hangs when a file path is provided with wildcard but without `-r` (recurse) flag during the multithreaded analysis file enumeration phase.
```
stacywray, line 4: - Add "a" before "wildcard"
- Add "," before "but
- Add "a" before `-r`
- Hyphenate "multi-threaded"

stacywray (changes requested):
Requested changes.

ReleaseHistory.md
shaopeng-gh: fixed

ReleaseHistory.md
```diff
@@ -1,4 +1,8 @@
 # SARIF Package Release History (SDK, Driver, Converters, and Multitool)
+
+## UNRELEASED
+* BUG: Resolve hangs when a file path is provided with a wildcard, but without a `-r` (recurse) flag during the multi-threaded analysis file enumeration phase.
```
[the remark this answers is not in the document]
stacywray: "Resolve program hangs when a file path..."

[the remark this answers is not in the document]
stacywray: "Resolve process hangs..."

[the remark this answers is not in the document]
shaopeng-gh: updated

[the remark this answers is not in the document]
stacywray: Honestly, we should not even use the word "hang" or "hangs" - poor term. I would prefer "Resolve unresponsive processes....", but this seems to be the norm here.

stacywray (approved):

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -103,10 +103,18 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
...
+                {
+                    try
+                    {
+                        WriteFilesInDirectoryToChannel(directory, filesToProcessChannel, filter, new SortedSet<string>(StringComparer.Ordinal));
+                    }
+                    finally
+                    {
+                        filesToProcessChannel.Writer.Complete();
+                        directoryEnumerationTask = Task.CompletedTask;
```
scottoneil-ms, line 115: Is setting this Task accomplishing anything?  I don't think it should.  Can be safely removed, I think.

Once the finally blocks are doing the same thing, I think you can simplify the factoring by checking if(recurse) within the Task.Run() delegate, and use the same finally block for both conditions.

Good change!

shaopeng-gh: was thinking the same, fixed.

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
```diff
@@ -35,5 +36,64 @@ public void Multitool_LaunchesAndRunsSuccessfully()
                 process.ExitCode.Should().Be(0);
             }
         }
+
+#if DEBUG
```
shaopeng-gh, line 40: We need to test the analyze command hang, in sdk we only have analyze-test which is also wrapped under `#if DEBUG` so my test will also need to be wrapped under it.
I have tested ` .\BuildAndTest.cmd -Configuration Debug` will all pass.

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
```diff
@@ -35,5 +36,64 @@ public void Multitool_LaunchesAndRunsSuccessfully()
...
+            {
+                WindowStyle = ProcessWindowStyle.Hidden,
+                CreateNoWindow = true,
+                RedirectStandardOutput = true,
+                UseShellExecute = false
+            };
+
+            using (var process = Process.Start(startInfo))
+            {
+                var timer = new System.Timers.Timer(30000); // 30 seconds.
```
shaopeng-gh, line 76: with fix, it will pass, it will only runs within 5 seconds so will not affect testing run total time:
image

before fix, it will fail and the message will be like below for dev:
image

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -87,27 +87,24 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
 
             Task directoryEnumerationTask;
 
-            if (this.recurse)
+            directoryEnumerationTask = Task.Run(() =>
```
shaopeng-gh, line 90: The issue and fix is simple, in the recurse false code path we missed to use `Task.Run` to run the file enumeration operation asynchronously on the separate thread.
Now we merged in a same code block so both 
recurse true and false will have it.

src/Sarif.Driver/OrderedFileSpecifier.cs
shaopeng-gh: added

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
scottoneil-ms: I was confused.  Shaopeng educated me offline.  Issue is we fill the channel without yielding our thread.  So you really do need 10k+ files, because that's the channel size.

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
```diff
@@ -35,5 +36,64 @@ public void Multitool_LaunchesAndRunsSuccessfully()
...
+            string directoryPath = Path.Combine(Path.GetTempPath(), "SarifMultitool12kTestFiles");
+             
+            if (Directory.Exists(directoryPath))
+            {
+                Directory.Delete(directoryPath, true);
+            }
+
+            Directory.CreateDirectory(directoryPath);
+
+            for (int i = 1; i <= 12000; i++)
```
scottoneil-ms, line 59: Soft-recommend (do not require) using internalsvisibleto to couple this parameter to the channel size.
`ChannelCapacity`.

Signing off.

scottoneil-ms (approved):

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
shaopeng-gh: Also: did you go back and verify that the UT hangs on the unmodified code? I think that's important for this tes.
----- yes, undo the fix and leave test can repro and have the error in the screenshot in the other comment.

as talk in IM we need 10241 files.

Will work on the change you recommended in IM.

src/Test.FunctionalTests.Sarif/MultitoolCommandLineTests.cs
shaopeng-gh: updated to use the ChannelCapacity + 1.
Tried a few times, only need + 1 to repro.

scottoneil-ms (approved):

PR #2801, opened 2024-04-03
Fix for Symbolic Link

shaopeng-gh:


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1099,62 +1099,100 @@ public void MultithreadedAnalyzeCommandBase_TargetFileSizeTestCases()
...
+                    maxFileSizeInKB = long.MaxValue,
+                    isSymbolicLink = true,
+                    actualFileContent = "ABC123"
+                },
+                new {
+                    expectedExitReason = ExitReason.NoValidAnalysisTargets,
+                    fileSizeInBytes = (long)ulong.MinValue,
+                    maxFileSizeInKB = long.MaxValue,
+                    isSymbolicLink = true,
+                    actualFileContent = ""
```
shaopeng-gh, line 1153: when the target file is also empty.

ReleaseHistory.md
```diff
@@ -2,6 +2,7 @@
 
 ## UNRELEASED
 * BUG: Resolve process hangs when a file path is provided with a wildcard, but without a `-r` (recurse) flag during the multi-threaded analysis file enumeration phase.
+* BUG: Fix error `ERR997.NoValidAnalysisTargets` when scanning symbolic link files.
```
stacywray, line 5: LGTM

shaopeng-gh:
@michaelcfanning @scottoneil-ms @suvamM 
Ping from Veronica on progress, please review.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -1191,6 +1259,7 @@ public void MultithreadedAnalyzeCommandBase_TargetFileSizeTestCases()
                 mockFileSystem.Setup(x => x.FileOpenRead(It.IsAny<string>())).Returns(mockStream.Object);
                 mockFileSystem.Setup(x => x.FileExists(tempFile.Name)).Returns(true);
                 mockFileSystem.Setup(x => x.FileInfoLength(It.IsAny<string>())).Returns((long)testCase.fileSizeInBytes);
+                mockFileSystem.Setup(x => x.IsSymbolicLink(It.IsAny<string>())).Returns((bool)testCase.isSymbolicLink);
```
scottoneil-ms, line 1262: The above is a very good test, but it's mocking out the symbolic link detection and there isn't any coverage making sure that is done correctly.  This change would really benefit from an augmentation to an E2E test that incorporates symbolic links, especially in a case where the only targets are symlinks.  If that's not practicable, then direct UT coverage of FileSystem.IsSymbolicLink would be helpful.

src/Sarif/FileSystem.cs
```diff
@@ -388,6 +388,21 @@ public long FileInfoLength(string path)
...
+        /// <param name="path">
+        /// The fully qualified name or relative path of the file.
+        /// </param>
+        /// <returns>
+        /// A boolean value indicating whether the file is a symbolic link.
+        /// </returns>
+        public bool IsSymbolicLink(string path)
+        {
+            var fileInfo = new FileInfo(path);
+            return (fileInfo.Attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint;
```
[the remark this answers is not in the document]
shaopeng-gh: Added the link and the comment that it is a basic implementation. 
This basic implementation is widely used in repos.

src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
shaopeng-gh: Added a test for FileSystem.IsSymbolicLink.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -164,7 +166,14 @@ private void TryRewindStream()
                 }
                 else if (Uri != null && Uri.IsAbsoluteUri && Uri.IsFile)
                 {
-                    this.sizeInBytes = (long)FileSystem.FileInfoLength(Uri.LocalPath);
+                    if (FileSystem.IsSymbolicLink(Uri.LocalPath))
+                    {
+                        GetArtifactData();
```
michaelcfanning, line 171: >GetArtifactData [](http://example.com/codeflow?start=24&length=15)

So you've broken a core (undocumented!) assumption in this code that we do *not* load the scan target to determine its file size. 

The reason is that the file enumeration phase is single-threaded and it is helpful to defer file loading until the multithreaded scanning phase.

What you need to do here is obtain the actual path pointed to by the symbolic link. And then perform the size in bytes against that location using `FileInfoLength`. 

Speaking of which, why doesn't new FileInfo(path).Length work here? i.e., why would file open read work properly but the length check fails?

shaopeng-gh: Fixed as proposed.

michaelcfanning (approved):
:shipit:

PR #2810, opened 2024-05-06
Add option to limit the number of results logged

joscol:


src/Sarif/Writers/BaseLogger.cs
```diff
@@ -64,6 +75,33 @@ public bool ShouldLog(Notification notification)
...
+                {
+                    locationUri = firstLocation.PhysicalLocation.ArtifactLocation.Uri;
+                }
+                var resultKey = new Tuple<string, Uri>(result.RuleId, locationUri);
+
+                if (_resultsCount.TryGetValue(resultKey, out int currentCount))
+                {
+                    if (currentCount >= _resultsLimitPerRuleTarget)
+                    {
+                        ++_resultsLimited;
```
eddynaka, line 93: >++_resultsLimited; [](http://example.com/codeflow?start=24&length=18)

Should we add a notification saying: "ruleX fileY reached limit"? This would allow us to understand why we have so many findings of that type.

aasim: I like the idea of notification, but will adding a notification here trigger one each time the results are limited? That would cause a lot of noisy notifications. I think when the results are finalized, we can send just one notification for the rule that says the results for this rule have been truncated

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -1123,7 +1114,15 @@ public static ISet<string> BuildDisabledSkimmersSet(TContext context, IEnumerabl
                 }
             }
 
-            return disabledSkimmers;
+            if (disabledSkimmers.Count == skimmers.Count())
```
eddynaka, line 1117: >Count() [](http://example.com/codeflow?start=51&length=7)

nit: since this is a sortedSet, you could use .Count instead

joscol: I must have done something wrong with Git because I did not make this change.

joscol: Initially I was trying to work with the repo directly but I wasn't allowed to push to a new branch.  So I created the fork and copied the files over.  They must not have been from the same commit so when I copied the files, this change crept in and I didn't notice it.

src/Sarif/Writers/BaseLogger.cs
joscol: I added one notification per rule per file.

PR #2811, opened 2024-05-22
Remove unused rule from github.config.xml

marcogario:
The rule is now checking for more than just the HelpUri. Specifically, it requires the name of the rule to use PascalCase. This is confusing because this is not a requirement we have in Code Scanning.

Because Code Scanning do not show the HelpUri in the UI and it is only provided in the API for backwards compatibility, this commit removes the rule altogether.

See https://github.com/github/docs/issues/20872#issuecomment-1534850272

### DIFF
--- a/policies/github.config.xml
+++ b/policies/github.config.xml
@@ -4,10 +4,6 @@
     <!-- Upgraded from Warning, because GitHub requires this. -->
     <Property Key='RuleEnabled' Value='Error' />
   </Properties>
-  <Properties Key='SARIF2012.ProvideHelpUris.Options'>
-    <!-- Upgraded from Note, because GitHub displays them if available. -->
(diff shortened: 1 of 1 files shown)


felicitymay (approved):
Thank you for proposing this change to remove the source of user confusion 💖

marcogario:
@michaelcfanning Can I ask you to have a look at this PR and merge it if applicable?

michaelcfanning:
Of course and sorry for the delay. I've added you as a contributor, would you mind creating a branch off `main` with your change? And I will approve. Am I correct that we need to update `sarif.info` to get this change asap? @easyrhino-gh

EasyRhinoMSFT:
Yes I can update once a new SDK is published.

PR #2813, opened 2024-06-24
Remove unused rule from github.config.xml

marcogario:
The rule is now checking for more than just the HelpUri. Specifically, it requires the name of the rule to use PascalCase. This is confusing because this is not a requirement we have in Code Scanning.

Because Code Scanning do not show the HelpUri in the UI and it is only provided in the API for backwards compatibility, this commit removes the rule altogether.

Replaces https://github.com/microsoft/sarif-sdk/pull/2811


michaelcfanning (approved):
:shipit:

PR #2815, opened 2024-07-13
chore: remove unused Kusto command and dep

rwoll:
Cleanup removing code that is not known to be used for a few years and also allows removal of NuGet dependencies from the project.


suvamM (approved):

shaopeng-gh (approved):
:shipit:

PR #2816, opened 2024-07-14
chore: remove unused coyote bits

rwoll:
This change is part of an iterative cleanup to remove unused dependencies and code. Unused dependencies directly or transitively create additional maintenance burden.

Coyote-based testing was first introduced in #2388 but became defunct in #2625. In the latter PR, `CoyoteTestOutputLogger` became unreferenced but was not deleted. There are not currently any references to it either.

Since the Coyote-related tests were deleted in #2625, it's my assumption that the testing was no longer considered to have good ROI. If this is not the case, an alternative to this PR would be to revive the Coyote tests deleted in #2625. However, I'd advise that this PR land and then Coyote bits selectively brought back when/if that work gets prioritized. It's been absent for ~1.5 years.

No release note has been authored since this is a test-related change/dep that should not impact customers.


suvamM:
@rwoll Your analysis is correct. Coyote is a systematic testing engine for concurrency bugs. I had started to introduce its usage, but need more bandwidth to fully leverage the tool for our concurrency tests. At this point, I don't know what our concurrency testing landscape looks like. At some point in the near future, I'd like to understand our test coverage in this domain and bring back the Coyote bits.

suvamM (approved):

PR #2817, opened 2024-07-24
Update code to use 'var' where type is explicit.

michaelcfanning:
No behavioral changes, hence no release notes.

This change also disables automated SARIF object model generation: this code is stable and the autogenerator emit doesn't conform currently to our style guidelines. To enforce them we therefore need to check in and update a stable copy.

Finally, we now enforce strict adherence to the guidelines by elevating problems to errors via CLI build.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -297,6 +297,7 @@ csharp_style_prefer_local_over_anonymous_function = true:suggestion
 dotnet_diagnostic.SA1602.severity = suggestion
 dotnet_diagnostic.SA1307.severity = silent
 csharp_style_implicit_object_creation_when_type_is_apparent = true:suggestion
+csharp_style_prefer_primary_constructors = true:suggestion
 
 [*Tests.cs]
 
@@ -324,3 +325,4 @@ dotnet_style_prefer_compound_assignment = true:suggestion
 dotnet_style_prefer_simplified_interpolation = true:suggestion
 dotnet_style_namespace_match_folder = true:suggestion
 dotnet_code_quality_unused_parameters = all:error
+dotnet_style_prefer_collection_expression = when_types_loosely_match:suggestion
--- a/src/Sarif/Core/Notification.cs
+++ b/src/Sarif/Core/Notification.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#if DEBUG
 using System;
+#endif
 
 namespace Microsoft.CodeAnalysis.Sarif
 {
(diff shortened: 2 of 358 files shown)


src/.editorconfig
```diff
@@ -297,6 +297,7 @@ csharp_style_prefer_local_over_anonymous_function = true:suggestion
 dotnet_diagnostic.SA1602.severity = suggestion
 dotnet_diagnostic.SA1307.severity = silent
 csharp_style_implicit_object_creation_when_type_is_apparent = true:suggestion
+csharp_style_prefer_primary_constructors = true:suggestion
```
michaelcfanning, line 300: >true [](http://example.com/codeflow?start=43&length=4)

interestingly, this item was added when I explicitly used 'dotnet format' to get all autofixes I could from the command-line. This must be a tool feature, to add new diagnostics as they are added to the toolchain.

src/Sarif/LineInfo.cs
```diff
@@ -51,7 +51,7 @@ public LineInfo(int startOffset, int lineNumber)
         /// <returns>true if the objects are considered equal, false if they are not.</returns>
         public override bool Equals(object obj)
         {
-            LineInfo? other = obj as LineInfo?;
```
michaelcfanning, line 54: >LineInfo [](http://example.com/codeflow?start=12&length=8)

here it is friends, the application of our var standard. we do not require the type to be expressed if the type is already quite explicit. we do this to reduce the # of chars on a line to and allow left-aligning var declarations nicely for this circumstance.

src/Sarif/Sarif.csproj
```diff
@@ -99,7 +100,13 @@
       <DependentUpon>SdkResources.resx</DependentUpon>
     </Compile>
   </ItemGroup>
-
-  <Import Project="ToDotNet\ToDotNet.targets" />
+  
+<!-- Disabling the automatic code generation. This code emit is generating
```
michaelcfanning, line 104: >Disabling [](http://example.com/codeflow?start=5&length=9)

Everyone: please note this change. The SARIF OM has been stable for years now. We don't need to keep updating it. We can't enforce stringent style guidelines on that code if the autogen keeps overwriting changes again (and I don't feel like investing in fixing the JSchema codegen for these style issues, though that would be nice if someone did it).

src/Sarif/Core/Notification.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#if DEBUG
```
michaelcfanning, line 4: >DEBUG [](http://example.com/codeflow?start=4&length=5)

Please note this distinction you may encounter, I typically build as DEBUG in the IDE, our command-line is release. And so you may see discrepancies in style issues between IDE and command-line builds.

suvamM (approved):

PR #2819, opened 2024-08-21
Fix FileNotFoundException when a file name or directory path contains URL-encoded characters

shaopeng-gh:

We use `Uri` class. When the path contains URL-encoded character, it will automatically decode, and the only reliable property is the  `OriginalString`: 

image

Note that file name can contain `%2D` and it can contain `-` as well.

This fix the disk read related to use `OriginalString`.

Note that we have many other places that use other than `OriginalString`: `LocalPath`, `AbsoluteUri` etc.

### DIFF
--- a/src/Sarif/EnumeratedArtifact.cs
+++ b/src/Sarif/EnumeratedArtifact.cs
@@ -86,7 +86,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                 }
 
                 // This is our client-side, disk-based file retrieval case.
-                this.Stream = FileSystem.FileOpenRead(Uri.LocalPath);
+                this.Stream = FileSystem.FileOpenRead(Uri.OriginalString);
             }
 
             RetrieveDataFromStream();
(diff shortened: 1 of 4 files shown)


src/Test.UnitTests.Sarif.Driver/Sdk/AnalyzeCommandBaseTests.cs
```diff
@@ -831,14 +831,25 @@ public Run AnalyzeFile(string fileName,
...
+            string content = "foo foo";
+            var mockFileSystem = new Mock<IFileSystem>();
+            mockFileSystem.Setup(x => x.IsSymbolicLink(path)).Returns(false);
+            mockFileSystem.Setup(x => x.FileStreamLength(path)).Returns(content.Length);
+            mockFileSystem.Setup(x => x.FileInfoLength(path)).Returns(content.Length);
+            mockFileSystem.Setup(x => x.FileReadAllText(path)).Returns(content);
+            mockFileSystem.Setup(x => x.FileOpenRead(path)).Returns(new MemoryStream(Encoding.UTF8.GetBytes(content)));
 
-            var target = new EnumeratedArtifact(FileSystem.Instance) { Uri = uri, Contents = "foo foo" };
+            var target = new EnumeratedArtifact(mockFileSystem.Object)
```
shaopeng-gh, line 849: >mockFileSystem.Object [](http://example.com/codeflow?start=48&length=21)

Before code fix, the updated test will trigger the exception.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -164,9 +164,9 @@ public long? SizeInBytes
                 }
                 else if (Uri != null && Uri.IsAbsoluteUri && Uri.IsFile)
                 {
-                    this.sizeInBytes = FileSystem.IsSymbolicLink(Uri.LocalPath)
-                        ? (long)FileSystem.FileStreamLength(Uri.LocalPath)
-                        : (long)FileSystem.FileInfoLength(Uri.LocalPath);
```
michaelcfanning, line 169: >LocalPath [](http://example.com/codeflow?start=62&length=9)

OK, well. LocalPath unescapes, I see.

[https://learn.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=net-8.0](https://learn.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=net-8.0)

> The value returned by this property is unescaped. If the path is recognized as a Windows file path, all forward slashes (/) are replaced by backward slashes (\).

michaelcfanning (approved):
:shipit:

PR #2831, opened 2024-10-25
Consolidate PR validation into one file with multiple jobs

nguerrera:
Turns out the job name is the only things that shows in mobile and branch protection. Worse, if you try to protection when there are duplicate job names, it doesn't seem to work.

This consolidates everything for PR into one .yml and gives the jobs names. I have already set the branch protection and you can see that the newly consolidated jobs are required.


nguerrera:
Possibly unclear from the title and description. Other PRs will be unmergeable until this one merges to sync with changes made in GitHub admin backend.

suvamM:
@nguerrera The problem this PR is trying to solve isn't immediately clear to me. I see all the existing jobs are present in the consolidated file though, so I won't block this.

suvamM (approved):

nguerrera:
For posterity:

The blocking issue was that both build-and-test and build-multitool-for-npm had jobs named simply "run". It turns out that the job name must be unique. And in some contexts, branch protection and mobile github app, only the job name shows. So you would see only "Run" and not build-and-test and build-multitool-for-npm. Worse, you could not require these duplicate "run" jobs.

I decided that putting everything in one file is a simple way to guarantee that these names are always good and unique. :)

nguerrera:
I put these details in the commit description too.

PR #2837, opened 2024-10-29
Enable core function for archive enumeration/scanning.

michaelcfanning:


src/Sarif/ArtifactProvider.cs
```diff
@@ -1,24 +1,89 @@
...
+
+        public virtual bool IsExcluded(IEnumeratedArtifact artifact)
+        {
+            if (artifact == null) { throw new ArgumentNullException(nameof(artifact)); }
+
+            string extension = Path.GetExtension(artifact.Uri.LocalPath);
+
+            if (ExtensionsAllowList != null)
+            {
+                return !ExtensionsAllowList.Contains(extension);
```
nguerrera, line 70: It's surprising to me that the deny list is ignored if there is an allow list.

If users are required to use one or the other but not both, then perhaps we should check that they don't try to use both.

src/Sarif/HashData.cs
```diff
@@ -1,3 +1,4 @@
+
 // Copyright (c) Microsoft. All rights reserved.
```
nguerrera, line 2: Revert

src/Sarif/SarifExtensionMethods.cs
```diff
@@ -253,12 +253,20 @@ public static string GetFileName(this Uri uri)
                 return newAbsoluteUri.AbsolutePath.Split('/').Last();
             }
 
+            if (uri.Query != null)
+            {
+                return uri.Query;
```
nguerrera, line 258: Why would the query information be considered the file name?

src/Sarif/ZipArchiveArtifact.cs
```diff
@@ -19,14 +19,17 @@ public class ZipArchiveArtifact : IEnumeratedArtifact
...
         {
             this.entry = entry ?? throw new ArgumentNullException(nameof(entry));
             this.archive = archive ?? throw new ArgumentNullException(nameof(archive));
 
             this.binaryExtensions = binaryExtensions ?? new HashSet<string>();
-            this.uri = new Uri(entry.FullName, UriKind.RelativeOrAbsolute);
-        }
+
+            this.uri = baseUri != null
+                ? new Uri($"{baseUri}?{entry.FullName}")
```
nguerrera, line 30: >?{entry.FullName}" [](http://example.com/codeflow?start=37&length=18)

Oh, I see why. Yikes. Is this is a workaround for the issues in the other PR?

src/Sarif/ZipArchiveArtifact.cs
```diff
@@ -50,7 +53,14 @@ public Stream Stream
 
                 lock (this.archive)
                 {
-                    return entry.Open();
+                    try
+                    {
+                        return entry.Open();
+                    }
+                    catch (InvalidDataException)
```
nguerrera, line 60: >InvalidDataException [](http://example.com/codeflow?start=27&length=20)

Do we not need to report this anywhere?

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -666,6 +667,32 @@ private async Task<bool> EnumerateFilesFromArtifactsProvider(TContext globalCont
...
+                    stream ??= new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+                    ZipArchive archive;
+
+                    try
+                    {
+                        archive = new ZipArchive(stream, ZipArchiveMode.Read, leaveOpen: true);
+                    }
+                    catch(InvalidDataException)
+                    {
+                        // TBD log corrupt zip file.
```
nguerrera, line 682: >TBD [](http://example.com/codeflow?start=27&length=3)

I guess the same TBD applies in the other catch I just commented on. Can you at least comment there too?

nguerrera:
I just saw that this is a draft. Silly GitHub sent me email about it. Sorry if you weren't ready for feedback.

PR #2840, opened 2024-11-18
Make the build compatible with .NET 9 SDK

nguerrera:
* Move all warning-as-error handling to build props
* Disable warning-as-error in Debug builds
* Lower AnalysisLevel to .NET 8 wave when treating warnings as errors in Release
* Pass --self-contained explicitly to dotnet publish with runtime identifier for forward compat

### DIFF
--- a/scripts/BuildMultitoolForNpm.ps1
+++ b/scripts/BuildMultitoolForNpm.ps1
@@ -38,7 +38,7 @@ $npmBuildFolder = "$BuildRoot\Publish\npm"
 if (-not $SkipBuild) {
     Write-Information "Building Sarif.Multitool for Windows, Linux, and MacOS..."
     foreach ($runtime in "win-x64", "linux-x64", "osx-x64") {
-        dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.1 -r $runtime
+        dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.1 -r $runtime --self-contained
     }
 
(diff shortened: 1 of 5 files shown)


scripts/BuildMultitoolForNpm.ps1
```diff
@@ -38,7 +38,7 @@ $npmBuildFolder = "$BuildRoot\Publish\npm"
 if (-not $SkipBuild) {
     Write-Information "Building Sarif.Multitool for Windows, Linux, and MacOS..."
     foreach ($runtime in "win-x64", "linux-x64", "osx-x64") {
-        dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.1 -r $runtime
+        dotnet publish $SourceRoot\$project\$project.csproj -c $Configuration -f netcoreapp3.1 -r $runtime --self-contained
```
nguerrera, line 41: This fixes a .NET 9 SDK warning that tells us that --self-contained will no longer be the default as before when runtime identifier is specified.

src/Directory.Solution.props
```diff
@@ -0,0 +1,3 @@
+<Project>
+  <Import Project="build.warnings_as_errors.props" />
```
nguerrera, line 2: It is maddeningly difficult to get the same behavior for building .sln as .csproj for these settings.

joscol (approved):

src/Directory.Solution.props
michaelcfanning: indeed!

michaelcfanning (approved):
:shipit:

PR #2852, opened 2025-02-19
handle ambiguous file/directory references

pbhasker-msft:
This PR improves our handling of ambiguous file and directory references by introducing a robust retry mechanism. In summary, the changes include:

- **Enhanced Resolution:** When a file specifier does not correspond to an existing file, the system now checks if the specifier actually points to a directory.
- **Directory Scanning:** If the specifier is indeed an existing directory, the scanner will open and enumerate its contents instead of treating it as a missing file.
- **URI Flexibility:** Provisions have been added to correctly interpret both relative and absolute URI paths, ensuring that file and directory references are resolved reliably regardless of their format.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -11,10 +11,10 @@
 * BUG: Resolve process hangs when a file path is provided with a wildcard, but without a `-r` (recurse) flag during the multi-threaded analysis file enumeration phase.
 * BUG: Fix error `ERR997.NoValidAnalysisTargets` when scanning symbolic link files.
 * BUG: Fix `ERR999.UnhandledEngineException: System.IO.FileNotFoundException: Could not find file` when a file name or directory path contains URL-encoded characters.
+* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there are ambiguous file/directory references.
 * NEW: Allow null archive uri in `MultithreadedZipArchiveArtifactProvider` (which indicates that enumerated artifact paths should not include the base archive).
 * NEW: Update `LogTargetParseError(IAnalysisContext, Region, string, Exception)` to include optional exception argument to denote code location where parse error occurred.
 * NEW: `MultithreadedAnalyzeCommandBase.EnumerateArtifact` now supports scanning into compressed (OPC) files. Initial support file extensions are: `.apk`, `.appx`, `.appxbundle`, `.docx`, `.epub`, `.jar`, `.msix`, `.msixbundle`, `.odp`, `.ods`, `.odt`, `.onepkg`, `.oxps`, `.pkg`, `.pptx`, `.unitypackage`, `.vsdx`, `.xps`, `.xlsx`, `.zip`.
-* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there are ambiguous file/directory references.
 
 ## **v4.5.4 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.4) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.4) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.4)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.4) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.4)
 * BUG: Fix incorrect base class in rule ADO2012.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -11,7 +11,7 @@
 * BUG: Resolve process hangs when a file path is provided with a wildcard, but without a `-r` (recurse) flag during the multi-threaded analysis file enumeration phase.
 * BUG: Fix error `ERR997.NoValidAnalysisTargets` when scanning symbolic link files.
 * BUG: Fix `ERR999.UnhandledEngineException: System.IO.FileNotFoundException: Could not find file` when a file name or directory path contains URL-encoded characters.
-* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there are ambiguous file/directory references.
+* BUG: Fix error `ERR997.NoValidAnalysisTargets` when ambiguous file/directory references are provided to `OrderedFileSpecifier`. Previously, the code required an explicit directory separator to be added to the end of a directory path. Now, the code inspects the file system and assumes that a reference to an existing directory was intended by the user (even without a trailing separator).
 * NEW: Allow null archive uri in `MultithreadedZipArchiveArtifactProvider` (which indicates that enumerated artifact paths should not include the base archive).
 * NEW: Update `LogTargetParseError(IAnalysisContext, Region, string, Exception)` to include optional exception argument to denote code location where parse error occurred.
 * NEW: `MultithreadedAnalyzeCommandBase.EnumerateArtifact` now supports scanning into compressed (OPC) files. Initial support file extensions are: `.apk`, `.appx`, `.appxbundle`, `.docx`, `.epub`, `.jar`, `.msix`, `.msixbundle`, `.odp`, `.ods`, `.odt`, `.onepkg`, `.oxps`, `.pkg`, `.pptx`, `.unitypackage`, `.vsdx`, `.xps`, `.xlsx`, `.zip`.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -14,6 +14,7 @@
 * NEW: Allow null archive uri in `MultithreadedZipArchiveArtifactProvider` (which indicates that enumerated artifact paths should not include the base archive).
 * NEW: Update `LogTargetParseError(IAnalysisContext, Region, string, Exception)` to include optional exception argument to denote code location where parse error occurred.
 * NEW: `MultithreadedAnalyzeCommandBase.EnumerateArtifact` now supports scanning into compressed (OPC) files. Initial support file extensions are: `.apk`, `.appx`, `.appxbundle`, `.docx`, `.epub`, `.jar`, `.msix`, `.msixbundle`, `.odp`, `.ods`, `.odt`, `.onepkg`, `.oxps`, `.pkg`, `.pptx`, `.unitypackage`, `.vsdx`, `.xps`, `.xlsx`, `.zip`.
+* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there are ambiguous file/directory references.
 
 ## **v4.5.4 [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v4.5.4) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v4.5.4) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v4.5.4)  | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v4.5.4) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v4.5.4)
 * BUG: Fix incorrect base class in rule ADO2012.
--- a/src/Sarif.Driver/OrderedFileSpecifier.cs
+++ b/src/Sarif.Driver/OrderedFileSpecifier.cs
@@ -52,8 +52,20 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 normalizedSpecifier = uri.GetFilePath();
             }
 
-            string filter = Path.GetFileName(normalizedSpecifier);
-            string directory = Path.GetDirectoryName(normalizedSpecifier);
+            string directory = string.Empty;
+            string filter = string.Empty;
+
+            if (FileSystem.DirectoryExists(normalizedSpecifier))
+            {
+                // Check if the specifier itself is a directory
+                directory = normalizedSpecifier;
+                filter = "*";
+            }
+            else
+            {
+                filter = Path.GetFileName(normalizedSpecifier);
+                directory = Path.GetDirectoryName(normalizedSpecifier);
+            }
 
             if (directory.Length == 0)
             {
--- a/src/Sarif.Driver/OrderedFileSpecifier.cs
+++ b/src/Sarif.Driver/OrderedFileSpecifier.cs
@@ -52,20 +52,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                 normalizedSpecifier = uri.GetFilePath();
             }
 
-            string directory = string.Empty;
-            string filter = string.Empty;
-
-            if (FileSystem.DirectoryExists(normalizedSpecifier))
-            {
-                // Check if the specifier itself is a directory
-                directory = normalizedSpecifier;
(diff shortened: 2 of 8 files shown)


pbhasker-msft:
> @pbhasker-msft please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

src/Test.UnitTests.Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBaseTests.cs
```diff
@@ -1571,6 +1576,53 @@ private static Notification BuildTestNotification(Uri uri)
...
         }
 
+        [Fact]
+        public void MultithreadedAnalyzeCommandBase_DirectoriesAreEnumerated()
+        {
+            string testDirectory = "TestDirectory";
+            string testFile = "TestFile.txt";
+            string directory = Path.GetDirectoryName(GetThisTestAssemblyFilePath());
+
+            // Create test dictionary
```
michaelcfanning, line 1586: >dictionary [](http://example.com/codeflow?start=27&length=10)

NIT: please always add a period to the end of sentences in code comments.

src/Test.UnitTests.Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBaseTests.cs
```diff
@@ -1571,6 +1576,53 @@ private static Notification BuildTestNotification(Uri uri)
             };
         }
 
+        [Fact]
+        public void MultithreadedAnalyzeCommandBase_DirectoriesAreEnumerated()
```
michaelcfanning, line 1580: >MultithreadedAnalyzeCommandBase_DirectoriesAreEnumerated [](http://example.com/codeflow?start=20&length=56)

This test doesn't really communicate your approach. I think what you're doing here is creating a wide range of high-level command-line arguments that always ensure that TestFile.txt is found and scanned, yes? 

If so, please put a code comment at the beginning that describes the test approach. A good test should help onboard the next developer who comes along when that test unexpectedly breaks.

This test has another problem, it breaks on the first failure.

Look at line 1708 for an example of a pattern in this file intended to overcome that issue. I will say that in general this code isn't following current best practices/patterns, but that isn't on you, it's a legacy code base.

In other repositories, where we've intended to land is to use fluent assertions and assertion scopes to get the behaviors that we want. You can see another pattern in this file that we actually have avoided mostly, the use of xUnit [Theory]. This is an elegant test pattern that has caused measurable performance degradation for us. This was observed several years ago, and no one has completed the analysis to see whether the problem has been fixed.

src/Sarif.Driver/OrderedFileSpecifier.cs
```diff
@@ -55,6 +55,12 @@ private IEnumerable<IEnumeratedArtifact> EnumeratedArtifacts()
             string filter = Path.GetFileName(normalizedSpecifier);
             string directory = Path.GetDirectoryName(normalizedSpecifier);
 
+            if (filter != "*" && filter != "" && !FileSystem.FileExists(normalizedSpecifier))
```
michaelcfanning, line 58: >filter [](http://example.com/codeflow?start=16&length=6)

You have made an important change to a low-level class but you haven't authored unit tests targeting this code directly, please do that.

As a rule, it's not a great idea to always have testing originate with your higher-level abstractions, as you have done here. I want to emphasize that your high-level testing in the analysis command looks pretty exhaustive from a code and data coverage pattern. The problem here is around future flexibility in this code. Imagine that we want to lift-and-shift this code elsewhere. Now the OrderedFileSpecifierTests, when copied, will be incomplete when they are moved as well.

michaelcfanning, line 58: >filter [](http://example.com/codeflow?start=33&length=6)

Under what circumstances will `Path.GetFileName` return an empty string? Did you observe one?

src/Test.UnitTests.Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBaseTests.cs
```diff
@@ -1571,6 +1576,53 @@ private static Notification BuildTestNotification(Uri uri)
...
+            string filePath = Path.Combine(testDirectoryPath, testFile);
+            if (!File.Exists(filePath))
+            {
+                File.WriteAllText(filePath, $"{Guid.NewGuid()}");
+            }
+
+            var testCases = new List<string>
+            {
+                // Absolute directory paths
+                testDirectoryPath,
```
michaelcfanning, line 1600: >testDirectoryPath [](http://example.com/codeflow?start=16&length=17)

I don't see any tests here using the standard windows directory separate, `\`

src/Test.UnitTests.Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBaseTests.cs
```diff
@@ -1571,6 +1576,53 @@ private static Notification BuildTestNotification(Uri uri)
...
+
+            // Create test dictionary
+            string testDirectoryPath = Path.Combine(directory, testDirectory);
+            Directory.CreateDirectory(testDirectoryPath);
+
+            // Create test file
+            string filePath = Path.Combine(testDirectoryPath, testFile);
+            if (!File.Exists(filePath))
+            {
+                File.WriteAllText(filePath, $"{Guid.NewGuid()}");
```
michaelcfanning, line 1594: >WriteAllText [](http://example.com/codeflow?start=21&length=12)

your test approach is leaving test artifacts behind and this isn't a good practice. instead, use a built-in temporary file that autodeletes, or at least add a try/finally block that ensures everything is cleaned up (even if there's an unhandled exception). there are patterns for this in this file, take a look at calls to `Path.GetTempFileName`

src/Test.UnitTests.Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBaseTests.cs
```diff
@@ -1571,6 +1576,53 @@ private static Notification BuildTestNotification(Uri uri)
...
+
+                // Relative file paths
+                Path.Combine(testDirectory, testFile),
+                "./" + Path.Combine(testDirectory, testFile),
+                $"{testDirectory}//{testFile}",
+            };
+
+            foreach (string testCase in testCases)
+            {
+                AnalyzeFile(testCase);
```
michaelcfanning, line 1622: >AnalyzeFile [](http://example.com/codeflow?start=16&length=11)

Add a comment to this like so:

// AnalyzeFile will ensure that a scan context was created for
// 'TestFile.txt`, indicating that it was successfully found at the
//  file enumeration phase (driven by 'OrderedFileSpecifier').

src/Sarif.Driver/OrderedFileSpecifier.cs
joscol, line 58: >FileExists [](http://example.com/codeflow?start=61&length=10)

What does this do with symbolic links and junction files?

You might need to do something like this: [https://stackoverflow.com/questions/1485155/check-if-a-file-is-real-or-a-symbolic-link](https://stackoverflow.com/questions/1485155/check-if-a-file-is-real-or-a-symbolic-link)

michaelcfanning: Of course, makes sense! Just do a length check here against 0, `filter.Length != 0', You could perform a comparison against the `string.Empty` constant as well (but the length check is more direct). It used to be that we expressly forbade use of `""` in code as it would result in a string literal compiled into MSIL where .NET already defines one (that's interned) in `string.Empty`. In practice, all this concern around the empty string may be a bit of theater. @nguerrera may have some comments.

In generally, comparisons to the empty string are interesting because this operation might mask a need for expanded comparison semantics, for example, checks for whitespace literals. That isn't the case here as we're exclusively working against the return values of `Path.GetFileName`.

pbhasker-msft: symbolic links / junction files are resolved correctly. Added unit tests to verify this behavior.

pbhasker-msft: Added a new OrderedFileSpecifierTests.cs file and moved UTs there.

src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            sb.Length.Should().Be(0, because: "all artifacts should be enumerated but the following cases failed." + Environment.NewLine + sb.ToString());
+        }
+
+        private void CreateSymbolicLink(string linkPath, string targetPath, string targetType)
+        {
+            var processStartInfo = new System.Diagnostics.ProcessStartInfo
+            {
+                FileName = "cmd.exe",
```
michaelcfanning, line 193: >cmd.exe [](http://example.com/codeflow?start=28&length=7)

I'm not super comfortable with tests that are invoking standalone processes.

can you describe what behavior you observed with symbolic links? if everything 'just works' then I'm not sure we need explicit tests for this.

michaelcfanning: To be clear, if you've empirically observed the .NET handles these conditions correctly and without exceptions, we don't require explicit tests. .NET has it covered.

you could add a comment to the unit test that mentions the approach works for symbolic links, in case someone wonders on inspecting the code.

src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
```diff
@@ -0,0 +1,204 @@
...
+                Directory.CreateDirectory(childDirectory);
+
+                for (int j = 0; j < 5; j++)
+                {
+                    string file = Path.Combine(childDirectory, "TestFile" + j.ToString() + ".txt");
+                    File.WriteAllText(file, "12345");
+                }
+            }
+        }
+        private static string GetThisTestAssemblyFilePath()
```
michaelcfanning, line 44: >static [](http://example.com/codeflow?start=16&length=6)

You can just use Path.GetTempFileName, and then if you call Path.GetDirectoryName on it, you have a good temp dir.

michaelcfanning:
>             foreach (string childDirectory in FileSystem.DirectoryEnumerateDirectories(directory, "*", SearchOption.TopDirectoryOnly))

If there's an issue with symbolic links/reparse points in our code, I tend to think it would surface here. Per Josh' point.

If you have time, create a manual test, Create a symlink in a child directory that points to its parent.

At least one danger here is endless file enumeration because we don't understand we've been in the directory before. 

@joscol based on your experience, what other use cases do we need to be watching out for?

---
Refers to: src/Sarif.Driver/OrderedFileSpecifier.cs:181 in 3668b00. [](commit_id = 3668b00451ca26b37b6ef8396ed8af8c3e90ec4c, deletion_comment = False)

src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
pbhasker-msft: Using this to resolve relative URl paths.

joscol:
>             foreach (string childDirectory in FileSystem.DirectoryEnumerateDirectories(directory, "*", SearchOption.TopDirectoryOnly))

My initial concern was that `FileSystem.FileExists` might return true for a link and wouldn't be treated as a directory.

---
In reply to: [2679949559](https://github.com/microsoft/sarif-sdk/pull/2852#issuecomment-2679949559) [](http://example.com/codeflow?ancestors=2679949559)

---
Refers to: src/Sarif.Driver/OrderedFileSpecifier.cs:181 in 3668b00. [](commit_id = 3668b00451ca26b37b6ef8396ed8af8c3e90ec4c, deletion_comment = False)

pbhasker-msft:
Created a user story on our board to handle endless enumeration due symbolic links - https://dev.azure.com/mseng/1ES/_workitems/edit/2254488.

We will address the above issue in a separate PR after we triage.

michaelcfanning (approved):
:shipit:

ReleaseHistory.md
```diff
@@ -11,6 +11,7 @@
 * BUG: Resolve process hangs when a file path is provided with a wildcard, but without a `-r` (recurse) flag during the multi-threaded analysis file enumeration phase.
 * BUG: Fix error `ERR997.NoValidAnalysisTargets` when scanning symbolic link files.
 * BUG: Fix `ERR999.UnhandledEngineException: System.IO.FileNotFoundException: Could not find file` when a file name or directory path contains URL-encoded characters.
+* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there are ambiguous file/directory references.
```
michaelcfanning, line 14: >NoValidAnalysisTargets [](http://example.com/codeflow?start=25&length=22)

you did not note the type that was modified. the relevant change is to `OrderedFileSpecifier`. we include this information because a user with a specific bug, like hey, `OrderedFileSpecifier` doesn't handle ambiguous file/directory references properly will likely look for that type name in our change tracking.

Writing a good note is an art. Generally, you want to clearly describe the bad/fixed behavior, using names and evidence that a user is likely to search on. Then you want to describe the fix. 

One of the most obvious symptoms of this issues is that we required users consuming `Sarif.Driver` command-lines to explicitly add a directory slash to the argument, otherwise it would be interpreted as a file, not a directory. And that's really the most obvious thing a user might notice and want fixed, yes?

Here you are:

* BUG: Fix error `ERR997.NoValidAnalysisTargets` when there ambiguous file/directory references are provided to `OrderedFileSpecifier`. Previously, the code required an explicit directory separator to be added to the end of a directory path. Now, the code inspects the file system and will assume a reference to an existing directory is that was intended by the user (even without a trailing separator).

pbhasker-msft: Noted! That’s a great point—I'll be sure to include the relevant type names and clearly highlight symptoms and fixes in future notes :)

PR #2854, opened 2025-02-27
Fixing non-determinism in OrderedFileSpecifier test

pbhasker-msft:
This PR addresses intermittent failures in our file enumeration tests caused by **non-deterministic resolution of relative paths.**
 Previously, tests using relative paths derived from the fixture’s _RootDirectoryRelativePath_ were failing when the working directory was not as expected. Specifically, when the test suite was run until failure, relative paths such as ".\\{GUID}" or "{GUID}/TestDirectory0/TestFile0.txt" resolved incorrectly.

This PR makes the following changes to address the issue above:
- Updated the test method ResolveRelativeFileAndDirectoryPaths to explicitly set the current working directory to the parent directory of _fixture.RootDirectory (using GetThisTestAssemblyFilePath()), ensuring that relative paths resolve to the correct location.
- Wrapped the test iteration in a try/finally block to restore the original working directory after each test case.

These changes ensure that the relative paths are consistently and correctly resolved during test execution, eliminating the non-deterministic behavior observed previously.

### DIFF
--- a/src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
+++ b/src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
@@ -30,7 +30,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
 
             for (int i = 0; i < 5; i++)
             {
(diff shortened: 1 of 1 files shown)


michaelcfanning (approved):
:shipit:

PR #2855, opened 2025-02-27
Test utility file diffing enforcement of required result fingerprints

cfaucon:
Technical description:
Adding an optional capability to enforce a required result fingerprint set during file diffing test runs.

Change justification / impact:
No change to existing tests by default. This optional utility improvement will allow tool authors and consumers of the sarif-sdk to author file diffing tests that in addition to existing additional enforcement of output quality such as no tool notifications could ensure and fail tests on requiring a set of fingerprints exist for every result produced.

### DIFF
--- a/src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
+++ b/src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
@@ -233,6 +233,7 @@ namespace Microsoft.CodeAnalysis.Sarif
 
             var filesWithErrors = new List<string>();
             var filesResultNotMatch = new List<string>();
+            var filesMissingFingerprints = new List<string>();
 
             // Reify the list of keys because we're going to modify the dictionary in place.
             var keys = expectedSarifTextDictionary.Keys.ToList();
@@ -269,7 +270,7 @@ namespace Microsoft.CodeAnalysis.Sarif
                         {
                             if (actual.Runs[0].Results.Any(r => r.Fingerprints.ContainsKey(expectedFingerprintKey) == false))
                             {
-                                filesWithErrors.Add(key);
+                                filesMissingFingerprints.Add(key);
                                 break;
                             }
                         }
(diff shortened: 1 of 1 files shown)


src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
```diff
@@ -256,6 +262,20 @@ private void CompareActualToExpected(
                                                      expectedSarifTextDictionary[key],
                                                      out SarifLog actual);
 
+                    if (expectedResultFingerprintKeys != null &&
+                        actual != null &&
+                        actual.Runs[0].Results != null)
```
michaelcfanning, line 267: use this single line check:
 actual?.Runs?[0]?.Results != null

michaelcfanning: you could also set a result local to this value.

src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
```diff
@@ -256,6 +262,20 @@ private void CompareActualToExpected(
...
+                    {
+                        foreach (string expectedFingerprintKey in expectedResultFingerprintKeys)
+                        {
+                            if (actual.Runs[0].Results.Any(r => r.Fingerprints.ContainsKey(expectedFingerprintKey) == false))
+                            {
+                                filesMissingFingerprints.Add(key);
+                                break;
```
michaelcfanning, line 274: >break [](http://example.com/codeflow?start=32&length=5)

add your reporting up here, get rid of this break, describe the missing data.

cfaucon: Fixed to have complete set of missing fingerprints.

src/Test.Utilities.Sarif/FileDiffingUnitTests.cs
cfaucon: Updated based on recommendation

michaelcfanning (approved):
:shipit:

PR #2856, opened 2025-02-27
Correctly setting working directory in OrderedFileSpecifierTests

pbhasker-msft:
The unit test in OrderedFileSpecifier fails when it is a submodule in the SPMI repo (Failing test pipeline: https://dev.azure.com/mseng/1ES/_build/results?buildId=29658428&view=ms.vss-test-web.build-test-results-tab&runId=168189729&resultId=100040&paneView=debug)

The error indicates that when the test code attempts to set the current working directory using,
`Environment.CurrentDirectory = _fixture.GetThisTestAssemblyFilePath();`
the directory returned by GetThisTestAssemblyFilePath() doesn't actually exist at that moment. This can happen if:
- The path is in a temporary folder: The test assembly might be running from a temporary directory that has been cleaned up or isn't guaranteed to exist at runtime.
- The directory was deleted: If any part of the test fixture's lifecycle (or the test runner) cleans up temporary directories before this line executes, setting the current directory will fail.

To fix this, this PR chooses a different base directory that is known to persist throughout the test execution by using Path.GetTempPath();

Since we are now using the temp path, both test files `OrderedFileSpecifierTests` and `FileSpecifiertests` rely on modifying the current working directory and accessing temporary directories, and they interfere with each other when run concurrently.

To resolve this, we also add the test collection attribute `[Collection("TestsUsingCurrentDirectory")]` to ensure that these tests run sequentially, reducing nondeterministic failures and resource conflicts.

### DIFF
--- a/src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
+++ b/src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
@@ -24,7 +24,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
         public OrderedFileSpecifierTestsFixture()
         {
             RootDirectoryRelativePath = Guid.NewGuid().ToString();
-            RootDirectory = Path.Combine(GetThisTestAssemblyFilePath(), RootDirectoryRelativePath);
+            RootDirectory = Path.Combine(Path.GetTempPath(), RootDirectoryRelativePath);
 
             Directory.CreateDirectory(RootDirectory);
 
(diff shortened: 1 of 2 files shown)


src/Test.UnitTests.Sarif.Driver/Sdk/OrderedFileSpecifierTests.cs
```diff
@@ -24,7 +24,7 @@ public class OrderedFileSpecifierTestsFixture : IDisposable
         public OrderedFileSpecifierTestsFixture()
         {
             RootDirectoryRelativePath = Guid.NewGuid().ToString();
-            RootDirectory = Path.Combine(GetThisTestAssemblyFilePath(), RootDirectoryRelativePath);
+            RootDirectory = Path.Combine(Path.GetTempPath(), RootDirectoryRelativePath);
```
joscol, line 27: Adding `TempPath = Path.GetTempPath();` would eliminate the need to call it for each test.

michaelcfanning (approved):

PR #2859, opened 2025-03-28
[Tests] Make tests work for all autocrlf settings (on Windows)

Penguinwizzard:
Test input files, baseline files, and multiline strings get their line endings changed based on git's core.autocrlf config setting, while the text generated by the Sarif tools generally has line endings set using Environment.NewLine.

Included in this PR is fixes for all of the tests that failed when run in a checkout set up with core.autocrlf=input on a Windows machine.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,6 +10,9 @@
 ###############################################################################
 src/Sarif/Taxonomies/SampleCode.cs eol=lf
 
+# Pinned: this resource is embedded and hashed; line-ending drift would change the hash.
+src/Test.UnitTests.Sarif/TestData/InsertOptionalDataVisitor/InsertOptionalDataVisitor.txt text eol=crlf
+
 ###############################################################################
 # Set default behavior for command prompt diff.
 #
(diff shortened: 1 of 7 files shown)


michaelcfanning:
Pushed a reshape that moves the canonicalization to the read boundary in `TestAssetResourceExtractor.GetResourceText` (single `\r\n` -> `\n` -> `Environment.NewLine` step) and rewrites the affected literal-string assertions to express newlines explicitly via `string.Join(Environment.NewLine, ...)` or `$@"...{Environment.NewLine}..."`. `.gitattributes` keeps the `eol=crlf` pin on `InsertOptionalDataVisitor.txt` since it's embedded and hashed.

Local run is green: `Test.UnitTests.Sarif` 898/901, `Test.UnitTests.Sarif.Converters` 276/276 (skips unchanged). Validate workflow now triggers on `dev`-targeted events (separate fix on `dev`), so the next push should pick up CI.

michaelcfanning (approved):

PR #2866, opened 2025-06-30
Eliminate MD5 support as this algorithm suffers from security issues and triggers security alerts.

michaelcfanning:
* BRK: Remove `HashData.MD5`, `HashUtilities.ComputeMD5Hash` due to the inherent insecurity of this algorithm.
* BRK: 'HashUtilities.ComputeHash' no longer generates MD5 hashes (only SHA1 and SHA256).


nguerrera:
I'd be inclined to rip out SHA1 now too if we can get away with it.

PR #2867, opened 2025-07-23
Refactor invalid artifact checks to prevent unnecessary OPC file loads

pbhasker-msft:
**Problem:**
Previously, the checks for empty or oversized artifacts were only performed after verifying and loading the artifact as an OPC file. As a result, even artifacts that exceeded the maximum allowed size were unnecessarily opened and processed as OPC files.

**Solution:**
Refactored MultithreadedCommandBase to perform empty and max-size checks before attempting to load or identify OPC artifacts. This prevents unnecessary overhead and ensures that invalid artifacts are skipped earlier in the flow.

Behavior before update:
image

Behavior after update:
image


michaelcfanning (approved):
:shipit:

nguerrera (approved):

PR #2872, opened 2025-09-18
Add filtering of symbolic links to (Ordered)FileSpecifier

cfaucon:
Add filtering of symbolic links to (Ordered)FileSpecifier.

This will prevent file and directory traversal from getting into an infinite recursive loop when a child directory symbolic links to a location previously within the traversal tree.

### DIFF
--- a/src/Sarif.Driver/FileSpecifier.cs
+++ b/src/Sarif.Driver/FileSpecifier.cs
@@ -100,7 +100,12 @@ namespace Microsoft.CodeAnalysis.Sarif.Driver
                     {
                         foreach (string subdir in FileSystem.DirectoryGetDirectories(dir))
                         {
-                            AddFilesFromDirectory(subdir, filter);
+                            // Skip subdirectories that are symbolic links to prevent infinite loops
+                            // and avoid unintentionally following symbolic links during recursion
+                            if (!FileSystem.IsSymbolicLink(subdir))
+                            {
+                                AddFilesFromDirectory(subdir, filter);
+                            }
                         }
                     }
                     catch (UnauthorizedAccessException)
--- a/src/Sarif/FileSystem.cs
+++ b/src/Sarif/FileSystem.cs
@@ -417,8 +417,26 @@ namespace Microsoft.CodeAnalysis.Sarif
             // https://learn.microsoft.com/en-us/dotnet/api/system.io.fileattributes
             // While symbolic links will have the ReparsePoint flag set, not all reparse points represent symbolic links.
             // This is a basic implementation.
-            var fileInfo = new FileInfo(path);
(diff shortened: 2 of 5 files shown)


src/Sarif/FileSystem.cs
```diff
@@ -417,8 +417,26 @@ public bool IsSymbolicLink(string path)
             // https://learn.microsoft.com/en-us/dotnet/api/system.io.fileattributes
             // While symbolic links will have the ReparsePoint flag set, not all reparse points represent symbolic links.
             // This is a basic implementation.
-            var fileInfo = new FileInfo(path);
-            return (fileInfo.Attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint;
+            try
+            {
+                if (File.Exists(path))
```
michaelcfanning: >File [](http://example.com/codeflow?start=20&length=4)

maybe qualify these calls with `this.` for clarify?

src/Sarif/FileSystem.cs
```diff
@@ -417,8 +417,26 @@ public bool IsSymbolicLink(string path)
...
+                }
+                else if (Directory.Exists(path))
+                {
+                    var directoryInfo = new DirectoryInfo(path);
+                    return (directoryInfo.Attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint;
+                }
+
+                return false;
+            }
+            catch
```
michaelcfanning: kill this catch

michaelcfanning:
>         /// Uses <see cref="FileInfo"/> to determine whether a file is a symbolic link.

file or directory now yes? #Closed

---
Refers to: src/Sarif/FileSystem.cs:407 in 7c7761a. [](commit_id = 7c7761aa95e180510dc1560ea03f69483aa90cac, deletion_comment = False)

src/Sarif.Driver/FileSpecifier.cs
```diff
@@ -100,7 +100,12 @@ private void AddFilesFromDirectory(string dir, string filter)
                     {
                         foreach (string subdir in FileSystem.DirectoryGetDirectories(dir))
                         {
-                            AddFilesFromDirectory(subdir, filter);
+                            // Skip subdirectories that are symbolic links to prevent infinite loops
+                            // and avoid unintentionally following symbolic links during recursion
+                            if (!FileSystem.IsSymbolicLink(subdir))
```
nguerrera, line 105: Came here to say that you can use `EnumerationOptions.AttributesToSkip` to make the BCL file system enumeration do this for you. However, that API is only 7 years old so it's not available in .NET Framework / .NET Standard 2.0. 😔

nguerrera: No action item. Resolving.

michaelcfanning (approved):
:shipit:

PR #2874, opened 2025-09-23
Adding EscapeNonAscii when using SarifLogger

eddynaka:
## Context

The team is using SarifLogger to generate a SARIF after scanning some content. From time to time, we scan content containing emojis and other symbols. In the scanner, when retrieving the context region snippet, you might retrieve a partial character, and with that, you break it, generating a problem when writing.

## Exception

```
System.Text.EncoderFallbackException
  HResult=0x80070057
  Message=Unable to translate Unicode character \\uD83D at index -1 to specified code page.
  Source=System.Private.CoreLib
  StackTrace:
...
...
...
   at Microsoft.CodeAnalysis.Sarif.Writers.ResultLogJsonWriter.WriteResult(Result result) in C:\repos\sarif-sdk\src\Sarif\Writers\ResultLogJsonWriter.cs:line 223
   at Microsoft.CodeAnalysis.Sarif.Writers.SarifLogger.Log(ReportingDescriptor rule, Result result, Nullable`1 extensionIndex) in C:\repos\sarif-sdk\src\Sarif\Writers\SarifLogger.cs:line 382
   at Microsoft.CodeAnalysis.Sarif.SarifLoggerTests.SarifLogger_ShouldNotThrowWhenWritingInvalidSequences() in C:\repos\sarif-sdk\src\Test.UnitTests.Sarif\Writers\SarifLoggerTests.cs:line 1194
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

## Solution

To prevent an unhandled exception when using SarifLogger, by enabling the option `StringEscapeHandling = StringEscapeHandling.EscapeNonAscii`, the writer properly handles broken characters, allowing us to write properly.

## Tests

The below text is the message written retrieved from the SARIF to prove that the content was kept:
```json
"message": {
    "text": "ASCII: Hello! Café Español München Ελληνικά Кириллица 漢字かなカナ MusicalSymbol: 𝄞 Rocket: 🚀 Snowman: ☃ Grin: 😀𐀀𐀀�"
}
```

As you can see the last character is the broken one that was introduced.

### DIFF
--- a/src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
+++ b/src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
@@ -995,7 +995,6 @@ namespace Microsoft.CodeAnalysis.Sarif
                 {
                     ValidateLoggerForExclusiveOption(logger, loggingOption);
                 }
-                ;
 
                 // Validates overload that accepts any 
                 // TextWriter (for example, one instantiated over a
@@ -1009,7 +1008,6 @@ namespace Microsoft.CodeAnalysis.Sarif
                 {
                     ValidateLoggerForExclusiveOption(logger, loggingOption);
                 }
-                ;
             }
             catch (Exception e)
             {
--- a/src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
+++ b/src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
@@ -1234,17 +1234,5 @@ namespace Microsoft.CodeAnalysis.Sarif
             string logText = stringBuilder.ToString();
             return JsonConvert.DeserializeObject<SarifLog>(logText);
         }
-
-
-        internal static readonly ResultKindSet s_kinds = [ResultKind.Fail];
-        internal static readonly FilePersistenceOptions s_logFilePersistenceOptions = FilePersistenceOptions.None;
-
-        public FailureLevelSet FailureLevels => [FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note];
-        public OptionallyEmittedData DataToInsert => OptionallyEmittedData.ComprehensiveRegionProperties
-                                                                        | OptionallyEmittedData.ContextRegionSnippets
-                                                                        | OptionallyEmittedData.Guids
-                                                                        | OptionallyEmittedData.Hashes
-                                                                        | OptionallyEmittedData.RegionSnippets
-                                                                        | OptionallyEmittedData.RollingHashPartialFingerprints;
     }
 }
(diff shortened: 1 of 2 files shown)


nguerrera:
I am just curious, is it possible for the code that retrieves/creates context snippet to extend the context by one char in the presence of a surrogate at the boundary?

I think this is a good change either way so that we're not dependent on callers doing that, but it might be additionally good to avoid putting half a surrogate pair in the context too where possible.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -994,7 +994,8 @@ private void TestForLoggingOption(FilePersistenceOptions loggingOption)
                                                 kinds: BaseLogger.Fail))
                 {
                     ValidateLoggerForExclusiveOption(logger, loggingOption);
-                };
+                }
+                ;
```
nguerrera: I'm guessing some auto-formatting did this. It looks like the semicolon is an unnecessary empty statement. Let's remove it.

eddynaka: Sad. Let me do it.

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
```diff
@@ -1007,7 +1008,8 @@ private void TestForLoggingOption(FilePersistenceOptions loggingOption)
...
```
nguerrera: Here too.

eddynaka:
> I am just curious, is it possible for the code that retrieves/creates context snippet to extend the context by one char in the presence of a surrogate at the boundary?
> 
> I think this is a good change either way so that we're not dependent on callers doing that, but it might be additionally good to avoid putting half a surrogate pair in the context too where possible.

The way our code works is that when we find a match, to construct the context region, we get 100 characters to the left and to the right. It would be hard to predict if a given context depends on the extra character or not because we would either need to inspect each context region or handle somehow when writing (where the exception happens).

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
eddynaka: Looked again, we could actually remove those extra `;`

src/Test.UnitTests.Sarif/Writers/SarifLoggerTests.cs
eddynaka: Removed the character.

nguerrera (approved):

nguerrera:
> It would be hard to predict if a given context depends on the extra character or not 

Couldn't this code inspect the char 100 chars to the left and right to see if they are char.IsLow/HighSurrogate and use 101 instead?

cfaucon (approved):

PR #2883, opened 2026-01-27
Fix thread-safety issue in SarifLogger for concurrent Log() and Dispose() calls

aasim:
### Summary
Fixes a concurrency bug where calling `SarifLogger.Dispose()` while `Log()` operations are in-flight causes `InvalidOperationException: Collection was modified; enumeration operation may not execute`.

Fixes [**2250448**](https://dev.azure.com/mseng/1ES/_workitems/edit/2250448)

### Problem
When `SarifLogger.Dispose()` is called, it serializes internal collections (like `Tool.Driver.Rules`) via JSON serialization. If concurrent threads are still calling `Log()` and adding new rules to those collections, the enumeration fails with:
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeDictionary(...)
at Microsoft.CodeAnalysis.Sarif.Writers.ResultLogJsonWriter.WriteTool(Tool tool)
at Microsoft.CodeAnalysis.Sarif.Writers.SarifLogger.Dispose()
```

### Solution
Implemented thread-safety in `SarifLogger` using:
1. **Interlocked counter** (`_operationsInFlight`) to track active `Log()` calls
2. **SpinWait** in `Dispose()` to wait for in-flight operations to complete
3. **Lock object** (`_writeLock`) to serialize writes between `Log()` and `Dispose()`
4. **State machine** to prevent new `Log()` calls once dispose begins

### Changes
- **SarifLogger.cs**: Added thread-safety mechanism with `_threadSafeLoggingEnabled` field, `TryEnterOperation()`/`ExitOperation()` methods, and thread-safe `Dispose()`
- **AnalyzeOptionsBase.cs**: Added `--thread-safe-logging` command-line option (default: `false`)
- **AnalyzeContextBase.cs**: Added `ThreadSafeLogging` property to flow the setting
- **MultithreadedAnalyzeCommandBase.cs**: Wires the option to `SarifLogger` constructor
- **CoyoteTest/Program.cs**: Added stress tests to verify the fix, will remove before merging

### Feature Flag
The fix is controlled via:
- **CLI**: `--thread-safe-logging true|false` (default: `true`)
- **Code**: `new SarifLogger(..., threadSafeLoggingEnabled: true)`

This allows disabling the fix if it causes unexpected issues in production.

### Testing
Verified with stress tests that:
1. ✅ Bug reproduces with `threadSafeLoggingEnabled: false`
2. ✅ Bug is fixed with `threadSafeLoggingEnabled: true`
3. ✅ `Log()` after `Dispose()` throws `ObjectDisposedException`

### DIFF
--- a/src/CoyoteTest/Program.cs
+++ b/src/CoyoteTest/Program.cs
@@ -1,0 +1,540 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Channels;
+
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.CodeAnalysis.Sarif.Writers;
+using Microsoft.Coyote;
+using Microsoft.Coyote.Specifications;
+using Microsoft.Coyote.SystematicTesting;
+
+namespace CoyoteTest;
+
+/// <summary>
+/// Coyote tests to reproduce the concurrency bug documented in work item 2250448.
+/// 
+/// The bug manifests as:
+///   System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
+///   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
+///   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeDictionary(...)
+///   at Microsoft.CodeAnalysis.Sarif.Writers.ResultLogJsonWriter.WriteTool(Tool tool)
+///   at Microsoft.CodeAnalysis.Sarif.Writers.SarifLogger.Dispose()
+/// 
+/// The race condition occurs when:
+/// - Thread A: SarifLogger.Dispose() -> CompleteRun() -> WriteTool() -> serializes Tool.Properties
+/// - Thread B: Analysis thread modifies Tool.Properties or other shared collections
+/// </summary>
+public static class SarifLoggerConcurrencyTests
+{
+    public static async Task Main(string[] args)
+    {
+        Console.WriteLine("Running Coyote systematic tests for SarifLogger concurrency bug...");
+        Console.WriteLine("=" + new string('=', 70));
+
+        // Run all test methods
+        await RunTestAsync(nameof(DisposeDuringConcurrentToolPropertiesMutation), DisposeDuringConcurrentToolPropertiesMutation);
+        await RunTestAsync(nameof(DisposeDuringConcurrentInvocationEnvironmentMutation), DisposeDuringConcurrentInvocationEnvironmentMutation);
+        await RunTestAsync(nameof(DisposeDuringToolComponentPropertiesMutation), DisposeDuringToolComponentPropertiesMutation);
+        await RunTestAsync(nameof(DisposeDuringConcurrentToolMutation), DisposeDuringConcurrentToolMutation);
+        await RunTestAsync(nameof(DisposeDuringConcurrentResultsMutation), DisposeDuringConcurrentResultsMutation);
+        await RunTestAsync(nameof(SimulateMultithreadedAnalyzeCommandScenario), SimulateMultithreadedAnalyzeCommandScenario);
+        await RunTestAsync(nameof(RepeatedDisposeCyclesWithConcurrentMutations), RepeatedDisposeCyclesWithConcurrentMutations);
+
+        Console.WriteLine("=" + new string('=', 70));
+        Console.WriteLine("All Coyote tests completed.");
+    }
+
+    private static async Task RunTestAsync(string testName, Func<Task> test)
+    {
+        Console.WriteLine($"\n--- Running: {testName} ---");
+
+        var configuration = Configuration.Create()
+            .WithTestingIterations(1000)
+            .WithMaxSchedulingSteps(1000);
+
+        var engine = TestingEngine.Create(configuration, test);
+        engine.Run();
+
+        Console.WriteLine($"Iterations: {engine.TestReport.NumOfFoundBugs} bugs found");
+
+        if (engine.TestReport.NumOfFoundBugs > 0)
+        {
+            Console.WriteLine("BUG FOUND! Reproducible trace available.");
+            Console.WriteLine($"Error: {engine.TestReport.BugReports.FirstOrDefault()}");
+
+            // Generate replay trace
+            string traceFile = $"{testName}_replay.schedule";
+            engine.TryEmitReports(".", testName, out _);
+            Console.WriteLine($"Replay trace saved to: {traceFile}");
+        }
+        else
+        {
+            Console.WriteLine("No bugs found in this test.");
+        }
+
+        await Task.CompletedTask;
+    }
+
+    /// <summary>
+    /// Test 1: Race between Dispose serializing Tool.Properties and concurrent mutation.
+    /// 
+    /// Matches the reported diagram:
+    ///  - Thread A: SarifLogger.Dispose -> CompleteRun -> WriteTool -> serializes Tool.Properties
+    ///  - Thread B: Mutates Tool.Properties
+    /// </summary>
+    [Microsoft.Coyote.SystematicTesting.Test]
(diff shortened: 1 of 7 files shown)


src/CoyoteTest/Program.cs
```diff
@@ -0,0 +1,338 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis.Sarif;
```
aasim, line 4: Will remove these tests files before merging, right now have them only to do this test

aasim:
```
cd "\sarif-sdk\src\CoyoteTest" ; dotnet run
Thread-Safety Fix Verification Tests for SarifLogger
=======================================================================

This fix can be controlled via the --thread-safe-logging option.
In code: new SarifLogger(..., threadSafeLoggingEnabled: true/false)

>>> PHASE 1: Testing WITHOUT fix (threadSafeLoggingEnabled: false)
    Expected: Tests should FAIL with 'Collection was modified' errors


--- Running: StressTest_ConcurrentLogAndDispose (20 iterations) ---
✓ EXPECTED FAILURE: 20/20 iterations failed (bug confirmed!)

--- Running: StressTest_ManyThreadsLogThenDispose (10 iterations) ---
✓ EXPECTED FAILURE: 10/10 iterations failed (bug confirmed!)

>>> PHASE 2: Testing WITH fix (threadSafeLoggingEnabled: true)
    Expected: Tests should PASS


--- Running: StressTest_ConcurrentLogAndDispose (100 iterations) ---
✓ PASSED: All 100 iterations succeeded (fix working!)

--- Running: StressTest_ManyThreadsLogThenDispose (50 iterations) ---
✓ PASSED: All 50 iterations succeeded (fix working!)

--- Running: StressTest_DisposeDuringActiveLogs (100 iterations) ---
 [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=2, rejected=998] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000] [logged=0, rejected=1000] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=1, rejected=999] [logged=0, rejected=1000]✓ PASSED: All 100 iterations succeeded (fix working!)       

--- Running: StressTest_DisposeRejectsNewLogs (50 iterations) ---
✓ PASSED: All 50 iterations succeeded (fix working!)
```

src/Sarif/Writers/SarifLogger.cs
```diff
@@ -21,6 +22,14 @@ public class SarifLogger : BaseLogger, IDisposable, IAnalysisLogger
         private JsonTextWriter _jsonTextWriter;
         private ResultLogJsonWriter _issueLogJsonWriter;
 
+        private int _operationsInFlight;
+        private int _state; // 0 = Active, 1 = Disposing, 2 = Disposed
+        private const int StateActive = 0;
+        private const int StateDisposing = 1;
+        private const int StateDisposed = 2;
+        private readonly object _writeLock = new object();
+        private readonly bool _threadSafeLoggingEnabled;
```
nguerrera, line 31: I have not reviewed in detail, but an initial thought is to consider using `ReaderWriterLockSlim`. Dipsose() could take write lock while Log() takes read lock. I think we'd get similar performance, but the code would be easier to understand a harder to get wrong as we'd let ReaderWriterLockSlim handle the low-level details.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
```diff
@@ -159,5 +159,11 @@ public IEnumerable<RuleKind> RuleKindOption
         public RuleKindSet RuleKinds => RuleKindOption != null ?
             new RuleKindSet(RuleKindOption) :
             new RuleKindSet(new List<RuleKind>(new[] { RuleKind.Sarif }));
+
+        [Option(
+            "thread-safe-logging",
+            Default = false,
+            HelpText = "Enable thread-safe logging to prevent concurrent access issues during analysis.")]
+        public bool ThreadSafeLogging { get; set; } = false;
```
nguerrera, line 167: Do we need this to be an option? It feels like it should be on always with no option to turn off since it fixes a bug.

nguerrera: I see in the description this is to have a feature flag initially.

src/Sarif/Writers/SarifLogger.cs
aasim: I had originally thought of a readerwriterlockslim approach, but interestingly went this route to make it simpler :)

I like the slim lock approach as well, I think that is definitely a better option. I will update this PR shortly with that fix and publish again.

src/Sarif.Driver/Sdk/AnalyzeOptionsBase.cs
aasim: Yeah, I just want to make sure we don't regress by first testing this behind a FF in DevSecOps. Once done, will remove this flag and make it permanent

michaelcfanning: nah just fix the bug. :) a bug fix isn't a feature. run more/better stress locally if you're really worried about regressing the code.

PR #2894, opened 2026-05-11
Fix AccessViolationException in EnumeratedArtifact when caller stream's Seek re-enters native code (e.g. IIS HttpBufferlessInputStream)

aasim:
## Root cause

`EnumeratedArtifact.RetrieveDataFromStream` calls `Stream.Seek(0, Begin)` whenever `CanSeek == true`. For some "seekable" wrapper streams (e.g. WebAPI's `SeekableBufferedRequestStream` over IIS's `HttpBufferlessInputStream`), `Seek` re-enters the underlying source. When that source is a native IIS request whose context has been released, the re-entry dereferences freed memory and crashes the host process with `AccessViolationException` in `iiscore!NOTIFICATION_CONTEXT::DecrementOutstandingIO`.

Related ICM: [791503236](https://portal.microsofticm.com/imp/v5/incidents/details/791503236/summary)

## Fix

Always wrap the caller-provided stream in `PeekableStream` and rewind in-memory. `EnumeratedArtifact` no longer calls `Seek` on a caller-provided stream.

Regression tests use a `HostileSeekStream` that advertises `CanSeek == true` but throws `AccessViolationException` from `Seek`. They fail on `main`, pass after the fix.


michaelcfanning:
Hello! Can you explain why our code is attempting to work in a situation where some underlying context has been released? What is the guarantee that the rest of our scanning will succeed against this stream? i.e., you have explained the crash but haven't explained why things work if we avoid the AV

aasim:
> Hello! Can you explain why our code is attempting to work in a situation where some underlying context has been released? What is the guarantee that the rest of our scanning will succeed against this stream? i.e., you have explained the crash but haven't explained why things work if we avoid the AV

The fix doesn't guarantee the scan succeeds when the underlying context is gone. It guarantees the failure is recoverable.

Two cases after the fix:

* Source still alive: scan completes normally (same as today's happy path).
* Source already released: the forward Read throws a normal managed exception (ObjectDisposedException, IOException, etc.). That propagates up through the SARIF analyze loop, the controller catches it, the request fails with a 5xx, telemetry is flushed, and the worker stays alive for the next request.
Today the same case causes the AcessViolation exception on the process: no catch handler runs, no telemetry, the worker dies along with every other in-flight request on it.

So, this is not to fix a scan for a large file that would have failed regardless, this fix mainly converts an unrecoverable exception to a managed one that we can handle in the caller (with retries etc.), and more importantly the IIS host process survives to continue to process other requests.

On why the SDK is even attempting Seek in this state in the first place, it isn't, its just that it can't tell if the source stream is terminated with confidence. The Stream API gives no way to know that a native resource behind a managed wrapper has been released; CanSeek == true only means "Seek is supported," not "Seek is safe right now," and that unknowability is fine for normal streams (MemoryStream, FileStream) where nothing external can yank the resource between calls, but for Http streams this is not the case. The fix sidesteps the problem by never calling Seek on the caller's stream, so any latent unhealthy state surfaces as a normal catchable Read failure instead of a process-killing native crash.

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -109,17 +109,17 @@ private bool IsZipHeader(byte[] header, int length)
 
     private void RetrieveDataFromStream()
     {
-        if (!this.Stream.CanSeek)
-        {
-            this.Stream = new PeekableStream(this.Stream, BinarySniffingHeaderSizeBytes);
-        }
+        // Always rewind via PeekableStream; CanSeek can lie (e.g. WebAPI's SeekableBufferedRequestStream over IIS).
```
cfaucon: nit; reword lie to something like 'be inaccurate'

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -109,17 +109,17 @@ private bool IsZipHeader(byte[] header, int length)
...
+        var peekable = this.Stream as PeekableStream
```
cfaucon: Is there a technical reason to assign to peekable var first. and then to this.Stream? and calling rewind off of peekable var and not this.Stream?

src/Sarif/EnumeratedArtifact.cs
aasim: Fixed

src/Sarif/EnumeratedArtifact.cs
aasim: peekable has rewind(), stream doesn't so we have to call Rewind() on peekable. We set stream to peekable to get the buffered stream that is used in the rest of the code

src/Sarif/EnumeratedArtifact.cs
```diff
@@ -109,17 +109,17 @@ private bool IsZipHeader(byte[] header, int length)
 
     private void RetrieveDataFromStream()
     {
-        if (!this.Stream.CanSeek)
-        {
-            this.Stream = new PeekableStream(this.Stream, BinarySniffingHeaderSizeBytes);
-        }
+        // Always rewind via PeekableStream; CanSeek can be inaccurate (e.g. WebAPI's SeekableBufferedRequestStream over IIS).
+        var peekable = this.Stream as PeekableStream
```
michaelcfanning: It seems like this isn't the bug. This stream was in fact seekable, it's just that *attempting* the seek for a dispose thing is causing an AV. right? what you're really doing here is preventing an AV by not calling seek not preventing an AV on CanSeek right? Mostly CanSeek refers to the semantics of the class when the object is not disposed. this value typically doesn't change (AFAIK!!) if the object itself isn't usable. that's my assumption anyway.

michaelcfanning: OK, so we discussed this, a better fix (to avoid the memory increase implied by what we have here) is to examine the type of the object and only make it use peekable stream as required. we might not have (or want to link to) a dep that provides us a simple type comparison. in which case object.GetType().FullName would be what we need to examine.

aasim: @michaelcfanning updated the PR to have a allowlist of streams that will be used as is, anything else will be wrapped in a Peekable

michaelcfanning (approved):

PR #2904, opened 2026-05-17
Static CWE taxonomy: ship MITRE data offline, drop network reach

michaelcfanning:
Replaces the network-fetching `CweNameResolver` with an offline lookup backed by a SARIF-§3.19 taxonomy file generated once from MITRE's authoritative CWE catalog. **Stacked on top of #2903** (MCP emit test fixture).

## Why

The previous resolver made live HTTP calls to `cwe.mitre.org` per CWE-id at runtime to scrape page titles into PascalCase rule names. That's needless network reliance for a slowly-changing enumeration: every scan paid latency, every CI run risked flake, every reproduction varied if MITRE retitled an entry. CWE-id-to-name is static text and should ship statically.

### 1. `src/Sarif/Taxonomies/CWE.sarif` -- a first-class SDK asset

A SARIF 2.1.0 log file whose single run carries a CWE taxonomy per 3.19. 944 entries from MITRE's CWE View-1000 (Research Concepts), each with:
- `id` (numeric, e.g. `"78"`)
- `name` -- the MITRE canonical name (preserved unchanged for fidelity)
- `shortDescription.text` -- one-paragraph description
- `fullDescription.text` -- extended description when available
- `helpUri` -- `https://cwe.mitre.org/data/definitions/{id}.html`
- `properties.cwe/status`, `properties.cwe/abstraction` -- catalog metadata so consumers can filter further

Attribution to MITRE preserved on the toolComponent (`organization`, `informationUri`, `downloadUri`). CWE is published under terms permitting redistribution with attribution.

**Reusable beyond the MCP server** -- any SARIF producer (in any language) can reference this taxonomy via `run.taxonomies[]` or `externalPropertyFileReferences.taxonomies[]` and stop maintaining its own CWE name table.

### 2. `CweNameResolver` -- offline lookup

Loads the taxonomy from an embedded resource at first use, builds an in-memory `Dictionary<string, string>` mapping `id` to PascalCase name. `HttpClient` gone. Regex gone. Network reliance gone.

PascalCase derivation preserves the prior behavior (truncate at the first `(` to elide the parenthetical alternate-name suffix). CWE-78 still resolves to `ImproperNeutralizationOfSpecialElementsUsedInAnOsCommand` rather than the doubled `...OsCommandOsCommandInjection` that a raw whole-name PascalCase would produce.

## `scripts/Generate-CweTaxonomy.ps1`

Self-contained PowerShell script:
1. Downloads the official CWE View-1000 CSV from `cwe.mitre.org`
[...]

PR #2905, opened 2026-05-17
MCP reliability: atomic finalize + create-run overwrite guard

michaelcfanning:
Two reliability hardenings on the MCP emit surface, both addressing concrete ways the current code could leave the user with a corrupt or silently-clobbered scan output. **Stacked on #2904** (CWE static taxonomy).

## 1. Atomic finalize (`SarifLogWriter.Save`)

**Before**: `File.Create(outputPath)` + `StreamWriter` + `JsonSerializer.Serialize`. Any mid-write fault -- process crash, disk full, serializer exception, network share hiccup on the target volume -- left a partial-and-invalid SARIF file at the canonical path. Readers ingesting that path would see a broken JSON document with no signal as to whether it was finished.

**After**: serialize to `<outputPath>.tmp-<guid>`, then `File.Move(temp, outputPath, overwrite: true)`. The atomic-rename pattern guarantees readers observe either the prior contents or the new contents -- never a partial write. On serializer failure the temp file is best-effort cleaned and the target is untouched.

## 2. Create-run overwrite guard (`CreateRunTool.CreateRun`)

**New parameter**: `allowOverwrite` (default `false`). When `false` and the resolved output path already exists, `sarif_create_run` throws `InvalidOperationException` *before* registering the run in the store. Prevents the accidental second-invocation clobber (re-running a scan over a prior result file without explicit intent). When `true`, the call proceeds and the eventual atomic finalize overwrites cleanly.

## Tests: 46/46 (4 new) under `/p:EnforceCodeStyleInBuild=true`

`Test.UnitTests.Sarif.Mcp.Server/ReliabilityTests.cs`:
- `CreateRun_FailsFast_IfOutputExists_AndAllowOverwriteFalse`
- `CreateRun_Succeeds_IfOutputExists_AndAllowOverwriteTrue`
- `Finalize_LeavesNoTempFilesBehind` (pins the sidecar-temp clean-up)
- `Finalize_AtomicallyOverwrites_PriorFile_WithAllowOverwrite` (end-to-end: prior contents gone, new SARIF round-trips, no temp residue)

## Notes

- The new `allowOverwrite` parameter has a `[Description]` attribute so it surfaces in `tools/list` discovery and the `EveryParameterOfEveryMcpTool_HasADescription` test continues to pass.
- No existing test triggered the exists-guard (each test creates a fresh scratch directory) -- the change is backward-compatible for all current callers.
- Considered, deferred: `CreateRunTool` writing a zero-result placeholder SARIF at create time so the file is visible as "in progress" between create and finalize. That's a UX-ergonomic feature, not a reliability one; can land separately if/when an external workflow needs it.


michaelcfanning:
Closing as part of a design reset. Reverting the MCP/SDK-fix chain back to PR #2894 (the unshipped AV fix) to establish a clean foundation for a redesigned approach: file-as-state via an append-only JSONL event log + finalize commit, direct CWE descriptor enrichment (no taxonomies mechanism), zero AI flag surface. New design will be planned and implemented after the revert lands. Thanks for the cycles.

PR #2914, opened 2026-05-20
Phase 1: incremental SARIF authoring foundation + emit-init/emit-finalize verbs

michaelcfanning:
## Phase 1: incremental SARIF authoring foundation

This PR delivers the foundation for **incremental SARIF authoring across multiple process invocations**, plus bookend multitool verbs that prove the end-to-end flow.

### The shape

```
multitool emit-init <output>.sarif         # creates <output>.sarif.wip.jsonl + seeds run-header
... external producers append events to the wip event log ...
multitool emit-finalize <output>.sarif     # replays  enriches  atomically writes <output>.sarif
```

The `.wip.jsonl` is the in-flight representation. The `.sarif` is the committed artifact. Two files during authoring; one file as the deliverable.

### What's in this PR

**New `Microsoft.CodeAnalysis.Sarif.Emit` namespace** (`src/Sarif/Emit/`):

| Type | Role |
|---|---|
| `SarifEventLogWriter` / `SarifEventLogReader` | Versioned newline-delimited JSON event log. Kinds in v1: `run-header`, `result`, `invocation`, `notification`. Writer rejects torn trailing lines; reader tolerates CRLF + blank lines + unknown kinds (forward-compatible). |
| `SarifEventReplayer` | Replays an event log into a `SarifLog`. Auto-registers `tool.driver.rules` from first-sighting `Result.RuleId` (mirrors `SarifLogger.LogRule`). Routes notifications to the last invocation, synthesizing one with `executionSuccessful = true` if none was reported. |
| `AtomicSarifWriter` | Writes destination SARIF via same-directory temp file + `File.Move(overwrite: true)`. Readers see prior-or-new contents, never partial. Tolerates consumers that dispose a `StreamWriter` wrapping the supplied stream. |

CWE enrichment in `emit-finalize` uses `Microsoft.CodeAnalysis.Sarif.Taxonomies.CweTaxonomyEnricher` (landed in #2916).

**Two new multitool verbs** (`src/Sarif.Multitool.Library/Emit/`):

- `emit-init <output>.sarif`  state-machine over `(.sarif  .wip.jsonl)  --allow-overwrite`. Seeds run-header with `--tool`, `--tool-version`, `--info-uri`, `--organization`, `--repo`, `--revision`, `--branch`, `--source-root`.
- `emit-finalize <output>.sarif`  replays wip  calls `Sarif.Taxonomies.CweTaxonomyEnricher.Enrich` (unless `--no-cwe-enrichment`)  atomically writes SARIF  deletes wip (unless `--keep-wip`). Supports `--minify`.

**Internal: `Sarif.Multitool/Program.cs` rewrite.** CommandLineParser 2.9.1 caps the generic `ParseArguments<T1..T16>` at 16 verbs. With 16 already registered (DEBUG path) and 2 more to add, the build broke. Replaced with the non-generic `Parser.ParseArguments(args, params Type[])` overload + a `Dispatch(object)` switch  removes the ceiling entirely and gives headroom for future verbs.

### Tests

- `Test.UnitTests.Sarif/Emit/`  event-log, replayer, atomic-writer coverage
  - `SarifEventLogWriterReaderTests`: round-trip, torn-line refusal, CRLF tolerance, blank-line skip, unknown-kind skip, malformed-JSON line-numbered errors, exclusive-lock rejection, wire-shape ordering.
  - `AtomicSarifWriterTests`: create, overwrite, consumer-disposes-stream tolerance, no-turd-on-failure, same-directory staging.
  - `SarifEventReplayerTests`: run-header handling, auto-registration, notification routing, synthetic invocation, header sanitization.
- `Test.UnitTests.Sarif.Multitool.Library/Emit/`  verb coverage
  - `EmitInitCommandTests`: state machine, `BuildRunHeader` shape, empty-flag defaults.
  - `EmitFinalizeCommandTests`: happy path with enrichment, missing wip, `--no-cwe-enrichment`, `--keep-wip`, atomic replace.

Both suites green: **821 / 3 skipped** in `Test.UnitTests.Sarif`, **118 / 1 skipped** in `Test.UnitTests.Sarif.Multitool.Library`.

### Why this shape (vs. a long-lived MCP stdio session)

Earlier in this design line we explored MCP-server-as-authoring-engine. That approach was reverted (#2907) because the intended producer workflow is **incremental authoring across multiple process invocations**, not a long-lived stdio session holding state in-memory. An event-log foundation is the right shape for that  and the same engine can later back an MCP transport too, if that becomes useful.

### Out of scope (future phases)

- Phase 2 verbs: `emit-add-results`, `emit-add-invocation`, `emit-add-notifications` (Phase 1 already covers the end-to-end shape via bookend verbs; producers append events themselves between them).
- AI* validation rules (separate concern; return when AI profile firms up).
- MCP transport (same event-log engine could back it later).
- JSONL chunking for long-running scans.

### A couple of small-but-noteworthy notes

1. **`AtomicSarifWriter` accepts dirty consumers.** If a consumer does `using var sw = new StreamWriter(stream)` (without `leaveOpen: true`), the underlying `FileStream` is disposed before the writer can call `Flush(flushToDisk: true)`. The bytes are already in the OS buffer (the consumer's dispose flushed), so we just catch `ObjectDisposedException` and skip the `fsync` hint. Documented in the XML comment.
2. **Verb naming**: hyphenated single-level (`emit-init`, `emit-finalize`), consistent with existing verbs (`apply-policy`, `absolute-uri`, `result-matching`, `file-work-items`, `rebase-uri`). Nested verbs (`multitool emit init`) would need hierarchical-verb plumbing in CommandLineParser that doesn't carry its weight at this scale.
3. **Rebased on top of #2915 and #2916** (both merged). The earlier draft of this branch carried its own pre-#2916 copy of `CweTaxonomyEnricher` under `Sarif.Emit`  that has been dropped in favor of the public `Microsoft.CodeAnalysis.Sarif.Taxonomies.CweTaxonomyEnricher` shipped by #2916.

### Validation

- Solution builds clean (`Sarif.Sdk.sln`) under `/p:EnforceCodeStyleInBuild=true`.
- No style violations from new code.
- End-to-end smoke validated locally: `emit-init`  external `Sarif.dll`-using appender  `emit-finalize` produces fully-enriched SARIF (CWE-79  MITRE name/desc/helpUri; CUSTOM-1 left bare) + deletes wip. Output round-trips through `SarifLog.Load`.


src/Sarif.Multitool.Library/Emit/EmitConventions.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    /// Conventions shared by the <c>emit-*</c> verbs.
+    /// </summary>
+    internal static class EmitConventions
+    {
+        /// <summary>
+        /// Returns the wip event-log path for a given destination SARIF path
+        /// (<c>foo.sarif</c> → <c>foo.sarif.wip.jsonl</c>).
+        /// </summary>
+        internal static string GetWipPath(string sarifOutputPath) => sarifOutputPath + ".wip.jsonl";
```
michaelcfanning: More descriptive name, I thnk, e.g., `GetWorkInProgressPath`, or `GetSarifJsonlPath`.

src/Sarif.Multitool.Library/Emit/EmitConventions.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    /// <summary>
+    /// Conventions shared by the <c>emit-*</c> verbs.
+    /// </summary>
+    internal static class EmitConventions
```
michaelcfanning: this class looks anemic to me, a new type, file, for a string constant (that should be DRY'ed out anyway and a simple path construction.

src/Sarif.Multitool.Library/Emit/EmitInitOptions.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using CommandLine;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    /// <summary>
+    /// Options for <c>emit-init</c>, which opens an append-only event log
```
michaelcfanning: maybe `emit-init-run`? is this clearer, that we reference the run object?

src/Sarif.Multitool.Library/Emit/EmitInitOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        [Value(
+            0,
+            MetaName = "<outputSarifPath>",
+            HelpText = "Path to the final SARIF file (the event log is staged alongside as '<output>.wip.jsonl').",
+            Required = true)]
+        public string OutputFilePath { get; set; }
+
+        [Option(
+            't',
+            "tool",
```
michaelcfanning: it's more verbose for incredibly consistent if we had full kebab ids that map to the SARIF interior, e.g., `tool-drive-name`, `information-uri` and yes by this convention simply `srcroot`

src/Sarif.Multitool.Library/Emit/EmitInitRunOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        public string Organization { get; set; }
+
+        [Option(
+            "repo",
+            HelpText = "Repository URI (run.versionControlProvenance[0].repositoryUri).")]
+        public string RepositoryUri { get; set; }
+
+        [Option(
+            "revision",
+            HelpText = "Revision identifier, typically a commit SHA (run.versionControlProvenance[0].revisionId).")]
```
michaelcfanning, line 59: `vcp-revisionid`? or `versioncontrolprovenance-revisionid` (getting pretty long with that last name)

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+    /// Implements <c>multitool emit-init</c>: creates an append-only SARIF event log
+    /// (<c>&lt;output&gt;.wip.jsonl</c>) seeded with a <c>run-header</c> event built from the
+    /// supplied tool / repo flags.
+    /// </summary>
+    /// <remarks>
+    /// <para>State table:</para>
+    /// <list type="table">
+    /// <listheader>
+    /// <term>State</term>
+    /// <term>No <c>--allow-overwrite</c></term>
```
michaelcfanning: look at the other commands, i think force overwrite is more typical.

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        public int Run(EmitInitOptions options, IFileSystem fileSystem = null)
+        {
+            fileSystem ??= Sarif.FileSystem.Instance;
+
+            try
+            {
+                if (string.IsNullOrWhiteSpace(options?.OutputFilePath))
+                {
+                    Console.Error.WriteLine("emit-init: output SARIF path is required.");
```
michaelcfanning: Why are we prefixing our messages with the verb name? is there a rationale or an appeal to an existing pattern>?

scripts/Generate-CweTaxonomy.ps1
```diff
@@ -0,0 +1 @@
... of software and hardware weakness types maintained by MITRE Corporation. Published under terms permitting redistribution with attribution.' }    taxa             = $taxa}$run = [ordered]@{    tool = [ordered]@{        driver = [ordered]@{            name = 'Microsoft.CodeAnalysis.Sarif.Taxonomies.CweGenerator'            informationUri = 'https://github.com/microsoft/sarif-sdk'        }    }    taxonomies = @($taxonomy)}$sarifLog = [ordered]@{    '$schema' = 'https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0.json'    version   = '2.1.0'    runs      = @($run)}Write-Host "[4/4] Writing taxonomy to $OutputPath ..."$outputDir = Split-Path $OutputPath -Parentif (-not (Test-Path $outputDir)){    New-Item -ItemType Directory -Path $outputDir -Force | Out-Null}# ConvertTo-Json with adequate depth; the taxa array nests at depth ~5# (sarifLog.runs.taxonomies.taxa.shortDescription.text).$json = $sarifLog | ConvertTo-Json -Depth 10[System.IO.File]::WriteAllText((Resolve-Path -LiteralPath $outputDir).Path + [System.IO.Path]::DirectorySeparatorChar + (Split-Path $OutputPath -Leaf), $json, [System.Text.UTF8Encoding]::new($false))Write-Host "Wrote $($taxa.Count) taxa to $OutputPath"
```
keegan-caruso: please format this file, also a bash script would be appreciated.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Text;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+using Microsoft.CodeAnalysis.Sarif.Emit;
+
+using Newtonsoft.Json;
```
keegan-caruso, line 15: we should be using System.Text.Json, not newtonsoft

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
```
keegan-caruso, line 4: a few things that can clean this up:

- https://csharpier.com/docs/About I like this, install the msbuild version, then never worry about the format, the format is what the tool gives you. There is a minimum config file surface
- Use <ImplicitUsings>enable</ImplicitUsings>, cleans up the using blocks
- add this to an editor config: 

```
# Use file-scoped namespaces
csharp_style_namespace_declarations = file_scoped:warning
```

Feel free to log issue and do as a follow up if you agree with any of it

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    /// <summary>
+    /// Implements <c>multitool emit-finalize</c>: replays <c>&lt;output&gt;.wip.jsonl</c>,
+    /// optionally enriches CWE-as-rule-id descriptors, and atomically writes the destination
+    /// SARIF file.
+    /// </summary>
+    public class EmitFinalizeCommand : CommandBase
+    {
+        public int Run(EmitFinalizeOptions options, IFileSystem fileSystem = null)
```
keegan-caruso, line 26: please add nullable analysis to this project.

Add this to your csproj/props file

```xml
  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>
```

And then annotate nullable types, e.g. IFileSystem?

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+                        string.Format(
+                            CultureInfo.CurrentCulture,
+                            "emit-finalize: event log '{0}' does not exist; run 'emit-init' first.",
+                            wipPath));
+                    return FAILURE;
+                }
+
+                SarifLog log = SarifEventReplayer.Replay(wipPath);
+
+                if (!options.NoCweEnrichment && log?.Runs != null)
```
keegan-caruso, line 53: I'd add this to the editor config:

```
dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
```

At least as suggestion, modern c# prefers `is null` / `is not null`

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    Console.Error.WriteLine(
+                        string.Format(
+                            CultureInfo.CurrentCulture,
+                            "emit-finalize: event log '{0}' does not exist; run 'emit-init' first.",
+                            wipPath));
+                    return FAILURE;
+                }
+
+                SarifLog log = SarifEventReplayer.Replay(wipPath);
```
keegan-caruso, line 51: if we make it invariant that SarifLog .Runs is never null (use an emtpy array for the default case) then we can remove all of these null checks and clean this file up

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
keegan-caruso: I'd suggest using the LoggerMessage Pattern and using ILogger<T> from DI

So this gets a ctor:

```cs
EmitInitCommand (ILogger<EmitInitCommand)
```

then logging is defined like so:

```cs
    [LoggerMessage(
        EventId = 1,
        Level = LogLevel.Error,
        Message = "Output SARIF path is required but was not supplied.")]
    public static partial void FailedToProcessWorkItem(
        ILogger logger);
```

It gives a consistent pattern for logging and moves configuration of logging to well known patterns in DI.

src/Sarif.Multitool.Library/Emit/EmitInitRunCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+            var driver = new ToolComponent
+            {
+                Name = options.ToolName,
+                Version = NullIfEmpty(options.ToolVersion),
+                Organization = NullIfEmpty(options.Organization),
+            };
+
+            if (!string.IsNullOrWhiteSpace(options.InformationUri))
+            {
+                driver.InformationUri = new Uri(options.InformationUri, UriKind.RelativeOrAbsolute);
```
keegan-caruso: if the uri fails to parse, do we want to throw an error in the form of an exception?

src/Sarif/Emit/AtomicSarifWriter.cs
```diff
@@ -0,0 +1,101 @@
...
+            try
+            {
+                if (File.Exists(path))
+                {
+                    File.Delete(path);
+                }
+            }
+            catch
+            {
+                // Best-effort cleanup; intentional swallow.
```
keegan-caruso: log at least maybe?

src/Sarif/Emit/AtomicSarifWriter.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>The staging file is placed in the same directory as the destination so the final
+    /// rename is a within-volume operation, which is atomic on every supported filesystem.</para>
+    /// <para>If the rename fails, the staging file is removed to avoid leaving turds behind.</para>
+    /// <para>The <c>writeContent</c> callback receives the underlying <see cref="FileStream"/>;
+    /// callers MAY dispose any wrapper they construct (e.g., a <see cref="StreamWriter"/>) — the
+    /// final fsync is best-effort and tolerates an already-disposed stream.</para>
+    /// </remarks>
+    public static class AtomicSarifWriter
```
keegan-caruso, line 23: This doesn’t quite provide the advertised atomic replacement guarantee. Deleting the destination before moving the staged file creates a window where readers can observe no file at all, and a crash in that window loses the previous committed SARIF. Could we use File.Replace(stagingPath, destinationPath, null) when the destination exists, and File.Move only for first creation? That would better match the “readers see prior-or-new contents, never partial” contract.

src/Sarif/Emit/CweTaxonomyEnricher.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// already supplied are preserved.</para>
+    /// <para>We are first-order producers of CWE-as-rule-id; the enricher does NOT add a
+    /// <c>descriptor.taxa[]</c> cross-reference. That pattern is for tools that map their own
+    /// rule IDs to CWE.</para>
+    /// </remarks>
+    public static class CweTaxonomyEnricher
+    {
+        private const string EmbeddedTaxonomyResourceName = "Microsoft.CodeAnalysis.Sarif.Taxonomies.CWE.sarif";
+
+        private static readonly Lazy<IReadOnlyDictionary<string, ReportingDescriptor>> s_index =
```
keegan-caruso: avoid interfaces for private /internals, the generated code will be less efficient and the abstraction doesn't bring a benefit. 
```suggestion
        private static readonly Lazy<ReadOnlyDictionary<string, ReportingDescriptor>> s_index =
```

src/Sarif/Emit/CweTaxonomyEnricher.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// <summary>
+        /// Enriches every CWE-shaped descriptor on <paramref name="run"/>.
+        /// </summary>
+        /// <returns>The number of descriptors enriched.</returns>
+        public static int Enrich(Run run)
+        {
+            if (run == null)
+            {
+                throw new ArgumentNullException(nameof(run));
+            }
```
keegan-caruso:
```suggestion
            ArgumentNullException.ThrowIfNull(run);
```

src/Sarif/Emit/CweTaxonomyEnricher.cs
```diff
@@ -0,0 +1,178 @@
...
+                return 0;
+            }
+
+            int enriched = 0;
+            foreach (ReportingDescriptor descriptor in run.Tool.Driver.Rules)
+            {
+                if (TryNormalizeCweId(descriptor.Id, out string cweNumber)
+                    && s_index.Value.TryGetValue(cweNumber, out ReportingDescriptor source))
+                {
+                    if (descriptor.Name == null && source.Name != null)
```
keegan-caruso: these can all be collapsed down to something like

descriptor.Name ??= source.Name;

If Name is null, use the value from source.Name, if source.Name is null, that is okay.

src/Sarif/Emit/CweTaxonomyEnricher.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                    enriched++;
+                }
+            }
+
+            return enriched;
+        }
+
+        /// <summary>
+        /// Recognizes <c>CWE-79</c>, <c>cwe-79</c>, <c>Cwe79</c>, or bare <c>79</c> and returns
```
keegan-caruso:
```suggestion
        /// Recognizes <c>CWE-N</c>, <c>cwe-N</c>, <c>CweN</c>, or bare <c>N</c> and returns
```

src/Sarif/Emit/CweTaxonomyEnricher.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// the canonical numeric form.
+        /// </summary>
+        internal static bool TryNormalizeCweId(string id, out string cweNumber)
```
keegan-caruso:
```suggestion
        internal static bool TryNormalizeCweId(string id, out [NotNullWhen(true)] string cweNumber)
```

This will help after nullables are added

keegan-caruso:
Left a few high level comments on some of the code shape, feel free to do those as follow ups of course.

src/Sarif/Emit/SarifEventKinds.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary>
+        /// A partial <see cref="Run"/> skeleton (everything except <c>results</c>, <c>invocations</c>,
+        /// and <c>notifications</c>, which arrive as separate events). MUST appear at most once per
+        /// event log; SHOULD be the first event in the log.
+        /// </summary>
+        public const string RunHeader = "run-header";
+
+        /// <summary>
+        /// A single self-contained <see cref="Result"/>. Self-contained means the result
+        /// SHALL NOT carry index references (<c>ruleIndex</c>, <c>artifactLocation.index</c>,
```
keegan-caruso, line 26: Can we make this enforceable? is it already enforced?

src/Sarif/Emit/SarifEventLogWriter.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// Append-only writer for a SARIF event log (<c>*.sarif.wip.jsonl</c>).
+    /// </summary>
+    /// <remarks>
+    /// <para>The writer opens the target file with <see cref="FileShare.Read"/> sharing so
+    /// concurrent appends from other processes are rejected with an <see cref="IOException"/>.</para>
+    /// <para>If the file exists and does not end with a newline, the prior writer was interrupted
+    /// mid-line; the writer rejects the file with a <see cref="SarifEventLogException"/> rather
+    /// than risk concatenating bytes to a torn line.</para>
+    /// <para>Every event is serialized to a single UTF-8 line terminated with <c>\n</c>. The line
+    /// is flushed to disk before the writer call returns to minimize the torn-line window on crash.</para>
```
keegan-caruso: This isn't quite true vs the code, would need flushToDisk: true, as well

src/Sarif/Emit/SarifEventLogWriter.cs
```diff
@@ -0,0 +1,143 @@
...
+                token = JToken.FromObject(payload, _serializer);
+            }
+
+            Append(kind, token);
+        }
+
+        public void Dispose()
+        {
+            _stream?.Dispose();
+        }
```
keegan-caruso:
```suggestion
        public void Dispose() => _stream?.Dispose();
```

keegan-caruso:
Any concerns with multiple agent sessions and exclusive locks?

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Definitely will take a look. We have mostly explored `dotnet format` for this kind of operation, there have been frustration with missing rewrite capabilities and keeping everything in sync between cmd-line, ide builds etc.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: We should be. This API didn't exist when we authored this code. If you'd like to author the conversion and allow us to drop our Newtonsoft dependency, I would very much welcome that contribution. Broadly, we have sufficient unit testing dropped in to have a fair degree of confidence in a transition you took this one (without adding significant supplemental tests, I'd guess).

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Another fine suggestion but one that we would drop in as a focused exercise. Please feel free to file an issue, any AI friend reading my feedback.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Yes, indeed, but another issue that extends broadly across the API surface area and good to work on in isolation. We can't do this naively, as empty arrays generally need not to serialize to JSON where they might bloat files unnecessarily or suggest a semantic meaning distinct from 'this array property was never initialized at all'.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: I am open to code style opinions but not eager to start dropping in patterns that aren't consistent with existing code without at least considering tackling this as a style clean-up effort. Dangling another thought to you, I have been eager for a code style/quality hygiene champion for this repo, due to being unable to serve as such myself. Lemme know if you're that here. :)

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
michaelcfanning: I think we should consider this, yes, and enforce broadly. I will have an AI friend file the work.

src/Sarif.Multitool.Library/Emit/EmitInitRunCommand.cs
michaelcfanning: As authored, without TryCreate, this code in fact will throw. I wonder about the looseness here, we can probably enforce a stricter standard, e.g., an absolute URL, no file schema, etc.

src/Sarif.Multitool.Library/Emit/EmitConventions.cs
michaelcfanning: Fixed in 91288cc2 — `EmitConventions` is gone, and the `.wip.jsonl` extension is now inlined at its two callsites (`EmitInitRunCommand` and `EmitFinalizeCommand`). No path-helper to name.

src/Sarif.Multitool.Library/Emit/EmitConventions.cs
michaelcfanning: Fixed in 91288cc2 — `EmitConventions` deleted. `SourceRootBaseId` lives on `EmitInitRunCommand`, the `.wip.jsonl` extension is inlined at its two callsites.

src/Sarif.Multitool.Library/Emit/EmitInitOptions.cs
michaelcfanning: Adopted in 91288cc2. Verb is now `emit-init-run`; types are `EmitInitRunOptions` / `EmitInitRunCommand`. `init` stays as the verb prefix so future siblings (`emit-init-policy`, `emit-init-result-batch`, etc.) land naturally.

src/Sarif.Multitool.Library/Emit/EmitInitOptions.cs
michaelcfanning: Adopted in 91288cc2. CLI flags now mirror SARIF interior paths: `--tool-driver-name` (the 3-segment form because `name` alone is too generic), `--information-uri`, `--tool-version`, `--organization`, and the VCP family below.

src/Sarif.Multitool.Library/Emit/EmitInitRunOptions.cs
michaelcfanning: Adopted `--vcp-revisionid` in 91288cc2 (short prefix + joined leaf — no kebab inside the leaf, since the VCP prefix already disambiguates). `--vcp-repositoryuri` and `--vcp-branch` follow the same shape.

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
michaelcfanning: Fixed in 91288cc2 — flag is now `--force-overwrite`, property is `ForceOverwrite`. Matches the convention everywhere else in the multitool.

src/Sarif.Multitool.Library/Emit/EmitInitCommand.cs
michaelcfanning: Verb-name prefixes dropped in 91288cc2 (5 sites in `EmitInitRunCommand`, 3 in `EmitFinalizeCommand`) — context is already verb-scoped so the prefix was noise. ILogger<T> DI thread filed as #2920 per Michael's commitment above.

scripts/Generate-CweTaxonomy.ps1
michaelcfanning: Filed #2921 — format the `.ps1` and add a bash sibling so non-Windows contributors can refresh the taxonomy without crossing the PowerShell boundary.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Filed #2917 — System.Text.Json migration across the SDK + multitool. Resolving the thread; the migration itself is its own PR.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Filed #2918 — CSharpier + ImplicitUsings + file-scoped namespaces as a bundled style-baseline sweep, including the `.editorconfig` settings called out in the related thread.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Filed #2919 — enable nullable repo-wide as a focused exercise, per Michael's reply.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Folded into the repo-wide style-baseline issue #2918 — the `.editorconfig` settings live alongside the CSharpier / file-scoped namespaces sweep so we don't fragment the cleanup.

src/Sarif.Multitool.Library/Emit/EmitFinalizeCommand.cs
michaelcfanning: Agreed this is a broad API-surface change that needs careful handling of the empty-array semantics Michael called out. Deferring to a focused PR rather than scope-creeping this one. Resolving.

src/Sarif.Multitool.Library/Emit/EmitInitRunCommand.cs
michaelcfanning: Addressed in 91288cc2. `EmitInitRunCommand.Run` now validates `--information-uri`, `--vcp-repositoryuri`, and `--srcroot` up front with `Uri.IsWellFormedUriString`, so a malformed URI produces a targeted error (`Invalid URI for --information-uri: '...'`) before the outer catch can swallow it as a generic 'emit-init failed'. New `Run_FailsOnMalformedInformationUri` test pins the behavior. The stricter standard Michael mused about (absolute-only, no `file:` scheme) is a worthwhile follow-up — happy to add it here or in a separate PR if you want me to lean in.

src/Sarif/Emit/AtomicSarifWriter.cs
michaelcfanning: Done in 31fdff37 — the `TryDelete` swallow now emits a `Trace.WriteLine` with the path + exception details. Symmetric with the `ObjectDisposedException` trace added on the `Flush(flushToDisk:true)` site in 786a69d7. Cleanup stays best-effort (the caller is already in its own catch), but leaked staging files are no longer invisible.

src/Sarif/Emit/AtomicSarifWriter.cs
michaelcfanning: Fixed in 786a69d7. `AtomicSarifWriter` now uses `File.Replace(stagingPath, destinationPath, null)` when the destination exists so the rename is a single filesystem operation — no more delete-then-move window where a reader observes no file at all. `File.Move` is retained only for the first-write case where there is nothing to replace.

src/Sarif/Emit/CweTaxonomyEnricher.cs
michaelcfanning: Moot — this comment was on `src/Sarif/Emit/CweTaxonomyEnricher.cs`, which was deleted in the rebase onto main once PR #2916 landed. The canonical enricher now lives at `src/Sarif/Taxonomies/CweTaxonomyEnricher.cs` and does not use the `Lazy<>`/interface pattern this comment targeted — it builds the taxa lookup per call via a private `BuildTaxaLookup`. Resolving.

src/Sarif/Emit/CweTaxonomyEnricher.cs
michaelcfanning: Moot — this was on the deleted `src/Sarif/Emit/CweTaxonomyEnricher.cs` (canonical version landed via #2916 at `src/Sarif/Taxonomies/CweTaxonomyEnricher.cs`). Separately: `Sarif.csproj` multi-targets `netstandard2.0;net462`, so `ArgumentNullException.ThrowIfNull` (.NET 6+) is not available in the SDK assembly until we drop those targets. Worth tracking but not in scope here. Resolving.

src/Sarif/Emit/CweTaxonomyEnricher.cs
michaelcfanning: Declined for now. The five enrichment branches in the merged `Taxonomies/CweTaxonomyEnricher.TryEnrich` each set `modified = true` to signal that the run was touched, and `??=` cannot carry that side effect. Four of the five branches also use type-specific emptiness checks (`IsEmptyMessage` for `MultiformatMessageString`, `IsEmptyMultiformat` for `Help`, `HelpUri.OriginalString` with a computed fallback URL), not plain `string.IsNullOrEmpty` — so collapsing only the `Name` branch loses the visual uniformity that makes the block readable. Resolving; happy to revisit if there's a shape that keeps both the `modified` signal and the per-type emptiness.

src/Sarif/Emit/CweTaxonomyEnricher.cs
michaelcfanning: Moot — the merged `Taxonomies/CweTaxonomyEnricher.TryNormalizeCweId` is strict: `^\s*[Cc][Ww][Ee]-(\d+)\s*$`. It does not accept `CweN` or bare `N` (per the conversation on #2916 about not inventing fuzzy variants — every accepted form must mirror an external CWE convention). The doc comment for the function reflects that. Resolving.

src/Sarif/Emit/CweTaxonomyEnricher.cs
michaelcfanning: Moot for now — the merged `Taxonomies/CweTaxonomyEnricher.TryNormalizeCweId` is `private static`, so the `[NotNullWhen(true)]` annotation doesn't observably help callers. It does become valuable once `<Nullable>enable</Nullable>` lights up across the SDK; tracked on #2919 so the nullable rollout sweeps over private-static helpers like this one too. Resolving.

src/Sarif/Emit/SarifEventKinds.cs
michaelcfanning: Enforceable as of 786a69d7. Added `SarifEventReplayerTests.Replay_HasSwitchCoverageFor`, a theory that reflects every `public const string` on `SarifEventKinds` and feeds each one through `SarifEventReplayer.Replay`. The replayer's `default:` branch throws `SarifEventLogException("Unexpected event kind '...'")`, so adding a new kind to `SarifEventKinds` without a matching `case` fails the test for that kind. No new API surface, just the test.

src/Sarif/Emit/SarifEventLogWriter.cs
michaelcfanning: Fixed in 786a69d7. The doc now accurately describes the behavior: `Append()` calls `_stream.Flush()` (managed buffer → OS buffer), not `Flush(flushToDisk:true)`; the durable-write contract lives on `AtomicSarifWriter` (which is what produces the final SARIF artifact). The torn-line check is explicitly framed as best-effort (it protects subsequent append sessions, not the in-progress one).

src/Sarif/Emit/SarifEventLogWriter.cs
michaelcfanning: Applied in 786a69d7 — `Dispose` is now `public void Dispose() => _stream?.Dispose();`.

src/Sarif.Multitool.Library/Emit/EmitInitRunCommand.cs
michaelcfanning: Leaning in per Michael's ack  tightened in 8099c43f:

* `--information-uri` and `--vcp-repositoryuri` now require an absolute URI with scheme `https`. http, file, ftp, etc. all rejected up front.
* `--srcroot` requires absolute with scheme in {`https`, `file`}. file:// remains because SARIF base IDs commonly anchor at a local checkout; http is gone.
* Empty/null still accepted (flags remain optional). Implementation pulled into a shared `TryValidateUri` helper so the three sites share their error shape.

Tests: kept the malformed-URI guard and added four positive blocks (relative info URI, http info URI, file: info URI, unsupported-scheme srcroot). Happy-path tests using https + file:// continue to pass unchanged.

cfaucon (approved):

cfaucon (approved):

PR #2924, opened 2026-05-23
Drop SHA-1 from default hashing; add opt-in git-blob-sha-1 (v5.0.0 breaking)

michaelcfanning:
## Summary

Removes SHA-1 from the default `artifact.hashes` emitted by `SarifLogger` and introduces a new `[Flags] HashAlgorithms` enum (`Sha1`, `Sha256`, `GitBlobSha1`, `Default = Sha256`) as the single opt-in mechanism for selecting which algorithms are computed. SHA-1 is now strictly opt-in.

Also adds `HashAlgorithms.GitBlobSha1`: computes `SHA1("blob " + length + "\0" + content)` over the raw on-disk bytes, emitted under the dictionary key `git-blob-sha-1` (lowercase, matching git's canonical form). When the source stream is at position zero, the value matches `git hash-object <file>` exactly. The hash is byte-for-byte sensitive (including line endings) — matching a server-persisted git blob SHA requires the on-disk bytes to match what git stored.

## Coherent v5.0.0 breaking API

Per maintainer decision to take a major-version break in service of a clean API. Every public surface that touched the old SHA-1+SHA-256 implicit default has been updated consistently — `hashAlgorithms` always defaults to `HashAlgorithms.Default` (SHA-256 only) and is always positioned at the end of the parameter list.

| Type / Member | Change |
|---|---|
| `HashAlgorithms` | NEW `[Flags]` enum |
| `HashUtilities.ComputeHashes(string, IFileSystem, HashAlgorithms)` | `algorithms` now optional, defaults to `Default` |
| `HashUtilities.ComputeHashes(Stream, HashAlgorithms)` | same |
| `HashUtilities.ComputeHashesForText(string, HashAlgorithms)` | same |
| `HashUtilities.MultithreadedComputeTargetFileHashes` | new optional `algorithms` parameter |
| `HashUtilities` legacy shims | `ComputeHashes(string, IFileSystem)`, `ComputeHashes(Stream)`, `ComputeHashesForText(string)` removed |
| `HashData` | new `GitBlobSha1` property + 3-arg ctor (2-arg ctor preserved) |
| `HashData.ToDictionary` | emits `git-blob-sha-1` (lowercase); `sha-1`/`sha-256` remain uppercase |
| `Artifact.Create` | new `hashAlgorithms` parameter at end; suppresses empty `"hashes": {}` |
| `Run.GetFileIndex` | new `hashAlgorithms` parameter at end |
| `SarifLogger` (both ctors) | new `hashAlgorithms` parameter at end; a caller-supplied `FileRegionsCache` wins (the cache's `HashAlgorithms` is the source of truth) |
| `FileRegionsCache` | new `hashAlgorithms` ctor parameter + public `HashAlgorithms` property |
| `InsertOptionalDataVisitor.VisitArtifact` | now fetches `HashData` from the cache and threads `FileRegionsCache.HashAlgorithms` to `Artifact.Create` |

### Latent fix bundled with this change

`FileRegionsCache.BuildHashDataForFile` previously hashed a UTF-8 re-encoding of decoded text, which was incorrect for files with BOMs or non-UTF-8 encoding. It now reads raw bytes via `IFileSystem.FileOpenRead`, so all hashes (including the new `git-blob-sha-1`) are computed over the bytes that are actually on disk. The single test fixture in this repo (`InsertOptionalDataVisitor.txt`) has no BOM, so existing SHA-256 snapshot values are unchanged.

## Stream-position semantics for `git-blob-sha-1`

`ComputeHashes(Stream, HashAlgorithms)` hashes from the stream's current position to the end. The git blob header length reflects only the bytes that will actually be hashed (`stream.Length - stream.Position`), so the result matches `git hash-object` exactly when the stream is positioned at zero (e.g., a freshly opened `FileStream`). Callers wishing to compute the git blob SHA-1 of a sub-region of a stream should treat that sub-region as a standalone blob. This contract is pinned down by `ComputeHashes_FromNonZeroStreamPosition_HashesSuffixAsStandaloneBlob`.

## Tests

- Updated 3 `InsertOptionalDataVisitor` expected-output snapshots to drop the `sha-1` entry: `Rel_Hashes.sarif`, `Rel_Hashes+TextFiles.sarif`, `Rel_All.sarif`.
- Updated `Artifact_ComputeHashes` and `SarifLogger_WritesFileData` expectations (now assert no `sha-1` key).
- Added 13 new `HashUtilitiesTests` covering: canonical empty-file and known-content git blob SHA-1 values; lowercase-hex emission; `HashAlgorithms.None` / `Sha256`-only / all-algorithms selection; stream-position restoration; file-vs-stream equivalence; new default-overload semantics (SHA-256 only); non-zero-position semantics; `ToDictionary` null-skip and all-null-empty behavior; `git-blob-sha-1` key emission; legacy 2-arg `HashData` ctor.
- Added `Artifact_Create_HashesNotRequested_DoesNotEmitEmptyHashesObject`.
- Added `SarifLogger_GitBlobSha1Requested_EmitsGitBlobShaInArtifactHashes` (asserts the canonical value `e7b1c4ad7a76616c5313bfa64ebd50fb478aef9d` for `#include "windows.h";`).

## Version

`build.props`: `VersionPrefix` bumped `4.6.4 → 5.0.0`; `PreviousVersionPrefix → 4.6.4`. Full `BRK` + `NEW` entries added to `ReleaseHistory.md` under `UNRELEASED`.

## Verification

- Full solution builds clean (0 warnings, 0 errors).
- **1,589 tests pass** across `Sarif`, `Sarif.Driver`, `Sarif.Converters`, `Sarif.Multitool`, `Sarif.Multitool.Library`, `WorkItems`, `Sarif.WorkItems`, and `FunctionalTests.Sarif` (5 pre-existing skips, 0 failures).

## Deferred items

- Newline normalization / server-stability of `git-blob-sha-1` (e.g., emitting both a raw-bytes form and a normalized form to match what the server stored regardless of working-tree config). Discussed and intentionally deferred — for now, callers wanting server-matching values must ensure on-disk bytes match what git stored.
- Whether to evolve `HashData` to a dictionary-backed shape. Kept additive (named properties) for now to minimize churn.

🤖 Generated with [GitHub Copilot CLI](https://github.com/github/gh-copilot)


src/Sarif/Core/Artifact.cs
```diff
@@ -70,16 +71,20 @@ public static Artifact Create(
...
                 {
-                    HashData hashes = hashData ?? HashUtilities.ComputeHashes(filePath);
+                    HashData hashes = hashData
+                        ?? HashUtilities.ComputeHashes(filePath, fileSystem, hashAlgorithms);
 
                     // The hash utilities will return null data in some test contexts.
                     if (hashes != null)
                     {
-                        artifact.Hashes = new Dictionary<string, string>
+                        IDictionary<string, string> hashDictionary = hashes.ToDictionary();
```
michaelcfanning: what's all this work for? couldn't we simply assign the result of NullIfEmpty(hashes) to artifact.Hashes?

src/Sarif/Core/Artifact.cs
```diff
@@ -112,5 +117,13 @@ public override string ToString()
             return this.Location?.ToString() ?? base.ToString();
         }
 #endif
+
+        // Avoid serializing an empty "hashes": {} object when no algorithm produced a value
```
michaelcfanning: so much code commentary when `NullIfEmpty` tells the whole story. please clean up other verbose comments that basically reframe obvious logic in natural language.

src/Sarif/FileRegionsCache.cs
```diff
@@ -364,14 +376,19 @@ private static void PopulateCharLength(NewLineIndex lineIndex, Region region)
         public HashData GetHashData(Uri uri, string fileText = null)
         {
             string path = uri.GetFilePath();
+
+            // Legacy / test path: caller supplied the text directly. We hash the
```
michaelcfanning: huh? why are we enforcing this limit? let's not do that

src/Sarif/FileRegionsCache.cs
```diff
@@ -421,8 +438,22 @@ private string RetrieveTextForFile(string path)
 
         private HashData BuildHashDataForFile(string path)
         {
+            // Read raw bytes from disk so that git-blob-sha-1 (if requested) matches
```
michaelcfanning: non useful code comment

src/Sarif/HashAlgorithms.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// dictionary key <c>git-blob-sha-1</c>. The value matches what
+        /// <c>git hash-object &lt;file&gt;</c> would produce for the same bytes.
+        /// </summary>
+        GitBlobSha1 = 0x4,
+
+        /// <summary>
+        /// The default set of algorithms computed by <see cref="SarifLogger"/> and related
+        /// infrastructure when only <see cref="OptionallyEmittedData.Hashes"/> is requested.
+        /// </summary>
+        Default = Sha256,
```
michaelcfanning, line 64: should we go ahead and add SHA512?

src/Sarif/HashData.cs
```diff
@@ -8,27 +8,45 @@ namespace Microsoft.CodeAnalysis.Sarif
     public class HashData
```
michaelcfanning, line 14: these parameterized constructors aren't super helpful, maybe consider a more modern init strategy?

michaelcfanning:
Pushed `3a3a3298` addressing all six inline comments + the explicit-null flag.

**Mapping to each review comment:**

| # | File:Line | Change |
|---|---|---|
| 1 | `Artifact.cs:80` | Collapsed if-block to `artifact.Hashes = NullIfEmpty(hashes?.ToDictionary());` |
| 2 | `Artifact.cs:121` | Dropped the multi-line comment on `NullIfEmpty`. Also condensed the verbose XML remarks on `ComputeHashes(Stream, HashAlgorithms)`. |
| 3 | `FileRegionsCache.cs:380` | Removed the `~HashAlgorithms.GitBlobSha1` mask in both `GetHashData(text)` and the text-fallback in `BuildHashDataForFile`. GitBlobSha1 is now computed over the UTF-8 bytes of supplied text if requested (won't match a server-stored blob SHA, but it's deterministic — caller's responsibility). |
| 4 | `FileRegionsCache.cs:441` | Removed the verbose "Read raw bytes from disk..." comment. |
| 5 | `HashAlgorithms.cs:59` | Added `Sha512 = 0x8`. Plumbed through `HashUtilities.ComputeHashes(Stream, ...)` as a new `IncrementalHash` participant in the single-pass loop. New canonical-value test (`HashUtilitiesTests.ComputeHashes_KnownContent_Sha512_MatchesCanonical`), plus updated None/Sha256-only/All-algorithms tests to cover Sha512. |
| 6 | `HashData.cs:8` | Modernized: positional ctors removed in favor of object-initializer syntax — `new HashData { Sha256 = "...", GitBlobSha1 = "..." }`. Properties are now settable (matching the autogenerated SARIF object idiom in the codebase). Added `Sha512` property. All call sites + tests updated. |
| — | `HashUtilities.cs:53` (your explicit-null flag) | Added `IFileSystem fileSystem = null` parameter to `MultithreadedComputeTargetFileHashes` (between `suppressConsoleOutput` and `algorithms`); threaded through to `ComputeHashes` instead of the hardcoded null. |

**Test results:** 1,590 passed / 5 skipped (pre-existing) / 0 failed across the solution. Clean build, 0 warnings.

`ReleaseHistory.md` UNRELEASED entry updated to call out the SHA-512 NEW item and the HashData modernization inside the existing BRK bullet.

One scope-check question: the `IFileSystem` threading audit shows other call sites of `Artifact.Create` (in `SarifLogger`, `InsertOptionalDataVisitor`, `Run.GetFileIndex`) that don't currently pass through an `IFileSystem`. The visitor has its own `_fileSystem` field, and `SarifLogger` relies on `FileRegionsCache._fileSystem` (which is currently private). Do you want me to extend the threading in a follow-up commit — exposing `FileRegionsCache.FileSystem` as a public property mirror of `HashAlgorithms`, and threading it through all internal `Artifact.Create` / `Run.GetFileIndex` call sites? I left that out to keep this commit focused on your explicit flag.

michaelcfanning:
Followed up with `7427668d` — threaded `IFileSystem` through the remaining `Artifact.Create` / `Run.GetFileIndex` call sites.

**Approach:** Exposed `FileRegionsCache.FileSystem` as **internal** (mirror of the existing public `HashAlgorithms` property). `Sarif.dll` exposes internals to `Sarif.Driver` and the test assemblies via `InternalsVisibleTo`, so the property reaches every caller that legitimately needs it without expanding the public surface area. External consumers don't see it — the cache continues to manage `IFileSystem` lifetime as plumbing.

**Updated call sites:**
- `SarifLogger.EnhanceRun` (3 sub-calls: 1× `Artifact.Create`, 1× `_run.GetFileIndex`) — passes `FileRegionsCache.FileSystem`.
- `SarifLogger.LogArtifact` (1× `_run.GetFileIndex`) — passes `FileRegionsCache?.FileSystem`.
- `SarifLogger.AnalyzingTarget` (1× `_run.GetFileIndex`) — passes `FileRegionsCache?.FileSystem`.
- `InsertOptionalDataVisitor.VisitArtifact` — now passes its own `_fileSystem` to `Artifact.Create`.
- `Run.GetFileIndex` — gains `IFileSystem fileSystem = null` parameter (positionally before `hashAlgorithms`); threads to its internal `Artifact.Create`. Named-arg callers (i.e., all internal callers) unaffected.

Tests: 1,590 / 5 / 0 still green.

`ReleaseHistory.md` BRK bullet updated to call out the `Run.GetFileIndex` parameter addition.

src/Sarif/Core/Artifact.cs
michaelcfanning: Done in `3a3a3298` — collapsed to `artifact.Hashes = NullIfEmpty(hashes?.ToDictionary());`.

src/Sarif/Core/Artifact.cs
michaelcfanning: Done in `3a3a3298` — dropped the multi-line comment on `NullIfEmpty`, and also condensed the verbose XML remarks on `HashUtilities.ComputeHashes(Stream, HashAlgorithms)` for the same reason.

src/Sarif/FileRegionsCache.cs
michaelcfanning: Done in `3a3a3298` — removed the `~HashAlgorithms.GitBlobSha1` mask in both `GetHashData(text)` and the text-fallback path of `BuildHashDataForFile`. Callers requesting `GitBlobSha1` with text-only input now get it computed over the UTF-8 bytes of the supplied text (won't match a server-stored git blob SHA, but it's deterministic — caller's responsibility).

src/Sarif/FileRegionsCache.cs
michaelcfanning: Done in `3a3a3298` — removed.

src/Sarif/HashAlgorithms.cs
michaelcfanning: Done in `3a3a3298` — added `Sha512 = 0x8`, plumbed through `HashUtilities.ComputeHashes` as a new `IncrementalHash` in the single-pass loop, emitted under the `sha-512` dictionary key, plus a canonical-value test (`HashUtilitiesTests.ComputeHashes_KnownContent_Sha512_MatchesCanonical`) and updated assertions across the None / Sha256-only / All-algorithms tests.

src/Sarif/HashData.cs
michaelcfanning: Done in `3a3a3298` — replaced positional constructors with object-initializer syntax (`new HashData { Sha256 = "...", GitBlobSha1 = "..." }`); properties are now settable, matching the autogenerated SARIF object idiom in this codebase. All call sites and tests updated.

cfaucon (approved):

PR #2925, opened 2026-05-24
CI: modernize dotnet format (drop deprecated dotnet-format 5.x)

michaelcfanning:
## What

Replaces the legacy standalone `dotnet-format` 5.1.250801 in `.github/workflows/validate.yml` with the built-in SDK `dotnet format` (added in .NET 6, the path going forward).

## Why

The two formatters disagree on some C# 12+ constructs. That divergence caused [PR #2923](https://github.com/microsoft/sarif-sdk/pull/2923) to hit a CI failure that didn't reproduce locally: contributors using the modern SDK formatter (which is what `dotnet format` invokes by default) produce output that the CI legacy tool rejects. Modernizing CI to the same formatter contributors actually have on disk closes that gap.

## What's in the PR

* `.github/workflows/validate.yml`  `check-format` job uses `dotnet format whitespace src\Sarif.Sdk.sln --verify-no-changes --exclude src/Sarif/Autogenerated/` instead of the legacy tool. `actions/checkout` bumped v2  v4. Explicit `setup-dotnet@v4 / 8.0.x` so the formatter version is pinned alongside the rest of the build.
* One-time normalization of 8 files where the new formatter wanted whitespace fixes the legacy tool let through. (The legacy tool was producing both false positives _and_ false negatives  these are the negatives.)

## Test plan

* CI green on this PR is the definition of correctness for this change.
* I also ran `dotnet format whitespace src\Sarif.Sdk.sln --verify-no-changes --exclude src/Sarif/Autogenerated/` locally; exit 0.

## Follow-ups

* If we want stricter style/analyzers enforcement in CI, `dotnet format style` and `dotnet format analyzers` are separate sibling commands. Scoped out of this PR.


michaelcfanning:
Superseded by #2927, which now subsumes the CI modernization (validate.yml change) along with the broader dotnet format pass and editorconfig hardening.

PR #2926, opened 2026-05-24
AI-origin-aware suppression for style-class SARIF validation rules

michaelcfanning:
## What

Adds SarifValidationSkimmerBase.IsAIOriginRun(Run) and wires it into four
style-class SARIF validation rules so AI-emitted SARIF fixtures can ship
**0 errors / 0 warnings / 0 notes** under
`validate --rule-kind Sarif;AI --level Error;Warning;Note`.

Affected rules (Note-level style guidance that doesn't apply to AI content):

| Rule | Why suppressed for AI content |
|---|---|
| `SARIF2002` ProvideMessageArguments | AI emits rendered `message.text` per-result, not via `message.id`+`arguments`. |
| `SARIF2009` ConsiderConventionalIdentifierValues | The `NOVEL-` prefix IS the documented AI ruleId escape hatch. |
| `SARIF2014` ProvideDynamicMessageContent | AI rule descriptors carry direct messages, not `{N}` placeholders. |
| `SARIF2015` EnquoteDynamicMessageContent | Companion to 2014. |

**Trigger:** any non-empty `run.properties["ai/origin"]` value. The marker is
intrinsic to the data, so downstream viewers that run validation by default get
correct behavior without needing a CLI flag.

**Orthogonal axes:**
- `--rule-kind` controls *which rules run*.
- `ai/origin` controls *how the rules interpret the data*.

## Not in scope (correctness, not style)

These continue to apply uniformly to AI-emitted SARIF and are NOT suppressed:
`SARIF2003`, `SARIF2006`, `SARIF2010`, `SARIF2011`, `SARIF2012`,
`SARIF2013`, `SARIF2016`, and the ADO ingestion rules (`ADO101x`).
ADO rules enforce the ingestion contract, not style.

## Determinism fix bundled

`CweGenerateSample.ps1` had a latent non-determinism bug: bare `@{...}`
hashtables fed to `ConvertTo-Json` have unspecified key enumeration order
across fresh PowerShell processes (stable within one process; varies between
subprocess invocations). All JSON-bound hashtables now use `[ordered]@{...}`
so the fixture is byte-identical across the eight-run determinism gate
(`CweGeneratedSampleTests`).

## Tests

- **11 new unit cases** on `SarifValidationSkimmerBaseTests` cover the helper:
  null run, no-properties, unrelated property, four non-empty `ai/origin`
  values via `[Theory]`, two empty/whitespace values via `[Theory]`.
- `CweSample.sarif` validates clean: **0/0/0**.
- Full sweep:
  - `Test.UnitTests.Sarif.Multitool.Library`: 157 passed / 1 skipped.
  - `Test.UnitTests.Sarif`: 887 passed / 3 skipped.

## Release notes

Added an `UNRELEASED` bullet at the top of `ReleaseHistory.md` summarizing
the new behavior and the helper.


src/Sarif.Multitool.Library/Rules/SARIF2002.ProvideMessageArguments.cs
```diff
@@ -37,6 +37,13 @@ public ProvideMessageArguments()
 
         protected override void Analyze(Result result, string resultPointer)
         {
+            if (IsAIOriginRun())
+            {
+                // AI emitters render message.text per-result; the id+arguments pattern is
```
michaelcfanning: A better explanation is that AI results are typically stochastic, not driven by tables for messages. the message string analysis is recommended for fixed format strings that help enable localization scenarios. I presume that AI will simply translate content on the fly as necessary, dunno, to handle loc scenarios in the future.

src/Sarif.Multitool.Library/Rules/SARIF2009.ConsiderConventionalIdentifierValues.cs
```diff
@@ -39,6 +39,14 @@ public ConsiderConventionalIdentifierValues()
 
         protected override void Analyze(Run run, string runPointer)
         {
+            if (IsAIOriginRun(run))
```
michaelcfanning: do you need this? do we really want to suspend all tool analysis at the run level? better to have processing within the tool analyze method (as you already do?)

src/Sarif.Multitool.Library/Rules/SARIF2014.ProvideDynamicMessageContent.cs
```diff
@@ -39,6 +39,14 @@ public ProvideDynamicMessageContent()
...
         {
+            if (IsAIOriginRun())
+            {
+                // AI rule descriptors carry direct, per-result message.text rather than
+                // {N}-templated messageStrings; the dynamic-content check is human-author
+                // guidance that does not apply.
+                return;
+            }
+
             if (tool.Driver != null)
```
michaelcfanning, line 42: same note, don't suspend processing at a parent level, inject the meaningful conditional execution as closely co-located with the rules at hand as possible.

src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -297,6 +297,40 @@ internal static bool IsWellFormedUriString(string uriString, UriKind uriKind)
...
+        /// that the containing run was produced by an AI-emitting tool. Style-class
+        /// validation rules (e.g. SARIF2002, SARIF2009, SARIF2014, SARIF2015) suppress
+        /// themselves when this marker is set: AI emitters carry per-result rendered
+        /// message text, use the NOVEL- ruleId escape hatch, and otherwise diverge from
+        /// the human-authoring guidance those rules encode.
+        ///
+        /// Correctness-class rules (snippets, hashes, provenance, relative URIs, etc.)
+        /// must NOT consult this marker — those checks apply uniformly to AI content.
+        /// </summary>
+        protected const string AIOriginPropertyName = "ai/origin";
```
michaelcfanning, line 314: does this helper more logically belong in the AI rule set, DRY'ed out possibly to support the actual AI origin check?

src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
```diff
@@ -297,6 +297,40 @@ internal static bool IsWellFormedUriString(string uriString, UriKind uriKind)
...
+
+        /// <summary>
+        /// Returns true when <paramref name="run"/> declares AI provenance via the
+        /// <c>ai/origin</c> run property. Any non-null/non-empty value counts; the
+        /// vocabulary (<c>generated</c>, <c>assisted</c>, <c>reviewed</c>, …) is open
+        /// by design so AI tooling can self-describe at any granularity.
+        /// </summary>
+        internal static bool IsAIOriginRun(Run run)
+        {
+            if (run == null) { return false; }
```
michaelcfanning: why is silent failure for a null run ok here?

src/Sarif.Multitool.Library/Rules/SARIF2002.ProvideMessageArguments.cs
michaelcfanning: Adopted your framing  the rationale now leads with stochastic per-result rendering vs fixed-format strings for localization (which AI handles out-of-band). Comment updated in f68c5b214b73cbc1cd75e6ae26228953aa10e523.

src/Sarif.Multitool.Library/Rules/SARIF2009.ConsiderConventionalIdentifierValues.cs
michaelcfanning: Good call  moved the suppression out of `Analyze(Run)` and down into `AnalyzeReportingDescriptor`, immediately adjacent to the slash/regex checks. The `AnalyzeTool`/`AnalyzeToolDriver` walk still happens; only the per-descriptor branches that actually mis-fire on `BASE/sub-id` and `NOVEL-<sub-id>` are gated. f68c5b214b73cbc1cd75e6ae26228953aa10e523.

src/Sarif.Multitool.Library/Rules/SARIF2014.ProvideDynamicMessageContent.cs
michaelcfanning: Same move applied  suppression now lives in `AnalyzeMessageString`, right next to the no-{N}-content check it gates. Also applied the same change to SARIF2015 (companion rule) for consistency, since the same pattern is there. f68c5b214b73cbc1cd75e6ae26228953aa10e523.

src/Sarif.Multitool.Library/Rules/SarifValidationSkimmerBase.cs
michaelcfanning: Fair  silent false was wrong. The static `IsAIOriginRun(Run)` now throws `ArgumentNullException` on null; that contract makes upstream lifecycle bugs loud rather than masking them as "not AI". The instance shim still explicitly handles the "no current run scope" case (e.g. an `Analyze(SarifLog)` dispatch) with a justified false. Test flipped from `NullRun_ReturnsFalse` to `NullRun_Throws`. f68c5b214b73cbc1cd75e6ae26228953aa10e523.

cfaucon (approved):

PR #2946, opened 2026-05-27
Promote dev to main: v5.0.1 release

michaelcfanning:
Promote `dev` to `main` to ship v5.0.1.

v5.0.1 was tagged but never published to NuGet. This PR brings every change accumulated on `dev` since the v5.0.0 ship into `main` so the v5.0.1 NuGet packages can be cut.

**Headlines (see `ReleaseHistory.md` for the full bullets):**

- **BRK:** `multitool emit-init-run` drops the twelve typed content flags it shipped with in v5.0.0 and consumes a SARIF `Run` JSON document via `--input` / stdin instead — same contract as `add-result` / `add-notification` / `add-reporting-descriptor`.
- **BRK:** `AdoPipelineContext.ApplyTo(Run)` → `bool TryApplyTo(Run, out string error)`; stamps `automationDetails.id` and the four `azuredevops/pipeline/build/*` properties only when absent and returns false on per-field conflict.
- **NEW:** `multitool add-invocation` verb (only path producers have to populate `run.invocations[]` since the replayer strips header-carried invocations).
- **NEW:** `multitool emit-init-run` accepts the full `Run` partial-document shape the replayer already supports — multiple `versionControlProvenance` entries and arbitrary `properties` bags now flow through.
- **NEW:** Reintroduce `RuleKind.Ado` as an `[Obsolete]` alias for `RuleKind.GHAzDO` (source-compat with v5.0.0 consumers — the v5.0.0 rename removed the value outright).
- **Infra:** `Validate` CI workflow now triggers on `dev`-targeted pushes / PRs in addition to `main`.


michaelcfanning:
Reopening with a single squashed release commit so the PR view shows one ship-shaped change instead of 23 intermediate commits.

PR #2955, opened 2026-05-27
Narrow SARIF1001 case-fold relaxation to AI-origin notification descriptor (#2951)

### LINKED ISSUE #2951: SARIF1001: case-insensitive 'name' vs 'id' comparison overrides spec text
## Summary

SARIF1001 currently performs a **case-insensitive** comparison when checking whether `reportingDescriptor.name` is identical to `reportingDescriptor.id`. A literal reading of the spec text suggests the comparison should be **case-sensitive**.

## Spec text

[SARIF v2.1.0 § 3.49.7](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317843):

> If `name` is provided, it **SHALL NOT be identical to** `id`. […] If `name` is provided and `id` is also provided, they SHALL NOT be the same string.

The word "identical" / "the same string" is naturally case-sensitive in software contexts. The current case-fold comparison expands the spec's constraint beyond its text.

## Why this matters

Producers using `SCREAMING-CAPS` ids (a natural convention for opaque-stable concept-token ids) are forced to invent two-word names when the natural `PascalCase` form would suffice:

```jsonc
// Today: flagged by SARIF1001 (case-folded comparison treats these as identical)
{ "id": "DECISION",    "name": "Decision" }
{ "id": "ESCALATION",  "name": "Escalation" }
{ "id": "ERROR",       "name": "Error" }

// Workaround producers adopt to pass the validator
{ "id": "DECISION",    "name": "DecisionEvent" }
{ "id": "ESCALATION",  "name": "SkillEscalation" }
{ "id": "ERROR",       "name": "UnhandledError" }
```

The workaround names carry no additional semantic information — they exist solely to dodge the case-fold. Meanwhile, hyphen-separated ids (which produce camel-case names that already differ from the id by hyphen presence) pass without issue:

```jsonc
// Already passes today (id contains hyphens; name does not)
{ "id": "RULED-OUT",         "name": "RuledOut" }
{ "id": "RULE-COVERAGE-GAP", "name": "RuleCoverageGap" }
```

The current behavior is **inconsistent**: producers with hyphen-separated ids get a clean rule; producers with single-word ids get pushed into ceremony.

## Proposed fix

Change SARIF1001 to use a **case-sensitive** comparison (the literal spec reading).

## Backward compatibility

**Monotonically relaxed.** No producer that passes today starts failing.

| Producer pattern | Today's behavior | After fix |
|---|---|---|
| `id`, no `name` | passes | passes |
| `id` and `name`, strictly identical strings | fails (correct) | fails (correct) |
| `id` and `name`, case-fold-equal but not identical (e.g. `DECISION` / `Decision`) | **fails** | **passes** |
| `id` and `name`, neither identical nor case-fold-equal | passes | passes |

Only the third row changes, and only in the direction of accepting more producer input.

## Out of scope for this issue

A small follow-on docs change to `generating-sarif.md` could note that the natural convention for AI / concept-token ids is `id = SCREAMING-CAPS-WITH-HYPHENS`, `name = PascalCase(id, hyphens stripped)`. That's covered in a separate proposal issue on AI notification taxonomy.

michaelcfanning:
Fixes #2951.

## Spec text

SARIF v2.1.0 [§ 3.49.7 (`reportingDescriptor.name`)](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317843):

> If `name` is provided, it **SHALL NOT** be identical to `id`. [...] If `name` is provided and `id` is also provided, they **SHALL NOT** be the same string.

"Identical" and "the same string" both read case-sensitively in the surrounding spec.

## Today

`SARIF1001.RuleIdentifiersMustBeValid` compares with `StringComparison.OrdinalIgnoreCase`, expanding the rule beyond what § 3.49.7 actually constrains. Producers that pair a `SCREAMING-CAPS` opaque id with the corresponding `PascalCase` end-user name — a natural shape for AI / concept-token reporting descriptors — get flagged today even though the two strings are literally different.

```csharp
// SARIF1001.RuleIdentifiersMustBeValid.cs, line 38 (before)
reportingDescriptor.Id.Equals(reportingDescriptor.Name, StringComparison.OrdinalIgnoreCase)
```

## Change

One token: `OrdinalIgnoreCase` → `Ordinal`.

The change is **monotonically relaxed**. No producer that passes today starts failing; producers that previously had to invent multi-word names solely to dodge the case-fold comparison can now use the natural form.

| Before fix | After fix |
|---|---|
| `id="DECISION", name="Decision"` → flagged | passes |
| `id="RULE-COVERAGE-GAP", name="RuleCoverageGap"` → flagged | passes |
| `id="RULE0001", name="RULE0001"` → flagged | flagged (unchanged) |
| `id="RULE0001", name="Rule one"` → passes | passes (unchanged) |

## Tests

The Valid functional-test input fixture adds the two case-fold-equal-but-not-identical pairs above. The expected-output baseline is unchanged (still `results: []`). Both `SARIF1001_RuleIdentifiersMustBeValid_Valid` and `..._Invalid` pass.

## Peer-rule audit (and a flagged follow-up)

While here, I grepped `StringComparison.OrdinalIgnoreCase` across `src/Sarif.Multitool.Library/Rules/SARIF*.cs`. One other hit is worth surfacing as a **follow-up, not a blocker**:

> **`SARIF1010.RuleIdMustBeConsistent.cs` line 53** has the inverse drift:
> `!result.RuleId.Equals(result.Rule?.Id, StringComparison.OrdinalIgnoreCase)`
>
> The SARIF spec says `result.ruleId` MUST equal `result.rule.id` when both are present. The current code silently tolerates `ruleId="cwe-79"` / `rule.id="CWE-79"`. Tightening it to `Ordinal` would match the spec but is a producer-breaking direction (unlike this PR, which is monotonically relaxed). Worth a separate issue with deprecation messaging.

I will file that as its own issue rather than fold it into this PR.

## Release notes

A `BUG:` bullet under `## **v5.0.2** UNRELEASED` in `ReleaseHistory.md`.


michaelcfanning:
Filed the SARIF1010 inverse-drift follow-up as #2956 with the two rollout options (clean BRK vs. soft-launch with a transition diagnostic).

PR #2964, opened 2026-06-01
Reconcile authored regions via OverwriteExistingData (#1784)

### LINKED ISSUE #1784: Define invalid log behavior found during Multitool rewrite mode
During Multitool rewrite, if Snippet or Region information is added, FileRegionsCache had Debug.Asserts to validate that Regions appear valid and included properties are consistent with one another.

This is undesirable. We don't want the Multitool to blow up and fail to complete if the SARIF a user is trying to transform is invalid in some way. We may want to emit detected problems to the console, in the output log as ToolNotifications, or elsewhere.

Users likely want validate to catch Region issues, among other things, but will want rewrite mode to leave invalid Results alone and fix others as requested.

FileRegionsCache's Debug.Asserts have been changed to redirect to an empty method as a stub to contain this logic, once fully defined.

michaelcfanning:
## What

Make `OptionallyEmittedData.OverwriteExistingData` govern reconciliation of authored region coordinates against the values computed from the source text.

`FileRegionsCache` previously discarded any divergence between an authored region coordinate and the computed value (a no-op `Assert`), silently shipping regions that point at the wrong span. This is a latent correctness hole: a producer that mis-authors `startColumn`/`charOffset`/etc. ships a SARIF result whose region highlights the wrong text.

## Behavior

- **`OverwriteExistingData` UNSET (default — `validate me`):** a divergent authored coordinate (`startLine`/`startColumn`/`endLine`/`endColumn`/`charOffset`/`charLength`), or a character span past end-of-file, throws `ArgumentException` (paramName `inputRegion`).
- **`OverwriteExistingData` SET (`trust me / recompute`):** the divergent authored coordinate is overwritten with the computed value. This also makes the flag honest — previously it only governed snippets/hashes/contents and never touched region coordinates.
- **Absent (`0`) coordinates** are still computed and filled exactly as before, so callers that omit coordinates are unaffected.

## Surface

- `FileRegionsCache.PopulateTextRegionProperties` gains an `overwriteExistingData` parameter (default `false`).
- `InsertOptionalDataVisitor` threads the already-computed `OverwriteExistingData` flag into it.
- `emit-finalize` does **not** set `OverwriteExistingData`, so it validates authored regions for free; a producer that mis-authors a region now fails loudly at finalize time.

## Tests

Rewrote the `FileRegionsCache` coverage around the overwrite axis: divergent coordinate throws when not overwriting; divergent coordinate is recomputed when overwriting; exact coordinates survive both modes. Full `FileRegionsCache` + `InsertOptionalDataVisitor` (38 pass / 1 skip) and `Sarif.Multitool.Library` (263 pass / 1 skip) suites green.

Closes #1784


cfaucon (approved):

michaelcfanning:
Superseded by #2966, which combines this region-reconciliation work with the AI fingerprints rule split and release-note guards into a single PR targeting `dev`. The region commit (`36c7bb48`) is carried verbatim there.

PR #2968, opened 2026-06-02
Add AI emit-profile input schemas for the invocation and reporting-descriptor verbs

michaelcfanning:
## What

SDK-owned, publishable Draft 2020-12 JSON **input schemas** for the remaining AI emit verbs, each pinned to its canonical C# receipt contract by an xUnit drift fact. Follows #2967 (`ai-result` + `ai-run-header`).

| Schema | Verb / path | Pins |
| --- | --- | --- |
| `ai-invocation.schema.json` | `add-invocation` | overlay on the canonical SARIF invocation (`allOf` `$ref`) requiring `executionSuccessful` (bool) + `commandLine` (non-whitespace) + `workingDirectory` (artifactLocation w/ non-whitespace `uri`) |
| `ai-reporting-descriptor.schema.json` | `add-reporting-descriptor` | object + non-empty string `id` |
| `ai-novel-rule-descriptor.schema.json` | `add-reporting-descriptor --rules` | the base (by `$ref`) + `id` starts with `NOVEL-` |

## `add-notification` is removed; notifications travel inline on the invocation

This PR also **removes the `add-notification` verb** (and the `ExecutionNotification`/`ConfigurationNotification` event kinds). SARIF has no run-level notifications array — a notification lives on the `invocation` that owns it (`toolExecutionNotifications[]` / `toolConfigurationNotifications[]`). A standalone, streamed `add-notification` could not be routed to one of several **parallel** invocations without a correlation key, so the design collapses to a single atomic carrier: the producer holds per-process state and emits **one fully-formed `invocation`** (notifications inline) when that process finishes. This drops the replayer's notification-synthesis path and gives one schema describing the complete enriched object.

## `add-invocation` is an overlay, not a thin schema

`add-invocation` is the sole path to populate `run.invocations[]`, so its schema `$ref`s the canonical SARIF 2.1.0 invocation by `allOf` (every base field keeps its type) and tightens it for the AI profile:

- **Required:** `executionSuccessful`, `commandLine` (non-whitespace), `workingDirectory` (artifactLocation with a non-whitespace `uri`), and a non-whitespace `timeUtc` on every inline notification. The AI producer knows all of these; `commandLine`/`workingDirectory` are the natural identity + path anchor of a launched process, so the profile requires them even though SARIF §3.20 makes them optional. A notification's `timeUtc` is **producer-owned** — generated in-flight while the producer assembles the invocation — so the verb never invents it; it is required by the schema overlay and by the receipt gate.
- **Not required:** `endTimeUtc` — the verb **auto-stamps** this at emit time from wall-clock `now` when the producer leaves it unset (time of receipt ≈ process end), preserving any producer-supplied value, so replay stays deterministic.
- `exitCode` / `startTimeUtc` etc. remain permitted-but-optional via the base `$ref`.

The verb's receipt gate (`AddInvocationCommand.TryValidateInvocationReceipt`) mirrors the three requireds; richer structural validation is deferred to `emit-finalize --validate`.

## The descriptor is one object with two storage targets

`add-reporting-descriptor` accepts **one** reportingDescriptor shape; the `--rules` flag only chooses where the verb files it — `run.tool.driver.rules[]` with `--rules`, `run.tool.driver.notifications[]` by default. So rather than two parallel schemas, this models it as a **base + minimal overlay**:

- `ai-reporting-descriptor.schema.json` is the general contract (object + non-empty string `id`).
- `ai-novel-rule-descriptor.schema.json` `$ref`s the base via `allOf` and adds the single tightening the `--rules` path enforces — `id` must start with `NOVEL-`. It reads as "a novel rule descriptor **is** a reporting descriptor whose id starts with `NOVEL-`", keeping one source of truth for the descriptor shape.

## Faithful-to-source nuances

- **`minLength: 1`, not `pattern: "\\S"`** on descriptor `id` — the verb gates `id` on `string.IsNullOrEmpty` (not `IsNullOrWhiteSpace`), so a whitespace-only `id` is **accepted**.
- **`^NOVEL-` prefix only** — `--rules` gates on `AIRuleIdConvention.IsNovel` (`StartsWith("NOVEL-")`), not the full novel grammar. So `NOVEL-foo/bar` and bare `NOVEL-` are accepted by this receipt overlay even though the result-side `ruleId` grammar in `ai-result.schema.json` rejects them. The drift facts assert exactly this separation.
- **AI1013** (associatedRule resolution, Error) is cross-document — it resolves `notification.associatedRule.id` against the assembled `run.tool.driver.rules[]` — so it cannot be encoded in a single-object schema; it stays at `emit-finalize --validate`.

## Directory paths are not scanned artifacts (two bug fixes)

Giving the sample a real `workingDirectory` surfaced two latent bugs in the enrichment pass, the second masked by the first:

- **Empty-string hash for a directory.** `FileRegionsCache.GetHashData` hashed a path that resolves to a directory or missing file as the sha-256 of the *empty string* (`E3B0C442…`). It now returns `null` when there is no cached text.
- **Location-only artifact (SARIF2004).** Once the bogus hash was gone, the `workingDirectory` surfaced as a bare location in `run.artifacts` and tripped `Warning_EliminateLocationOnlyArtifacts`. Root cause: `AddFileReferencesVisitor` promoted *every* artifactLocation — including an invocation's `workingDirectory` — into `run.artifacts`. It now skips the `workingDirectory` (process context, not a scanned artifact), so it stays a standalone `{uri, uriBaseId}` and never lands in `run.artifacts`.

After both fixes the regenerated fixtures validate with **0 errors / 0 warnings / 0 notes**; `SampleCode.cs` is the sole artifact and `workingDirectory` is a standalone artifactLocation.

## Tests

`AIInputSchemaDriftTests`: per-schema drift facts (one `[Fact]` per clause walking a case table, no `[Theory]`). The invocation fact pins the requireds (including notification `timeUtc`) + accept of an inline-notification-carrying invocation + reject of non-objects + reject of notifications missing/empty `timeUtc` + that `endTimeUtc` is **not** required; `AddInvocationCommandTests` pins the matching receipt-gate behavior (including `missing-workingDirectory`, whitespace-`uri`, and notification-`timeUtc` rejects, and that a supplied notification `timeUtc` is preserved). `FileRegionsCacheTests` adds directory/missing-file null-hash regression cases; `AddFilesVisitorTests` adds a case asserting `workingDirectory` is not promoted into `run.artifacts`. The `CweGenerateSample` byte-identical fixtures were regenerated (the sample invocation now carries a real `workingDirectory` + an `arguments[]` array).

## Release notes

`NEW` (invocation + reporting-descriptor schemas) and `BRK` (removal of `add-notification`) entries in v5.0.3 (UNRELEASED), plus two `BUG` entries for the directory-hash and location-only-artifact fixes.


src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/microsoft/sarif-sdk/main/src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json",
+  "title": "SARIF AI emit profile \u2014 invocation object",
+  "description": "Input contract for a single SARIF invocation an AI producer supplies to the add-invocation verb. SARIF \u00a73.20 makes every invocation field optional, so the verb's receipt check is intentionally minimal: any JSON object is accepted. Whole-log invocation validation is enforced at emit-finalize --validate, not on a single invocation here.",
+  "type": "object"
```
michaelcfanning: really? we have no requirements for an invocation object? what about return value? start and end time? command line? working directory? see my comments offline.

src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/microsoft/sarif-sdk/main/src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json",
+  "title": "SARIF AI emit profile \u2014 reporting descriptor object",
+  "description": "Input contract for a SARIF reportingDescriptor an AI producer supplies to the add-reporting-descriptor verb. The verb stores it under run.tool.driver.notifications[] by default, or run.tool.driver.rules[] with --rules. SARIF \u00a73.49.3 requires a non-empty string id; the verb rejects a missing, null, non-string, or empty id. On the --rules path the verb additionally requires the id to begin with NOVEL- (AIRuleIdConvention.IsNovel); ai-novel-rule-descriptor.schema.json overlays this schema to pin that tighter gate.",
```
michaelcfanning: this description shouldn't talk about NOVEL rules, that's handled in the other schema. Did we agree to trim the description content in most cases?

ReleaseHistory.md
```diff
@@ -17,6 +17,7 @@ Entries are terse by design: one line per change, present-tense behavior, comple
 * BRK: `FileRegionsCache.PopulateTextRegionProperties` gains an `overwriteExistingData` parameter (default `false`) and `OptionallyEmittedData.OverwriteExistingData` now suppresses exceptions for divergence in pre-populated versus computed region properties.
 * BRK: Split `AI2011.DoNotPersistFingerprints` (Note) into `AI1007.DoNotPersistFingerprints` (Error; `result.fingerprints`) and `AI2011.DoNotPersistPartialFingerprints` (Warning; `result.partialFingerprints`).
 * BRK: Remove `AI1011.RedactedRunMarker`; the AI emit profile no longer validates the `ai/redacted` run marker or `ai/fullLogLocation`.
+* NEW: Add AI emit-profile input schemas for the `add-notification`, `add-invocation`, and `add-reporting-descriptor` verbs (notification, invocation, reporting-descriptor, and a novel-rule-descriptor overlay that `$ref`s the reporting-descriptor and adds the `--rules` `NOVEL-` id gate), each pinned to its C# receipt contract by a drift test.
```
michaelcfanning: we only need this lovely terse first sentence ` NEW: Add AI emit-profile input schemas for the `add-notification`, `add-invocation`, and `add-reporting-descriptor` verbs.`

michaelcfanning: removing add-notification of course!

src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json
michaelcfanning: Resolved in a0c05912. The schema now overlays the canonical SARIF invocation (allOf \, so exitCode/startTimeUtc/working dir etc. are all permitted via the base) and **requires** executionSuccessful + commandLine + workingDirectory (uri non-whitespace). endTimeUtc and inline notification timeUtc are intentionally not required — the verb auto-stamps them at emit time. Per your call we did not additionally require exitCode/arguments/startTimeUtc (left permitted-but-optional). The receipt gate + drift/command tests pin all three requireds.

src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json
michaelcfanning: Done — trimmed in cb1491ce; the description no longer mentions NOVEL rules (that gate lives in the ai-novel-rule-descriptor overlay).

ReleaseHistory.md
michaelcfanning: Done in cb1491ce — the NEW entry is now your terse one-sentence form, with add-notification dropped (see reply below). A separate BRK entry records the add-notification removal.

michaelcfanning: Yep — dropped add-notification from the NEW sentence; it now reads dd-invocation and dd-reporting-descriptor only.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
```diff
@@ -61,7 +77,7 @@ public int Run(AddInvocationOptions options, IFileSystem fileSystem = null)
                     string.Format(
                         CultureInfo.CurrentCulture,
                         "Appended invocation (executionSuccessful='{0}') to '{1}'.",
-                        executionSuccessful,
+                        executionSuccessful.ToString().ToLowerInvariant(),
```
michaelcfanning: how about execution successful ? "true" : "false" and save string construction

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
```diff
@@ -71,5 +87,71 @@ public int Run(AddInvocationOptions options, IFileSystem fileSystem = null)
                 return FAILURE;
             }
         }
+
+        // Mirrors the load-bearing requireds of ai-invocation.schema.json. Full structural
+        // validation is deferred to emit-finalize --validate; this is a cheap receipt gate that
+        // rejects the three fields the AI profile makes mandatory.
+        private static bool TryValidateInvocationReceipt(JObject payload)
```
michaelcfanning, line 77: so we do *not* prefer to validate against the schema> and prefer instead to duplicate what the schema enforces and then add whatever additional imperative validation is required?

what are we buying with this approach? speed?

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
```diff
@@ -71,5 +87,71 @@ public int Run(AddInvocationOptions options, IFileSystem fileSystem = null)
...
+                    "Invalid invocation: 'workingDirectory' is required and must be an artifactLocation with a non-whitespace 'uri'.");
+                return false;
+            }
+
+            return true;
+        }
+
+        // Fills wall-clock fields the producer left unset, using a single 'now' so endTimeUtc and
+        // every inline notification's timeUtc agree. Producer-supplied values are preserved.
+        private static void StampWallClock(JObject payload, string now)
```
michaelcfanning: How about `StampEndTimeUtc` instead of introducing a competing term, wallclock

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
```diff
@@ -71,5 +87,71 @@ public int Run(AddInvocationOptions options, IFileSystem fileSystem = null)
...
+            if (payload["endTimeUtc"] == null || payload["endTimeUtc"].Type == JTokenType.Null)
+            {
+                payload["endTimeUtc"] = now;
+            }
+
+            StampNotificationTimes(payload["toolExecutionNotifications"] as JArray, now);
+            StampNotificationTimes(payload["toolConfigurationNotifications"] as JArray, now);
+        }
+
+        private static void StampNotificationTimes(JArray notifications, string now)
```
michaelcfanning: 'StampNotificationsTimeUtc`

src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json
```diff
@@ -0,0 +1,43 @@
...
+      "pattern": "\\S"
+    },
+    "workingDirectory": {
+      "description": "The directory the process ran in, as a SARIF artifactLocation. Required by the AI profile (it anchors relative paths in the scan) even though SARIF \u00a73.20.19 makes it optional; the AI producer knows it. The uri must be a non-empty string.",
+      "type": "object",
+      "required": [
+        "uri"
+      ],
+      "properties": {
+        "uri": {
```
michaelcfanning, line 34: what's going on here? I don't understand this definition

src/Sarif.Multitool.Library/GetSchema/ai-novel-rule-descriptor.schema.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "Input contract for a reportingDescriptor an AI producer supplies to add-reporting-descriptor --rules (stored under run.tool.driver.rules[]). It overlays ai-reporting-descriptor.schema.json by $ref \u2014 inheriting the non-empty string id requirement \u2014 and adds the single tightening the --rules path enforces: the id must begin with the NOVEL- escape-hatch prefix (AIRuleIdConvention.IsNovel). Taxonomy-mapped rule descriptors (e.g. CWE-89) come from the finalize-time taxonomy enricher, not this verb. The prefix gate is matched as-is (StartsWith), not the full novel grammar.",
+  "allOf": [
+    {
+      "$ref": "https://raw.githubusercontent.com/microsoft/sarif-sdk/main/src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json"
+    }
+  ],
+  "properties": {
+    "id": {
+      "description": "Novel-finding descriptor id. The non-empty string requirement is inherited from ai-reporting-descriptor; this overlay adds the verb's IsNovel prefix gate. Prefix only, so 'NOVEL-foo/bar' is accepted here even though the richer result-side ruleId grammar rejects it.",
+      "pattern": "^NOVEL-"
```
michaelcfanning: there must be at least one character following this. and maybe we should constrain follow on values to hyphens and caps alpha, ensuring no trailing hyphen,

Did I get it right?

^NOVEL-[A-Z-]*[A-Z]

src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/microsoft/sarif-sdk/main/src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json",
+  "title": "SARIF AI emit profile \u2014 reporting descriptor object",
+  "description": "Input contract for a SARIF reportingDescriptor an AI producer supplies to the add-reporting-descriptor verb, stored under run.tool.driver.notifications[] (default) or run.tool.driver.rules[] (--rules). Requires a non-empty string id (SARIF \u00a73.49.3); a missing, null, non-string, or empty id is rejected.",
```
michaelcfanning, line 5: in general, it's a nice value for everything to have a proper reference to the specese for the construct.

michaelcfanning: meaning, nice work where that's been done, let's scrub everything to make it a consistent standard if not already

src/Sarif/Emit/SarifEventKinds.cs
```diff
@@ -30,22 +30,14 @@ public static class SarifEventKinds
...
-        /// A self-contained <see cref="Notification"/> destined for
-        /// <c>invocations[last].toolConfigurationNotifications</c>. The replay engine routes
-        /// events of this kind to the configuration-notifications array.
-        /// </summary>
-        public const string ConfigurationNotification = "configuration-notification";
-
-        /// <summary>
-        /// A complete <see cref="Invocation"/> object. Producers may append multiple
-        /// invocations per run.
+        /// A complete <see cref="Invocation"/> object. Producers may append multiple invocations
```
michaelcfanning: Can you run a nice constructively adversarial review to ensure all doc comments encode useful, relevant information with no duplication, over-explication, etc.

src/Sarif/Emit/SarifEventReplayer.cs
```diff
@@ -116,17 +111,11 @@ public static SarifLog Replay(IEnumerable<SarifEvent> events)
                         break;
 
                     case SarifEventKinds.Invocation:
+                        // The invocation carries its own toolExecutionNotifications /
```
michaelcfanning: unnecessary code comments. of course every object carries its properties

src/Sarif/Taxonomies/CweGenerateSample.ps1
```diff
@@ -380,18 +384,30 @@ foreach ($e in $events) {
...
-# timeUtc preset so SarifEventReplayer leaves it alone (AI2019 auto-stamp).
-$notifPayload = [ordered]@{
-    level   = 'note'
-    message = [ordered]@{
-        text     = "Analyzed $($events.Count) findings across 1 file."
-        markdown = "Analyzed **$($events.Count)** findings across **1** file."
-    }
-    timeUtc = '2024-01-01T00:00:00.000Z'
+# A single fully-formed invocation carries the run's notification INLINE on
+# toolExecutionNotifications. The add-notification verb was removed: SARIF has no
```
michaelcfanning: this code comment references historical edits, like removing add-notification. don't do that please! scrub for this issue elsewhere too

src/Sarif/Taxonomies/CweGHAzDoSample.sarif
```diff
@@ -143,6 +149,14 @@
         }
       },
       "artifacts": [
+        {
+          "location": {
+            "uri": "file:///src/"
+          },
+          "hashes": {
+            "sha-256": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855"
```
michaelcfanning: oh we have a hash of a directory? that's a bug

src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json
michaelcfanning: Addressed. ai-invocation.schema.json now requires `executionSuccessful`, `commandLine`, and `workingDirectory` (with a non-whitespace `uri`), and requires a producer-supplied `timeUtc` on every inline notification. `endTimeUtc` is intentionally not required — the verb auto-stamps it on omission.

src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json
michaelcfanning: Addressed. The base ai-reporting-descriptor description now covers only the non-empty-string `id` contract (SARIF 3.49.3); all NOVEL- grammar lives in the ai-novel-rule-descriptor overlay.

ReleaseHistory.md
michaelcfanning: Addressed. Trimmed to the single terse sentence: `NEW: Add AI emit-profile input schemas for the add-invocation and add-reporting-descriptor verbs.`

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Done — `executionSuccessful ? "true" : "false"`.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Design call, leaving this unresolved for your sign-off.

We keep the cheap imperative receipt gate rather than running full JSON-Schema validation at this verb. What it buys:

1. Speed and zero schema-runtime dependency on the per-event hot path.
2. One tailored stderr line per missing field instead of a schema validator's nested error blob.
3. Consistency with the verb-chain design: every add-* verb does a minimal receipt check, and full structural validation is the explicit opt-in gate at `emit-finalize --validate` (which runs the SARIF + AI rule kinds against the assembled log).

The receipt gate asserts only the three load-bearing requireds; the schema stays the source of truth for consumers and for finalize-time validation. Happy to switch to per-event schema validation if you'd rather pay that cost here.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Done — the helper is `StampEndTimeUtcIfOmitted` and the competing `wallclock` term is gone from the code (only a descriptive `wall-clock now` remains in the doc remark).

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Obsolete — the verb no longer stamps notification `timeUtc` at all. Producers supply it (the AI profile requires it on every inline notification) and the verb preserves it verbatim, so there is no notification-stamping method left to name.

src/Sarif.Multitool.Library/GetSchema/ai-invocation.schema.json
michaelcfanning: Clarified. `workingDirectory` is a SARIF artifactLocation object; the AI profile requires it and requires its `uri` to be a non-whitespace string (`pattern: \S`). I rewrote the description and dropped the redundant `minLength: 1` — the `\S` pattern is the real non-whitespace gate (mirroring the verb's IsNullOrWhiteSpace check); `minLength: 1` would wrongly accept a single space.

src/Sarif.Multitool.Library/GetSchema/ai-novel-rule-descriptor.schema.json
michaelcfanning: Addressed, with one deliberate divergence from the sketch. The pattern is now `^NOVEL-[a-z0-9]+(-[a-z0-9]+)*$` — at least one character after `NOVEL-`, constrained to a lowercase-alphanumeric kebab sub-id.

I went lowercase rather than caps on purpose: a rule descriptor `id` IS the `result.ruleId` that references it, and the canonical result-side NOVEL grammar (`AIRuleIdConvention.s_novelPrefix`) is lowercase-kebab. A caps tail here would never match a conformant result ruleId, so reusing the lowercase grammar keeps the descriptor id byte-identical to the ruleId it pairs with. The verb gate (`IsNovel` + `IsAcceptable`), the schema, and the drift/command tests now all enforce this same grammar.

src/Sarif.Multitool.Library/GetSchema/ai-reporting-descriptor.schema.json
michaelcfanning: Addressed. Spec references are now consistent across the AI schemas: invocation cites 3.20.x per field; ai-reporting-descriptor and the ai-novel-rule-descriptor overlay both cite 3.49.3 for `id`.

src/Sarif/Emit/SarifEventKinds.cs
michaelcfanning: Ran the pass. Trimmed the SarifEventKinds doc comments and the verb remarks to encode only non-obvious intent/invariants: removed the obvious property-restating comment in the replayer, scrubbed historical add-notification narration, and reframed the notification-routing prose as a positive invariant (notifications inline because SARIF has no run-level notifications array, plus parallel-invocation routing). What remains covers the load-bearing requireds and the novel-grammar gate rationale.

src/Sarif/Emit/SarifEventReplayer.cs
michaelcfanning: Done — removed.

src/Sarif/Taxonomies/CweGenerateSample.ps1
michaelcfanning: Done — scrubbed both occurrences in CweGenerateSample.ps1 (around lines 55 and 388) and the `there is no add-notification verb because…` framing in AddInvocationCommand and SarifEventKinds, restated as positive invariants. Swept the tree to confirm no other historical-edit references remain in code comments.

src/Sarif/Taxonomies/CweGHAzDoSample.sarif
michaelcfanning: Fixed on this branch. `FileRegionsCache.GetHashData` now returns null (rather than the sha-256 of the empty string) for a path that resolves to a directory or missing file with no cached text, so the regenerated samples no longer carry a directory hash. `CweGeneratedSampleTests` passes byte-identically.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
```diff
@@ -71,5 +81,90 @@ public int Run(AddInvocationOptions options, IFileSystem fileSystem = null)
...
+                            arrayName));
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        // Fills endTimeUtc with the time of receipt when the producer left it unset.
+        private static void StampEndTimeUtcIfOmitted(JObject payload, string now)
```
michaelcfanning: I think I'd expect to co-locate the date string construction here to verify it's in the proper SARIF friendly format. so change this helper to accept a DateTime param is my suggestion, then do the proper formatting here.

Also, this all assumes that we are in fact required to process this value as a string, not a DateTime?

src/Sarif.Multitool.Library/Emit/AddReportingDescriptorCommand.cs
```diff
@@ -91,16 +94,17 @@ public int Run(AddReportingDescriptorOptions options, IFileSystem fileSystem = n
...
                 }
 
-                // The --rules path is reserved for NOVEL- ids. Reject anything else with a
-                // tailored envelope that mirrors the AI-RULEID-001 shape but is specific to
-                // descriptor authoring (the result-side AIRuleIdConventionException message
-                // talks about taxonomy sub-id form, which doesn't apply to a descriptor id).
-                if (isRules && !AIRuleIdConvention.IsNovel(id))
+                // The --rules path is reserved for well-formed NOVEL- ids (full grammar, not just
+                // the prefix), so a descriptor id is byte-identical to the result ruleId that
+                // references it. The error envelope is tailored to descriptor authoring rather than
```
michaelcfanning: this comment doesn't encode anything useful in it: change my mind if you can.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Decision: going with the lightweight receipt gate at `add-invocation` (required-field checks: `executionSuccessful`, `commandLine`, `workingDirectory.uri`, and `timeUtc` on every inline notification), deferring full structural/schema validation to `emit-finalize --validate`. This keeps the per-event verbs cheap and fail-fast on the common authoring mistakes. We'll revisit heavier up-front schema validation if real-world usage shows it's needed.

src/Sarif.Multitool.Library/Emit/AddInvocationCommand.cs
michaelcfanning: Done in 57df333d. `StampEndTimeUtcIfOmitted` now accepts a `DateTime` and performs the SARIF formatting (`SarifDateTimeFormatMillisecondsPrecision`) itself, so the SARIF-friendly construction is co-located with the stamp; `Run` just passes `DateTime.UtcNow`. Output is byte-identical (18/18 AddInvocation tests pass).

On the string-vs-DateTime question: the producer-supplied value must stay a string. `TryReadJsonPayload` parses with `DateParseHandling.None` specifically to preserve ISO-8601 text exactly and stop Json.NET from normalizing date-looking strings; the payload flows through the event log as an opaque JToken and is replayed verbatim. Only the value the verb itself authors (the stamped `endTimeUtc`) is ours to format, which is what this helper now does.

Issue #2970, opened 2026-06-04
Author finalized whole-log AI SARIF schema (ai-log.schema.json) and serve via get-schema emit-finalize


michaelcfanning:
## Summary

Author a high-fidelity JSON Schema for a **fully-enriched SARIF AI log** — the output of `emit-finalize` — and serve it via `get-schema emit-finalize` (`ai-log.schema.json`). This completes the `get-schema` verb→schema mapping to **6↔6** (every emit verb has a schema): the input verbs serve *what you feed them*; `emit-finalize` serves *what it emits*.

This is the immediate fast-follow to the `get-schema` + verb/schema-rename PR (branch `ai-emit-get-schema`), which intentionally ships only the **five input** schemas. The `get-schema` verb→schema map is built as an extensible dictionary, so wiring `emit-finalize → ai-log` is a one-line add here.

## Why this is a distinct artifact (not a union of the input schemas)

Finalize **adds** exactly the fields the per-object input schemas **reject**:

- `result.ruleIndex` and `result.guid` — replayer-assigned (input `ai-result` bounces producer-supplied values).
- Taxonomy rule descriptors (e.g. `CWE-89`) — injected by the taxonomy enricher; producers cannot author them (that's why `add-rule-reporting-descriptor` is NOVEL-only).
- `run.taxonomies[]`, region enrichment (`snippet`, `contextRegion`, sha-256), optionally embedded `artifacts[].contents.text`.
- `run.properties["ai/origin"]`, `automationDetails` + the four `azuredevops/pipeline/build/*` GHAzDO props.

So `ai-log.schema.json` is a **post-enrichment** contract: in effect, *"what `emit-finalize --validate` enforces, expressed as one whole-log JSON Schema."* It's the headline artifact a consumer (e.g. GHAzDO ingestion) validates the final log against.

## Scope

- [ ] Author `src/Sarif.Multitool.Library/GetSchema/ai-log.schema.json` (Draft 2020-12 overlay `$ref`-ing the schemastore SARIF base, tightened to the AI Error-level rules at whole-log scale).
- [ ] Pin it with a whole-log drift test (sibling to `AIInputSchemaDriftTests`) that validates accept/reject verdicts against the C# validation rules (AI100x, AIRuleIdConvention, GHAzDO1019/1020, ai/origin) using JsonSchema.Net.
- [ ] Embed it as a resource and wire `emit-finalize → ai-log` into `get-schema` (+ extend the in-sync test to 6↔6).
[...]

PR #2971, opened 2026-06-04
Add get-schema verb and align AI emit verb/schema names

michaelcfanning:
## What

Adds a `multitool get-schema <verb>` verb and brings the AI emit surface into verb↔schema naming parity. Follows #2968 (which added the five AI input schemas). No back-compat aliases — this surface has no real consumers yet, so each rename/split is a clean `BRK:`.

### `get-schema` (NEW)

`get-schema <verb>` streams the embedded JSON Schema that validates a named emit verb's input to stdout (raw bytes, no trailing newline) or to `--output` (refuses to clobber without `--force-overwrite`). `--list` enumerates the verbs that have a schema. The served bytes are the assembly's embedded resources, byte-identical to the schema files under `GetSchema/`.

### Verb ↔ schema parity

| verb | schema |
|---|---|
| `emit-run` (was `emit-init-run`) | `ai-run` |
| `add-result` | `ai-result` |
| `add-invocation` | `ai-invocation` |
| `add-notification-reporting-descriptor` (was `add-reporting-descriptor`) | `ai-notification-reporting-descriptor` |
| `add-rule-reporting-descriptor` (was `add-reporting-descriptor --rules`) | `ai-rule-reporting-descriptor` |
| `emit-finalize` | reserved — `ai-log`, fast-follow (#2970) |

- **`emit-init-run` → `emit-run`.**
- **`add-reporting-descriptor` split** into `add-notification-reporting-descriptor` and `add-rule-reporting-descriptor`; the `--rules` flag is gone (the verb selects the target array and id gate). Shared body factored into the internal `ReportingDescriptorEmitter`.
- **Schema files renamed** to mirror their verbs. The rule schema is now standalone (dropped the cross-schema `allOf`/`$ref`).

`emit-finalize` is reserved in the catalog with a `null` schema this release; serving it returns a failure that points at #2970. Adding its whole-log `ai-log` schema next is a one-line catalog change.

## Why the descriptor split

Producer-authored descriptors are only notifications (default) and novel rules. Taxonomy/CWE rule descriptors are SDK-injected at finalize by the taxonomy enricher, never producer-authored — so a single `--rules` toggle conflated two distinct contracts. The split makes the target array and the id gate (non-empty string vs. well-formed `NOVEL-` grammar) explicit per verb.

## Tests / verification

- **New in-sync test** (`GetSchemaCommandTests`, 8 `[Fact]`s): for every catalog verb, bytes are identical across the embedded resource, the on-disk `GetSchema/` file, and what the verb writes; plus catalog↔embedded-manifest parity (the authoritative anti-drift check) and catalog↔disk parity, so the surface cannot silently drift. Also covers `--list` content, unknown/no-verb failure, the reserved-verb pending path, and the overwrite guard.
- Drift and descriptor tests updated for the renames/split.
- Multitool builds clean; `Test.UnitTests.Sarif.Multitool.Library` green; ReleaseHistory style test green.
- Both Cwe sample fixtures regenerate **byte-identical** (verb name is not part of SARIF output).

## Follow-up

- #2970 — finalized whole-log `ai-log.schema.json` for `emit-finalize --validate`.


skills/emit-sarif-findings/SKILL.md
```diff
@@ -38,13 +38,13 @@ Apply this skill when an agent is the **originating** detector (not post-process
 
 ## Method
 
-The skill uses four multitool verbs: `emit-init-run` → `add-result` / `add-invocation` (per finding or scan phase) → `emit-finalize --validate`. Each verb either appends to an event log (`<output>.wip.jsonl`) or replays the log into a finished SARIF file.
+The skill uses four multitool verbs: `emit-run` → `add-result` / `add-invocation` (per finding or scan phase) → `emit-finalize --validate`. Each verb either appends to an event log (`<output>.wip.jsonl`) or replays the log into a finished SARIF file.
```
michaelcfanning: why doesn't this reference the other emit verb? for notification and rule reporting descriptors?

src/Sarif.Multitool.Library/Emit/AddNotificationReportingDescriptorOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Options for <c>add-notification-reporting-descriptor</c>, which appends a SARIF
+    /// <c>reportingDescriptor</c> to <c>run.tool.driver.notifications[]</c> in a staged event log
+    /// (<c>&lt;output&gt;.wip.jsonl</c>) created by <c>emit-run</c>.
+    /// </summary>
+    /// <remarks>
+    /// The descriptor is supplied as a JSON document (file via <c>--input</c> or piped on stdin).
+    /// Each <c>id</c> may appear at most once in the notifications array.
+    /// </remarks>
+    [Verb("add-notification-reporting-descriptor", HelpText = "Append a SARIF reportingDescriptor (JSON) to run.tool.driver.notifications[] in a staged event log.")]
+    public class AddNotificationReportingDescriptorOptions
```
michaelcfanning: All these emit things have an input and output arg yes? Base class?

src/Sarif.Multitool.Library/Emit/EmitFinalizeOptions.cs
```diff
@@ -39,7 +39,7 @@ public class EmitFinalizeOptions
 
         [Option(
             "srcroot",
-            HelpText = "Rewrite originalUriBaseIds[\"SRCROOT\"].uri to this value AFTER enrichment runs. Use a local file:// SRCROOT on emit-init-run so InsertOptionalDataVisitor can read sources for snippets/hashes, then pass the canonical portable URI here (e.g. https://github.com/<org>/<repo>/blob/<branch>/) so the shipped SARIF anchors at a stable, host-independent location.")]
+            HelpText = "Rewrite originalUriBaseIds[\"SRCROOT\"].uri to this value AFTER enrichment runs. Use a local file:// SRCROOT on emit-run so InsertOptionalDataVisitor can read sources for snippets/hashes, then pass the canonical portable URI here (e.g. https://github.com/<org>/<repo>/blob/<branch>/) so the shipped SARIF anchors at a stable, host-independent location.")]
```
michaelcfanning, line 42: this raises a new validation requirement, which is that on first receipt, SRCROOT should resolve to an observable disk location. then we have confidence finalize will enrich properly.

src/Sarif.Multitool.Library/GetSchema/GetSchemaCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
+namespace Microsoft.CodeAnalysis.Sarif.Multitool
+{
+    /// <summary>
+    /// Implements <c>multitool get-schema</c>: emits the embedded JSON Schema that validates the
```
michaelcfanning: I believe that `sarif` is the dotnet tool verb if dotnet tool installed, and Sarif.Multitool is the package name for dnx commands.

skills/emit-sarif-findings/SKILL.md
michaelcfanning: Done in de8cb7e2. The verb list now names dd-notification-reporting-descriptor and dd-rule-reporting-descriptor, and a new `Step 4 — Register reporting descriptors` documents that the first appends to `run.tool.driver.notifications[]` and the second appends a `NOVEL-` rule to `run.tool.driver.rules[]` (taxonomy/CWE rule descriptors stay SDK-injected at finalize).

src/Sarif.Multitool.Library/Emit/AddNotificationReportingDescriptorOptions.cs
michaelcfanning: Yes — factored `EmitInputOptionsBase` (`[Value(0)] OutputFilePath` + `[Option('i',\input\)] InputFilePath`). All five emit option classes inherit it; `EmitRunOptions` adds only `--force-overwrite`. CommandLineParser discovers the inherited attributes (same pattern as the repo's `SingleFileOptionsBase`). de8cb7e2.

src/Sarif.Multitool.Library/Emit/EmitFinalizeOptions.cs
michaelcfanning: Implemented in de8cb7e2. `emit-run` now rejects a `file:` `originalUriBaseIds[\SRCROOT\]` whose path does not resolve to an existing directory on disk at receipt (`TryValidateSourceRootResolvesOnDisk`), so finalize has confidence it can enrich result locations against an observable checkout. Added passing/failing receipt tests; `https` and absent SRCROOT are unaffected. Terse BRK bullet added to ReleaseHistory.

src/Sarif.Multitool.Library/GetSchema/GetSchemaCommand.cs
michaelcfanning: Correct — the dotnet tool command is `sarif` (per `DotnetToolSettings.xml`) and `Sarif.Multitool` is the package/assembly. Fixed the inaccurate `multitool <verb>` wording: the command doc-comment summaries across the emit family and get-schema now read `Implements <verb>`. de8cb7e2.

Issue #2972, opened 2026-06-05
Fast-follow: broaden VCP-driven emit-finalize rebasing (receipt-time validation, non-github hosts, multi-repo/submodule fixtures)


michaelcfanning:
Follow-up work deferred from the VCP-driven `emit-finalize` rebasing change (branch `ai-emit-vcp-rebase`). That PR landed the finalize-time rebase visitor, conditional `mappedTo=SRCROOT` stamping at `emit-run`, the Cwe fixture regen, and ReleaseHistory. `emit-finalize` now HARD-enforces the VCP + `mappedTo` contract at the terminal step, so the intermediate state is safe (clear failure, never a silent machine-path leak). The items below strengthen and broaden it.

### 1. Receipt-time VCP/`mappedTo` validation at `emit-run`
Today `emit-finalize` is the first place the "≥1 `versionControlProvenance` entry, each with `mappedTo` binding a declared `originalUriBaseIds` root" contract is enforced. Move a (warn-or-fail TBD) check to `emit-run` receipt so producers get feedback at the point of authorship rather than at finalize. Entangles with the ~53 emit-run tests and the ADO-stamping flow — scope carefully.

### 2. `github.com`-host early failure at `emit-run`
`emit-finalize` portable-URL derivation is github.com-only (v1). A non-github `repositoryUri` only fails at finalize. Consider surfacing an early, host-aware diagnostic (or explicit "rebasing unsupported for this host") at `emit-run`.

### 3. Non-github portable URL construction (ADO, GHE)
Derive portable permalinks for `dev.azure.com` (query-string `?path=&version=GC<sha>` addressing) and GitHub Enterprise hosts. `emit-run` already stays host-agnostic; only finalize derivation is github-only.

### 4. Multi-repo / submodule fixtures + determinism
- Add multi-repo and submodule sample fixtures (`SRCROOT_<REPO>` minting, nested-checkout longest-prefix attribution).
- Multi-repo collision-suffix (`SRCROOT_<LEAF>_2`) is currently assigned in `versionControlProvenance` array order. Consider a stable ordering key (normalized `repositoryUri` + local root tie-breaker) so same-leaf repos (`org-a/utils`, `org-b/utils`) get deterministic suffixes independent of array order.
- Submodule repos can nest other repos; verify longest-prefix attribution and base minting hold.

### 5. Chained-base-alias base-only locations
[...]

PR #2974, opened 2026-06-06
Harden emit-finalize portable-root host handling and reject embedded credentials

michaelcfanning:
## What & why

Centralizes emit-finalize portable-root derivation in a new VcpPortableRoot helper and tightens the supported-host contract so a finalized SARIF can never ship a credential or a confidently-wrong repository link.

### Host scope (explicit allow-list)
| Family | Accepted | Rejected |
| --- | --- | --- |
| GitHub | github.com, <slug>.ghe.com (data-residency / EMU) | custom-hostname GitHub Enterprise Server |
| Azure DevOps | dev.azure.com/<org>/<project>/_git/<repo> only | legacy <org>.visualstudio.com (insecure), on-prem/collection prefixes |

Users must supply the `dev.azure.com` URL, and that is the only Azure DevOps form produced.

### repositoryUri hardening (before a root is derived)
- **https-only** — `http` is rejected.
- **Embedded credentials rejected outright.** Any `account@`, `account:password@`, or token userinfo on an https repositoryUri is **rejected**, not stripped — a credential must never ride in a repositoryUri, and an account segment is not part of the repository identity.
- GitHub **ssh/scp** clone URLs normalized to https (ssh login dropped); an explicit port, query, or fragment (which the https form cannot represent) is rejected rather than silently dropped.
- **scp authority** validated tightly: `git@evil@host` (multi-`@`), `user:pass@host` (colon-before-`@`), empty user, and bad host characters are all rejected — so a crafted login cannot smuggle a different host or echo a secret in a diagnostic.
- Azure DevOps ssh is rejected with a pointer to the https clone URL (normalization deferred to #2972).

### Credentials: reject caller input, sanitize the env boundary
A repositoryUri is a caller-authored identity, so a credential on it is a misconfiguration we fail closed on. But `BUILD_REPOSITORY_URI` is untrusted pipeline-environment input: modern Azure DevOps emits no userinfo, while legacy/migrated or PAT-auth setups emit `https://<org>@dev.azure.com/...` or a token. `emit-run` strips that userinfo at the env boundary in `AdoPipelineContext` before stamping `versionControlProvenance`, so the env-driven path keeps working with a clean identity while only caller-authored credentials fail the receipt gate.

### Finalized run
`emit-finalize` normalizes a GitHub ssh/scp `repositoryUri` to its https identity in the finalized run; a clean `https://github.com` URL is left **byte-identical** (guards the Cwe sample byte-gate). `emit-run` validates the same shapes at receipt and fails closed on an unsupported host, embedded credential, or malformed path.

## Validation
- Full `Test.UnitTests.Sarif.Multitool.Library`: **339 pass / 1 skip** (incl. Cwe byte-gate + ReleaseHistory style).
- `Release -p:EnforceCodeStyleInBuild=true`: clean (no IDE0005).
- New tests cover `*.ghe.com` success, custom-GHES reject, legacy `visualstudio.com` reject, on-prem reject, https credential reject (GitHub + ADO), ssh explicit-port/query reject, scp smuggled-host reject, scp password-authority reject (no secret in diagnostics), and env-boundary userinfo strip (`<org>@dev.azure.com` -> clean identity).

## Deferred -> #2972
ADO ssh normalization; custom-hostname GHES; multi-repo/submodule fixtures.


src/Sarif.Multitool.Library/Emit/EmitRunCommand.cs
```diff
@@ -155,6 +155,13 @@ public int Run(EmitRunOptions options, IFileSystem fileSystem = null)
                     }
                 }
 
+                // Receipt-time shape gate: every present repositoryUri (caller-supplied or
```
michaelcfanning: gee did we basically recap the whole doc XML comment here? :)

michaelcfanning: Ha, guilty — it was a near-verbatim recap of the method's XML doc. Removed the inline comment entirely; the method name + doc carry it. Also ran the ruthless pass over the rest of this segment's new comments (the AdoPipelineContext env-strip and its test) to trim them to current-design intent. Fixed in e232d294.

PR #2978, opened 2026-06-07
Surface skill descriptions in get-skill --list

michaelcfanning:
## What

get-skill --list now prints each skill's frontmatter `description` beside its name in an aligned two-column layout:

```
Available skills:
  emit-sarif         Serialize AI-detected security findings as SARIF v2.1.0 conforming to the AI-generated-findings profile, using the Sarif.Multitool emit verbs.
  publish-to-ghazdo  Uploads a finalized SARIF file to GitHub Advanced Security for Azure DevOps (GHAzDO) using the Sarif.Multitool publish-to-ghazdo verb.
  validate-sarif     Validates SARIF files against the SARIF 2.1.0 schema and the AI-generated-findings profile rules shipped by Sarif.Multitool.
```

## Why

The catalog previously emitted bare names. Names are intentionally terse, so an agent consuming the catalog had to fetch every skill and parse its frontmatter just to triage relevance. Surfacing the description inline lets a consumer self-select from a single `--list` call.

## Design

- **Single source of truth.** The description is read from each embedded skill's YAML frontmatter `description:` scalar -- the same value a consumer already reads from the emitted document. Nothing new to author or keep in sync.
- **Graceful degradation.** A skill that opens no frontmatter, declares no description, or uses a multi-line block scalar falls back to name-only rather than emitting a dangling indicator.
- **Conciseness.** Trimmed the `publish-to-ghazdo` frontmatter description to one clause to match its siblings; the env-var bearer-secret detail is a how (covered in the skill body), not a selection signal.

## Tests

Added direct `ExtractFrontmatterDescription` unit facts (unquoted scalar, quote stripping, no-frontmatter, block-scalar, post-frontmatter `description:`) and an integration fact asserting each catalog skill's description is surfaced beside its name. Full GetSkill suite: 20/20 green. Release build clean (0/0).

## Versioning

Bump to **5.0.4**.

### DIFF
--- a/src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
+++ b/src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
@@ -61,13 +61,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
         // `description:` appearing in skill prose cannot leak into the catalog.
         private static readonly Regex s_frontmatterDescription = new Regex(
             @"^description:[ \t]*(?<value>.+?)[ \t]*$",
-            RegexOptions.Compiled);
+            RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.ExplicitCapture);
 
         // A bare YAML block-scalar header (">", "|", optionally with a chomping/indentation indicator)
         // means the value spills onto following lines; the terse catalog skips such a description.
         private static readonly Regex s_blockScalarHeader = new Regex(
             @"^[>|][0-9+\-]*$",
-            RegexOptions.Compiled);
+            RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.ExplicitCapture);
 
         public int Run(GetSkillOptions options, IFileSystem fileSystem = null)
         {
--- a/src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
+++ b/src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
@@ -56,6 +56,19 @@ namespace Microsoft.CodeAnalysis.Sarif.Multitool
             @"\+(?<sha>[0-9a-fA-F]{7,64})",
             RegexOptions.Compiled);
 
+        // The frontmatter `description:` scalar surfaced beside each skill in the get-skill --list
+        // catalog. Matched only against lines inside the leading YAML frontmatter block, so a
+        // `description:` appearing in skill prose cannot leak into the catalog.
+        private static readonly Regex s_frontmatterDescription = new Regex(
+            @"^description:[ \t]*(?<value>.+?)[ \t]*$",
+            RegexOptions.Compiled);
+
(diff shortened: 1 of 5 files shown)


src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
```diff
@@ -56,6 +56,19 @@ public class GetSkillCommand : CommandBase
             @"\+(?<sha>[0-9a-fA-F]{7,64})",
             RegexOptions.Compiled);
 
+        // The frontmatter `description:` scalar surfaced beside each skill in the get-skill --list
+        // catalog. Matched only against lines inside the leading YAML frontmatter block, so a
+        // `description:` appearing in skill prose cannot leak into the catalog.
+        private static readonly Regex s_frontmatterDescription = new Regex(
+            @"^description:[ \t]*(?<value>.+?)[ \t]*$",
+            RegexOptions.Compiled);
```
michaelcfanning: please add invariant culture and explicitnamedgroup

src/build.props
```diff
@@ -13,8 +13,8 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF SDK</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix>5.0.3</VersionPrefix>
-    <PreviousVersionPrefix>5.0.2</PreviousVersionPrefix>
+    <VersionPrefix>5.0.4</VersionPrefix>
+    <PreviousVersionPrefix>5.0.3</PreviousVersionPrefix>
```
michaelcfanning: wait we clobbered previousversionprefix previously and build theater, non-functional. did it come back?

src/Sarif.Multitool.Library/GetSkill/GetSkillCommand.cs
michaelcfanning: Done in 6c30fd92 — both new regexes (`s_frontmatterDescription`, `s_blockScalarHeader`) now carry `RegexOptions.CultureInvariant | RegexOptions.ExplicitCapture` alongside `Compiled`. Scoped to the two regexes this PR introduced.

src/build.props
michaelcfanning: Good catch, and yes — it came back. History:

- #2899 (`d4af9789`, *"zap PreviousVersionPrefix theater"*) deliberately deleted it plus its "not because our MSBuild files use it" comment.
- The revert in #2907 (`fe70ce8b`) resurrected it (`+ <PreviousVersionPrefix>4.6.3</PreviousVersionPrefix>`), and it has ridden along since.

Confirmed still non-functional: a repo-wide grep for `PreviousVersionPrefix` returns exactly one hit — its own declaration. Nothing reads the property. So in 6c30fd92 I re-zapped it (removed the line) rather than keep bumping a dead property. `VersionPrefix` stays at 5.0.4.

PR #2983, opened 2026-06-08
Name the canonical sub-id fallback in the AI-RULEID-001 envelope

michaelcfanning:
When an AI producer emits a bare `CWE-<n>` ruleId, the emit chain rejects it with the `AI-RULEID-001` envelope. The envelope told the producer to add a sub-classifier but never said what to do when it has no sharper sub-pattern — leaving the obvious fallback (kebab-case the CWE name) as inference rather than instruction.

## What changed

The envelope now states the fallback explicitly and canonically: for `CWE-89`, `CWE-89/sql-injection`.

The framing preserves the deliberate design — **the emit chain never fills the sub-id in for the producer** (Option B, not auto-injection). Authoring the slug, even the generic fallback, is the producer's on-record classification: it records that a finer sub-classification was *weighed and the generic one chosen*, rather than that sub-classification was never considered. That explicit-declination signal is exactly what silent auto-injection would erase.

The `AI1012` validation rule already taught this fallback (\otherwise the kebab-cased rule name\); this aligns the emit-time envelope with the in-tool validator so both tell one story.

## Files

- `src/Sarif/Emit/AIRuleIdConventionException.cs` — adds the canonical fallback lines under shape #1 of the envelope.
- `docs/AI-RuleId-Convention.md` — softens the scolding *Why*, blesses the kebab-name fallback as a conformant (not lesser) choice, adds an accepted-example row, and syncs the envelope sample block.
- `src/Test.UnitTests.Sarif/Emit/AIRuleIdConventionTests.cs` — asserts the envelope now contains the fallback recipe and the `CWE-89/sql-injection` example.
- `ReleaseHistory.md` — NEW bullet.

## Verification

- `Test.UnitTests.Sarif` ~`AIRuleIdConvention`: 44/44 pass.
- Core `Sarif.csproj` Release build with `EnforceCodeStyleInBuild=true`: 0 warnings / 0 errors.
- ReleaseHistory style test: pass.


michaelcfanning:
Retired the hand-coined `AI-RULEID-001` error code. The emit-time `AIRuleIdConventionException` now reports **`AI1012`** — the canonical `ProvideRuleSubId` rule id — so the emit-time throw and the post-hoc `sarif validate` surface one id instead of two parallel coinages for the same concern.

Most call sites already referenced `AIRuleIdConventionException.ErrorCode`, so flipping the constant cascaded; also updated the prose comments, the convention-doc sample output, the copilot-instructions moniker, and the `SarifEventReplayer` test (asserts the constant now, not a literal). ReleaseHistory references corrected in place (v5.0.2-4 are unreleased).

Also merged `dev` to absorb #2985 (RollingHash fingerprints), resolving the ReleaseHistory conflict by keeping both NEW bullets. Affected suites green: AIRuleIdConvention + SarifEventReplayer (62), emit/AI Multitool.Library tests (94), ReleaseHistory style (2).

PR #2985, opened 2026-06-08
Add --insert RollingHashPartialFingerprints to stamp primaryLocationLineHash

michaelcfanning:
## What

Adds `--insert RollingHashPartialFingerprints` (also reachable via `--insert All`) to `InsertOptionalDataVisitor`, stamping each result's `primaryLocationLineHash` partial fingerprint using the CodeQL rolling line-hash.

GitHub code scanning (GHAS) and GitHub Advanced Security for Azure DevOps (GHAzDO) consume the `primaryLocationLineHash` key to match results across runs. The enum member `OptionallyEmittedData.RollingHashPartialFingerprints` already existed and is already part of the `All` composite, but nothing ever wired it to emit a value — this PR closes that gap.

## How

- For each result, when the flag is set (and the fingerprint is absent or `OverwriteExistingData` is also set), resolve the primary location's artifact, read its text through `FileRegionsCache`, compute `HashUtilities.RollingHash`, and store the hash for the primary location's `startLine`.
- Rolling hashes are **memoized per file** so a source file shared by many results is read and hashed only once per run.
- Results whose primary location lacks a `startLine` are **skipped** (CodeQL parity — no defaulting to the first line).
- The `primaryLocationLineHash` key is fixed by the consumers and carries **no version suffix** (documented on the const).

## Tests

- New `[Fact]`s in `InsertOptionalDataVisitorTests` covering the stamp, the skip-when-no-startLine case, memoization, and overwrite semantics.
- Rebaselined `Rel_All.sarif` expected output.
- Full `Test.UnitTests.Sarif` suite: 922 pass / 3 skip / 0 fail.

## Release note

`NEW: --insert RollingHashPartialFingerprints (also in All) stamps each result's primaryLocationLineHash partial fingerprint using the CodeQL rolling line-hash, consumed by GitHub code scanning and GHAzDO for result matching. Results lacking a primary-location startLine are skipped.`


michaelcfanning:
Follow-up commit (`b75b5e43`): fingerprint region-less results from line 1 to match the canonical `github/codeql-action`.

While verifying the `:N` occurrence suffix against `src/fingerprints.ts`, I found a parity bug: my port **skipped** any result whose primary location lacked a `region.startLine`, and a code comment falsely claimed that matched CodeQL. The canonical `locationUpdateCallback` does the opposite — a missing `startLine` defaults to **line 1** ("the alert pertains to the entire file"), so the result is fingerprinted, not dropped.

Changes:
- Default a missing/non-positive `startLine` to 1 before computing the rolling hash; gated only on a resolvable artifact location.
- Corrected the comment.
- Ported the `fingerprints.test.ts` `missingRegions` semantics: the two former skip assertions are now line-1 assertions, plus a genuine skip case (unresolvable artifact) and a mixed end-to-end pass.

Full `Test.UnitTests.Sarif` suite: 924 pass / 0 fail / 3 skip.

PR #2784, opened 2024-02-15
Add functional test data for new validation rules + code polish

EasyRhinoMSFT:

PR #2989, opened 2026-06-09
Make RuleKind a [Flags] enum and remove RuleKind.Gh

michaelcfanning:
## What

Cleans up the `RuleKind` enum:

- Makes it a `[Flags]` enum with single-bit values: `None=0, Sarif=1, Ghas=2, GHAzDO=4, AI=8`.
- Removes the stray `RuleKind.Gh`.
- Repoints the seven GH#### rules that used `Gh` (GH1001–GH1007) onto `RuleKind.Ghas`, so every GitHub Advanced Security validation rule now reports a single, consistent kind.

## Why

`Gh` and `Ghas` were two kinds for one family — `GH1001–GH1007` claimed `Gh` while `GH1011`/`GH1013–GH1018`/`GH2012` already claimed `Ghas`. Folding them onto `Ghas` lets a consumer (e.g. an AI validation rule, or `emit-finalize`) reason about "is GHAS in play" by testing for a single kind. Making the enum `[Flags]` gives the kinds clean, independent bit values.

## Compatibility

`BRK`. `RuleKind.Gh` is gone; callers that selected it must use `RuleKind.Ghas`. `GHAzDO` keeps value `4`, so the obsolete `RuleKind.Ado` alias and its case-insensitive `--rule-kind` parsing are unchanged. CLI `--rule-kind` parsing is by name, so the renumbering of `AI` (5→8) and `Ghas` (3→2) is not observable through the command line or any persisted artifact.

## Validation

- `RuleKindTests` 3/3 (Ado-alias value + parse, new single-bit `[Flags]` invariant guard).
- Functional `GH100*` validation tests 14/14 (GH1001–GH1007 still fire under `--rule-kind Ghas`).
- Release build with `EnforceCodeStyleInBuild=true` clean (0/0).
- `ReleaseHistory.md` `UNRELEASED` `BRK` bullet added.

### DIFF
--- a/src/Sarif/RuleKind.cs
+++ b/src/Sarif/RuleKind.cs
@@ -9,16 +9,17 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// Identifies the validator family that applies a SARIF reporting-descriptor rule.
     /// </summary>
     /// <remarks>
-    /// Not a <c>[Flags]</c> enum; combinations are expressed via <see cref="RuleKindSet"/>.
+    /// <c>[Flags]</c>: each kind is an independent bit. The active kinds for an
+    /// analysis are expressed as a <see cref="RuleKindSet"/>.
     /// </remarks>
+    [Flags]
     public enum RuleKind
     {
         None = 0,
         Sarif = 1,
-        Gh = 2,
-        Ghas = 3,
+        Ghas = 2,
         GHAzDO = 4,
-        AI = 5,
+        AI = 8,
 
         [Obsolete("Use RuleKind.GHAzDO instead.")]
         Ado = GHAzDO,
--- a/src/Sarif/RuleKind.cs
+++ b/src/Sarif/RuleKind.cs
@@ -8,10 +8,6 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// <summary>
     /// Identifies the validator family that applies a SARIF reporting-descriptor rule.
     /// </summary>
-    /// <remarks>
-    /// <c>[Flags]</c>: each kind is an independent bit. The active kinds for an
-    /// analysis are expressed as a <see cref="RuleKindSet"/>.
-    /// </remarks>
     [Flags]
     public enum RuleKind
     {
@@ -20,6 +16,9 @@ namespace Microsoft.CodeAnalysis.Sarif
         Ghas = 2,
         GHAzDO = 4,
         AI = 8,
+
+        [Obsolete("Use RuleKind.Ghas instead.")]
+        Gh = Ghas,
 
         [Obsolete("Use RuleKind.GHAzDO instead.")]
         Ado = GHAzDO,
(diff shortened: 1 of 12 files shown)


src/Sarif/RuleKind.cs
```diff
@@ -9,16 +9,17 @@ namespace Microsoft.CodeAnalysis.Sarif
     /// Identifies the validator family that applies a SARIF reporting-descriptor rule.
     /// </summary>
     /// <remarks>
-    /// Not a <c>[Flags]</c> enum; combinations are expressed via <see cref="RuleKindSet"/>.
+    /// <c>[Flags]</c>: each kind is an independent bit. The active kinds for an
```
michaelcfanning: Unnecessary comment, you just defined a [Flags] enum here against in readable terms.

src/Sarif/RuleKind.cs
```diff
@@ -9,16 +9,17 @@ namespace Microsoft.CodeAnalysis.Sarif
...
         None = 0,
         Sarif = 1,
-        Gh = 2,
-        Ghas = 3,
+        Ghas = 2,
         GHAzDO = 4,
-        AI = 5,
+        AI = 8,
 
         [Obsolete("Use RuleKind.GHAzDO instead.")]
```
michaelcfanning, line 23: Want to leave behind a [Obsolete]Gh = Ghas; back compat item? We don't need to do this unless GH shipped pre-%.x

src/Test.UnitTests.Sarif/RuleKindTests.cs
```diff
@@ -30,6 +30,21 @@ public void RuleKind_AdoAlias_ParsesCaseInsensitively()
             AssertParsesToGHAzDO("ADO");
         }
 
+        [Fact]
+        public void RuleKind_IsFlags_WithDistinctSingleBitValues()
```
michaelcfanning, line 43: [Flags] s/be [Flags] test? ok. low value but ok.

src/Sarif/RuleKind.cs
michaelcfanning: Done — dropped the <remarks> block in ab87af5.

src/Sarif/RuleKind.cs
michaelcfanning: Yes — Gh shipped in v5.0.4 (it's in origin/main, introduced by #2784), so I kept it as `[Obsolete("Use RuleKind.Ghas instead.")] Gh = Ghas` for source back-compat, mirroring `Ado = GHAzDO`. Value stays 2, so existing references compile unchanged. Added a Gh-alias equivalence fact too.

src/Test.UnitTests.Sarif/RuleKindTests.cs
michaelcfanning: Kept it — guards the single-bit/non-overlapping invariant so a future edit can't silently re-collide a kind (the bug we just fixed).

PR #2990, opened 2026-06-09
Check for successive null bytes in IsTextualData

ycholette:
IsTextualData returns true for a lot of dlls, like Microsoft.CodeAnalysis.ExternalAccess.RazorCompiler.dll 5.300.26.7805.  
 
I've added a check for 4 successive number of null bytes and it seems to be much more robust.


michaelcfanning:
After a struggle on this topic in other contexts, we eventually simply inverted detection by building confidence we have binary data, not textual. UTF16/32 present complications, of course, but if you provide the optional BOM this data can be clearly marked as textual. Check out #2993 which I created based on your observations and other experience, your comments/review appreciated.

PR #2993, opened 2026-06-10
Define FileEncoding.IsTextualData as the inverse of a new IsBinaryData

michaelcfanning:
Inspired by @ycholette (thank you!).

## Summary

`FileEncoding.IsTextualData` previously decoded the buffer under UTF-16/UTF-32 and inspected for control characters and `U+FFFD` replacement chars. That positive-detection heuristic produced **false negatives on real binaries** — e.g. it classified some DLLs (`Microsoft.CodeAnalysis.ExternalAccess.RazorCompiler.dll`) as textual, because some binary data happens to decode as valid UTF-16/UTF-32 without a replacement char.

Detecting "is text" positively is arguably harder than detecting "is binary" so perhaps we can invert the contract.

## What changed

- **New public API:** `FileEncoding.IsBinaryData(byte[])` and `IsBinaryData(byte[], int, int)`.
- **`IsTextualData` is now defined as `!IsBinaryData(...)`** (both overloads).
- **`IsBinaryData` algorithm:**
  1. **BOM fast-path** — if the inspected window starts with a UTF-8, UTF-16, or UTF-32 byte-order mark, the data is **textual**.
  2. **NUL scan** — otherwise, the data is **binary** if any `0x00` byte occurs within the first 8000 bytes. This is the same heuristic Git and GitHub apply.

## Why a BOM fast-path

A bare NUL scan is text/binary signal there is — ASCII, Latin-1, and UTF-8 never contain `0x00`, while essentially every real binary does, early. One structural blind spot: UTF-16/UTF-32, where `0x00` is *padding*, not signal.

The disambiguator there is the BOM, not content sniffing — BOM-less wide text is genuinely indistinguishable from some binaries by content alone.

```
| Encoding   | BOM (preamble)   | NUL in ASCII text? | Verdict        |
|------------|------------------|--------------------|----------------|
| UTF-16 LE  | FF FE            | yes (0x00 padding) | textual w/ BOM |
| UTF-16 BE  | FE FF            | yes                | textual w/ BOM |
| UTF-32 LE  | FF FE 00 00      | yes                | textual w/ BOM |
| UTF-32 BE  | 00 00 FE FF      | yes                | textual w/ BOM |
```
## Tests

- `FileEncoding_UnicodeDataIsTextual` now prepends each encoding's `GetPreamble()` so it asserts the realistic contract: **wide text with a BOM is textual**. (`GetBytes()` does not emit a preamble; only `GetPreamble()` / `StreamWriter` do.)
- Null / `ArgumentOutOfRangeException` guards preserved.
- Both call sites (`EnumeratedArtifact`, `ZipArchiveArtifact`) compile unchanged.
- All 7 `FileEncodingTests` pass.

---


michaelcfanning:
@ychollete your review here appreciated.

ycholette:
Just tested your changes and it seems to solve the issue, thanks!

michaelcfanning:
Thanks @ycholette! Going next into 5.0.5, shipping tomorrow AM latest.

PR #3004, opened 2026-06-11
AI rules: full AI1012 grammar + eliminate remaining silent evidence-catch swallows

michaelcfanning:
## What

This PR carries two related strands of AI-rule work that share resource,
designer, test, and doc surface:

### 1. AI1012/ProvideRuleSubId — full grammar validation
`AI1012` now validates the complete `AIRuleIdConvention` grammar rather than
only checking for a missing sub-id. It distinguishes:
- `Error_Malformed` — a non-`CWE`/`NOVEL` base, an uppercase or underscore
  sub-id, a `NOVEL-` id carrying a slash, or an empty sub-id.
- `Error_Missing` — a bare `CWE-<n>` (repairable by appending a sub-id).

This matches the convention the `emit` chain hard-enforces at authoring time,
so `SarifLogger` callers and hand-authored SARIF are held to the same grammar
at validation time. Id and name are unchanged.

### 2. Silent-catch sweep — eliminate the remaining swallowed evidence parses
The AI evidence rules each parsed `ai/evidence` inside a
`try { JArray.Parse(...) } catch { return; }`. A producer emitting a
present-but-malformed value (not a JSON array) got **zero feedback** — every
consumer silently dropped it. That is the general, silent, no-op catch our
doctrine forbids: never author a bare/`catch (Exception)` with no logging in
library code; prefer a single logging handler at the CLI boundary, or a typed
catch that takes a meaningful action.

Resolution — one rule owns the report, the rest skip cleanly:
- **`EvidenceJsonReader.TryParseEvidenceArray`** is the single parse chokepoint.
  It catches exactly `Newtonsoft.Json.JsonReaderException` (what `JArray.Parse`
  throws for any non-array input) and returns `false`; unrelated faults
  propagate to the engine's logging handler.
- **`AI2016/ProvideEvidenceBacking`** (Warning) **owns** the well-formedness
  report: on parse failure it logs the new `Warning_MalformedEvidence`.
- **`AI1010/ProvideEvidenceBackingUri`** (Error) skips a malformed value
  cleanly and defers to AI2016 (no double report). Its inner `JsonPointer`
  catch is narrowed from a bare `catch` to `catch (ArgumentException)` — the
  type the pointer ctor/`Evaluate` throw for a malformed or unresolvable
  pointer, which is exactly the violation this rule already reports.

This closes the gap where broken evidence JSON produced no diagnostic at all.

### 3. Doc slim — `AI-RuleId-Convention.md`
The §Enforcement section reproduced `AIRuleIdConventionException.BuildMessage`'s
stderr output verbatim — a hand-synced copy that can only drift. Replaced with
a one-line pointer to the source. The doc's unique value (TL;DR table, BNF
grammar, accept/reject oracle tables) is unchanged.

## Tests
- `AIValidationRuleTests`: 49 pass (+3 — AI2016 reports malformed evidence,
  AI1010 skips it, well-formed array produces no malformed report).
- `ValidateCommand` functional tests: 90 pass. AI2016 `_Invalid` baseline
  regenerated — only the `messageStrings` catalog grows by the new entry; no
  new result rows.

## Notes
The two strands were authored on the same branch and are entangled across
`RuleResources.resx/.Designer.cs`, `AIValidationRuleTests.cs`,
`ReleaseHistory.md`, and `AI-RuleId-Convention.md`. Folding them realizes
"resolve all the swallowed stuff in a single change" — this PR is the single
silent-catch sweep for the AI rules. Happy to split if a narrower review is
preferred.


docs/AI-RuleId-Convention.md
```diff
@@ -15,7 +15,7 @@ If an AI tool emits anything else — bare `CWE-89`, missing entirely, `cwe-89/f
 
 ## Why
 
-A SARIF result's sub-classifier — the part after `CWE-89/` — is where the producer records *which kind* of CWE-89 finding it emitted. The taxonomy entry is necessary but rarely sufficient, so AI-produced findings should name the sub-pattern they actually observed (`CWE-89/orm-string-interpolation`, not just `CWE-89`). The `AI1012` validation rule encodes exactly this expectation: any well-shaped AI finding either has a slash-bearing `ruleId` or extends its descriptor id with a slash-separated sub-id.
+A SARIF result's sub-classifier — the part after `CWE-89/` — is where the producer records *which kind* of CWE-89 finding it emitted. The taxonomy entry is necessary but rarely sufficient, so AI-produced findings should name the sub-pattern they actually observed (`CWE-89/orm-string-interpolation`, not just `CWE-89`). The `AI1012` validation rule is the validation-time counterpart to the emit-time enforcement: it holds `result.ruleId` to this same grammar, distinguishing a bare CWE base id (repairable by appending a sub-id) from a malformed id that no appended sub-id can rescue.
```
michaelcfanning: why are we maintaining this separate file?

PR #3009, opened 2026-06-12
Merge main into dev to true up after the v5.0.7 release

michaelcfanning:
Merge main back into dev after the v5.0.7 release (#3008), restoring main as a linear ancestor of dev.

The only delta merging in is the release stamp: the ReleaseHistory.md topmost section becomes ## **v5.0.7** (the five bullets are unchanged — the UNRELEASED header on dev was renamed in place) and src/build.props VersionPrefix is 5.0.7. The merge conflict was a pure header rename (UNRELEASED -> 5.0.7); resolved by keeping the shipped v5.0.7 section. After this merges, the dev tree is byte-identical to main, and the next feature PR recreates the UNRELEASED section above v5.0.7.


michaelcfanning:
Superseded: the true-up was landed as a real merge commit pushed directly to dev (1918d9a5a), matching every prior release true-up. A squash merge here would have flattened the merge commit and broken main's linear-ancestor relationship to dev.

PR #3130, opened 2026-07-07
Omit non-superset contextRegion in TypeScript constructMultilineContextSnippet

aasim:
The `@microsoft/sarif` port of `FileRegionsCache.constructMultilineContextSnippet` could emit a `contextRegion` that is not a proper superset of its `region`, tripping `SARIF1008` on pathologically long source lines. This ports the two guards the C# reference already has but the TS port had dropped.

## Root cause

Two divergences from `src/Sarif/FileRegionsCache.cs`:

| Variant | Trigger | C# (correct) | TS port (buggy) |
|---|---|---|---|
| identical | `region.charLength >= 512` (minified line) | omit | returns a clone byte-identical to `region` |
| non-superset | ±128/512-char window clamped short of a wide region | omit unless proper superset | returns the window unconditionally |

## Fix

In `npm/sarif/src/regions.ts`: return `undefined` in the `>= 512` branch, and add an `isProperSupersetOf` helper (a port of `Region.IsProperSupersetOf`) gating the char-offset-window return. `contextRegion` is optional (§3.29.5), so omitting it leaves `region` + snippet intact.

## Tests

Two cases in `regions.test.js` (one per variant) asserting a returned `contextRegion` strictly contains its `region` or is omitted. Both failed before, suite green now (41/41).

## Scope

TypeScript port only — the .NET `FileRegionsCache` was already correct and is untouched. No change to normal-length-line rendering.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -14,7 +14,7 @@ Each release entry below is prefixed with one of:
 Entries are terse by design: one line per change, present-tense behavior, complete but only essential data. No issue/PR archaeology or narrative — that history lives in the engineering system.
 
 ## **UNRELEASED**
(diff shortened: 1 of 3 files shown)


aasim:
Related bug: https://dev.azure.com/acs-mdash/acs-mdash/_workitems/edit/2565/

cfaucon (approved):

cfaucon (approved):

PR #3135, opened 2026-07-13
Reject path-traversal escapes in ArtifactLocation.TryReconstructAbsol…

aasim:
TryReconstructAbsoluteUri resolves a relative artifactLocation uri by concatenating it onto an originalUriBaseIds base. It never checked that the result stayed under that base, so a uri with ../ segments could resolve to an arbitrary path outside the base. When enrichment (insert / rewrite snippet and content population) reads that path, it copies file contents into the output SARIF.

This change validates containment. After the absolute uri is reconstructed, it must be at or beneath the base it resolved through (Uri.IsBaseOf). A reference that escapes its base returns false with a null resolvedUri, which is the method's existing "cannot reconstruct" result, so callers just skip enrichment for it.

The change is non-breaking. The absolute-uri passthrough at the top of the method is untouched, so SARIF with legitimate absolute paths is unaffected, and in-range .. that does not escape the base still resolves.

Reuses the same Uri.IsBaseOf containment check already used by RebaseUriVisitor and EmitFinalizeRebaseVisitor.

Added ArtifactLocationTests covering single, deep, and chained escapes (rejected), in-range .. (allowed), and confirmed the existing InsertOptionalDataVisitor and RebaseUri tests still pass.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -13,6 +13,9 @@ Each release entry below is prefixed with one of:
 
 Entries are terse by design: one line per change, present-tense behavior, complete but only essential data. No issue/PR archaeology or narrative — that history lives in the engineering system.
 
+## **UNRELEASED**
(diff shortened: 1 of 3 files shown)


michaelcfanning:
Nice fix.

cfaucon (approved):

PR #3136, opened 2026-07-15
Synthesize SarifLogger invocation only when the run has none

### LINKED ISSUE #2519: SarifLogger always add a new Invocation when SarifLogger is used
# The issue:

When I updated the sarif.sdk from 2.2.2 to 2.4.16, the SarifLogger changed its behavior and it always add a new invocation in the run.

https://github.com/microsoft/sarif-sdk/blob/acf13371915233ebaf1b99d00b7506e93baf3f97/src/Sarif/Writers/SarifLogger.cs#L227

# How to simulate:

Use the following snippet:
```
_run = new Run()
{
    Invocations = new List<Invocation>
    {
        Invocation.Create(
            emitMachineEnvironment: false,
            emitTimestamps: true,
            propertiesToLog: new[] { "commandLine", "workingDirectory" }),
    }, 
    VersionControlProvenance = new[] { versionControlDetails }
};

this.logger = new SarifLogger(
    streamWriter,
    run: _run,
    logFilePersistenceOptions: LogFilePersistenceOptions.PrettyPrint,
    levels,
    kinds);
```

# Current behavior:

The output SARIF will have two invocations: one that was created by the snippet and one that got created inside SarifLogger.

# Expected behavior:

Only add a new invocation if the `run.invocations` are null/empty. Otherwise, nothing should happen.

sukanth:
## Summary

`SarifLogger.EnhanceRun` unconditionally created a new `Invocation` and appended it to `run.Invocations` every time the logger enhanced a run. When a caller supplied a `Run` that already carried one or more `Invocation` objects, the logger appended a second, machine-synthesized invocation — duplicating invocation data the caller had already provided and producing a SARIF run with a spurious trailing invocation.

This change makes the synthesis conditional: `SarifLogger` now synthesizes and appends an `Invocation` only when `run.Invocations` is `null` or empty. When the caller already supplied at least one invocation, the logger leaves the caller's invocations untouched.

Fixes #2519.

## Details

`EnhanceRun` still normalizes `_run.Invocations` to a non-null list (`??= new List<Invocation>()`) because downstream code paths — `Dispose`/`AnalysisStopped` end-time stamping, the notification writers, SARIF v1 conversion, and `ResultProvenance.InvocationIndex` — read `Invocations[0]` / `Invocations.Count` and would `NullReferenceException` against a null list. Only the *creation* of a new invocation (plus its redaction and append) is now guarded by `if (_run.Invocations.Count == 0)`.

- The index-0 caller-supplied invocation is fully preserved, so end-time stamping, v1 conversion, and provenance indexing continue to operate on the caller's object.
- Redaction was never applied to caller-supplied invocations before this change, so scoping redaction to the synthesized invocation is not a behavioral regression.

## Tests

- `SarifLogger_DoesNotAddInvocationWhenRunAlreadyHasInvocation` — asserts no second invocation is appended when the caller supplies one.
- `SarifLogger_RetainsAllCallerSuppliedInvocations` — asserts multiple caller-supplied invocations are all retained and none are added.
- `SarifLogger_EnhancesRunWithInvocation` — refactored to share a private helper (`RunSarifLoggerOverRun`) with the new tests; still asserts the synthesized-invocation path for a run that starts with none.

Per repo convention, the new coverage is expressed as multiple `[Fact]` methods delegating to a shared helper (no `[Theory]`/`[InlineData]`). All `SarifLoggerTests` pass locally (33/33); the solution builds clean.

## Validation

The change was validated with a design review before implementation and an independent diff review afterward across two model families (OpenAI GPT-5.6 and Anthropic Claude Opus 4.8); both reached approve with no blocking findings.

### DIFF
--- a/src/Sarif/Writers/SarifLogger.cs
+++ b/src/Sarif/Writers/SarifLogger.cs
@@ -252,38 +252,44 @@ namespace Microsoft.CodeAnalysis.Sarif.Writers
                 }
             }
 
-            var invocation = Invocation.Create(
-                emitMachineEnvironment: dataToInsert.HasFlag(OptionallyEmittedData.EnvironmentVariables),
-                emitTimestamps: !dataToRemove.HasFlag(OptionallyEmittedData.NondeterministicProperties),
-                invocationPropertiesToLog);
-
-            // TODO we should actually redact across the complete log file context
-            // by a dedicated rewriting visitor or some other approach.
-
-            if (invocationTokensToRedact != null)
+            // Synthesize an invocation to capture command-line, machine, and timing
+            // data only when the caller has not already provided one. Adding one
+            // unconditionally would duplicate a caller-supplied invocation in the run.
+            if (_run.Invocations.Count == 0)
             {
-                invocation.CommandLine = Redact(invocation.CommandLine, invocationTokensToRedact);
-                invocation.Machine = Redact(invocation.Machine, invocationTokensToRedact);
-                invocation.Account = Redact(invocation.Account, invocationTokensToRedact);
+                var invocation = Invocation.Create(
+                    emitMachineEnvironment: dataToInsert.HasFlag(OptionallyEmittedData.EnvironmentVariables),
+                    emitTimestamps: !dataToRemove.HasFlag(OptionallyEmittedData.NondeterministicProperties),
+                    invocationPropertiesToLog);
 
-                if (invocation.WorkingDirectory != null)
+                // TODO we should actually redact across the complete log file context
+                // by a dedicated rewriting visitor or some other approach.
+
+                if (invocationTokensToRedact != null)
                 {
-                    invocation.WorkingDirectory.Uri = Redact(invocation.WorkingDirectory.Uri, invocationTokensToRedact);
-                }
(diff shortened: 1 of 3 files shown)


src/Sarif/Writers/SarifLogger.cs
```diff
@@ -252,38 +252,44 @@ private void EnhanceRun(IEnumerable<string> analysisTargets,
...
 
-                if (invocation.WorkingDirectory != null)
-                {
-                    invocation.WorkingDirectory.Uri = Redact(invocation.WorkingDirectory.Uri, invocationTokensToRedact);
-                }
+                // TODO we should actually redact across the complete log file context
+                // by a dedicated rewriting visitor or some other approach.
 
-                if (invocation.EnvironmentVariables != null)
+                if (invocationTokensToRedact != null)
```
michaelcfanning, line 268: You've created some API confusion here as now arguments such as 'invocationsToRedact' are only used in cases where an invocation is created by this code. So, if you provide a non-null Invocations list and also invocationTokensToRedact, this latter argument goes silently unused. There are two obvious possibilities, attempt to apply invocationsToRedact over all invocations, whether passed in or created *or* enforce that arguments such as invocationsToRedact are only provided when the incoming invocation list is null or empty.

PR #3145, opened 2026-07-28
Add RunAsync to MultithreadedAnalyzeCommandBase and serialize RuntimeErrors merges

eddynaka:
## Summary

Adds an async entry point to `MultithreadedAnalyzeCommandBase` and fixes a lost-update race on `RuntimeErrors`. Neither change alters the behavior of an existing successful scan.

## Async entry point

The class exposed only a synchronous `Run`, which built its channel pipeline and then blocked on the resulting tasks, so hosts already on an async stack had to occupy a thread for the length of the scan.

`RunAsync` returns `(int ExitCode, TContext GlobalContext)` rather than taking the context by `ref`, because an async method cannot declare a `ref` parameter. An overload without the context argument is provided for callers that discard it.

`Run` keeps its signature and continues to dispatch to `ValidateContext`, `AnalyzeTargets` and `PostLogFile`, so **subclasses overriding those virtuals are unaffected**. That is why this is not a sync-over-async shim, which would silently change which overrides get invoked for downstream tools. `RunAsync` dispatches to new `*Async` virtuals holding the real work, and the synchronous virtuals delegate to them wherever `Run` already blocked on a task, so the two paths share one implementation rather than drifting.

New awaits use `ConfigureAwait(false)`, since the synchronous wrappers block on those tasks and capturing a calling `SynchronizationContext` would deadlock a UI or ASP.NET host.

## The `RuntimeErrors` race

`ScanTargetsAsync` merged each target's `RuntimeErrors` into the global context with an unsynchronized `|=`. Every scan worker performs that read-modify-write concurrently, so two workers reading the same value lose one of the two merges. `LogResultsAsync` re-merges the same per-file contexts under `lock (globalContext)`, which restores most dropped flags and is why this has not surfaced as a reported defect, but flags raised against a context that never reaches the consumer are lost outright. The merge now takes the same lock the consumer holds, as does the merge on the archive-open failure path.

`MultithreadedAnalyzeCommandBaseConcurrencyTests` covers `Run` and `RunAsync`. The tests drive a genuine race: `MergeProbingContext` detects a write that clears a flag its writer never observed, and `MergeProbingCommand` tags each target with a distinct condition so a dropped merge is observable at all, a merge of `RuntimeConditions.None` being idempotent. `MergeProbingContext` lives in `Test.Utilities.Sarif` because `ExportConfigurationCommandBaseTests` counts MEF convention-discovered `IOptionsProvider` exports in its own assembly, and `AnalyzeContextBase` implements `IOptionsProvider`.

Also renames `readyToScanChannel` to `_readyToScanChannel` to match the surrounding fields, and drops a duplicate semicolon, an empty statement and a null check on a value dereferenced two lines earlier.

## Validation

`Test.UnitTests.Sarif.Driver`: 162 passed, 1 skipped, 0 failed.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -16,6 +16,7 @@ Entries are terse by design: one line per change, present-tense behavior, comple
 ## **UNRELEASED**
 * NEW: `MultithreadedAnalyzeCommandBase.RunAsync` analyzes without blocking the caller, dispatching to new async virtuals that hold the work; `Run` keeps its signature and dispatches to their synchronous counterparts, so existing subclasses are unaffected.
 * BUG: `ArtifactLocation.TryReconstructAbsoluteUri` returns false (leaving `resolvedUri` null) when a relative `uri`'s `../` segments escape the `originalUriBaseIds` base it resolves through, so enrichment no longer reads files outside a declared base.
+* BUG: `MultithreadedAnalyzeCommandBase` merges per-target `RuntimeErrors` into the global context under a lock, so concurrent scan workers no longer lose each other's flags.
 
 ## **v5.5.0** [Sdk](https://www.nuget.org/packages/Sarif.Sdk/v5.5.0) | [Driver](https://www.nuget.org/packages/Sarif.Driver/v5.5.0) | [Converters](https://www.nuget.org/packages/Sarif.Converters/v5.5.0) | [Multitool](https://www.nuget.org/packages/Sarif.Multitool/v5.5.0) | [Multitool Library](https://www.nuget.org/packages/Sarif.Multitool.Library/v5.5.0)
 * BUG: `@microsoft/sarif`'s `FileRegionsCache.constructMultilineContextSnippet` omits `contextRegion` when the region meets the 512-char cap or the window is not a proper superset of `region`, so long lines no longer emit SARIF that `SARIF1008.PhysicalLocationPropertiesMustBeConsistent` rejects.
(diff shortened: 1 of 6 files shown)


ReleaseHistory.md
```diff
@@ -14,7 +14,9 @@ Each release entry below is prefixed with one of:
 Entries are terse by design: one line per change, present-tense behavior, complete but only essential data. No issue/PR archaeology or narrative — that history lives in the engineering system.
 
 ## **UNRELEASED**
```
michaelcfanning, line 16: this release note is not added in the proper order, all `NEW` items follow all `BUG` items. even AI can't get this right smh. :)

src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs
```diff
@@ -168,6 +136,115 @@ public virtual int Run(TOptions options, ref TContext globalContext)
                     : FAILURE;
         }
 
+        /// <summary>
+        /// Runs analysis without blocking the calling thread. Unlike <see cref="Run(TOptions)"/>,
+        /// no stage of this pipeline waits on a task: target enumeration, scanning and result
+        /// logging are awaited, as are the post-uri health check and the log file post.
+        /// </summary>
+        public virtual async Task<int> RunAsync(TOptions options)
```
michaelcfanning: why are we introducing two new virtual methods? let's just add one.

eddynaka: Copilot followed the same methods exist for sync. Just removed them.

ReleaseHistory.md
eddynaka: Yup. Manually fixed.

michaelcfanning (approved):

aasim (approved):

